PR #23, opened 2015-08-28
Add ES2015 support to CLI

### LINKED ISSUE #20: Built-in ES2015 support
Would be nice to support ES6 built-in at runtime by using the [Babel require hook](https://babeljs.io/docs/using-babel/#babel_register).

That way you can just write your tests in ES6 and have it run in any node version.

This is so nice:

```js
test(t => {
  t.ok();
  t.end();
});
```

andrepolischuk:
Fix #20


andrepolischuk:
I have no idea why execFile failed in travis

test/test.js
```diff
@@ -343,3 +344,15 @@ test('promise support - reject', function (t) {
 		t.end();
 	});
 });
+
+test('es2015 support', function (t) {
+	t.plan(3);
+
+	execFile('./cli.js', ['test/es2015.js'], {
+		cwd: __dirname
```
sindresorhus: `cwd: path.resolve(__dirname, '..')`

test/test.js
```diff
@@ -343,3 +344,15 @@ test('promise support - reject', function (t) {
 		t.end();
 	});
 });
+
+test('es2015 support', function (t) {
+	t.plan(3);
+
+	execFile('./cli.js', ['test/es2015.js'], {
```
sindresorhus: `execFile('../cli.js', ['es2015.js'], {`

test/test.js
```diff
@@ -343,3 +344,15 @@ test('promise support - reject', function (t) {
 		t.end();
 	});
 });
+
+test('es2015 support', function (t) {
```
sindresorhus: ES2015

test/test.js
```diff
@@ -343,3 +344,15 @@ test('promise support - reject', function (t) {
...
+	t.plan(3);
+
+	execFile('./cli.js', ['test/es2015.js'], {
+		cwd: __dirname
+	}, function (err, stdout, stderr) {
+		t.assert(!err, err);
+		t.assert(stdout.trim().length);
+		t.assert(!stderr.trim().length);
```
sindresorhus: This check is moot and can be removed.

test/es2015.js
```diff
@@ -0,0 +1,12 @@
+import test from 'tape';
+import ava from '../lib/test';
+
+test('run test', t => {
+	ava('foo', a => {
+		a.true(false);
```
sindresorhus: `a.pass();` instead

package.json
```diff
@@ -22,7 +22,7 @@
     "node": ">=0.10.0"
   },
   "scripts": {
-    "test": "xo && node test/test.js | tap-dot"
+    "test": "xo test/test.js && node test/test.js | tap-dot"
```
sindresorhus: Don't change this. XO should test all JS files.

cli.js
```diff
@@ -5,6 +5,7 @@ var path = require('path');
 var globby = require('globby');
 var meow = require('meow');
 var updateNotifier = require('update-notifier');
+require('babel/register');
```
sindresorhus: I think we should try to load the local `babel` before using the built-in one. So the user can provide their own.

You can use https://github.com/sindresorhus/resolve-from to achieve this.

sindresorhus:
Awesome! Thanks for working on this :)

Would you mind adding something about it to the readme too?

test/test.js
andrepolischuk: You mean that I should remove this test? Or rewrite?

package.json
andrepolischuk: I changed this because XO without --esnext failed on checking test/es2015.js

sindresorhus:
Why did you remove the test? Can you fix the merge conflict?

cli.js
```diff
@@ -4,8 +4,15 @@ var fs = require('fs');
 var path = require('path');
 var globby = require('globby');
 var meow = require('meow');
+var resolveFrom = require('resolve-from');
 var updateNotifier = require('update-notifier');
 
+try {
+	require(resolveFrom('.', 'babel/register'));
+} catch (e) {
```
sindresorhus: `} catch (err) {`

package.json
```diff
@@ -46,6 +46,7 @@
   ],
   "dependencies": {
     "async-each-series": "^1.0.0",
+    "babel": "^5.8.23",
```
sindresorhus: Use `babel-core` instead. It doesn't include the Babel CLI junk.

readme.md
```diff
@@ -165,6 +165,54 @@ test.serial(function (t) {
...
+```
+
+
+### ES2015 support
+
+You can write your tests in ES2015:
+
+```js
+test(t => {
+	t.ok();
```
sindresorhus: `t.pass();`

andrepolischuk:
I will fix

andrepolischuk:
@sindresorhus I fixed conflicts

sindresorhus:
Can you add back the test? We need a test to make sure it's working ok.

cli.js
```diff
@@ -4,8 +4,15 @@ var fs = require('fs');
 var path = require('path');
 var globby = require('globby');
 var meow = require('meow');
+var resolveFrom = require('resolve-from');
 var updateNotifier = require('update-notifier');
 
+try {
+	require(resolveFrom('.', 'babel/register'));
```
sindresorhus: This should also be `babel-core`.

sindresorhus: Actually, should probably do both, in case users use plain `babel`.

```
require(resolveFrom('.', 'babel-core/register') || resolveFrom('.', 'babel/register'));
```

sindresorhus:
Looks good. Thank you @andrepolischuk :)

andrepolischuk:
Thank you for review and comments

sindresorhus:
:beers:

PR #29, opened 2015-08-29
Show only the useful lines of stack trace - fixes #22

### LINKED ISSUE #22: Error context
Tape has this nice output on errors:

```
at: Test.<anonymous> (/Users/sindresorhus/dev/hypot/test.js:9:4)
```

Currently we output the full stack trace or none.

-

This might be something `claim` should handle.

https://github.com/substack/tape/blob/ccfcf4f76581abc3842163dfe065198764247c85/lib/test.js#L207

mdibaiee:
Fix "Assertion count does not match planned" error's stack trace not showing

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,6 +3,7 @@ var chalk = require('chalk');
 var figures = require('figures');
 var Squeak = require('squeak');
 var plur = require('plur');
+var path = require('path');
 var Runner = require('./lib/runner');
 var log = new Squeak({separator: ' '});
 var runner = new Runner();
@@ -38,8 +39,14 @@ function stack(results) {
 
 		// Don't print the full stack but the only useful line showing
 		// the actual test file stack
+		var dir = path.dirname(module.parent.filename);
+
 		var split = (result.error.stack || '').split('\n');
-		var beautiful = result.error.message + '\n' + (split[2] || '');
+		var related = split.filter(function (line) {
+			return line.indexOf(dir) > -1;
+		});
+
+		var beautiful = result.error.message + '\n' + related.join('\n');
 		result.error.stack = beautiful;
 
 		log.writelpad(chalk.red(i + '.', result.title));
--- a/index.js
+++ b/index.js
@@ -36,6 +36,13 @@ function stack(results) {
 
 		i++;
 
+
+		// Don't print the full stack but the only useful line showing
+		// the actual test file stack
+		var split = (result.error.stack || '').split('\n');
+		var beautiful = result.error.message + '\n' + (split[2] || '');
+		result.error.stack = beautiful;
+
 		log.writelpad(chalk.red(i + '.', result.title));
 		log.writelpad(chalk.red(result.error.stack));
 		log.write();
--- a/index.js
+++ b/index.js
@@ -36,7 +36,6 @@ function stack(results) {
 
 		i++;
 
-
 		// Don't print the full stack but the only useful line showing
 		// the actual test file stack
 		var split = (result.error.stack || '').split('\n');
--- a/index.js
+++ b/index.js
@@ -3,7 +3,6 @@ var chalk = require('chalk');
 var figures = require('figures');
 var Squeak = require('squeak');
 var plur = require('plur');
-var path = require('path');
 var Runner = require('./lib/runner');
 var log = new Squeak({separator: ' '});
 var runner = new Runner();
@@ -39,15 +38,16 @@ function stack(results) {
 
 		// Don't print the full stack but the only useful line showing
 		// the actual test file stack
-		var dir = path.dirname(module.parent.filename);
-
-		var split = (result.error.stack || '').split('\n');
-		var related = split.filter(function (line) {
-			return line.indexOf(dir) > -1;
-		});
-
-		var beautiful = result.error.message + '\n' + related.join('\n');
-		result.error.stack = beautiful;
+		var reg = /\((?:[\\\/](?:(?!node_modules[\\\/]ava[\\\/])[^:\\\/])+)+:\d+:\d+\)/;
+		function beautifulStack(stack) {
+			var found = false;
+			return stack.split('\n').filter(function (line) {
+				var relevant = reg.test(line);
+				found = found || relevant;
+				return !found || relevant;
+			}).join('\n');
+		}
+		result.error.stack = beautifulStack(result.error.stack);
 
 		log.writelpad(chalk.red(i + '.', result.title));
 		log.writelpad(chalk.red(result.error.stack));
--- a/index.js
+++ b/index.js
@@ -7,6 +7,8 @@ var Runner = require('./lib/runner');
 var log = new Squeak({separator: ' '});
 var runner = new Runner();
 
+Error.stackTraceLimit = Infinity;
+
 log.type('success', {
 	color: 'green',
 	prefix: figures.tick
@@ -38,7 +40,7 @@ function stack(results) {
 
 		// Don't print the full stack but the only useful line showing
 		// the actual test file stack
-		var reg = /\((?:[\\\/](?:(?!node_modules[\\\/]ava[\\\/])[^:\\\/])+)+:\d+:\d+\)/;
+		var reg = /(?:^(?! {4}at\b).{6})|(?:\((?:[\\\/](?:(?!node_modules[\\\/]ava[\\\/])[^:\\\/])+)+:\d+:\d+\))/;
 		function beautifulStack(stack) {
 			var found = false;
 			return stack.split('\n').filter(function (line) {
@@ -47,10 +49,10 @@ function stack(results) {
 				return !found || relevant;
 			}).join('\n');
 		}
-		result.error.stack = beautifulStack(result.error.stack);
+		var stack = beautifulStack(result.error.stack);
 
 		log.writelpad(chalk.red(i + '.', result.title));
-		log.writelpad(chalk.red(result.error.stack));
+		log.writelpad(chalk.red(stack));
 		log.write();
 	});
 }
--- a/lib/test.js
+++ b/lib/test.js
@@ -101,7 +101,6 @@ Test.prototype.end = function () {
 
 Test.prototype.exit = function () {
 	if (this.planCount !== null && this.planCount !== this.assertCount) {
-		console.log(this.planStack);
 		this.assertError = new assert.AssertionError({
 			actual: this.assertCount,
 			expected: this.planCount,
--- a/lib/test.js
+++ b/lib/test.js
@@ -54,6 +54,9 @@ Test.prototype.plan = function (count) {
 	}
 
 	this.planCount = count;
+	// in case the planCount doesnt match assertCount, we need the stack of this
+	// function to throw with a useful stack
+	this.planStack = new Error().stack;
 };
 
 Test.prototype.skip = function () {
@@ -75,8 +78,7 @@ Test.prototype.run = function (cb) {
 				this.assertError = new assert.AssertionError({
 					actual: err,
 					message: 'Promise rejected → ' + err,
-					operator: 'promise',
-					stackStartFunction: this
+					operator: 'promise'
 				});
 
 				this.exit();
@@ -99,14 +101,15 @@ Test.prototype.end = function () {
 
 Test.prototype.exit = function () {
 	if (this.planCount !== null && this.planCount !== this.assertCount) {
+		console.log(this.planStack);
 		this.assertError = new assert.AssertionError({
 			actual: this.assertCount,
 			expected: this.planCount,
 			message: 'Assertion count does not match planned',
-			operator: 'plan',
-			// TODO: find out why it doesn't show any stack
-			stackStartFunction: this.fn
+			operator: 'plan'
 		});
+
+		this.assertError.stack = this.planStack;
 	}
 
 	if (!this.ended) {


mdibaiee:
About promises, I couldn't find a way of finding the stack trace for them, and as it's self-explainatory that the the _returned promise_ was rejected, I think we can just ignore printing and stack trace for that.

Qix-:
Without looking too deeply at the code, is there an option to show full stacks? This is one thing about Mocha I actually hate; a lot of the time I'm writing two modules that complement each other concurrently, and the stack traces never show the other's stack entries unless I use `--full-trace`.

sindresorhus:
@Qix- What would you like to keep?

Currently it looks like:

```
  AssertionError: false === true
    at Test.(anonymous function) [as assert] (/Users/sindresorhus/dev/ava/lib/test.js:49:14)
    at Test.fn (/Users/sindresorhus/dev/acosh/test.js:8:4)
    at Test.run (/Users/sindresorhus/dev/ava/lib/test.js:78:18)
    at Runner.<anonymous> (/Users/sindresorhus/dev/ava/lib/runner.js:39:8)
    at Immediate.immediate._onImmediate (timers.js:418:16)
    at processImmediate [as _immediateCallback] (timers.js:367:17)
```

With this PR, it looks like:

```
false === true
    at Test.fn (/Users/sindresorhus/dev/acosh/test.js:8:4)
```

All it hides is useless stack info coming from AVA and Node.js core libs.

lib/test.js
```diff
@@ -99,14 +101,15 @@ Test.prototype.end = function () {
 
 Test.prototype.exit = function () {
 	if (this.planCount !== null && this.planCount !== this.assertCount) {
+		console.log(this.planStack);
```
sindresorhus: Leftover

Qix-:
I'd have to see a full stack. If it's just removing lines caused by the tests then that should be just fine.

sindresorhus:
Can you provide some code that result in a "full stack"? I definitely agree it should not remove that.

Qix-:
Well more specifically with Mocha in general, it [removes any stack entries that have `node_modules` in them](https://github.com/mochajs/mocha/blob/2f458ab38cd191c4071742caa3d101a96a832c2b/lib/utils.js#L684-L740). I suppose it makes sense, but at the same time it's somewhat futile and time-wasting because, in the case of errors with suppressed `node_modules` stack lines, the error message doesn't normally correspond with the actual error site at the "top" of the trace, causing confusion.

mdibaiee:
I see that this approach doesn't work in some cases, consider this example:

```javascript
test('something', function(t) {
  t.is('hey', 'hey');
  doSomeTest(t, 2);
  doSomeTest(t, {});
  t.end();
});

function doSomeTest(t, x) {
  t.same({x: x}, {x: {}});
}
```

This is the output for this example:
```
1. something
  { x: 2 } === { x: {} }
    at doSomeTest (/Users/mahdi/Documents/Workshop/github/ava-test/test.js:13:5)
```
But we should also show the line for:
```
at Test.fn (/Users/mahdi/Documents/Workshop/github/ava-test/test.js:9:5)
```

I'm going to see if I can find a better approach.

Qix-:
Not quite, but that is true. For instance,

```javascript
// mocha

// Note, `my-module' is not part of the source tree, but it's still mine
// in this example. Maybe a sister/plugin module, etc.
var myModule = require('my-module');

it('should xyz', function () {
  myModule(); // throws
});
```

stack:
```
Error: some sort of error
    at function (test.js:8:4)
    at <mocha stuff>
```

It doesn't show where inside the module `my-module` the error was thrown. I have to pass `--full-trace` in order to get that information.

mdibaiee:
Okay, I think I found the ultimate solution ✊

We can filter out the stack lines which start with the directory of the test file, see:

```javascript
let dir = path.dirname(module.parent.filename);

var split = (result.error.stack || '').split('\n');
let related = split.filter(line => line.indexOf(dir) > -1);

var beautiful = result.error.message + '\n' + related.join('\n');
```

Qix-:
Then assertions won't show up correctly. :tangerine:

mdibaiee:
Why? I tested it with the example I said above, it's working correctly, have you tested this?

Output:
```
0:50 » ~/Documents/Workshop/github/ava-test $ node test.js                                                                                                                                            ‹1›
  ✖ something { x: 2 } === { x: {} }

  1 test failed

  1. something
  { x: 2 } === { x: {} }
    at doSomeTest (/Users/mahdi/Documents/Workshop/github/ava-test/test.js:13:5)
    at Test.fn (/Users/mahdi/Documents/Workshop/github/ava-test/test.js:7:3)
```

mdibaiee:
@sindresorhus Do you think I should squash my commits into one?

sindresorhus:
@mdibaiee No need. I squash one merge. Easier to see individual changes this way.

mdibaiee:
What's the plan now? Anything else has to be done in this PR?

sindresorhus:
Looks good to me now :)

@Qix- ?

index.js
```diff
@@ -36,6 +37,18 @@ function stack(results) {
 
 		i++;
 
+		// Don't print the full stack but the only useful line showing
+		// the actual test file stack
+		var dir = path.dirname(module.parent.filename);
```
Qix-: Sorry for my thickness, but what is this *actually* doing? Isn't it just filtering out anything that's not in the project directory? What is this improving/fixing?

mdibaiee: This is fixing [this](https://github.com/sindresorhus/ava/issues/22).

Qix-:
I understand the desire for clean stack traces. I just heavily question the usefulness.

Tape's stack errors are even more useless than Mocha's if all they're showing is the relevant call site of the error. Many times, the parts of the stack that make the error's cause **immediately clear** are the parts above (deeper) the deepest user script. For example:

```
TypeError: Cannot convert object to primitive value
    at String (unknown source)
    at util.js:38:25
    at String.replace (native)
    at Object.<anonymous> (util.js:35:23)
    at Object.<anonymous> (console.js:25:36)
    at EventEmitter.<anonymous> (/project/src/routines/debug/boot.js:16:21)
```

Without looking at `boot.js` in the above project, I can safely assume the affected line is probably not clear what exactly is going on. Take a look.

```javascript
'use strict';

var path = require('path');

var message = 'TypeError: Cannot convert object to primitive value';
var stack = [
  '    at String (unknown source)',
  '    at util.js:38:25',
  '    at String.replace (native)',
  '    at Object.<anonymous> (util.js:35:23)',
  '    at Object.<anonymous> (console.js:25:36)',
  '    at EventEmitter.<anonymous> (' + __filename + ':16:21)',
].join('\n');

var dir = path.dirname(module.filename);
console.log('DIR', dir);

var split = (stack || '').split('\n');
var related = split.filter(function (line) {
	return line.indexOf(dir) > -1;
});

var beautiful = message + '\n' + related.join('\n');

console.log(beautiful);
```

which outputs

```
$ node test.js
DIR /private/tmp/avatest
TypeError: Cannot convert object to primitive value
    at EventEmitter.<anonymous> (/private/tmp/avatest/test.js:16:21)
```

I'd have no idea what was going on, making the stack trace **absolutely useless** - hence why I asked if there was a way to show full traces.

---

If the intent here is to only show relevant stack lines, so that we exclude:

- Lines that are part of Node.js proper
- Lines that belong to the AVA module

then why not filter on lines that have absolute paths, starting at the first found? The logic here is that, beginning at the deepest source file (going upward through the stack), we filter out anything that isn't a user script. We also show anything deeper than the deepest user script.

This is an approach I've used elsewhere for clean error printing and I've never run into problems with not seeing enough information.

```regex
\((?:\/(?:(?!node_modules[\/\\]ava[\/\\])[^:\/])+)+:\d+:\d+\)
```

Example:

```javascript
'use strict';

var path = require('path');

var message = 'TypeError: Cannot convert object to primitive value';
var stack = [
  '    at String (unknown source)',
  '    at util.js:38:25',
  '    at String.replace (native)',
  '    at Object.<anonymous> (util.js:35:23)',
  '    at Object.<anonymous> (console.js:25:36)',
  '    at EventEmitter.<anonymous> (' + __filename + ':16:21)',
  '    at Object.<anonymous> (console.js:25:36)',
  '    at Object.<anonymous> (console.js:25:36)',
  '    at Object.<anonymous> (console.js:25:36)',
  '    at Object.<anonymous> (console.js:25:36)',
  '    at EventEmitter.<anonymous> (' + __filename + ':16:21)',
  '    at EventEmitter.<anonymous> (' + __filename + ':16:21)',
  '    at Object.<anonymous> (console.js:25:36)',
  '    at Object.<anonymous> (console.js:25:36)',
  '    at Object.<anonymous> (console.js:25:36)'
].join('\n');

var reg = /\((?:\/(?:(?!node_modules[\/\\]ava[\/\\])[^:\/])+)+:\d+:\d+\)/;
function beautifulStack(stack) {
  var found = false;
  return stack.split('\n').filter(function (line) {
    var relevant = reg.test(line);
    found = found || relevant;
    return !found || relevant;
  }).join('\n');
}

console.log(message + '\n' + beautifulStack(stack));
```

which yields

```
$ node test.js
TypeError: Cannot convert object to primitive value
    at String (unknown source)
    at util.js:38:25
    at String.replace (native)
    at Object.<anonymous> (util.js:35:23)
    at Object.<anonymous> (console.js:25:36)
    at EventEmitter.<anonymous> (/private/tmp/avatest/test.js:16:21)
    at EventEmitter.<anonymous> (/private/tmp/avatest/test.js:16:21)
    at EventEmitter.<anonymous> (/private/tmp/avatest/test.js:16:21)
```

For @sindresorhus's stack example, this yields

```
$ node test.js
AssertionError: false === true
    at Test.(anonymous function) [as assert] (/Users/sindresorhus/dev/ava/lib/test.js:49:14)
    at Test.fn (/Users/sindresorhus/dev/acosh/test.js:8:4)
    at Test.run (/Users/sindresorhus/dev/ava/lib/test.js:78:18)
    at Runner.<anonymous> (/Users/sindresorhus/dev/ava/lib/runner.js:39:8)
```

---

Another example, [this stacktrace](http://stackoverflow.com/a/22726190/510036):

```
Error: `foo` has been removed in favorof `bar`
    at function() (/home/ray/dev/test/error.js:53:3)
    at b() (domain.js:183:18)
    at Domain.run() (domain.js:123:23)
    at function() (/home/ray/dev/test/error.js:52:3)
    at Module._compile() (module.js:456:26)
    at Module._extensions..js() (module.js:474:10)
    at Module.load() (module.js:356:32)
    at Module._load() (module.js:312:12)
    at Module.runMain() (module.js:497:10)
```

gets reduced down to

```
$ node test.js
Error: `foo` has been removed in favorof `bar`
    at function() (/home/ray/dev/test/error.js:53:3)
    at function() (/home/ray/dev/test/error.js:52:3)
```

That seems much more appropriate to me.

mdibaiee:
@Qix- Can you give me the actual test that produces this stack?

```
at String (unknown source)',
  '    at util.js:38:25',
  '    at String.replace (native)',
  '    at Object.<anonymous> (util.js:35:23)',
  '    at Object.<anonymous> (console.js:25:36)',
  '    at EventEmitter.<anonymous> (' + __filename + ':16:21)',
```

Qix-:
@mdibaiee No, it's an example stack trace I found on SO. My point is is that you don't need to look at the code to know that the first 5 frames are important for debugging the error.

mdibaiee:
@Qix- My point is knowing what *your fault* was in the error is good enough for debugging most codes, and *starting* debugging of others.

Qix-:
@mdibaiee aren't tests a vital part of debugging code?

mdibaiee:
@Qix- You know the answer, yes. And how does that relate?

Qix-:
Because a first glance view of all of the *relevant* parts of an error are important.

mdibaiee:
I don't think so. I don't understand your worries, almost all commands have a `--verbose` option, do you know why? Because I don't need to know all the details everytime, I don't want `ls` to be like `ls -l` everytime, it takes spaces, it takes my time and my attention.

If you really think all relevant paths are important, `alias ava='ava --full-trace'`. Problem solved. I'm not continuing this discussion.

Qix-:
Which is what I requested to begin with, however

```
$ grep -r "full-trace" *
``

You don't have the option in there.

**Comment by Qix-** [comment]
I'd also recommend setting infinity length stack traces before filtering.

```javascript
Error.stackTraceLimit = Infinity;
```

sindresorhus:
Minimal vs verbose is a hard one, but from experience with debugging a lot of failed assertions I have to agree with @Qix-. I often have to digg deeper than just the test file. I'm willing to sacrifice visual beauty in favor of practicality here. We should preserve any stack info referencing user code as @Qix- has outlined above. I'm allergic to options and I'd rather go with a good default. We can always change it in the future if it turns out to be the wrong choice.

@mdibaiee Would you be willing to update your PR even though you seem to strongly disagree?

> Error.stackTraceLimit = Infinity;

:+1:

Qix-:
One last thing (I know I'm persistent) but since the introduction of [`error-ex`](/Qix-/node-error-ex) we might run into exceptions with non-standard lines that start with something other than `at`; we shouldn't filter those out either.

sindresorhus:
> but since the introduction of error-ex we might run into exceptions with non-standard lines that start with something other than at; we shouldn't filter those out either.

:+1: I've seen other error modules add stuff too. Better to be on the safe side.

mdibaiee:
@sindresorhus So what's the plan now? Don't filter anything out? Use @Qix- 's RegExp?

sindresorhus:
> Use @Qix- 's RegExp?

Yes

mdibaiee:
@sindresorhus Alright, sorry for the delay, I was busy working on another projet, are we ready to go?

index.js
```diff
@@ -36,6 +36,19 @@ function stack(results) {
 
 		i++;
 
+		// Don't print the full stack but the only useful line showing
+		// the actual test file stack
+		var reg = /\((?:[\\\/](?:(?!node_modules[\\\/]ava[\\\/])[^:\\\/])+)+:\d+:\d+\)/;
```
Qix-: Mind updating the regex?

```
(?:^(?! {4}at).{6})|(?:\((?:[\\\/](?:(?!node_modules[\\\/]ava[\\\/])[^:\\\/])+)+:\d+:\d+\))
```

This won't kill the error message and won't filter out lines that don't start with `at`.

index.js
```diff
@@ -36,6 +36,19 @@ function stack(results) {
...
+		function beautifulStack(stack) {
+			var found = false;
+			return stack.split('\n').filter(function (line) {
+				var relevant = reg.test(line);
+				found = found || relevant;
+				return !found || relevant;
+			}).join('\n');
+		}
+		result.error.stack = beautifulStack(result.error.stack);
```
Qix-: Something makes me feel iffy about re-assigning it back to the error itself. Thoughts @sindresorhus?

sindresorhus: Agreed. Just use it inline below:

```js
log.writelpad(chalk.red(beautifulStack(result.error.stack)));
```

mdibaiee:
@sindresorhus done

sindresorhus:
Landed! Thanks for the PR @mdibaiee. Keep'em coming ;)

And thanks for the review @Qix-.

Qix-:
No problem! Thanks @mdibaiee :dancer: :+1:

mdibaiee:
✊✌

Issue #14, opened 2014-12-06
Execution time


kevva:
Add execution time for each test.

PR #30, opened 2015-09-01
Output test execution time

vdemedes:
Hello Sindre,

This pull request closes issue #14. It measures each test's total execution time (duration) using `process.hrtime()` and then outputs nicely formatted time near test's title:

screen shot 2015-09-01 at 12 27 17 pm

[pretty-hrtime](https://npmjs.org/package/pretty-hrtime) module is used for converting `process.hrtime()` result to a human-readable form.

Thanks for a great project, will be switching off mocha to ava soon.

### DIFF
--- a/index.js
+++ b/index.js
@@ -31,7 +31,9 @@ function test(err, title, duration) {
 	var threshold = 100;
 
 	if (duration > threshold) {
-		var formattedDuration = prettyMs(duration, { secDecimalDigits: 2 });
+		var formattedDuration = prettyMs(duration, {
+			secDecimalDigits: 2
+		});
 
 		timeSpent = chalk.gray.dim(' (' + formattedDuration + ')');
 	}
--- a/index.js
+++ b/index.js
@@ -31,7 +31,9 @@ function test(err, title, duration) {
 	var threshold = 100;
 
 	if (duration > threshold) {
-		timeSpent = chalk.gray.dim(' (' + prettyMs(duration) + ')');
+		var formattedDuration = prettyMs(duration, { secDecimalDigits: 2 });
+
+		timeSpent = chalk.gray.dim(' (' + formattedDuration + ')');
 	}
 
 	log.success(title + timeSpent);
--- a/index.js
+++ b/index.js
@@ -24,7 +24,7 @@ function test(err, title, duration) {
 		return;
 	}
 
-	log.success(title + ' (' + hrtime(duration) + ')');
+	log.success(title + ' ' + chalk.dim('(' + hrtime(duration) + ')'));
 }
 
 function stack(results) {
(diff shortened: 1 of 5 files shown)


Qix-:
:+1: Neat. Code looks good to me.

Could there be a way to optionally specify that I want to always show *at least* millisecond output?

index.js
```diff
@@ -17,13 +18,13 @@ log.type('error', {
...
 
-function test(err, title) {
+function test(err, title, duration) {
 	if (err) {
 		log.error(title, chalk.red(err.message));
 		return;
 	}
 
-	log.success(title);
+	log.success(title + ' (' + hrtime(duration) + ')');
```
sindresorhus:
```js
log.success(title + ' ' + chalk.dim('(' + hrtime(duration) + ')'));
```

sindresorhus:
Was thinking about doing something like this, but instead only show the time when it passes some kind of threshold, as the number is really only useful for identifying slow tests. Thoughts?

Qix-:
That's how mocha does it, and it makes sense. I don't care if a test took 200us. I do care if it took 200ms.

index.js
vdemedes: Pushed an update.

sindresorhus:
If so, what kind of threshold makes sense? Not interested in adding an option to change it, so it needs to cover most use-cases.

vdemedes:
Added code to display test duration only when a threshold (100ms) is reached. If you want, you can easily change it here: https://github.com/sindresorhus/ava/commit/2791234f08d8d474ae7558522e366b253a741443#diff-168726dbe96b3ce427e7fedce31bb0bcR31

sindresorhus:
Looks very good @vdemedes! Can you add a test?

sindresorhus:
Here's how it looks:

screen shot 2015-09-01 at 17 49 55

index.js
```diff
@@ -17,13 +18,23 @@ log.type('error', {
...
 
-	log.success(title);
+	// format a time spent in the test
+	var timeSpent = '';
+
+	// display duration only over a threshold
+	var threshold = 100;
+
+	if (duration > threshold) {
+		timeSpent = chalk.dim('(' + prettyMs(duration) + ')');
```
sindresorhus: Just remembered that `dim` doesn't work everywhere. Can you make it `chalk.gray.dim`?

sindresorhus: And the below `' '` should be here instead.

vdemedes: Sure, going to add that now.

vdemedes: Pushed an update.

vdemedes:
@sindresorhus of course, will add a test for sure

vdemedes:
@sindresorhus Added a test for measuring duration, let me know what you think!

sindresorhus:
Excellent work @vdemedes. Keep'em coming! ;)

More than happy to receive addition suggestion on how AVA can be better in form of issues or PRs.

vdemedes:
Thank you @sindresorhus, will definitely contribute more!

Issue #38, opened 2015-09-06
Use promises


sindresorhus:
We do a lot of async stuff in AVA. Would be cleaner by using promises instead.

For example here: https://github.com/sindresorhus/ava/pull/36/files#diff-b3b53682a18f203ac8d29b0e277cad26R97

Should use https://github.com/floatdrop/pinkie-promise and https://github.com/sindresorhus/pify.

PR #39, opened 2015-09-06
Promisification

vdemedes:
This PR is related to #38. After this PR is complete and callbacks are removed to the maximum extent, AVA's code should be much cleaner.

### Overview

As stated in #38, `pinkie-promise` and `pify` modules are used in this PR.

### Changes

- [x] Promisify `Runner.prototype.serial`
- [x] Promisify `Runner.prototype.concurrent`
- [x] Promisify `Runner.prototype.run`
- [x] Promisify `Test.prototype.run`
- [x] Use promises in tests

As usual, let me know what you think and let's discuss this!


sindresorhus:
> Use mz module when promises are needed for core modules (e.g. fs)

You can just use `pify` for that. See `pify.all()`.

vdemedes:
Don't know why, but I knew you would say that, haha

vdemedes:
@sindresorhus this PR is ready for you to review, let me know what you think ;)

lib/runner.js
```diff
@@ -1,8 +1,10 @@
 'use strict';
+var Promise = require('pinkie-promise');
```
sindresorhus: Actually. I think it's worth using https://github.com/petkaantonov/bluebird here for performance reasons.

lib/runner.js
```diff
@@ -1,8 +1,10 @@
...
+var promisify = require('pify');
 var util = require('util');
 var EventEmitter = require('events').EventEmitter;
-var each = require('each-async');
-var eachSerial = require('async-each-series');
+var each = promisify(require('each-async'), Promise);
```
sindresorhus: Just use `Promise.all()`

lib/runner.js
```diff
@@ -1,8 +1,10 @@
...
+var eachSerial = promisify(require('async-each-series'), Promise);
```
sindresorhus: https://github.com/petkaantonov/bluebird/blob/master/API.md#eachfunction-iterator---promise

lib/runner.js
```diff
@@ -44,67 +46,84 @@ Runner.prototype.addAfterHook = function (title, cb) {
...
+		})
+		.then(function () {
+			self.stats.passCount = self.stats.testCount - self.stats.failCount;
+		});
 };
+
+// Runner.prototype.end = function (cb) {
+// 	this.stats.passCount = this.stats.testCount - this.stats.failCount;
+// 	cb(this.stats, this.results);
+// };
```
sindresorhus: ?

vdemedes: Forgot to remove that

lib/test.js
```diff
@@ -66,34 +67,39 @@ Test.prototype.skip = function () {
...
+	this.promise = {};
 
-	if (!this.fn || this.skipTest) {
-		this.exit();
-	}
-
-	this._timeStart = Date.now();
+	return new Promise(function (resolve, reject) {
+		this.promise.resolve = resolve;
+		this.promise.reject = reject;
```
sindresorhus: I can't really think of a better way right now, but this feels weird.

lib/runner.js
```diff
@@ -44,67 +46,84 @@ Runner.prototype.addAfterHook = function (title, cb) {
...
+			return self.serial(serial);
+		})
+		.then(function () {
+			return self.concurrent(concurrent);
+		})
+		.then(function () {
+			return self.serial(after);
+		})
+		.catch(function () {
+			return;
```
sindresorhus: Why?

lib/test.js
vdemedes: Yeah, it does. My goal was to switch to promises with a minimum code changed.

lib/runner.js
```diff
@@ -44,67 +46,84 @@ Runner.prototype.addAfterHook = function (title, cb) {
...
-};
+	var self = this;
 
-Runner.prototype.end = function (cb) {
-	this.stats.passCount = this.stats.testCount - this.stats.failCount;
-	cb(this.stats, this.results);
+	return this.serial(before)
+		.then(function () {
+			if (self.stats.failCount > 0) {
+				return Promise.reject();
```
sindresorhus: Should we reject with an error with a error message here?

lib/runner.js
vdemedes: Do you prefer `.then(this.concurrent.bind(this, concurrent))`?

lib/runner.js
vdemedes: There was no error message before, so I did not insert it too. Also, there wouldn't be a meaningful message, only smth like "Error happened during tests" and that's it. The goal of this line is just to prevent execution of the following tests.

lib/runner.js
Qix-: I think he means why just return from 119 and not doing anything on the catch.

sindresorhus: Yup, I should have been clearer.

vdemedes: Oh, right, I am just checking notes from the PR, not from the files themselves.

Because previously, when error happened, `this.end` executed and that's it. So that's exactly what I did here. In case of an error, just exit.

lib/test.js
sindresorhus: Ok. Can you at least add a `TODO` in there so it's not forgotten that it needs improving?

vdemedes: Yeah, sure

lib/runner.js
```diff
@@ -44,67 +43,82 @@ Runner.prototype.addAfterHook = function (title, cb) {
...
-				title: test.title,
-				error: err
-			});
-
-			this.emit('test', err, test.title, duration);
-			next();
-		}.bind(this));
-	}.bind(this), cb);
+Runner.prototype.serial = function (tests) {
+	return Promise.resolve(tests).each(function (test) {
```
sindresorhus: This inner function can be DRYed up as the `Runner.prototype.concurrent` has the exact same function content.

vdemedes: Yeah, thought of that too

vdemedes: Just wanted to get a quick feedback on the PR as a whole

sindresorhus:
You can also remove `pinkie-promise` from the tests and devDependency and use Bluebird there too.

vdemedes:
@sindresorhus Right, good catch, missed that

sindresorhus:
The code base is so much cleaner with promises. Excellent work @vdemedes :)

Gonna give it a more thorough review tomorrow. Need sleep now.

vdemedes:
Thank you @sindresorhus ;)

sindresorhus:
I tried running this on [`pageres`](https://github.com/sindresorhus/pageres) and got an error:

```
Unhandled rejection TypeError: Cannot read property 'testCount' of undefined
    at exit (/Users/sindresorhus/dev/ava/index.js:52:11)
    at tryCatcher (/Users/sindresorhus/dev/ava/node_modules/bluebird/js/main/util.js:26:23)
    at Promise._settlePromiseFromHandler (/Users/sindresorhus/dev/ava/node_modules/bluebird/js/main/promise.js:503:31)
    at Promise._settlePromiseAt (/Users/sindresorhus/dev/ava/node_modules/bluebird/js/main/promise.js:577:18)
    at Promise._settlePromises (/Users/sindresorhus/dev/ava/node_modules/bluebird/js/main/promise.js:693:14)
    at Async._drainQueue (/Users/sindresorhus/dev/ava/node_modules/bluebird/js/main/async.js:123:16)
    at Async._drainQueues (/Users/sindresorhus/dev/ava/node_modules/bluebird/js/main/async.js:133:10)
    at Immediate.Async.drainQueues [as _onImmediate] (/Users/sindresorhus/dev/ava/node_modules/bluebird/js/main/async.js:15:14)
    at processImmediate [as _immediateCallback] (timers.js:371:17)
```

It seems the `stats` and `results` arguments in the `exit` function are `undefined`.

lib/runner.js
```diff
@@ -44,67 +43,76 @@ Runner.prototype.addAfterHook = function (title, cb) {
...
+	}.bind(this));
 };
 
-Runner.prototype.serial = function (tests, cb) {
-	eachSerial(tests, function (test, next) {
-		test.run(function (err, duration) {
-			if (err) {
-				this.stats.failCount++;
-			}
+Runner.prototype.addTestResult = function (test) {
```
sindresorhus: This should be a private method, meaning prefixed with `_`: `Runner.prototype._addTestResult`

vdemedes:
@sindresorhus Trying to figure out the right way to test `index.js`. It is a bit problematic, since it exposes only functions to add tests, so I can't test it without `child_process.exec`. What do you think?

sindresorhus:
Using `exec` is fine.

vdemedes:
@sindresorhus Pushed critical fixes and updated es6 test to use `index.js`, instead of `lib/test.js` directly. Also, pageres tests succeed now:

screen shot 2015-09-07 at 1 49 22 pm


But now we have another problem. There's no continuous output, meaning that all output is displayed at the end, when all tests finish. This happens, because `Promise.all` (https://github.com/sindresorhus/ava/pull/39/files#diff-b3b53682a18f203ac8d29b0e277cad26R56) resolves only when all promises are resolved.

vdemedes:
First thing that I came up with to fix that is to use `each-async` again, instead of `Promise.all`.

vdemedes:
@sindresorhus Oh my God, I know how to fix this, never mind ;)

sindresorhus:
Yeah, let's use `each-async` for now. Add a TODO about improving it though.

vdemedes:
I came up with a much better solution and without use of `each-async` ;) Just a few seconds and you'll see the commit!

vdemedes:
Here's that change: https://github.com/sindresorhus/ava/pull/39/files#diff-b3b53682a18f203ac8d29b0e277cad26R52.

The logic here is to output test results immediately, as they are available. But, use `Promise.all` to return from `Runner#concurrent`, when all the promises are resolved (tests are completed).

sindresorhus:
Awesome! I'm really happy about this change. Super work @vdemedes :D

Would you happen to be interested in joining the project? I think you have a lot of good to bring to AVA.

687474703a2f2f7777772e7265616374696f6e676966732e636f6d2f77702d636f6e74656e742f67616c6c6572792f64616e63652d70617274792f74756d626c725f6d306f73727268714d6b31717a6376376e6f345f3235302e676966

Qix-:
:+1: for @vdemedes on the project. Lots of great stuff there.

vdemedes:
@sindresorhus @Qix- Thank you, guys!

Yes, I am definitely interested in joining AVA! I really like it and would love to contribute more!

PR #41, opened 2015-09-08
readme - add AVA feature display

sindresorhus:
Some marketing. Anything I'm missing? Anything that could be improved?

// @vdemedes @Qix- @kevva


vdemedes:
Looks good. I'd only change the text example to a test with some no-brainer assertion, not `t.pass()`:

```js
test(t => {
  t.same([1, 2], [1, 2]);
  t.end();
});
```

And I'd also change `import test from 'ava'` to `var test = require('ava')` to make things completely straightforward. I think, not everyone is familiar with ES6 module system.

sindresorhus:
> And I'd also change import test from 'ava' to var test = require('ava') to make things completely straightforward. I think, not everyone is familiar with ES6 module system.

If they're not familiar with the module system they're probably not familiar with arrow functions either. I'd rather go all in with the ES2015 syntax than awkwardly something in the middle.

Qix-:
:+1:

PR #42, opened 2015-09-08
Use pinkie-promise again

floatdrop:
Is there a reason to not use native Promise (if available)? Am I missing something?


sindresorhus:
Bluebird is faster than native (i know...) and includes some useful utilities. Also no point in including multiple promise libs.

floatdrop:
Fair enough.

sindresorhus:
@floatdrop I'm btw slowly migrating my modules over to promises using `pinkie-promise` ;)

https://www.npmjs.com/browse/depended/pinkie-promise

PR #43, opened 2015-09-08
remove EventEmitter from Test

floatdrop:
`EventEmmitter` is not used in `Test` class and seems like won't be used in future (nor exposed in readme).


sindresorhus:
Nice catch.

Issue #25, opened 2015-08-28
Support registering a custom assertion module


ColCh:
How to use it?  For example,  Chai is very nice. Or assert.js

sindresorhus:
Yes you can. Updated the readme about it: https://github.com/sindresorhus/ava/commit/3deba5189186ce73ef6a7bc18c4044f713c63eed

It won't work with planned assertions though. For that, we'll need an AVA method that hooks those assertion so we can do the assert counting. Shouldn't be that hard, but not top priority.

ColCh:
Thanks. Any tutorial on how to do that? Code sample will be cool.

sindresorhus:
How to do what? With assert counting is not possible yet as outlined above. Without assert counting is documented in the readme as already mentioned.

sindresorhus:
Please read the discussion in https://github.com/sindresorhus/ava/pull/49 before commenting.

novemberborn:
I think we've become quite comfortable with AVA's built-in assertions since this issue was raised. We won't be adding tight integration for other assertion libraries any time soon, so I'm closing this issue.

sindresorhus:
People can already use their favorite assertion module with AVA. It just won't work with planned assertions. That being said, I almost never use planned assertions anymore, now that we have async/await.

PR #46, opened 2015-09-09
Less API using power-assert

uiureo:
I just want to share my demo, so feel free to close this PR. Just ignore it if you don't like it.

Added `t.ok(value, [message])` using [power-assert](https://github.com/power-assert-js/power-assert). It shows descriptive assertion message by parsing the JS expression.

## example
```javascript
test((t) => {
  var a = { name: 'foo' }

  t.ok(a.name === 'bar')

  t.end()
})
```

shows a message like this.

```javascript
> ava

  ✖ [anonymous]   # test.js:26
  t.ok(a.name === 'bar')
       | |    |         
       | |    false     
       | "foo"          
       Object{name:"foo"}
  
  --- [string] 'bar'
  +++ [string] a.name
  @@ -1,3 +1,3 @@
  -bar

```

## apis
```javascript
t.true(true)
t.is('foo', 'foo');
t.not('foo', 'bar');
t.same({a: 'a'}, {a: 'a'});
t.regexTest(/^abc$/, 'abc');
```

with power-assert: 

```javascript
var deepEqual = require('deep-equal');

t.ok(true);
t.ok('foo' === 'foo');
t.ok('foo' !== 'bar');
t.ok(deepEqual({a: 'a'}, {a: 'a'}));
t.ok(/^abc$/.test('ab'));
```

## more example
```javascript
  ✖ [anonymous]   # test.js:44
  
  t.ok(deepEqual(a, b))
       |         |  |  
       |         |  ["a","a"]
       false     ["a","b"]
  
  ✖ [anonymous]   # test.js:50
  
  t.ok(/^abc$/.test(`ab${ s }`))
               |    |     |     
               |    "aba" "a"   
               false            
  
```

I know custom assertion is supported on ava and it worked pretty well.
But I wanted to use `t.plan()` and that assertion function with ava's minimal interface.


Qix-:
It'd be nice if this translated into some way to hook into Ava to allow it to use any arbitrary assertion library, including support for `t.plan()`.

Qix-:
@uiureo why did you close? :o

uiureo:
Because I don't intend this PR to be merged. I don't want the opened state to bother you guys.

> It'd be nice if this translated into some way to hook into Ava to allow it to use any arbitrary assertion library, including support for t.plan().

That's the very thing I want.
I'm looking into the code and I would try if I come up with some way.

Related: #25

Qix-:
:+1: sounds good

sindresorhus:
I'd like to keep this open for a while so people can see it. I think `power-assert` is really cool and I actually considered including it in AVA early on. The reasons I didn't were the compile step (which is not a problem now that we include Babel by default) and the result, while being very informative, is a bit noisy. `foo === bar` is a lot clearer and faster to grasp. Is there any way to simplify the output?

uiureo:
> the result, while being very informative, is a bit noisy. foo === bar is a lot clearer and faster to grasp. Is there any way to simplify the output?

I agree its output is a little bit noisy. It seems you can write custom output formatter in power-assert.

https://github.com/power-assert-js/power-assert#optionsoutput
https://github.com/power-assert-js/power-assert-formatter

FYI @twada

twada:
Hi, I'm creator of power-assert.
Thank you for mentioning me @uiureo.
I'm very interested in AVA now.

power-assert output consists of 4 parts, rendered by 4 renderers (`file`, `assertion`, `diagram`, and `binary-expression`).
And you can omit each part from output through customization API.

For example, output below 
```
  # test.js:26
  t.ok(a.name === 'bar')
       | |    |         
       | |    false     
       | "foo"          
       Object{name:"foo"}
  --- [string] 'bar'
  +++ [string] a.name
  @@ -1,3 +1,3 @@
  -bar
```

`file` renderer produces,
```
   # test.js:26
```

`assertion` renderer produces,
```
  t.ok(a.name === 'bar')
```


`diagram` renderer produces,
```
       | |    |         
       | |    false     
       | "foo"          
       Object{name:"foo"}
```

and `binary-expression` renderer produces
```
  --- [string] 'bar'
  +++ [string] a.name
  @@ -1,3 +1,3 @@
  -bar
```

To disable verbose graph, remove `diagram` renderer from `output.renderers` by using `customize` method. For example, this configuration

```javascript
var assert = require('power-assert').customize({
    output: {
        renderers: [
            './built-in/file',
            './built-in/assertion',
            './built-in/binary-expression'
        ]
    }
});
```

produces output as below.

```
  # test.js:26
  t.ok(a.name === 'bar')
  
  --- [string] 'bar'
  +++ [string] a.name
  @@ -1,3 +1,3 @@
  -bar
```

And of course, I want to hear your request, and create custom formatter for AVA!

Qix-:
To be honest, the binary renderer is the part that seems noisy to me. The diagram is what makes it super useful.

sindresorhus:
Instead of:

```
  # test.js:26
  t.ok(a.name === 'bar')
       | |    |         
       | |    false     
       | "foo"          
       Object{name:"foo"}
  --- [string] 'bar'
  +++ [string] a.name
  @@ -1,3 +1,3 @@
  -bar
```

I think I would want something like:

```
  # test.js:26
  a.name === 'bar'
     |
     |
   "foo"          
```

twada:
Hi @Qix- 

> To be honest, the binary renderer is the part that seems noisy to me. The diagram is what makes it super useful.

Understood. You can disable `binary-expression` renderer by removing `'./built-in/binary-expression'` from `output.renderers`

```javascript
var assert = require('power-assert').customize({
    output: {
        renderers: [
            './built-in/file',
            './built-in/assertion',
            './built-in/diagram'
        ]
    }
});
```

You'll see output like

```
  # test.js:26
  t.ok(a.name === 'bar')
       | |    |         
       | |    false     
       | "foo"          
       Object{name:"foo"}
```

> Very tempting to switch from should. I'll play with it a bit more.

I'm glad to hear that! :smile:

twada:
Hi @sindresorhus, thank you for your request.
I'll create succinct output renderer for AVA!

sindresorhus:
@twada Any updates? Would really like to use `power-assert` to improve AVAs assert output.

twada:
@sindresorhus I've extracted [power-assert-renderers](https://github.com/twada/power-assert-renderers) module out and now creating less-noisy succinct-diagram renderer for AVA. 

I would like you to see [succinct-diagram-test.js](https://github.com/twada/power-assert-renderers/blob/master/test/succinct-diagram-test.js).
(You'll see `assert` or `t.ok` calls are still remaining in output, but it's a bit difficult to remove for now)

If you like it, I'm going to integrate it into this pull-req and also power-assert master.

sindresorhus:
@twada Yes, looks very good. I'm fine with this for now. Would be awesome if you would integrate into this PR and fix the merge conflict.

kevva:
Really exciting stuff.

uiureo:
I did quick hack to use the succinct renderer and fixed the merge conflict.

Now the output looks like this:

```sh
❯ npm test    

> ava

  ✖ [anonymous]   
  t.ok(a.name === 'bar')
         |              
         "foo"          
  

  1 test failed

  1. [anonymous]
  AssertionError:   
  t.ok(a.name === 'bar')
         |              
         "foo"          
    at decoratedAssert (/Users/zat/.ghq/github.com/sindresorhus/ava/node_modules/power-assert/node_modules/empower/lib/decorate.js:42:30)
    at Function.powerAssert (/Users/zat/.ghq/github.com/sindresorhus/ava/node_modules/power-assert/node_modules/empower/index.js:58:32)
    at Test.ok (/Users/zat/.ghq/github.com/sindresorhus/ava/lib/test.js:85:15)
    at Test.fn (/Users/zat/.ghq/github.com/uiureo/ava-power-assert-example/test.js:26:5)
    at Test.<anonymous> (/Users/zat/.ghq/github.com/sindresorhus/ava/lib/test.js:124:19)
    at tryCatcher (/Users/zat/.ghq/github.com/sindresorhus/ava/node_modules/bluebird/js/main/util.js:26:23)
    at Promise._resolveFromResolver (/Users/zat/.ghq/github.com/sindresorhus/ava/node_modules/bluebird/js/main/promise.js:480:31)
    at new Promise (/Users/zat/.ghq/github.com/sindresorhus/ava/node_modules/bluebird/js/main/promise.js:70:37)
    at Test.run (/Users/zat/.ghq/github.com/sindresorhus/ava/lib/test.js:110:9)
    at tryCatcher (/Users/zat/.ghq/github.com/sindresorhus/ava/node_modules/bluebird/js/main/util.js:26:23)
    at ReductionPromiseArray._promiseFulfilled (/Users/zat/.ghq/github.com/sindresorhus/ava/node_modules/bluebird/js/main/reduce.js:105:38)
    at ReductionPromiseArray.init [as _init$] (/Users/zat/.ghq/github.com/sindresorhus/ava/node_modules/bluebird/js/main/promise_array.js:92:18)
    at ReductionPromiseArray.init (/Users/zat/.ghq/github.com/sindresorhus/ava/node_modules/bluebird/js/main/reduce.js:42:10)
    at Async._drainQueue (/Users/zat/.ghq/github.com/sindresorhus/ava/node_modules/bluebird/js/main/async.js:128:12)
    at Async._drainQueues (/Users/zat/.ghq/github.com/sindresorhus/ava/node_modules/bluebird/js/main/async.js:133:10)
    at Async.drainQueues (/Users/zat/.ghq/github.com/sindresorhus/ava/node_modules/bluebird/js/main/async.js:15:14)

npm ERR! Test failed.  See above for more details.
```

lib/babel.js
```diff
@@ -15,7 +17,15 @@ try {
 
 	require(localBabel)(options);
 } catch (err) {
-	require('babel-core/register')(options);
+	require('babel-core/register')(assign({}, options, {
```
sindresorhus: The options here need to be put in a variable and used in line 18 too.

lib/babel.js
```diff
@@ -1,6 +1,8 @@
 'use strict';
 
 var resolveFrom = require('resolve-from');
+var assign = require('object-assign');
```
sindresorhus: `assign` => `objectAssign`

twada:
@uiureo Thanks for responding so quickly, nice to see this being merged!

lib/test.js
```diff
@@ -70,6 +78,19 @@ Object.keys(assert).forEach(function (el) {
 	};
 });
 
+Test.prototype.ok = function () {
```
sindresorhus: This should probably be in the `assert.js` file.

lib/test.js
```diff
@@ -70,6 +78,19 @@ Object.keys(assert).forEach(function (el) {
...
+
+	try {
+		powerAssert.apply(powerAssert, arguments);
+	} catch (err) {
+		this.assertError = err;
+	}
+};
+
+Test.prototype._capt = powerAssert._capt;
+Test.prototype._expr = powerAssert._expr;
```
sindresorhus: What are these? Add a comment inline in the code.

package.json
```diff
@@ -73,6 +74,8 @@
     "meow": "^3.3.0",
     "object-assign": "^4.0.1",
     "plur": "^2.0.0",
+    "power-assert": "^1.0.0",
+    "power-assert-renderers": "twada/power-assert-renderers",
```
sindresorhus: @twada Can this be published on npm? I don't want to depend on a git repo.

readme.md
```diff
@@ -449,4 +449,3 @@ Concurrency is not parallelism. It enables parallelism. It's about dealing with,
 	<br>
 	<br>
 </div>
-
```
sindresorhus: ?

lib/test.js
```diff
@@ -2,6 +2,14 @@
...
 var Promise = require('bluebird');
 var setImmediate = require('set-immediate-shim');
+var powerAssert = require('power-assert').customize({
+	output: {
+		renderers: [
+			require('power-assert-renderers').AssertionRenderer,
+			require('power-assert-renderers').SuccinctRenderer
+		]
+	}
+});
```
sindresorhus: Why can't this be done at require time in https://github.com/sindresorhus/ava/pull/46/files#diff-c56ba0fef1bb510acefb4680eb31e9c6R21?

sindresorhus:
@twada Can you review?

package.json
twada: @sindresorhus Agreed. I'm going ahead.

lib/test.js
twada: @sindresorhus power-assert architecture consists of two parts. Instrumentation (i.e. transformation) side and runtime side. This code is a runtime side. The diff you mentioned is an instrumentation side. Output rendering belongs to runtime side.

twada:
@sindresorhus

> @twada Can you review?

Sure!

readme.md
uiureo: oops

lib/test.js
```diff
@@ -70,6 +70,9 @@ Object.keys(assert).forEach(function (el) {
 	};
 });
 
+Test.prototype._capt = assert._capt
+Test.prototype._expr = assert._expr
```
uiureo: @twada Do you have any idea to avoid this? I don't understand much what they are...

uiureo:
Squashed commits.  I'll write some test.

lib/assert.js
```diff
@@ -32,6 +50,9 @@ x.ok = x.assert = function (val, msg) {
 	test(val, create(val, true, '==', msg, x.ok));
 };
 
+x._capt = assert._capt;
+x._expr = assert._expr;
```
twada: These two lines aren't needed.

lib/test.js
twada: @uiureo Hmm, this workaround is required for now, but looks a bit messy.
To clear this up, I have to create a module to enhance prototype, not function nor object.
Would you wait some more days?

lib/assert.js
```diff
@@ -20,6 +24,20 @@ function test(ok, opts) {
 	}
 }
 
+function enhanceAssert(assert) {
+	return empower(assert,
```
sindresorhus: No need to return since the return value is not used.

lib/assert.js
```diff
@@ -20,6 +24,20 @@ function test(ok, opts) {
...
+		powerAssertFormatter({
+			renderers: [
+				powerAssertRenderers.AssertionRenderer,
+				powerAssertRenderers.SuccinctRenderer
+			]
+		}),
+		{
+			destructive: true
+		}
+	);
```
sindresorhus: Looks like we'll have to add patterns for all the non-standard assert methods: https://github.com/power-assert-js/empower#optionspatterns

@twada Is this correct?

sindresorhus:
@twada Can you explain what the [`modifyMessageOnRethrow`](https://github.com/power-assert-js/empower#optionsmodifymessageonrethrow) option does? Should we enable it?

twada:
@sindresorhus `modifyMessageOnRethrow` option indicates whether to [modify AssertionError message](https://github.com/power-assert-js/empower/blob/master/lib/decorator.js#L81) on (re)throwing AssersionError or not.

> Should we enable it?

You are right. Should be `true` in this case.

lib/assert.js
twada: @sindresorhus Yes. It would be better to configure them explicitly.
(Is `t.assert` public API?)

```js
function enhanceAssert(assert) {
    empower(assert,
		powerAssertFormatter({
			renderers: [
				powerAssertRenderers.AssertionRenderer,
				powerAssertRenderers.SuccinctRenderer
			]
		}),
		{
			destructive: true,
			modifyMessageOnRethrow: true,
			saveContextOnRethrow: false,
			patterns: [
				't.ok(value, [message])',
				't.assert(value, [message])'
			]
		}
	);
}
```


And if you want to enhance output of assertion methods other than `ok`, you can enhance them.

```js
function enhanceAssert(assert) {
    empower(assert,
		powerAssertFormatter({
			renderers: [
				powerAssertRenderers.AssertionRenderer,
				powerAssertRenderers.SuccinctRenderer
			]
		}),
		{
			destructive: true,
			modifyMessageOnRethrow: true,
			saveContextOnRethrow: false,
			patterns: [
				't.ok(value, [message])',
				't.notOk(value, [message])',
				't.true(value, [message])',
				't.false(value, [message])',
				't.is(value, expected, [message])',
				't.not(value, expected, [message])',
				't.same(value, expected, [message])',
				't.notSame(value, expected, [message])',
				't.regexTest(regex, contents, [message])'
			]
		}
	);
}
```

In this case, intrumentation patterns in `lib/babel.js` should be the same.
```js
var options = {
	only: /(test|test\-.+|test\/.+)\.js$/,
	optional: hasGenerators ? ['asyncToGenerator'] : [],
	plugins: [
		createEspowerPlugin(require('babel-core'), {
			patterns: [
				't.ok(value, [message])',
				't.notOk(value, [message])',
				't.true(value, [message])',
				't.false(value, [message])',
				't.is(value, expected, [message])',
				't.not(value, expected, [message])',
				't.same(value, expected, [message])',
				't.notSame(value, expected, [message])',
				't.regexTest(regex, contents, [message])'
			]
		})
	]
};
```

sindresorhus: Yeah, I don't see why we wouldn't want to decorate all of them, even though `t.ok`/`notOk` are the ones with the most gain.

> (Is t.assert public API?)

No, it's only there for legacy reasons.

package.json
twada: Just released [power-assert-renderers 0.1.0](https://www.npmjs.com/package/power-assert-renderers)

@uiureo Would you update `package.json`?

lib/test.js
uiureo: Ok. I'll add a comment for now.

twada: Thanks! These methods capture each node value in assertion expressions then pass the values to power-assert formatter/renderer.

package.json
uiureo: Thanks, I've updated it.

uiureo:
I've tried to add patterns like this:

```js
	plugins: [
		createEspowerPlugin(require('babel-core'), {
			patterns: [
				't.ok(value, [message])',
				't.notOk(value, [message])',
				't.true(value, [message])',
				't.false(value, [message])',
				't.is(value, expected, [message])',
				't.not(value, expected, [message])',
				't.same(value, expected, [message])',
				't.notSame(value, expected, [message])',
				't.regexTest(regex, contents, [message])'
			]
		})
	]
```

```
test((t) => {
  var a = 'foo'
  t.is(a, 'bar')
  t.end()
})
```

But I got weird output like this. 

```
  ✖ [anonymous]   
  t.ok(a.name !== 'foo')
         |              
         "foo"          
  
  ✖ [anonymous] { powerAssertContext: { value: 'foo', events: [ [Object] ] },
  source: { content: 't.is(a, \'bar\')', filepath: 'test.js', line === 'bar'
```

 `t.ok()` works correctly.

twada:
@uiureo Thank you for reporting.
In this case, intrumentation patterns in `lib/babel.js` should be the same.
So please configure `lib/babel.js` and try again.

```js
function enhanceAssert(assert) {
    empower(assert,
        powerAssertFormatter({
            renderers: [
                powerAssertRenderers.AssertionRenderer,
                powerAssertRenderers.SuccinctRenderer
            ]
        }),
        {
            destructive: true,
            modifyMessageOnRethrow: true,
            saveContextOnRethrow: false,
            patterns: [
                't.ok(value, [message])',
                't.notOk(value, [message])',
                't.true(value, [message])',
                't.false(value, [message])',
                't.is(value, expected, [message])',
                't.not(value, expected, [message])',
                't.same(value, expected, [message])',
                't.notSame(value, expected, [message])',
                't.regexTest(regex, contents, [message])'
            ]
        }
    );
}
```

uiureo:
@twada 
Ah, I missed the point. 
It works right. Thanks for help.

lib/test.js
twada: @uiureo 

> To clear this up, I have to create a module to enhance prototype, not function nor object.
Would you wait some more days?

Yesterday I tried to add prototype enhancement feature to `empower`.
However, enhancing prototype does not work well with AVA.
So this workaround is needed for a while (Thank you for the appropriate inline comment).

By the way, I just released [empower 1.0.2](https://www.npmjs.com/package/empower).
This prevents `_capt` and `_expr` from being wrapped by `Object.keys(assert).forEach(function (el) {` loop.
A little progress.

To recap, the workaround 
```js
Test.prototype._capt = assert._capt
Test.prototype._expr = assert._expr
```
is still needed, even if empower is updated to 1.0.2, but it makes a little progress.

Would you update `package.json` again?

uiureo: I've updated empower. Thanks.

twada: @uiureo Thank you!

test/test.js
```diff
@@ -774,3 +774,12 @@ test('fail-fast mode', function (t) {
 		t.end();
 	});
 });
+
+test('power-assert support', function (t) {
+	t.plan(2);
+
+	execCli('fixture/power-assert.js', function (err, stdout, stderr) {
+		t.ok(err);
+		t.true(stderr.indexOf('t.ok(a === \'bar\')\n') !== -1);
```
sindresorhus: How is this testing that it's working? It's only the normals output.

And should be:

````js
t.not(stderr.indexOf('t.ok(a === \'bar\')\n'), -1);
```

test/fixture/power-assert.js
```diff
@@ -0,0 +1,8 @@
+import test from '../../';
+
+test(t => {
+  const a = 'foo';
```
sindresorhus: wrong indent

lib/test.js
```diff
@@ -70,6 +70,11 @@ Object.keys(assert).forEach(function (el) {
 	};
 });
 
+// Workaround for power-assert
```
sindresorhus: @twada Is there anything you can do in `power-assert` to make this unnecessary?

lib/babel.js
```diff
@@ -1,13 +1,29 @@
...
+				't.ok(value, [message])',
+				't.notOk(value, [message])',
+				't.true(value, [message])',
+				't.false(value, [message])',
+				't.is(value, expected, [message])',
+				't.not(value, expected, [message])',
+				't.same(value, expected, [message])',
+				't.notSame(value, expected, [message])',
+				't.regexTest(regex, contents, [message])'
+			]
```
sindresorhus: This list is duplicated. Maybe move it out into a separate file and require it in both places.

sindresorhus: Maybe move the whole `function enhanceAssert(assert) {` function into a file called `enhance-assert` and also export the list. Then that file could be both used here and in `lib/assert.js`.

test/fixture/power-assert.js
uiureo: Thanks. xo ignores `test/fixture/*` by default and I missed it.
Is there any way to exclude this file from lint ignores?

test/test.js
uiureo: Without power-assert, the output should be `false == true` because this assert is `t.ok()`.
I thought probably it renders `t.ok(a === 'bar')` only when power-assert works.

```js
test(t => {
	const a = 'foo';

	t.ok(a === 'bar');
	t.end();
});
```

Should I test this renders the diagram output somehow?

```
  t.ok(a === 'bar')
       |           
       "foo"       
```

lib/babel.js
uiureo: nice idea.

test/test.js
uiureo: I've added a test that it renders diagram output by regexp.
Let me know if you want some changes.

uiureo:
@sindresorhus I've done some tweaks. Can you review again?

sindresorhus:
Landed. Woop woop. Great work @uiureo. And thanks for reviewing @twada :)

uiureo:
Thanks a lot @sindresorhus @twada. I couldn't do this without your great help :)

twada:
@sindresorhus Wow I'm very glad to see this pull-req is getting merged!! :tada: 
Thank you so much @uiureo for your great work!

lib/test.js
twada: Unfortunately, this workaround is necessary for now (`_capt` and `_expr` methods are needed at the top level).

To make them look less messy, I'll introduce small helper method like
```js
empower.makeCapturable(Test.prototype);
```
(looking for a better name though)

Issue #1, opened 2014-11-18
Parallelism


sindresorhus:
Would be cool to implement actual parallelism. Could be as simple as executing each test file in a separate node process using `child_process.fork`. Each node sends back its result and the main process merges it and presents it to the user.

Doing it in sub-processes comes with the added benefit of each file being isolated and not being able to screw up the environment for others.

> Concurrency is not parallelism. It enables parallelism. It's about dealing with, while parallelism is about *doing*, lots of things at once.

PR #47, opened 2015-09-10
Parallelism

vdemedes:
This PR is related to #1.

### Description

Test files supplied to `ava` are now run in their own `node` process. This is implemented using [child_process#fork](https://nodejs.org/api/child_process.html#child_process_child_process_fork_modulepath_args_options).

### Changes

- [x] Separate output functions to `lib/logger.js`
- [x] Fork for each test file using `child_process#fork`
- [x] Create an intermediary [babel.js](https://github.com/sindresorhus/ava/pull/47/files#diff-f6ab9a89a7b25074280e693908d6cb01R1) script, which registers a require hook (`babel/register`) for ES6 support
- [x] Exit `ava` CLI with a correct exit code

### Output

Output remains the same as before, you'll see no visual differences. Although, I do think that we should prefix/suffix test titles with a test file in case of failure, to know where this test actually is. For example:

Command:

```
$ ava
```

`test` directory contents:

```
- test-requests.js
- test-options.js
```

```
  ✔ send a request
  ✖ parse options - error title (test-options.js)
```

Other than that, let's discuss this! I think this is a killer feature for AVA.


Qix-:
I remember saying something about it before, but we just need to make it completely obvious you'll need to re-check your coverage tools, e.g. `istanbul`. I think @sindresorhus had an answer for that a while back. Most coverage tools don't work with forked scripts.

sindresorhus:
Yup, we should add a tip about using [`nyc`](https://github.com/bcoe/nyc), which wraps Istanbul with childprocess support.

vdemedes:
@Qix- Great catch!

vdemedes:
@sindresorhus @Qix- @kevva Any thoughts on this PR?

Qix-:
Totally on @sindresorhus for this one, I still haven't looked too closely at the ava code to know whether or not this is good. :dancer:

cli.js
```diff
@@ -56,8 +79,43 @@ function run(file) {
...
+		};
+
+		files++;
+
+		var babel = join(__dirname, 'babel.js');
+
+		var ps = fork(babel, [file], options);
+
+		ps.stdout.on('data', test);
+		ps.stderr.on('data', test);
```
sindresorhus: I was thinking that only the CLI would print to stdout. The forked processes would message status updates through the message channel. That way the parallelism is contained to the API and the CLI only calls the API normally.

lib/logger.js
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	if (err) {
+		log.error(title, chalk.red(err.message));
+		return;
+	}
+
+	// if (runner.stats.testCount === 1) {
+	// 	return;
+	// }
```
sindresorhus: ?

cli.js
vdemedes: I also thought to use some kind of IPC, but turned out that you can send messages only parent -> fork way, but not the opposite. So in a PR, fork process detects if it is a fork (by checking `process.env.AVA_FORK`) and outputs passed tests to `stdout` and failed ones to `stderr`. If you could recommend a better way of exchanging data between processes, that would be great! I am not a huge fan of the current method either.

lib/logger.js
vdemedes: For `logger` to access a `runner` object, there would need to be some kind of initialization to get a reference to it. For example:

```js
var Runner = require('./lib/runner');
var log = require('./lib/logger');

var runner = new Runner();

log.init(runner);
```

So I decided to skip it for now.

cli.js
sindresorhus: It should be bi-directional: http://stackoverflow.com/a/10394457/64949

lib/logger.js
sindresorhus: k

cli.js
vdemedes: Damn, missed this somehow. Will convert it to `.send()` later.

vdemedes:
@sindresorhus Implemented IPC with `process.send`.

vdemedes:
For some unknown reason tests fail on Travis with node 4.0.0. On my laptop, they run just fine with the latest node:

screen shot 2015-09-12 at 2 02 59 pm

babel.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
```
sindresorhus: This file needs to be added to `files` in package.json.

cli.js
```diff
@@ -56,10 +74,44 @@ function run(file) {
 			return;
 		}
 
-		require(file);
```
sindresorhus: The below should be in a separate function.

babel.js
sindresorhus: Or better yet, just move it to `/lib`

cli.js
```diff
@@ -56,10 +74,44 @@ function run(file) {
 			return;
 		}
```
sindresorhus: Instead of running it directly here, how about we just return the file, and use `[].map` here: https://github.com/vdemedes/ava/blob/2651df6749af48ee985bad9b92d107d4770ea0eb/cli.js#L125 We can then run everything from here https://github.com/vdemedes/ava/blob/2651df6749af48ee985bad9b92d107d4770ea0eb/cli.js#L128 instead and use `Promise.all` for the execution instead of a global `files` counter.

cli.js
```diff
@@ -35,11 +33,31 @@ var cli = meow({
...
+		failed++;
+	} else {
+		log.test(null, data.title, data.duration);
+
+		passed++;
+	}
+}
+
 function run(file) {
 	fs.stat(file, function (err, stats) {
```
sindresorhus: For ease, this can just be `fs.statSync`.

cli.js
```diff
@@ -56,10 +74,44 @@ function run(file) {
...
+		};
+
+		files++;
+
+		var babel = join(__dirname, 'babel.js');
+
+		var ps = fork(babel, [file], options);
+
+		ps.on('message', test);
+		ps.on('close', exit);
```
sindresorhus: I think all the fork logic should be outside `cli.js`, in `index.js` or maybe a separate `run.js`/`fork.js` file or something. Separation of concerns and all. This way the CLI could just use a promise interface and get back the result when done without having to know at all that there's forking going on behind the scenes.

sindresorhus: This will also make the fork logic testable, which it should be.

cli.js
vdemedes: So you are saying `fork.js` should return a Promise? But it also needs to emit `test` event for each test.

vdemedes:
For some unknown reason tests on Travis CI (v0.12.7) failed. Just did a clean test run in Docker with node v0.12.7:

ava

cli.js
sindresorhus: It can both return a promise and emit events. Events are emitted on the object, not promise.

vdemedes: How can it both return a promise and emit events? I had this in mind:

```js
function fork (file) {
    var ps = child_process.fork(file);

    // need also to emit test events somehow

    ps.on('close', function (code) {
      if (code > 0) return reject();

      resolve();
    });
  });
}
```

As a workaround, we could accept test listener:

```js
function fork (file, test) {
  ...
  ps.on('data', test);
  ...
}
```

vdemedes:
Pushed a change, which now shows output from forked processes. Before, all output (`console.log`, `process.stdout.write`) was hidden, when tests are executed via `ava` command. Now developers won't be confused, why their `console.log`s are not showing up.

lib/fork.js
```diff
@@ -0,0 +1,34 @@
...
+
+	var options = {
+		silent: true,
+		env: assign({}, process.env, {AVA_FORK: 1})
+	};
+
+	var ps = childProcess.fork(babel, [file], options);
+
+	ps.stdout.pipe(process.stdout);
+	ps.stderr.pipe(process.stderr);
```
sindresorhus: Only `cli.js` should write to stdout/stderr. This could maybe also be attached on the returned promise. Unless you can come up with a better solution.

index.js
```diff
@@ -60,23 +20,43 @@ function stack(results) {
...
+		process.send({
+			err: err || {},
+			title: title,
+			duration: duration
+		});
+
+		return;
+	}
+
+	log.test(err, title, duration);
```
sindresorhus: This should also not log. Should happen in the CLI.

lib/fork.js
```diff
@@ -0,0 +1,34 @@
...
+
+	var promise = new Promise(function (resolve) {
+		ps.on('exit', resolve);
```
sindresorhus: This needs to take the exit code into account and resolve/reject based on that. And shouldn't the resolved promise contain the result?

cli.js
```diff
@@ -69,15 +101,17 @@ function init(files) {
...
-	return globby(files).then(function (files) {
-		files.forEach(function (file) {
-			run(path.resolve(file));
-		});
+	return globby(files)
+		.then(function (files) {
+			// convert relative paths to absolute
+			files = files.map(function (file) {
+				return path.join(process.cwd(), file);
+			});
```
sindresorhus: I think we should first resolve all paths before we do anything. That would make the code cleaner and less coupled. Then with that array of files, run fork (`fork(file).on('message', test)`) on them here.

sindresorhus: So one function (`handlePaths()`?) that globs the files, filters out js files (like `run` does now), stats.isDirectory() (like `run` does now), convert paths to absolute, then returns them. This function should be used here.

cli.js
```diff
@@ -35,29 +32,64 @@ var cli = meow({
...
+
+	if (stats.isDirectory()) {
+		return init(path.join(file, '*.js'));
+	}
+
+	if (path.extname(file) !== '.js') {
+		return Promise.resolve();
+	}
+
+	return fork(file).on('message', test);
```
sindresorhus: Instead of a `message` event, the result should be as the resolved promise.

cli.js
```diff
@@ -35,29 +32,64 @@ var cli = meow({
 	string: ['_']
 });
 
+var errors = [];
+var failed = 0;
+var passed = 0;
```
sindresorhus: These should be returned from the API as some kind of stats, not here.

lib/fork.js
sindresorhus: Actually, all information should just be in `message` events, not stdout/stderr.

lib/logger.js
sindresorhus: Can you add a `TODO` for this so we don't forget?

lib/fork.js
vdemedes:
```js
ps.stdout.pipe(process.stdout);
ps.stderr.pipe(process.stderr);
```

This is not for sending data from forked processes. I am sending data via `process.send()`. This is for piping output from forked processes, so that `console.log()`s won't be hidden, when executed via CLI.

index.js
vdemedes: But then, if tests are executed directly, via `node test.js`, there won't be any output.

lib/fork.js
vdemedes: I decided to skip `reject()`, because there was no logic for that previously. There was just an `exit()` function, that outputs number of passed/failed tests. And what kind of result you are thinking of?

cli.js
vdemedes: Good point.

cli.js
vdemedes: Don't understand. This function returns a promise. It also emits `message` event on completed/failed test, like we agreed earlier. Returned promise resolves, when all tests are completed.

cli.js
vdemedes: We would still need to sum all results and display all errors at the end.

lib/fork.js
sindresorhus: Disregard my second comment. My first still stands though. Only `cli.js` should write to stdout/stderr.

index.js
sindresorhus: Hmm, I think we're way past supporting `node test.js` now, with parallelism and ES2015. Unless you can come up with a good way to support both cases. Maybe detect that it's executed directly and log output. You can detect direct execution with: `require.main === module`.

lib/fork.js
sindresorhus: It should get the final stats (passed/failed tests, etc), where we write `1 test passed successfully`.

cli.js
sindresorhus: Oh, yeah, my mistake.

sindresorhus:
Looks good to me now. Awesome work on this @vdemedes!

unicorn

PR #49, opened 2015-09-10
Allow custom assert modules

jenslind:
Allows you to set a custom assert module so it can be used together with `plan()`.
```
var test = require('ava')
var assert = require('assert');
test.setAssertModule(assert);
```

#25

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -92,11 +92,6 @@ Runner.prototype._addTestResult = function (test) {
 };
 
 Runner.prototype.run = function () {
-	// Set default assert module if needed
-	if (!this._assertModule) {
-		Test._setAssertModule();
-	}
-
 	var self = this;
 	var tests = this.tests;
 	var stats = this.stats;
@@ -126,5 +121,5 @@ Runner.prototype.run = function () {
 
 // Set custom assert module
 Runner.prototype.setAssertModule = function (assertModule) {
-	this._assertModule = Test._setAssertModule(assertModule);
+	Test._setAssertModule(assertModule);
 };
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -24,6 +24,8 @@ function Runner(opts) {
 		before: [],
 		after: []
 	};
+
+	this._assertModule = null;
 }
 
 util.inherits(Runner, EventEmitter);
@@ -31,20 +33,20 @@ module.exports = Runner;
 
 Runner.prototype.addTest = function (title, cb) {
 	this.stats.testCount++;
-	this.tests.concurrent.push(new Test(title, cb));
+	this.tests.concurrent.push(new Test(title, this._assertModule, cb));
 };
 
 Runner.prototype.addSerialTest = function (title, cb) {
 	this.stats.testCount++;
-	this.tests.serial.push(new Test(title, cb));
+	this.tests.serial.push(new Test(title, this._assertModule, cb));
 };
 
 Runner.prototype.addBeforeHook = function (title, cb) {
-	this.tests.before.push(new Test(title, cb));
+	this.tests.before.push(new Test(title, this._assertModule, cb));
 };
 
 Runner.prototype.addAfterHook = function (title, cb) {
-	this.tests.after.push(new Test(title, cb));
+	this.tests.after.push(new Test(title, this._assertModule, cb));
 };
 
 Runner.prototype.concurrent = function (tests) {
@@ -121,5 +123,5 @@ Runner.prototype.run = function () {
 
 // Set custom assert module
 Runner.prototype.setAssertModule = function (assertModule) {
-	Test._setAssertModule(assertModule);
+	this._assertModule = assertModule;
 };
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -92,6 +92,11 @@ Runner.prototype._addTestResult = function (test) {
 };
 
 Runner.prototype.run = function () {
+	// Set default assert module if needed
+	if (!this._assertModule) {
+		Test._setAssertModule();
+	}
+
 	var self = this;
 	var tests = this.tests;
 	var stats = this.stats;
@@ -117,4 +122,9 @@ Runner.prototype.run = function () {
 		.then(function () {
 			stats.passCount = stats.testCount - stats.failCount;
 		});
+};
+
+// Set custom assert module
+Runner.prototype.setAssertModule = function (assertModule) {
+	this._assertModule = Test._setAssertModule(assertModule);
 };
--- a/lib/test.js
+++ b/lib/test.js
@@ -45,17 +45,25 @@ Test.prototype._assert = function () {
 
 Test.prototype._setAssertModule = function (assertModule) {
 	var lib = assertModule || assert;
-	Object.keys(lib).forEach(function (el) {
-		Test.prototype[el] = function () {
-			this._assert();
 
-			try {
-				lib[el].apply(lib, arguments);
-			} catch (err) {
-				this.assertError = err;
-			}
-		};
-	});
+	if (typeof lib === 'function') {
+		setPrototype(lib.name, lib);
+	} else if (typeof lib === 'object') {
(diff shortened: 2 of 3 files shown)


Qix-:
:O Yay. I don't know enough about the code base to see whether or not this is correct; @vdemedes and/or @sindresorhus will look it over, but this looks exciting.

vdemedes:
@jenslind First of all, thanks for PR! How would you implement a support for `should` interface?

```js
a.should.equal(b);
should.not.exist(c);
```

lib/runner.js
```diff
@@ -118,3 +118,8 @@ Runner.prototype.run = function () {
 			stats.passCount = stats.testCount - stats.failCount;
 		});
 };
+
+// Set custom assert module
+Runner.prototype.setAssertModule = function (assertModule) {
+	Test._setAssertModule(assertModule);
```
vdemedes: Why have 2 functions for a one task? You can just leave it as `setAssertModule`.

lib/test.js
```diff
@@ -35,17 +35,20 @@ Test.prototype._assert = function () {
...
+			this._assert();
+
+			try {
+				lib[el].apply(lib, arguments);
+			} catch (err) {
+				this.assertError = err;
+			}
+		};
+	});
+})();
```
vdemedes: What is a self-executing function for?

Qix-:
Should in and of itself doesn't have `.plan()`. It's on their radar, but it doesn't as of now (unfortunately). The question to be asked is what to do when `plan()` is called. Doing *nothing* isn't ideal.

lib/runner.js
jenslind: You mean like this?
```
Runner.prototype.setAssertModule = Test.setAssertModule;
```

lib/test.js
jenslind: To set the default assert lib on start. Might not be the best solution.

sindresorhus: Yeah, would be better if `Test._setAssertModule` were explicitly called below.

sindresorhus:
@Qix- Most assert libs doesn't have a `.plan()`. That's usually part of the test runner. What asked here is how to mixin `should` methods, since they're chained on the native prototype of objects/string/etc. The native `assert` module works here, since it's only methods on an object. But it doesn't work for more advanced cases, like `should`. Maybe we should also have an adapter interface of some kind that can handle more advanced cases. How would that look like though?

Qix-:
`should` is actually just the `Should` object returned by `require('should')`, and a binding of it is added to `Object.prototype`. Doing whatever wrapping you have for `Should.prototype` should work, and propagate to the `Object.prototype` uses.

sindresorhus:
@Qix- Oh, I didn't realize. Yeah, that makes things easier. But there still might be other assertion libs where the easy way doesn't work. Maybe we should add support for the most popular assertion lib, so that you can just pass in the main export, and we'll handle whatever's nessecary to decorate it.

lib/runner.js
vdemedes: Oops, I missed that there are `Runner` and `Test` there, sorry. I would leave `setAssertModule` on `Runner` and when you create a new instance of `Test`, assign it there - https://github.com/sindresorhus/ava/blob/master/lib/runner.js#L34.

lib/test.js
vdemedes: You can do it in the constructor of a `Test`.

jenslind: @vdemedes most tests fails if I run it there.

vdemedes: In that case we need to know why it fails :)

jenslind: Yes ;) At work atm so haven't had time to investigate.

jenslind: Okey my misstake, tests passing now. But I have one concern about putting it in the constructor, it will run for every declared test. Not sure if we want that, ofc we could have some `bool` to check if  it has already been set but that doesn't make any sense to me.

vdemedes: That's totally fine, it is not a resource-expensive task.

jenslind:
Updated the PR. The following should now work:
```
test('chai', function (t) {
  t.plan(2);
  t.assert.equal(true, true);
  t.expect(true).to.equal(true);
})

test('should', function (t) {
  t.plan(1);
  t.should(true).equal(true);
})

test('assert', function (t) {
  t.plan(1);
  t.equal(true, true);
})
```
Implementing support for `a.should.equal(b)` will need some more work. :cat2:

lib/test.js
```diff
@@ -35,17 +43,28 @@ Test.prototype._assert = function () {
...
-});
+	if (typeof lib === 'function') {
+		setPrototype(lib.name, lib);
+	} else if (typeof lib === 'object') {
+		Object.keys(lib).forEach(function (el) {
+			setPrototype(el, lib[el]);
+		});
+	}
+
+	function setPrototype(el, method) {
```
vdemedes: Functions inside functions is a really no-go, please separate it out.

lib/test.js
```diff
@@ -4,14 +4,19 @@ var setImmediate = require('set-immediate-shim');
 var fnName = require('fn-name');
 var assert = require('./assert');
 
-function Test(title, fn) {
+function Test(title, assertModule, fn) {
```
vdemedes: `title` and `fn` are more common properties, while `assertModule` is absolutely optional. I think it needs to be the last one in the argument list.

vdemedes:
@jenslind In my opinion, `t.should`, `t.expect`, `t.assert` is really not a desired solution for supporting custom assert modules. It shouldn't assign those modules to a test context. All it has to do is to patch them with an assert count for `t.plan()` to work.

jenslind:
@vdemedes hmm okey so how would you patch the modules? Should we rather aim to add a counter to the module itself and then let `t.plan()` get the count from the module? But how would that work? A counter on the module won't know about which test it's in.

Qix-:
Remember - not all modules report *successful* asserts, which is something required for `t.plan()` to work. For instance, this was the problem with `should` - it doesn't, by default, report successful asserts.

sindresorhus:
The intention was to decorate the assert methods with an assertion counter, not mix them into the test context. I guess we could expose a `t.countAssert()` (better name?) or something too.

vdemedes:
I'd implement something like the following.

To use **should**:

```js
var should = chai.should();
var ava = require('ava');

ava.should(should);
```

To use **expect**:

```js
var expect = require('chai').expect;
var ava = require('ava');

ava.expect(expect);
```

or any other module, which does not require custom magic like `should` or `expect` does:

```js
var assert = require('assert');
var ava = require('ava');

ava.assert(assert);
```

To avoid duplication, we could prefix those methods with `setup`:

```js
ava.setupShould(should);
ava.setupExpect(expect);
ava.setupAssert(assert);
```

Qix-:
That seems messy for two reasons:

1. `should(should)` is redundant. What else would you pass to `should()`?
2. If you take that out, then you have to somehow enforce `should` be installed before calling `ava.should()`, unless you package all of them together, which... don't. :)

sindresorhus:
> should(should) is redundant. What else would you pass to should()?

It's to be explicit. If we only have `ava.setupAssert`, we would have to detect `should`. Is that possible? If yes, how? Having only one setup method would indeed be nicer.

> If you take that out, then you have to somehow enforce should be installed before calling ava.should(), unless you package all of them together, which... don't. :)

Not sure I understand what you mean, but this is only for mixing in the assert counter into the assert lib, not the other way. After calling `ava.setupAssert(should)`, you would just call `should` normally, not through AVA.

-

I want to get this moving. What's the minium thing we can do first? Maybe ignore the should/except stuff and just expose a generic `ava.countAssert()` and `ava.setupAssert()`.

sindresorhus:
Sorry @jenslind. Seems we aren't going to be able to reach a good solution for this right now. Closing for now, but happy to open it again when we have a decision on how to handle custom assertion libs.

Let's continue the discussion in #25.

PR #51, opened 2015-09-15
Easier debugging when not overwriting already existing assertErrors

joakimbeng:
Having a test with a `plan` which throws an unexpected error like so:

```javascript
var test = require('ava');

test(function unexpected(t) {
  t.plan(1);
  var o = {};
  o();
});
```

Will at the moment only report an error about assert count not matching planned count, even though an `assertError` already have been set. This addresses that by not overwriting an existing assert error, which gives the correct error message:

```
TypeError: o is not a function
```

What do you think?

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -107,7 +107,7 @@ Test.prototype.exit = function () {
 	// calculate total time spent in test
 	this.duration = Date.now() - this._timeStart;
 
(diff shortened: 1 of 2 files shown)


sindresorhus:
:+1:

vdemedes:
Good one!

PR #52, opened 2015-09-16
workarounds for browserify

### LINKED ISSUE #24: Browser support
Trying to get this working in the browser, and then later [hihat](https://github.com/Jam3/hihat) for debugging/breakpoints/etc.

After browserifying the demo in the readme, I get the following:

`Uncaught ReferenceError: setImmediate is not defined`

[relevant line](https://github.com/sindresorhus/ava/blob/97add1fcec62d510d72f1d26e1ea5ca972c495c6/index.js#L65)

(weird cause I thought browserify would have shimmed this)

danielepolencic:
fixes #24 

- if `stderr` is not available, inject the console in `squeak`
- `process.exit` is not available in the browser. Don't call the method if it is not available.

### DIFF
--- a/index.js
+++ b/index.js
@@ -84,7 +84,9 @@ function exit() {
 		stack(results);
 	}
 
-	process.exit(stats.failCount > 0 ? 1 : 0);
+	if (process.exit) {
+		process.exit(stats.failCount > 0 ? 1 : 0);
+	}
 }
 
 setImmediate(function () {
--- a/index.js
+++ b/index.js
@@ -6,7 +6,8 @@ var figures = require('figures');
 var Squeak = require('squeak');
 var plur = require('plur');
 var Runner = require('./lib/runner');
-var log = new Squeak({separator: ' '});
+var consoleStream = require('console-stream');
+var log = new Squeak({separator: ' ', stream: process.stderr || consoleStream()});
 var runner = new Runner();
 
 Error.stackTraceLimit = Infinity;
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "babel-core": "^5.8.23",
     "bluebird": "^2.9.34",
     "chalk": "^1.0.0",
+    "console-stream": "^0.1.1",
     "figures": "^1.3.5",
     "fn-name": "^2.0.0",
     "globby": "^3.0.1",


sindresorhus:
I still think this is better fixed in `squeak`: https://github.com/sindresorhus/ava/issues/24#issuecomment-140687957

index.js
```diff
@@ -83,7 +84,9 @@ function exit() {
 		stack(results);
 	}
 
-	process.exit(stats.failCount > 0 ? 1 : 0);
+	if (process.exit) {
```
sindresorhus: How will this work with `hihat`. According to its docs:

> The process will stay open until you call window.close()

Wouldn't it be better to call `window.close` here if in the `browser`. Or even better, keep it as before, and have `hihat` shim `process.exit` => https://github.com/Jam3/hihat/issues/27

sindresorhus:
// @mattdesl

mattdesl:
Hmm.. `window.close` would also work with [smokestack](https://github.com/hughsk/smokestack) but it seems like a pretty big assumption. 

In `hihat` the goal is interactive testing (fast save -> reload workflow). If `ava` exits the process on fail, it would break this development cycle. Maybe `ava` should not quit at all in the browser?

sindresorhus:
@mattdesl Good point. So what should we do here?

Qix-:
Hmm. Perhaps `ava.on('exit', ...)`, where ava detects if it's in node and registers the default handler of `process.exit()`? Let test implementers decide; we can add sensible defaults as necessary.

Qix-:
Or even shim `process.exit()`? It fires the `exit` event (`process.on('exit', ...)`) so that might even be the ideal way anyway to keep full compatibility.

vdemedes:
Agree with @mattdesl, on client-side `process.exit()` should do nothing or send an `exit` event, but definitely not close a window.

mattdesl:
It seems sorta weird that an API would ever call `process.exit`. I feel like this should be left up to the user.

I feel like only the CLI should exit and handle errors itself, since in that case the user isn't handling them.

vdemedes:
@mattdesl this happens, because we want to cover a case when test is being executed directly.

```
$ node test-something.js
```

So if only CLI would use `process.exit()`, in the above case process would exit always with zero (success).

mattdesl:
I see. It's a bit of a special case then.

arthurvr:
> Or even shim process.exit()? It fires the exit event (process.on('exit', ...)) so that might even be the ideal way anyway to keep full compatibility.

To me that sounds the most reasonable way to me. Though I guess this is kind of a rare case anyways.

Qix-:
It is, but it's still probably the *correct* way. Just a thought.

scottcorgan:
Just curious if there's been any progress for the browser. Only thing keeping me from Ava.

sindresorhus:
@kevva said he would make `squeak` browserifyable and this PR needs to be updated to only include the exit change.

sindresorhus:
Might be more, though, as we recently introduced parallism, which forks the processes. Not sure whether browserify handles that or if we need to do some special casing. Help welcome. Our main focus is elsewhere.

kevva:
Should be fixed in https://github.com/kevva/squeak/commit/fd7333e343b1ff5f73cd14736ccc049c45dead3b.

kevva:
> ... this PR needs to be updated to only include the exit change.

So what should we decide on? Just check for `process.exit` and do nothing if it doesn't exist?

sindresorhus:
Closing in favor of #97.

PR #59, opened 2015-09-18
use the test file path in the Babel `only` prop instead of a pattern

sindresorhus:
since we're only using this once per file now

http://babeljs.io/docs/usage/options/#options

### DIFF
--- a/lib/babel.js
+++ b/lib/babel.js
@@ -4,9 +4,10 @@ var resolveFrom = require('resolve-from');
 var createEspowerPlugin = require('babel-plugin-espower/create');
 
 var hasGenerators = parseInt(process.version.slice(1), 10) > 0;
+var path = process.argv[2];
 
 var options = {
-	only: /(test|test\-.+|test\/.+)\.js$/,
+	only: path,
 	blacklist: hasGenerators ? ['regenerator'] : [],
 	optional: hasGenerators ? ['asyncToGenerator'] : [],
 	plugins: [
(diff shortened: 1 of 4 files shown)


sindresorhus:
@vdemedes Any thoughts on how to best test to ensure that this is working correctly? Meaning it only transpiles the tested files and nothing more.

vdemedes:
Well, if we are executing babel per file, we should just set file name to `only` property.
`whitelist` property is not for files, it is a list of transform to include.

vdemedes:
@sindresorhus Does passing a string to `only`, instead of a pattern, works?

sindresorhus:
By whitelist I meant `only`.

sindresorhus:
> Does passing a string to only, instead of a pattern, works?

It should in theory as the prop is being arrayified, but we need a test for the behavior anyways.

ahdinosaur:
what about for projects that are using `babel` to transpile the rest of their app (e.g. by running their code with `babel-node`), wouldn't this force babel to only transpile the test files?

sindresorhus:
> what about for projects that are using babel to transpile the rest of their app (e.g. by running their code with babel-node), wouldn't this force babel to only transpile the test files?

We already limit that in master. This PR only changes it to only ignore the path explicitly instead of using a ignore pattern. You would just do whatever you with your existing test framework that doesn't support ES2015 builtin, not sure what that is?

PR #61, opened 2015-09-20
Use `co` for running generator functions

vdemedes:
Bluebird's `coroutine()` turned out to be pretty minimal in its functionality. It can only `yield` promises. `co` is more mature with handling generators. In addition, it supports `yield`ing not just promises, but other generator functions too, which is essential.

I also added `regenerator` to babel's blacklist to not transform generators into ugly ****.


vdemedes:
I considered using `Bluebird.coroutine.addYieldHandler()` at first. We would need to detect type of input and return another coroutine or promise afterwards. So I decided to ditch it and go with battle-tested `co`.

PR #65, opened 2015-09-23
Display error message when there are no test files

### LINKED ISSUE #55: Crash when setting an non-existent path
```
$ ava does-not-exist
```

vdemedes:
Fixes #55 

Output:

screen shot 2015-09-23 at 9 21 37 pm

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -106,6 +106,11 @@ function init(files) {
 			return path.join(process.cwd(), file);
 		})
 		.then(function (files) {
+			if (files.length === 0) {
+				log.error('Couldn\'t find any files to test\n');
+				process.exit(1);
+			}
+
 			var tests = files.map(run);
 
 			return Promise.all(tests);


cli.js
```diff
@@ -106,6 +106,12 @@ function init(files) {
 			return path.join(process.cwd(), file);
 		})
 		.then(function (files) {
+			if (files.length === 0) {
+				log.error('Couldn\'t find any files to test');
```
sindresorhus: `log.error('Couldn\'t find any files to test\n');` instead of the extract `log.write()` call.

cli.js
```diff
@@ -106,6 +106,11 @@ function init(files) {
 			return path.join(process.cwd(), file);
 		})
 		.then(function (files) {
+			if (files.length === 0) {
```
Qix-: @sindresorhus don't we go for `!files.length`? Or am I just thinking back to my C days?

sindresorhus: I prefer to be explicit.

Qix-: Fair enough.

PR #66, opened 2015-09-23
Prefix test title with test location

### LINKED ISSUE #53: Prefix test results with filename when running multiple files
We should prefix the tests with the filename minus extension and prefix `test-` so [this](https://github.com/sindresorhus/got/pull/104/files#diff-4b237e7473a059a1f82a01d75d69c5f1R26) isn't needed.

`test-url.js` with `test('foo bar')` would be:

```
✔ url - foo bar
```

-

// @floatdrop

vdemedes:
Fixes #53 

Uses `figures.pointerSmall` to replace `/` and as a separator between prefix and test title.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -2,6 +2,7 @@
 'use strict';
 var fs = require('fs');
 var path = require('path');
+var figures = require('figures');
 var flatten = require('arr-flatten');
 var globby = require('globby');
 var meow = require('meow');
@@ -33,6 +34,7 @@ var cli = meow({
 	string: ['_']
 });
 
+var fileCount = 0;
 var errors = [];
 
 function error(err) {
@@ -40,15 +42,45 @@ function error(err) {
 	process.exit(1);
 }
 
+function prefixTitle(file) {
+	var separator = ' ' + chalk.gray.dim(figures.pointerSmall) + ' ';
+
+	var base = path.dirname(cli.input[0]);
+
+	if (base === '.') {
+		base = cli.input[0] || 'test';
+	}
+
+	base += path.sep;
+
+	var prefix = path.relative(process.cwd(), file)
+		.replace(/test\-/g, '')
+		.replace(/\.js$/, '')
+		.replace(base, '')
+		.replace(new RegExp(path.sep, 'g'), separator);
+
+	if (prefix.length > 0) {
+		prefix += separator;
+	}
+
+	return prefix;
+}
+
 function test(data) {
 	var isError = data.err.message;
 
+	var prefix = '';
+
+	if (fileCount > 1) {
+		prefix = prefixTitle(data.file);
+	}
+
 	if (isError) {
-		log.error(data.title, chalk.red(data.err.message));
+		log.error(prefix + data.title, chalk.red(data.err.message));
 
 		errors.push(data);
 	} else {
-		log.test(null, data.title, data.duration);
+		log.test(null, prefix + data.title, data.duration);
 	}
 }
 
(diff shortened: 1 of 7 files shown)


cli.js
```diff
@@ -43,12 +45,24 @@ function error(err) {
 function test(data) {
 	var isError = data.err.message;
 
+	var prefix = '';
+
+	if (fileCount > 1) {
+		var separator = ' ' + figures.pointerSmall + ' ';
+
+		prefix = data.file
+			.replace(process.cwd() + '/', '')
```
Qix-: Should use `path.relative()` here.

cli.js
```diff
@@ -43,12 +45,24 @@ function error(err) {
...
+			.replace(/test\-|test\//g, '')
```
Qix-: If the directory has `test-` then that's a problem. If they're on windows, `\/` might be a problem.

cli.js
```diff
@@ -43,12 +45,24 @@ function error(err) {
...
+			.replace(/\.js$/, '')
```
Qix-: Use `path.basename(..., '.js')`.

cli.js
```diff
@@ -43,12 +45,24 @@ function error(err) {
...
+			.replace(/\//g, separator) + separator;
```
Qix-: Use `path.sep` here.

vdemedes:
@Qix- all good points, I should not submit PRs after 10pm. Will review this again tomorrow, thanks!

cli.js
```diff
@@ -33,6 +34,7 @@ var cli = meow({
 	string: ['_']
 });
 
+var fileCount = 0;
```
Qix-: Is this going to pose some sort of problem with concurrency? // @sindresorhus

vdemedes: No issues here, since it is in `cli.js`. This is an entrypoint file.

Qix-: :+1: disregard then.

Qix-:
Haha no problem; common path stuff.

vdemedes:
Added changes to `babel.js` to detect if generators are supported. If they are supported, babel's gonna use `asyncToGenerator` transform and blacklist `regenerator` transform. Before that, tests on 0.10 and 0.12 failed, because they don't support generators.

Qix-:
Lgtm :+1:  @sindresorhus will be the ultimate decider.

cli.js
```diff
@@ -40,15 +42,31 @@ function error(err) {
 	process.exit(1);
 }
 
+function prefixTitle(file) {
+	var separator = ' ' + figures.pointerSmall + ' ';
```
sindresorhus: `var separator = ' ' + chalk.dim(figures.pointerSmall) + ' ';`

sindresorhus:
The prefixing doesn't seem to work when specifying a direct path:

```
$ ava test/foo/http.js

  ✔ test title
```

Expected:

```
$ ava test/foo/http.js

  ✔ foo › http › test title
```

vdemedes:
@sindresorhus This PR is related to issue #53, which is "Prefix test results with filename when running multiple files". So this output is expected. When running a single test file, I don't think we need prefixes, since direct path is specified.

Qix-:
@vdemedes while I agree that that is sound behavior, I'm with @sindresorhus on it; consistency is important.

sindresorhus:
Oh right. That makes sense. If only one file, it should not show the path.

sindresorhus:
Super work @vdemedes. I really like this feature! :)

sindresorhus:
tumblr_lxx3edkjhr1qdlh1io1_400

Qix-:
That gif is life... :baby:

Issue #15, opened 2015-01-27
Emit error when there are more assertions than planned


kevva:
Currently we aren't keeping track of all the pending assertions. `tape` is using this https://github.com/substack/tape/blob/master/lib/test.js#L174-L181.

mdibaiee:
We do emit an error when the assertion count doesn't match planned. Why do we need the `_pendingAsserts` in the first place?

I have probably misunderstood this bug, can you please clarify? @kevva

sindresorhus:
@mdibaiee I think it's related to this: https://github.com/sindresorhus/ava/blob/ff5eebba5d06ea22a1ed1ca5f76f33386bee8381/test/test.js#L245-L247

mdibaiee:
@sindresorhus: I see, I'll take a look to see if I can fix this when I get some free time.

Qix-:
If I'm understanding this correctly, the way I've seen it done before is even if `t.end()` has been called and the test passed, an assertion called after that point will re-show the test as failed. You don't need to know the point of last assertion, and you will *never* know thanks to the halting problem.

sindresorhus:
@mdibaiee Did you have a chance to look into it? No worries if not :)

mdibaiee:
@sindresorhus: Hey, sorry I had forgot about this, busy on a few projects.

I took a look, couldn't find a solution, it *might* be possible if we have a timeout, but then, we can't just wait for the timeout to pass for every test only because it might be doing something wrong.

A funny idea hit my head! Inspecting the test body's code, searching for `<first-argument>.assert` and similar method calls :laughing:

Qix-:
> A funny idea hit my head! Inspecting the test body's code, searching for <first-argument>.assert and similar method calls :laughing:

Going to assume that was a joke :sos: 

---

If things are blocking and async functions are still at play, node won't terminate. You don't have to wait for anything unless it's a faulty test.

mdibaiee:
@Qix- yeah that was a joke.

How can we know if there are async functions still in progress? I tried calling `process._getActiveHandles()` and `process._getActiveRequests()` but these functions seem to only work for IO handles/requests.

Qix-:
Because node won't terminate unless they're done executing.

PR #67, opened 2015-09-24
Wait for pending tests

### LINKED ISSUE #63: Test silently ignored
Found here - https://github.com/kevva/download/pull/77#issuecomment-141836144

Consider this test file: 

```js
var test = require('ava');

test('no asserts', function (t) {
    // still passes
});
```

`ava` will exit silently without `no asserts` test in output.

vdemedes:
Fixes #63, related to #15.

### Changes

Before test `.run()`, timer with maximum possible value is started to keep the process alive. When the test is finished, `clearTimeout()` removes the timeout.

### Fixed cases

Now, the following tests won't just produce empty output:

```js
test('empty test', function (t) {

});

test('no assertions', function (t) {
  t.plan(1);
});
```

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -6,6 +6,8 @@ var fnName = require('fn-name');
 var assert = require('./assert');
 var co = require('co');
 
+var TIMEOUT_MAX_VALUE = 2147483647;
+
 function Test(title, fn) {
 	if (!(this instanceof Test)) {
 		return new Test(title, fn);
@@ -75,6 +77,9 @@ Test.prototype.run = function () {
 
 		this._timeStart = Date.now();
 
+		// wait until all assertions are complete
+		this._timeout = setTimeout(function () {}, TIMEOUT_MAX_VALUE);
+
 		try {
 			var ret = this.fn(this);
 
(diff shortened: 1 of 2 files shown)


lib/test.js
```diff
@@ -75,6 +75,9 @@ Test.prototype.run = function () {
 
 		this._timeStart = Date.now();
 
+		// wait until all assertions are complete
+		this._timeout = setTimeout(function () {}, 2147483647);
```
sindresorhus: Can you make the number a const: `TIMEOUT_UPPER_LIMIT` or something.

sindresorhus:
Can you reenable thsi test: https://github.com/sindresorhus/ava/blob/5dc0873b74d59a3f3cc33a36020c78cb305153aa/test/test.js#L262 and add some more?

vdemedes:
https://github.com/sindresorhus/ava/blob/5dc0873b74d59a3f3cc33a36020c78cb305153aa/test/test.js#L262 this test is not related to this PR (this PR won't fix this problem).

test/test.js
```diff
@@ -470,3 +470,22 @@ test('async/await support', function (t) {
...
+
+		avaTest = a;
+	}).run().then(function (a) {
+		t.is(a.planCount, 1);
+		t.is(a.assertCount, 1);
+		t.true(a.duration >= 1234);
+		t.end();
+	});
+
+	setTimeout(function () {
```
floatdrop: Is this test relevant? `setTimeout` will block Node from exit anyway.

vdemedes: If test wouldn't wait for pending assertion, `a.duration` would be (or close to) zero. So if `a.duration` is more than `1234`, it means test was waiting.

sindresorhus:
:+1: :tada:

PR #71, opened 2015-10-04
Add beforeEach/afterEach hooks

vdemedes:
This PR adds `beforeEach()` and `afterEach()` methods to Ava.

As names hint, these functions add hooks that are executed before & after each test.

There is one difference though, if `beforeEach` hook fails, tests don't stop (unlike `before()`).

So, let's discuss this!


vdemedes:
I am now thinking that these hooks must run serially, just like `before()` and `after()`.

vdemedes:
Closing this (until fixed) as I discovered more issues.

PR #72, opened 2015-10-04
Add beforeEach/afterEach hooks

vdemedes:
This PR adds `beforeEach()` and `afterEach()` methods to Ava.

These functions allow to add hooks that are executed before & after each test.
`beforeEach` and `afterEach` hooks execute serially, just like `before` and `after`.

There is one difference though, if `beforeEach` hook fails, tests don't stop (unlike `before`).

So, let's discuss this!

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -4,6 +4,14 @@ var EventEmitter = require('events').EventEmitter;
 var Promise = require('bluebird');
 var Test = require('./test');
 
+function each(items, fn) {
+	return Promise.all(items.map(fn));
+}
+
+function eachSeries(items, fn) {
+	return Promise.resolve(items).each(fn);
+}
+
 function Runner(opts) {
 	if (!(this instanceof Runner)) {
 		return new Runner(opts);
@@ -22,7 +30,9 @@ function Runner(opts) {
 		concurrent: [],
 		serial: [],
 		before: [],
-		after: []
+		after: [],
+		beforeEach: [],
+		afterEach: []
 	};
 }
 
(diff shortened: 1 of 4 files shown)


lib/runner.js
```diff
@@ -47,36 +49,54 @@ Runner.prototype.addAfterHook = function (title, cb) {
 	this.tests.after.push(new Test(title, cb));
 };
 
-Runner.prototype.concurrent = function (tests) {
-	var self = this;
+Runner.prototype.addBeforeEachHook = function (title, cb) {
+	this.tests.beforeEach.push({
```
vdemedes: I am not doing `new Test(title, cb)` here, because we need multiple instances of that test for `*each` hooks. So instead I am storing title and function and creating `Test` instances later.

lib/runner.js
```diff
@@ -47,36 +49,54 @@ Runner.prototype.addAfterHook = function (title, cb) {
...
+		return new Test(hook.title, hook.fn);
+	});
+
+	var tests = [];
+
+	tests.push.apply(tests, beforeHooks);
+	tests.push(test);
+	tests.push.apply(tests, afterHooks);
+
+	return eachSeries(tests, function (test) {
```
vdemedes: Executing `beforeEach` and `afterEach` hooks serially.

lib/runner.js
```diff
@@ -118,3 +137,11 @@ Runner.prototype.run = function () {
 			stats.passCount = stats.testCount - stats.failCount;
 		});
 };
+
```
vdemedes: Created these helpers to DRY up.

vdemedes:
I will add info to readme when this PR will be complete ;)

lib/runner.js
```diff
@@ -118,3 +138,11 @@ Runner.prototype.run = function () {
...
+function each(items, fn) {
+	return Promise.all(items.map(fn));
+}
+
+function eachSeries(items, fn) {
+	return Promise.resolve(items).each(fn);
+}
```
sindresorhus: Can you put these at the top of the file or at least above where they're used?

sindresorhus:
:+1: This looks very good! Pretty much exactly how I would have done it :)

lib/runner.js
vdemedes: Sure, np

sindresorhus:
> There is one difference though, if beforeEach hook fails, tests don't stop (unlike before).

What's the reasoning for that? Since the test depend on the beforeEach hook, and if that fails silently, the test might end in a false positives.

vdemedes:
Glad you like it!

    What's the reasoning for that? Since the test depend on the beforeEach hook, and if that fails silently, the test might end in a false positives.

Just did not implement that yet, maybe I should include it in this PR ;)

sindresorhus:
Yeah, I would personally want it to fail. Failing hooks should be the same as a failing tests. Hooks are really just implicit function calls:

```js
test(t => {
  before();
  t.pass();
  after();
});

**Comment by vdemedes** [comment]
Updated to fail when `beforeEach` hook fails ;)

File: test/test.js [hunk 5]
```diff
@@ -448,6 +448,162 @@ test('hooks - stop if before hooks failed', function (t) {
...
+		a.end();
+	});
+
+	runner.addTest(function (a) {
+		arr.push('b');
+		a.pass();
+		a.end();
+	});
+
+	runner.run().then(function () {
```
sindresorhus: Can you add an `err` check here too? To make sure the `fail` is propagated.

sindresorhus:
One tiny inline comment and needs docs, otherwise ready to be merged.

vdemedes:
Added info to readme

sindresorhus:
Landed. Woohooo! The feature set is really coming together now.

a77b1e625aa3d9739706cfe23888917d

kasperlewau:
Any ETA on when this might be published?

Currently running with `$ npm install git+ssh://git@github.com/sindresorhus/ava.git` (*which isn't that big of a deal*), though I did get a wee bit stumped after having referred to the README and then receiving `TypeError: _ava2.default.beforeEach is not a function` :smile:.

sindresorhus:
When we're ready to do another release. Probably in a few days. As with every other OSS project, the readme is master, if you want to see docs for a specific release go to the git tag or the npm page.

kasperlewau:
> When we're ready to do another release. Probably in a few days.

Coolio, thanks!

Issue #73, opened 2015-10-06
throws assertion does not work with await/async


SamVerschueren:
When using `await`/`async`, the throws() assertion doesn't work.

```js
test('throws error', async t => {
    t.throws(await fn(), Error);
});
```

vdemedes:
`throws()` expects a function you are executing to throw. This will not work here, because `async/await` is converted to generator functions, thanks to `asyncToGenerator` babel transform. So, your example is transformed to this:

```js
test('throws error', function * (t) {
  t.throws(yield fn(), Error);
});
```

For `t.throws()` to catch your error, it needs to check if promise rejects and that's not supported at the moment.

sindresorhus:
> For t.throws() to catch your error, it needs to check if promise rejects and that's not supported at the moment.

Could be useful though. Let's leave this as a feature request?

SamVerschueren:
It would make testing for an error really nice and clean.

vdemedes:
@sindresorhus Yep, let's!

silvenon:
I tried the provided solution, but it doesn't work for me:

```js
test(function * (t) {
  t.throws(yield Promise.reject(new Error()), Error);
});
```

It doesn't output the test. Am I doing something wrong?

floatdrop:
@SamVerschueren why not try-catch?

```js
test('throws error', async t => {
    try { await fn(); } catch (err) { /* ... */ }
});
```

silvenon:
@floatdrop that works, thanks :+1:

SamVerschueren:
@floatdrop Yes I know that's possible, and I already do that at the moment. But I think it would be a very nice addition that the `throws` assertion method can handle rejected promises. This would make some tests methods a lot cleaner.

```js
test('throws error', async t => {
    t.throws(await fn(), Error);
});

test('throws error 2', async t => {
    try {
        await fn();
        t.fail();
    } catch (err) {
        t.ok(err);
    }
});
```

sindresorhus:
In `got`, we have to [repeat `t.fail('Exception is not thrown');`](https://github.com/sindresorhus/got/commit/f7500d474e4f54003682c998d81a3aa9ce61612c#diff-bbc23533bfd4986af19324ccf2e7c4baR27) in a lot of tests, which is pretty sub-optimal.

sindresorhus:
Anyone wanna try implementing this?

vdemedes:
Submitting a PR in a minute ;)

[...]

PR #79, opened 2015-10-17
Handle promises in throws() and doesNotThrow()

vdemedes:
This PR introduces the ability to handle promises in `t.throws()` and `t.doesNotThrow()` methods.

Refs #73.

Changes:

- In `lib/test.js` check, if assertion function returns promise and act accordingly
- Improve `t.throws()` and `t.doesNotThrow()` methods to handle promises
- Change test names related to functions above to be more clear

Let me know what you think ;)

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -69,6 +69,24 @@ x.notSame = function (val, expected, msg) {
 };
 
 x.throws = function (fn, err, msg) {
+	if (fn && fn.then) {
+		err = err || Error;
+
+		return fn
+			.catch(err, function (err) {
+				return err;
+			})
+			.then(function (value) {
+				if (value instanceof err) {
+					return;
+				}
+
+				x.throws(function () {
+					throw value;
+				}, err, msg);
+			});
+	}
+
 	try {
 		assert.throws(fn, err, msg);
 	} catch (err) {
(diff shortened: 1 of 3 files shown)


lib/assert.js
```diff
@@ -69,6 +69,22 @@ x.notSame = function (val, expected, msg) {
...
+		return fn
+			.catch(err, function (err) {
+				return err;
+			})
+			.then(function (value) {
+				if (value instanceof err) {
+					return;
+				}
+
+				test(false, create(undefined, undefined, undefined, msg || 'Missing expected exception', x.throws));
```
sindresorhus: Why is `actual` and `expected` undefined? Shouldn't we follow the sync method: https://github.com/sindresorhus/core-assert/blob/494c1eaea60090d4d4a187f03de9c2f4051b6f9b/index.js#L321

As for `msg`, see: https://github.com/sindresorhus/core-assert/blob/494c1eaea60090d4d4a187f03de9c2f4051b6f9b/index.js#L317-L318

Maybe it would be easier to just use `assert.throws(fn, err, msg);` here.

lib/test.js
```diff
@@ -41,12 +41,26 @@ Test.prototype._assert = function () {
 
 Object.keys(assert).forEach(function (el) {
 	Test.prototype[el] = function () {
-		this._assert();
+		var self = this;
 
 		try {
-			assert[el].apply(assert, arguments);
+			var promise = assert[el].apply(assert, arguments);
```
sindresorhus: `var promise` => `var fn`

sindresorhus:
// @silvenon @floatdrop @SamVerschueren @kevva

vdemedes:
PR updated

sindresorhus:
Looks good. I'll add some docs tomorrow.

SamVerschueren:
Awesome! Thanks @vdemedes for the PR and thanks @sindresorhus for merging! Will make testing for erros way cleaner!

PR #88, opened 2015-10-21
Allow es2015 sources

### LINKED ISSUE #87: Allow non-test files to take advantage of babel compilation
I'm using AVA to write test some ES2015 source code but I get an error every time I try to do so.

I believe the offending line is in [lib/babel.js](https://github.com/sindresorhus/ava/blob/e9b3e5019de71e00725fa77b4f705c64e86e99c3/lib/babel.js#L8)

Not sure what the appropriate way to expose this should be, other then simply removing that line.

Whoaa512:
Fix #87

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -23,6 +23,7 @@ var cli = meow({
 		'Options',
 		'  --init       Add AVA to your project',
 		'  --fail-fast  Stop after first test failure',
+		'  --es-src     Allow testing of ES2015 source files',
 		'',
 		'Examples',
 		'  ava',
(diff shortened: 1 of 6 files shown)


kevva:
Not so sure I agree with this. You'd still have to compile your files with Babel afterwards so why not setup a npm script to compile them before tests? See [this](https://github.com/sindresorhus/pageres/blob/master/package.json#L16-L17) for an example, and then `require` the compiled code in your tests.

vdemedes:
I agree with @kevva. 

I don't think this PR will get merged. In the related issue, @sindresorhus said that you shouldn't rely on built-in babel, it is just there for tests. Therefore, people disagree on this feature.

Whoaa512:
@kevva thanks for the example, I think this will suffice

PR #91, opened 2015-10-22
Fix t.throws() behavior to avoid false-positives and regex suport

### LINKED ISSUE #85: assert.throws and promises which resolve to error objects
[Currently](https://github.com/sindresorhus/ava/blob/d14c6e37fa4c9666599e78256cc0c9ccc39d45e5/lib/assert.js#L76-L87), the following test passes:
```js
test('resolve error value', t => {
	t.plan(1);
	t.throws(Promise.resolve(new Error('abc'))); // pass (false negative)
});
```

### LINKED ISSUE #86: assert.throws, promises and regex
[Currently](https://github.com/sindresorhus/ava/blob/d14c6e37fa4c9666599e78256cc0c9ccc39d45e5/lib/assert.js#L76-L87), the last test fails:

```js
test('throw Error and regex', t => {
	t.plan(1);
	t.throws(() => { throw new Error('abc'); }, /b/); // pass (as expected)
});

test('reject error', t => {
	t.plan(1);
	t.throws(Promise.reject(new Error('abc'))); // pass (as expected)
});

test('reject error and regex', t => {
	t.plan(1);
	t.throws(Promise.reject(new Error('abc')), /b/); // fails (false positive)
});
```

vdemedes:
This PR fixes #86 and fixes #85.

It improves `t.throws()`, so that it avoids false-positives results and supports regex, just like native `throws()`.

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -70,19 +70,21 @@ x.notSame = function (val, expected, msg) {
 
 x.throws = function (fn, err, msg) {
 	if (fn && fn.then) {
-		err = err || Error;
+		var isFailed = false;
 
 		return fn
-			.catch(err, function (err) {
+			.catch(function (err) {
+				isFailed = true;
+
 				return err;
 			})
-			.then(function (value) {
-				if (value instanceof err) {
-					return;
+			.then(function (fnErr) {
+				if (!isFailed) {
+					x.throws(function () {}, err, msg);
 				}
 
 				x.throws(function () {
-					throw value;
+					throw fnErr;
 				}, err, msg);
 			});
 	}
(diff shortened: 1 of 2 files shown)


sindresorhus:
Super! :)

schnittstabil:
@vdemedes Thanks a lot.

lib/assert.js
```diff
@@ -70,19 +70,21 @@ x.notSame = function (val, expected, msg) {
...
+			.then(function (fnErr) {
+				if (!isFailed) {
+					x.throws(function () {}, err, msg);
 				}
 
 				x.throws(function () {
-					throw value;
+					throw fnErr;
 				}, err, msg);
 			});
```
schnittstabil: @vdemedes This seems unnecessarily complicated to me. Just a suggestion:
```js
	if (fn && fn.then) {
		return fn
			.then(function () {
				x.throws(function () {}, err, msg);
			}, function (fnErr) {
				x.throws(function () {
					throw fnErr;
				}, err, msg);
			});
	}
```

PR #93, opened 2015-10-23
add xo coverage to the esnext fixtures

dylang:
I'm thinking of doing this for my own project `test` directories. The projects aren't `esnext` but the AVA tests are. 

This seemed more simple than implementing path-specific config in XO, and makes me feel better that I'm not lint-ignoring my tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "node": ">=0.10.0"
   },
   "scripts": {
-    "test": "xo && nyc tape test/*.js | tap-dot",
+    "test": "xo && xo --esnext true ./test/fixture/*.js && nyc tape test/*.js | tap-dot",
     "coveralls": "nyc report --reporter=text-lcov | coveralls"
   },
   "files": [


sindresorhus:
Using CLI flags will make it not co-operate with the editor plugins and other tools.

package.json
```diff
@@ -32,7 +32,7 @@
     "node": ">=0.10.0"
   },
   "scripts": {
-    "test": "xo && nyc tape test/*.js | tap-dot",
+    "test": "xo && xo --esnext true ./test/fixture/*.js && nyc tape test/*.js | tap-dot",
```
sindresorhus: No need for `true`.

sindresorhus:
I don't really see the need for this tbh. Hopefully ESLint will soon be switching to Acorn, which will give us better ES2015+ compat by default.

dylang:
@sindresorhus Thanks for the feedback!

PR #98, opened 2015-10-24
Support running the test file only via CLI

### LINKED ISSUE #97: Remove support for running the test file directly with node
It's nice in theory to be able to `node test.js`, but the reality is that it complicates the code a lot and by using it that way you miss out on the best features of AVA. It also create diverging code paths that just becomes harder to maintain.

This would also fix #52.

@vdemedes I guess we can remove https://github.com/sindresorhus/ava/blob/c96622d6d8785719d16aa83d129cb5ffccbf9add/index.js#L75-L86 now. Anything else that can be removed/simplified because of this?

vdemedes:
This PR closes #97.

After this PR, `node test.js` is not possible. Only `ava test.js`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,10 +2,8 @@
 var setImmediate = require('set-immediate-shim');
 var Runner = require('./lib/runner');
 var runner = new Runner();
-var log = require('./lib/logger');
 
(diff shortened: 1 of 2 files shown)


sindresorhus:
:+1:

vdemedes:
Cool!

PR #101, opened 2015-10-24
Add serial mode

### LINKED ISSUE #31: CLI `--debug` flag
For running everything serially so it's easier to debug concurrency issues.

In the future it might do other things too to aid in debugging. Suggestions welcome.

vdemedes:
This PR closes #31.

It adds `--serial` option, which forces AVA to run all tests serially by default.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -23,6 +23,7 @@ var cli = meow({
 		'Options',
 		'  --init       Add AVA to your project',
 		'  --fail-fast  Stop after first test failure',
+		'  --serial     Run tests serially',
 		'',
 		'Examples',
 		'  ava',
@@ -36,7 +37,7 @@ var cli = meow({
 	]
 }, {
 	string: ['_'],
-	boolean: ['fail-fast']
+	boolean: ['fail-fast', 'serial']
 });
 
 var fileCount = 0;
@@ -94,6 +95,10 @@ function run(file) {
 
 	if (cli.flags.failFast) {
 		args.push('--fail-fast');
+	}
+
+	if (cli.flags.serial) {
+		args.push('--serial');
 	}
 
 	return fork(args)
--- a/readme.md
+++ b/readme.md
@@ -92,6 +92,7 @@ $ ava --help
   Options
     --init       Add AVA to your project
     --fail-fast  Stop after first test failure
+    --serial     Run tests serially
 
   Examples
     ava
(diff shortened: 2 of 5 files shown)


cli.js
```diff
@@ -23,6 +23,7 @@ var cli = meow({
 		'Options',
 		'  --init       Add AVA to your project',
 		'  --fail-fast  Stop after first test failure',
+		'  --serial     Execute all tests serially',
```
sindresorhus: > Run tests serially

cli.js
```diff
@@ -96,6 +97,10 @@ function run(file) {
 		args.push('--fail-fast');
 	}
 
+	if (cli.flags.serial) {
+		args.push('--serial');
+	}
```
sindresorhus: Instead of doing this for every option I think we should just JSON.stringify the options and pass them as the first argument.

readme.md
```diff
@@ -468,6 +469,14 @@ Have fun!
 
 Running tests concurrently comes with some challenges, doing IO is one. Usually, serial tests just create temp directories in the current test directory and cleans it up at the end. This won't work when you run tests concurrently as tests will conflict with each other. The correct way to do it is to use a new temp directory for each test. The [`tempfile`](https://github.com/sindresorhus/tempfile) and [`temp-write`](https://github.com/sindresorhus/temp-write) modules can be helpful.
 
+### Debugging
+
+AVA runs tests concurrently by default, which is not optimal when you need to debug something. To make AVA run all tests serially, use `--serial` option:
```
sindresorhus: > AVA runs tests concurrently by default, which is suboptimal when you need to debug something. Instead, run tests serially with the `--serial` option:

test/fixture/serial.js
```diff
@@ -0,0 +1,26 @@
+import test from '../../';
+
+function randomDelay() {
+	return parseInt(Math.random() * 1000, 10);
+}
+
+let tests = [];
```
sindresorhus: Use `const`

`const` only means no new assignments, it's not immutable.

test/fixture/serial.js
```diff
@@ -0,0 +1,26 @@
...
+		tests.push('first');
+		t.end();
+	}, randomDelay());
+});
+
+test('second', t => {
+	setTimeout(() => {
+			tests.push('second');
+			t.end();
+		}, randomDelay());
```
sindresorhus: wrong indent

cli.js
vdemedes: Ooh, this will not look good... I propose to postpone this, until we have more options, what do you say?

sindresorhus: k

sindresorhus:
Awesome! Glad we could get this in :)

PR #103, opened 2015-10-25
Replace babel require hook with transformFileSync

floatdrop:
Start of discussion - https://github.com/sindresorhus/ava/commit/52b0ccb33ca30bba5185056361f1c2976b7cf56f#commitcomment-13971327

If you have `babel` and want to use require hook in application code and want to run tests with `ava` - you can get syntax errors (`SyntaxError: Unexpected reserved word`) because babel will get options from `ava`.

### DIFF
--- a/lib/babel.js
+++ b/lib/babel.js
@@ -1,11 +1,11 @@
 'use strict';
 var resolveFrom = require('resolve-from');
 var createEspowerPlugin = require('babel-plugin-espower/create');
+var requireFromString = require('require-from-string');
 var hasGenerators = parseInt(process.version.slice(1), 10) > 0;
 var path = process.argv[2];
 
 var options = {
-	only: path,
 	blacklist: hasGenerators ? ['regenerator'] : [],
 	optional: hasGenerators ? ['asyncToGenerator'] : [],
 	plugins: [
@@ -15,12 +15,14 @@ var options = {
 	]
 };
 
-try {
-	var localBabel = resolveFrom('.', 'babel-core/register') || resolveFrom('.', 'babel/register');
+var babel;
 
-	require(localBabel)(options);
+try {
+	var localBabel = resolveFrom('.', 'babel-core') || resolveFrom('.', 'babel');
+	babel = require(localBabel);
 } catch (err) {
-	require('babel-core/register')(options);
+	babel = require('babel-core');
 }
 
-require(path);
+var transpiled = babel.transformFileSync(path, options);
+requireFromString(transpiled.code, path);
--- a/lib/babel.js
+++ b/lib/babel.js
@@ -7,7 +7,7 @@ var path = process.argv[2];
 
 var options = {
 	blacklist: hasGenerators ? ['regenerator'] : [],
-	optional: hasGenerators ? ['asyncToGenerator'] : [],
+	optional: hasGenerators ? ['asyncToGenerator', 'runtime'] : ['runtime'],
 	plugins: [
 		createEspowerPlugin(require('babel-core'), {
 			patterns: require('./enhance-assert').PATTERNS
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "babel-runtime": "^5.8.29",
     "bluebird": "^3.0.0",
     "chalk": "^1.0.0",
-    "co": "^4.6.0",
+    "co": "floatdrop/co#e6c2a50",
     "core-assert": "^0.1.0",
     "empower": "^1.0.2",
     "figures": "^1.4.0",
--- a/package.json
+++ b/package.json
@@ -64,6 +64,7 @@
     "ava-init": "^0.1.0",
     "babel-core": "^5.8.23",
     "babel-plugin-espower": "^1.0.0",
+    "babel-runtime": "^5.8.29",
     "bluebird": "^3.0.0",
     "chalk": "^1.0.0",
     "co": "^4.6.0",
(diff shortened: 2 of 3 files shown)


lib/babel.js
```diff
@@ -1,13 +1,13 @@
...
 var createEspowerPlugin = require('babel-plugin-espower/create');
+var requireFromString = require('require-from-string');
 var hasGenerators = parseInt(process.version.slice(1), 10) > 0;
 var path = process.argv[2];
 
 var options = {
-	only: path,
 	blacklist: hasGenerators ? ['regenerator'] : [],
-	optional: hasGenerators ? ['asyncToGenerator'] : [],
+	optional: hasGenerators ? ['asyncToGenerator'] : ['runtime'],
```
sindresorhus: Runtime should still be included, even if `hasGenerators` is true.

sindresorhus:
Does the caching still work with this solution? Since Babel is relatively slow, the caching is very important.

floatdrop:
@sindresorhus I'm afraid, cache should be implemented by hands.

`co` must be replaced (or patched) to use Promise polyfill, since babel now will not do this for us.

vdemedes:
@floatdrop in order to support the stuff that "generator community" is used to, `co` is required. Babel's built-in implementation is very limited.

sindresorhus:
@floatdrop Maybe you could either extract [this](https://github.com/babel/babel/blob/332bde6a0dc30cfb6f89ac70b22f2b795a5e7a0b/packages/babel/src/api/register/node.js#L71-L107) into a separate module we can depend on or do a PR on Babel to expose it?

floatdrop:
@vdemedes yeah, I know, but you can't use `co` in Node.js 0.10 without some transpiler (and I don't think, that `co` will adapt to it soon https://github.com/tj/co/pull/244). So we need to come up with something. I believe, that transpiling only test code will improve performance compared to patched `require` function.

@sindresorhus okay, but need to do some performance tests - I wonder, how fast cache file will fill up and parsing time eat all the benefit.

vdemedes:
@floatdrop generators are only being used in node 0.11, iojs and node v4.x projects, so we don't have to worry about that.

floatdrop:
@vdemedes `co` uses global [Promise](https://github.com/tj/co/blob/master/index.js#L50) object - so you can't use it right away in Node.js 0.10.

vdemedes:
@floatdrop Hmm, did not understand the point. I am just saying that nobody uses generators on the old node.js versions.

floatdrop:
@vdemedes generators are not a problem, yes. But ava does [require co](https://github.com/sindresorhus/ava/blob/master/lib/test.js#L6) at the top of `test.js` and this leads to errors in `0.10` - https://travis-ci.org/sindresorhus/ava/jobs/87312158

floatdrop:
JFYI: seems like babel require hook have some polyfill side-effect - https://github.com/floatdrop/ava-in-node-0.10

floatdrop:
Just to keep track - waiting for https://github.com/tj/co/pull/250

sindresorhus:
Doesn't look like `co` has much activity, so I don't think we should wait on that tbh.

sindresorhus:
Why do we even depend on `co` again? Can't Babel+Regenerator do this for us?

vdemedes:
Because co implements some [very nice features](https://github.com/tj/co#yieldables), which babel's regenerator does not implement. Babel's implementation only makes Promises yieldable, but nothing else.

Why don't we use @floatdrop's fixed fork for now, until they merge his fix into master?

sindresorhus:
> Why don't we use @floatdrop's fixed fork for now, until they merge his fix into master?

:+1:

package.json
```diff
@@ -64,9 +64,10 @@
     "ava-init": "^0.1.0",
     "babel-core": "^5.8.23",
     "babel-plugin-espower": "^1.0.0",
+    "babel-runtime": "^5.8.29",
     "bluebird": "^3.0.0",
     "chalk": "^1.0.0",
-    "co": "^4.6.0",
+    "co": "git://github.com/floatdrop/co",
```
sindresorhus:
```
"co": "floatdrop/co#343xsv324",
```

use the shorthand format and use a specific commit hash (first 6 chars of the hash is enough), otherwise the npm cache will become stale.

floatdrop: :ok:

sindresorhus:
> okay, but need to do some performance tests - I wonder, how fast cache file will fill up and parsing time eat all the benefit.

@floatdrop I'm ok with merging this now if you commit to looking into caching later. Compilation is a lot larger overhead than just reading in a cached file from disk. I do think we can do caching better than Babel itself, though. Mainly using separate cache files for each test file (Babel uses one large file) and a [LRU cache](https://github.com/isaacs/node-lru-cache).

package.json
```diff
@@ -64,9 +64,10 @@
     "ava-init": "^0.1.0",
     "babel-core": "^5.8.23",
     "babel-plugin-espower": "^1.0.0",
+    "babel-runtime": "^5.8.29",
     "bluebird": "^3.0.0",
     "chalk": "^1.0.0",
-    "co": "^4.6.0",
+    "co": "floatdrop/co#343xsv324",
```
sindresorhus: lol, that was a fake hash, i don't know which hash you want to use.

floatdrop:
@sindresorhus sure. I'm started to use `ava` in some work related projects, so I will hang around for some time :)

sindresorhus:
> I'm started to use ava in some work related projects

:heart_eyes:

sindresorhus:
This looks good to me. @vdemedes Can you take a final look?

vdemedes:
Good to go! 👍

sindresorhus:
Thanks again @floatdrop :)

PR #106, opened 2015-10-25
Shared context between tests and beforeEach/afterEach hooks

vdemedes:
This PR resolves the following use case:

```js
test.beforeEach(t => {
  this.data = generateUniqueData();
});

test('some data', t => {
  // use this.data
});
```

With this PR, context (`this` object) is shared between test and **beforeEach**/**afterEach** hooks (not **before** or **after**).

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -90,10 +90,20 @@ Runner.prototype._runTestWithHooks = function (test) {
 	tests.push(test);
 	tests.push.apply(tests, afterHooks);
 
-	return eachSeries(tests, this._runTest.bind(this)).catch(noop);
+	var context = {};
+
+	return eachSeries(tests, function (test) {
+		return this._runTest(test, context);
+	}.bind(this)).catch(noop);
 };
 
-Runner.prototype._runTest = function (test) {
+Runner.prototype._runTest = function (test, context) {
+	// shared context (only applies to tests with hooks around them)
+	// set only if `context` is an object (can be index number)
+	if (typeof context === 'object') {
+		test.context = context;
+	}
+
 	// add test result regardless of state
 	// but on error, don't execute next tests
 	return test.run()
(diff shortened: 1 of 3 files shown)


sindresorhus:
Perfect.

schnittstabil:
Arrow functions work different! From [MDN](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Functions/Arrow_functions):
> An arrow function expression… and **lexically binds the this value** (does not bind its own this, arguments, super, or new.target).

An example:
```js
const fn = () => {this.unicorns = 'many'};
global.unicorns; // => undefined
fn();
global.unicorns; // => 'many'
```

Therefore, at least the examples are misleading.

sindresorhus:
@schnittstabil Good catch. That totally slipped my mind.

@vdemedes Maybe we should use a property on `t` for holding the context instead then? `t.context` or `t.data`?

schnittstabil:
Nice, `t.context` already works like expected:

```js
test(function(t) {
	t.is(t.context, this);
	t.end();
});
```

sindresorhus:
Oh, right. I guess all we have to do is revert https://github.com/sindresorhus/ava/pull/106/files#diff-dd2e881af3311f834199c7b515cff0c7R109 then.

Maybe we should make `t.context` non-writable, so people don't accidentally do:

```js
test.beforeEach(t => {
	t.context = 'foo';
	t.end();
});
```

schnittstabil:
> I guess all we have to do is revert…

We haven't to, but I think it is cleaner. 

> Maybe we should make t.context non-writable

That would be reasonable. Or we support it:
```js
test.beforeEach(t => {
	t.context = 'foo';
	t.end();
});

test('some data', t => {
	t.is(t.context, 'foo');
	t.end();
});
```

sindresorhus:
Yeah, I think it would be nicer if the user could use the context for whatever they wanted, it being an object or something else.

PR #119, opened 2015-11-01
Don't display test title, if there's only one anonymous test

### LINKED ISSUE #112: Don't show title when there's only one anonymous test
This used to work, but was disabled when we introduced the forking. We should bring it back.

https://github.com/sindresorhus/ava/blob/7a3f4444997099cd571118508f08783bce8a5526/lib/logger.js#L42-L44

vdemedes:
Fixes #112

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -43,6 +43,7 @@ var cli = meow({
 	]
 });
 
+var testCount = 0;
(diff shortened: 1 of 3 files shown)


sindresorhus:
Yay! Been missing this :)

PR #124, opened 2015-11-04
Refactor promises in throws method

SamVerschueren:
Regardless of issue https://github.com/sindresorhus/ava/issues/123, I think the promise code in `x.throws` might be written more cleaner.

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -70,19 +70,11 @@ x.notSame = function (val, expected, msg) {
 
 x.throws = function (fn, err, msg) {
 	if (fn && fn.then) {
-		var isFailed = false;
-
 		return fn
-			.catch(function (err) {
-				isFailed = true;
-
-				return err;
+			.then(function () {
+				x.throws(function () {}, err, msg);
 			})
-			.then(function (fnErr) {
-				if (!isFailed) {
-					x.throws(function () {}, err, msg);
-				}
-
+			.catch(function (fnErr) {
 				x.throws(function () {
 					throw fnErr;
 				}, err, msg);
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -73,8 +73,7 @@ x.throws = function (fn, err, msg) {
 		return fn
 			.then(function () {
 				x.throws(function () {}, err, msg);
-			})
-			.catch(function (fnErr) {
+			}, function (fnErr) {
 				x.throws(function () {
 					throw fnErr;
 				}, err, msg);


SamVerschueren:
This is odd, because of this minor change, 40 tests are failing.

schnittstabil:
@SamVerschueren see my comment [here](https://github.com/sindresorhus/ava/pull/91) if you want to fix that.

lib/assert.js
```diff
@@ -70,19 +70,11 @@ x.notSame = function (val, expected, msg) {
...
+			.then(function () {
+				x.throws(function () {}, err, msg);
 			})
-			.then(function (fnErr) {
-				if (!isFailed) {
-					x.throws(function () {}, err, msg);
-				}
-
+			.catch(function (fnErr) {
 				x.throws(function () {
```
schnittstabil: That will catch errors from `fn` **and** errors thrown at line 75…

lib/assert.js
```diff
@@ -70,19 +70,11 @@ x.notSame = function (val, expected, msg) {
...
-				return err;
+			.then(function () {
+				x.throws(function () {}, err, msg);
 			})
-			.then(function (fnErr) {
-				if (!isFailed) {
-					x.throws(function () {}, err, msg);
-				}
-
+			.catch(function (fnErr) {
```
schnittstabil: That will catch errors from `fn` **and** errors thrown at line 75…

SamVerschueren: In what situation does line 75 throw an error? It seems you are correct though, if I change it to your suggestion, it works.

```js
			.then(function () {
				x.throws(function () {}, err, msg);
			}, function (fnErr) {
				x.throws(function () {
					throw fnErr;
				}, err, msg);
			});
```

schnittstabil: `x.throws(function () {}, err, msg);` throws an error, and is evaluated if `fn` isn't rejected.

vdemedes:
Could you update it to:

```js
return fn
  .catch()
  .then()
```

instead of shortcut form?

sindresorhus:
@vdemedes See: https://github.com/sindresorhus/ava/pull/124#discussion_r43855517

vdemedes:
@sindresorhus this applies if we have `.then()` and `.catch()` in that order.

`.then(yes, no)` is a shortcut for `.catch(no).then(yes)`.

schnittstabil:
@vdemedes It is not a shortcut:
```js
Promise.reject('err').catch(() => null).then(() => console.log('success'))
//=> outputs: success

Promise.resolve(() => console.log('success'), () => null)
//=> no output
```

vdemedes:
@schnittstabil https://github.com/petkaantonov/bluebird/wiki/Promise-anti-patterns#the-thensuccess-fail-anti-pattern

vdemedes:
My suggestion is to convert this:

```js
.then(function () {
			x.throws(function () {}, err, msg);
			}, function (fnErr) {
 				x.throws(function () {
 					throw fnErr;
 				}, err, msg);
 			});
```

to this:
```js
.catch(function (fnErr) {
	x.throws(function () {
		throw fnErr;
	}, err, msg);
})
.then(function () {
	x.throws(function () {}, err, msg);
})
```

schnittstabil:
> …  and rationalize to yourself that at least the code is "less coupled" or something.
I don't do that :bowtie:

schnittstabil:
@vdemedes if you really, really, really want to avoid `.then(…, …)` then you may prefer:

```js
	if (fn && fn.then) {
			fn.then(function () {
				try {
					x.throws(function () {}, err, msg);
				} catch (e) {
					reject(e);
				}
			}).catch(function (errFn) {
				try {
					x.throws(function () {
						throw errFn;
					}, err, msg);
					resolve();
				} catch (e) {
					reject(e);
				}
			});
		});
	}
```

SamVerschueren:
@schnittstabil The reason I created this PR was to make the code cleaner, not make it more complex :). So your example is definitely a no-go.

@vdemedes Suggestion doesn't work

#### Example

```js

Promise.reject(new Error('foo'))
    .catch(function (err) {
        console.log(err.message);
    })
    .then(function (success) {
        console.log(success);
    });
// => foo
// => undefined

Promise.resolve('bar')
    .catch(function (err) {
        console.log(err.message);
    })
    .then(function (success) {
        console.log(success);
    });
// => bar

Promise.reject(new Error('foo'))
    .then(function (success) {
        console.log(success);
    }, function (err) {
        console.log(err.message);
    });
// => foo

Promise.resolve('bar')
    .then(function (success) {
        console.log(success);
    }, function(err) {
        console.log(err.message);
    });
// => bar
```

It looks like the only option is the `then(success, err)` one. If you don't want to use that one, just keep it the way it is :).

sindresorhus:
> It looks like the only option is the then(success, err) one. 

I'm fine with using this pattern. It's IMHO only an anti-pattern when you use it thinking it's a shortcut for `.catch()`.

vdemedes:
Ok, let's merge this and we'll see what we can do later.

PR #125, opened 2015-11-04
Add string support to throws

SamVerschueren:
I added the possibility to pass in the error message as string to the `throws` method.

#### Example

```js
test('test it', t => {
	t.throws(throws, 'foo bar');
	t.end();

	function throws() {
		throw new Error('foo bar');
	}
});
```

#### Benefit
If you want to test the returned error message you either should use a regex or a validation function. The validation function spoils the test file in my opinion, and the regex is, well yeah, it's a regex. People have to escape some characters and a string just looks cleaner.

#### Downside
Not sure if it is a downside, but it's not standard `assert` stuff which might confuse people. Although, I don't really see it as downside, it's more an improvement.

// @sindresorhus @vdemedes @kevva @Qix-

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -90,7 +90,13 @@ x.throws = function (fn, err, msg) {
 	}
 
 	try {
-		assert.throws(fn, err, msg);
+		if (typeof err === 'string') {
+			assert.throws(fn, function (e) {
+				return e.message === err;
+			}, msg);
+		} else {
+			assert.throws(fn, err, msg);
+		}
 	} catch (err) {
 		test(false, create(err.actual, err.expected, err.operator, err.message, x.throws));
 	}
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -91,12 +91,13 @@ x.throws = function (fn, err, msg) {
 
 	try {
 		if (typeof err === 'string') {
-			assert.throws(fn, function (e) {
-				return e.message === err;
-			}, msg);
-		} else {
-			assert.throws(fn, err, msg);
+			var e = err;
+			err = function (err) {
+				return err.message === e;
+			};
 		}
+
+		assert.throws(fn, err, msg);
 	} catch (err) {
 		test(false, create(err.actual, err.expected, err.operator, err.message, x.throws));
 	}
--- a/readme.md
+++ b/readme.md
@@ -423,7 +423,7 @@ Assert that `value` is not deep equal to `expected`.
 
 Assert that `function` throws an error or `promise` rejects.
 
-`error` can be a constructor, regex or validation function.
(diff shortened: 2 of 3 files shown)


Qix-:
`should` does this and to be honest it's the only way I could see reliably testing for expected exceptions. Expecting an exception but not specifying which one introduced a few regressions before I realized I needed to specify the message.

sindresorhus:
> Not sure if it is a downside, but it's not standard assert stuff which might confuse people.

It's not. We only use the core assert so not to duplicate code. If anything can be done better, we'll do it. Also the core Node.js people have said `assert` is done and won't change anymore.

-

:+1: from me.

Qix-:
> Also the core Node.js people have said `assert` is done and won't change anymore.

One of my problems with the Node team. /offtopic

SamVerschueren:
Things can always be done better!

sindresorhus:
@Qux- In their defense it was a mistake to expose it as it was only really meant to be used to test core, from what I've read.

Also why ideally almost everything should be in user-land so it can be individually versioned and evolved.

Qix-:
That makes sense.

lib/assert.js
```diff
@@ -90,7 +90,13 @@ x.throws = function (fn, err, msg) {
...
 
 	try {
-		assert.throws(fn, err, msg);
+		if (typeof err === 'string') {
+			assert.throws(fn, function (e) {
+				return e.message === err;
+			}, msg);
+		} else {
+			assert.throws(fn, err, msg);
+		}
```
sindresorhus:
```js
if (typeof err === 'string') {
  err = function (e) {
    return e.message === err;
  };
}

assert.throws(fn, err, msg);
```

SamVerschueren: Definitely cleaner, but then I guess it should be like this.

```js
if (typeof err === 'string') {
  var e = err;
  err = function (err) {
    return err.message === e;
  };
}

assert.throws(fn, err, msg);
```

On your code example, at the time the function is executed, `err` is not the string anymore but it is the function. So it will check if `err.message` is equal to the function instead of to the original string.

sindresorhus: Yup

sindresorhus:
:+1: Thanks Sam :)

SamVerschueren:
No problem, always happy to contribute ;).

Any idea when these latest changes (`string` support in throws, promise fix in throws) will be released? They are needed in `pageres` and `gh-user` :). (and many others off course)

sindresorhus:
@SamVerschueren Hopefully today. Just want to get in the two open PRs. Any help reviewing https://github.com/sindresorhus/ava/pull/120 welcome ;)

SamVerschueren:
That's not an easy one :). Will have a look at it later today.

PR #126, opened 2015-11-04
Add test

SamVerschueren:
According to issue https://github.com/sindresorhus/ava/issues/123, long-running promises doesn't seem to work.

I added a new test to test that behaviour, and guess what, it works.

-----
Just wanted to share this test and it might be merged in. If you don't want to, just reject :).

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -198,6 +198,23 @@ test('handle throws with rejected promise', function (t) {
 	});
 });
 
+test('handle throws with long running rejected promise', function (t) {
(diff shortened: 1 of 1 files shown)


sindresorhus:
More tests are always welcome. Can't get enough. Nom nom.

SamVerschueren:
0adca243bf45b1023d1e8b6e01e69e1f94773df69bbbdd1e4489b8d1fd53225d

Issue #123, opened 2015-11-04
`throws` doesn't work correctly with promises


SamVerschueren:
So, I tried to use the `throws` method in some libraries and found out it doesn't work as it should be or as we expect it should work. Haven't found a solution for this but I have some interesting information (I guess).

#### Immediately rejected promises

This works correctly, if a promise is rejected immediately, the test is done correctly.

```js
test('immediate rejection', async t => {
	await t.throws(throws(), /help me/);

	function throws() {
		return Promise.reject(new Error('hello world'));
	}
});
```

> Test will fail because `hello world` !== 'help me'
> Note 2: Can't pass in the string `help me` as second argument of throws because then the test succeeds. But that is another issue.

#### Long running promises

Think about network requests, etc. This does not work, the `throws` assertion does not wait untill the promise is rejected or resolved.

```js
test('test me', async t => {
	await t.throws(throws(), /help me/);

	function throws() {
		return new Promise(function (resolve, reject) {
			setTimeout(function () {
				reject(new Error('hello world'));
			}, 2000);
		});
	}
});
```

> This test succeeds.

An interesting thing I noticed is that the test succeeds immediately, it doesn't wait 2 seconds untill the promise is rejected.

Any thoughts?

SamVerschueren:
Apparently, this prints `undefined` instead of `Promise`.

```js
test(t => {
    console.log(t.throws(Promise.resolve()));
    t.end();
});
```

SamVerschueren:
Can be closed as #127 was merged.

PR #127, opened 2015-11-04
Throws returns the promise

SamVerschueren:
Finally found the solution for #123 

By returning the promise, it's possible to check the throwed error like this which succesfully waits untill the promise is resolved/rejected.

```js
test(async t => {
	await t.throws(throws(), /hello world/);

	function throws() {
			setTimeout(function () {
				reject(new Error('hello world'));
			}, 2000);
		});
	}
});
```


vdemedes:
lgtm

@sindresorhus?

sindresorhus:
Looks good :)

PR #134, opened 2015-11-05
Kill forked processes after tests are done

vdemedes:
@jamestalmage here's the refactored solution to the issue you mentioned - https://github.com/sindresorhus/ava/pull/134/files#diff-a48173105195b7a6481a3ad85b45144eR27.

The "trick" is to kill the forked process after the test results are received and not let `cli.js` handle this mess.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -133,8 +133,7 @@ function sum(arr, key) {
 	return result;
 }
 
-function exit(testRun) {
-	var results = testRun.results;
+function exit(results) {
 	// assemble stats from all tests
 	var stats = results.map(function (result) {
 		return result.stats;
@@ -159,10 +158,7 @@ function exit(testRun) {
 	// correctly flush the output when multiple test files
 	process.stdout.write('');
 
-	// wait for the child processes to exit
-	testRun.childProcesses.finally(function () {
-		process.exit(failed > 0 ? 1 : 0);
-	});
+	process.exit(failed > 0 ? 1 : 0);
 }
 
 function init(files) {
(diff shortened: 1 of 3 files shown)


jamestalmage:
That works. My thought was to leave long running processes alone until we were all done (that's why I did it the way I did). Looking back, I can't remember why I thought that was important. If prematurely killing the process for `test-file-a.js` adversely affects `test-file-b.js`, you are doing something wrong anyway.

jamestalmage:
Oh wait - this is not going to work.

You need to do something like this:

```js
// fork.js
new Promise(resolve, reject) {
  var storedResult;
  ps.on('result', function (result) {
    // don't resolve here we MUST wait for the process to exit.
    storedResult = result;
    ps.kill();
  });
  ps.on('exit', function(code) {
    if (code > 0) {
      reject();
    }
    else {
      resolve(storedResult);
   }
  });
}
```

vdemedes:
Why do we need to wait for the process to exit? `results` event is being sent after all tests are done.

jamestalmage:
Because then this: https://github.com/sindresorhus/ava/pull/134/files#diff-2cce40143051e25f811b56c79d619bf5R161

Gets executed before all the child processes finish exiting.

If you kill the parent, the child processes are killed without warning, despite where they may be wrapping up the kill signal that was sent to the process.

`nyc` intercepts the kill signal, and delays letting the process actually die until it is done writing coverage results back to disk.

vdemedes:
Ok, I'll push an update, thanks for the tip!

jamestalmage:
@vdemedes 
OK, I was working on a PR as well.
See #135.
It has a good failing test.
I can carry that through with the fix or let you handle it.

PR #136, opened 2015-11-05
Wait for fork to exit and then return results

vdemedes:
Improves #134.

// @jamestalmage

### DIFF
--- a/lib/fork.js
+++ b/lib/fork.js
@@ -19,8 +19,10 @@ module.exports = function (args) {
 	var ps = childProcess.fork(babel, args, options);
 
 	var promise = new Promise(function (resolve, reject) {
+		var testResults;
+
 		ps.on('results', function (results) {
-			resolve(results);
+			testResults = results;
 
 			// after all tests are finished and results received
 			// kill the forked process, so AVA can exit safely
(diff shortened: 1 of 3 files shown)


jamestalmage:
That should do it.
Once you merge this, I will merge master back into #135 and it should pass.

lib/fork.js
```diff
@@ -19,8 +19,16 @@ module.exports = function (args) {
 	var ps = childProcess.fork(babel, args, options);
 
 	var promise = new Promise(function (resolve, reject) {
+		// define backup results, when there are no tests defined
+		var testResults = {
+			stats: {
+				passCount: 0
```
jamestalmage: Does `ava` already warn if a test file has no tests? (i.e. `passCount === 0`)?

I would think that should be a failure, otherwise putting:
```js
process.exit(0); 
```
at the top of your test makes it pass. (contrived I know).

Should we be warning if the process exits with code `0`, but never sends a `result` message?

vdemedes: I think I will revert this one, because it seems unrelated to this PR. I think this question deserves its own space.

jamestalmage: I would agree, it needs a separate issue/discussion.

I think your code is fine, no need to revert. Lets just create a new issue and follow up on it.

jamestalmage: @vdemedes #137 created to track that discussion.

vdemedes: I will revert it, just for the sake of keeping things clear.

PR #146, opened 2015-11-07
bump power-assert dependencies.

jamestalmage:
Good news: 

I have worked with @twada to get yield, and assert expressions [working in power-assert](https://github.com/power-assert-js/power-assert-formatter/blob/master/CHANGELOG.md#130-2015-11-07), and it is [pretty slick](https://github.com/power-assert-js/babel-plugin-espower/issues/6#issuecomment-154598875).

Bad news:

[I broke semver](https://github.com/power-assert-js/babel-plugin-espower/issues/8) (just a little).

The best course of action is to just bump your dependencies so none of them


twada:
I really appreciate @jamestalmage's great contribution.

FYI: I just released [babel-plugin-espower 1.0.1](https://github.com/power-assert-js/babel-plugin-espower/blob/master/CHANGELOG.md#101-2015-11-07) to fix the little breaking. Thanks!

twada:
@jamestalmage You mean `yield, and async/await expressions`? ;)

jamestalmage:
Yeah - that! Also not sure what happened in the middle of my last sentence.

sindresorhus:
Awesome work @jamestalmage! I use `await` inline a lot.

Issue #144, opened 2015-11-06
Missing module error with the babelification


sindresorhus:
@floatdrop after 1d5ef4c5f6fcb70f9e90e584dfd08865db1e93b4 landed I'm now getting this error when testing `got` using master.

```
~/dev/got master
❯ ava

module.js:339
    throw err;
    ^

Error: Cannot find module 'babel-runtime/helpers/async-to-generator'
    at Function.Module._resolveFilename (module.js:337:15)
    at Function.Module._load (module.js:287:25)
    at Module.require (module.js:366:17)
    at require (module.js:385:17)
    at Object.<anonymous> (/Users/sindresorhus/dev/got/test/unix-socket.js:3:25)
    at Module._compile (module.js:435:26)
    at requireFromString (/Users/sindresorhus/dev/ava/node_modules/require-from-string/index.js:14:4)
    at Object.<anonymous> (/Users/sindresorhus/dev/ava/lib/babel.js:28:1)
    at Module._compile (module.js:435:26)
    at Object.Module._extensions..js (module.js:442:10)
module.js:339
    throw err;
    ^

Error: Cannot find module 'babel-runtime/helpers/async-to-generator'
    at Function.Module._resolveFilename (module.js:337:15)
    at Function.Module._load (module.js:287:25)
    at Module.require (module.js:366:17)
    at require (module.js:385:17)
    at Object.<anonymous> (/Users/sindresorhus/dev/got/test/retry.js:3:25)
    at Module._compile (module.js:435:26)
    at requireFromString (/Users/sindresorhus/dev/ava/node_modules/require-from-string/index.js:14:4)
    at Object.<anonymous> (/Users/sindresorhus/dev/ava/lib/babel.js:28:1)
    at Module._compile (module.js:435:26)
    at Object.Module._extensions..js (module.js:442:10)
TypeError: Cannot read property 'stack' of undefined
    at error (/Users/sindresorhus/dev/ava/cli.js:51:19)
From previous event:
    at Object.<anonymous> (/Users/sindresorhus/dev/ava/cli.js:218:34)
    at Module._compile (module.js:435:26)
    at Object.Module._extensions..js (module.js:442:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:311:12)
    at Function.Module.runMain (module.js:467:10)
    at startup (node.js:134:18)
    at node.js:961:3
```

floatdrop:
@sindresorhus babel transpile files with `async` and add

```js
[...]

PR #150, opened 2015-11-07
Remove runtime from lib/babel

floatdrop:
Fixes for #144 

- [x] Remove `runtime` with polyfills
- [x] Add `regenerator` to test files
- [ ] Make it less horrible

### DIFF
--- a/lib/babel.js
+++ b/lib/babel.js
@@ -16,7 +16,7 @@ try {
 
 var options = {
 	blacklist: hasGenerators ? ['regenerator'] : [],
-	optional: hasGenerators ? ['asyncToGenerator', 'runtime'] : ['runtime'],
+	optional: hasGenerators ? ['asyncToGenerator'] : [],
 	plugins: [
 		createEspowerPlugin(babel, {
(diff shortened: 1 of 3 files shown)


sindresorhus:
Couldn't it just read the runtime from https://github.com/facebook/regenerator ?

test/fixture/async-await.js
```diff
@@ -1,5 +1,6 @@
 'use strict';
 const test = require('../../');
+const Promise = require('bluebird');
```
sindresorhus: Why? I don't want to manually require a Promise polyfill in all my tests.

floatdrop: Yes, it is lame to include polyfills in test files - but it is bad to include global polyfills either (because they leak to production code).

floatdrop:
> Couldn't it just read the runtime from https://github.com/facebook/regenerator ?

This is what `Make it less horrible` part for ;)

test/fixture/async-await.js
sindresorhus: I guess there's really no good solution for this...

Issue #152, opened 2015-11-07
Improve babelification


sindresorhus:
Continuation of https://github.com/sindresorhus/ava/issues/144.

From #151:

> Drawback - now you can require any module from ava/node_module in tests.

I don't think this is optimal. We should look into better solutions long-term.

novemberborn:
I believe this was fixed by #390.

PR #153, opened 2015-11-07
WIP: Improve babel runtime lookup

jamestalmage:
Attempts to address #152 

It needs a bit more testing / follow up (I think there are a couple extra `requireFrom`s in there).

Submitted for feedback.

Instead of mucking with module paths, it finds the dependencies and builds some wrapper code that will require them with absolute paths (pretty straightforward)

The difficult part is making sure we get module resolution correct.

Basically, the way I see it there can be a few things happening:

 1. User has `babel-runtime` installed locally
 2. User has `regenerator` installed locally.
 3. User has `babel-core` installed locally
 4. User has `babel` installed locally
 5. User has nothing installed and will rely on the local ava implementation.

Here is what we should then resolve

1.  `babel-runtime/regenerator` from `.`
2.  `regenerator/runtime-module` from `.`
3.  `regenerator/runtime-module` from `node_modeules/babel-core`
4.  deeply nested at this point: `n_m/babel/n_m/babel-core/n_m/regenerator/runtime-module` (`n_m === node_modules`)

And who knows how `npm dedupe` has flattened things (some of those deeply nested items may get promoted to higher level folders by the dedupe process).

### DIFF
--- a/lib/babel.js
+++ b/lib/babel.js
@@ -1,22 +1,30 @@
 'use strict';
-var resolveFrom = require('resolve-from');
+var unsafeResolveFrom = require('resolve-from');
 var createEspowerPlugin = require('babel-plugin-espower/create');
 var requireFromString = require('require-from-string');
+var path = require('path');
 var hasGenerators = parseInt(process.version.slice(1), 10) > 0;
-var path = process.argv[2];
+var testFile = process.argv[2];
 
-var babel;
-
-try {
-	var localBabel = resolveFrom('.', 'babel-core') || resolveFrom('.', 'babel');
-	babel = require(localBabel);
-} catch (err) {
-	babel = require('babel-core');
+function resolveFrom(dir, p) {
+	try {
+		return unsafeResolveFrom(dir, p);
+	} catch (e) {}
 }
+
+var babelPath = resolveFrom('.', 'babel-core') || resolveFrom('.', 'babel') || require.resolve('babel-core');
+var babel = require(babelPath);
+
+var rtp1 = 'babel-runtime/regenerator';
+var rtp2 = 'regenerator/runtime-module';
+var rtp3 = 'babel-core/node_modules/regenerator/runtime-module';
+var babelDir = path.dirname(babelPath);
+var runtimePath = resolveFrom('.', rtp1) || resolveFrom('.', rtp2) || resolveFrom('.', rtp3) || resolveFrom(babelDir, rtp1) || resolveFrom(babelDir, rtp2) || resolveFrom(babelDir, rtp3) || require.resolve(rtp1);
+var bluebirdPath = require.resolve('bluebird');
 
 var options = {
 	blacklist: hasGenerators ? ['regenerator'] : [],
-	optional: hasGenerators ? ['asyncToGenerator', 'runtime'] : ['runtime'],
+	optional: hasGenerators ? ['asyncToGenerator'] : [],
 	plugins: [
 		createEspowerPlugin(babel, {
 			patterns: require('./enhance-assert').PATTERNS
@@ -24,5 +32,16 @@ var options = {
 	]
 };
 
-var transpiled = babel.transformFileSync(path, options);
-requireFromString(transpiled.code, path);
+var transpiled = babel.transformFileSync(testFile, options);
+
+var code = [
+	'var regeneratorRuntime=require("',
+	runtimePath,
+	'"); var Promise=require("',
+	bluebirdPath,
+	'"); (function(){',
+	transpiled.code,
+	'})()'
+].join('');
+
+requireFromString(code, testFile);


lib/babel.js
```diff
@@ -1,22 +1,30 @@
...
+var rtp1 = 'babel-runtime/regenerator';
+var rtp2 = 'regenerator/runtime-module';
+var rtp3 = 'babel-core/node_modules/regenerator/runtime-module';
+var babelDir = path.dirname(babelPath);
+var runtimePath = resolveFrom('.', rtp1) || resolveFrom('.', rtp2) || resolveFrom('.', rtp3) || resolveFrom(babelDir, rtp1) || resolveFrom(babelDir, rtp2) || resolveFrom(babelDir, rtp3) || require.resolve(rtp1);
+var bluebirdPath = require.resolve('bluebird');
+
 var options = {
 	blacklist: hasGenerators ? ['regenerator'] : [],
-	optional: hasGenerators ? ['asyncToGenerator', 'runtime'] : ['runtime'],
```
floatdrop: We need `runtime`, because of polyfils - https://github.com/sindresorhus/ava/issues/144#issuecomment-154691249

jamestalmage: Does `runtime` have an option NOT to put the require statements in at the beginning.
That would be nice.

lib/babel.js
```diff
@@ -1,22 +1,30 @@
...
+		return unsafeResolveFrom(dir, p);
+	} catch (e) {}
 }
 
+var babelPath = resolveFrom('.', 'babel-core') || resolveFrom('.', 'babel') || require.resolve('babel-core');
+var babel = require(babelPath);
+
+var rtp1 = 'babel-runtime/regenerator';
+var rtp2 = 'regenerator/runtime-module';
+var rtp3 = 'babel-core/node_modules/regenerator/runtime-module';
```
sindresorhus: This won't work with npm@3 as it might put `regenerator

lib/babel.js
```diff
@@ -1,22 +1,30 @@
...
+var babelDir = path.dirname(babelPath);
+var runtimePath = resolveFrom('.', rtp1) || resolveFrom('.', rtp2) || resolveFrom('.', rtp3) || resolveFrom(babelDir, rtp1) || resolveFrom(babelDir, rtp2) || resolveFrom(babelDir, rtp3) || require.resolve(rtp1);
+var bluebirdPath = require.resolve('bluebird');
```
sindresorhus: core.js (which is the runtime polyfill) already includes a small promise polyfill. Wouldn't it be better to use that? Bluebird is huge and slow to `eval`.

lib/babel.js
jamestalmage: which is why I requireFrom('.', 'regenerator/runtime-module');

lib/babel.js
jamestalmage: I thought I saw bluebird as a babel dependency and just assumed it was their polyfill. I'm very likely wrong on that.

sindresorhus: Yup, it's only our.

lib/babel.js
sindresorhus: :+1:

lib/babel.js
jamestalmage: We need to use `core.js` anyways with the `runtime` transform.

jamestalmage: [core.js aliasing](https://github.com/babel/babel.github.io/blob/862b43db93e48762671267034a50c30c00e433e2/docs/usage/runtime.md#core-js-aliasing)

jamestalmage:
Here is a snippet of the runtime plugin (I would link to the repo, but it is 404'ing, since they went monorepo - old url was: https://github.com/babel-plugins/babel-plugin-runtime)

```
  return new Plugin("runtime", {
    metadata: {
      group: "builtin-post-modules"
    },

    pre: function pre(file) {
      file.set("helperGenerator", function (name) {
        return file.addImport(RUNTIME_MODULE_NAME + "/helpers/" + name, name, "absoluteDefault");
      });

      file.setDynamic("regeneratorIdentifier", function () {
        return file.addImport(RUNTIME_MODULE_NAME + "/regenerator", "regeneratorRuntime", "absoluteDefault");
      });
    },
```

jamestalmage:
maybe we can wrap that plugin / override the `pre` method (to do nothing), and introduce our own requires that we resolve.

jamestalmage:
What if we just link `./node_modules/ava/node_modules/babel-runtime => ./node_modules/babel-runtime`

Then it resolves for them no matter what.

`npm prune` would blow it away, but we would just put it back each time.

jamestalmage:
Wrapping the Babel-6 plugin with our own require paths would be easy enough:

https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-runtime/src/index.js

just override the `pre` fuction

jamestalmage:
@sindresorhus @floatdrop 
See https://github.com/jamestalmage/hack-babel-5-runtime-plugin

It is possible to wrap the existing babel plugins and manipulate the require paths they put in the source.

floatdrop:
@jamestalmage yes, this is the way to fix this :+1:

lib/babel.js
jamestalmage: @sindresorhus 

> Bluebird is huge and slow to eval

I am not `eval`'ing the entire bluebird codebase just a require statement to that resolved path.

Actually this way is probably faster since we already require `bluebird` in `test.js` (which I had to do to make unhandledRejections work on Node <= 0.12).

Having the babelfied code require a different implementation would mean Node is parsing two different implementations per forked process.

Also, in some ways this is a feature - you can use all the fancy sugar bluebird adds in your test, but we hide it from your production code so you can make sure your production code is not relying on bluebird specific features. 

If babels implementation supports `unhandledRejection` notifications, maybe we should switch `test.js` to that.

sindresorhus:
> It is possible to wrap the existing babel runtime plugins (both v5 and v6) and manipulate the require paths they put in the source.

:+1:

sindresorhus:
I think we should upgrade to Babel 6 first, though, before we spend to much time making it work with Babel 5, which is going to be replaced.

lib/babel.js
sindresorhus: > Also, in some ways this is a feature - you can use all the fancy sugar bluebird adds in your test

Not a big fan of implicitly exposing Bluebird to users. If they need the extra functionality they can require it themselves.

jamestalmage: Yep. If the Babel polyfill doesn't support `unhandledRejection`, at least `pinkie` is a viable option now.

jamestalmage: Note that `pinkie-promise` probably isn't a great alternative for us, since it defaults to `global.Promise` if available (which won't fire `unhandledRejection` on Node < `4.0.0`)

sindresorhus: @jamestalmage Can you open an issue about it over at core.js? Would be more convenient to use that one, otherwise we'll need to figure out a way to manually alias it to pinkie.

jamestalmage: Are we sure it doesn't support it? (I certainly don't know).

sindresorhus: > If the Babel polyfill doesn't support unhandledRejection

I'm just going off your comment :p

jamestalmage: @sindresorhus 
FYI, looks like `core-js` [does support unhandled rejection](https://github.com/zloirock/core-js/blob/14f0bcb4b5bed7b92b9c9f2c1f59d5d76f0a5fde/library/modules/es6.promise.js#L123)

sindresorhus: :+1:

PR #165, opened 2015-11-08
Support Observable

### LINKED ISSUE #84: Support Observable
Re https://github.com/mochajs/mocha/issues/1898

-

https://github.com/zenparsing/es-observable

@thejameskyle What do we need to do to support it?

I would guess calling `.forEach()` on the observable as that returns a promise.

BarryThePenguin:
Fixes #84

Still working through it. Happy to hear any feedback. I've just duplicated the tests for Promises. I started out writing tests using RxJs which all passed successfully. Switched to `zen-observable` half way through so some tests are currently skipped as a result.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -7,6 +7,7 @@ var test = require('tape');
 var Runner = require('../lib/runner');
 var fork = require('../lib/fork');
 var ava = require('../lib/test');
+var Observable = require('./fixture/observable');
 
 function execCli(args, cb) {
 	if (!Array.isArray(args)) {
@@ -90,6 +91,25 @@ test('plan assertions with support for promises', function (t) {
 		}, 200);
 
 		return promise;
+	}).run().then(function (a) {
+		t.is(a.planCount, 2);
+		t.is(a.assertCount, 2);
+		t.end();
+	});
+});
+
+test('plan assertions with support for observables', function (t) {
+	ava(function (a) {
+		a.plan(2);
+
+		var observable = Observable.of();
+
+		setTimeout(function () {
+			a.pass();
+			a.pass();
+		}, 200);
+
+		return observable;
 	}).run().then(function (a) {
 		t.is(a.planCount, 2);
 		t.is(a.assertCount, 2);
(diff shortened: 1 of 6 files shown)


test/test.js
```diff
@@ -96,6 +97,25 @@ test('plan assertions with support for promises', function (t) {
 	});
 });
```
sindresorhus: Might we worth extracting all the Observable related test into a separate test file, since there are so many now.

sindresorhus:
Generally looks good :)

Will need docs and tests are currently failing.

test/test.js
BarryThePenguin: Ok, and the Promise related tests too? I'm thinking the end result would be:
- test.js
- test.Promise.js
- test.Observable.js

BarryThePenguin:
Happy to write some documentation. I'll keep looking at the failing tests, I'm not sure what the problem is yet...

BarryThePenguin:
I think 'es6-symbol' is causing the tests to fail. What is the preferred solution for Symbols in 0.10 and 0.12?

BarryThePenguin:
Hmm... also zen-observable requires a Promise polyfill...

jamestalmage:
We will have to run zen-observables through the Babel `runtime` plugin. As well as anything using Symbol.

test/test.js
sindresorhus: Yeah, but that should be done in a separate PR.

- test.js
- promise.js
- observable.js

jamestalmage:
@BarryThePenguin #200 - will fix the Node 0.10 error - it exists on master (I don't know how the latest master build passed)

BarryThePenguin:
Great, I'll squash and rebase again once that gets in

test/test.js
```diff
@@ -8,6 +8,18 @@ var Runner = require('../lib/runner');
...
+	Object.defineProperty(global, 'Promise', {
+		value: Promise,
+		configurable: true,
+		enumerable: false,
+		writable: true
+	});
+}
+
+var Observable = require('zen-observable');
+
```
BarryThePenguin: I'm not sure if this is a Great Idea™

Any opinions?...

jamestalmage: Nah - just set it. That doesn't do anything to prevent pollution of globals.

BarryThePenguin: It's in test/test.js for Observables only as we want a lightweight implementation of the observable spec to test with. `zen-observable` "[Requires ES6 Promises or a Promise polyfill](https://github.com/zenparsing/zen-observable#zen-observable)" and this looked like a simple way to do that without also having a Symbol polyfil at the same time.

Maybe @blesh or @zenparsing have a suggestion?

blesh: I'm unsure of what the goal is here for "Supporting Observables". What do you need them to do? Do you need to support subscription to them? Do you need to support returning them?

If it's the former, you should be able to test for `obj[Symbol.observable]` and call it (like a function) to get back an `observable` object with a `subscribe()` function on it that follows the `zen-observable` spec. For that, there's really no need to import a library.

On the other hand, if you want to emit them, you'll want to use the zen-observable polyfill or the RxJS 5 implementation as they follow the spec. And you'll need to do something like you're doing here.

blesh:
Okay, I think I'm caught up with what the desire is here.

Observables can "end" in one of three ways: 
They terminate because: 

1. the producer errors.
2. the producer completes successfully.
3. the consumer unsubscribed.


You only really need to care about 1 and 2, because 3 would be caused by the test itself. To handle 1 and 2 for zen-observable, RxJS 5+, Kefir, Most and Bacon, you'd just need:

```js
let obj = getSomethingThatMightBeObservable();

if (obj[Symbol.observable]) {
  obj[Symbol.observable]().subscribe({
    complete() { /* done successfully */ },
    error(err) { reportError(err); /* done because error */ }
  });
}
```

But the problem there is while you've determined that it's "done", you haven't analyzed the output.

If the idea is to analyze the output, the best short-term solution is to subscribe to `RxJS 5` or `zen-observable` Observables with `forEach`, which returns a promise:

```js
return myObservable.forEach(x => doSomeAssertion(x === 'whatever'));
```

There would be no mileage from Kefir, Most or Bacon from that unless the spec changes to include that `forEach` must be implemented on returns from `[Symbol.observable]()`. cc/ @zenparsing

blesh:
> There would be no mileage from Kefir, Most or Bacon from that unless the spec changes to include that forEach must be implemented on returns from [Symbol.observable]()

... actually in that case it wouldn't matter anyhow, because the libraries that aren't RxJS 5 or zen-observable would need to be subscribed to in a different way.

I think that using `forEach` is probably your safest bet. It returns a promise, so you wouldn't have to make a change.

sindresorhus:
> I think that using forEach is probably your safest bet. It returns a promise, so you wouldn't have to make a change.

:+1:

sindresorhus:
OT to this PR, but kinda weird that you have to call `Observable.forEach(function () {});` to convert it to a promise. I would have preferred something like `Promise.resolve(observable)`.

blesh:
> I would have preferred something like Promise.resolve(observable).

Good luck with that one. It's not a bad idea, but I think it will be hard to get traction on changes to Promise to support Observable.

However, with RxJS, you can just use `Observable.prototype.toPromise()` which will convert the Observable to a Promise that resolves with the last value of the Observable when the Observable completes, or reject with the Observable's error if it errors.

blesh:
Also, I think it would be `Promise.from(observable)` ;)

sindresorhus:
@BarryThePenguin Can you rebase from master and fix the merge conflict? This looks good to me.

@vdemedes @jamestalmage Anything final you'd like to point out?

vdemedes:
Everything looks good to me, very straightforward, thanks @BarryThePenguin!

BarryThePenguin:
:+1:

BarryThePenguin:
Feel free to cancel [build 168](https://ci.appveyor.com/project/sindresorhus/ava/build/168)

That was a mistake

sindresorhus:
> Feel free to cancel build 168

Done.

sindresorhus:
Landed! Thank you @BarryThePenguin :)

BarryThePenguin:
:tada:

sindresorhus:
I've added some docs for it https://github.com/sindresorhus/ava/commit/e57eff059f73e53f1c8db1af096dc53e2608a322, but I don't really know how Observables work, so any improvements welcome! For example, can you `await` an Observable in an async function?

// @thejameskyle

zenparsing:
@sindresorhus You can't directly `await` an observable.  We've discussed it at length and implicit conversion to promise for use by `await` is controversial.  You can consume the sequence with `forEach` and `await` the returned promise, or we might add something like `observable.toPromise()` in the future.

sindresorhus:
> We've discussed it at length and implicit conversion to promise for use by await is controversial.

That's too bad... I was hoping `await` could be used directly by any async primitive. Do you happen to have a link to that discussion?

In that case, `observable.toPromise()` is definitely needed. The `forEach`-to-convert-to-promise thing is super weird.

Issue #140, opened 2015-11-06
Planned assertion documentation issue (or planned assertion bug)


ArtemGovorov:
Readme says:

> And this will result in an error because the code called more assertions than planned:

``` JavaScript
test(t => {
    t.plan(1);

    t.pass();

    setTimeout(() => {
        t.pass();
    }, 100);
});
```

however running the test gives:
```
✔ [anonymous]

  1 test passed

```

vdemedes:
This is happening, because when first `t.pass()` is executed, AVA exits the test and treats it as finished. We had a TODO test for this case before, but removed it because we could not find a good & stable solution for this one.

rmg:
I think the docs could use a little more clarifying. A common problem with test plans is forgetting to update them when new assertions are added. In cases where the assertions happen asynchronously (the primary use case for test plans), they are not seen even if they fail. The suite will continue to pass, even though failing tests have been added. 

While this is less of a problem if one is disciplined about verifying their failing tests before fixing them, it would be nice if the tools helped. And if the tools can't help, then at least the readme should offer a warning about this scenario.

I suspect that if this was "fixed" in the code there would be a number of bug reports claiming AVA suddenly broke but really it just stopped hiding broken tests. This is a problem we have had to address in node-tap and it "broke" a lot of test suites when we did.

sindresorhus:
@rmg As far as I know there's no way to actually fix this (would be so happy to be proven wrong), as we would never know whether an async planned assertion will happen at some point. Currently, we end the test in the next event loop when all planned assertions are fulfilled.

I removed the test in https://github.com/sindresorhus/ava/commit/7a3f4444997099cd571118508f08783bce8a5526 when I realized it's not really possible. I had this idea of checking `process._getActiveHandles().length === 1`, but I don't think that'll work when tests run concurrently and there might be a lot of other async things going on in AVA itself.

How do you think we could further improve the docs about this? Would you be interested in doing a documentation pull request?

rmg:
I'm planning on submitting a docs PR if I can remember to when I'm on a real computer. Meant to mention that in my comment above.

jamestalmage:
If we are killing the process because we think we are done that's one thing, but the following should detect an issue:

```js
test('a', t => {
	t.plan(1);
	t.ok(true);

	setTimeout(() => t.ok(true), 1000);  // test-b will still be running when this happens.
});

test('b', t => {
	setTimeout(() => t.end(), 2000 )
});
```

[...]

PR #169, opened 2015-11-08
Fix Windows onExit failure.

jamestalmage:
SIGHUP seems to be the only kill signal supported by Windows:
https://nodejs.org/api/process.html#process_process_kill_pid_signal

This is a stab in the dark while I wade through setting up Node on Windows.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -158,7 +158,10 @@ function exit(results) {
 	// correctly flush the output when multiple test files
 	process.stdout.write('');
 
-	process.exit(failed > 0 ? 1 : 0);
+	// timeout required to correctly flush stderr on Node 0.10 Windows
+	setTimeout(function () {
+		process.exit(failed > 0 ? 1 : 0);
+	}, 0);
 }
 
 function init(files) {
--- a/lib/babel.js
+++ b/lib/babel.js
@@ -28,3 +28,9 @@ var transpiled = babel.transformFileSync(testPath, options);
 requireFromString(transpiled.code, testPath, {
 	appendPaths: module.paths
 });
+
+process.on('message', function (message) {
+	if (message['ava-kill-command']) {
+		process.exit(0);
+	}
+});
--- a/test/fork.js
+++ b/test/fork.js
@@ -48,8 +48,8 @@ test('exit after tests are finished', function (t) {
 
 	fork(fixture('long-running.js'))
 		.on('exit', function () {
-			t.ok(Date.now() - start < 10000, 'did NOT wait for setTimeout(fn, 15000');
-			t.ok(cleanupCompleted, 'did wait for onExit(fn) to complete');
+			t.ok(Date.now() - start < 10000, 'should NOT wait for setTimeout(fn, 15000)');
+			t.ok(cleanupCompleted, 'should wait for onExit(fn) to complete');
 		})
 		.on('cleanup-completed', function (event) {
 			cleanupCompleted = event.completed;
--- a/test/fork.js
+++ b/test/fork.js
@@ -48,8 +48,8 @@ test('exit after tests are finished', function (t) {
 
 	fork(fixture('long-running.js'))
 		.on('exit', function () {
-			t.ok(Date.now() - start < 10000, 'should NOT wait for setTimeout(fn, 15000)');
-			t.ok(cleanupCompleted, 'should wait for onExit(fn) to complete');
+			t.ok(Date.now() - start < 10000, 'test waited for a pending setTimeout');
+			t.ok(cleanupCompleted, 'cleanup did not complete');
 		})
 		.on('cleanup-completed', function (event) {
 			cleanupCompleted = event.completed;
--- a/test/test.js
+++ b/test/test.js
@@ -12,7 +12,7 @@ function execCli(args, cb) {
 		args = [args];
 	}
 
-	childProcess.execFile('../cli.js', args, {cwd: __dirname}, cb);
+	childProcess.execFile(process.execPath, ['../cli.js'].concat(args), {cwd: __dirname}, cb);
 }
 
 test('run test', function (t) {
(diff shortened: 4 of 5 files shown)


jamestalmage:
:cry: If only it were that easy.

sindresorhus:
You're dealing with computers. It's never easy :p

jamestalmage:
trying again.
I'm closer. The remaining issue might be a problem with my box.

jamestalmage:
arghh - so close!
Working on windows except in Node 0.10

test/test.js
```diff
@@ -6,13 +6,16 @@ var figures = require('figures');
...
+var cli = path.join(__dirname, '../cli.js');
 
 function execCli(args, cb) {
 	if (!Array.isArray(args)) {
 		args = [args];
 	}
+	args.unshift(cli);
 
-	childProcess.execFile('../cli.js', args, {cwd: __dirname}, cb);
+	childProcess.execFile(node, args, {cwd: __dirname}, cb);
```
sindresorhus: You can just use `childProcess.execFile(process.execPath, args, {cwd: __dirname}, cb);` instead of `which`.

jamestalmage:
Finally!
@sindresorhus - I'm gonna let the latest push go green and then squash

cli.js
```diff
@@ -158,7 +158,10 @@ function exit(results) {
 	// correctly flush the output when multiple test files
 	process.stdout.write('');
```
jamestalmage: I might have actually fixed this line too.

Do we have a test that covers this?

Everything passed on OSX / Node `4.1` without this and without my change (in other words, unless the problem this addresses is specific to an OS / Node version we do NOT have test coverage for it).

jamestalmage: @vdemedes, this line is from #47 

Can you remember what you were addressing there? Tests pass without it. (even before my changes).

test/fork.js
```diff
@@ -48,8 +48,8 @@ test('exit after tests are finished', function (t) {
 
 	fork(fixture('long-running.js'))
 		.on('exit', function () {
-			t.ok(Date.now() - start < 10000, 'did NOT wait for setTimeout(fn, 15000');
-			t.ok(cleanupCompleted, 'did wait for onExit(fn) to complete');
+			t.ok(Date.now() - start < 10000, 'should NOT wait for setTimeout(fn, 15000)');
+			t.ok(cleanupCompleted, 'should wait for onExit(fn) to complete');
```
jamestalmage: I was supposed to clean these up in #135 but it got merged before I could.

lib/babel.js
```diff
@@ -28,3 +28,9 @@ var transpiled = babel.transformFileSync(testPath, options);
 requireFromString(transpiled.code, testPath, {
 	appendPaths: module.paths
 });
+
+process.on('message', function (message) {
+	if (message['ava-kill-command']) {
+		process.exit(0);
```
jamestalmage: This means the test process is now receiving messages.

Do we need to monkey patch `process.on` so tests never see them?

cli.js
vdemedes: `process.stdout.write('');` was there before me, @sindresorhus should have a reason for it.

test/fork.js
vdemedes: These descriptions still look like test descriptions. Those messages are error messages, not assert descriptions. imho smth like this would be better: `t.ok(cleanupCompleted, 'clean up did not complete')`.

Would be great, if you improved those ;)

jamestalmage: done!

cli.js
jamestalmage: oops. Sorry - `blame` attributed it to you, and I didn't scan down the #47 far enough.

jamestalmage: @sindresorhus ?

test/fork.js
```diff
@@ -48,8 +48,8 @@ test('exit after tests are finished', function (t) {
 
 	fork(fixture('long-running.js'))
 		.on('exit', function () {
-			t.ok(Date.now() - start < 10000, 'did NOT wait for setTimeout(fn, 15000');
-			t.ok(cleanupCompleted, 'did wait for onExit(fn) to complete');
+			t.ok(Date.now() - start < 10000, 'did not short-circuit pending setTimeout');
```
vdemedes: how about `test waited for a pending setTimeout` or `test should've exited without waiting for a timer`?

vdemedes: Nah, last one is crap, first one is simple & good. What do you think?

jamestalmage: went with the first one

jamestalmage:
@sindresorhus @vdemedes 
I've made the recommended changes. Anything else holding this up? It would be nice if peoples PR's could go green again.

cli.js
sindresorhus: Yeah, I added it as a quick hack because the `1 test passed` message wouldn't output without it when running multiple tests and I couldn't reproduce it in a test. If it works fine now you can just remove it.

lib/babel.js
sindresorhus: Not unless it interferes with them. @vdemedes ?

sindresorhus:
Landed! Superb work tracking this down @jamestalmage. Node.js 0.10 is getting so annoying to support...

lib/babel.js
jamestalmage: It won't interfere with AVA test code.

But if someone attaches a process.on('message', handler), and then emits their own message events on process, that would interfere. Though it would be really odd behavior.

jamestalmage: It wouldn't really ever make sense for them to do that though (not unless they fork an additional process).

vdemedes:
I just realized something. Why don't we just let forks exit themselves, without interaction with a parent process?

```js
process.send(results);
process.exit();
```

jamestalmage:
I don't think there is anything stopping us from doing that now.

I did have one thought with regard to #140, and that is to leave processes running as long as possible to give us the best chance of catching extra assertion calls (beyond plan), and repetitive calls to `t.end()`

In that case it would be up to the parent to close.

But if we think that is a bad idea, then no - there is probably no reason to do it the way we are.

PR #174, opened 2015-11-09
made "concurernvy vs. parallelism" section clearer

alexbooker:
For many people, the concept of concurrency vs. parallelism can be hard to grok.

I do not think the old descriptin read very well, nor did I think it would be _that_ helpful to someone who needed the answer in the frirst place (i.e. someone who doesn't know the difference.)

I tried writing an answer of my own before I realized that someone else had already put it best, so I quoted an aclaimed StackOverflow answer and gave appropiate credit (as Creative Commons requires).

I think linking to a SO question will also make it easy for people to read other answers, which they may understand better.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -520,7 +520,13 @@ AVA, not Ava or ava. Pronounced [`/ˈeɪvə/` ay-və](media/pronunciation.m4a?ra
 
 ### Concurrency vs. parallelism
 
-Concurrency is not parallelism. It enables parallelism. It's about dealing with, while parallelism is about doing, lots of things at once.
+Concurrency is not parallelism. It enables parallelism.
+
+**Concurrency** is when two tasks can start, run, and complete in overlapping time periods. It doesn't necessarily mean they'll ever both be running at the same instant. (e.g. multitasking on a single-core machine.) 
+
+**Parallelism** is when tasks literally run at the same time. (e.g. on a multicore processor.)
+
+[(Source)](http://stackoverflow.com/a/1050257)
 
 
 ## Support
--- a/readme.md
+++ b/readme.md
@@ -520,7 +520,7 @@ AVA, not Ava or ava. Pronounced [`/ˈeɪvə/` ay-və](media/pronunciation.m4a?ra
 
 ### Concurrency vs. parallelism
 
-Concurrency is not parallelism. It enables parallelism. To learn more about the difference, see [this](http://stackoverflow.com/a/1050257) Stack Overflow answer. 
+Concurrency is not parallelism. It enables parallelism. To learn more about the difference, see [this](http://stackoverflow.com/q/1050222) Stack Overflow answer. 
 
 
 ## Support
--- a/readme.md
+++ b/readme.md
@@ -522,7 +522,7 @@ AVA, not Ava or ava. Pronounced [`/ˈeɪvə/` ay-və](media/pronunciation.m4a?ra
 
 Concurrency is not parallelism. It enables parallelism.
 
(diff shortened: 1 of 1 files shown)


schnittstabil:
Nice, I like clarifications, but I don't think he *put it best*. Mainly because the examples explain more than the descriptions, but introduces (probably unknown) terms: *multitasking*, *single-core*, *multicore*, *processor*.

readme.md
```diff
@@ -520,7 +520,13 @@ AVA, not Ava or ava. Pronounced [`/ˈeɪvə/` ay-və](media/pronunciation.m4a?ra
 
 ### Concurrency vs. parallelism
 
-Concurrency is not parallelism. It enables parallelism. It's about dealing with, while parallelism is about doing, lots of things at once.
+Concurrency is not parallelism. It enables parallelism.
+
+**Concurrency** is when two tasks can start, run, and complete in overlapping time periods. It doesn't necessarily mean they'll ever both be running at the same instant. (e.g. multitasking on a single-core machine.)
```
schnittstabil: 1. *two or more tasks*?
2. I wouldn't say *starting* and *completing* is necessary
3. I'm not a native speaker, but I wouldn't use the word `when` in that context to express a condition.

alexbooker:
Thanks for the input, @schnittstabil - I apprecaite it :smile:.

I _completely_ agree that it should be _"two ore more tasks"_ - I edited the original [answer](https://stackoverflow.com/posts/1050257/revisions) and this pull request in response.

Bare in mind the answer has **400:heavy_plus_sign:**   positive responses which tells me the answer _is_ useful.

I think it reads fine, to be honest. What's more, because I have included a link to the SO thread someone can go there for additional context and alternative answers or analogies.

readme.md
```diff
@@ -520,7 +520,13 @@ AVA, not Ava or ava. Pronounced [`/ˈeɪvə/` ay-və](media/pronunciation.m4a?ra
 
 ### Concurrency vs. parallelism
 
-Concurrency is not parallelism. It enables parallelism. It's about dealing with, while parallelism is about doing, lots of things at once.
+Concurrency is not parallelism. It enables parallelism.
+
+**Concurrency** is when two or more tasks can start, run, and complete in overlapping time periods. It doesn't necessarily mean they'll ever both be running at the same instant. (e.g. multitasking on a single-core machine.)
```
schnittstabil: *two or more tasks can start, run, and complete in overlapping time periods*, that is also true for **Parallelism**, isn't it?

alexbooker: @schnittstabil As is stated: _"concurrency enables parallism"_. Of course there is an overlap.

What do you suggest?

vdemedes:
I think this information should not belong in the AVA's Readme. Project's readme is about project, not about explaining concepts. What I would do here, is to link to that stack overflow answer, instead of copying it.

alexbooker:
@vdemedes You know what, mate. That is a good idea. Done.

schnittstabil:
I thought about that description and has come to the conclusion that it is at least misleading.
*Sequential* and *Concurrent* is about algorithms and language design.<br/>*Serial* and *Parallel* is about hardware and run-time.

Instead of covering these aspects they are messed up.

Also the answer mentions *Sun's* **Multithreaded** *Programming Guide*, which (obviously) only talking about Threads and states that: Concurrency is a more generalized form of parallelism. That is only valuable in that guide, but can't be transferred to newer concepts used in Node or Go…

alexbooker:
@schnittstabil I meant to link to the question.

Issue #156, opened 2015-11-07
Allow babel transpilation of test helper files.


jamestalmage:
We are doing all this work to ensure babel does not leak into any file but the test file, but there is a downside for "helper" files, etc.

The other option would be to hook require and make them provide some sort of `include` / `exclude` glob.

sindresorhus:
Why did you close?

jamestalmage:
Didn't mean to.

sindresorhus:
I'm open to it, as it would be very nice to be able to use ES2015 stuff in test helper files too. I think a `whitelist` would be better so we could still make use of the ES2015 `import` semantics.

SamVerschueren:
Would be very nice indeed! Maybe use something like `.es6.js` as extension? This way you don't need to update the whitelist.

jamestalmage:
Or just default the whitelist to include `test/_*`, since the underscore prefix is already set aside to denote helpers. 

Or just everything in the `test` directory? Event `test/fixtures/**`?

sindresorhus:
@jamestalmage I think defaulting the whitelist to `test/_*` is a good idea, not so sure about the whole test directory, though. I would definitely not want my fixtures implicitly transpiled.

jamestalmage:
Thats my feeling as well.

just-boris:
+1, I want this thing too. But why do you need all that stuff around `require-from-string`?

Why not just add babel hook using `require("babel-core/register");`?

jamestalmage:
Because, one goal of this library is not to pollute any of your production code with babel transforms/polyfills (unless you specifically ask it to).

just-boris:
But all my source code is written for babel too. Will there an option to transpile files from my `src/` folder in addition to `test/`?

sparty02:
> Because, one goal of this library is not to pollute any of your production code with babel transforms/polyfills (unless you specifically ask it to).

I actually want my source code (non-test) to be transpiled by babel as well..... i.e. I want to be able to ```import``` a module that is written in ES6 into my ava test module.  Is there a suggested technique for this today?  Or is this a pending enhancement?

sindresorhus:
@sparty02 Pending enhancement. Definitely something we want to support, but haven't had the chance yet. See https://github.com/sindresorhus/ava/issues/111. Any thoughts there much appreciated.

sparty02:
@sindresorhus  ahh.... thanks for the reference to #111.  I'll take a look!

[...]

Issue #154, opened 2015-11-07
No longer working on Node.js 0.10


sindresorhus:
Not sure why the tests didn't catch this.

```js
// test.js
import test from 'ava';

test(t => {
	t.pass();
	t.end();
});
```

```
❯ node -v
v0.10.40

❯ ava test.js

TypeError: Cannot read property 'stack' of undefined
    at error (/usr/local/lib/node_modules/ava/cli.js:51:19)
From previous event:
    at Object.<anonymous> (/usr/local/lib/node_modules/ava/cli.js:218:34)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Function.Module.runMain (module.js:497:10)
    at startup (node.js:119:16)
    at node.js:935:3
```

// @floatdrop

floatdrop:
@sindresorhus most likely, because [this line](https://github.com/sindresorhus/ava/blob/master/lib/fork.js#L36) rejects to `undefined` and `cli` trys to get `stack` from `undefined`.

sindresorhus:
Yup, but I'm unsure why. That test should pass, but it's rejecting somehow.

floatdrop:
@sindresorhus probably because we [kill](https://github.com/sindresorhus/ava/blob/master/lib/fork.js#L29) process after all tests are done - and sometimes Node.js 0.10 will return `SIGTERM` exit code (143) which is trigger `error` handler.

jamestalmage:
Should we be sending a different kill signal?

https://nodejs.org/api/process.html#process_signal_events

> * SIGTERM and SIGINT have default handlers on non-Windows platforms that resets the terminal mode before exiting with code 128 + signal number. If one of these signals has a listener installed, its default behaviour will be removed (node will no longer exit).
> * SIGPIPE is ignored by default, it can have a listener installed.
> * SIGHUP is generated on Windows when the console window is closed, and on other platforms under various similar conditions, see signal(7). It can have a listener installed, however node will be unconditionally terminated by Windows about 10 seconds later. On non-Windows platforms, the default behaviour of SIGHUP is to terminate node, but once a listener has been installed its default behaviour will be removed.
> * SIGTERM is not supported on Windows, it can be listened on.
[...]

PR #177, opened 2015-11-10
wip - Support compilation of "helper" modules

jamestalmage:
This solves the basic requirements of #156.

It relies on https://github.com/floatdrop/require-from-string/pull/3

Along the way I uncovered a few other bugs that should get fixed in master (I will call those out in line comments). 

// @floatdrop

### DIFF
--- a/lib/babel.js
+++ b/lib/babel.js
@@ -1,10 +1,13 @@
 'use strict';
+var path = require('path');
 var resolveFrom = require('resolve-from');
 var createEspowerPlugin = require('babel-plugin-espower/create');
 var requireFromString = require('require-from-string');
 
 var hasGenerators = parseInt(process.version.slice(1), 10) > 0;
 var testPath = process.argv[2];
+var testDir = path.dirname(testPath);
+var Module = module.constructor;
 var babel;
 
 try {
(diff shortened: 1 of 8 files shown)


lib/babel.js
```diff
@@ -14,17 +17,58 @@ try {
...
+				patterns: require('./enhance-assert').PATTERNS
+			})
+		]
+	};
+}
+
+// TODO: This needs to be smarter.
+function shouldTranspile(filePath) {
+	return testDir === path.dirname(filePath) && filePath[0] === '_' || filePath === testPath;
+}
```
jamestalmage: This `shouldTranspile` definitely needs to be improved. Very naive.

I'm holding off investing time on it until we have a decision what the API and CLI should look like.

lib/fork.js
```diff
@@ -33,7 +33,7 @@ module.exports = function (args) {
 
 		ps.on('exit', function (code) {
 			if (code > 0 && code !== 143) {
-				reject();
+				reject(new Error(file + ' exited with error code ' + code));
```
jamestalmage: This was the cause of the unhelpful message reported in #154 (not the cause of that particular failure - just the bad message).

test/fork.js
```diff
@@ -35,7 +35,7 @@ test('rejects on error and streams output', function (t) {
 			buffer += data;
 		})
 		.catch(function () {
-			t.ok(/no such file or directory/.test(buffer));
+			t.ok(/Cannot find module/.test(buffer));
```
jamestalmage: The require hook now outputs a different message when it can't find a file.

lib/babel.js
```diff
@@ -14,17 +17,59 @@ try {
...
+		plugins: [
+			createEspowerPlugin(babel, {
+				patterns: require('./enhance-assert').PATTERNS
+			})
+		]
+	};
+}
+
+// TODO: This needs to be smarter.
+function shouldTranspile(filePath) {
```
jamestalmage: This shouldTranspile definitely needs to be improved. Very naive.

I'm holding off investing time on it until we have a decision what the API and CLI should look like.

PR #181, opened 2015-11-10
Support use of t.end as callback.

### LINKED ISSUE #180: Support use of t.end as callback
Passing `t.end` as a callback is a nice feature of `tape`. It'll end the test & automatically check first arg for an error.

If one wants to use AVA's `t.end` as a callback though, firstly, it doesn't check for error arg, but secondly it will break due to references to `this`.

This would provide built-in support for node's errback pattern, echoing AVA's built-in support for promises.

Current implementation:
```js
Test.prototype.end = function () {
	if (this.endCalled) {
		throw new Error('.end() called more than once');
	}

	this.endCalled = true;
	this.exit();
};
```
https://github.com/sindresorhus/ava/blob/7575ca82855da86e359a4ff8b17bb4588d626e8d/lib/test.js#L137-L144

All AVA Test methods face this issue, perhaps they should all be forcibly bound to the current instance, though `t.end` is probably by far the most useful one to have bound.

This may be is too much magic though, mainly raising as I noticed this didn't work when migrating my tape tests.

timoxley:
* Binds all Test.prototype methods to current test.
* Checks for truthy error as first argument to Test.prototype.end.

Closes #180.

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -35,6 +35,10 @@ function Test(title, fn) {
 	if (this.title === 'callee$0$0') {
 		this.title = '[anonymous]';
 	}
+
+	Object.keys(Test.prototype).forEach(function (key) {
+		this[key] = this[key].bind(this);
+	}, this);
 }
 
 module.exports = Test;
@@ -134,7 +138,11 @@ Test.prototype.run = function () {
 	}.bind(this));
 };
 
-Test.prototype.end = function () {
+Test.prototype.end = function (err) {
+	if (arguments.length >= 1 && err) {
+		this.ifError(err);
+	}
+
 	if (this.endCalled) {
 		throw new Error('.end() called more than once');
 	}
--- a/readme.md
+++ b/readme.md
@@ -319,6 +319,19 @@ test(async t => {
 
 *You don't have to manually call `t.end()`.*
 
+### Callback support
+
+AVA supports using `t.end` as the final callback when using node-style
+error-first callback APIs. AVA will consider any truthy value passed as
+the first argument to `t.end` to be an error.
+
+```js
+test(t => {
+	// t.end automatically checks for error as first argument
+	fs.readFile('data.txt', t.end);
+});
+```
+
 
 ## API
 
--- a/test/test.js
+++ b/test/test.js
@@ -119,6 +119,24 @@ test('handle non-assertion errors', function (t) {
 	});
 });
 
+test('end can be used as callback without maintaining thisArg', function (t) {
+	ava(function (a) {
+		setTimeout(a.end);
+	}).run().then(function (a) {
+		t.false(a.assertError);
+		t.end();
+	});
+});
+
+test('end can be used as callback with error', function (t) {
+	ava(function (a) {
+		a.end(new Error('failed'));
+	}).run().catch(function (err) {
+		t.true(err instanceof Error);
+		t.end();
+	});
+});
+
 test('handle non-assertion errors even when planned', function (t) {
 	ava(function (a) {
 		a.plan(1);


lib/test.js
```diff
@@ -134,7 +138,11 @@ Test.prototype.run = function () {
 	}.bind(this));
 };
 
-Test.prototype.end = function () {
+Test.prototype.end = function (err) {
+	if (arguments.length >= 1 && Boolean(err)) {
```
sindresorhus: `if (arguments.length >= 1 && err) {`?

sindresorhus:
Looks good to me. @vdemedes @jamestalmage ?

lib/test.js
jamestalmage: why not just `if (err)` ?

lib/test.js
```diff
@@ -134,7 +138,11 @@ Test.prototype.run = function () {
...
+		this.ifError(err);
```
jamestalmage: If they've used `t.plan()`, won't this increase their assertion count causing a `plan error` message?

lib/test.js
timoxley: oh yeah wtf am I doing

timoxley: Originally this was `!!err` but linter told me to `Boolean(err)` so I followed blindly

vdemedes:
Except suggestions in comments, everything is good.

lib/test.js
timoxley: What happens if you have a plan and call `t.end`?

lib/test.js
```diff
@@ -134,7 +138,11 @@ Test.prototype.run = function () {
 	}.bind(this));
 };
 
-Test.prototype.end = function () {
+Test.prototype.end = function (err) {
+	if (arguments.length >= 1 && err) {
```
jamestalmage: I'm still genuinely curious why you are checking argument length.
If `arguments.length === 0`, isn't `err` going to be falsy no matter what?

timoxley:
Also note we're opting for checking & printing error *before* checking if `.end` has been called more than once. This is how tape does it and I think it makes sense as the error message + stack is going to be more helpful messaging than simply 'called end more than once'. They'll get this message too, but only after they fix the actual source of the error.

lib/test.js
timoxley: Updated.

jamestalmage:
> What happens if you have a plan and call t.end?

I don't know, but we probably should have a test for it.

jamestalmage:
@timoxley looking at the code again I'm pretty sure you are going to get a bad error message if you use plan and end with this.

Please write a test that does this:

```js
test(t => {
  t.plan(1);
  t.ok(1);
  t.end(new Error('you should get this error message, not a plan count error'));
});
```

And check the message of the thrown exception.

lib/test.js
timoxley: Idea is to only log the assertion if there's some arguments passed. Alternative is have every test log an additional ifError assertion at the end.

jamestalmage: No, I'm saying

```js
if (err) {
  this.ifErr(err)
}
```

timoxley: You're right.

timoxley: Perhaps it should always log the assertion if there's `arguments.length`?

jamestalmage:
@timoxley I think it's down to that one `if` statement and another test ensuring your use `ifError` does not generate a `plan count error` message.

lib/test.js
jamestalmage: Hmmm.

We need it to behave consistently whether or not it's being used as a callback

```js
t.plan(1);
t.ok(true);
t.end();
```

should perform the same as

```js
t.plan(1)
t.ok(true);
setImmediate(t.end);
```

timoxley: in neither of those situations will the t.end be called with any arguments though.

timoxley: The use-case I'm thinking about goes something like this:

```js
test(t => {
	fs.readFile('data.txt', t.end); // will error if t.end is passed truthy first arg
});
```

jamestalmage: that is fine, but for your plan test do something that is not async:
```js
test(t => {
    t.plan(1);
    t.ok(true);
    t.end(new Error('something bad happened'));
});
```

vdemedes: To avoid conflicts with `t.plan()`, `this.ifError(err)` could be replaced with:

```js
this.assertError = new assert.AssertionError({
  actual: err,
  message: 'Callback called with an error → ' + err,
  operator: 'callback'
});

this.exit();
```

This is the same stuff we do when promise rejects.

vdemedes:
Hey @timoxley, could you please fix this suggestion (https://github.com/sindresorhus/ava/pull/181#discussion-diff-44397790), so that we can merge it in? Thanks!

vdemedes:
Closed via https://github.com/sindresorhus/ava/commit/403a28c8cebf9a8ff736aaaf2c4cb79ca30c5a47

Thanks @timoxley!

PR #187, opened 2015-11-11
Fix circular references on assertions before sending across the wire.

jamestalmage:
We serialize and send copies of assertions errors across the wire on failure.
If `assertionError.actual`, or `assertionError.expected` have a circular reference, this causes a problem when `process.send` converts to JSON.

This replaces `serialize-error` with `destroy-circular`. I found one [tiny bug](https://github.com/jonpacker/destroy-circular/pull/1) with their implementation, but it seems good otherwise.


jamestalmage:
:angry:

Windows and Node 0.10 are quickly becoming annoying.

jamestalmage:
It looks like AppVeyor is having some issues. Those pass just fine in my Windows VM. It's not making it through npm install on AppVeyor

jamestalmage:
What the heck? It was green after rebase! All I did was squash!

jamestalmage:
Crap. Adding the `tap-spec` reporter should not have made this pass.

We have intermittent test failures on CI that I can not reproduce locally at all.

jamestalmage:
@sindresorhus 
I think this is ready for merge.

I can drop the `tap-spec` commits, but I wanted you to see the brittle test issue I'm having before I do. (and if it fails again when I revert to @2f16758, I may toss my computer out the window).

sindresorhus:
:+1:

PR #189, opened 2015-11-11
Cache transformed files

floatdrop:
I have some tests hanging with cache, but implementation will look something like this. Opening to gather suggestions.

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,12 +3,25 @@ var setImmediate = require('set-immediate-shim');
 var relative = require('path').relative;
 var hasFlag = require('has-flag');
 var chalk = require('chalk');
+var join = require('path').join;
 var serializeError = require('./lib/serialize-value');
 var Runner = require('./lib/runner');
 var send = require('./lib/send');
 var log = require('./lib/logger');
 
 var runner = new Runner();
+var xdgBasedir = require('xdg-basedir');
+var cache = require('cacha')(join(xdgBasedir.cache, 'ava'));
+var Configstore = require('configstore');
+var config = new Configstore('ava', {
+	lastCacheClean: Date.now()
+});
+var WEEK = 7 * 24 * 60 * 60 * 1000;
+
+if (Date.now() - config.get('lastCacheClean') > WEEK) {
+	config.set('lastCacheClean', Date.now());
+	cache.clean();
+}
 
 // note that test files have require('ava')
 require('./lib/babel').avaRequired = true;
--- a/lib/babel.js
+++ b/lib/babel.js
@@ -6,6 +6,11 @@ var resolveFrom = require('resolve-from');
 var hasGenerator = require('has-generator');
 var serializeValue = require('./serialize-value');
 var send = require('./send');
+var hasha = require('hasha');
+var hashObj = require('hash-obj');
+var xdgBasedir = require('xdg-basedir');
+var path = require('path');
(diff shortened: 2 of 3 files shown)


lib/babel.js
```diff
@@ -2,6 +2,8 @@
 var resolveFrom = require('resolve-from');
 var createEspowerPlugin = require('babel-plugin-espower/create');
 var requireFromString = require('require-from-string');
+var hasha = require('hasha');
+var cache = require('cacha')('.ava/cache');
```
sindresorhus: Can you make sure it follow XDG?

https://github.com/sindresorhus/xdg-basedir#cache

index.js
```diff
@@ -4,6 +4,9 @@ var hasFlag = require('has-flag');
 var serializeError = require('serialize-error');
 var Runner = require('./lib/runner');
 var runner = new Runner();
+var cache = require('cacha')('.ava/cache');
+
+cache.clean();
```
sindresorhus: Why?

floatdrop: `cache.clean()` removes outdated files from cache directory (so it will not grow endlessly). Function name a little misleading, true.

sindresorhus: Makes sense.

Why can't it just do that on every call for expired `ttl`s?

floatdrop: > Why can't it just do that on every call for expired ttls?

Calling `clean` on every `get` call will be expensive (for example, if cache is big enough, like 1000 files) - we need to read cache directory and check every file `atime`.

sindresorhus: Does it do that every time `cache.clean()` is called? If yes, it might be worth only doing it once in awhile and checking with a file the last time it was done.

floatdrop: @sindresorhus yes, this makes sense. I'm up to something like [`update-notifier` Configstore](https://github.com/yeoman/update-notifier/blob/master/index.js#L39).

sindresorhus: :+1: Once a week or less sounds sensible. It doesn't generate that much junk.

sindresorhus:
And I dig the module name ;)

lib/babel.js
jamestalmage: We are going to need a way to include the settings for the transform in the hash as well.

Perhaps it is:
```
  MD5(MD5(source_file_contents) + MD5(
    ava_version + '\n' +
    settings_file_contents + '\n' +
    ...???...
  ));
```
How do we handle changing dependency versions? What happens when the `babel-plugin-transform-runtime` has a patch level bump (meaning the output of the transform may have changed some for the same input). Are we going to have to scan all the plugin version numbers and make those part of the hash? I can't think of an other way.

However we do it, we should compute the part of the hash that is consistent across the test run once (the `ava_version` + `plugin_hash` part), and pass that to the forked processes so they only need to compute the hash of the main file.

floatdrop: @sindresorhus done.

@jamestalmage you are perfectly right. Until #202 is resolved we can hash with `ava` version.

jamestalmage:
https://github.com/jamestalmage/babel-plugin-detective

It is based on Browserify's algorithm for finding dependencies.

Once we make the move to Babel 6, we can use this to precompile dependencies before ever launching the test. We will need other strategies for dynamic requires (possibly the server / client option I discussed earlier), but initially we can demand static requires if they want them transpiled (should cover 90% of use cases anyways).

sindresorhus:
Closing in favor of https://github.com/sindresorhus/ava/pull/352.

PR #190, opened 2015-11-11
Display only failed hooks

### LINKED ISSUE #105: after/before/afterEach/beforeEach hooks should not have a title
It doesn't make sense for them to have a title as you're just hooks.

Should we just prevent it being set in the first place?

vdemedes:
This PR fixes #105.

From now on, AVA displays only hooks that failed. This results to a dramatically cleaner output:

**Before**:

screen shot 2015-11-11 at 5 04 02 pm

**After**:

screen shot 2015-11-11 at 5 04 30 pm

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -53,14 +53,25 @@ Runner.prototype.addSerialTest = function (title, cb) {
 };
 
 Runner.prototype.addBeforeHook = function (title, cb) {
-	this.tests.before.push(new Test(title, cb));
+	var test = new Test(title, cb);
+	test.type = 'hook';
+
+	this.tests.before.push(test);
 };
 
 Runner.prototype.addAfterHook = function (title, cb) {
-	this.tests.after.push(new Test(title, cb));
+	var test = new Test(title, cb);
+	test.type = 'hook';
+
+	this.tests.after.push(test);
 };
 
 Runner.prototype.addBeforeEachHook = function (title, cb) {
+	if (!cb) {
+		cb = title;
+		title = undefined;
+	}
+
 	this.tests.beforeEach.push({
 		title: title,
 		fn: cb
@@ -68,6 +79,11 @@ Runner.prototype.addBeforeEachHook = function (title, cb) {
 };
 
 Runner.prototype.addAfterEachHook = function (title, cb) {
+	if (!cb) {
+		cb = title;
+		title = undefined;
+	}
+
 	this.tests.afterEach.push({
 		title: title,
 		fn: cb
@@ -76,11 +92,19 @@ Runner.prototype.addAfterEachHook = function (title, cb) {
 
 Runner.prototype._runTestWithHooks = function (test) {
 	var beforeHooks = this.tests.beforeEach.map(function (hook) {
-		return new Test(hook.title, hook.fn);
+		var title = hook.title || 'beforeEach for "' + test.title + '"';
+		hook = new Test(title, hook.fn);
+		hook.type = 'eachHook';
+
+		return hook;
 	});
 
 	var afterHooks = this.tests.afterEach.map(function (hook) {
-		return new Test(hook.title, hook.fn);
+		var title = hook.title || 'afterEach for "' + test.title + '"';
+		hook = new Test(title, hook.fn);
+		hook.type = 'eachHook';
+
+		return hook;
 	});
 
 	var tests = [];
@@ -137,10 +161,11 @@ Runner.prototype._addTestResult = function (test) {
 	this.results.push({
 		duration: test.duration,
 		title: test.title,
-		error: test.assertError
+		error: test.assertError,
+		type: test.type
 	});
 
-	this.emit('test', test.assertError, test.title, test.duration);
+	this.emit('test', test.assertError, test.title, test.duration, test.type);
 };
 
 Runner.prototype.run = function () {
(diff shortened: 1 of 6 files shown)


lib/runner.js
```diff
@@ -137,10 +161,11 @@ Runner.prototype._addTestResult = function (test) {
 	this.results.push({
 		duration: test.duration,
 		title: test.title,
-		error: test.assertError
+		error: test.assertError,
+		type: test.type
 	});
 
-	this.emit('test', test.assertError, test.title, test.duration);
+	this.emit('test', test.assertError, test.title, test.duration, test.type);
```
sindresorhus: Not for this PR, but what do you think about pushing and object here instead of all the arguments? We could just reuse the object in `this.results.push`.

test/test.js
```diff
@@ -918,6 +953,32 @@ test('don\'t display test title, if there is only one anonymous test', function
 	});
 });
 
+test('don\'t display hook title, if it did not fail ', function (t) {
```
sindresorhus: no comma needed here and extraneous space at the end.

test/test.js
```diff
@@ -918,6 +953,32 @@ test('don\'t display test title, if there is only one anonymous test', function
...
+		.on('test', function (test) {
+			t.deepEqual(test.err, {});
+			t.is(test.title, 'pass');
+		})
+		.then(function () {
+			t.end();
+		});
+});
+
+test('display hook title, if it failed', function (t) {
```
sindresorhus: no comma needed

test/test.js
```diff
@@ -918,6 +953,32 @@ test('don\'t display test title, if there is only one anonymous test', function
...
+	t.plan(2);
+
+	fork(path.join(__dirname, 'fixture', 'hooks-passing.js'))
+		.on('test', function (test) {
+			t.deepEqual(test.err, {});
+			t.is(test.title, 'pass');
+		})
+		.then(function () {
+			t.end();
+		});
```
sindresorhus: `.then(t.end);`?

test/test.js
```diff
@@ -918,6 +953,32 @@ test('don\'t display test title, if there is only one anonymous test', function
...
+	t.plan(2);
+
+	fork(path.join(__dirname, 'fixture', 'hooks-failing.js'))
+		.on('test', function (test) {
+			t.is(test.err.name, 'AssertionError');
+			t.is(test.title, 'beforeEach for "pass"');
+		})
+		.then(function () {
+			t.end();
+		});
```
sindresorhus: `.then(t.end);`?

test/test.js
```diff
@@ -824,6 +825,40 @@ test('hooks - shared context of any type', function (t) {
...
+	runner.on('test', function (err, title, duration, type) {
+		var test = tests.shift();
+
+		t.is(test.title, title);
+		t.is(test.type, type);
+	});
+
+	runner.run().then(function () {
+		t.end();
+	});
```
sindresorhus: `.then(t.end);`?

sindresorhus:
Looks good! Commented some nitpicks inline.

sindresorhus:
// @floatdrop @jamestalmage

lib/runner.js
```diff
@@ -53,14 +53,25 @@ Runner.prototype.addSerialTest = function (title, cb) {
 };
 
 Runner.prototype.addBeforeHook = function (title, cb) {
-	this.tests.before.push(new Test(title, cb));
+	var test = new Test(title, cb);
+	test.type = 'hook';
```
floatdrop: How about moving types to something like `new Test.hook(title, cb)` (can be done in separate PR)?

sindresorhus: @floatdrop What would `new Test.hook(title, cb)` be? Just a wrapper function that sets the internal `this.type`?

floatdrop: @sindresorhus yep. May be this is not worth a shot yet, but in future we could want to eliminate these string constants.

lib/runner.js
vdemedes: Absolutely, was going to do that here at first, but then thought about a new PR :D

test/test.js
vdemedes: #181 is not merged yet, so it will call `t.end()` with a global context (`t.end()` is not binded yet).

lib/runner.js
vdemedes: imho `new Test.hook(title, cb)` does not look good. In such case, `Test.hook` is a class (function), which has to be uppercase.

test/test.js
sindresorhus: But this is tape, not AVA.

vdemedes: Oh damn it.

lib/runner.js
sindresorhus: Yeah, we can look into this later. It could be a subclass too, `new Hook()`, but I'm not a big fan of that either. Maybe using Symbols as constants. `ava.HOOK_TYPE = Symbol('hook')` or something. But meh.

test/test.js
vdemedes: Just tried it, does not work, because `fork` resolves with test results: `t.end(results)`.

sindresorhus: Oh, nvm then. Sorry about the noise.

vdemedes: No probs ;) `runner.then(t.end)` works though!

vdemedes:
@sindresorhus PR updated ;)

sindresorhus:
Ace :)

vdemedes:
giphy

Issue #192, opened 2015-11-11
Add test.skip()


vdemedes:
This issue's source is #9.

The aim is to add ability to skip certain tests.

Issue #132, opened 2015-11-05
Add test.only()


unlight:
I could not find `test.only` in api documentation. Similar to [tape's test.only](https://github.com/substack/tape#testonlyname-cb) or mocha's `it.only`.
Is it exists?

vdemedes:
`test.only` does not exist at the moment. We're having a long discussion on test modifiers at the moment - #9.

lijunle:
+1 on this feature.

I am considering to migrate from mocha to AVA. However, keep easy test failure debug experience is one of my concerns - `test.only` is the key to do so.

sindresorhus:
:+1: We plan to add this. Help welcome as we have a lot of other things to deal with right now.

vdemedes:
Closing via #204.

PR #193, opened 2015-11-11
Event data improvements

vdemedes:
This PR lays groundwork for #132 and #192 by allowing to easily send & receive custom test properties.

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -158,14 +158,15 @@ Runner.prototype._addTestResult = function (test) {
 		this.stats.failCount++;
 	}
 
-	this.results.push({
+	var props = {
 		duration: test.duration,
 		title: test.title,
 		error: test.assertError,
 		type: test.type
-	});
+	};
 
-	this.emit('test', test.assertError, test.title, test.duration, test.type);
+	this.results.push(props);
+	this.emit('test', props);
 };
 
 Runner.prototype.run = function () {
--- a/test/test.js
+++ b/test/test.js
@@ -405,10 +405,10 @@ test('runner have test event', function (t) {
 		a.end();
 	});
 
-	runner.on('test', function (err, title, duration) {
-		t.error(err);
-		t.equal(title, 'foo');
-		t.ok(duration !== undefined);
+	runner.on('test', function (props) {
+		t.error(props.error);
+		t.equal(props.title, 'foo');
+		t.notEqual(props.duration, undefined);
 		t.end();
 	});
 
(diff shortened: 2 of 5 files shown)


lib/runner.js
```diff
@@ -158,14 +158,15 @@ Runner.prototype._addTestResult = function (test) {
 		this.stats.failCount++;
 	}
 
-	this.results.push({
+	var props = {
 		duration: test.duration,
 		title: test.title,
-		error: test.assertError,
+		err: test.assertError,
```
sindresorhus: I would prefer `error` for consistency.

vdemedes: Haha, I changed it to `err` for consistency too :D

lib/runner.js
```diff
@@ -158,14 +158,15 @@ Runner.prototype._addTestResult = function (test) {
...
 		type: test.type
-	});
+	};
 
-	this.emit('test', test.assertError, test.title, test.duration, test.type);
+	this.results.push(props);
+	this.emit('test', test.assertError, props);
```
sindresorhus: Why is `test.assertError` left in here? It's already part of `props`.

vdemedes: To conform to errback style. However, since it's the event, not a callback, I can skip it.

test/test.js
```diff
@@ -405,10 +405,10 @@ test('runner have test event', function (t) {
 		a.end();
 	});
 
-	runner.on('test', function (err, title, duration) {
+	runner.on('test', function (err, props) {
 		t.error(err);
-		t.equal(title, 'foo');
-		t.ok(duration !== undefined);
+		t.equal(props.title, 'foo');
+		t.ok(props.duration !== undefined);
```
sindresorhus: This should use `t.not()`.

test/test.js
```diff
@@ -405,10 +405,10 @@ test('runner have test event', function (t) {
 		a.end();
 	});
 
-	runner.on('test', function (err, title, duration) {
+	runner.on('test', function (err, props) {
 		t.error(err);
-		t.equal(title, 'foo');
-		t.ok(duration !== undefined);
+		t.equal(props.title, 'foo');
```
sindresorhus: Just fyi, tape supports `t.is` as an alias, so we could use that. Probably better as a separate PR though.

vdemedes: Yep, as part of that whole cleanup process.

lib/runner.js
sindresorhus: I feel it's different. This is more like passing around data than variables. Think of it as a table of data and the keys are the headers. At least how it's like in my mind.

sindresorhus:
Clean code is happy code :D

vdemedes:
Couldn't agree more!

PR #194, opened 2015-11-11
Add test.skip()

### LINKED ISSUE #192: Add test.skip()

vdemedes:
This PR fixes #192.

It allows you to use `test.skip()` to skip the whole test. The output for skipped tests is like that:

screen shot 2015-11-11 at 9 16 46 pm

### DIFF
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -39,6 +39,11 @@ x.test = function (props) {
 		return;
 	}
 
+	if (props.skip) {
+		log.write('  ' + chalk.cyan('- ' + props.title));
+		return;
+	}
+
 	// if (runner.stats.testCount === 1) {
 	// 	return;
 	// }
--- a/index.js
+++ b/index.js
@@ -77,3 +77,4 @@ module.exports.before = runner.addBeforeHook.bind(runner);
 module.exports.after = runner.addAfterHook.bind(runner);
 module.exports.beforeEach = runner.addBeforeEachHook.bind(runner);
 module.exports.afterEach = runner.addAfterEachHook.bind(runner);
+module.exports.skip = runner.addSkippedTest.bind(runner);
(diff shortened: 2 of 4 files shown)


lib/logger.js
```diff
@@ -39,6 +39,11 @@ x.test = function (props) {
 		return;
 	}
 
+	if (props.skip) {
+		log.write('  ' + chalk.cyan('- ' + props.title));
+		return;
+	}
+
 	// if (runner.stats.testCount === 1) {
```
sindresorhus: Unrelated, but we can probably remove this now, right? Because of https://github.com/sindresorhus/ava/commit/bb1304c5be056689c39211d89c673f2f732e4c4a

lib/logger.js
```diff
@@ -39,6 +39,11 @@ x.test = function (props) {
 		return;
 	}
 
+	if (props.skip) {
+		log.write('  ' + chalk.cyan('- ' + props.title));
```
sindresorhus: Is cyan really the best color for skipped? Just asking. It should at least be `.dim` too.

lib/logger.js
vdemedes: Yeah, spotted this crap, will remove later in cleanup PRs ;)

lib/logger.js
vdemedes: I am afraid `.dim` will make it less noticeable, and the purpose of `.skip()` is to temporarily disable it. So if we keep it bright, it will stay in developer's focus, that something needs fixing.

sindresorhus: Ok, let's try it out like that for now.

sindresorhus:
You're on :fire:!

vdemedes:
I was, I was! Now my day is complete :D

PR #200, opened 2015-11-12
fork.js: Reject child-process completion promise with an error.

jamestalmage:
It is generally recommended to reject Promises with an error.
This becomes especially true with the move to `async/await`,
as `Promise.reject(1)` essentially becomes `throw 1` in that context.

Not doing so was causing a weird bug on Node 0.10.

Ref: https://travis-ci.org/sindresorhus/ava/jobs/90709377#L779

### DIFF
--- a/lib/fork.js
+++ b/lib/fork.js
@@ -33,7 +33,7 @@ module.exports = function (args) {
 
 		ps.on('exit', function (code) {
 			if (code > 0 && code !== 143) {
-				reject();
+				reject(new Error(file + ' exited with a non-zero exit code: ' + code));
 			} else {
 				resolve(testResults);
 			}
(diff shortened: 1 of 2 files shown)


sindresorhus:
Agreed. Rejecting without an error makes it impossible to see where it happened.

sindresorhus:
Fails on Node 0.10 :/

https://ci.appveyor.com/project/sindresorhus/ava/build/96/job/3yun8im3pcpp3o8w

jamestalmage:
Passes on Windows 7 / Node 10 here.
I think it is just AppVeyor again.

Give me a moment. to get those logged in the tracking issue.

sindresorhus:
Alright

Issue #161, opened 2015-11-08
100% coverage


sindresorhus:
We're currently at 96% coverage. Which is amazing. Getting us closer to 100% would be a good beginner task. 

**Please leave this for first time contributors.**

See https://coveralls.io/github/avajs/ava?branch=master for uncovered code.

alejandronanez:
@sindresorhus your link is outdated. This should be the right one: https://coveralls.io/github/avajs/ava?branch=master

EEtancelin:
Fine, i dev beginner and i will be happy to take à look and see how improve code coverage.

platy11:
@sindresorhus I was looking at [cli.js100](https://github.com/avajs/ava/blob/master/lib/cli.js#L100) and in my testing (in your `cat-names` repo) I cannot get the deprecation error to appear - is this just me?

novemberborn:
@platy11 that line is in `master` but not yet in a release.

platy11:
@novemberborn Ok, thanks! :smiley:

wildaces215:
Where can I start writing tests for ava it seems like a good place to start contributing to open source software!

novemberborn:
@wildaces215 you can browse our current code coverage at https://codecov.io/gh/avajs/ava. Look for a file that has little coverage, try and figure out what it does, and see if you can write a test that extends the coverage.

We'd be interested in covering larger pieces of logic rather than small edge cases.

PR #203, opened 2015-11-12
More tests for runner.js (issue #161)

sohamkamani:
Added unit tests for methods of Runner. 
Increased branch coverage from 90% to 100% :)

### DIFF
--- a/test/runner.js
+++ b/test/runner.js
@@ -21,19 +21,46 @@ test('runner.addTest adds a new test', function (t) {
 });
 
 test('runner.addSerialTest adds a new serial test', function (t) {
-var runner = new Runner();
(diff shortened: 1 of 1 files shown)


sindresorhus:
According to Coveralls there's no change in the coverage:

>  coverage/coveralls — Coverage remained the same at 95.866%

sohamkamani:
@sindresorhus It's for branch coverage, which is metioned in the report you get after running ```npm test``` (% Branch). 
Also, if you, for example remove [this](https://github.com/sindresorhus/ava/blob/master/lib/runner.js#L76), the other tests still pass.

sindresorhus:
:+1: Alright. Thanks :)

PR #204, opened 2015-11-12
Implement the only modifier.

### LINKED ISSUE #132: Add test.only()

lijunle:
**DO NOT MERGE**

Resolve #132 

I hack the test count number to show the *right* number and passed cases number. From my understand, such information should be pass through the promise chain.

Besides, `ava.serial.only` has no support in this PR yet.

/cc @vdemedes @sindresorhus

---

Screenshot:

screenshot_2015-11-12_23-48-19

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -97,6 +97,9 @@ Runner.prototype.addSkippedTest = function (title, cb) {
 	this.tests.concurrent.push(test);
 };
 
+Runner.prototype.addOnlyTest = function () {
+};
+
 Runner.prototype._runTestWithHooks = function (test) {
 	if (test.skip) {
 		this._addTestResult(test);
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -32,6 +32,7 @@ function Runner(opts) {
 	this.tests = {
 		concurrent: [],
 		serial: [],
+		only: [],
 		before: [],
 		after: [],
 		beforeEach: [],
@@ -97,7 +98,9 @@ Runner.prototype.addSkippedTest = function (title, cb) {
 	this.tests.concurrent.push(test);
 };
 
-Runner.prototype.addOnlyTest = function () {
+Runner.prototype.addOnlyTest = function (title, cb) {
+	this.stats.testCount++;
+	this.tests.only.push(new Test(title, cb));
 };
 
 Runner.prototype._runTestWithHooks = function (test) {
@@ -206,16 +209,20 @@ Runner.prototype.run = function () {
 			}
 		})
 		.then(function () {
-			return self.serial(tests.serial);
+			return self.concurrent(tests.only);
 		})
 		.then(function () {
-			return self.concurrent(tests.concurrent);
+			return tests.only.length ? [] : self.serial(tests.serial);
+		})
+		.then(function () {
+			return tests.only.length ? [] : self.concurrent(tests.concurrent);
 		})
 		.then(function () {
 			return eachSeries(tests.after, self._runTest.bind(self));
 		})
 		.catch(noop)
 		.then(function () {
+			stats.testCount = tests.only.length ? tests.only.length : stats.testCount;
 			stats.passCount = stats.testCount - stats.failCount;
 		});
 };
--- a/index.js
+++ b/index.js
@@ -79,3 +79,4 @@ module.exports.after = runner.addAfterHook.bind(runner);
 module.exports.beforeEach = runner.addBeforeEachHook.bind(runner);
 module.exports.afterEach = runner.addAfterEachHook.bind(runner);
 module.exports.skip = runner.addSkippedTest.bind(runner);
+module.exports.only = runner.addOnlyTest.bind(runner);
--- a/test/test.js
+++ b/test/test.js
@@ -1034,6 +1034,30 @@ test('skip test', function (t) {
 	});
 });
 
+test('only test', function (t) {
+	t.plan(3);
+
+	var runner = new Runner();
+	var arr = [];
+
+	runner.addTest(function (a) {
+		arr.push('a');
+		a.end();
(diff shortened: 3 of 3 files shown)


lib/runner.js
```diff
@@ -203,16 +209,21 @@ Runner.prototype.run = function () {
...
-			return self.concurrent(tests.concurrent);
+			return tests.only.length ? [] : self.concurrent(tests.concurrent);
 		})
 		.then(function () {
 			return eachSeries(tests.after, self._runTest.bind(self));
 		})
 		.catch(noop)
 		.then(function () {
+			// HACK the test count number if only modifier is shown
+			stats.testCount = tests.only.length ? tests.only.length : stats.testCount;
```
lijunle: The variable `testCount` is maintain by the `addTest`, `addOnlyTest`, `addSerialTest` function, which I think that is not suitable. A better way is to return such information (`testCount`, `failCount`) from the promise, then calculate the final stats from at the end of the chain.

However, it involves more code change if doing so. Any ideas?

sindresorhus:
Awesome @lijunle :) Both @vdemedes and I are traveling, so we probably won't be able to review this until monday.

// @jamestalmage

lijunle:
@sindresorhus That is OK, please keep in mind that, there is a hack. I am not going to merge until resolve the hack with a good way (unless a direction).

lib/runner.js
```diff
@@ -203,16 +209,21 @@ Runner.prototype.run = function () {
 			}
 		})
 		.then(function () {
-			return self.serial(tests.serial);
+			return tests.only.length ? self.concurrent(tests.only) : [];
```
vdemedes: it can be just `return self.concurrent(tests.only)`, since `tests.only` defaults to `[]`.

lib/runner.js
vdemedes: I think it's perfectly fine, I wouldn't call it a hack.

vdemedes:
Everything looks good so far, I pointed out some things in the diff. One question I have in mind, should `only` tests be executed serially on in parallel? Btw, mocha allows only one `only` test, so maybe we should also do that? And it would solve the serial vs parallel question too.

lijunle:
> One question I have in mind, should only tests be executed serially on in parallel?

My question is *no*. If the guy want to do so, he should do `ava.serial.only`. However such API is not implemented in this RP.

> Btw, mocha allows only one only test, so maybe we should also do that?

Mocha allow several `only` modifiers. For example, if there are 10 test case, and 3 of them are marked as `only`, those 3 test cases will be run.

vdemedes:
> Mocha allow several only modifiers. For example, if there are 10 test cases, and 3 of them are marked as only, those 3 test cases will be run.

I guess I was running some out-dated version then.

vdemedes:
@lijunle Could you please rebase your PR (to pick up latest fixes for windows tests) and address comments in the diff? Happy to merge this ASAP.

lijunle:
OK, do you want to modify the comment about the *hack*?

vdemedes:
Yeah, the comment is not really needed and this - https://github.com/sindresorhus/ava/pull/204/files#r45035490

lijunle:
Rebased, let us wait for the CI. :smile_cat: 

---

However, although I removed the comment, I still consider that is a *hack*.

The stats should be collected from `_runTestWithHooks` function and pass back to the promise chain in the `Runner.prototype.run` function. At the end of the promise chain, calculate the final stats from the input.

Pseudo code like this:

```js
Runner.prototype.run = function () {
	var self = this;
	return eachSeries(tests.before, this._runTest.bind(this)) // resolve with { pass: 1, failed: 2, skipped: 3 }
		.then(function (stats) {
			return Object.assign(stats, self.serial(tests.serial)); // i.e. { pass: 2, failed: 3, skipped: 4 }
		})
		.then(function (stats) {
			return Object.assign(stats, self.concurrent(tests.concurrent)); // object with { pass, failed, skipped }
		})
		.then(function (stats) {
			return Object.assign(stats, eachSeries(tests.after, self._runTest.bind(self))); // object with { pass, failed, skipped }
		})
		.then(function (stats) {
			self.stats.passCount = stats.pass; // calculate from the input of promise chain.
			return stats; // external check could depends on return value of run function, more stable. 
		});
};
```

vdemedes:
The refactoring and cleanup work is ahead, so don't worry about this 1 line ;)

lijunle:
I understand. It is up to you to design the pattern, that is just my suggestion.

jamestalmage:
@lijunle 

I was actually taking a stab at this earlier when I got pulled away to some other stuff.
Since you are pretty far down the road, I am holding off.

I just squashed my work and pushed it up: https://github.com/jamestalmage/ava/commit/4568347c6ae230e8f03bb20f93f5430ce7bfff58

Use it if it helps.

My other thoughts.

1. I would **really** like to avoid the Mocha behavior of only allowing a single test in exclusive mode. I want to be able to put `only` on however many tests I chose and have them be the ones that run. This is especially important since we don't have grouping yet.

2. I like your idea of `test.serial.only`, etc.

3. I think creating stats at the beginning is making all this hard. Just put flags in each individual test and use `this.tests.concurrent.filter` to create the stats and get at the tests you want (see the diff I linked).

Good work! And thanks for the help.

vdemedes:
@lijunle Thank you!

sindresorhus:
@jamestalmage Any improvements welcome in a follow-up pull request ;)

jamestalmage:
Yeah, sorry! 
I've got a lot going on today and am lagging way behind the conversation.

PR #205, opened 2015-11-13
regression: logger throws on files with both passing and failing tests.

jamestalmage:
log.errors() cycles through all the results from a file (passing and failing)
and prints out a summary of the failing tests and the assertion errors.

It was misidentfying all results as errors, and then throwing
when result.error turned out to be an empty object ({}).

I think the result object can be improved such that
result.error === false when there were no errors.
But that is for another PR.


sindresorhus:
:+1: Good catch :)

PR #206, opened 2015-11-13
Mega PR: unhandledRejection, uncaughtException, reliable IO capture, consistent tests.

jamestalmage:
Combination of other PR's: #211, #213, #215

It adds the following features:
 - Reporting of uncaught exceptions in forked processes.
 - Reporting of unhandled promise rejections in forked processes.
 - Increased likelihood of receiving and printing all data from forked process stdout and stderr.
 - Improved reliability of stdout/stderr capture in AppVeyor CI by detecting AppVeyor and slowing down process.exit events.
 - Since the above solution is not perfect, increase likelihood of passing tests on AppVeyor by running the tests a second time if they fail.

I have some ideas for refactoring that might make some of this cleaner, but I know there are a number of pending PR's, and I don't want to go crazy until those land.


cli.js
```diff
@@ -145,21 +160,30 @@ function exit(results) {
...
+	flushIoAndExit(
+		failed > 0 || unhandledRejectionCount > 0 || uncaughtExceptionCount > 0 ? 1 : 0
+	);
+}
+
+function flushIoAndExit(code) {
 	// TODO: figure out why this needs to be here to
 	// correctly flush the output when multiple test files
 	process.stdout.write('');
+	process.stderr.write('');
```
sindresorhus: Wut, another one? :p

cli.js
```diff
@@ -145,21 +160,30 @@ function exit(results) {
...
 
-	// timeout required to correctly flush stderr on Node 0.10 Windows
+	// timeout required to correctly flush io on Node 0.10 Windows
 	setTimeout(function () {
-		process.exit(failed > 0 ? 1 : 0);
-	}, 0);
+		process.exit(code);
+	}, process.env.APPVEYOR ? 500 : 0);
```
sindresorhus: Agh, so much hacks to work around Windows and AppVeyor flakiness...

sindresorhus:
Why combine the unhandledRejection/uncaughtException stuff with the rest? Seems it could be a separate PR? Or is there a coupling with the other fixes?

lib/fork.js
```diff
@@ -67,11 +79,23 @@ module.exports = function (args) {
...
-		ps.on.apply(ps, arguments);
+	var stdout = new Promise(function (resolve) {
+		ps.stdout.on('end', resolve);
+	});
+
+	var stderr = new Promise(function (resolve) {
+		ps.stderr.on('end', resolve);
+	});
 
+	var endPromise = Promise.all([promise.reflect(), stdout, stderr]).then(function () {
```
sindresorhus: This is weird. Why not just wait on `ps.on('exit')`?

lib/logger.js
```diff
@@ -70,10 +70,41 @@ x.errors = function (results) {
...
 	if (failed > 0) {
 		log.writelpad(chalk.red(failed, plur('test', failed), 'failed'));
 	} else {
 		log.writelpad(chalk.green(passed, plur('test', passed), 'passed'));
 	}
+	if (unhandled > 0) {
+		log.writelpad(chalk.red(unhandled, 'unhandled', plur('rejection', unhandled)));
+	}
+	if (uncaught > 0) {
+		log.writelpad(chalk.red(uncaught, 'uncaught', plur('exceptions', uncaught)));
```
sindresorhus: `exceptions` → `exception`

test/fixture/loud-rejection.js
```diff
@@ -0,0 +1,9 @@
+const test = require('../../');
+
+test('creates an unhandled rejection', t => {
+	Promise.reject(new Error(`You can't handle this!`));
+
+	setTimeout(function () {
+		t.end();
+	}, 0);
```
sindresorhus: `0` can be dropped as it's the same as none

test/test.js
```diff
@@ -1058,14 +1058,32 @@ test('change process.cwd() to a test\'s directory', function (t) {
...
+		t.ok(/1 unhandled rejection[^s]/.test(stderr));
+		t.end();
+	});
+});
+
+test('uncaught exception will be reported to console', function (t) {
+	execCli('fixture/uncaught-exception.js', function (err, stdout, stderr) {
+		t.ok(err);
+		t.ok(/Can't catch me!/.test(stderr));
+		// TODO: The promise ends up rejected, so we need to track this differently
```
sindresorhus: Differently how?

lib/serialize-value.js
```diff
@@ -0,0 +1,15 @@
...
+var destroyCircular = require('destroy-circular');
+
+// Make a value ready for JSON.stringify() / process.send()
+
+module.exports = function serializeValue(value) {
+	if (typeof value === 'object') {
+		return destroyCircular(value);
+	}
+	if (typeof value === 'function') {
+		// JSON.stringify discards functions
```
sindresorhus: Why do *we* need it for?

sindresorhus:
// @vdemedes @floatdrop

lib/fork.js
jamestalmage: I've come at this at least 5 times, and every time, it has required all these fixes.

I don't know that I've ever applied this one last, so it could be unnecessary. I can pull it out with another commit and see what happens.

cli.js
jamestalmage: Yes, so much.

test/test.js
jamestalmage: Not reject the promise, and instead fulfill it with the error information.
Eventually, I'd like to see `cli.js -> error()` callback only get called in the case of a bug (something we can't handle). In a testing framework, getting an error inside the test is an expected response, and shouldn't mean we bail.

lib/serialize-value.js
jamestalmage: If a user decides the do `function foo(){}; throw foo`. Terrible idea, but somebody is bound to. Only happens if they actually throw a function (nested functions will be discarded as normal).

jamestalmage: needs a test.

jamestalmage:
> Why combine the unhandledRejection/uncaughtException stuff with the rest? Seems it could be a separate PR?

The uncaught exception stuff is definitely coupled. These all started as separate PR's (see description).

jamestalmage:
My preference would be to get this merged so everybody can start getting green PR's again. I can then work backwards and see what can be pulled out. Working backwards from green is going to be easier than going the other direction.

lib/fork.js
jamestalmage: #219 will tell us if this was necessary.

lib/serialize-value.js
jamestalmage: TDIL:
```
let foo = function () {};
let bar = foo;

console.log(bar.name);
// => foo
```

the more you know

jamestalmage: not sure if it's just a babel thing or what.

BarryThePenguin: [function names](http://www.2ality.com/2015/09/function-names-es6.html)
```js
const bar = function baz() {};
console.log(bar.name); // baz
```

jamestalmage: Does not work in node `5.0.0`. Empty strings all around.
Works in Babel instrumented code though.

jamestalmage:
Just Sayin...

screenshot 2015-11-16 01 18 19

jamestalmage:
@sindresorhus 
Every comment except the separating out loud-rejection promises is addressed.

I think I can pull it out if you really think it is necessary.

floatdrop:
Great work! - reliable builds on Windows is must have. Rough edges can be improved from there.

I'm think that unhandled related code (and `serialize-value` replacement too?) should be moved in separate PR thou - it will be easier to find corresponding code to fixes later. Not critical thou.

jamestalmage:
`serialize-value` is part of the fix for uncaught exceptions, and that is part of the solution for reliable builds. Allowed to go uncaught, they cause the child to exit unpredictably (we are not sure exactly what messages the parent has received). With this we notify parent "hey, uncaught error here, bailing!", wait for acknowledgement, and then shut down. Otherwise we must rely on stderr / stdout from the child to be properly piped to stderr/stdout on the parent. For reasons unknown to me, that is just too unreliable on Windows to be a solution.

Unfortunately, I committed the `unhandled-rejection` stuff before the `uncaught-exception`, and there a number of conflicts between the two ([here](https://github.com/sindresorhus/ava/pull/206/files#diff-2cce40143051e25f811b56c79d619bf5R163), [here](https://github.com/sindresorhus/ava/pull/206/files#diff-2cce40143051e25f811b56c79d619bf5R173), etc.). Part of the frustration was that everything passed on my machine, yet failed on AppVeyor. Scroll through the [AppVeyor history](https://ci.appveyor.com/project/sindresorhus/ava/history) if you wanna see my name 3 dozen times next to a lot of red :angry:. It is probably not a huge deal to pull out the unhandled rejection stuff, but there will be merge conflicts to resolve before and after, and it just seems like extra work for not a lot of benefit.

I think I did do a pretty good job maintaining a good git history, and have squashed relevant commits together. Perhaps do not squash this PR so `git blame` and the commit log maintain finer grained details for each change?

floatdrop:
@jamestalmage no offence at all (sorry if it seems so, English is not my strongest part), really appreciate work you done in `ava` and related modules. I understand, that rebasing unhandledRejection is hard and not insisting on it. Thanks for explaining `serialize-value` part.

jamestalmage:
> no offence

none taken :smile: 

The only thing that has offended me during this process is Windows!

lib/serialize-value.js
```diff
@@ -0,0 +1,16 @@
...
+
+// Make a value ready for JSON.stringify() / process.send()
+
+module.exports = function serializeValue(value) {
+	if (typeof value === 'object') {
+		return destroyCircular(value);
+	}
+	if (typeof value === 'function') {
+		// JSON.stringify discards functions, leaving no context information once we serialize and send across.
+		// We replace thrown functions with a string to provide as much information to the user as possible.
```
sindresorhus: I'm just curious, when would anyone throw a function?

test/test.js
```diff
@@ -1058,14 +1058,50 @@ test('change process.cwd() to a test\'s directory', function (t) {
...
+		t.ok(/1 unhandled rejection[^s]/.test(stderr));
+		t.end();
+	});
+});
+
+test('uncaught exception will be reported to console', function (t) {
+	execCli('fixture/uncaught-exception.js', function (err, stdout, stderr) {
+		t.ok(err);
+		t.ok(/Can't catch me!/.test(stderr));
+		// TODO: This should get printed, but we reject the promise (ending all tests) instead of just ending that one test and reporting.
```
sindresorhus: For TODO's I prefer to include the username so we know who added it without having to git blame.

```
TODO(jamestalmage):
```

sindresorhus:
Looks good to me. Let's get this merged asap. Just waiting for @vdemedes to review.

lib/fork.js
```diff
@@ -18,6 +18,10 @@ module.exports = function (args) {
 
 	var ps = childProcess.fork(babel, args, options);
 
+	function send(command, data) {
+		ps.send({'ava-child-process-command': command, 'data': data});
```
vdemedes: can just be:

```
ps.send({
  name: command,
  data: data
});
```

lib/babel.js
```diff
@@ -32,20 +34,46 @@ module.exports = {
...
+	var command = message['ava-child-process-command'];
+	if (command) {
+		process.emit('ava-' + command, message.data);
+	}
+});
+
+process.on('ava-kill', function () {
+	setTimeout(function () {
 		process.exit(0);
+	}, process.env.AVA_APPVEYOR ? 100 : 0);
```
vdemedes: AppVeyor doesn't set env variables like Travis to identify itself?

cli.js
vdemedes: Yeah, there's 3 of `process.stdout.write('')` now for no explained reason.

appveyor.yml
```diff
@@ -7,9 +7,10 @@ environment:
 install:
   - ps: Install-Product node $env:nodejs_version
   - set CI=true
-  - npm -g install npm@latest
+  - set AVA_APPVEYOR=true
+  - npm -g install npm@latest || (timeout 30 && npm -g install npm@latest)
   - set PATH=%APPDATA%\npm;%PATH%
-  - npm install
+  - npm install || (timeout 30 && npm install)
```
vdemedes: how about just: `timeout 30 && npm install`?

lib/babel.js
kevva: Yes, it does. http://www.appveyor.com/docs/environment-variables

vdemedes: Great, in that case we don't need our own `AVA_APPVEYOR`.

sindresorhus: @vdemedes I think the intention is to only do it when *we* run AVA's tests on AppVeyor, not when the user run their tests there.

vdemedes: @sindresorhus Hmm, don't those fixes also needed for user's tests?

appveyor.yml
jamestalmage: this tries to reinstall after 30 seconds if the first npm install fails (which it does about every 10 times or so on AppVeyor).

vdemedes: Can we make this timeout shorter?

lib/babel.js
jamestalmage: @sindresorhus got the intent right.
I originally had it as APPVEYOR, until a conversation with @BarryThePenguin 

The only tests this affects is ones that parse the stdout / stderr from our CLI.
There is a chance by running the tests fast, that they miss the last line or two (a summary or the failed stack trace). It's highly unlikely they will be grepping our code CLI output to get a pass/fail on their own tests.

They will still always get the correct exit code - they might just be missing the last "1 test failed".
If users report it as a problem it is easy enough to correct down the road.

lib/fork.js
jamestalmage: My only thought there was wether or not we wanted to monkeypatch `process.on` on the child process, so users never saw our communication.

I can imagine a scenario where you are writing some code that *might* be a child process (or might be top level), and so they would will listen to `process.on('message')`. By choosing a unique property name here we could monkeypatch `on` and filter out these messages.  I reemit these on the child process as specific `ava-` prefixed events(`process.on('ava-kill', ...)`, `process.on('ava-cleanup')`, etc) - that it is really unlikely users will listen to.

None of that is implemented. Just leaving the door open.

cli.js
jamestalmage: oops - I meant to delete that other one.

There should be one each `stderr` and `stdout`. This belongs in the "it seamed to help" pile. Definitely room to experiment with removing it down the road.

sindresorhus:
Landed! Superb work tracking down all these issues @jamestalmage :)

I would honestly have just given up and removed AppVeyor :p

appveyor.yml
jamestalmage: I chose 30 seconds pretty arbitrarily, so I"m not opposed to shortening it and seeing what happens.
Why do you want it shorter?

jamestalmage:
Awesome!

> I would honestly have just given up and removed AppVeyor :p

Now you tell me!! :wink:

appveyor.yml
sindresorhus: I don't think it matters to be honest. AppVeyor is sooooo slow regardless. But we can try with 10 seconds to see if it still works.

sindresorhus: @jamestalmage Probably because it's annoying to always have to wait for AppVeyor before merging a PR.

vdemedes: > Why do you want it shorter?

So that tests are executed sooner.

sindresorhus:
> Now you tell me!! :wink:

Automated Windows testing is especially important for a test runner, though, so glad we have it, regardless of what I feel about Windows.

appveyor.yml
jamestalmage: > Probably because it's annoying to always have to wait for AppVeyor before merging a PR. 

Yes. That.

I got a lot of TV watched this weekend. Push a commit, watch 20 minutes... Still broken... Push a commit, watch 20 minutes...

test/test.js
jamestalmage: cool, I'm assuming you took care of it in the merge? Or do you want me to go handle it now?

appveyor.yml
sindresorhus: Haha :p

Sounds like me too. I was bumping dependencies on a gadzillion modules this weekend on a super slow connection. `$ david update`, wait 10 minutes, commit, push, `npm publish`, wait 5 minutes, then next module...

test/test.js
sindresorhus: I'll do it.

PR #208, opened 2015-11-13
Properly handle empty results from test files - fixes #198

jamestalmage:
SEE #213 FIRST
Fixes #198.

PR #210, opened 2015-11-13
ignore non test files, fixes #209

### LINKED ISSUE #209: error thrown when files do not require ava
When running AVA against empty files or files that do not require ava, this error is thrown on v.0.4.2.
I tried with master, and it just hangs up.
getting the same error as in #198 (I opened a new issue because I'm not sure it's the same cause.):
``` 
TypeError: Cannot read property 'stats' of undefined
    at C:\Users\Frontend\AppData\Roaming\nvm\v0.12.5\node_modules\ava\cli.js:139:16
    at Array.map (native)
    at exit (C:\Users\Frontend\AppData\Roaming\nvm\v0.12.5\node_modules\ava\cli.js:138:22)
From previous event:
    at Object.<anonymous> (C:\Users\Frontend\AppData\Roaming\nvm\v0.12.5\node_modules\ava\cli.js:216:18)
    at Module._compile (module.js:460:26)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Function.Module.runMain (module.js:501:10)
    at startup (node.js:129:16)
    at node.js:814:3 
```

In master, I saw that if the file transpiled by babel does is not processed by AVA and thus emits the needed events to end, the process will just hang up.

GabiGrin:
I took a long shot with this PR but hopefully it helps :)

This fix has 2 parts:

1) When a file that didn't include AVA ran inside the babel.js child  process the process will never exit as no piece of code will ever emit the needed events. 
The only way I found around this, because of the complex process nature of the code, is to send a message once AVA runs on a file, and another message when babel finishes. The result is that inside fork.js we can now know if a file that did not include AVA was read and exit if so.
PS: this only happens in master, not in v0.4.2, probably because the fix in [here](https://github.com/sindresorhus/ava/commit/6a8f0e7a7caf833627c4645e74f8632e1c0c02b8)

2) If a file didn't include AVA, test results are not emitted and thus an the error in the mentioned issue is is thrown. I filtered the all undefined results to fix this.

Also, added a test that will fail on master / latest published version.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -132,6 +132,10 @@ function sum(arr, key) {
 }
 
 function exit(results) {
+	// in case of non-test files, the result will be undefined so we remove them
+	results = results.filter(function (result) {
+		return result !== undefined;
+	});
 	// assemble stats from all tests
 	var stats = results.map(function (result) {
 		return result.stats;
--- a/index.js
+++ b/index.js
@@ -66,6 +66,11 @@ function exit() {
 	});
 }
 
+process.send({
+	name: 'started',
+	data: {}
+});
+
 setImmediate(function () {
 	runner.on('test', test);
 	runner.run().then(exit);
--- a/lib/babel.js
+++ b/lib/babel.js
@@ -25,8 +25,14 @@ var options = {
 };
 
 var transpiled = babel.transformFileSync(testPath, options);
+
 requireFromString(transpiled.code, testPath, {
 	appendPaths: module.paths
+});
+
+process.send({
+	name: 'babel-ended',
+	data: {}
 });
 
 process.on('message', function (message) {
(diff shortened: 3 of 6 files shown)


jamestalmage:
IMO, the proper thing to do in a file without tests is to fail. Helper files should be prefixed with a `_` character.

> The only way I found around this, because of the complex process nature of the code, is to send a message once AVA runs on a file, and another message when babel finishes

See #208 for my solution to the problem. It's equally convoluted, but does not require sending messages between processes.

GabiGrin:
Much more elegant!

I actually got into this issue while running AVA on directly on a folder (via gulp-ava) with other files, but that was just by mistake.

I'll close thi. Thanks!

Issue #182, opened 2015-11-10
Avoid loading babel in forked processes.


jamestalmage:
We are paying a hefty penalty to load Babel once per forked process.

I have played around with a way to share one Babel instances across every forked process (though it does not necessarily need to be limited to a single instance).

https://github.com/jamestalmage/__share_transpile_across_process

With some very naive benchmarks I am seeing < 3 seconds using this method compared to > 16 loading up a Babel instance per fork.

floatdrop:
@jamestalmage thought of this too. We can transpile files with babel in master process and load transpiled code from cache (as soon as I deal with #118).

sindresorhus:
That's a cool idea. I like it.

jamestalmage:
What is your plan for the cache? Write it all to disk before launching forks?

I'm actually firing up an http server and then using `child_process.execSync('cur', ...)` to fetch the transpiled code.

jamestalmage:
The major problem there is that it requires actually forking an entirely new process for every fetch, which is not free. Using a direct execution of `curl` vs forking another node process is way better, but still not great.

For the initial test file, we could use IPC (maybe even do static analysis of the Babel AST to find additional require calls).

Dynamic requires will be a problem. That is where I think my strategy will help. If somebody does a lot of dynamic calls there is even a chance my strategy may hurt, so we probably need a way for people to turn it off.

floatdrop:
@jamestalmage yes, but do it step by step (transpile one test and launch it), not all at once.

jamestalmage:
Gotcha.
Seems reasonable.

jamestalmage:
@sindresorhus We need a good real-world library to performance test this against. I know pageres is mentioned in the readme, but given the nature of that project (I am just guessing), that a lot of time is spent on I/O. Is there one with a high test-file count where I/O is minimal (where improvements to the load time of AVA forks would be most noticeable).

sindresorhus:
@jamestalmage https://github.com/sindresorhus/got has a lot of tests and uses a local server for most, so the tests are pretty fast.

vdemedes:
Really good idea! Well done @jamestalmage

jamestalmage:
This was fixed in ~`0.10.0`

Issue #148, opened 2015-11-07
Node.js version specific Babel presets


sindresorhus:
So Babel only transpiles features not already supported by the platform.

@jamestalmage said he's interested in looking into this.

jamestalmage:
Looks like work has already begun:

https://www.npmjs.com/package/babel-preset-es2015-node5
https://www.npmjs.com/package/babel-preset-es2015-node4
https://www.npmjs.com/package/babel-preset-node5

Still, I think it might be better to automate this.

I am thinking of create a test-file for each of the [available plugins](https://babeljs.io/docs/plugins/#es2015) that uses the specific feature that plugin provides.

Then a script to require each in a try/catch and see if they throw / fail to compile. That gives us a nice automated update path as future versions are released. I am hoping babel's own test suite will make this all a simple matter of cutting and pasting code samples... There are a lot of plugins.

sindresorhus:
Wouldn't live feature testing be slow? I had a similar idea of doing the feature testing for every major Node.js version before publishing.

jamestalmage:
Yes - as part of the update process. Not live testing.

After the test suite runs we output a file for that specific version in a subfolder:

```js
// index.js
module.exports = require('./plugin-lists/' + process.version);
```

or something like that.

sindresorhus:
:+1: That's exactly what I was thinking.

jamestalmage:
It might be cool to fetch the actual plugins from `npm` on demand as well (so `npm install` goes faster in newer versions of node). We could give back a few CPU cycles to travis-ci.org that way.

Definitely not something to target for the first release, but might be interesting to play with down the road.

hzoo:
Haven't implemented it yet but I would like to see at least a preset where you can specify that environment you want to target and it would have a predetermined list of transforms based on what it supports. You could just use the data or use `env` to do different things. https://github.com/babel/babel/pull/3476

novemberborn:
Just saw this: https://github.com/gabmontes/babel-preset-latest-minimal

sindresorhus:
@novemberborn I'm a little bit sceptical of feature testing. New features might be buggy or slow. I would be more comfortable with a preset that choose plugins based on vetted testing on various Node.js versions. Maybe https://github.com/babel/babel-preset-env when it gets more mature.

novemberborn:
Is fixed by #1193.

PR #212, opened 2015-11-13
WIP - Babel 6 upgrade

jamestalmage:
This works but Babel is really slow.

Also, I think there are some problems with the latest babel-plugin-transform-runtime that required me to pin back a minor release.

https://github.com/babel/babel/issues/2954

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -40,6 +40,10 @@ function Test(title, fn) {
 	if (this.title === 'callee$0$0') {
 		this.title = '[anonymous]';
 	}
+
+	Object.keys(Test.prototype).forEach(function (key) {
+		this[key] = this[key].bind(this);
+	}, this);
 }
 
 module.exports = Test;
@@ -147,7 +151,17 @@ Test.prototype.run = function () {
 	}.bind(this));
 };
 
-Test.prototype.end = function () {
+Test.prototype.end = function (err) {
+	if (err) {
+		this.assertError = new assert.AssertionError({
+			actual: err,
+			message: 'Callback called with an error → ' + err,
+			operator: 'callback'
+		});
+
+		return this.exit();
+	}
+
 	if (this.endCalled) {
 		throw new Error('.end() called more than once');
 	}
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -63,9 +64,12 @@
...
     "ava-init": "^0.1.0",
-    "babel-core": "^5.8.23",
-    "babel-plugin-espower": "^1.1.0",
-    "babel-runtime": "^5.8.29",
+    "babel-core": "^6.1.20",
+    "babel-plugin-espower": "^2.0.0",
+    "babel-plugin-transform-runtime": "^6.1.18",
+    "babel-preset-es2015": "6.0.14",
+    "babel-preset-stage-3": "6.0.14",
+    "babel-runtime": "^5.8.34",
```
jamestalmage: @floatdrop I think the reason I need to depend on `babel-runtime@5` here is because of the way we are mucking with module paths. Many of babels transforms depend on `babel-runtime@5` internally, so I think our hack is forcing unintentionally pushing them back to `v6`.

That's my best guess right now anyways.

jamestalmage:
Babel 6 has crazy long bootup.
#148 and #182 are going to become really important.

package.json
jamestalmage: > the reason I need to depend on babel-runtime@5 here is because of the way we are mucking with module paths

This may not be true. I just discovered that running the babel test suite had created a whole bunch of global symlinks. Not sure if it was what was causing this, I will retest later.

sindresorhus:
> Babel 6 has crazy long bootup.

Oh, that sucks... You might want to open an issue on Babel about it. You can use [time-require](https://github.com/jaguard/time-require) to see what's slowing it down.

jamestalmage:
~850ms: https://github.com/jamestalmage/__time-require-babel-5

~3.6 seconds: https://github.com/jamestalmage/__time-require-babel-6

jamestalmage:
Never Mind:

`npm dedupe` improves babel 6 times drastically.

I still had npm@2 lying around in my node `4.1.2` install

So it seems you trade a longer `npm install`, for faster requires.  Worth it, it seems to me.

sindresorhus:
Cool, I guess it's fine then :)

PR #213, opened 2015-11-14
(DO NOT MERGE): Reliable output

jamestalmage:
This combines:
#206 
#208 
#211 

In hopes that the combined strategies are the answer to brittle AppVeyor tests.

DO NOT MERGE:
Over time this PR Became a dump for every random thought I had on how to make AppVeyor work. I need to keep it open so AppVeyor continues to build, but once I find a reliable strategy, I will submit separate, cleaner PR's.

PR #215, opened 2015-11-14
Make AppVeyor retry failed runs.

jamestalmage:
Tests are just too brittle on AppVeyor. `npm install` fails frequently,
and Node on Windows does not seem to reliably flush stdout/stderr
of forked child processes.

I don't like this, but I'm at a loss for other solutions.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,9 +7,9 @@ environment:
 install:
   - ps: Install-Product node $env:nodejs_version
   - set CI=true
-  - npm -g install npm@latest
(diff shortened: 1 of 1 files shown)


jamestalmage:
Argghhh!!

NODE 10 WINDOWS!!!

Despite the failure - This should be merged.

There is obviously something else in #213 that let it pass 5 times in a row, but this was definitely part of the solution.

jamestalmage:
Closing in favor of #206

PR #216, opened 2015-11-15
add support for *.spec.js spec file naming

jankuca:
This change adds support for `whatever.spec.js` test file names (in the context of prefixing test results by the CLI) which is a quite common practice really. The former logic which trims `test-` prefixes stays in place.

```
test-whatever.js => whatever
test-whatever.spec.js => test-whatever
```

It might even make sense to include support for the `whateverSpec.js` variation.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -64,9 +64,16 @@ function prefixTitle(file) {
 	base += path.sep;
 
 	var prefix = path.relative(process.cwd(), file)
-		.replace(/test\-/g, '')
+		.replace(base, '');
+
+	if (/\.spec\./.test(prefix)) {
+		prefix = prefix.replace(/\.spec\./, '.');
+	} else {
+		prefix = prefix.replace(/test\-/g, '');
+	}
+
+	prefix = prefix
 		.replace(/\.js$/, '')
-		.replace(base, '')
 		.split(path.sep).join(separator);
 
 	if (prefix.length > 0) {


sindresorhus:
> which is a quite common practice really

Never seen it before. Can you link to something that proves your point?

jankuca:
A quick Google query popped out a few examples:

http://backbone-testing.com/
> Chapter 1: Setting up a Test Infrastructure
> - Trying out the test libraries: Some first basic unit tests using Mocha, Chai, and SinonJS.
>   - **hello.spec.js**
> - Test Failures: Different types of test failures.
>   - **failure.spec.js**
> - Test Timing: Tests that take different times, which Mocha annotates for "medium" and "slow" tests. Also has one test timeout failure.
>   - **timing.spec.js**

http://blog.revolunet.com/blog/2013/12/05/unit-testing-angularjs-directive/
> Here’s our example “files” section :
> files: [
>   ...
>   "src/angular-stepper.js",                           <-- our component source code
>   "src/**angular-stepper.spec.js**"                       <-- our component test suite
> ]

https://review.openstack.org/#/c/184543/28/openstack_dashboard/static/openstack-service-api/common-test.spec.js,unified

The latter (`whateverSpec.js`) convention is used by jasmine-node for instance – https://github.com/mhevery/jasmine-node
> Note: your specification files must be named as **spec.js**, **spec.coffee** or **spec.litcoffee**, which matches the regular expression /spec\.(js|coffee|litcoffee)$/i; otherwise jasmine-node won't find them! For example, sampleSpecs.js is wrong, sampleSpec.js is right.

jankuca:
> And where do you keep these *.spec.js files? In a test folder? In root?

I personally keep them in the `test/` folder, yes.

sindresorhus:
> I personally keep them in the test/ folder, yes.

Ok, so it seems it's for marking something as a test when you have both the test and the actual code in the same directory. Not saying we won't accept this PR, but you are not doing this, so you really don't need the `.spec` postfix in your case.

jankuca:
> …when you have both the test and the actual code in the same directory

Yes, I've seen that a couple of times and even tried it. I've come to the conclusion that it is more practical to have spec files in a separate top-level folder.

The real reason I keep using the suffix is mostly because it is possible to determine the type of a file (implementation/spec/mock) from its basename.

screen shot 2015-11-15 at 9 34 03 pm

schnittstabil:
> I personally keep them in the test/ folder, yes.

@jankuca Aren't your test files run? I thought they should, because:
```
  Usage
    ava [<file|folder|glob> ...]
…
  Default patterns when no arguments:
  test.js test-*.js test/*.js
```

jankuca:
> Aren't your test files run?

No, they are, everything's working fine. Only the result prefixes are wrong:

screen shot 2015-11-15 at 9 43 57 pm

cli.js
```diff
@@ -64,9 +64,16 @@ function prefixTitle(file) {
 	base += path.sep;
 
 	var prefix = path.relative(process.cwd(), file)
-		.replace(/test\-/g, '')
+		.replace(base, '');
+
```
vdemedes: to avoid the `if` condition, you could just do this:

```js
.replace(/\.spec/, '')
.replace(/test\-/g, '')
```

vdemedes:
I am now wondering if it's not just easier to rename `.spec` suffix from file names. That way we won't tailor AVA to every possible use case out there.

vdemedes:
But for now, @sindresorhus and I think that this is an easy fix'n'win, so I guess we'll just merge this PR ;)

vdemedes:
@jankuca could you please fix the comments in this PR?

vdemedes:
@jankuca Never mind, I made the changes myself ;)

jankuca:
Thanks!

The reason I've put the `if` there was to handle a single spec file naming convention at once.

PR #221, opened 2015-11-16
WIP - Babel 6

jamestalmage:
See #212 for an earlier PR.

Blocking issue is: 
https://github.com/babel/babel/issues/2954

I have to rely on `babel-polyfill` to get tests to pass until that is taken care of.

### DIFF
--- a/lib/babel.js
+++ b/lib/babel.js
@@ -4,7 +4,6 @@ var requireFromString = require('require-from-string');
 var loudRejection = require('loud-rejection/api')(process);
 var resolveFrom = require('resolve-from');
 var serializeValue = require('./serialize-value');
-require('babel-polyfill');
 var send = require('./send');
 
 var testPath = process.argv[2];
@@ -13,7 +12,7 @@ var testPath = process.argv[2];
 var babel;
 
 try {
-	var localBabel = resolveFrom('.', 'babel-core') || resolveFrom('.', 'babel');
+	var localBabel = resolveFrom('.', 'babel-core');
 	babel = require(localBabel);
 } catch (err) {
 	babel = require('babel-core');
@@ -26,7 +25,7 @@ var powerAssert = createEspowerPlugin(babel, {
 
 var options = {
 	presets: ['stage-3', 'es2015'],
-	plugins: [powerAssert]
+	plugins: [powerAssert, 'transform-runtime']
 };
 
 // check if test files required ava and show error, when they didn't
--- a/lib/babel.js
+++ b/lib/babel.js
@@ -3,8 +3,8 @@ var createEspowerPlugin = require('babel-plugin-espower/create');
 var requireFromString = require('require-from-string');
 var loudRejection = require('loud-rejection/api')(process);
 var resolveFrom = require('resolve-from');
-var hasGenerator = require('has-generator');
 var serializeValue = require('./serialize-value');
+require('babel-polyfill');
 var send = require('./send');
 
 var testPath = process.argv[2];
@@ -24,10 +24,8 @@ var powerAssert = createEspowerPlugin(babel, {
 	patterns: require('./enhance-assert').PATTERNS
 });
 
-// if generators are not supported, use regenerator
 var options = {
-	blacklist: hasGenerator ? ['regenerator'] : [],
-	optional: hasGenerator ? ['asyncToGenerator', 'runtime'] : ['runtime'],
+	presets: ['stage-3', 'es2015'],
 	plugins: [powerAssert]
 };
 
@@ -40,9 +38,7 @@ process.on('uncaughtException', function (exception) {
 
 // include test file
 var transpiled = babel.transformFileSync(testPath, options);
-requireFromString(transpiled.code, testPath, {
-	appendPaths: module.paths
-});
+requireFromString(transpiled.code, testPath);
 
 // if ava was not required, show an error
 if (!exports.avaRequired) {
--- a/package.json
+++ b/package.json
@@ -73,9 +73,12 @@
   "dependencies": {
     "arr-flatten": "^1.0.1",
     "ava-init": "^0.1.0",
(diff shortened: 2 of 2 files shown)


sindresorhus:
Relevant: https://github.com/babel/babel/pull/3094

sindresorhus:
> Blocking issue is: babel/babel#2954

Is this still blocking? Would you mind updating the link? I have no idea what issue on the new issue tracker it is.

jamestalmage:
Ugh, why did they do that!

I looked and can't find the original issue in the new tracker.

I will look into this later today.

sindresorhus:
They were fed up with GitHub being non-responsive on improving the issue system and they needed something more powerful.

If it helps, the issue title were: `(0 , _typeof3.default) is not a function`

thejameskyle:
It's the same issue number. https://phabricator.babeljs.io/T2954 which was closed as a duplicate of https://phabricator.babeljs.io/T6644

jamestalmage:
Yeah, I searched for that and did not find.

They have released a couple runtimes since, so probably the easiest option
is for me to try again. Won't take me long. Just has to happen after work.


James Talmage
Owner
JR Technical
cell: 248-259-3137
fax:  248-498-6071

jamestalmage:
~~Yeah, I searched for that and did not find~~

~~They have released a couple runtimes since, so probably the easiest option is for me to try again. Won't take me long. Just has to happen after work.~~

Thanks @thejameskyle

jamestalmage:
Looks like there is a [pending PR](https://github.com/babel/babel/pull/3126/files) to fix our blocking issue.

lib/babel.js
```diff
@@ -13,7 +12,7 @@ var testPath = process.argv[2];
 var babel;
 
 try {
-	var localBabel = resolveFrom('.', 'babel-core') || resolveFrom('.', 'babel');
+	var localBabel = resolveFrom('.', 'babel-core');
 	babel = require(localBabel);
 } catch (err) {
 	babel = require('babel-core');
```
sindresorhus: I think we should drop using the the user's local Babel version. Instead we should just document that to use npm@3 or `npm deduplicate`. The user can then just enforce any Babel 6 version and we will use the same since our dependency is loose.

lib/babel.js
```diff
@@ -24,11 +23,9 @@ var powerAssert = createEspowerPlugin(babel, {
 	patterns: require('./enhance-assert').PATTERNS
 });
 
-// if generators are not supported, use regenerator
 var options = {
-	blacklist: hasGenerator ? ['regenerator'] : [],
-	optional: hasGenerator ? ['asyncToGenerator', 'runtime'] : ['runtime'],
```
sindresorhus: Why is the generator check no longer needed?

package.json
```diff
@@ -73,9 +73,13 @@
   "dependencies": {
     "arr-flatten": "^1.0.1",
     "ava-init": "^0.1.0",
-    "babel-core": "^5.8.23",
-    "babel-plugin-espower": "^1.1.0",
-    "babel-runtime": "^5.8.29",
+    "babel-core": "6.0.14",
```
sindresorhus: Use caret `^`

lib/babel.js
jamestalmage: I agree.

lib/babel.js
vdemedes: Also curious about this one.

jamestalmage: Oh, it still would be if we want to avoid the unnecessary transform. I think maybe `asyncToGenerator` was rolled in a preset? There's certainly a way around it.

sindresorhus: http://babeljs.io/docs/plugins/transform-async-to-generator/

jamestalmage: http://babeljs.io/docs/plugins/preset-stage-3/

jamestalmage: It was included in that preset, that's why I dropped it. Every intention of adding it back.

We are still waiting on https://phabricator.babeljs.io/T6644. I will approach this again when that closes.

sindresorhus: Oh, I didn't realize async functions went to stage3. That's great! :)

package.json
jamestalmage: Yep, at the time, latest was `6.0.18` which I couldn't get to work at all. That is why it is pinned.
They are on `6.3.?` now, so this likely isn't necessary anymore.

sindresorhus: @jamestalmage Yeah, was only commenting what I want the final PR to look like. We can wait on doing it until this PR is ready.

sindresorhus:
Another take on fixing the issue => https://github.com/babel/babel/pull/3142

sindresorhus:
I believe this is also blocking => https://github.com/babel/babel/pull/3139

jamestalmage:
> I believe this is also blocking => babel/babel#3139

No, shouldn't be. `nyc` already works around it. Babel's hook is perfectly fine if installed first. `nyc` has a solution in place that allows that allows the babel hook to be installed second, after `nyc` (but breaks if further hooks are installed). 

[`capture-require`](https://github.com/jamestalmage/capture-require/blob/master/test/execution-order.js) will solve that for them, but it wreaks havoc on their test suite. `capture-require` can be used multiple times without breaking previous hooks. Their current test suite relies on the fact that `nyc.wrapRequire()` breaks the previously installed hook. That is [basically fixed](https://github.com/bcoe/nyc/pull/89), just need to get some consensus around it.

jamestalmage:
> Another take on fixing the issue => babel/babel#3142

Yeah, I think that's the one that will land

sindresorhus:
Oh ok, that's great. So as soon as the Babel issue is fixed we can upgrade then.

jamestalmage:
Closing in favor of #333 

There is only one concern mentioned here that still applies, and I have transferred that to the new PR.

PR #225, opened 2015-11-16
add covenience util for debugging

jamestalmage:
Launches a single test in the same process

Logs what would be sent to the parent process (in magenta!).

Makes sure the `ava-cleanup` method gets called on exit (so you can verify unhandled rejections).


sindresorhus:
You need to add it to `files` in package.json.

jamestalmage:
Oh sorry, I wasn't clear - I intended it just for internal development use. Not public consumption. I guess it would not be a terrible idea to publish it. We could ask users to execute tests with it and send the output. Might help hunt down some difficult bug someday.

What do you think?

sindresorhus:
@jamestalmage Oh ok. Then let's keep it internal like this for now until a need arises.

sindresorhus:
@jamestalmage Can you add a short example on how to use this to https://github.com/sindresorhus/ava/blob/master/maintaining.md ?

@vdemedes Thoughts?

vdemedes:
I did not have a need for something like that for now, but if you find it useful, why not. But still, I don't think it should be included on npm. Perhaps leave it as an open PR?

sindresorhus:
@vdemedes It's currently not included in the npm package as it's not listed in the `files` field in package.json. I would rather have it in the repo than an open PR.

vdemedes:
Oh wait, I have an idea. Why don't we use [debug](https://npmjs.com/package/debug) in AVA? It will allow for an extremely easy debugging and a simple way to enable it. Plus, we won't need to ship/include one more file.

```
$ DEBUG=ava ava test.js
$ DEBUG=ava:fork ava test.js
$ DEBUG=ava:* ava test.js
```

And it's been more or less a standard way to enable logs for a certain module (check `debug` readme).

sindresorhus:
:+1:

jamestalmage:
That still requires forking a process. It doesn't capture everything sent via process.send.

vdemedes:
```js
fork.on('message', function (message) {
  // log message
});
```

?

vdemedes:
It will capture everything that a forked process sends via `process.send()`, won't it?

jamestalmage:
Did somebody force push to master?

vdemedes:
Nope, but somehow `maintaining.md` ended up in your PR. Rebase gone wrong?

vdemedes:
Anyway, I think this PR should be closed in favor of using `debug` utility, so don't bother fixing it.

sindresorhus:
Closing in favor of https://github.com/sindresorhus/ava/issues/242.

PR #227, opened 2015-11-16
simplify context sharing between beforeEach and test

jamestalmage:
The wrapper is not necessary, even for primitives.

// @sindresorhus


vdemedes:
I like this change, @sindresorhus what do you think?

sindresorhus:
Yeah, not sure what I was thinking there.

PR #228, opened 2015-11-16
Output the first assertion. fixes #121

### LINKED ISSUE #121: Output every failed assertion or just the first one (currently outputs just the last one)
```js
test('foo', t => {
	t.plan(2);
	var a = 2;
	var b = 3;
	var c = 4;
	t.ok(a === b);
	t.ok(b === c);
});
```

output:
```
 ✖ foo   
  t.ok(b === c)
       |     | 
       3     4 
  

  1 test failed
```

It would be better if either:
 1. Both assertion failures printed.
 2. If there can only be one, print the first.

jamestalmage:
Currently the final assertion error is shown, this instead shows the first encountered assertion error.
This is generally more convenient than showing the final assertion error, since a final error may 
not be that helpful without knowing which preconditions failed.

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -166,17 +166,18 @@ Runner.prototype.serial = function (tests) {
 };
 
 Runner.prototype._addTestResult = function (test) {
-	if (test.assertError) {
-		this.stats.failCount++;
-	}
-
 	var props = {
 		duration: test.duration,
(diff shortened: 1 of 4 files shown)


jamestalmage:
Known issue:
https://github.com/scottcorgan/tap-spec/issues/40

PR #230, opened 2015-11-17
This should fail!

jamestalmage:
on my machine this exits after 43 tests with a zero exit code.

it definitely should fail the build.

it is likely a problem in tape. I am just making sure the problem is not local to me before submitting.

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -24,11 +24,6 @@ function Runner(opts) {
 
 	this.results = [];
 
-	this.stats = {
-		failCount: 0,
-		testCount: 0
-	};
-
 	this.tests = {
 		concurrent: [],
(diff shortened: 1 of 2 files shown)


vdemedes:
Please, try to write better titles and describe your PRs properly. There's no rush and any bonuses over creating PRs fast. From this title and lack of description, it seems that this PR demonstrates some issue, stating that AVA should fail, but it's not.

It would be great, if we did not have to look at changed files to get a sense of what is this PR about. Thanks.

vdemedes:
And I see that this PR contains the same code as #231, is it a duplicate?

jamestalmage:
Yes sorry,
Got pulled away. I'll clean house here shortly.

PR #231, opened 2015-11-17
cause build to fail if `tape` exits with a non-zero exit code

jamestalmage:
This is the only solution I have found so far.

Demonstration of the issue: #230

This is a known problem: https://github.com/scottcorgan/tap-spec/issues/40

Note that this does not work on Windows. The build will still exhibit false positives on Windows. I added it to the `test-win` script only because I often use that as a shortcut to skip the linter. It does not interfere with operation on Windows.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,8 +32,8 @@
     "node": ">=0.10.0"
   },
   "scripts": {
-    "test": "xo && nyc tape test/*.js | tap-spec",
-    "test-win": "tape test/*.js | tap-spec",
+    "test": "xo && nyc tape test/*.js | tap-spec; test \"${PIPESTATUS[0]} -eq 0\"",
+    "test-win": "tape test/*.js | tap-spec; test \"${PIPESTATUS[0]} -eq 0\"",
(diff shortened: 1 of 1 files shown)


sindresorhus:
Can't we just use another non-broken reporter? Or fix the `tap-spec` one.

jamestalmage:
> Or fix the tap-spec one

I am minutes out from submitting a PR for exactly that (need to fix pinkie first though).

I can not find any way of actually checking what the exit code of the process being piped in is. The issue with tap-spec is that it is not inspecting the `plan` line, and validating it against the number of results it sees. 

For those who are interested, this is a plan line in TAP:
```
1...48
```
It means "plan for tests 1 to 48"

It always comes at the end in tape, not sure about other reporters.

vdemedes:
Could you post some example test code to reproduce this?

jamestalmage:
Yes, if you clone this and run tests it passes (on OSX at least).

Travis does not like this, so I've opted for [another solution](https://github.com/scottcorgan/tap-out/pull/18).

PR #233, opened 2015-11-17
use faucet tap reporter

jamestalmage:
Until this is resolved: https://github.com/scottcorgan/tap-spec/issues/40

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,8 +32,8 @@
     "node": ">=0.10.0"
   },
   "scripts": {
-    "test": "xo && nyc tape test/*.js | tap-spec",
-    "test-win": "tape test/*.js | tap-spec",
+    "test": "xo && nyc tape test/*.js | faucet",
+    "test-win": "tape test/*.js | faucet",
     "coveralls": "nyc report --reporter=text-lcov | coveralls"
   },
   "files": [
@@ -102,9 +102,9 @@
   },
   "devDependencies": {
     "coveralls": "^2.11.4",
+    "faucet": "0.0.1",
     "nyc": "^3.2.2",
     "signal-exit": "^2.1.2",
-    "tap-spec": "^4.1.0",
     "tape": "^4.0.0",
     "xo": "*",
     "zen-observable": "^0.1.6"


vdemedes:
The output on Travis is a bit crappy - https://travis-ci.org/sindresorhus/ava/jobs/91561792. Although faucet is really cool.

jamestalmage:
Yeah, we can switch back when `tap-spec` gets fixed. I have a PR in for it's dependency, so hopefully he follows through and pulls that upstream.

I would prefer to continue working on AVA for now and just go with this. If he doesn't merge and pull the fix upstream in a few days, I will patch `tap-spec` myself and we can use that.

vdemedes:
Regarding this tap-spec issue #231, is there a way I can reproduce this? I wrote a failing test and tap-spec exited correctly with `1` exit code.

jamestalmage:
You need to throw an uncaught error a `tape` test (so the `tape` command exits suddenly).

A failing test that does not cause the process to exit will still fail the build.

jamestalmage:
See this build: https://travis-ci.org/sindresorhus/ava/jobs/91510425

Exits passing after only 45 tests.

vdemedes:
Damn, still can't reproduce. My test file:

```js

test(function (t) {
	t.end();
});

test(function (t) {
	var obj = {};
	obj.child.value = 1;
});

test(function (t) {
	t.end();
});

```

vdemedes:
Ok, got it, I was using `tap` instead of `tape`. Maybe we should just use `tap` in AVA tests? I can't find the difference between those too.

// @sindresorhus

sindresorhus:
@vdemedes I'm fine with switching to `tap`. It's more actively developed and generally better.

`tape` tl;dr, `tape` was initially made to add browser support to `tap`. `tap` also had a long period of inactivity, but after npm decided to beef up their test suite it has gotten a lot of <3.

vdemedes:
Great, so let's just use `tap` and `tap-spec` reporter.

jamestalmage:
closed in favor of #234

PR #234, opened 2015-11-17
use tap test runner instead of tape.

jamestalmage:


vdemedes:
Instead of:

```js
var tap = require('tap');
var test = tap.test;
```

you can just:

```js
```

jamestalmage:
done.
Why does `tap` set the timeout per file? That is crazy. Why not per individual test?

I understand why you guys were arguing for 30 seconds now.

jamestalmage:
Let's do "per test" timeouts in AVA.

vdemedes:
`$ tap -t 0 test/*.js` should solve the timeout issue.

jamestalmage:
I did 150 seconds.

vdemedes:
`tap -t150` is not a good way to specify a value. It should be `tap -t 150`.

sindresorhus:
I'm fine with 150s. That way it doesn't hang the CI forever if we have some kind of timeout issue.

vdemedes:
Yeah, global timeout should be there, was just pointing out a better way to set an argument value.

sindresorhus:
I prefer to full flag with `=` too, for readability: `tap --timeout=150`.

vdemedes:
Sure, that works too!

jamestalmage:
done

vdemedes:
Thanks!

Issue #229, opened 2015-11-17
Consider allowing custom test preprocessors


billyjanitsch:
I understand the philosophy of keeping the test environment separate from the lib environment. If I'm writing a lib in ES5 but want to write tests in ES6, it makes sense, and I think it should be kept as a default behavior. However, there are cases where AVA's custom transpiling and non-polluting `require` are more of a hinderance than a benefit.

Suppose I'm already transpiling my lib with a transpiler more advanced than AVA's (e.g. ES6/7 stage 0 + JSX + additional babel plugins). I probably want to transpile my tests with the same transpiler to let them make use of new language features, and for the sake of consistent style and linting across my project. I also probably want my tests' dependencies transpiled automatically. #202 and #111 would help, but only if I'm using `babel`.

Suppose I'm using `webpack` instead (with `babel-loader`, but also `css-loader` to handle requiring CSS modules). I can't compile my build before running tests without losing test access to all of my lib's non-exported internals, and there's no `import 'babel-core/register'`-type hook I can include in each of my test files. A custom compiler for my tests' dependencies as suggested in #111 would be inefficient, since if a test imported more than one file, webpack would have to compile them separately.

Instead, I want an option to specify a custom pre-processor _for my test files_, overwriting the one that AVA provides by default. For example, I could pass `babel-core/register` (which would make my tests' dependencies compile automatically, using my own `babel` settings), or a custom one which handles `webpack` integration. This is what Mocha and Karma do, and I think it's a very clean solution.

jamestalmage:
> A custom compiler for my tests' dependencies as suggested in #111 would be inefficient, since if a test imported more than one file, webpack would have to compile them separately.

Off topic here, but Babel suffers from the same issue across processes. Since AVA runs each test in a separate process, if you import it `test-a.js`, and `test-b.js` it compiles it once in each process. There is work on a solution (#189), that might be usable to help with that.

Right now, the intricacies of getting this to work well and work fast are challenging enough targeting Babel only.

Any custom pre-processor you wrote would have to handle the IPC protocol we already have in place, would have to apply the `power-assert` transform on its own. As well as integrate communication of uncaught exceptions, and unhandled promise rejections. All that is currently bundled in `babel.js`. Some would be fairly easy to extract and make generic, some would not.

An additional concern is that this project is currently moving REALLY fast. Maintaining multiple preprocessor solutions at this point will slow us down, which I think would be bad for the project overall.

[...]

PR #236, opened 2015-11-17
Clean up forking and messages

vdemedes:
This is a clean up kind of PR, not refactoring.

### DIFF
--- a/lib/babel.js
+++ b/lib/babel.js
@@ -1,12 +1,16 @@
 'use strict';
-var loudRejection = require('loud-rejection/api')(process);
-var resolveFrom = require('resolve-from');
 var createEspowerPlugin = require('babel-plugin-espower/create');
 var requireFromString = require('require-from-string');
+var loudRejection = require('loud-rejection/api')(process);
+var resolveFrom = require('resolve-from');
 var serializeValue = require('./serialize-value');
+var send = require('./send');
 
+// if node's major version part is >= 1, generators are supported
 var hasGenerators = parseInt(process.version.slice(1), 10) > 0;
 var testPath = process.argv[2];
+
+// include local babel and fallback to ava's babel
 var babel;
 
 try {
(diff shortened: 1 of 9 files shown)


lib/runner.js
```diff
@@ -1,9 +1,10 @@
 'use strict';
-var util = require('util');
 var EventEmitter = require('events').EventEmitter;
 var Promise = require('bluebird');
 var hasFlag = require('has-flag');
+var util = require('util');
```
sindresorhus: I prefer to have builtin requires grouped first.

lib/babel.js
```diff
@@ -16,64 +20,69 @@ try {
...
+	if (!message.ava) {
+		return;
 	}
+
+	process.emit(message.name, message.data);
 });
 
-process.on('ava-kill', function () {
+process.on('ava-exit', function () {
+	// use a little delay when running on AppVeyor (reason unknown)
```
sindresorhus: We know the reason, AppVeyor is just super bad and flaky :p

lib/fork.js
```diff
@@ -1,54 +1,43 @@
 'use strict';
 var childProcess = require('child_process');
 var Promise = require('bluebird');
-var path = require('path');
+var dirname = require('path').dirname;
+var join = require('path').join;
```
sindresorhus: Can you just import `path` instead and use methods on it?

vdemedes: yeah, did that before, but we need `path` here - https://github.com/sindresorhus/ava/pull/236/files#diff-a48173105195b7a6481a3ad85b45144eR13

When we will support #229, `babel` variable name won't be a good choice.

lib/babel.js
vdemedes: will update the comment ;)

lib/runner.js
vdemedes: Np ;) I just order stuff by module name length in my projects.

lib/fork.js
```diff
@@ -56,20 +45,36 @@ module.exports = function (args) {
...
 		});
 	});
 
 	// emit `test` and `stats` events
 	ps.on('message', function (event) {
+		if (!event.ava) {
+			return;
+		}
+
+		event.name = event.name.replace('ava-', '');
```
sindresorhus: `'ava-'` → `/^ava-/` just to be sure we don't strip any future message with `ava-` in the middle. I know, edge case, but this gives clearer intent too.

vdemedes: Ok!

lib/send.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+// utility to send messages to processes
+function send(ps, name, data) {
+	if (typeof ps === 'string') {
```
sindresorhus: Is this really needed? Do we really pass a custom `process` anywhere? And even if we do, wouldn't it be better to just have it as the third argument instead? For simplicity.

vdemedes: We pass messages in `lib/fork.js`. Yeah, I can move it to 3rd argument.

lib/fork.js
sindresorhus: Call it `filepath` or `pth`.

test/fixture/long-running.js
```diff
@@ -11,7 +11,11 @@ test('long running', function (t) {
 		while(Date.now() - start < 2000) {
 			//synchronously wait for 2 seconds
 		}
-		process.send({name:'cleanup-completed', data: {completed: true}});
+		process.send({
+			name:'cleanup-completed',
```
sindresorhus: needs a space after `:`

sindresorhus:
// @jamestalmage

lib/send.js
vdemedes: Turns out it will conflict with the following:

```js
send('teardown', ps);
```

where events don't have any data.

sindresorhus: Oh, ok, nvm then.

vdemedes:
@jamestalmage thanks for the tip!

jamestalmage:
Nice!

You taking a break now, or are you moving on to a major refactor?

vdemedes:
No no, I want to clean up everything, all files and tests, then move on to refactoring.

jamestalmage:
Leave `runner.js` to me for a while if you don't mind.

vdemedes:
Unfortunately, code interferes between files, so I can't guarantee I can leave unaffected runner.js. I was hoping to do it all in one wave, so that I don't wait for anyone else (that is why we cleaned up our PR queue) and after that, any suggestions and contributions are welcome ;)

jamestalmage:
OK

PR #239, opened 2015-11-18
Clean up runner

vdemedes:
Small cleanup of `lib/runner.js`

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -8,12 +8,12 @@ var send = require('./send');
 
 function noop() {}
 
-function each(items, fn) {
-	return Promise.all(items.map(fn));
+function each(items, fn, context) {
+	return Promise.all(items.map(fn, context));
 }
 
-function eachSeries(items, fn) {
-	return Promise.resolve(items).each(fn);
+function eachSeries(items, fn, context) {
+	return Promise.resolve(items).each(fn.bind(context));
 }
 
 function Runner(opts) {
@@ -27,6 +27,7 @@ function Runner(opts) {
 
 	this.stats = {
 		failCount: 0,
+		passCount: 0,
 		testCount: 0
 	};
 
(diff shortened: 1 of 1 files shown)


lib/runner.js
```diff
@@ -207,17 +210,21 @@ Runner.prototype.run = function () {
...
+			}
 		})
 		.then(function () {
-			return eachSeries(tests.after, self._runTest.bind(self));
+			return eachSeries(tests.after, self._runTest, self);
 		})
 		.catch(noop)
 		.then(function () {
-			stats.testCount = tests.only.length ? tests.only.length : stats.testCount;
+			stats.testCount = hasOnlyTests ? tests.only.length : stats.testCount;
```
BarryThePenguin: What about...
```js
stats.testCount = tests.only.length || stats.testCount;
```

vdemedes: Good one!

BarryThePenguin: It's the small things :sparkles:

vdemedes: Fully agree!

sindresorhus:
@jamestalmage Node.js 0.10 seems to be failing again on Windows :(

https://ci.appveyor.com/project/sindresorhus/ava/build/job/vrwhv5konnjgylcc

jamestalmage:
Yes, failure is different now. "Missing plan".

In TAP, the plan line can be emitted at the beginning or end. If it is the end, we already know forked processes often drop a line or two of IO from the end on Windows (especially Node 0.10). It could be something else entirely, but that is my initial suspicion.

sindresorhus:
@jamestalmage Yeah, my suspicion too. Might be relevant: https://github.com/nodejs/node/issues/3670

jamestalmage:
Yes. Looks very relevant. Perhaps we need to avoid explicitly ending the process unless we have to.

PR #240, opened 2015-11-18
Clean up lib/test.js

vdemedes:


lib/test.js
```diff
@@ -104,51 +112,52 @@ Test.prototype.plan = function (count) {
 };
 
 Test.prototype.run = function () {
-	this.promise = {};
+	var self = this;
+
+	this.promise = Promise.pending();
```
sindresorhus: `Promise.pending()`? Couldn't find this anywhere in the Bluebird docs: http://bluebirdjs.com/docs/api-reference.html

jamestalmage: https://github.com/petkaantonov/bluebird/wiki/Promise-anti-patterns

lib/test.js
```diff
@@ -104,51 +112,52 @@ Test.prototype.plan = function (count) {
...
 
 	// TODO(vdemedes): refactor this to avoid storing the promise
-	return new Promise(function (resolve, reject) {
-		this.promise.resolve = resolve;
-		this.promise.reject = reject;
+	if (!this.fn) {
+		return this.exit();
```
sindresorhus: I prefer not an empty `return` if it actually doesn't return anything. It's more explicit and has a better intent.

```js
this.exit();
return;
```

lib/test.js
```diff
@@ -104,51 +112,52 @@ Test.prototype.plan = function (count) {
...
+					actual: err,
+					message: 'Promise rejected → ' + err,
+					operator: 'promise'
+				});
+
+				self.exit();
+			});
+	}
+
+	return this.promise.promise;
```
sindresorhus: Why `promise.promise`?

lib/test.js
```diff
@@ -192,11 +203,11 @@ Test.prototype.exit = function () {
 		this.ended = true;
 
 		setImmediate(function () {
-			if (this.assertError) {
-				return this.promise.reject(this.assertError);
+			if (self.assertError) {
+				return self.promise.reject(self.assertError);
```
sindresorhus: Same as https://github.com/sindresorhus/ava/pull/240/files#r45325062

lib/test.js
BarryThePenguin: Would it be simpler to use `Promise.resolve` and `Promise.reject`?

lib/test.js
```diff
@@ -192,11 +203,11 @@ Test.prototype.exit = function () {
 		this.ended = true;
 
 		setImmediate(function () {
```
sindresorhus: Wonder if we should use https://github.com/sindresorhus/immediate-promise and have `.exit` return a promise. Related to: https://github.com/sindresorhus/ava/pull/240/files#r45325153

lib/test.js
vdemedes: @BarryThePenguin we need to return a promise by the end of function and resolve/reject it later, so we can't use these right away.

lib/test.js
vdemedes: Np ;)

lib/test.js
vdemedes: because `Promise.pending()` returns something called a `resolver`, it's not an actual promise, it's not thenable. To get a promise from resolver, we need to get it from `promise` property:

```js
var resolver = Promise.pending();
var promise = resolver.promise;

promise.then(...);
```

lib/test.js
vdemedes: The main reason for using `Promise.pending()` is to have less nesting/indentation in this function, that's it.

sindresorhus:
tumblr_ml9za9ecsb1qdlh1io1_400

PR #241, opened 2015-11-19
Try and let processes exit on their own instead of calling process.exit();

jamestalmage:
See discussion on #239

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,9 +1,9 @@
 environment:
   matrix:
+    - nodejs_version: '0.10'
     - nodejs_version: '5'
     - nodejs_version: '4'
     - nodejs_version: '0.12'
-    - nodejs_version: '0.10'
 install:
   - ps: Install-Product node $env:nodejs_version
   - set CI=true
--- a/lib/babel.js
+++ b/lib/babel.js
@@ -60,29 +60,19 @@ process.on('message', function (message) {
 });
 
 process.on('ava-exit', function () {
-	// use a little delay when running on AppVeyor (because it's shit)
-	var delay = process.env.AVA_APPVEYOR ? 100 : 0;
-
-	setTimeout(function () {
-		process.exit(0);
-	}, delay);
+	process.exit(0);
 });
 
 process.on('ava-teardown', function () {
 	var rejections = loudRejection.currentlyUnhandled();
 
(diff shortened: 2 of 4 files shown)


sindresorhus:
Still failing. I've had enough, let's just remove Node.js 0.10 from AppVeyor. It's already tested on Linux. ?

jamestalmage:
I begrudgingly agree. Very disappointed this is the route we have to take:

http://help.appveyor.com/discussions/problems/1352-output-of-commands-is-often-trimmed

https://github.com/mapnik/node-mapnik/issues/257#issuecomment-45571226

http://help.appveyor.com/discussions/problems/1257-console-output-suppressed

jamestalmage:
Here is the core issue: https://github.com/nodejs/node-v0.x-archive/pull/7196

The fix needs to be back-ported to the Node `0.10` branch. A number of people have plus oned the issue, without a lot of traction.

@sindresorhus Do you know anybody working on node that could push that along? This will impact our ability to provide TAP output on Node 0.10 / Windows.

If we can't get this fixed upstream, we should at least do our users the courtesy of officially NOT supporting it, and documenting the lack of support. Something along the lines of "We support all LTS versions of Node, with the exception of `0.10` on Windows." (and a link to the relevant issue).

sindresorhus:
> Do you know anybody working on node that could push that along?

@silverwind

> This will impact our ability to provide TAP output on Node 0.10 / Windows.

Yeah, let's just document it. I honestly couldn't care less about an ancient Node.js version on Windows. Most people test on Travis anyways. I think we've wasted more than enough time dealing with this.

sindresorhus:
There's https://github.com/cowboy/node-exit, but `process.on('exit')` won't work correctly when using it.

jamestalmage:
>  I honestly couldn't care less about an ancient Node.js version on Windows

I am perfectly fine leaving it at that.

> but `process.on('exit')` won't work correctly ... 

I don't think `process.on('exit')` is being used anywhere in our code, we are using `signal-exit` which does things a bit different (not sure how it would be affected without testing).

silverwind:
@jamestalmage you could just do a PR on node with that backport, targeting the `v0.10` branch. Maybe it'll even make it in time for the upcoming release, see https://github.com/nodejs/node/pull/2805. I don't think anyone would object merging it.

PR #243, opened 2015-11-20
Make runner methods chainable.

jamestalmage:
This flattens the entire `runner.tests` data structure into a single array.

Elements of the array are then chosen using `runner.select(selectorObject)`.

The selector currently just compares its own properties to each test, and selects those that have matching properties. It will be easy to expand with `groupMembership` queries, etc.

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -82,7 +82,7 @@ Object.keys(chainableFunctions).forEach(function (key) {
 Runner.prototype._addFn = function (opts, title, fn) {
 	var Constructor = (opts && /Each/.test(opts.type)) ? Hook : Test;
 	var test = new Constructor(title, fn);
-	objectAssign(test, opts);
+	test.metadata = objectAssign({}, opts);
 	this.tests.push(test);
 };
 
@@ -146,7 +146,7 @@ Runner.prototype._addTestResult = function (test) {
 		duration: test.duration,
 		title: test.title,
 		error: test.assertError,
-		type: test.type,
+		type: test.metadata.type,
 		skip: test.skip
 	};
 
@@ -196,7 +196,7 @@ Runner.prototype.run = function () {
 Runner.prototype.select = function (filter) {
 	function filterFn(test) {
 		return Object.keys(filter).every(function (key) {
-			return filter[key] === test[key];
+			return filter[key] === test.metadata[key];
 		});
 	}
 	return this.tests.filter(filterFn);
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -28,26 +28,24 @@ function Runner(opts) {
 	this.results = [];
 
 	this.tests = [];
+
+	var add = (function add(test) {
+		this.tests.push(test);
+	}).bind(this);
+
+	this.test = makeChain({
+		type: 'test',
+		serial: false,
+		exclusive: false,
+		skipped: false
+	}, add);
 }
 
 util.inherits(Runner, EventEmitter);
 module.exports = Runner;
 
-Runner.prototype._addTest = function (Constructor, title, fn, opts) {
-	var test = new Constructor(title, fn);
-
-	objectAssign(test, {
-		type: 'test',
-		serial: false,
-		exclusive: false,
-		skipped: false
-	}, opts);
-
-	this.tests.push(test);
-};
-
 var chainableFunctions = {
-	serial: {serial: 'true'},
+	serial: {serial: true},
 	before: {type: 'before'},
 	after: {type: 'after'},
 	skip: {skipped: true},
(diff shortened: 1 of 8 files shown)


lib/runner.js
```diff
@@ -5,6 +5,7 @@ var Promise = require('bluebird');
 var hasFlag = require('has-flag');
 var Test = require('./test');
 var send = require('./send');
+var objectAssign = require('object-assign');
```
jamestalmage: TODO: We probably should look in to replacing some of these polyfills with `core-js` equivalents. Hopefully cutting down on duplicate functionality.

jamestalmage:
I'm getting close to turning this into some cool chaining functionality.

`only.serial(...)`
`serial.only(...)`
`skip.serial(...)`
`afterEach.skip(...)`
...
etc. 


I need sleep now though. More tomorrow.

test/runner.js
```diff
@@ -14,63 +14,65 @@ test('returns new instance of runner without "new"', function (t) {
 test('runner.addTest adds a new test', function (t) {
 	var runner = new Runner();
 	runner.addTest(mockTitle, noop);
-	t.is(runner.stats.testCount, 1);
-	t.is(runner.tests.concurrent.length, 1);
-	t.true(runner.tests.concurrent[0] instanceof Test);
+	t.is(runner.tests.length, 1);
+	t.true(runner.tests[0] instanceof Test);
+	t.false(runner.tests[0].serial);
```
sindresorhus: Only use `t.true/false` when they're actually `true`/`false`. I know `tap` doesn't enforce this, but AVA does, so it will fail when we eventually move to self testing. Instead use `t.ok/notOk` to check truthy/falsy.

I just cleaned this up yesterday: https://github.com/sindresorhus/ava/commit/21885b8c7dd929a7496ea9105196dd2d67d02db5

// @vdemedes fyi

sindresorhus:
I like where this is going :)

Chaining is an important step as it will let you skip a test without losing the type of test. Right now you can't have a skipped test that is serial. You have to remove the `serial` part, which loses some important information.

jamestalmage:
OK - saw my error - chaining should now work.

This needs some cleanup (see the various TODO's), documentation, and more test coverage.

I should be able to wrap it up this weekend.

When it comes time to make nested groups, we will continue along the same pattern here, but replacing the base `_addTest` with one that merges the group settings as well (so you can skip an entire group, or set `serial` as the default for a group for example).

I also need to add the inverse of most of the boolean flags. (i.e. concurrent = {serial: false}). That would be used for making a few tests concurrent in a group of tests where the default is serial.

In all, I think doing it this way is going to open up some really cool stuff.

OK - bedtime for real now.

jamestalmage:
I am starting to think all these flags belong on a `metadata` object, i.e.:

```
test.skipped => test.metadata.skipped
test.exclusive => test.metadata.exclusive
```

Otherwise we need to always be careful the metadata properties never share a name with a useful property. (For example we can not use `test.skip` as the metadata property, because `test.skip()` is the method that sets that flag).

jamestalmage:
> Otherwise we need to always be careful the metadata properties never share a name with a useful property

Useful properties are going to be hard to predict as well, especially once we enable custom assertions.

lib/runner.js
```diff
@@ -207,24 +179,25 @@ Runner.prototype.run = function () {
...
 		});
 };
+
+Runner.prototype.select = function (filter) {
+	function filterFn(test) {
+		return Object.keys(filter).every(function (key) {
+			return filter[key] === test.metadata[key];
+		});
+	}
+	return this.tests.filter(filterFn);
```
sindresorhus: Just put the filter function inline.

test/runner.js
```diff
@@ -116,26 +174,71 @@ test('run serial tests before concurrent ones', function (t) {
...
+	runner.beforeEach.skip(pusher('beforeEach.skip'));
+
+	runner.test(pusher('concurrent'));
+	runner.test.skip(pusher('concurrent.skip'));
+
+	runner.serial(pusher('serial'));
+	runner.serial.skip(pusher('serial.skip'));
+
+	runner.run().then(function () {
+		// Note that afterEach and beforeEach run twice because their are two actual tests - "serial" and "concurrent"
```
sindresorhus: `their` => `there`

sindresorhus:
@jamestalmage Reviewed. This looks very good. Could you update the readme about the methods being chainable?

lib/runner.js
jamestalmage: Wish I could use arrow functions :disappointed:

vdemedes:
I don't really understand the point of `Hook` class, since it returns a `Test` instance in `test` method. @jamestalmage Could you please explain?

Otherwise, I really like the idea behind this PR.

jamestalmage:
It's for beforeEach / afterEach. Which need to be created multiple times. It's just storing the constructor args for later invoke (multiple times)

vdemedes:
@jamestalmage makes sense!

sindresorhus:
:+1: Landed. Superb work on this one @jamestalmage :)

tumblr_mqmk6j8kkz1qdlh1io1_250

PR #247, opened 2015-11-20
Document workaround for transpiling ES2015 module imports in tests

sparty02:
Adverstise workaround for transpiling ES6 module imports in tests

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -313,7 +313,7 @@ You can also use your own local Babel version:
 
 #### Transpiling Dependent Modules
 
-AVA currently only transpiles the tests that you ask it to run.  *It will not transpile modules that you ```import``` from outside of the test (that also may happen to be written in ES2015 and/or use the ES2015 module syntax).*  While there are valid reasons for taking this approach, it may not be what you expect!
+AVA currently only transpiles the tests that you ask it to run.  *It will not transpile modules that you ```import``` from outside of the test.*  While there are valid reasons for taking this approach, it may not be what you expect!
 
 As a simple workaround, you can use [Babel's require hook](https://babeljs.io/docs/usage/require/) in order to do on-the-fly transpiling of modules that are subsequently required.  Because AVA supports ES6 module syntax, you can use it to import the require hook itself:
 
--- a/readme.md
+++ b/readme.md
@@ -311,14 +311,16 @@ You can also use your own local Babel version:
 }
 ```
 
-*Note:  AVA currently only transpiles the tests that you ask it to run.  In other words, it won't transitively transpile modules that you ```import``` from outside of the test (that also may happen to be written in ES6 and/or use the ES6 module syntax).  While there are valid reasons for taking this approach (no impact/modification to the code under test), it may not be what you expect!*
+#### Transpiling Dependent Modules
+
+*Note:  AVA currently only transpiles the tests that you ask it to run.  In other words, it won't transitively transpile modules that you ```import``` from outside of the test (that also may happen to be written in ES2015 and/or use the ES2015 module syntax).  While there are valid reasons for taking this approach (no impact/modification to the code under test), it may not be what you expect!*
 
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -311,6 +311,23 @@ You can also use your own local Babel version:
 }
 ```
 
+*Note:  AVA currently only transpiles the tests that you ask it to run.  In other words, it won't transitively transpile modules that you ```import``` from outside of the test (that also may happen to be written in ES6 and/or use the ES6 module syntax).  While there are valid reasons for taking this approach (no impact/modification to the code under test), it may not be what you expect!*
+
+*As a simple workaround, you can use [Babel's require hook](https://babeljs.io/docs/usage/require/) in order to do on-the-fly transpiling of modules that are subsequently required.  Because AVA supports ES6 module syntax, you can use it to import the require hook itself:*
+
+```js
+import 'babel-core/register';
```
sindresorhus: Place this below 'ava'. No point in wasting time transpiling AVA.

readme.md
```diff
@@ -311,6 +311,23 @@ You can also use your own local Babel version:
 }
 ```
 
+*Note:  AVA currently only transpiles the tests that you ask it to run.  In other words, it won't transitively transpile modules that you ```import``` from outside of the test (that also may happen to be written in ES6 and/or use the ES6 module syntax).  While there are valid reasons for taking this approach (no impact/modification to the code under test), it may not be what you expect!*
```
sindresorhus: Can you put a subheading above here. `####`

sindresorhus:
Can you give the PR a proper title and use ES2015 over ES6?

readme.md
sparty02: Can you clarify?

sparty02:
Do you want me to squash these updates into a single commit?

readme.md
sindresorhus: If you move it down it will not have an effect on the below import which is AVA. AVA is pretty large, so it will definitely slow it down if it needs to be transpiled too.

sindresorhus:
@sparty02 Nah, I'll squash merge myself when it's done.

readme.md
```diff
@@ -311,6 +311,25 @@ You can also use your own local Babel version:
 }
 ```
 
+#### Transpiling Dependent Modules
+
+*Note:  AVA currently only transpiles the tests that you ask it to run.  In other words, it won't transitively transpile modules that you ```import``` from outside of the test (that also may happen to be written in ES2015 and/or use the ES2015 module syntax).  While there are valid reasons for taking this approach (no impact/modification to the code under test), it may not be what you expect!*
```
jamestalmage: A few nits:

- IMO, you are overusing *emphasis* here. Emphasizing whole paragraphs is overkill. Emphasize the one/two sentences I need to read to understand what the paragraph is about, so I can move on if it does not concern me.

- ... run. In other words, it won't   `=>` ... run. It will **not** (`In other words` is superfluous language)

- `transitively transpile` - let's think of a simpler way to say that 

> Note: *AVA currently only transpiles the tests that you ask it to run.* It will **not** ...

sparty02:
@jamestalmage  Good feedback, I've addressed this in the latest updates.

jamestalmage:
:+1:

BarryThePenguin:
Trying to make it more concise. Use what you like:

AVA currently only transpiles the tests that you ask it to run. *AVA does not transpile modules that you `import` from outside a test.* While there are valid reasons for taking this approach, it may not be what you expect!

As a workaround, you can use Babel's require hook to transpile required modules on-the-fly. Because AVA supports ES2015 module syntax, you can use it to import the require hook itself:

jamestalmage:
@sparty02 - I think we are ready to merge this once you incorporate @BarryThePenguin's advice.

sparty02:
@jamestalmage done!

sindresorhus:
Thank you @sparty02 and @BarryThePenguin :)

PR #251, opened 2015-11-21
context should not exist on before/after tests.

jamestalmage:
IMO, since context is not shared in `before` and `after` hooks.
It should be an error to access it in either scenario.

(cherry picked from commit f298d72 on #243)

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -26,7 +26,6 @@ function Test(title, fn) {
 	this.assertCount = 0;
 	this.planCount = null;
 	this.duration = null;
-	this.context = {};
 
 	// test type, can be: test, hook, eachHook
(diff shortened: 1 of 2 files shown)


jamestalmage:
This becomes easier to solve with a helpful Error message with the attached metadata from #243 . 

```js
function checkContextAccess(test) {
    if (test.metadata.type === 'before' || test.metadata.type === 'after') {
      throw new Error('test.context access is not allowed in before/after hooks');
    }
}

Object.define(Test.prototype, 'context', {
  get: function () {
    checkContextAccess(this);
    return this._context;
  }, 
  set: function (newContext) {
    checkContextAccess(this);
    this._context = context;
  }
});

PR #253, opened 2015-11-22
Breaking: `t.end` requires `test.async`, and `t.plan` will not auto end tests. Fixes #244.

### LINKED ISSUE #244: Proposal to make test() behavior more explicit and optimize for the future
Continued from https://github.com/sindresorhus/ava/issues/133.

-

I've been thinking a lot about sync vs. async tests and I really would like to make it more explicit and optimize for the future. Currently when using `test()` it's ambiguous whether the test is sync or async. The reality is also that most tests are sync and it's annoying and easy to forget the `t.end()` in every test (even I forget it all the time).

I propose we change the interface:

#### Before

- `test(t => {})` - async/sync depending of whether you do any async operation
- `test(async t => {})` - promise async

#### After

- `test(t => {})` - sync (no longer required to use `t.end()`)
- `test(async t => {})` - promise async (what you really should be using)
- `test.async(t => {})` - callback async (legacy)

The main change here is that `test()` goes from being async to being sync with async support through async functions.

I know it's a breaking change, but isn't now the time to do that?

This will make AVA more future friendly as we'll eventually just use async functions for async operations. We'll support `test.async()` for legacy, but have a nice main interface for the main use-cases.

-

// @vdemedes @jamestalmage @kevva @arthurvr @SamVerschueren @af @taylorhakes and anyone else with opinions.

jamestalmage:
See #244

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -6,6 +6,7 @@ var fnName = require('fn-name');
 var co = require('co-with-promise');
 var observableToPromise = require('observable-to-promise');
 var isPromise = require('is-promise');
+var isObservable = require('is-observable');
 var assert = require('./assert');
 var globals = require('./globals');
 
@@ -30,7 +31,6 @@ function Test(title, fn) {
 
 	// TODO(jamestalmage): make this an optional constructor arg instead of having Runner set it after the fact.
 	// metadata should just always exist, otherwise it requires a bunch of ugly checks all over the place.
-	// I will fix this correctly in the upstream PR.
 	this.metadata = {};
 
 	// store the time point before test execution
@@ -51,12 +51,6 @@ module.exports = Test;
 
 Test.prototype._assert = function () {
 	this.assertCount++;
-
-	// TODO(jamestalmage): no only auto-ends in "declared" async mode `test.async(...)`. Should we drop auto-end behavior all together?
-	// https://github.com/sindresorhus/ava/issues/244#issuecomment-158705839
-	if (this.metadata.async && this.assertCount === this.planCount) {
-		globals.setImmediate(this.exit.bind(this));
-	}
 };
 
 // patch assert methods to increase assert count and store errors
@@ -139,24 +133,33 @@ Test.prototype.run = function () {
 		this.exit();
 	}
 
-	ret = observableToPromise(ret);
+	var asyncType = 'promises';
+
+	if (isObservable(ret)) {
+		asyncType = 'observables';
+		ret = observableToPromise(ret);
+	}
 
 	if (isPromise(ret)) {
-		if (!this.metadata.async) {
-			ret = ret.then(function () {
-				self.exit();
-			});
+		if (this.metadata.async) {
+			self._setAssertError(new Error('Do not return ' + asyncType + ' from tests declared via `test.async(...)`, if you want to return a promise simply declare the test via `test(...)`'));
+			this.exit();
+			return this.promise.promise;
 		}
 
-		ret.catch(function (err) {
-			self._setAssertError(new assert.AssertionError({
-				actual: err,
-				message: 'Promise rejected → ' + err,
-				operator: 'promise'
-			}));
+		ret
+			.then(function () {
+				self.exit();
+			})
+			.catch(function (err) {
+				self._setAssertError(new assert.AssertionError({
+					actual: err,
+					message: 'Promise rejected → ' + err,
+					operator: 'promise'
+				}));
 
-			self.exit();
-		});
+				self.exit();
+			});
 	} else if (!this.metadata.async) {
 		this.exit();
 	}
--- a/lib/test.js
+++ b/lib/test.js
@@ -28,6 +28,11 @@ function Test(title, fn) {
 	this.duration = null;
 	this.assertError = undefined;
 
+	// TODO(jamestalmage): make this an optional constructor arg instead of having Runner set it after the fact.
+	// metadata should just always exist, otherwise it requires a bunch of ugly checks all over the place.
+	// I will fix this correctly in the upstream PR.
+	this.metadata = {};
+
 	// store the time point before test execution
 	// to calculate the total time spent in test
 	this._timeStart = null;
@@ -47,7 +52,9 @@ module.exports = Test;
 Test.prototype._assert = function () {
 	this.assertCount++;
 
-	if (this.assertCount === this.planCount) {
+	// TODO(jamestalmage): no only auto-ends in "declared" async mode `test.async(...)`. Should we drop auto-end behavior all together?
+	// https://github.com/sindresorhus/ava/issues/244#issuecomment-158705839
+	if (this.metadata.async && this.assertCount === this.planCount) {
 		globals.setImmediate(this.exit.bind(this));
 	}
 };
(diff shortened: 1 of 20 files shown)


jamestalmage:
Overall, I like it. 

There are a few places where it creates some weirdness. All resolve around our async assertions (`t.throws(rejectedPromise)`, `t.throws(rejectedObservable)`).

```js
test('this does not work', t => {
  var promise = Promise.reject(new Error('foo'));
  
  t.throws(promise);
  // we do not return a promise, so AVA assumes this is a synchronous function
});
```

There are a number of easy ways around this this. The issue will be preventing user confusion.

(Ordered from best to worst workaround IMHO)

```js
test('use async/await', async t => {
  var promise = Promise.reject(new Error('foo'));
  
  await t.throws(promise); 
});
```

```js
test('return the assertion promise', t => {
  var promise = Promise.reject(new Error('foo'));

  // t.throws returns a promise, so return it from our test function.
  return t.throws(promise);
});
```

```js
test.async('legacy async with a plan', t => {
  t.plan(1);
  var promise = Promise.reject(new Error('foo'));

  // test is auto-ended once this resolves and we reach our plan count.
  t.throws(promise);
});
```

I like this last one the least. As [previously stated](https://github.com/sindresorhus/ava/issues/244#issuecomment-158705839), I think allowing `t.plan()` to auto-end tests is a bad choice that causes confusing false positives. Since we are biting the bullet with a major breaking change here, I would love to see us correct that as well.

sindresorhus:
> (Ordered from best to worst workaround IMHO)

#1 is how I already use it and how it was intended. We should just document it better.

sindresorhus:
I think we have a resolution in https://github.com/sindresorhus/ava/issues/244, unless there's anything else that's still undecided? Could you rebase this PR?

jamestalmage:
This is ready for review.

// @sindresorhus @vdemedes 

I have updated the docs to reflect the new behavior, with a few language tweaks along the way.

readme.md
```diff
@@ -385,31 +432,34 @@ test(async t => {
...
-		return n * n;
-	});
+	t.plan(3);
+	return Observable.of(1, 2, 3, 4, 5, 6)
+		.filter(n => {
+			// only even numbers
+			return n % 2 === 0;
+		})
+		.map(n => {
+			t.pass();
```
jamestalmage: The previous Observable example was so trivial that I don't feel it adequately explained the advantage AVA offered.
I think this:

 1. Is a simple enough usage of observable that you don't necessarily need to be familiar with them to understand what is happening.
 2. Shows a novel use of `t.plan()` in coordination with the Observable filter. It's a little closer to a test someone might actually right.

@BarryThePenguin 
What I know about Observables I learned in the last 20 minutes, so feel free to tell me where I missed the mark.

BarryThePenguin: You could change map with `.do(() => t.pass());`. Otherwise, looks good.

RxJS has some [docs](https://github.com/ReactiveX/RxJS/blob/master/doc/writing-marble-tests.md) on marble test. But I'm not sure how they would fit with AvA

jamestalmage: Good call! Thanks!

lib/test.js
```diff
@@ -135,27 +135,38 @@ Test.prototype.run = function () {
...
+Object.defineProperty(Test.prototype, 'end', {
+	get: function () {
+		if (this.metadata.async) {
+			return this._end;
+		}
+		throw new Error('t.end is not supported in this context. To use t.end as a callback, you must explicitly declare the test asynchronous via `test.async(testName, fn)` ');
+	}
+});
+
+Test.prototype._end = function (err) {
```
sindresorhus: Why not just this here instead?

```js
if (!this.metadata.async) {
	throw new Error('t.end is not supported in this context. To use t.end as a callback, you must explicitly declare the test asynchronous via `test.async(testName, fn)` ');
}
```

readme.md
```diff
@@ -70,15 +70,15 @@ import test from 'ava';
 
 test('foo', t => {
 	t.pass();
-	t.end();
 });
 
-test('bar', t => {
+test.async('bar', t => {
```
sindresorhus: We should rather use async functions in this example.

lib/test.js
```diff
@@ -135,27 +135,38 @@ Test.prototype.run = function () {
...
-					operator: 'promise'
-				}));
-
+		if (!this.metadata.async) {
+			ret = ret.then(function () {
 				self.exit();
 			});
+		}
+
+		ret.catch(function (err) {
```
sindresorhus: Any reason to allow promises in `test.async`? Wouldn't it be better to contain it to `test()`?

readme.md
```diff
@@ -152,13 +155,13 @@ You can also choose to use a named function instead:
...
 test(function name(t) {
-	t.end();
+	t.pass();
 });
 ```
 
 ### Planned assertions
 
-Planned assertions are useful for being able to assert that all async actions happened. It also comes with the benefit of not having to manually end the test.
+Planned assertions are useful for being able to assert that all async actions happened.
```
sindresorhus: Can you add something about that it's also useful to make sure not too many assertions were run?

readme.md
```diff
@@ -166,19 +169,62 @@ This will result in a passed test:
...
+
+test.async(t => {
 	setTimeout(() => {
 		t.pass();
+		t.end();
 	}, 100);
 });
 ```
 
+*WARNING: Recent breaking change.*
```
sindresorhus: Make this a subheading

readme.md
```diff
@@ -385,31 +432,32 @@ test(async t => {
 });
 ```
 
-*You don't have to manually call `t.end()`.*
-
 ### Observable support
 
 AVA comes with builtin support for [observables](https://github.com/zenparsing/es-observable).
+If you return an observable from a test, AVA will automatically consume it to completion before ending the test.
+*You don't have to use legacy `test.async` mode or `t.end()`.*
```
sindresorhus: Can't we just prevent it? Throw an error if it's used with legacy mode?

sindresorhus:
Looks good to me, except for a few inline comments :)

@vdemedes ?

vdemedes:
✅from me.

lib/test.js
sindresorhus: // @vdemedes

vdemedes: well promises are async, that's why `test.async` allows them, right?

sindresorhus: Yes, but I see no reason you wouldn't rather use 'test()' with an async function in that case. Or even 'test' with a normal function and return a promise. I want to enforce best practices whenever possible.

Unless I'm missing something?

lib/test.js
jamestalmage: You mean just have `t.end` throw when executed?

Because we want to fail fast, and if passed as a callback:

```js
   fs.readFile(path, t.end);
```

We are waiting for the function to actually be called instead of throwing ASAP.

lib/test.js
jamestalmage: So you are thinking?

```js
if (isPromise(ret)) {
  if (this.metadata.async) {
    this._setAssertError(new Error('Do not return promises from....'));
  }
}

readme.md
jamestalmage: Yes - I think that is what you discussed above.

jamestalmage:
Today is a major holiday in the US, so I won't be wrapping this up untill tomorrow.

lib/test.js
sindresorhus: Yes

And error message should tell the user to use "test()" instead.

lib/test.js
sindresorhus: Does it matter? It shouldn't be used with `test()` anyways. So I don't see how it makes a difference.

jamestalmage: If they don't return a promise - we assume sync. We might mark the test done before the callback returns.

sindresorhus: Yes, they shouldn't use callback with `test()`, so I still don't see how it matter. I'm not totally focused and on my phone, so I might very well be confused though.

sindresorhus: Ok, never mind. Got it now. Sorry about the noise.

sindresorhus:
@jamestalmage No worries at all. Enjoy Thanksgiving :D

readme.md
sindresorhus: Yup

jamestalmage: FYI - I did make the change so we throw if they return a promise / observable. However, I don't think the wording here needs to be changed. It is not an error to use Observables in `test.async` tests, it's just an error to **return** an Observable.

We could add more language here fully explaining that, but I'm not sure it is necessary. The semantics of returning an async object are now well defined (the test ends when the async object ends). It doesn't make sense to combine that behavior with `t.end()`. Telling them they can't do something that doesn't make sense seems to be more noise than helpful.

sindresorhus: Ok

jamestalmage:
Ahhh, Windows!

@sindresorhus - I'll start looking into it here, but can you just retrigger the build? I was watching that one, and it was running REALLY slow. I think something was wrong on that VM.

sindresorhus:
Done

jamestalmage:
OK, CI passed.

Are we good here? I think all concerns are addressed.

sindresorhus:
:+1: Excellent. I'm super excited about this!

tumblr_mkwooez15j1qdlh1io1_400

Issue #250, opened 2015-11-21
PowerAssert shows incorrect line


Qix-:
I assume it's a mapping issue since we're transpiling things with babel.

<img width="138" alt="screen shot 2015-11-21 at 06 28 14" src="https://cloud.githubusercontent.com/assets/885648/11318672/12a429e4-9019-11e5-85ce-9bb75c0988ac.png">
<img width="433" alt="screen shot 2015-11-21 at 06 28 28" src="https://cloud.githubusercontent.com/assets/885648/11318675/1a841250-9019-11e5-8e53-0bc2b6297a66.png">

As you can see the source file in question only goes to line 120, but the error message shows 299.

jamestalmage:
Duplicate of #223.

PR for Fix: #254

Qix-:
:+1: couldn't find it in the issue search for some reason. Thanks.

PR #254, opened 2015-11-22
Add source map support. Fixes: #223, #250

### LINKED ISSUE #223: Add source-map-support
<https://www.npmjs.com/package/source-map-support> could be added to `babel.js`. Configured properly this would work with any source maps in the files under test, as well as the transformed test file.

jamestalmage:
Fixes #223, #250 

This adds basic source map support. Since transpilation of dependencies
is not merged yet, the `sourceMapCache` object and lookup is a little
premature (since we will only ever have one file that we can apply
apply source maps to).

### DIFF
--- a/lib/babel.js
+++ b/lib/babel.js
@@ -1,4 +1,18 @@
 'use strict';
+var sourceMapCache = Object.create(null);
+
+require('source-map-support').install({
+	retrieveSourceMap: function (source) {
+		if (sourceMapCache[source]) {
+			return {
+				url: source,
+				map: sourceMapCache[source]
+			};
+		}
+		return null;
+	}
+});
+
 var createEspowerPlugin = require('babel-plugin-espower/create');
 var requireFromString = require('require-from-string');
 var loudRejection = require('loud-rejection/api')(process);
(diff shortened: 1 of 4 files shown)


jamestalmage:
Because we use `require-from-string`, our source maps are never written to disk. This requires the implementation of the `retrieveSourceMap` callback. Final solutions for #189 and #177 will need to expand on this as well.

I do not know how expensive it is to load up `source-map-support` in each forked process like this.

Since all errors are serialized and transmitted over IPC before logging, It may be possible to defer the stack-trace transform until it has been transmitted back to the main thread.

jamestalmage:
This needs #255 before it will pass on Windows. The Travis failure looks like a fluke.

sindresorhus:
Looks good :+1:

sindresorhus:
> I do not know how expensive it is to load up source-map-support in each forked process like this.

> Since all errors are serialized and transmitted over IPC before logging, It may be possible to defer the stack-trace transform until it has been transmitted back to the main thread.

Let's do some benchmarking before we start investigating this. Might not make a difference.

PR #258, opened 2015-11-22
Add ability to skip individual assertions. Fixes #9.

### LINKED ISSUE #9: Test modifiers
`this._skip` is still `false` after running `.skip()`. It needs `.bind(this)` to work.

```js
var i = 0;

ava(function (a) {
	a.skip(); // a.skip.bind(this) works
	i++;
	a.end();
}).run(function () {
	t.is(i, 0);
	t.end();
});
```

This fails with `i` being `1`.

jamestalmage:
Fixes #9

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -202,7 +202,7 @@ Test.prototype.exit = function () {
 
 Test.prototype._publicApi = function () {
 	var self = this;
-	var api = {};
+	var api = {skip: {}};
 
 	// Getters
 	[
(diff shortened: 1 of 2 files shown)


sindresorhus:
I don't think we ever made a decision about `t.skip()` in #9. I guess it's about time we do. I was initially against it, but I find myself often having to comment out single test assertions when testing and manually updating the `t.plan` count. So I'm leaning towards :+1:. @vdemedes ?

vdemedes:
@sindresorhus we discussed this a few weeks ago and agreed to support only test-wide skip(), not per assertion.

Although, I see that implementation does not add much code, so I guess we can merge it in.

sindresorhus:
@vdemedes Ah, forgot, but seeing how simple the implementation is and the fact I've had a need for it recently, I'm :+1:.

@jamestalmage Can you rebase?

sindresorhus:
Totally slipped my mind when :+1:, but this needs some docs.

PR #259, opened 2015-11-23
print the first assertion failure of each test, instead of the last. Fixes #121, #220

### LINKED ISSUE #121: Output every failed assertion or just the first one (currently outputs just the last one)

### LINKED ISSUE #220: Stop on first failed assertion
I noticed that test cases are not stopped when an assertion fails. To be exact, they are stopped asynchronously. Here is an example:

```
test(t => {
  let arr = [ { name: 'abc' } ];
  t.is(arr.length, 2);
  t.is(arr[1].name, 'xyz');
  t.end();
});
```

The error I'd expect to see is "1 !== 2" while the actual error I get is "cannot read property 'name' of undefined" because execution is not stopped on the failed assertion (on the line above).

Is there a reason why `Test#exit()` is called asynchronously (by `Test#_assert()`) rather than synchronously (which is the case for errors thrown outside assertions)?

jamestalmage:
Fixes #121.
Fixes #220.

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -83,12 +83,15 @@ Object.keys(assert).forEach(function (el) {
 });
 
 Test.prototype._setAssertError = function (err) {
-	if (this.assertError === undefined) {
-		if (err === undefined) {
-			err = 'undefined';
-		}
-		this.assertError = err;
+	if (this.assertError !== undefined) {
+		return;
 	}
+
+	if (err === undefined) {
+		err = 'undefined';
+	}
+
+	this.assertError = err;
 };
 
 // Workaround for power-assert
--- a/lib/test.js
+++ b/lib/test.js
@@ -26,6 +26,7 @@ function Test(title, fn) {
 	this.assertCount = 0;
 	this.planCount = null;
 	this.duration = null;
+	this.assertError = undefined;
 
 	// test type, can be: test, hook, eachHook
 	this.type = 'test';
@@ -67,19 +68,28 @@ Object.keys(assert).forEach(function (el) {
 			if (isPromise(fn)) {
 				return Promise.resolve(fn)
 					.catch(function (err) {
-						self.assertError = err;
+						self._setAssertError(err);
 					})
 					.finally(function () {
 						self._assert();
 					});
 			}
 		} catch (err) {
-			this.assertError = err;
+			this._setAssertError(err);
 		}
 
 		this._assert();
 	};
 });
+
+Test.prototype._setAssertError = function (err) {
+	if (this.assertError === undefined) {
+		if (err === undefined) {
+			err = 'undefined';
+		}
+		this.assertError = err;
+	}
+};
 
 // Workaround for power-assert
 // `t` must be capturable for decorated assert output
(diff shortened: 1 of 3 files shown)


test/test.js
```diff
@@ -312,3 +312,33 @@ test('wait for test to end', function (t) {
 		avaTest.pass();
 	}, 1234);
 });
+
+test('promise is rejected with the first assertError', function (t) {
```
sindresorhus: Not sure why `promise` is mentioned here. It's not really relevant. Should be more like:

> fails with the first assertError

Sem with the below.

lib/test.js
```diff
@@ -68,20 +69,29 @@ Object.keys(assert).forEach(function (el) {
...
 	};
 });
 
+Test.prototype._setAssertError = function (err) {
+	if (this.assertError === undefined) {
+		if (err === undefined) {
+			err = 'undefined';
+		}
+		this.assertError = err;
+	}
```
sindresorhus: I usually prefer early returns over nesting when possible:

```js
if (this.assertError !== undefined) {
	return;
}

if (err === undefined) {
	err = 'undefined';
}

this.assertError = err;
```

sindresorhus:
@jamestalmage Looks good except for some minor nitpick :) Would you mind adding the assertion failure behavior to the docs so it's explicit how it works?

When you reference tickets in the the title, can you do so in the description too? It makes it easier to click through. The issue references in the title aren't clickable.

@vdemedes ?

jamestalmage:
> When you reference tickets in the the title, can you do so in the description too? It makes it easier to click through. The issue references in the title aren't clickable.

Errr...

> James Talmage commented 6 hours ago ...

sindresorhus:
> Errr...

Ah yes. I should have mentioned. I added the issue references to the PR description for you here.

jamestalmage:
OK, I just followed up on a number of other PR's

sindresorhus:
Can you fix the merge conflict?

jamestalmage:
Working on that and the documentation.

test/test.js
```diff
@@ -312,3 +312,33 @@ test('wait for test to end', function (t) {
...
+		a.is(1, 2);
+		a.is(3, 4);
+	}).run().catch(function (err) {
+		t.is(err.actual, 1);
+		t.is(err.expected, 2);
+		t.end();
+	});
+});
+
+test('fails with thrown falsie value', function (t) {
```
sindresorhus: super minor, but for consistency `falsie` → `falsy`

sindresorhus:
@jamestalmage You are so productive! :D

jamestalmage:
OK, I think that covers it.

PR #260, opened 2015-11-23
Ignore possible test files under node_modules folder. Fixes #226.

### LINKED ISSUE #226: Automatically ignore test files in node_modules folder
See #224 

The glob string was picking up a test file that was deployed with a dependency. (Unclear if they were using the default glob string or not - that should be looked into as well).

Should we automatically ignore `basedir/node_modules`? I can't see why a user would ever want that.

sotojuan:
As noted in #226, this pull request does two things:

* Add `!node_modules` to the array [https://github.com/sindresorhus/ava/blob/master/cli.js#L203](here), so that it ignores the folder when ava is ran with no arguments.

* Filter out any matched file that has the string `node_modules` in it. This works in case the user does something like `ava **`, which is rare but can happen.

Perhaps there's a more efficient way to do this, but I have tested this in various cases and it works, though I haven't been able to write a proper test for it.

Thanks and let me know if there's anything I can improve on.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -201,11 +201,17 @@ function init(files) {
 function handlePaths(files) {
 	if (files.length === 0) {
 		files = [
+			'!node_modules',
 			'test.js',
 			'test-*.js',
 			'test/*.js'
 		];
 	}
+
+	// ignore all possible test files in node_modules folder
+	files = files.filter(function (file) {
+		return (file.indexOf('node_modules') === -1);
+	});
 
 	// convert pinkie-promise to Bluebird promise
 	files = Promise.resolve(globby(files));
--- a/cli.js
+++ b/cli.js
@@ -201,12 +201,13 @@ function init(files) {
 function handlePaths(files) {
 	if (files.length === 0) {
 		files = [
-			'!node_modules',
 			'test.js',
 			'test-*.js',
 			'test/*.js'
 		];
 	}
+
+	files.push('!**/node_modules/**');
 
 	// convert pinkie-promise to Bluebird promise
 	files = Promise.resolve(globby(files));
@@ -221,7 +222,7 @@ function handlePaths(files) {
 		})
 		.then(flatten)
 		.filter(function (file) {
-			return file.indexOf('node_modules') === -1 && path.extname(file) === '.js' && path.basename(file)[0] !== '_';
+			return path.extname(file) === '.js' && path.basename(file)[0] !== '_';
 		});
 }
 
--- a/cli.js
+++ b/cli.js
@@ -208,11 +208,6 @@ function handlePaths(files) {
 		];
 	}
 
-	// ignore all possible test files in node_modules folder
-	files = files.filter(function (file) {
-		return (file.indexOf('node_modules') === -1);
-	});
-
 	// convert pinkie-promise to Bluebird promise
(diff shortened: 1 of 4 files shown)


cli.js
```diff
@@ -201,12 +201,18 @@ function init(files) {
...
 			'test.js',
 			'test-*.js',
 			'test/*.js'
 		];
 	}
 
+	// ignore all possible test files in node_modules folder
+	files = files.filter(function (file) {
+		return (file.indexOf('node_modules') === -1);
+	});
```
jamestalmage: This is code is actually before matched files are found.

At this point, `files` is not an actually a list of files, but a list of glob strings (like `"**/*-test.js`).

You may be seeing results on your machine here because your operating system is automatically expanding glob strings for you (I believe that is the case on OSX, but not on Windows).

cli.js
```diff
@@ -201,12 +201,18 @@ function init(files) {
...
+
 	// convert pinkie-promise to Bluebird promise
 	files = Promise.resolve(globby(files));
```
jamestalmage: This line is where we trade the incoming glob strings for a Promise of all the files found by `globby` using said glob strings.

Your filter above should be moved down into the `then` method.

jamestalmage:
You are on the right track. You need to expand on this using my comments from above.

You should also try to create a test or two. This will likely involve adding some files to the `test\fixture` folder. Look at some of the tests in `test/cli.js` for inspiration, and get creative from there.

A good way to verify you have written an effective test is to get rid of your changes to the production code (keep your tests in place). Your new tests should fail without your changes (otherwise what was the point?!), then add back in your changes and verify your changes to the production code take you from failing tests to passing tests.

Done this way, everybody can understand the need for the new code you propose, and your tests become a way of documenting the importance of your PR.

cli.js
```diff
@@ -201,6 +201,7 @@ function init(files) {
 function handlePaths(files) {
 	if (files.length === 0) {
 		files = [
+			'!node_modules',
```
sindresorhus: I think we can use `'!**/node_modules',` to match `node_modules` at any level and not having to do the filtering below.

sotojuan:
Wouldn't that only work if the user runs `ava` with no arguments? When you specify files like in my test or have some sort of custom match (which was the source of this issue), it fails (just tested it).

sindresorhus:
@sotojuan With how this is currently implemented, yes, but it would work if you instead just always pushed it as the last item to the `files` array before passing it to `globby`.

sotojuan:
Thanks, `!**/node_modules/**` seems to work best if pushed before passing the files to `globby`.

sindresorhus:
@sotojuan That's what I said ;)

sotojuan:
Ha, I meant adding the extra `/**` at the end made it worked!

sindresorhus:
Right, yeah.

jamestalmage:
:+1:

sindresorhus:
Landed. Very nice work @sotojuan ;)

687474703a2f2f7777772e7265616374696f6e676966732e636f6d2f77702d636f6e74656e742f67616c6c6572792f64616e63652d70617274792f74756d626c725f6d306f73727268714d6b31717a6376376e6f345f3235302e676966

PR #262, opened 2015-11-23
Support source map pragmas for stack traces

novemberborn:
This is a follow-up to #254. `source-map-support` can also resolve source maps from pragmas in the source file (or indeed embedded source maps). Now if the files under test are compiled separately but include the pragma, and the source maps are on disk, stack traces will be accurate. Same goes for third-party dependencies that include their source maps.

First commit makes the test explicit, the second commit adds another test for a file that includes a source map pragma. Without the modifications to `lib/babel` this test fails, but with the modifications it passes.


sindresorhus:
Looks good to me.

jamestalmage:
Hey @novemberborn,

Can we also make sure we are covering the case where the test file itself may have a source map?

This will require making Babel consume incoming source maps. Looking at [the Babel 5 options](https://github.com/babel/babel.github.io/blob/5.0.0/docs/usage/options.md), it seems there is an `inputSourceMap` option. So it is probably just a matter of retrieving it via `sourceMapSupport.retrieveSourceMap(...)`

sindresorhus:
Looks good. @novemberborn Would be cool if you could do a followup pull request with :arrow_up: :)

novemberborn:
@jamestalmage 

> Can we also make sure we are covering the case where the test file itself may have a source map?

To clarify, do you mean a case where the test file itself is the result of a compilation step, and then AVA's Babel hook transpiles it again? So we need to provide the input source map, if any, making sure the output source map can map all the way back to the original source?

jamestalmage:
yep. It's probably a niche case, but it just makes our source-map support complete. If we ever implement alternate compilers, it will be good to have, since we will likely always apply the power-assert transforms.

PR #264, opened 2015-11-23
Fix coverage

jamestalmage:
Due to NYC defaults, we were missing coverage on lib/test.js

I also added `npm run coverage` as a build target, which is for use
on developer machines. It launches a browser window with coverage
from the last run.

We no longer need NYC ~~or coveralls~~ as dev dependencies. `tap` handles
it all for us magically.

`maintaining.md` updated to reflect all these changes.

Also - I dropped coverage from the Windows tests. We don't need to make AppVeyor any slower.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules
 .nyc_output
+coverage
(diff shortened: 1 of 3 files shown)


sindresorhus:
:+1:

Issue #263, opened 2015-11-23
regression: array.includes no longer working.


jamestalmage:
Reference: https://github.com/sindresorhus/serialize-error/pull/3#discussion_r45643804

The following no longer works:

```js
test(t => {
  var keys = Object.keys(obj);
  var hasName = keys.includes('name');
  t.true(hasName);
  t.end();
});
```

I get the following error:

>  TypeError: keys.includes is not a function

We are using the `runtime` plugin, to avoid any polyfills, so it makes sense that Array.prototype.includes does not actually exist. I would have thought the runtime plugin would have picked this up and provided it via some helper function.

We are still on Babel 5, so I am punting until we land Babel 6 support before digging in.

// @sebmck @thejameskyle

jamestalmage:
So it looks like the last time this passed, was on `ava@0.3.0`. 
[At that point](https://github.com/sindresorhus/ava/blob/v0.3.0/lib/babel.js) we were using `babel-core/register`, which loads polyfills.

So maybe the `runtime` plugin has never handled this.

thejameskyle:
>  I would have thought the runtime plugin would have picked this up and provided it via some helper function.

Babel can't resolve instance methods on references to polyfills.

Think about it this way, how can a compiler **know** what `keys` is in this case? Until it knows that it can't possibly know what `includes` is referring to.

```js
keys.includes()
```

There is the potential in this particular case to do something about it because we can infer that the result of `Object.keys` is an array. However, if you had a case like the following we couldn't possibly get it working.

```js
function includesName(keys) {
  return keys.includes('name');
}
```

This inability to do things consistently is why Babel will never auto-polyfill like this.

jamestalmage:
Couldn't you transform it to something like this:

```js
var x = arr.includes('foo');

// =>

var x = includesHelper(arr, 'name');

function includesHelper(maybeArray, ...args) {
  if (Array.isArray(maybeArray) && !maybeArray.includes) {
    // use polyfill
  } else {
   return maybeArray.includes.apply(maybeArray, args);
  }
}
``` 

Obviously there is a performance hit for this way. Is that the only reason it is considered unacceptable, or are there other reasons?

thejameskyle:
If you're going to include the code to polyfill it, why not just polyfill it?

Also, I don't think we'd ever be interested in going down that route in core, you're welcome to try your hand at in in your own plugin.

jamestalmage:
[...]

PR #266, opened 2015-11-24
WIP: add test suite for es2015 polyfills

jamestalmage:
Based on the conversation in #263, I'm starting to work up a test suite that 
actually excercises ALL the ES2015 features, so we know what we are missing.

This is a rough start. If someone wants to help divide and conquer, just ping me on 
Gitter to discuss.

### DIFF
--- a/test/fixture/es2015-polyfills.js
+++ b/test/fixture/es2015-polyfills.js
@@ -1,0 +1,53 @@
+import test from '../../';
+
+test('Array.of support', t => {
+	const arr10 = Array.of(10);
+	t.is(arr10.length, 1);
+	t.is(arr10[0], 10);
+	t.end();
+});
+
+test('Array.from support', t => {
+	const arr = Array.from({
+		length: 2,
+		0: 0,
+		1: 1
+	});
+
+	t.true(Array.isArray(arr));
+	t.is(arr.length, 2);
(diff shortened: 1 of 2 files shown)


vdemedes:
Oooh, I am not sure about this one. Our test suite will grow tremendously, if we test every ES6 feature. That's definitely up to babel project to test these. We just need to fix our babel issues and get it right.

sindresorhus:
Good idea! But I think this would be better as a separate project. Something like a ES2015 tester project that both we and Babel could consume.

sindresorhus:
There's also https://github.com/kangax/compat-table which already does this. Should probably just reuse that one instead.

vdemedes:
@sindresorhus +1

thejameskyle:
@sindresorhus https://github.com/kangax/compat-table is a bad test suite for ES6 features, it's very shallow. https://github.com/tc39/test262 is better

sindresorhus:
I think we should put this on hold until we have upgraded to Babel 6.

sindresorhus:
> But I think this would be better as a separate project. Something like a ES2015 tester project that both we and Babel could consume.

Closing for the above reason. @jamestalmage If you're no longer interested in doing this and think it's still relevant, can you open an issue?

PR #270, opened 2015-11-25
Guard against clock manipulation by fake timer libraries. Fixes #128.

### LINKED ISSUE #128: Time is not calculated correctly when using a fake clock
I use [sinonjs](http://sinonjs.org/) for stubbing and stuff. But also to use a [fake timer](http://sinonjs.org/docs/#clock) that I can manipulate.

When doing this, times are not calculated correctly.

screen shot 2015-11-04 at 14 51 42

Will have a look at it myself, not sure when though.

jamestalmage:
Fixes #128

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,9 +1,9 @@
 'use strict';
-var setImmediate = require('set-immediate-shim');
 var relative = require('path').relative;
 var hasFlag = require('has-flag');
 var chalk = require('chalk');
 var serializeError = require('serialize-error');
+var globals = require('./lib/globals');
 var Runner = require('./lib/runner');
 var send = require('./lib/send');
 var log = require('./lib/logger');
@@ -65,7 +65,7 @@ function exit() {
 	});
 }
 
-setImmediate(function () {
+globals.setImmediate(function () {
 	runner.on('test', test);
 	runner.run().then(exit);
 });
--- a/test/fork.js
+++ b/test/fork.js
@@ -59,3 +59,12 @@ test('exit after tests are finished', function (t) {
 			cleanupCompleted = event.completed;
 		});
 });
+
+test('fake timers do not break duration', function (t) {
+	fork(fixture('fake-timers.js'))
+		.then(function (info) {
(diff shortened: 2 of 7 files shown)


index.js
```diff
@@ -1,5 +1,5 @@
 'use strict';
-var setImmediate = require('set-immediate-shim');
+var globals = require('./lib/globals');
```
sindresorhus: Can you place it with the other relative requires?

test/fork.js
```diff
@@ -59,3 +59,12 @@ test('exit after tests are finished', function (t) {
 			cleanupCompleted = event.completed;
 		});
 });
+
+test('fake timers do not break duration', function (t) {
+	fork(fixture('fake-timers.js'))
+		.then(function (info) {
+			var duration = info.tests[0].duration;
+			t.true(duration < 1000, duration + ' < 1000');
```
sindresorhus: Wish we had `power-assert` here so we didn't need the `duration + ' < 1000'`.

jamestalmage: Yep. I'm so tempted to go create tap support. We will self test someday, and then all will be right with the world.

jamestalmage:
@sindresorhus - made the change.

test/fork.js
jprichardson: > Yep. I'm so tempted to go create tap support. 

Pretty please, with sugar on top :blush:

jamestalmage: Only so much time in the day. Don't hold your breath. :relaxed:

sindresorhus:
:+1:

jamestalmage:
The only hole I still see in this is someone using bluebird [`setScheduler`](http://bluebirdjs.com/docs/api/promise.setscheduler.html) and mucking things up.

sindresorhus:
@jamestalmage Too much of an edge-case to care about IMHO.

jamestalmage:
I don't think it's likely to be an issue until we actually start supporting the Browser. That angular example is literally the only way I've ever seen `setScheduler` used.

PR #273, opened 2015-11-25
Stack traces for input source maps

novemberborn:
Follow-up to #262, [as requested](https://github.com/sindresorhus/ava/pull/262#issuecomment-159053622) by @jamestalmage:

> Can we also make sure we are covering the case where the test file itself may have a source map?

---

Note the second commit which is unrelated to input source map support but removes an unnecessary stringification of the transpiled source map.


sindresorhus:
Looks good to me. @jamestalmage ?

jamestalmage:
:+1: 

> Note the second commit which is unrelated to input source map support but removes an unnecessary stringification of the transpiled source map

Bah! I meant to fix that. Thanks @novemberborn.

PR #275, opened 2015-11-27
Pipe forked IO streams instead of forwarding `data` events.

jamestalmage:
This passes terminal width information to, and enables ansi-color
support in, the forked process. It makes the output of `time-require`
much prettier when used in a fork.

### DIFF
--- a/lib/fork.js
+++ b/lib/fork.js
@@ -14,8 +14,8 @@ module.exports = function (args) {
 	var file = args[0];
 
 	var options = {
-		silent: true,
-		cwd: path.dirname(file)
+		cwd: path.dirname(file),
+		stdio: ['ignore', process.stdout, process.stdout]
 	};
 
 	var ps = childProcess.fork(filepath, args, options);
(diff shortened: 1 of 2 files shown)


lib/fork.js
```diff
@@ -14,8 +14,8 @@ module.exports = function (args) {
 	var file = args[0];
 
 	var options = {
-		silent: true,
-		cwd: path.dirname(file)
+		cwd: path.dirname(file),
+		stdio: ['ignore', process.stdout, process.stdout]
```
jamestalmage: I am not sure why we forward both child output streams to `stdout`. I do it here simply because it is what we were doing before. Making that change will impact a lot of tests, and is a separate PR.

In my mind what we are currently doing makes the *least* sense of the the available options. I think we should either:

 1. pipe `child.stdout` -> `process.stdout`, and `child.stderr` -> `process.stderr`.
 2. pipe both to `process.stderr`. This keeps `process.stdout` pristine and will not interfere with our TAP output.

jamestalmage:
@sindresorhus - this seems like a no brainer to me. What we were doing before was essentially piping to stdout. This just uses the `child_process

lib/fork.js
sindresorhus: Can you open an issue with the contents of this comment? I'm not sure to be honest.

jamestalmage:
Or that! :smile:

PR #276, opened 2015-11-27
Integrate time-require into the CLI.

jamestalmage:
I want to start focusing on getting our require times down,
both in the main thread and forked tests.

This adds `time-require` support directly to the CLI to assist in that.
This is especially important for forked test processes, where I currently
have to manually add a line of code each time I want to test an optimization.

I don't think this should be a publicly documented move. I think we want
the freedom to remove this once we are happy with performance.

I have placed comments above all `time-require` references, indicating
it is for internal use only. Users have been given fair warning against
relying on its inclusion long term.

Usage:

```sh
$ ./cli.js test/fixture/es2015.js --time-require --sorted
```

`--sorted` is optional. It sorts the requires from most expensive to least. The default is sequential sorting.

The output will display multiple `time-require` reports.
One for each forked process, and the last one for the main thread.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -1,5 +1,12 @@
 #!/usr/bin/env node
 'use strict';
+
+var debug = require('debug')('ava');
+
+if (debug.enabled) {
+	require('time-require');
+}
+
 var fs = require('fs');
 var path = require('path');
 var figures = require('figures');
@@ -107,6 +114,12 @@ function run(file) {
 
 	if (cli.flags.serial) {
 		args.push('--serial');
+	}
+
+	// Forward the `time-require` `--sorted` flag.
+	// Intended for internal optimization tests only.
+	if (cli.flags.sorted) {
+		args.push('--sorted');
 	}
 
 	return fork(args)
--- a/lib/babel.js
+++ b/lib/babel.js
@@ -1,5 +1,11 @@
 'use strict';
 
+var debug = require('debug')('ava');
+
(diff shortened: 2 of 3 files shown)


sindresorhus:
Can't we just use `debug` for this instead of a separate flag?

jamestalmage:
I am not sure I understand how to use the `debug` module to accomplish that.
 
This way `time-require` is not setting it's require-hook unless the flag is set.

sindresorhus:
```js

if (debug.enabled) {
  require('time-require');
}
```

jamestalmage:
Got it. That works.

How should I handle `--sorted` vs not? It is nice to have both, but not required.

If I have to pick one, then I guess I'd prefer the default sequential sorting (easier to figure out which module is requiring `acorn` that way).

sindresorhus:
> How should I handle --sorted vs not? It is nice to have both, but not required.

If you feel it would be useful, just forward it to the forked processes like we do with the other ones.

sindresorhus:
There's also:

- https://github.com/iarna/require-timer
- https://github.com/maxogden/require-times

jamestalmage:
I just checked out both of those. They are a lot more verbose and don't provide the sort option. `time-require` provides nicer output, especially once #275 gets merged.

jamestalmage:
I think AppVeyor ate too much turkey today and is feeling sluggish.

PR #277, opened 2015-11-27
Prefer the locally installed version of AVA.

### LINKED ISSUE #157: global cli broken if there is any babel version mismatch
If there is any difference in the babel-version installed globally, vs that installed locally, you get:

```js
TypeError: Plugin "babel-plugin-espower" is resolving to a different Babel version than what is performing the transformation.
 ```
We could use [`fallback-cli`](https://www.npmjs.com/package/fallback-cli) to always use the local install of ava if there is one. Things might get weird if they had no local install, but did have `babel-runtime` installed locally. I'd need to play with it.

@sindresorhus I'm happy to tackle this, but I know [you have reservations](https://github.com/sindresorhus/xo/pull/32#issuecomment-144939822) with the idea.

### PR DESCRIPTION
**Pull request by jamestalmage**
This hunts for AVA in the locally installed project, and runs that
CLI instead of the globally installed one.

Fixes #157.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -3,6 +3,16 @@
 
 var debug = require('debug')('ava');
 
+// Prefer the local installation of AVA.
+var resolveFrom = require('resolve-from');
+var localCLI = resolveFrom('.', 'ava/cli');
+
+if (localCLI && localCLI !== __filename) {
+	debug('Using local install of AVA.');
+	require(localCLI);
+	return;
+}
+
 if (debug.enabled) {
 	require('time-require');
 }
--- a/readme.md
+++ b/readme.md
@@ -116,6 +116,7 @@ $ ava --help
 
 Files starting with `_` are ignored. This can be useful for having helpers in the same directory as your test files.
 
+*WARNING: NON-STANDARD BEHAVIOR:* The AVA CLI will always try to find and use your projects local install of AVA. This is true even when you run the global `ava` command. This non-standard behavior solves an important [issue](https://github.com/sindresorhus/ava/issues/157), and should have no impact on everyday use.
 
 ## Documentation
 
--- a/package.json
+++ b/package.json
@@ -105,7 +105,7 @@
     "power-assert-renderers": "^0.1.0",
     "pretty-ms": "^2.0.0",
     "require-from-string": "^1.1.0",
-    "resolve-from": "^1.0.0",
+    "resolve-from": "^2.0.0",
     "serialize-error": "^1.1.0",
     "set-immediate-shim": "^1.0.1",
     "source-map-support": "^0.3.3",
@@ -128,5 +128,10 @@
         "test[/\\\\]"
       ]
     }
+  },
+  "xo": {
+    "ignore": [
+      "cli.js"
+    ]
   }
 }

### CODE REVIEW

**Comment by jamestalmage** [comment]
The second commit wraps the entire CLI in an IIFE, to avoid a `ParseError` from `xo`.

https://github.com/eslint/eslint/issues/1158

**Comment by jamestalmage** [comment]
If https://github.com/sindresorhus/xo/issues/51 were to be implemented, the second commit would not be necessary.

**Comment by jamestalmage** [comment]
This needs a test suite, which is doable but will involve a number of fixtures. I would like agreement on my general approach before completing.

Right now, it outputs the following:

- Run **locally**, or globally **without** a local version: `no warning issued`

- Run globally, **with** a local version:

 ```
Using local install of AVA (v0.6.1), which differs from global (v0.6.0)
```
 or 
 ```
Using local install of AVA (v0.6.0), which matches global (v0.6.0)
```

**Comment by sindresorhus** [comment]
Instead of wrapping it in an IIFE, just remove XO for now. We can add it back when we've sorted this out.

**Comment by sindresorhus** [comment]
> Right now, it outputs the following:

I don't think it matters if they differ. All we're doing with the global CLI if local is available is to require the local one. As long as the user has AVA 0.7.0 containing this pull request installed globally they should be fine.

File: cli.js [hunk 1]
```diff
@@ -1,248 +1,269 @@
...
-
-	var base = path.dirname(cli.input[0]);
-
-	if (base === '.') {
-		base = cli.input[0] || 'test';
+(function () {
+	// Prefer the local installation of AVA.
+	var resolveFrom = require('resolve-from');
+	var localCLI;
+	try {
```
sindresorhus: Been thinking about changing `resolve-from` to just return `null` when not found instead of throwing. Thoughts?

readme.md
```diff
@@ -116,6 +116,9 @@ $ ava --help
 
 Files starting with `_` are ignored. This can be useful for having helpers in the same directory as your test files.
 
+*WARNING: NON-STANDARD BEHAVIOR:* The AVA CLI will always try to find and use your projects local install of AVA.
+ This is true even when you run the global `ava` command. This non-standard behavior solves an important [issue](https://github.com/sindresorhus/ava/issues/157),
+ and should have no impact on everyday use.
```
sindresorhus: Can you not hard-wrap?

jamestalmage: Sure. Webstorm has only global and per file setting for soft wrapping. No ability to set for just `.md` etc. That's why I have gotten in that habit. 

https://github.com/nicoulaj/idea-markdown/issues/57

cli.js
shinnn: FYI [try-resolve-from](https://github.com/shinnn/try-resolve-from/blob/50cd2cb89934ec0172e43db86fe78263759908ee/index.js) works as you expected. https://github.com/shinnn/try-resolve-from/blob/50cd2cb89934ec0172e43db86fe78263759908ee/index.js

jamestalmage: I think `resolve-from` should adopt that behavior. I can't think of very many scenarios where you would want `resolve-from`, and not want to wrap the thrown exception.

Also, it looks like someone was already assuming that behavior here: [babel.js#L41](https://github.com/sindresorhus/ava/blob/306f28d25cbf5b798dbaf0db3480f973547848b1/lib/babel.js#L41). This like the more common use-case.

Perhaps retain the old behavior somewhere:

```js
require('resolve-from/throws');
```

or something like that.

jamestalmage:
> Instead of wrapping it in an IIFE, just remove XO for now. We can add it back when we've sorted this out.

This is what led me to the `shim` behavior in `fallback-cli`, it allows you to do this:

```js
// cli-shim.js
require('fallback-cli')('ava/cli.js', function (opts) {
  // returns are allowed inside the callback - linters stay happy.
  if (opts.globalCli && opts.localCli) {
    console.warn('Using the local install of AVA instead of global.');
  }
  require(opts.cli); // opts.cli === opts.localCli || opts.globalCli
});
```

cli.js
```diff
@@ -1,6 +1,19 @@
...
 
+// Prefer the local installation of AVA.
+var resolveFrom = require('resolve-from');
+var localCLI;
+try {
+	localCLI = resolveFrom('.', 'ava/cli');
+} catch (e) {}
+
+if (localCLI && localCLI !== __filename) {
+	console.warn('Using local install of AVA.');
```
vdemedes: I don't think we need that warning, it gives no helpful information to the test output. Plus, it will not look good.

sindresorhus:
@jamestalmage Bump `resolve-from` and it will no longer throw ;)

cli.js
sindresorhus: Maybe this is something we could have in a future `--verbose` mode or just add it as a `debug()`?

sindresorhus:
Generally looks good to me.

cli.js
jamestalmage: OK. Honestly, that line was there to appease you: https://github.com/sindresorhus/xo/pull/32#issuecomment-155032045.

I thought you brought up a good point, that silently changing standard behavior is a bad thing.

@sindresorhus, what do you think?

cli.js
shinnn: OK I'll unpublish try-resolve-from. https://github.com/sindresorhus/resolve-from/commit/4cb3cf0ff07f05b64463bb680f650a77c2da5c80

cli.js
```diff
@@ -3,6 +3,18 @@
 
 var debug = require('debug')('ava');
 
+// Prefer the local installation of AVA.
+var resolveFrom = require('resolve-from');
+var localCLI = resolveFrom('.', 'ava/cli');
+
+if (localCLI && localCLI !== __filename) {
+	if (debug.enabled) {
```
sindresorhus: You don't have to check if debug is enabled. That's already done by `debug`

sindresorhus:
One minor comment. Otherwise it's good. Feel free to merge when that's resolved.

Issue #83, opened 2015-10-19
Programmatic API


sindresorhus:
So other tools can use it without having to spawn. Right now I [spawn](https://github.com/sindresorhus/gulp-ava/blob/68dab5150299c7954492421b115889036f20b538/index.js#L20) in `gulp-ava`.

This would also help us have a better separation between the CLI and the API.

The API should be high-level and do whatever the CLI does now, like forking, etc. The CLI would just be a super-thin wrapper calling the API.

```js
const Ava = require('ava').Ava;

const instance = new Ava({
  files: ['foo.js', 'test/**'],
  failFast: true
});

instance.run();

instance.log().pipe(process.stdout);
```

The above is just a quick example. I hope it will end up looking better than that.

kevva:
I'll try work on this tomorrow. Definitely needed.

Issue #27, opened 2015-08-28
TAP support


sindresorhus:
Should be a flag, like `$ ava --tap`, to get TAP output. This would be a good way to enable reporters without actually supporting reporters. Can just refer people to all the available TAP reporters instead of reinventing the wheel.

scottcorgan:
I know that a PR is better than this question, BUT, is this still a priority?

jprichardson:
:+1:

jamestalmage:
From my reading of the TAP spec, I it does not seem very well suited to concurrency. You would need to buffer individual tests and then log them all at once so they don't walk all over each-other.

vdemedes:
Coming Soon!

<img width="378" alt="screen shot 2015-11-29 at 7 18 50 pm" src="https://cloud.githubusercontent.com/assets/697676/11459052/179e680e-96ce-11e5-8a87-b092db191964.png">

vdemedes:
TAP support is ready, waiting for #279.

jprichardson:
After I saw this closed by @vdemedes I assumed it was part of `0.7.0`. But didn't see it listed https://github.com/sindresorhus/ava/releases/tag/v0.7.0. Since this issue is still part of `0.7.0`, does that mean that it's in `0.7.0` and hence forgotten about in the release announcement or that a new milestone should be associated with this issue?

jamestalmage:
@jprichardson - We decided to push #279 to [the next milestone](https://github.com/sindresorhus/ava/pull/279#issuecomment-160478598).

sindresorhus:
It was closed by a mistake.

jonathanKingston:
> @jamestalmage From my reading of the TAP spec, I it does not seem very well suited to concurrency.

That is the case certainly, I am happy to discuss some of my suggestions that have been aired on TAP specification discussions however likely won't be adopted very quickly at all (Having a different output might be simpler)

PR #279, opened 2015-11-28
Add programmatic API

vdemedes:
This PR adds a long-awaited programmatic API (#83).

To-do:

- [x] resolve errors on node v0.10
- [ ] add tests
- [ ] use API in tests instead of `child_process.exec`
- [x] implement #282

It also opens a road for #27 (TAP support).

### DIFF
--- a/api.js
+++ b/api.js
@@ -1,0 +1,230 @@
+'use strict';
+var EventEmitter = require('events').EventEmitter;
+var path = require('path');
+var util = require('util');
+var fs = require('fs');
+var flatten = require('arr-flatten');
+var Promise = require('bluebird');
+var figures = require('figures');
+var assign = require('object-assign');
+var globby = require('globby');
+var chalk = require('chalk');
+var fork = require('./lib/fork');
+
+function Api(files, options) {
+	if (!(this instanceof Api)) {
+		return new Api(files, options);
+	}
+
+	EventEmitter.call(this);
+
+	assign(this, options);
+
+	this.rejectionCount = 0;
+	this.exceptionCount = 0;
+	this.passCount = 0;
+	this.failCount = 0;
+	this.fileCount = 0;
+	this.testCount = 0;
+	this.errors = [];
+	this.stats = [];
+	this.tests = [];
+	this.files = files || [];
+
+	Object.keys(Api.prototype).forEach(function (key) {
+		this[key] = this[key].bind(this);
+	}, this);
+}
+
+util.inherits(Api, EventEmitter);
+module.exports = Api;
+
+Api.prototype._runFile = function (file) {
+	var args = [file];
+
+	if (this.failFast) {
+		args.push('--fail-fast');
+	}
+
+	if (this.serial) {
+		args.push('--serial');
+	}
+
+	// Forward the `time-require` `--sorted` flag.
+	// Intended for internal optimization tests only.
+	if (this._sorted) {
+		args.push('--sorted');
+	}
+
+	return fork(args)
+		.on('stats', this._handleStats)
+		.on('test', this._handleTest)
+		.on('unhandledRejections', this._handleRejections)
+		.on('uncaughtException', this._handleExceptions);
+};
+
+Api.prototype._handleRejections = function (data) {
+	this.rejectionCount += data.rejections.length;
+
+	data.rejections.forEach(function (err) {
+		err.type = 'rejection';
+		err.file = data.file;
+		this.emit('error', err);
+		this.errors.push(err);
+	}, this);
+};
+
+Api.prototype._handleExceptions = function (err) {
+	this.exceptionCount++;
+	err.type = 'exception';
+	this.emit('error', err);
+	this.errors.push(err);
+};
+
+Api.prototype._handleStats = function (stats) {
+	this.testCount += stats.testCount;
+};
+
+Api.prototype._handleTest = function (test) {
+	test.title = this._prefixTitle(test.file) + test.title;
+
+	var isError = test.error.message;
+
+	if (isError) {
+		this.errors.push(test);
+	} else {
+		test.error = null;
+	}
+
+	this.emit('test', test);
+};
+
+Api.prototype._prefixTitle = function (file) {
+	if (this.fileCount === 1) {
+		return '';
+	}
+
+	var separator = ' ' + chalk.gray.dim(figures.pointerSmall) + ' ';
+
+	var base = path.dirname(this.files[0]);
+
+	if (base === '.') {
+		base = this.files[0] || 'test';
+	}
+
+	base += path.sep;
+
+	var prefix = path.relative('.', file)
+		.replace(base, '')
+		.replace(/\.spec/, '')
+		.replace(/test\-/g, '')
+		.replace(/\.js$/, '')
+		.split(path.sep)
+		.join(separator);
+
+	if (prefix.length > 0) {
+		prefix += separator;
+	}
+
+	return prefix;
+};
+
+Api.prototype.run = function () {
+	var self = this;
+
+	return handlePaths(this.files)
+		.map(function (file) {
+			return path.resolve(file);
+		})
+		.then(function (files) {
+			if (files.length === 0) {
+				return Promise.reject(new Error('Couldn\'t find any files to test'));
+			}
+
+			self.fileCount = files.length;
+
+			var tests = files.map(self._runFile);
+
+			// receive test count from all files and then run the tests
+			var statsCount = 0;
+			var deferred = Promise.pending();
+
+			tests.forEach(function (test) {
+				var counted = false;
+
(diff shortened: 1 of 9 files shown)


jamestalmage:
I think this should go in the base directory:

```js
require('ava/api');
// is better than
require('ava/lib/api');
```

lib/api.js
```diff
@@ -0,0 +1,213 @@
...
+	this.passCount = 0;
+	this.failCount = 0;
+	this.fileCount = 0;
+	this.testCount = 0;
+	this.errors = [];
+	this.stats = [];
+	this.tests = [];
+
+	Object.keys(Api.prototype).forEach(function (key) {
+		this[key] = this[key].bind(this);
```
jamestalmage: Maybe add a typeof check?
```js
if (typeof this[key] === 'function') {
  this[key] = this[key].bind(this);
}
```

lib/api.js
```diff
@@ -0,0 +1,213 @@
...
+	if (isError) {
+		this.errors.push(test);
+	} else {
+		test.error = null;
+	}
+
+	this.emit('test', test);
+};
+
+Api.prototype._prefixTitle = function (file) {
```
jamestalmage: `_prefixTitle` introduces new behavior not related to exposing an API. I like where you are headed, but it should probably be a separate PR.

index.js
```diff
@@ -66,8 +66,17 @@ function exit() {
...
 globals.setImmediate(function () {
+	send('stats', {
+		testCount: runner.select({type: 'test'}).length
+	});
+
 	runner.on('test', test);
-	runner.run().then(exit);
+
+	process.on('message', function (message) {
+		if (message.name === 'ava-run') {
```
jamestalmage: Just use:

```js
process.on('ava-run', ...);
```

lib/api.js
```diff
@@ -0,0 +1,213 @@
...
+			var statsCount = 0;
+			var deferred = Promise.pending();
+
+			tests.forEach(function (test) {
+				test.on('stats', function () {
+					if (++statsCount === self.fileCount) {
+						self.emit('ready');
+
+						tests.forEach(function (test) {
+							test.run();
```
jamestalmage: Looks like `test.run()` gets called a bunch of times.

jamestalmage: Why are you waiting for every test file to load before kicking off the tests? Once an individual test is `ready`, why not let it start?

jamestalmage:
@vdemedes - see my change. 

The issue you were having was `stats` never gets emitted if there is syntax error in the test file (that is why it started failing on the `babel-hook` fixture, which intentionally has ES2015 syntax in a file that will not be instrumented by babel).

jamestalmage:
I propose we change this up just a bit.

Rather than `api` tracking counts for `unhandledRejection`, `uncaughtException`, etc.

Use the array: `api.errors`, and every time one is encountered, push to it:

```js
on('unhandledRejection', function(e) {
  self.errors.push({type: 'unhandledRejection', error: e});
});

on('uncaughtException', function(e) {
  self.errors.push({type: 'uncaughtException', error: e});
});

// ...
self.errors.push({
  type: 'nonZeroExit',
  error: new Error('process exited with code: ' + code)
});

// ...
self.errors.push({
  type: 'assertionError',
  // ...
});
```

Then extend API with methods like:
```js
api.hasErrors(); // true / false
api.getErrors(); // returns all {type: typeName, error: errorObj}
api.getErrors(typeName); // returns all errors of matching type. 
```

I think the promise returned by the API should almost always resolve, and almost never reject. Users will then use `hasErrors`, and decide what to do. This makes `unhandledRejection`, `uncaughtException`, and `nonZeroExit` consistent with what we do with a failed assertion (we resolve the promise with a result that indicates there were errors).

We should continue emitting events so people can implement their own realtime logging, but it should be possible to create basically the same log output after the fact using just the array returned from  `getErrors()`.

Thoughts?

jamestalmage:
I've shared a number of thoughts, but overall, very much :+1:

jamestalmage:
>  use API in tests instead of child_process.exec

I'm not sure of this one. I like that `test/cli.js` exercises the full stack. It currently acts as more of an integration test.

cli.js
```diff
@@ -48,133 +43,60 @@ var cli = meow([
...
-
-	return prefix;
-}
+updateNotifier({pkg: cli.pkg}).notify();
 
-function stats(stats) {
-	testCount += stats.testCount;
+if (cli.flags.init) {
+	require('ava-init')();
+	process.exit();
```
sindresorhus: You can't `process.exit()` here as `ava-init` won't have a time to finish. Just `return` and XO will pass when https://github.com/sindresorhus/ava/pull/277/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R132 lands.

cli.js
```diff
@@ -48,133 +43,60 @@ var cli = meow([
...
-function test(data) {
-	var isError = data.error.message;
-
-	if (fileCount > 1) {
-		data.title = prefixTitle(data.file) + data.title;
-	}
-
-	if (isError) {
-		log.error(data.title, chalk.red(data.error.message));
+var api = new Api({
```
sindresorhus: Maybe `new Api(files, opts)`?

lib/api.js
sindresorhus: @jamestalmage We already do this in https://github.com/sindresorhus/ava/blob/e2e45151dc5ac249d49a7cd94baca8c70f9ed2ce/cli.js#L57

lib/api.js
```diff
@@ -0,0 +1,223 @@
...
+	var self = this;
+
+	return handlePaths(this.files)
+		.map(function (file) {
+			return path.resolve(file);
+		})
+		.then(function (files) {
+			if (files.length === 0) {
+				var err = new Error('Couldn\'t find any files to test');
+				return Promise.reject(err);
```
sindresorhus: `return Promise.reject(new Error('Couldn\'t find any files to test'));`

lib/api.js
```diff
@@ -0,0 +1,223 @@
...
+			var deferred = Promise.pending();
+
+			tests.forEach(function (test) {
+				var counted = false;
+
+				function count() {
+					if (counted) {
+						return;
+					}
+					if (++statsCount === self.fileCount) {
```
sindresorhus: add an empty line above this

sindresorhus:
> We might be able to move the majority of them to test/api.js, but some should definitely stay behind. 

@jamestalmage Yup, that's the plan. We still want to explicitly test the CLI, but many of the current CLI tests really just test "API" behavior.

sindresorhus:
:+1: Overall looks really good!

lib/api.js
jamestalmage: so we do :+1:

lib/api.js
vdemedes: Copied from `lib/test.js`.

vdemedes:
@jamestalmage I appreciate your suggestions, but I'd like to ask you not to modify WIP-kind of PRs. I already had a fix in development, but I had to cancel all my changes to merge yours in.

jamestalmage:
Sure no - problem. Sorry for messing you up.

vdemedes:
@jamestalmage Good suggestion about errors!

vdemedes:
I'd really like to merge this now and add TAP support for the next release. I could add/convert tests in a separate PR. @sindresorhus @jamestalmage what do you think?

jamestalmage:
I would wait for the release. We already have one major refactor in this release. I think that we let major changes sit on master for at least a few days before release.

sindresorhus:
> I think that we let major changes sit on master for at least a few days before release.

I agree. Not really any user-facing changes here anyways.

sindresorhus:
Landed! This is exactly what I had in mind when I opened the issue about it. Really nice work @vdemedes :)

sindresorhus:
tumblr_mlf8zt33uq1qdlh1io1_400

jamestalmage:
I need to step up my GIF game.

api.js
```diff
@@ -0,0 +1,230 @@
...
+		.map(function (file) {
+			if (fs.statSync(file).isDirectory()) {
+				return handlePaths([path.join(file, '*.js')]);
+			}
+
+			return file;
+		})
+		.then(flatten)
+		.filter(function (file) {
+			return path.extname(file) === '.js' && path.basename(file)[0] !== '_';
```
azizhk: How do .jsx test files run?
@vdemedes 

Debugged to find this, but unfortunately/luckily this file has been refactored by @jamestalmage 
Hoping that its fixed in next release or minor release to patch this is released soon.

azizhk: Ok the code has been moved to https://github.com/avajs/ava/blob/e01ee00/lib/ava-files.js#L248
Will see if I can account extension from files or source.

PR #282, opened 2015-11-29
Force serial execution of child processes on --serial flag

vkholodkov:
This forces serial execution of child processes if --serial flag is specified.

Without control over concurrency test suites are unable to instantiate identical fixtures in a shared database.

Ideally this must be a property of a test suite, e.g.:

test.serialSuite() // Force serial execution of a test suite

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -212,8 +212,8 @@ function init(files) {
 
 			fileCount = files.length;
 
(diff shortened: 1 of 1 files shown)


vdemedes:
I like it! That's how it should've been done in the first place ;) Thank you for this PR!

PR #283, opened 2015-11-29
Don't crash when there's no local AVA

vdemedes:
When there's no local `ava` installed, global CLI crashes:

screen shot 2015-11-29 at 6 55 35 pm


vdemedes:
// @jamestalmage

jamestalmage:
This was an unnecessary change. You needed to `npm install` to get latest require-from

vdemedes:
@jamestalmage You are right! Reverting these now.

PR #286, opened 2015-11-30
Hide test internals. Fixes #252.

### LINKED ISSUE #252: proposal: Hide `Test` internals. Expose a delegate object instead.
We expose instances of `Test` to the user. It's the `t` in the following code:

```js
test('title', t => {
  t.is(foo, bar);
});
```

That has all kinds of internal information on it, including

 - `assertCount`
 - `planCount`
 - `duration`
 - `_timeStart`
 - `planStack`
 - etc, etc, etc.

We are already looping over all the public methods and binding them to the test [here](https://github.com/sindresorhus/ava/blob/0caf0cf7b29615ca3da94812d91c4fc288d6a7b3/lib/test.js#L43-L45) so users can pass around methods without worrying about `this` bindings.

I suggest we copy those values (perhaps being a bit more careful to select only that which we truly want public), to a secondary object that we expose.

jamestalmage:
I considered the delegates module, but that still requires exposing the Test
object itself as a member of `api`, and does not bind `this` for the exposed
functions.

Fixes #252.

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -41,10 +41,6 @@ function Test(title, fn) {
 	if (this.title === 'callee$0$0') {
 		this.title = '[anonymous]';
 	}
-
-	Object.keys(Test.prototype).forEach(function (key) {
-		this[key] = this[key].bind(this);
-	}, this);
 }
 
 module.exports = Test;
@@ -52,33 +48,6 @@ module.exports = Test;
 Test.prototype._assert = function () {
 	this.assertCount++;
 };
-
-// patch assert methods to increase assert count and store errors
-Object.keys(assert).forEach(function (el) {
-	Test.prototype[el] = function () {
-		var self = this;
-
-		try {
-			var fn = assert[el].apply(assert, arguments);
-
-			fn = observableToPromise(fn);
-
-			if (isPromise(fn)) {
-				return Promise.resolve(fn)
-					.catch(function (err) {
-						self._setAssertError(err);
-					})
-					.finally(function () {
-						self._assert();
-					});
-			}
-		} catch (err) {
-			this._setAssertError(err);
-		}
-
-		this._assert();
-	};
-});
 
 Test.prototype._setAssertError = function (err) {
 	if (this.assertError !== undefined) {
@@ -128,7 +97,7 @@ Test.prototype.run = function () {
 	var ret;
 
 	try {
-		ret = this.fn(this);
+		ret = this.fn(this._publicApi());
 	} catch (err) {
 		this._setAssertError(err);
 		this.exit();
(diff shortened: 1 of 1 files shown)


lib/test.js
```diff
@@ -230,3 +199,60 @@ Test.prototype.exit = function () {
...
+	// Getters
+	['assertCount', 'title', 'end', '_capt', '_expr']
+		.forEach(function (name) {
+			Object.defineProperty(api, name, {
+				enumerable: !/^_/.test(name),
+				get: function () {
+					return self[name];
+				}
+			});
+		}, this);
```
sindresorhus: This `this` is moot.

lib/test.js
```diff
@@ -230,3 +199,60 @@ Test.prototype.exit = function () {
 		});
 	}
 };
+
+Test.prototype.publicApi = function () {
+	var api = {};
+	var self = this;
+
+	// Getters
+	['assertCount', 'title', 'end', '_capt', '_expr']
```
sindresorhus:
```js
[
	'assertCount',
	'title',
	'end',
	'_capt',
	'_expr'
]
```

lib/test.js
```diff
@@ -230,3 +199,60 @@ Test.prototype.exit = function () {
...
+		.forEach(function (name) {
+			Object.defineProperty(api, name, {
+				enumerable: !/^_/.test(name),
+				get: function () {
```
sindresorhus: `get() {`

lib/test.js
```diff
@@ -230,3 +199,60 @@ Test.prototype.exit = function () {
...
+					return self[name];
+				}
+			});
+		}, this);
+
+	// Get / Set
+	Object.defineProperty(api, 'context', {
+		enumerable: true,
+		get: function () {
```
sindresorhus: `get() {`

lib/test.js
```diff
@@ -230,3 +199,60 @@ Test.prototype.exit = function () {
...
+			return self.context;
+		},
+		set: function (context) {
```
sindresorhus: `set() {`

sindresorhus:
Generally looks good to me.

lib/test.js
```diff
@@ -230,3 +199,60 @@ Test.prototype.exit = function () {
 		});
 	}
 };
+
+Test.prototype.publicApi = function () {
```
sindresorhus: Maybe make this `_publicApi`? I don't see why anyone would need it.

lib/test.js
```diff
@@ -230,3 +199,60 @@ Test.prototype.exit = function () {
...
+	var api = {};
+	var self = this;
```
sindresorhus: My OCD, can you put the `self` statement before `api`?

lib/test.js
jamestalmage: That doesn't work in Node `0.10` or `0.12`

Issue #288, opened 2015-11-30
Alias names to ".true" and ".false".


jamen:
These are tokens, so in some usages they fail...

For instance with how I use AVA, I destructure the functions at the top.  This way I don't have to prefix every operation with `t.` (which looks ugly in my opinion).  So I do this:

```javascript
test('test name', ({is, same, not}) => {
  // is(), same(), not()
});
```

But, whenever you when using functions with token-like names, it fails.  For example, this destructuring fails getting `.true`: 

```javascript
test('test name', ({is, true, not}) => {
  // is(), true(), not()
});
```

```
SyntaxError: Unexpected token true
  at requireFromString (~/projects/rpv/node_modules/require-from-string/index.js:26:4)
```

Would be nice if there were some alias functions for any token-similar named functions for cases like this...  `.true()` and `.false()`

jamen:
I was playing around, and discovered if you reassign the name in the destruction, it doesn't fail.  For instance:

```javascript
test('test name', ({is, true: isTrue}), {
  // is(), isTrue()
});
```

This passes.

So you can choose to merge this or not.

jamestalmage:
It seems like you have found a good solution to your problem. I don't think it is a good idea for us to start aliasing methods. Eventually we will introduce the ability to fully integrate custom assertions. When that happens, you will be able to modify our default assert plugin with the aliases you desire and apply it once across your entire project.

sindresorhus:
I don't see the point of using destructuring when it's more verbose than just doing `t.true()`. We're also not interested in adding aliases as we want to keep it consistent.

https://github.com/sindresorhus/ava/issues/114#issuecomment-152706993

jamen:
Sounds good :+1:

PR #290, opened 2015-11-30
Add aliases to .true and .false: .isTrue and .isFalse

jamen:
Aliases for `.true` and `.false`, for cases where they act like tokens instead of functions (such as destructuring): `.isTrue` and `.isFalse`.

Related #288


sindresorhus:
https://github.com/sindresorhus/ava/issues/288#issuecomment-160843135

PR #292, opened 2015-12-01
WIP - allow custom pre-processors

jamestalmage:
This adds custom preprocessor support.

No automated test yet. Run the demo like this:

    ./cli.js --compilers=coffee:cofee-script/register ./test/fixture/custom-processor.coffee

Currently it runs our own Babel transform on the coffee-script transpiled code.  This allows the use of generators, async/await, etc. without users having to manually configure Babel. I am not sure this is the right strategy. While it provides an easy path to use ES2015 goodness in your coffee-script tests, there will likely be scenarios where users want to specify their own Babel setup, and just use our runner.

 1. Should we include a flag to disable our Babel hooks?
 2. Should automatically disable our Babel hook when we see `--compilers=` or `--require=`?
 3. If we do **2**, should we provide a way to reenable?

### DIFF
--- a/api.js
+++ b/api.js
@@ -50,6 +50,12 @@ Api.prototype._runFile = function (file) {
 		args.push('--serial');
 	}
 
+	if (this.compilers) {
+		Object.keys(this.compilers).forEach(function (ext) {
+			args.push('--compilers=' + ext + ':' + this.compilers[ext]);
+		}, this);
+	}
+
 	// Forward the `time-require` `--sorted` flag.
 	// Intended for internal optimization tests only.
 	if (this._sorted) {
@@ -217,7 +223,7 @@ function handlePaths(files) {
 		})
 		.then(flatten)
 		.filter(function (file) {
-			return path.extname(file) === '.js' && path.basename(file)[0] !== '_';
+			return path.extname(file) === '.coffee' || path.extname(file) === '.js' && path.basename(file)[0] !== '_';
 		});
 }
 
--- a/cli.js
+++ b/cli.js
@@ -62,10 +62,27 @@ if (cli.flags.init) {
 
 log.write();
 
-var api = new Api(cli.input, {
+var apiOpts = {
 	failFast: cli.flags.failFast,
 	serial: cli.flags.serial
-});
+};
+
+if (cli.flags.compilers) {
+	var compilers = cli.flags.compilers;
+	if (!Array.isArray(compilers)) {
+		compilers = [compilers];
+	}
+	apiOpts.compilers = {};
+	compilers.forEach(function (compilerDef) {
+		var colonIdx = compilerDef.indexOf(':');
+		if (colonIdx === -1) {
+			throw new Error('compilers must be specified with: --compilers=<ext>:<require-path>');
+		}
+		apiOpts.compilers[compilerDef.substring(0, colonIdx)] = compilerDef.substring(colonIdx + 1);
+	});
+}
+
+var api = new Api(cli.input, apiOpts);
 
 api.on('test', function (test) {
 	if (test.error) {
--- a/lib/babel.js
+++ b/lib/babel.js
@@ -30,6 +30,8 @@ var loudRejection = require('loud-rejection/api')(process);
 var resolveFrom = require('resolve-from');
 var hasGenerator = require('has-generator');
 var serializeError = require('serialize-error');
+var path = require('path');
+var fs = require('fs');
 var send = require('./send');
 
 var testPath = process.argv[2];
@@ -55,7 +57,8 @@ var options = {
 	optional: hasGenerator ? ['asyncToGenerator', 'runtime'] : ['runtime'],
 	plugins: [powerAssert],
 	sourceMaps: true,
-	inputSourceMap: null
+	inputSourceMap: null,
+	filename: testPath
 };
 
 // check if test files required ava and show error, when they didn't
@@ -74,8 +77,33 @@ if (inputSourceMap) {
 	options.inputSourceMap = JSON.parse(inputSourceMap.map);
 }
 
+// load up custom compilers
+process.argv.slice(3).forEach(function (arg) {
+	var match = /^--compilers=.+?:(.+)$/i.exec(arg);
+	if (match) {
+		require(match[1]);
+	}
+});
+
 // include test file
-var transpiled = babel.transformFileSync(testPath, options);
+var Module = module.constructor;
+var source = fs.readFileSync(testPath, 'utf8');
+var extension = path.extname(testPath);
+
+var compiler = Module._extensions[extension];
+if (compiler) {
+	compiler(
+		{
+			_compile: function (s, f) {
+				source = s;
+				testPath = f;
+			}
+		},
+		testPath
+	);
+}
+
+var transpiled = babel.transform(source, options);
 sourceMapCache[testPath] = transpiled.map;
 requireFromString(transpiled.code, testPath, {
 	appendPaths: module.paths
--- a/package.json
+++ b/package.json
@@ -115,7 +115,9 @@
     "update-notifier": "^0.5.0"
   },
   "devDependencies": {
+    "coffee-script": "^1.10.0",
     "coveralls": "^2.11.4",
+    "delay": "^1.1.0",
     "signal-exit": "^2.1.2",
     "sinon": "^1.17.2",
     "tap": "^2.2.1",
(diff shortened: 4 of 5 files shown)


api.js
```diff
@@ -215,7 +221,7 @@ function handlePaths(files) {
 		})
 		.then(flatten)
 		.filter(function (file) {
-			return path.extname(file) === '.js' && path.basename(file)[0] !== '_';
+			return path.extname(file) === '.coffee' || path.extname(file) === '.js' && path.basename(file)[0] !== '_';
```
jamestalmage: TODO: That was a quick hack I meant to fix

cli.js
```diff
@@ -62,10 +62,27 @@ if (cli.flags.init) {
...
+	if (!Array.isArray(compilers)) {
+		compilers = [compilers];
+	}
+	apiOpts.compilers = {};
+	compilers.forEach(function (compilerDef) {
+		var colonIdx = compilerDef.indexOf(':');
+		if (colonIdx === -1) {
+			throw new Error('compilers must be specified with: --compilers=<ext>:<require-path>');
+		}
+		apiOpts.compilers[compilerDef.substring(0, colonIdx)] = compilerDef.substring(colonIdx + 1);
```
jamestalmage: I could just use `string.split()` here for a lot less code. But that would break absolute paths on Windows. Mocha just uses `split`. Absolute paths seem unlikely.

Thoughts?

lib/babel.js
```diff
@@ -74,8 +77,33 @@ if (inputSourceMap) {
 	options.inputSourceMap = JSON.parse(inputSourceMap.map);
 }
 
+// load up custom compilers
+process.argv.slice(3).forEach(function (arg) {
+	var match = /^--compilers=.+?:(.+)$/i.exec(arg);
```
jamestalmage: I used `--compilers` before really researching why that is what mocha used.
Turns out the only reason they use that over `--require` is for `mocha --watch` (they add `.coffee` to the list of extensions that should be watched).

That means, really only `--require` is needed until we add watch support ourselves.

lib/babel.js
```diff
@@ -74,8 +77,33 @@ if (inputSourceMap) {
...
+
+var compiler = Module._extensions[extension];
+if (compiler) {
+	compiler(
+		{
+			_compile: function (s, f) {
+				source = s;
+				testPath = f;
+			}
+		},
```
jamestalmage: This `_compile` is a mock `Module` implementation that prevents it from actually loading up the module (letting us further transpile the source).

lib/babel.js
```diff
@@ -74,8 +77,33 @@ if (inputSourceMap) {
...
+var source = fs.readFileSync(testPath, 'utf8');
+var extension = path.extname(testPath);
+
+var compiler = Module._extensions[extension];
+if (compiler) {
+	compiler(
+		{
+			_compile: function (s, f) {
+				source = s;
+				testPath = f;
```
novemberborn: Throw if any of these are empty? This isn't a documented Node API so people probably copy/paste from other require hooks and will dutifully pass the same filename but if they don't it may be quite tricky to debug.

lib/babel.js
```diff
@@ -74,8 +77,33 @@ if (inputSourceMap) {
...
+
 // include test file
-var transpiled = babel.transformFileSync(testPath, options);
+var Module = module.constructor;
+var source = fs.readFileSync(testPath, 'utf8');
+var extension = path.extname(testPath);
+
+var compiler = Module._extensions[extension];
+if (compiler) {
+	compiler(
```
novemberborn: Interestingly even without the `--compiler` flag being used for `.js` extensions this would load the default compiler. You could fall back to the `.js` compiler and never `readFileSync` the source above.

lib/babel.js
novemberborn: > That means, really only --require is needed until we add watch support ourselves.

I'd prefer `--require`. It's easier to reason about when writing your own test bootstrap code and still works with other require hooks.

Mocha's watcher does weird things when it reloads the tests so I've just been using nodemon to restart the process entirely. I'm not convinced watching specific extensions is even necessary.

novemberborn:
Rather than depending on coffeescript is it more illustrative to code up a simple `.txt` "compiler"?

novemberborn:
> I am not sure this is the right strategy. While it provides an easy path to use ES2015 goodness in your coffee-script tests, there will likely be scenarios where users want to specify their own Babel setup, and just use our runner.

This would work though wouldn't it? I bring my own Babel set up, and make sure my require hook is loaded first. Then AVA compiles power assertions and async/await on top of that. I don't have to worry about configuring either of those and AVA doesn't have to care how I write my code.

lib/babel.js
jamestalmage: > Mocha's watcher does weird things when it reloads the tests

By nature of the way we fork new processes for each test, that should not be an issue for us.

>  I'm not convinced watching specific extensions is even necessary.

There is [discussion](https://github.com/sindresorhus/ava/issues/115) of including an option to have the watcher only run tests affected by the change. Though perhaps a better solution would be to just monitor the keys of `Module._extensions`, or allow a list of extensions to be passed to the watcher flag `--watch=jsx,coffee,...`

jamestalmage:
> Rather than depending on coffeescript is it more illustrative to code up a simple .txt "compiler"?

Sure - I will probably steel from whatever simplified transforms you end up adding to the `nyc` test suite. :stuck_out_tongue_winking_eye: 

> This would work though wouldn't it? 

I'm not sure, but I really doubt it. The issue blocking our Babel 6 adoption is actually related to Babel running a particular transform twice. (It transforms all `typeof` statements, but the generated code contains a `typeof` statement, which gets transformed again).

Perhaps (with Babel 6), we could fallback to only running the `power-assert` transform, instead of the full `es2015` preset.

lib/babel.js
jamestalmage: Yep. This was quick and dirty. Final solution would need that. I didn't want to invest a lot of time in it, because I'm leery of mocking an undocumented API this way. To be honest, I anticipated a little more of an "Are you nuts?" reaction to this.

Do you see any need to mock more of the implementation? I have not yet seen an extension that uses anything beyond `_compile`.

jamestalmage:
The more I think about this, we really can't run the `power-assert` transform after `regenerator` or `asyncToGenerator` transforms. `power-assert` needs to see the original AST - it uses it to pass a copy of the source to it's renderers.

```js
t.is(await foo, await bar);
```

becomes (very approximately):

```js
t.is(t._expr(t._capt(await foo), {source: 't.is(await foo, await bar'}), ...
```

applying after the asyncToGenerator transform, the source passed to `power-assert` would end up scrambled.

```js
t.is(ctx.$0$0, ctx.$0$1)
```

I am beginning to wonder if it would not be best to just adopt a "take it or leave it" approach to the transform setup we provide. If users want to provide their own transform they need to provide a *complete* one, and we just don't touch it. We can provide documentation (and possibly convenience methods) for including the `power-assert` transform.

At a minimum, I think we need to provide a way for them to tell us to skip our transform via a flag in the CLI.

// @sindresorhus @vdemedes

novemberborn:
> Perhaps (with Babel 6), we could fallback to only running the power-assert transform, instead of the full es2015 preset.

That's what I was thinking, yes.

> The more I think about this, we really can't run the power-assert transform after regenerator or asyncToGenerator transforms. power-assert needs to see the original AST - it uses it to pass a copy of the source to it's renderers.

Yes that sounds like a show-stopper for this. Too bad! (Unless we revert that line using a source map… but that gets too complicated.)

> I am beginning to wonder if it would not be best to just adopt a "take it or leave it" approach to the transform setup we provide. If users want to provide their own transform they need to provide a complete one, and we just don't touch it. We can provide documentation (and possibly convenience methods) for including the power-assert transform.
> 
> At a minimum, I think we need to provide a way for them to tell us to skip our transform via a flag in the CLI.

:+1:

lib/babel.js
novemberborn: [Proxyquire does](https://github.com/thlorenz/proxyquire/blob/b9b47a48901059a411f0f88870009db9f03f02d2/lib/proxyquire.js#L241) but that wouldn't impact the test file.



> I'm leery of mocking an undocumented API this way.

[nyc mocks it](https://github.com/bcoe/nyc/blob/17dd92eb358b3c8e0fb4abdb34228bbbadcb689b/index.js#L151) but ignores the second argument to `_compile()`.

PR #296, opened 2015-12-03
Node.js-style --require CLI argument

jokeyrhyme:
I figured I'd have an attempt at being able to pass the `--require` through to the Node executable used for the child process (#229).

My test project uses `import` and `export`, so doesn't execute in Node v5 without babel/register. I can manually confirm using `npm link` that the changes I have so far function as expected.

Besides general comments (very much appreciated), I'm a little bit at a loss as where to start with the tests. Do I need to confirm (with tests) that the `ava --require moduleId` works? Or is it sufficient to use mocks and confirm that arguments are being passed through as expected?

### DIFF
--- a/api.js
+++ b/api.js
@@ -18,6 +18,7 @@ function Api(files, options) {
 	EventEmitter.call(this);
 
 	this.options = options || {};
+
 	this.rejectionCount = 0;
 	this.exceptionCount = 0;
 	this.passCount = 0;
--- a/cli.js
+++ b/cli.js
@@ -17,6 +17,7 @@ if (debug.enabled) {
 	require('time-require');
 }
 
+var arrify = require('arrify');
 var meow = require('meow');
 var updateNotifier = require('update-notifier');
 var chalk = require('chalk');
@@ -35,6 +36,7 @@ var cli = meow([
 	'  --init       Add AVA to your project',
 	'  --fail-fast  Stop after first test failure',
 	'  --serial     Run tests serially',
+	'  --require    Module to preload (Can be repeated)',
 	'',
 	'Examples',
 	'  ava',
@@ -46,7 +48,10 @@ var cli = meow([
 	'Default patterns when no arguments:',
 	'test.js test-*.js test/*.js'
 ], {
-	string: ['_'],
+	string: [
+		'_',
+		'require'
+	],
 	boolean: [
 		'fail-fast',
 		'serial'
@@ -64,7 +69,8 @@ log.write();
 
 var api = new Api(cli.input, {
 	failFast: cli.flags.failFast,
-	serial: cli.flags.serial
+	serial: cli.flags.serial,
+	require: arrify(cli.flags.require)
 });
 
 api.on('test', function (test) {
--- a/lib/babel.js
+++ b/lib/babel.js
@@ -16,6 +16,11 @@ if (debug.enabled) {
 // Bind globals first, before anything has a chance to interfere.
 var globals = require('./globals');
 
+var resolveCwd = require('resolve-cwd');
+(opts.require || []).forEach(function (moduleId) {
+	require(resolveCwd(moduleId));
+});
+
 var sourceMapCache = Object.create(null);
 
 var sourceMapSupport = require('source-map-support');
@@ -33,7 +38,6 @@ sourceMapSupport.install({
 var createEspowerPlugin = require('babel-plugin-espower/create');
 var requireFromString = require('require-from-string');
 var loudRejection = require('loud-rejection/api')(process);
-var resolveCwd = require('resolve-cwd');
 var hasGenerator = require('has-generator');
 var serializeError = require('serialize-error');
 var send = require('./send');
--- a/readme.md
+++ b/readme.md
@@ -102,6 +102,7 @@ $ ava --help
     --init       Add AVA to your project
     --fail-fast  Stop after first test failure
     --serial     Run tests serially
+    --require    Module to preload (Can be repeated)
 
   Examples
     ava
--- a/test/fixture/install-global.js
+++ b/test/fixture/install-global.js
@@ -1,0 +1,1 @@
+global.foo = 'bar';
--- a/test/fixture/validate-installed-global.js
+++ b/test/fixture/validate-installed-global.js
@@ -1,0 +1,3 @@
+import test from '../../';
+
+test(t => t.is(global.foo, 'bar'));
--- a/package.json
+++ b/package.json
@@ -78,6 +78,7 @@
   ],
   "dependencies": {
     "arr-flatten": "^1.0.1",
+    "arrify": "^1.0.0",
     "ava-init": "^0.1.0",
     "babel-core": "^5.8.23",
     "babel-plugin-espower": "^1.1.0",
--- a/test/api.js
+++ b/test/api.js
@@ -260,3 +260,17 @@ test('test file in node_modules is ignored', function (t) {
 			t.true(/Couldn't find any files to test/.test(err.message));
 		});
 });
+
+test('Node.js-style --require CLI argument', function (t) {
+	t.plan(1);
+
+	var api = new Api(
+		[path.join(__dirname, 'fixture/validate-installed-global.js')],
+		{require: [path.join(__dirname, 'fixture', 'install-global.js')]}
+	);
+
+	api.run()
+		.then(function () {
+			t.is(api.passCount, 1);
+		});
+});


jamestalmage:
> Do I need to confirm (with tests) that the ava --require moduleId works?

That would be my approach, does not need to be fancy.

```js
// test/fixture/install-global.js

global.foo = 'bar'
```

```js
// test/fixture/validate-installed-global.js

import test from '../../'

test(t => t.is(global.foo, 'bar'));
````

Then use the API to call `validate-...` using the require hook you've just implemented.

Overall looks pretty good.

File: api.js [hunk 1]
```diff
@@ -56,7 +61,15 @@ Api.prototype._runFile = function (file) {
 		args.push('--sorted');
 	}
 
-	return fork(args)
+	var options = {};
+
+	if (Array.isArray(this.require) && this.require.length) {
```
jamestalmage: Just do 
```js
if (this.require) {
 // ...
}
```

You have already guaranteed it is an array of at least length one above.

api.js
```diff
@@ -56,7 +61,15 @@ Api.prototype._runFile = function (file) {
...
+		options.execArgv = this.require.reduce(function (execArgv, moduleId) {
+			return execArgv.concat(['--require', moduleId]);
+		}, []);
```
jamestalmage: My preference would be that you not use `reduce`.

```js
options.execArgv = [];
this.require.foreEach(moduleId => options.execArgv.push('--require', moduleId));
```

cli.js
```diff
@@ -35,6 +35,7 @@ var cli = meow([
 	'  --init       Add AVA to your project',
 	'  --fail-fast  Stop after first test failure',
 	'  --serial     Run tests serially',
+	'  --require m  passed through to Node.js executable, multiples allowed',
```
jamestalmage: Tell people what it does. Don't make them go read the Node documentation.

lib/fork.js
```diff
@@ -5,18 +5,19 @@ var Promise = require('bluebird');
 var debug = require('debug')('ava');
 var send = require('./send');
 
-module.exports = function (args) {
+module.exports = function (args, options) {
 	if (!Array.isArray(args)) {
 		args = [args];
 	}
+	if (!options || typeof options !== 'object') {
```
jamestalmage: Overkill for something not public facing. Just do:

```js
options = options || {}
```

cli.js
```diff
@@ -64,6 +65,7 @@ log.write();
 
 var api = new Api(cli.input, {
 	failFast: cli.flags.failFast,
+	require: cli.flags.require,
```
sindresorhus: Put it below `serial`

lib/fork.js
```diff
@@ -5,18 +5,19 @@ var Promise = require('bluebird');
 var debug = require('debug')('ava');
 var send = require('./send');
 
-module.exports = function (args) {
+module.exports = function (args, options) {
```
sindresorhus: `options` => `opts`

lib/fork.js
sindresorhus: It should use https://github.com/sindresorhus/object-assign so not to mutate the user-supplied object.

sindresorhus:
Not sure we ever got an agreement on this:

> Would exposing the Node.js binary --require flag **while disabling our Babel hook solve your problem**

@jamestalmage ^

jamestalmage:
I don't know if our Babel hook should be disabled automatically. I still have not decided the best way to make it all work.

My current thinking is using something along the lines of #297, but using a better version of `istanbul-lib-hook`. https://github.com/jamestalmage/transform-chain would be part of it. Combine `transform-chain` with a way of detecting when hooks are installed, similar to how @novemberborn did it [here](https://github.com/bcoe/nyc/blob/952850a6a74c92afca361efdd74c63a3a1de8b18/index.js#L168-L180).

jokeyrhyme:
Unless I somehow accomplished this accidentally, I did not attempt to disable ava's babel hook. My only purpose with this PR was the pass through the `--require`.

Also, I hope to get time to address the PR feedback over the next few days. I'm sorry for not being very responsive.

jamestalmage:
>  Unless I somehow accomplished this accidentally, I did not attempt to disable ava's babel hook

Understood, we are still deciding if / when / how to do that. Once a decision on that gets made we will merge this, and augment it with whatever behavior we finally decide on. 

> I'm sorry for not being very responsive.

That is not a problem. We still have to answer a few questions, so we are in no rush to merge.

jokeyrhyme:
I still haven't pushed tests yet. They don't seem quite as trivial as I'd hoped, even with @jamestalmage's suggestions. Soon...

jokeyrhyme:
Really stuck with the testing part. I'm getting similar errors when I try to use `api.js` instead of `cli.js`.

I've tried a few simple approaches, and none of them work. But this functionality doesn't seem complicated enough to warrant a complicated test. Is there an obvious flaw in this approach? Or do I need to take this up a notch?

jokeyrhyme:
Thanks for the hint, @jamestalmage. At least now, the errors are all the same one:
> Test files must be run with the AVA CLI

jokeyrhyme:
I could use mockery and just confirm that Node.js is called with the expected parameters, without actually calling Node.js? /shrug

jamestalmage:
Well, I'm having a hard time getting this to work on travis. Test #4 seems to work on OSX. Note that the `--require` flag does not work on Node `<4`

jamestalmage:
@sindresorhus @vdemedes - either of you develop on linux? Any advice on how he should fix this?

jokeyrhyme:
So I should detect the version of Node in the tests and only run them if for >=4.0?
And there's still something funny going on with the paths. The fact that it's so hard to test properly makes me wonder if I've taken a good approach in the first place. :S

jamestalmage:
I don't know. Maybe it would just be better to manually implement the same semantics as the `--require` flag ourselves. You could just do `require(somePath)`, for every passed flag inside `babel.js`. Might end up being more reliable.

sindresorhus:
> either of you develop on linux? Any advice on how he should fix this?

I have Linux in a VM, but I almost never use it, so pretty noobish with it.

-

> Maybe it would just be better to manually implement the same semantics as the --require flag ourselves.

:+1: This is where the `--require` magic happens: https://github.com/nodejs/node/blob/04b1a2f7564d70dfdab838d2830e04af34b97fe6/lib/module.js#L469-L488 Which is pretty much just `require(resolveCwd(cli.flags.require));`

jokeyrhyme:
Okay, well I'll have another crack at it over the next few days. I'll try implementing `--require` internally so that it works regardless of the Node.js version.

api.js
```diff
@@ -31,6 +31,11 @@ function Api(files, options) {
 	this.tests = [];
 	this.files = files || [];
 
+	if (this.require && !Array.isArray(this.require)) {
+		// normalise single `--require moduleId` usage
+		this.require = [this.require];
+	}
```
sindresorhus: Just use [`arrify`](https://github.com/sindresorhus/arrify) on line 69 and drop this whole thing.

api.js
```diff
@@ -56,7 +61,17 @@ Api.prototype._runFile = function (file) {
...
 
-	return fork(args)
+	var options = {
+		execArgv: []
+	};
+
+	if (this.require) {
+		this.require.forEach(function (moduleId) {
+			options.execArgv.push('--require', moduleId);
+		});
```
sindresorhus: I think it's time we simplify argument passing to the forked process.

@jamestalmage @vdemedes I propose we just JSON.stringify() the data we want to pass instead of using CLI args. Thoughts?

api.js
```diff
@@ -56,7 +61,17 @@ Api.prototype._runFile = function (file) {
 		args.push('--sorted');
```
sindresorhus: Sidenote: I think it's time we simplify argument passing to the forked process. Too much code now just proxying CLI args.

@jamestalmage @vdemedes I propose we just JSON.stringify() the data we want to pass instead of using CLI args. Thoughts?

sindresorhus:
```js
requires.forEach(function (el) {
  require(resolveCwd(el));
));
```

api.js
jokeyrhyme: Worth switching to [execa](https://github.com/sindresorhus/execa) at the same time?

test/fixture/install-global.js
```diff
@@ -0,0 +1 @@
+global.foo = 'bar'
```
sindresorhus: semicolon

test/fixture/validate-installed-global.js
```diff
@@ -0,0 +1,3 @@
+import test from '../../'
```
sindresorhus: semicolon

test/fixture/install-global.js
jokeyrhyme: Weird. I love my semi-colons, too. I wonder why ESLint didn't bust my chops about this? :)

cli.js
```diff
@@ -35,6 +35,7 @@ var cli = meow([
 	'  --init       Add AVA to your project',
 	'  --fail-fast  Stop after first test failure',
 	'  --serial     Run tests serially',
+	'  --require    module to preload (option can be repeated)',
```
sindresorhus:
```
'  --require    Module to preload [Can be repeated]',
```

test/fixture/validate-installed-global.js
jamestalmage: @sindresorhus is on a semicolon rampage today.

@semisorhus?

api.js
sindresorhus: @jokeyrhyme No, we use use `child_prcess.fork` which is different. We could probably use it in the CLI tests, though.

test/fixture/validate-installed-global.js
sindresorhus: @jamestalmage Yeah, just XO that ignores it because it's fixtures. https://github.com/sindresorhus/xo/blob/0a5060c717876f77d9bcc9b8b453ba14eec332f1/index.js#L20 Starting to question that.

test/fixture/install-global.js
sindresorhus: https://github.com/sindresorhus/ava/pull/296#discussion_r47035400

api.js
jamestalmage: > I propose we just JSON.stringify() the data we want to pass instead of using CLI args.

:+1:

test/fixture/validate-installed-global.js
jamestalmage: I thought we only ignored tests/fixtures because it might have some `esnext` stuff in it, while production might not?

It would be nice if XO had an easy way to multiple configs (and a way to specify which got applied would depend on glob patterns).

api.js
sindresorhus: Ok, opened an issue: https://github.com/sindresorhus/ava/issues/318

test/fixture/validate-installed-global.js
sindresorhus: > I thought we only ignored tests/fixtures because it might have some esnext stuff in it, while production might not?

Yes, that's why XO ignores them, because fixtures might not follow code style or anything.

> It would be nice if XO had an easy way to multiple configs (and a way to specify which got applied would depend on glob patterns).

Yup, I've been planning this for a long time, just haven't had the time to implement it yet.

https://github.com/sindresorhus/xo/issues/22

lib/fork.js
```diff
@@ -5,20 +5,19 @@ var Promise = require('bluebird');
 var debug = require('debug')('ava');
 var send = require('./send');
 
-module.exports = function (args) {
+module.exports = function (args, opts) {
 	if (!Array.isArray(args)) {
 		args = [args];
 	}
+	opts = opts || {};
```
sindresorhus: put an empty line above this one

billyjanitsch:
Sorry for the late reply; I didn't realize that progress was being made on #229. FWIW, I would definitely prefer that the Babel hook be disabled. Otherwise, if I'm using Babel as part of my custom register script, it needs to get loaded once to compile the tests, and then again for each individual test require. Isn't this is a *huge* perf loss (due to all of the Babel spin-ups, and the loss of its internal module caching)?

For reference, using the `require('babel-core/register')` workaround at the top of each of my tests (which, as I understand, is functionally equivalent to what passing `--require babel-core/register` would look like without disabling the hook) makes my test suite run around 12x slower in AVA than in its Mocha equivalent (passing `--compilers`).

Also, unless I happen to be using exactly the same Babel options as AVA, I'd prefer not to write my test and lib code in different subsets of ES6/7/JSX.

jokeyrhyme:
As an interim solution, we could check to see if "babel-core/register" has been `--require`d, and automatically disable the built-in babel hook?

billyjanitsch:
I'm not sure that's the best solution. I want to use a custom register hook that includes the Babel hook along with other patches (extending `require` for non-js imports), so I won't be passing `babel-core/register` to `--require` directly.

jamestalmage:
> Isn't this is a huge perf loss (due to all of the Babel spin-ups, 

Babel does not take long to spin up. It does take a good long while to `require` the first time. After that babel is in the cache, whether you register again or not. A call to `require('babel-core/register')` would result in a small eval of only a few additional files. Not worth worrying about.

> and the loss of its internal module caching)?

Not sure what you mean here. Node caches modules for you already. `babel-core/register` shouldn't be going and deleting things from your require cache. (if it does, I'm surprised).

billyjanitsch:
@jamestalmage Any idea why Mocha would be outperforming AVA by such a huge factor, in that case? Are there existing perf issues that you're aware of? (Sorry, not directly related to this PR)

jamestalmage:
Yes. It is definitely Babel related, but also architectural in some regard. We spawn a whole new process for every test file. Each new spawned process currently loads up Babel to perform transforms in that individual process. Doing this for every process is redundant and is expensive. 

Also, mocha is not inherently faster, it depends on what you are testing. If you have lots of small test files that execute really really quickly, then the expensive Babel load up times per process are likely what is killing your performance. If you have a bunch of slow tests (i.e. with network or disk access), spread across just a few files, then AVA will likely be far more performant, because you will be better utilizing your CPU's by running lots of tests concurrently across multiple forked processes.

We have plans to mitigate the Babel load up times by only loading it in the main thread, and having child threads defer to it for transformation. That is not an easy task, since `require` must happen synchronously, and inter process communication is async. [I have a solution for that](https://github.com/jamestalmage/__share_transpile_across_process), but it will take some time for it to find it's way into AVA.

jokeyrhyme:
I'll take another look at this once #319 is merged.

jokeyrhyme:
Okay, so I think this is working now and the tests are passing.

I had to compensate for the new api.js tests in #313, as they call `new Api()` without a 2nd parameter.

jokeyrhyme:
Also, I added 4 tests to check different argument orders, but we should probably just pick the one we like and get rid of the rest?

lib/babel.js
```diff
@@ -16,6 +16,11 @@ if (debug.enabled) {
 // Bind globals first, before anything has a chance to interfere.
 var globals = require('./globals');
 
+var resolveCwd = require('resolve-cwd');
+(opts.require || []).forEach(function (moduleId) {
```
sindresorhus: I think `opts.require` is already guaranteed to be an array, so this check is moot.

https://github.com/sindresorhus/ava/pull/296/files#diff-2cce40143051e25f811b56c79d619bf5R73

test/cli.js
```diff
@@ -93,3 +94,57 @@ test('throwing a anonymous function will report the function to the console', fu
...
+	execCli([
+		'--require',
+		path.join(__dirname, 'fixture', 'install-global.js'),
+		'fixture/validate-installed-global.js'
+	], function (err, stdout, stderr) {
+		t.ifError(err);
+		t.is(stderr.trim(), '1 test passed');
+		t.end();
+	});
+});
```
sindresorhus: Can you make these `api.js` tests instead? Spawning CLI for each test is slow and we now have an API we can test directly instead.

cli.js
```diff
@@ -35,6 +36,7 @@ var cli = meow([
 	'  --init       Add AVA to your project',
 	'  --fail-fast  Stop after first test failure',
 	'  --serial     Run tests serially',
+	'  --require    Module to preload (Can be repeated)',
```
sindresorhus: This needs to be updated in the readme.md too.

api.js
```diff
@@ -19,7 +19,7 @@ function Api(files, options) {
 	EventEmitter.call(this);
 
 	objectAssign(this, options);
-	this.options = options;
+	this.options = options || {};
```
sindresorhus: I already fixed this in https://github.com/sindresorhus/ava/commit/3fdf44d1acbe8c6f14eb3865352ce49812c4a956. Can you rebase to fix the merge conflict?

test/cli.js
```diff
@@ -93,3 +94,57 @@ test('throwing a anonymous function will report the function to the console', fu
 		t.end();
 	});
 });
+
+test('Node.js-style --require CLI argument #1', function (t) {
```
sindresorhus: Instead of `#1`, `#2`, etc, it would be better if they just said specifically what they were testing differently than the others.

lib/babel.js
jokeyrhyme: Nope. In the API tests, `new Api()` is called without the second parameter, which means it ends up being an empty Object, without a "require" property. Should I change the fallback default value in api.js to include an empty "require" array? e.g. `this.options = options || { require: [] };` ?

cli.js
jokeyrhyme: Done.

lib/babel.js
sindresorhus: Oh, never mind then. I read it wrong.

jokeyrhyme:
Okay, so in total I have 1 api test and 1 cli test. I left the cli test in there to stress the end-to-end process a bit. Of course, if you think this is overkill then I'm happy to drop the cli test completely.

test/cli.js
jokeyrhyme: Done, see https://github.com/sindresorhus/ava/pull/296#issuecomment-163786259

readme.md
```diff
@@ -697,4 +698,3 @@ Concurrency is not parallelism. It enables parallelism. [Learn more.](http://sta
 	<br>
 	<br>
 </div>
-
```
sindresorhus: Don't remove

sindresorhus:
> I left the cli test in there to stress the end-to-end process a bit

I think it's overkill to CLI tests all options, but I'm ok with this if you rework it into being a generic CLI option test, meaning change the test description.

jokeyrhyme:
I gave it a whirl, but it seemed infeasible to mock the boundary from cli.js to the others, and my other approach to CLI argument tests seemed unnecessarily complex (and still didn't work).

If we still feel strongly about having CLI tests, we can have a separate Issue / PR.

sindresorhus:
LGTM

vdemedes:
LGTM

sindresorhus:
Thank you @jokeyrhyme! :) 💥👌

Keep the quality contributions coming. We really appreciate your help.

vdemedes:
Thank you for your work, @jokeyrhyme!

ariporad:
Hey, @sindresorhus, @vdemedes, @jamestalmage: Any idea when this might ship? I'd really prefer to not depend directly on master. Thanks!

sindresorhus:
@ariporad We're planning a new release when https://github.com/sindresorhus/ava/pull/326 lands.

ariporad:
Thanks @sindresorhus!

sindresorhus:
@ariporad Actually, I did a release now. https://github.com/sindresorhus/ava/releases/tag/v0.8.0

ariporad:
Yay! Thanks!

jamestalmage:
> Any idea why Mocha would be outperforming AVA by such a huge factor, in that case? Are there existing perf issues that you're aware of? (Sorry, not directly related to this PR)

@billyjanitsch 

There have been major performance improvements since this was discussed. If possible, could you give AVA a try again and tell us how we're doing?

billyjanitsch:
@jamestalmage 

That's great to hear!

Currently [this](https://github.com/sindresorhus/ava/pull/398#issuecomment-169255283) is the biggest migration blocker for us: it's important that our tests run in the same transpilation environment as our lib code (we now use several custom Babel transforms). Our test suite is large enough (>100 files) that I'd rather not rewrite it in AVA until I know this will be possible.

That said, I'd be happy to convert a couple of our tests to see what the perf looks like relative to Mocha.

novemberborn:
@billyjanitsch 

> Currently this is the biggest migration blocker for us: it's important that our tests run in the same transpilation environment as our lib code (we now use several custom Babel transforms).

My approach to this is to transpile using `babel-cli` and run the tests on the transpiled code as it sits on the disk. That way I know that I'm testing the code that I'll end up publishing.

See [`identifierfy`](https://github.com/novemberborn/identifierfy) for an example.

jamestalmage:
> Our test suite is large enough (>100 files) ...

Wow! I would love to see a project that big converted. Is it open source? What is your current test setup? (mocha with chai for assertions, etc)

You may also want to wait for customizable assertions, so we can create one that matches your current assertion api exactly.

billyjanitsch:
@novemberborn 

> My approach to this is to transpile using babel-cli and run the tests on the transpiled code as it sits on the disk.

I think you're talking about transpiling [lib code](https://github.com/novemberborn/identifierfy/tree/master/src) (imported by tests). This PR provides an easier way to do that: `--require 'babel-core/register'`.

I'm talking about having [*test code*](https://github.com/novemberborn/identifierfy/blob/master/test.js) transpiled in the same way that your lib code is, rather than letting AVA transpile it. See the comment that I linked to for more info.

@jamestalmage 

> Is it open source? What is your current test setup? 

Unfortunately not. I work on [Kensho](http://kensho.com)'s front-end and we're currently using Mocha/Chai (assert style). I'd like to move to a flat Tape/AVA-style interface.

We do have an open source [React chart lib](https://github.com/kensho/orama) -- also using Mocha but [faking](https://github.com/kensho/orama/blob/master/src/Tooltip/index.test.js#L2) a Tape-ish API -- that we might be able to convert as well.

> You may also want to wait for customizable assertions, so we can create one that matches your current assertion api exactly.

I'm actually not a huge fan of our current assertions, so I don't mind rewriting them when the time comes. AVA's defaults seem great.

jamestalmage:
@billyjanitsch https://github.com/sindresorhus/ava/issues/424#issuecomment-172107059

PR #297, opened 2015-12-03
WIP - use istanbul-require-hook

jamestalmage:
This is just another exploratory PR.
Trying to figure out the best way to accommodate custom preprocessors.

`istanbul-require-hook` was mentioned here: https://github.com/bcoe/nyc/issues/70.

I think it shoes great promise. Especially if altered to include hook combining/chaining with something similar to what was done here https://github.com/bcoe/nyc/pull/65

I think we are nearing a real solution. Until a clear winner emerges from the solutions proposed in  https://github.com/bcoe/nyc/issues/70, I think I will fork `istanbul-require-hook` with some changes specific to what we need.

### DIFF
--- a/lib/babel.js
+++ b/lib/babel.js
@@ -25,7 +25,8 @@ sourceMapSupport.install({
 });
 
 var createEspowerPlugin = require('babel-plugin-espower/create');
-var requireFromString = require('require-from-string');
+var libHook = require('istanbul-lib-hook');
+var path = require('path');
 var loudRejection = require('loud-rejection/api')(process);
 var resolveFrom = require('resolve-from');
 var hasGenerator = require('has-generator');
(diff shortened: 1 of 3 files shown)


ariporad:
Two things:
1. It's called `istanbul-lib-hook`, not `istanbul-require-hook`, FYI
2. I think that it would be worthwhile to use `pirates` instead, as `istanbul-lib-hook` will (hopefully) soon just be a wrapper around pirates.

jamestalmage:
@ariporad - yep, this was an exploratory thing I did when that whole discussion kicked off.

I am going to be opposed to integrating pirates in any projects until https://github.com/ariporad/pirates/issues/5 is resolved.

ariporad:
@jamestalmage: Fair. I've just posted an update in that issue, in the interest of now flooding this one.

PR #299, opened 2015-12-03
Display skipped tests

vdemedes:
Fix regression introduced in #243. Skipped tests were ignored and not displayed in the output. This PR brings back skipped tests:

screen shot 2015-12-03 at 12 08 31 pm

### DIFF
--- a/index.js
+++ b/index.js
@@ -66,7 +66,7 @@ function exit() {
 }
 
 globals.setImmediate(function () {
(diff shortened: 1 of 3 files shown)


jamestalmage:
Needs a test so the regression doesn't happen again.

Otherwise :+1:

sindresorhus:
:+1: when test is added.

vdemedes:
Added test ;)

Issue #298, opened 2015-12-03
`t.same` masks RangeError that is thrown when testing objects with circular references. 


jamen:
I'm not sure how far this goes outside of context, but this is what I experienced with my project...  When you create a constructor object in JavaScript and give it a circular property, AVA fails where it normally wouldn't...  Take this example for instance:

```javascript
// Example of a object that has a circular property.
let Example = function(){
  this.test = this;
};

test('circular testing', ({same}) => {

  let foo = new Example();
  let bar = new Example();

  same(foo, bar);

});
```

```
  1. circular testing
  AssertionError: Example { test: [Circular] } === Example { test: [Circular] }
```

jamestalmage:
That assertion will fail with just about any assertion library.  Using `assert.deepStrictEqual` (on which `same` is based).

``` 
  1 test failed

  1. circular testing
  RangeError: Maximum call stack size exceeded
```

So, this is not a problem with our assertion per se, but that we are masking the true source of the error. I am pretty sure that this is actually related to `power-assert`.

@twada - how should we handle this? A simple `assertError instanceof assert.AssertionError` would work for `AVA`, but not every assertion library extends `assert.AssertionError`

twada:
@jamestalmage Thank you for clarification.

> this is not a problem with our assertion per se, but that we are masking the true source of the error. I am pretty sure that this is actually related to power-assert.

You mean AVA is masking the source of error [here](https://github.com/sindresorhus/ava/blob/v0.7.0/lib/assert.js#L62-L63). I think so too. It converts `RangeError` to `AssertionError`, so `power-assert` (wrapping `same`) report it with power-assert context.

To avoid it, yes, `assertError instanceof assert.AssertionError` may help.
`power-assert` will rethrow `RangeError` without any modification.

> how should we handle this? A simple assertError instanceof assert.AssertionError would work for AVA,

Yes, with AVA's default, [underlying assert module](https://github.com/sindresorhus/ava/blob/v0.7.0/lib/assert.js#L2) is `core-assert` so you can trust it.

>  but not every assertion library extends assert.AssertionError

Is this topic related to AVA? Or talking in generality?

jamestalmage:
Ahh - you are right. I forgot we did that. Thanks @twada

twada:
@jamestalmage My pleasure! :smiley:

jamestalmage:
#300 prevents the error masking.
#301 uses `deeper` to allow comparison of circular objects.

[...]

PR #300, opened 2015-12-03
Do not mask RangeError when comparing circular values.

jamestalmage:
#300 has an alternate solution (allow circular comparison).

Fixes: #298


sindresorhus:
> #300 has an alternate solution (allow circular comparison).

This is #300 :p

PR #301, opened 2015-12-03
Allow deepEqual comparison of values with circular references. Fixes #298.

### LINKED ISSUE #298: `t.same` masks RangeError that is thrown when testing objects with circular references.

jamestalmage:
Introduces an additional dependency: `deeper`.

Fixes #298.

Alternate to #300.

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -1,5 +1,6 @@
 'use strict';
 var assert = require('core-assert');
+var deepEqual = require('deeper');
 var observableToPromise = require('observable-to-promise');
 var isPromise = require('is-promise');
 var x = module.exports;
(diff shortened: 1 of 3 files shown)


sindresorhus:
I will always favor solutions that just makes things work over throwing, as long as it's not too magic :)

PR #302, opened 2015-12-04
allow destructuring of `t` parameter.

jamestalmage:
There were two problems with using destructuring on the `t` parameter passed to tests:

 1. `t.end` is enumerable, but throws when accessed. 
This created problems when destructuring using rest params `{ok, ...t}`. 
The solution: `t.end` is enumerable only in "callback mode".

 2. `t._capt`, `t._expr` are not enumerable, 
and therefore not exposed as a member of the `...t` rest param.
 The only solution was to make them enumerable. 
Not ideal, but better than the confusing error that previously occurred.

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -214,7 +214,7 @@ Test.prototype._publicApi = function () {
 	]
 		.forEach(function (name) {
 			Object.defineProperty(api, name, {
-				enumerable: !/^_/.test(name),
+				enumerable: name === 'end' ? self.metadata.callback : true,
 				get: function () {
 					return self[name];
 				}
--- a/test/fixture/destructuring-public-api.js
+++ b/test/fixture/destructuring-public-api.js
@@ -1,0 +1,17 @@
+import test from '../../';
+
+test.beforeEach(t => t.context = 'bar');
+
+test.cb('callback mode', ({context: foo, ... t}) => {
+	t.is(foo, 'bar');
+	t.end();
+});
+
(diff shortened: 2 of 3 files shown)


sindresorhus:
> The only solution was to make them enumerable. 

Why do they need to be enumerable?

vdemedes:
I also don't understand why would anyone want destructuring.

jamestalmage:
They don't get copied in the `rest` param

```js
test(({context: foo, ...t}) => {
  t.is(...); // throws "t._expr is not a function"
});

jamestalmage:
> I also don't understand why would anyone want destructuring.

It's available, people will use it. Anybody who hasn't hacked on power-assert is unlikely to guess that error message is related to their use of destructuring.

Honestly, if #304 happened, I would likely never destructure `t`. Still, I think given the potential confusion, we should still guard against it.

vdemedes:
> Still, I think given the potential confusion, we should still guard against it.

Fair point.

sindresorhus:
> t._capt, t._expr are not enumerable, 

@jamestalmage I remember @twada saying it might be possible to remove them at some point. Is there any `power-assert` issue I can follow about that?

jamestalmage:
> Is there any power-assert issue I can follow about that?

Not that I am aware of. I'll open one if I can't find it.

twada:
@sindresorhus @jamestalmage [self-assigned](https://github.com/twada/empower-core/issues/1)  ;)

Issue #272, opened 2015-11-25
Add filename to the names of failing tests


alubbe:
Due to the test reports of successful runs using the filenames, I've DRYed up my test names, e.g.,
```
 ✔ updateUser › should reject malformed requests (109ms)
```
However, when an assertion fails, it gives only the test name, making it a bit harder to quickly see where the failure happened:
```
  1. should reject malformed requests 
```
What do you think about printing this instead:
```
  1. updateUser › should reject malformed requests
```

alubbe:
Awesome, thanks!
Am 04.12.2015 10:44 vorm. schrieb "Sindre Sorhus" <notifications@github.com

PR #303, opened 2015-12-04
Adds filename to failed test title.

mcmathja:
Addresses #272 by sending the logger the tests stored in *api.errors*, which already include the formatted file prefix in their titles, instead of the whole list included in *api.tests*.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -91,7 +91,7 @@ api.run()
 		log.write();
 
 		if (api.failCount > 0) {
-			log.errors(api.tests);
+			log.errors(api.errors);
 		}
 
 		process.stdout.write('');
--- a/lib/logger.js
+++ b/lib/logger.js
@@ -52,18 +52,14 @@ x.test = function (props) {
 	log.success(props.title + dur);
 };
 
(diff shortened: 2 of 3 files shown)


jamestalmage:
@mcmathja This looks good. Can you add a test or two covering this? It will need to go in `cli.js` (model your tests after those). To validate that it fixes #272 you will need to call two test fixtures in a single test (there are a couple examples of that in there). This is because we do not prefix the filename if you only run a single file.

If you add tests, I'm :+1: for merging.

mcmathja:
Hmm, test passes without issue on my box. I'll look at it again in the morning.

mcmathja:
Today I learned AppVeyor doesn't properly handle unicode in console output. =)

jamestalmage:
AppVeyor console output is my nemesis.

vdemedes:
@mcmathja Great work! Clean & simple, I like it. Thank you for your contribution and welcome to the AVA project!

@sindresorhus all good here, so that we can merge?

sindresorhus:
Looks good. Thank you @mcmathja :)

PR #317, opened 2015-12-08
Renamed paths to relative instead of global. Fixes #207

### LINKED ISSUE #207: babel.Transformer is not a function
error
```
node: v4.2.2
npm: v1.4.9


C:\Users\mnzt\~code\ava-test\node_modules\ava\node_modules\babel-plugin-espower\create.js:6
    return new babel.Transformer('babel-plugin-espower', createEspowerVisitor(babel, options));
           ^

TypeError: babel.Transformer is not a function
    at createEspowerPlugin (C:\MEDIA\~code\Binaries\node_modules\ava\node_modules\babel-plugin-espower\create.js:6:12)
    at Object.<anonymous> (C:\MEDIA\~code\Binaries\node_modules\ava\lib\babel.js:21:3)
    at Module._compile (module.js:435:26)
    at Object.Module._extensions..js (module.js:442:10)
    at Function.Module._load (module.js:311:12)
    at Function.Module.runMain (module.js:467:10)
    at startup (node.js:136:18)
    at node.js:963:3
```
package.json
```
  "devDependencies": {
    "ava": "^0.4.2"
  },
```

mnzt:


mnzt:
Closes #207. 

Hopefully to fix the CI break, ensuring all packages are installed beforehand will suffice. Open to criticism.

sindresorhus:
Can you explain your changes? We're not interested in changing the paths like that, which I also really doubt is the solution to anything. Would like to hear your thinking behind it though.

mnzt:
This, at least for me, allowed me to run Babel 6 in my project and still be able to execute my tests via `ava test.js`. While not the most engineered nor perfect solution, it is one that worked for me, and from my view point does not impede Ava in anyway.

jamestalmage:
@mnzt Nice thought. But in practice, this is not a good idea. NPM 3 and later automatically dedupes and maximally flattens your dependency tree. In simpler terms:

instead of:

```
projectDir
    node_modules
        ava
            node_modules
                core_assert
```

you end up with:

```
projectDir
    node_modules
        ava
        core_assert
```

With `npm@3` the only reason you end up with nested `node_modules` folders is if two dependencies both want the same dependency but have incompatible `semver` ranges. (i.e. `foo-dep` wants `core-assert@2.x` and `bar-dep` wants `core-assert@3.x`).

Your changes here are broken by the new behavior.

mnzt:
:boom: Oops! I wasn't aware of these changes to NPM3, time for me to update. Thank you.

PR #319, opened 2015-12-09
process is forked with a single JSON argument

### LINKED ISSUE #318: Improve passing data to forked processes
From: https://github.com/sindresorhus/ava/pull/296#discussion_r47034570

Currently we just proxy the CLI args [manually](https://github.com/sindresorhus/ava/blob/ff5b8dfa5cc9c49a0919c9cad0896cb9c3e971ee/api.js#L45-L57). We should instead just JSON.stringify all the data we want to pass and pass it as the only argument.

jokeyrhyme:
This PR attempts to fix #318 .

Initially, the JSON string is converted back into `process.argv` values in lib/babel.js. I'm happy to go further and actually try to eliminate `--fail-fast` and `--serial` altogether in the child process, but I figured it was worth sharing what I had so far.

### DIFF
--- a/api.js
+++ b/api.js
@@ -6,7 +6,7 @@ var fs = require('fs');
 var flatten = require('arr-flatten');
 var Promise = require('bluebird');
 var figures = require('figures');
-var assign = require('object-assign');
+var objectAssign = require('object-assign');
 var globby = require('globby');
 var chalk = require('chalk');
 var fork = require('./lib/fork');
@@ -18,7 +18,8 @@ function Api(files, options) {
 
 	EventEmitter.call(this);
 
-	assign(this, options);
+	objectAssign(this, options);
+	this.options = options;
 
 	this.rejectionCount = 0;
 	this.exceptionCount = 0;
@@ -40,23 +41,7 @@ util.inherits(Api, EventEmitter);
 module.exports = Api;
 
 Api.prototype._runFile = function (file) {
-	var args = [file];
-
-	if (this.failFast) {
-		args.push('--fail-fast');
-	}
-
-	if (this.serial) {
-		args.push('--serial');
-	}
-
-	// Forward the `time-require` `--sorted` flag.
-	// Intended for internal optimization tests only.
-	if (this._sorted) {
-		args.push('--sorted');
-	}
-
-	return fork(args)
+	return fork(file, this.options)
 		.on('stats', this._handleStats)
 		.on('test', this._handleTest)
 		.on('unhandledRejections', this._handleRejections)
@@ -162,7 +147,7 @@ Api.prototype.run = function () {
 					if (++statsCount === self.fileCount) {
 						self.emit('ready');
 
-						var method = self.serial ? 'mapSeries' : 'map';
+						var method = self.options.serial ? 'mapSeries' : 'map';
 
 						deferred.resolve(Promise[method](files, function (file, index) {
 							return tests[index].run();
--- a/index.js
+++ b/index.js
@@ -1,6 +1,5 @@
 'use strict';
 var relative = require('path').relative;
-var hasFlag = require('has-flag');
 var chalk = require('chalk');
 var serializeError = require('serialize-error');
 var globals = require('./lib/globals');
@@ -8,10 +7,12 @@ var Runner = require('./lib/runner');
 var send = require('./lib/send');
 var log = require('./lib/logger');
 
-var runner = new Runner();
-
 // note that test files have require('ava')
 require('./lib/babel').avaRequired = true;
+
+var opts = JSON.parse(process.argv[2]);
+
+var runner = new Runner(opts);
 
 // check if the test is being run without AVA cli
 var isForked = typeof process.send === 'function';
@@ -45,7 +46,7 @@ function test(props) {
 
 	send('test', props);
 
-	if (props.error && hasFlag('fail-fast')) {
+	if (props.error && opts.failFast) {
 		isFailed = true;
 		exit();
 	}
--- a/lib/babel.js
+++ b/lib/babel.js
@@ -2,7 +2,14 @@
 
 var debug = require('debug')('ava');
 
+var opts = JSON.parse(process.argv[2]);
+
 if (debug.enabled) {
+	// Forward the `time-require` `--sorted` flag.
+	// Intended for internal optimization tests only.
+	if (opts._sorted) {
+		process.argv.push('--sorted');
+	}
 	require('time-require');
 }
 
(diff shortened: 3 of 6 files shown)


api.js
```diff
@@ -40,23 +40,17 @@ util.inherits(Api, EventEmitter);
 module.exports = Api;
 
 Api.prototype._runFile = function (file) {
-	var args = [file];
+	var opts = {
+		file: file,
+		failFast: this.failFast,
+		serial: this.serial,
```
sindresorhus: I think we should do `this.options = options` in the `Api` constructor so we can just pass the whole options object here instead of explicitly specify each option.

lib/babel.js
```diff
@@ -2,7 +2,21 @@
...
 var debug = require('debug')('ava');
 
+// unpack JSON argument for `hasFlag` compatibility
+var opts = JSON.parse(process.argv[2]);
+if (opts.failFast) {
+	process.argv.push('--fail-fast');
+}
+if (opts.serial) {
+	process.argv.push('--serial');
+}
```
sindresorhus: The point of this change is not having to do this.

api.js
```diff
@@ -40,23 +40,17 @@ util.inherits(Api, EventEmitter);
...
-	if (this._sorted) {
-		args.push('--sorted');
-	}
+		// Forward the `time-require` `--sorted` flag.
+		// Intended for internal optimization tests only.
+		sorted: this._sorted
+	};
 
-	return fork(args)
+	return fork(opts)
```
sindresorhus: I think the first argument should be the `file` and second should be `opts`.

jokeyrhyme:
Okay, I've removed the direct dependency on "has-flag" (although istanbul still pulls it in), and the child process just uses the parsed JSON options instead of using `hasFlag`.

sindresorhus:
See: https://github.com/sindresorhus/ava/pull/319#discussion_r47169248

jokeyrhyme:
Okay, fork now has the signature: `fork(file: String, opts?: Object) => Promise`, which means I was able to undo the changes I had to make to tests previously.

api.js
```diff
@@ -18,7 +17,7 @@ function Api(files, options) {
 
 	EventEmitter.call(this);
 
-	assign(this, options);
```
sindresorhus: You can't just remove this. The options are used directly on `this` in the code and needs to be changed to use `this.options.x`.

index.js
```diff
@@ -1,17 +1,17 @@
...
 var send = require('./lib/send');
 var log = require('./lib/logger');
 
-var runner = new Runner();
-
 // note that test files have require('ava')
-require('./lib/babel').avaRequired = true;
+babel.avaRequired = true;
+
+var runner = new Runner(babel.opts);
```
sindresorhus: It's weird that `babel` exports the options. I think we should just get them directly here too.

jokeyrhyme:
I very thoroughly checked for "serial" and "failFast" property access before removing that `assign(this, options);` in api.js, but I've reverted as you suggested.

sindresorhus:
@jokeyrhyme It's for example used here: https://github.com/jokeyrhyme/ava/blob/af4e70377a2b0143005c5fad52913821fcbdfbeb/api.js#L150

I would like to remove it though as I feel it's pretty dirty to have random option mixed into `this`. @vdemedes You ok with that? Any other places it's used?

jokeyrhyme:
It's weird that the tests were still passing when I dropped the `assign(this, options)` in api.js...

lib/babel.js
```diff
@@ -3,6 +3,11 @@
 var debug = require('debug')('ava');
 
 if (debug.enabled) {
+	// Forward the `time-require` `--sorted` flag.
+	// Intended for internal optimization tests only.
+	if (exports.opts._sorted) {
```
sindresorhus: This is outdated

lib/fork.js
```diff
@@ -1,23 +1,22 @@
 'use strict';
 var childProcess = require('child_process');
 var path = require('path');
+var assign = require('object-assign');
```
sindresorhus:
```
var objectAssign = require('object-assign');
```

lib/fork.js
```diff
@@ -1,23 +1,22 @@
...
 
-module.exports = function (args) {
-	if (!Array.isArray(args)) {
-		args = [args];
-	}
-
+module.exports = function (file, opts) {
 	var filepath = path.join(__dirname, 'babel.js');
-	var file = args[0];
+	opts = assign({file: file}, opts);
```
sindresorhus: Doesn't really matter, but:

```js
opts = assign(opts, {file: file});
```

lib/babel.js
jokeyrhyme: Should I completely drop the "time-require" module from package.json dependencies and the related `--sorted` stuff?

lib/runner.js
```diff
@@ -24,6 +23,8 @@ function Runner(opts) {
 
 	EventEmitter.call(this);
 
+	this.opts = opts || {};
```
sindresorhus:
```js
this.options = opts || {};
```

lib/babel.js
sindresorhus: No, meant that `exports.opts._sorted` is outdated with the latest PR changes.

api.js
jokeyrhyme: `assign(this, options);` has been replaced with `objectAssign(this, options);`

sindresorhus:
Very nice work @jokeyrhyme :) Can you squash it down to one commit?

sindresorhus:
LGTM

jokeyrhyme:
Done. :)

jokeyrhyme:
@sindresorhus I think your "LGTM" came before my squash :)

jamestalmage:
LGTM

vdemedes:
@sindresorhus This (https://github.com/jokeyrhyme/ava/blob/af4e70377a2b0143005c5fad52913821fcbdfbeb/api.js#L150) is required for serial execution of forks (#282).

jokeyrhyme:
@vdemedes in my squashed commit, I include moving that over to `this.options.serial`:
https://github.com/jokeyrhyme/ava/blob/eeeec706c285ed30889c6bf602f7797f31720270/api.js#L150

vdemedes:
@jokeyrhyme Oh, my bad. Everything is fine then!

vdemedes:
LGTM

vdemedes:
@jokeyrhyme Ron, thank you for your valuable contribution to AVA, we really appreciate it! We'd love to see your new PRs soon!

sindresorhus:
Thank you! :D

jokeyrhyme:
:tada: :smiley_cat:

PR #320, opened 2015-12-10
add contains assert

brunoqueiros:
A new assert to check if a string contains certain string.

t.contains(value, expected, [message]);

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -120,4 +120,8 @@ x.ifError = x.error = function (err, msg) {
 	test(!err, create(err, 'Error', '!==', msg, x.ifError));
 };
 
+x.contains = function (val, expected, msg) {
+	test(val.indexOf(expected) > -1, create(val, expected, 'indexOf', msg, x.contains));
+};
+
 require('./enhance-assert')(x);
--- a/readme.md
+++ b/readme.md
@@ -570,6 +570,10 @@ Assert that `regex` matches `contents`.
 
 Assert that `error` is falsy.
 
+### .contains(value, expected, [message])
+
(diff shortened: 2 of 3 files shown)


sindresorhus:
Hmm, not sure about this. Now that we have [`power-assert` support](https://github.com/sindresorhus/ava#enhanced-asserts) it doesn't really make sense to add a bunch more assertion methods as you can just use `t.true()`. In this case it would be `t.true(foo.includes('bar'));`, although `[].includes` is [currently not working](https://github.com/sindresorhus/ava/issues/263), but will be fixed.

brunoqueiros:
I think that is easier to have more assertions because the test file will be cleaner and more easy to maintaining. I think there should have both, that way the user can choose what use.

sindresorhus:
```js
t.contains(foo, 'bar');
// vs
t.ok(foo.includes('bar'));
```

// @twada @uiureo

shinnn:
Contrary to the PR author's opinion, `t.contains` won't simplify the test file, I think.

There are a lot of JavaScript objects that have `indexOf` method (array, string, Node's buffer, ...), but this `t.contains` function doesn't check its argument type. So whenever we use `t.contains` we should also check the object type.

brunoqueiros:
It's not even necessary, your arguments are right. I'll close the PR.

twada:
I agree with @sindresorhus and @shinnn.

I've developed [power-assert](https://github.com/power-assert-js/power-assert) to prevent tests from bloating assertion methods. Less is more.

I know it's a bit opinionated. But AVA is opinionated too ;)

PR #321, opened 2015-12-10
don't use local Babel directly

sindresorhus:
Instead depend on npm to dedupe. This is better as we will no longer fail if the user has Babel 6 and we Babel 5, and npm just won't dedupe. Same situation in the future when we're on Babel 6, but the user is still on Babel 5.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -363,16 +363,7 @@ test(t => {
 });
 ```
 
-You can also use your own local Babel version:
-
-```json
-{
-	"devDependencies": {
-		"ava": "^0.6.0",
-		"babel-core": "^5.8.0"
-	}
-}
-```
+We don't yet [support Babel 6](https://github.com/sindresorhus/ava/pull/221), but you can use any Babel version in your project. We use our own bundled one.
 
 #### Transpiling Imported Modules
 
--- a/lib/babel.js
+++ b/lib/babel.js
@@ -33,22 +33,12 @@ sourceMapSupport.install({
 var createEspowerPlugin = require('babel-plugin-espower/create');
 var requireFromString = require('require-from-string');
 var loudRejection = require('loud-rejection/api')(process);
-var resolveCwd = require('resolve-cwd');
 var hasGenerator = require('has-generator');
 var serializeError = require('serialize-error');
+var babel = require('babel-core');
 var send = require('./send');
 
 var testPath = opts.file;
-
-// include local babel and fallback to ava's babel
-var babel;
-
-try {
-	var localBabel = resolveCwd('babel-core') || resolveCwd('babel');
-	babel = require(localBabel);
-} catch (err) {
-	babel = require('babel-core');
-}
 
 // initialize power-assert
 var powerAssert = createEspowerPlugin(babel, {


readme.md
```diff
@@ -363,13 +363,15 @@ test(t => {
 });
 ```
 
-You can also use your own local Babel version:
+To use your own local Babel version, specify the version you want in your package.json, and ensure you're using npm@3 (or run `$ npm dedupe`), so the dependency will be deduped.
```
novemberborn: I assume you're advocating deduping so that if the code-under-test requires `babel-core` it doesn't load two versions? If so, is that really worth mentioning?

readme.md
```diff
@@ -363,13 +363,15 @@ test(t => {
...
+
+*We don't yet [support Babel 6](https://github.com/sindresorhus/ava/pull/221), but you can still use it in your own project.*
 
 ```json
 {
 	"devDependencies": {
 		"ava": "^0.6.0",
-		"babel-core": "^5.8.0"
+		"babel-core": "5.8.11"
```
novemberborn: Should the example use `^6` instead?

readme.md
sindresorhus: No, I'm advocating dedupe so the user is able to set an explicit Babel 5 version that we will also use. Since we depend on a loose Babel 5 version, the user can set an exact version and it will be deduped to the user's choice.

readme.md
sindresorhus: @novemberborn That's not really the point of that example. It could use either, but by using an exact version of Babel 5, I'm illustrating my above point of it being deduped.

readme.md
novemberborn: Ah, thanks for elaborating. I did not find that obvious from the documentation though. Same with the version pinning below which follows a note on Babel 6 support.

sindresorhus: Good feedback. I'll try to make that clearer. Any suggestions?

novemberborn: > Since we depend on a loose Babel 5 version, the user can set an exact version and it will be deduped to the user's choice.

This just doesn't seem very relevant to me. AVA uses Babel, fine. The version may change, even in patch releases, as long as the test behavior remains the same. Why should a user be concerned with setting an exact version in the project in order to influence the version AVA uses?

What I find more important as a user is knowing that I can install whatever Babel version I like and it doesn't break AVA.

readme.md
jamestalmage: Either bump this to show `^6`, or move this line

> We don't yet support Babel 6...

Down below the example code and preface it with `Note:`

> ```js
> {
>   "devDependencies": {
>    ...
> } 
> ```
>   
> *Note: We don't yet support Babel 6, but you can still use it in your own project*

I personally think it would be better to just give the Babel 6 example. In essence, all we are doing right now is explaining to people how semver and dedupe works. I think it undersells the best part of this change: that you can now use Babel 6.

novemberborn: > I think it undersells the best part of this change: that you can now use Babel 6.

:100:

jamestalmage:
@sindresorhus - your call on switching the example to Babel 6, otherwise LGTM

sindresorhus:
Ok, updated. I went with just removing the example as it didn't give much.

jamestalmage:
LGTM. 
Do we have to reapprove new commits like this, or is my original LGTM still valid?

sindresorhus:
@jamestalmage Not sure. @bradrydzewski ?

bradrydzewski:
@sindresorhus @jamestalmage the original LGTM will suffice. I've gotten a request to support `-1` and `+1` to handle these sort of situations, if that is of interest to you guys

jamestalmage:
Like -1 to change your vote?

bradrydzewski:
@jamestalmage yes, exactly. Open to other suggestions as well :smile:

jamestalmage:
My only other thought would be a conditional approval, maybe using the markdown checklist syntax.

LGTM if:
 - [ ] fix this
 - [ ] change that

Other collaborators can then check those off as the submitter handles the feedback. I think that would be especially helpful on the more heavily debated PRs with wall-of-text posts, where feedback can be lost amongst the noise.

vdemedes:
LGTM

PR #322, opened 2015-12-10
Show total execution time

brunoqueiros:
Hello,

This PR shows the total execution time of all tests. It's a nice information to show.

image

### DIFF
--- a/api.js
+++ b/api.js
@@ -24,6 +24,7 @@ function Api(files, options) {
 	this.failCount = 0;
 	this.fileCount = 0;
 	this.testCount = 0;
+	this.startTime = new Date();
 	this.errors = [];
 	this.stats = [];
 	this.tests = [];
--- a/api.js
+++ b/api.js
@@ -9,6 +9,7 @@ var figures = require('figures');
 var globby = require('globby');
 var chalk = require('chalk');
 var fork = require('./lib/fork');
+var globals = require('./lib/globals');
 
 function Api(files, options) {
 	if (!(this instanceof Api)) {
@@ -24,7 +25,7 @@ function Api(files, options) {
 	this.failCount = 0;
 	this.fileCount = 0;
 	this.testCount = 0;
-	this.startTime = new Date();
+	this.startTime = globals.now();
 	this.errors = [];
 	this.stats = [];
 	this.tests = [];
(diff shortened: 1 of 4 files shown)


lib/logger.js
```diff
@@ -64,11 +64,14 @@ x.errors = function (tests) {
 	});
 };
 
-x.report = function (passed, failed, unhandled, uncaught) {
+x.report = function (passed, failed, unhandled, uncaught, startTime) {
+	var duration = new Date() - startTime;
```
vdemedes: Logger should not calculate total time, it should only write output.

api.js
```diff
@@ -24,6 +24,7 @@ function Api(files, options) {
 	this.failCount = 0;
 	this.fileCount = 0;
 	this.testCount = 0;
+	this.startTime = new Date();
```
vdemedes: `lib/globals.js` and its `now()` method should be used instead of `new Date()`.

test/cli.js
```diff
@@ -69,7 +70,7 @@ test('don\'t display test title if there is only one anonymous test', function (
 
 	execCli(['fixture/es2015.js'], function (err, stdout, stderr) {
 		t.ifError(err);
-		t.is(stderr.trim(), '1 test passed');
+		t.true(stderr.trim().includes('1 test passed'));
```
vdemedes: This functionality should be tested by using API (`api.js`), without executing AVA's CLI.

sindresorhus:
> It's a nice information to show.

We don't add things on the merit of it being "nice to have". Does it have any actual use-case for every test run? I think knowing the test time can be useful once in awhile to make sure your tests aren't getting too slow, but then you can just do `$ time ava`. I don't want to see the time every time I run a test and I don't really see how it's useful either.

brunoqueiros:
In my opinion some users want to see the total execution time, if you don't want show the time for all users could be a good solution have a flag, that way only users which set the flag will see.

vdemedes:
@brunoqueiros If users will need to use a flag (for example, `--show-time`), they might as well just:

```js
$ time ava
```

which will also show total execution time.

sindresorhus:
> In my opinion some users want to see the total execution time

Why?

brunoqueiros:
OK, if you doesn't see that is a useful feature I'll close the PR.

Issue #315, opened 2015-12-08
Change weird `false fail false` message.


jamestalmage:
See here: https://github.com/sindresorhus/ava/pull/313#discussion_r46928584

Calling `t.fail()` gives a really odd message `false fail false`. It should be better.

sotojuan:
What would be better? Also I'm trying to look where that message is generated, but can't seem to find it. Seems like this would be an easy fix!

jamestalmage:
```js
import test from 'ava';

test(t => t.fail())
```

sotojuan:
Ah so it's an internal thing.

Is something like this better? Got it working.

```javascript
import test from 'ava';

test(t => t.fail('I failed fast!'));
```

jamestalmage:
> Ah so it's an internal thing.

Not sure what you mean. `false fail false` is the message AVA generates if you use `t.fail()`

> Is something like this better?

I believe `t.fail(message)` should already show the custom user-provided message. The problem is the default message becomes `false fail false` which is really useless.

I think the solution is simply defining a default message [here](https://github.com/sindresorhus/ava/blob/master/lib/assert.js#L33)

```js
x.fail = function (msg) {
    msg = msg || 'some better message';
    test(false, create(false, false, 'fail', msg, x.fail));
};
```

sotojuan:
> false fail false is the message AVA generates if you use t.fail()

That's what I meant! I thought "false fail false" was something someone wrote, not something AVA generated. 

So I guess a better message would just be "Test failed" or "Assertion failed"? Let me know!

jamestalmage:
Maybe: `Test failed via t.fail()` ?

I think it's best the message includes a hint that `t.fail()` was used. A simple "Assertion failed" would be confusing if the test contained multiple other assertions in addition to `t.fail()`

sotojuan:
That sounds good. I'll send the PR in a few minutes.

PR #323, opened 2015-12-11
Change `false fail false` message

sotojuan:
This helps out with #315 by adding a default message to `x.fail()` so that the failing message is always helpful.

If there's a better message to use, let me know! It's a quick change.


jamestalmage:
LGTM

vdemedes:
LGTM

vdemedes:
@sotojuan Thank you for a quick fix!

Issue #308, opened 2015-12-06
Stacktrace for syntax errors incorrectly points to the test file


ecowden:
If I have a function under test with a syntax error like:

```
// demo.js
module.exports = function () { // missing closing brace
```

...and a test like:

```
// demoTest.js
const test = require('ava')
const demo = require('./demo')

test(async function (t) {
  t.ok(true)
})
```

Then the syntax error is reported pointing to the `require` statement in the test file instead of the file with the error:

```
Uncaught Exception:  /Users/cowdene/src/avaIssue/demoTest.js
  SyntaxError: Unexpected token )
    at Object.<anonymous> (/Users/cowdene/src/avaIssue/demoTest.js:2:14)
```

This is pretty confusing and is especially tricky for newer Node developers.

novemberborn:
Either of these report the error:

```js
// requireDemo.js
const demo = require('./demo')
```

```js
// requireDemoFromString.js
var requireFromString = require('require-from-string');
requireFromString("const demo = require('./demo')", "hi.js")
```

However `lib/babel` uses `process.on('uncaughtException')` to catch the error, serialize it, and send it to the cli. Here's where the information gets lost.

I think Node runs [this code](https://github.com/nodejs/node/blob/322b36c0a163bc723a6d612d740f98775e0750f6/src/node.cc#L1382:L1388) when writing the error to `stderr` if a process crashes. This can somehow access the original source file and whatnot. Not sure how to get access that from JS.

jamestalmage:
Related: https://github.com/nodejs/node-v0.x-archive/issues/3452

jamestalmage:
It's impossible to fix this in a clean way in JS, the V8 related issues are here: 

https://code.google.com/p/v8/issues/detail?id=1914
https://code.google.com/p/v8/issues/detail?id=1281

The only workaround I can see is to wrap `Module.prototype._compile` to temporarily disable our `uncaughtException` handler. Here is a quick prototype of what I am thinking:

```js
var compileDepth = 0;
process.on('uncaughtException', function (e) {
  if (compileDepth !== 0) {
    throw e;
  } else {
    console.log('error swallowed this will not work');
  }
});

var Module = module.constructor;

var oldCompile = Module.prototype._compile;

Module.prototype._compile = function (code, filename) {
  compileDepth ++;
  oldCompile.apply(this, arguments);
  compileDepth --;
};

var requireFromString = require('require-from-string');

var badUserCode = 'function () {';

[...]

Issue #256, opened 2015-11-22
start using `tap.match` in unit tests


jamestalmage:
We have lots of assertions on regular expressions, but we are using `tap.true(regex.test(...))` and similar. When these fail, the reporter outputs really terrible info ("expected true got false" is the gist).
Using `tap.match` provides more information.

There are a number of PR's open right now, so let's wait to do that cleanup until some of the big ones have been merged.

sindresorhus:
Sure, since we don't have `power-assert` with `tap`.

vdemedes:
Fixed by #331.

PR #325, opened 2015-12-12
Stop hiding source of syntax errors

sotojuan:
This helps out with #308. It does the following things:

* Make the changes noted [here](https://github.com/jamestalmage/ava/commit/b1986249b6b59efba05413be3a7b746c4f2bc27d)
* Move babel hook test to CLI (needed after making change above)
* Add an API test for nonexistent files

However, there's one issue. The `fork.js` test that calls a nonexistent file ("rejects on error and streams output") doesn't seem to get a "file does not exist error" (it just gets "error code 1 returned") and worse, a stack of it shows up in the [test results](http://i.imgur.com/5Pe6Ifq.png). The testing continues fine though, but I am not sure how to contain it or test it.

Originally the test had an `.on('uncaughtException')` clause but after making the `babel.js` changes, it is not called at all so I removed it.

Any suggestions and help is appreciated!

### DIFF
--- a/test/api.js
+++ b/test/api.js
@@ -241,7 +241,7 @@ test('testing nonexistent files rejects', function (t) {
 	api.run()
 		.catch(function (err) {
 			t.ok(err);
-			t.true(/Couldn't find any files to test/.test(err.message));
+			t.match(err.message, /Couldn't find any files to test/);
 		});
 });
 
--- a/test/api.js
+++ b/test/api.js
@@ -233,7 +233,7 @@ test('test file that immediately exits with 0 exit code ', function (t) {
 		});
 });
 
-test('testing nonexisting files rejects', function (t) {
+test('testing nonexistent files rejects', function (t) {
 	t.plan(2);
 
 	var api = new Api([path.join(__dirname, 'fixture/broken.js')]);
--- a/test/api.js
+++ b/test/api.js
@@ -107,22 +107,6 @@ test('change process.cwd() to a test\'s directory', function (t) {
 		});
 });
 
-test('babel require hook only applies to the test file', function (t) {
-	t.plan(3);
-
-	var api = new Api([path.join(__dirname, 'fixture/babel-hook.js')]);
-
-	api.on('error', function (data) {
-		t.is(data.name, 'SyntaxError');
-		t.true(/Unexpected token/.test(data.message));
-	});
-
-	api.run()
-		.catch(function (err) {
-			t.ok(err);
-		});
-});
-
 test('unhandled promises will throw an error', function (t) {
 	t.plan(3);
 
@@ -246,6 +230,18 @@ test('test file that immediately exits with 0 exit code ', function (t) {
 		.catch(function (err) {
 			t.ok(err);
 			t.true(/Test results were not received from/.test(err.message));
+		});
+});
+
+test('testing nonexisting files rejects', function (t) {
+	t.plan(2);
+
+	var api = new Api([path.join(__dirname, 'fixture/broken.js')]);
+
+	api.run()
+		.catch(function (err) {
+			t.ok(err);
+			t.true(/Couldn't find any files to test/.test(err.message));
 		});
 });
 
--- a/test/cli.js
+++ b/test/cli.js
@@ -89,7 +89,7 @@ test('babel require hook only applies to the test file', function (t) {
 
 	execCli('fixture/babel-hook.js', function (err, stdout, stderr) {
 		t.ok(err);
-		t.true(/Unexpected token/.test(stderr));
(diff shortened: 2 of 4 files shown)


test/fork.js
```diff
@@ -38,11 +38,9 @@ test('rejects on error and streams output', function (t) {
...
 	fork(fixture('broken.js'))
 		.run()
-		.on('uncaughtException', function (data) {
-			t.true(/no such file or directory/.test(data.exception.message));
-		})
-		.catch(function () {
-			t.pass();
+		.catch(function (err) {
+			t.true(/exited with a non-zero exit code: 1/.test(err.message));
+			t.ok(err);
```
jamestalmage: Are you talking about this test here? If this test passes, that seems fine to me. We are getting a failure. Users of the CLI or API will get "can't find any files to test", like they have. Anybody using `lib/fork` from outside AVA is already asking for trouble.

We need to investigate why it fails on Node 0.10.

sotojuan: > We need to investigate why it fails on Node 0.10

It seems to fail in the regex test for the error message, so maybe the error message changed from 0.10 to others. I can give it a try with nvm.

> If this test passes, that seems fine to me.

Yep it works and passes but I was just curious about the fact that a stack is printed to the screen during that error. Maybe it's not a big deal.

jamestalmage:
> worse, a stack of it shows up in the test results.

Yep, that is anticipated. Basically due to some V8 issues, it is impossible to extract the filename and/or line number from caught SyntaxErrors in JavaScript. Node includes a hack that will print that information to stderr, but it relies on native C++ code. This is why we are changing the position of where we start listening for `uncaughtException`, it allows the error to be handled natively by Node, so we see the correct filename and line-number. Ideally we wouldn't have to do it this way, but we are picking the lesser of two evils.

sotojuan:
Seems like my suspicion was right. Node 0.10 gives an error code of 8 on no file fount, while other Node versions give an error code of 0, which is why my test was failing. It's a quick regex change luckily.

test/api.js
```diff
@@ -249,6 +233,18 @@ test('test file that immediately exits with 0 exit code ', function (t) {
...
 
+test('testing nonexistent files rejects', function (t) {
+	t.plan(2);
+
+	var api = new Api([path.join(__dirname, 'fixture/broken.js')]);
+
+	api.run()
+		.catch(function (err) {
+			t.ok(err);
+			t.true(/Couldn't find any files to test/.test(err.message));
```
jamestalmage: use taps `t.match(string, regexp)` in all new tests. It gives us better error messages. See #256

test/fork.js
jamestalmage: That is because we have to let Node print it to stderr natively. We could clean up the output of our own tests by allowing `fork` to take `stdio` options like `child_process.fork` does (and pipe them to something else besides stderr for that particular test). That's a future PR, IMO.

test/cli.js
```diff
@@ -84,6 +84,17 @@ test('throwing a named function will report the to the console', function (t) {
 	});
 });
 
+test('babel require hook only applies to the test file', function (t) {
+	t.plan(3);
+
+	execCli('fixture/babel-hook.js', function (err, stdout, stderr) {
+		t.ok(err);
+		t.true(/Unexpected token/.test(stderr));
```
jamestalmage: t.match

jamestalmage:
looks like you missed 1 `t.match`

jamestalmage:
> Node 0.10 gives an error code of 8 on no file fount, while other Node versions give an error code of 0

`0`? Really? I would have expected `1`

jamestalmage:
It [looks like it was 1](https://github.com/sotojuan/ava/commit/67574d98fd6dacfc39f6b60185b173ce9e727978) - that's good. As long as it is non-zero.

sotojuan:
Ah yeah sorry I mistyped. Well either way I got that fixed. The AppVeyor builds all passed—not sure why the check isn't there yet.

jamestalmage:
AppVeyor takes about 20 minutes per commit to complete, you pushed a few commits in rapid succession, so it was way behind. I cancelled all of them. Go ahead and fix that last `t.match` and push again, and we should be good.

jamestalmage:
LGTM

jamestalmage:
Thanks @sotojuan!

sotojuan:
I was playing around with/learning about syncing forks and accidentally pushed, so I guess I'll need your LGTM back @jamestalmage. Sorry about that ~_~.

sindresorhus:
LGTM

sindresorhus:
Great work @sotojuan :) :cake:

vdemedes:
Leonardo says thank you!

PR #326, opened 2015-12-12
TAP support

### LINKED ISSUE #27: TAP support

vdemedes:
This PR fixes #27.

It adds `--tap` flag to a CLI to generate TAP-compatible output, which can then be used with any [3rd-party reporters](https://github.com/substack/tape#pretty-reporters).

```
$ ava --tap example.js | tap-nyan
```

screen shot 2015-12-12 at 2 51 24 pm

Here are the things to be resolved before this PR can be merged:

- [ ] Test tracked assertions in `test/assert.js`
- [ ] [Decide on assertion output](https://github.com/sindresorhus/ava/pull/326#issuecomment-164236043)
- [x] Test TAP output (and probably separate into own module)
- [x] Include unhandled rejections and exceptions into TAP output
- [ ] Support fail-fast mode via `Bail out!`
- [x] Add information on enabling TAP output to Readme

Let's discuss!


lib/assert.js
```diff
@@ -19,62 +17,85 @@ function create(val, expected, operator, msg, fn) {
...
+	}
+
+	enhanceAssert(this);
+
+	this.asserts = [];
+}
+
+module.exports = Assert;
+
+var x = Assert.prototype;
```
vdemedes: Need to remove this.

lib/assert.js
```diff
@@ -19,62 +17,85 @@ function create(val, expected, operator, msg, fn) {
 	};
 }
 
-function test(ok, opts) {
+function Assert() {
```
vdemedes: Need `Assert` class, because assertions need to be tracked per test.

lib/tap.js
```diff
@@ -0,0 +1,58 @@
+'use strict';
+var format = require('util').format;
+
+function getSourceFromStack(stack) {
```
vdemedes: Parses stack trace and extracts original function name, file name and line.

lib/tap.js
```diff
@@ -0,0 +1,58 @@
...
+		output.push(format('not ok %d - %s', ++i, test.error.message));
+		output.push('  ---');
+		output.push('    operator: ' + test.error.operator);
+		output.push('    expected: ' + test.error.expected);
+		output.push('    actual: ' + test.error.actual);
+		output.push('    at: ' + getSourceFromStack(test.error.stack));
+		output.push('  ...');
+	}
+
+	console.log(output.join('\n'));
```
vdemedes: Need to collect output first and then display it all at once to avoid collisions with other concurrent `console.log()`s.

lib/test.js
```diff
@@ -41,6 +43,11 @@ function Test(title, fn) {
 	if (this.title === 'callee$0$0') {
 		this.title = '[anonymous]';
 	}
+
+	this._capt = this.assert._capt;
+	this._expr = this.assert._expr;
+	// Test.prototype._capt = assert._capt;
```
vdemedes: Comments need to be removed.

lib/test.js
```diff
@@ -63,8 +70,8 @@ Test.prototype._setAssertError = function (err) {
 
 // Workaround for power-assert
 // `t` must be capturable for decorated assert output
-Test.prototype._capt = assert._capt;
-Test.prototype._expr = assert._expr;
+// Test.prototype._capt = assert._capt;
```
vdemedes: Comments need to be removed.

test/assert.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 var test = require('tap').test;
-var assert = require('../lib/assert');
+var assert = require('../lib/assert')();
```
vdemedes: Workaround to avoid modifying each test in this file. Will need to create new instance of `Assert` in every test.

sindresorhus:
Looks very promising! :) :cake: 

> Decide on assertion output

What specifically needs to be decided on?

lib/tap.js
sindresorhus: Sounds like the above should be a code comment ;)

lib/assert.js
sindresorhus: :+1:

lib/test.js
```diff
@@ -240,11 +247,11 @@ Test.prototype._publicApi = function () {
 	}
 
 	// Patched assert methods: increase assert count and store errors.
-	Object.keys(assert).forEach(function (el) {
+	Object.keys(Assert.prototype).forEach(function (el) {
 		api.skip[el] = skipFn;
 		api[el] = function () {
 			try {
-				var fn = assert[el].apply(assert, arguments);
+				var fn = self.assert[el].apply(self.assert, arguments);
```
jamestalmage: Do we have to make `assert` a class. Can't we just track assertions here? Demanding the assert class track assertions is going to make it harder to support custom assertions down the road.

Is the issue simply that it is hard to know which argument is the message here?

lib/tap.js
vdemedes: True, I just commented it all here to have some kind of todo list ;)

lib/test.js
vdemedes: > Do we have to make assert a class? Can't we just track assertions here?

I'm open to suggestions!

> Demanding the assert class track assertions is going to make it harder to support custom assertions down the road.

It will, see my comment below.

> Is the issue simply that it is hard to know which argument is the message here?

The issue lies in AVA's feature - concurrency. As you know, tests run concurrently, so we have to track which assertions happened during that particular test. That's why `Assert` class is needed, because that way we can safely get a list of test's assertions.

vdemedes:
> Decide on assertion output
> What specifically needs to be decided on?

Sorry for not making that clear. We need to decide what title do we display for assertions:

```
test something
  ✅ 'yes' === 'no'
  or
  ✅ t.is()
  or
   ✅ t.is('yes', 'no')
```

So, test have titles and assertion have titles too, so we need to decide on the latter.

sindresorhus:
> We need to decide what title do we display for assertions

`'yes' === 'no'` and the `power-assert` output when `t.ok()`/`t.notOk()`. We should also show the assert message when provided.

jamestalmage:
@vdemedes - I know I linked it before somewhere, but here are the docs you need for the [`empower-core`](https://github.com/twada/empower-core#optionsonerror). 

FYI: @twada keeps stuff he still considers "beta" under his GitHub account and only moves stuff to the `power-assert-js` GitHub group when he is committing to a stable, long-term API. That is why you had trouble finding those docs.

As for `powerAssertContext`, there is a good example of it's contents [here](https://github.com/power-assert-js/power-assert-formatter#var-formattedtext--formatterpowerassertcontext). @twada may be able to provide a better link. You will likely need `powerAssertContext.source.content`.

Finally, we are going to need to come up with a solution for assertions like this:

```js
t.same(foo, {
  blah: 'blah',
  bar: 'foo',
  zippo: [1, 2, 3]
});
```

That is probably a more than we want to log, even when compacted to a single line. We can probably do some pretty clever collapsing of the code eventually (i.e. `t.same(foo, {blah: ...})`). We could even change color to light gray for the parts we collapse.

For the first pass, maybe we don't use the `power-assert` source (or only if `source.length` is less than some arbitrary number).

twada:
For multiline assertions, power-assert normalizes the assertion lines to single line to render diagrams so powerAssertContext.source.content is always in a single line, no matter how much lines original assertion have.

jamestalmage:
> powerAssertContext.source.content is always in a form of single line

Yes, my point was that even that will end up verbose given a large enough object declaration. I'm sure our eventual solution will share code with SuccinctReporter.

lib/test.js
```diff
@@ -20,13 +21,14 @@ function Test(title, fn) {
 		title = null;
 	}
 
-	assert.is(typeof fn, 'function', 'you must provide a callback');
+	assert.equal(typeof fn, 'function', 'you must provide a callback');
```
sindresorhus: nitpick, but `equal` => `strictEqual`

vdemedes:
@jamestalmage So I tried to use latest changes to assertions and turned out it's not actually usable for tracking assertions. I inserted `console.log(event)` at the beginning of `onAssertionEvent` and it shows no useful information (in case there was no error):

```js
{ originalMessage: undefined,
  enhanced: true,
  args: [ true, undefined ],
  assertionThrew: false,
  returnValue: undefined }
```

`event.powerAssertContext` is `undefined`.

The assertion is a simple `t.true(true)`. Am I missing something?

jamestalmage:
I looked into this a bit.
Turns out, `powerAssertContext` only attaches when the code is sufficiently complex to warrant it:

These will **not** attach a powerAssertContext:

```js
t.ok(true);
t.is('a', 'a');
```

These **will**:

```js
var a = 'a';
t.ok(a);
t.is(a, 'a');
t.is({a: 'a'}.a, 'a');
```

It should still be possible to build something close to the original source using `args`.

vdemedes:
> It should still be possible to build something close to the original source using args.

I don't think this will have good consequences, we need a stable solution. Otherwise we'll end up with a babel-like stream of issues.

sindresorhus:
Can we defer the assert title complications for later and focus on just getting some TAP output in this PR? We can just go for naive assert titles for now.

jprichardson:
> Can we defer the assert title complications for later and focus on just getting some TAP output in this PR? We can just go for naive assert titles for now.

:+1: It's (no tap) the only thing that's preventing me from deep diving into Ava :)

jamestalmage:
> defer the assert title complications for later

I agree. I think it will take us a while to figure that out correctly, and I don't want to hold up TAP support. Unfortunately, `onAssert` currently doesn't even tell you which method was called. @vdemedes - see #340 for a workaround on that (the `identity` function hack, not the whole PR).

vdemedes:
Ok, let's postpone it for now, I'll come up with something simpler. I'm off tomorrow, so I'll post updates the day after.

@jamestalmage will take a look, thanks.

sindresorhus:
Alright. I've opened an issue for it so we don't forget. https://github.com/sindresorhus/ava/issues/341

vdemedes:
PR updated with additions to Readme. Now it can be merged and we'll handle assertion output later.

readme.md
```diff
@@ -456,6 +458,14 @@ test.cb(t => {
 });
 ```
 
+### Optional TAP output
+
+AVA can generate TAP output via `--tap` option for use with any 3rd-party reporters.
+
+```bash
```
sindresorhus: Don't use `bash` as it's not valid shell script with the `$`.

readme.md
```diff
@@ -456,6 +458,14 @@ test.cb(t => {
 });
 ```
 
+### Optional TAP output
+
+AVA can generate TAP output via `--tap` option for use with any 3rd-party reporters.
```
sindresorhus: `3rd-party reporters` => `TAP reporter` and linkify it to `https://github.com/sindresorhus/awesome-tap#reporters`

lib/tap.js
```diff
@@ -0,0 +1,59 @@
...
+
+	output.push('# ' + err.message);
+	output.push(format('not ok %d - %s', ++i, err.message));
+	output.push('  ---');
+	output.push('    name: ' + err.name);
+	output.push('    at: ' + getSourceFromStack(err.stack, 1));
+	output.push('  ...');
+
+	console.log(output.join('\n'));
+};
```
sindresorhus: I think something like this would look nicer:

```js
console.log([
	'# ' + err.message,
	format('not ok %d - %s', ++i, err.message),
	'  ---',
	'    name: ' + err.name,
	'    at: ' + getSourceFromStack(err.stack, 1),
	'  ...'
].join('\n'));
```

lib/tap.js
```diff
@@ -0,0 +1,59 @@
...
+
+exports.finish = function (passCount, failCount, rejectionCount, exceptionCount) {
+	console.log();
+	console.log('1..' + (passCount + failCount));
+	console.log('# tests ' + (passCount + failCount));
+	console.log('# pass ' + passCount);
+	console.log('# fail ' + (failCount + rejectionCount + exceptionCount));
+	console.log();
+};
```
sindresorhus: Same as https://github.com/sindresorhus/ava/pull/326/files#r48151151

sindresorhus:
Can you add some tests?

sindresorhus:
And add `tap` as a keyword to package.json.

readme.md
```diff
@@ -456,6 +458,14 @@ test.cb(t => {
...
+
+```bash
+$ ava --tap | tap-spec
```
sindresorhus: Would be more fun to illustrate with the `nyan` reporter, right? You also need to show installing the reporter. Otherwise we'll end up with support issues on how it doesn't work...

readme.md
vdemedes: Was using it, because sometimes it highlights executable name and symbols like `|`. Will remove it, no probs ;)

readme.md
vdemedes: Should we include a screenshot or that'd be too much?

vdemedes:
@sindresorhus 
> Can you add some tests?

I will!

readme.md
sindresorhus: Yes, but the highlighting is very random.

readme.md
sindresorhus: Actually, drop the install instruction. We can't know if they want to use it locally or globally. It has to be globally installed if used like the above, or locally if used in a npm run script.

sindresorhus: > Should we include a screenshot or that'd be too much?

Sure! Here you go. Make sure to use `` and half its width so it becomes retina.

screen shot 2015-12-21 at 15 58 00

vdemedes:
I'm thinking if we should return generated output from `lib/tap.js`, instead of `console.log()`ing it from there. Would also be easier to test TAP output.

sindresorhus:
:+1:

vdemedes:
PR updated according to all suggestions:

- Improve readme and add a screenshot
- Return TAP output, instead of `console.log()`ing it directly from `lib/tap.js`
- Add tests for TAP
- Add "tap" keyword to package.json
- Clean up `lib/tap.js`

cli.js
```diff
@@ -36,7 +37,8 @@ var cli = meow([
 	'  --init       Add AVA to your project',
 	'  --fail-fast  Stop after first test failure',
 	'  --serial     Run tests serially',
-	'  --require    Module to preload (Can be repeated)',
+	'  --require    Module to preload (Can be repeated)',,
```
sindresorhus: double comma

sindresorhus:
LGTM

sindresorhus:
>  Support fail-fast mode via Bail out!

:arrow_up: Can you open an issue for this?

sindresorhus:
> (and probably separate into own module)

I'm surprised there's no compliance suite any TAP producer could use. Would be so useful. Something like https://github.com/promises-aplus/promises-tests

jamestalmage:
LGTM

I have run it against a number of projects, and the output looks great when everything passes. Something feels off when there are errors though:

`classic` reporter:

screenshot 2015-12-21 18 13 31

tail end of `spec` reporter: 

screenshot 2015-12-21 18 18 01

Neither tells you which test the failure came from (these are not uncaught exceptions, but assertion errors).

jamestalmage:
:shipit: :+1: LGTM

We can improve the messages incrementally.

vdemedes:
@sindresorhus 

> Support fail-fast mode via Bail out!
> :arrow_up: Can you open an issue for this?

I figured there's no need for this one, as we end the tests "safely". And not all reporters handle this stuff.

@jamestalmage Try to use `tap-*` reporters, not reporters for mocha, as it had some out-dated TAP output.

vdemedes:
> I'm surprised there's no compliance suite any TAP producer could use. Would be so useful (hint hint). 

@sindresorhus good idea!

vdemedes:
@sindresorhus @jamestalmage guys, need you to LGTM one more time. It does not want to merge it otherwise.

vdemedes:
@sindresorhus @jamestalmage Never mind, merged!

sindresorhus:
tap-dancing

So much tap! :dancer:

PR #330, opened 2015-12-13
use `empower-core` for better assertion tracking and performance

jamestalmage:
This switches us to the ([much](https://github.com/twada/empower-assert/pull/2#issuecomment-159793918)) lighter weight `empower-core`.

`empower-core` does not render the pretty assertion graphs in the forked processes. So we need to use `empower-formatter`s in the main process to finalize the enhancement. The browserified build of `empower-core` is [90% smaller than empower](https://github.com/twada/empower-assert/pull/2#issuecomment-159793918). However, we won't see much of that benefit until we move Babel transpiling to the main process as well (since the `power-assert` transform has a largely overlapping dependency graph with `empower`).

The immediate benefit is that this provides a very easy way to track assertions and methods which will be beneficial to TAP support.

It currently relies on the initial changes made in https://github.com/twada/empower-core/pull/2, but it will be trivial to update once that is merged and deployed. Currently it points to a dependency in the `@jamestalmage` namespace so as not to reintroduce a git dependency which seems to give some people problems.

### DIFF
--- a/api.js
+++ b/api.js
@@ -70,12 +70,21 @@ Api.prototype._handleStats = function (stats) {
 	this.testCount += stats.testCount;
 };
 
+var formatter = require('./lib/enhance-assert').formatter();
+
 Api.prototype._handleTest = function (test) {
 	test.title = this._prefixTitle(test.file) + test.title;
 
 	var isError = test.error.message;
 
 	if (isError) {
+		if (test.error.powerAssertContext) {
+			var message = formatter(test.error.powerAssertContext);
+			if (test.error.originalMessage) {
+				message = test.error.originalMessage + ' ' + message;
+			}
+			test.error.message = message;
+		}
 		this.errors.push(test);
 	} else {
 		test.error = null;
--- a/api.js
+++ b/api.js
@@ -9,6 +9,7 @@ var figures = require('figures');
 var globby = require('globby');
 var chalk = require('chalk');
 var fork = require('./lib/fork');
+var formatter = require('./lib/enhance-assert').formatter();
 
 function Api(files, options) {
 	if (!(this instanceof Api)) {
@@ -69,8 +70,6 @@ Api.prototype._handleExceptions = function (data) {
 Api.prototype._handleStats = function (stats) {
 	this.testCount += stats.testCount;
 };
-
-var formatter = require('./lib/enhance-assert').formatter();
 
 Api.prototype._handleTest = function (test) {
 	test.title = this._prefixTitle(test.file) + test.title;
--- a/lib/enhance-assert.js
+++ b/lib/enhance-assert.js
@@ -22,7 +22,7 @@ module.exports.NON_ENHANCED_PATTERNS = [
 ];
 
 function enhanceAssert(opts) {
-	var empower = require('@jamestalmage/empower-core');
+	var empower = require('empower-core');
 	var enhanced = empower(
 		opts.assert,
 		{
--- a/lib/enhance-assert.js
+++ b/lib/enhance-assert.js
@@ -1,4 +1,5 @@
 module.exports = enhanceAssert;
+module.exports.formatter = formatter;
 
 module.exports.PATTERNS = [
 	't.ok(value, [message])',
@@ -12,23 +13,40 @@ module.exports.PATTERNS = [
 	't.regexTest(regex, contents, [message])'
 ];
 
-function enhanceAssert(assert) {
-	var empower = require('empower');
+module.exports.NON_ENHANCED_PATTERNS = [
+	't.pass([message])',
+	't.fail([message])',
+	't.throws(fn, [message])',
+	't.doesNotThrow(fn, [message])',
+	't.ifError(error, [message])'
+];
+
+function enhanceAssert(opts) {
+	var empower = require('@jamestalmage/empower-core');
+	var enhanced = empower(
+		opts.assert,
+		{
+			destructive: false,
+			onError: opts.onError,
+			onSuccess: opts.onSuccess,
+			patterns: module.exports.PATTERNS,
+			additionalMethods: module.exports.NON_ENHANCED_PATTERNS
+		}
+	);
+
+	enhanced.AssertionError = opts.assert.AssertionError;
+
+	return enhanced;
+}
+
+function formatter() {
 	var powerAssertFormatter = require('power-assert-formatter');
 	var powerAssertRenderers = require('power-assert-renderers');
 
-	empower(assert,
-		powerAssertFormatter({
-			renderers: [
-				powerAssertRenderers.AssertionRenderer,
-				powerAssertRenderers.SuccinctRenderer
-			]
-		}),
-		{
-			destructive: true,
-			modifyMessageOnRethrow: true,
-			saveContextOnRethrow: false,
-			patterns: module.exports.PATTERNS
-		}
-	);
+	return powerAssertFormatter({
+		renderers: [
+			powerAssertRenderers.AssertionRenderer,
+			powerAssertRenderers.SuccinctRenderer
+		]
+	});
 }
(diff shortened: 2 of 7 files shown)


api.js
```diff
@@ -70,12 +70,21 @@ Api.prototype._handleStats = function (stats) {
 	this.testCount += stats.testCount;
 };
 
+var formatter = require('./lib/enhance-assert').formatter();
```
jamestalmage: crap, meant to move that to the top of the file.

lib/test.js
```diff
@@ -239,31 +233,45 @@ Test.prototype._publicApi = function () {
...
+			self._assert();
+			return null;
+		}
+
+		var fn = observableToPromise(event.returnValue);
+
+		if (isPromise(fn)) {
+			return Promise.resolve(fn)
+				.catch(function (err) {
+					self._setAssertError(err);
```
vdemedes: Will errors here also have `powerAssertContext`?

lib/test.js
```diff
@@ -239,31 +233,45 @@ Test.prototype._publicApi = function () {
 		self._assert();
 	}
 
+	function onAssertionEvent(event) {
+		if (event.type === 'error') {
```
jamestalmage: @vdemedes - This event now happens for every assertion. Pass or fail.

We are changing `event.type === 'error' / 'success'` to `event.assertionThrew === true/false`

see https://github.com/twada/empower-core/pull/2 for further discussion.

`event.originalMessage` will contain the original message the user passed in. (It will be undefined if the user didn't supply one). That will help decide what we display in TAP.

`event.originalMessage` is provided for every assertion, including the "non-enhanced" ones.

lib/enhance-assert.js
```diff
@@ -12,23 +13,40 @@ module.exports.PATTERNS = [
...
 		{
-			destructive: true,
-			modifyMessageOnRethrow: true,
-			saveContextOnRethrow: false,
-			patterns: module.exports.PATTERNS
+			destructive: false,
+			onError: opts.onError,
+			onSuccess: opts.onSuccess,
+			patterns: module.exports.PATTERNS,
+			additionalMethods: module.exports.NON_ENHANCED_PATTERNS
```
vdemedes: What is `additionalMethods` option for? Couldn't find it in empower's docs.

lib/enhance-assert.js
```diff
@@ -12,23 +13,40 @@ module.exports.PATTERNS = [
 	't.regexTest(regex, contents, [message])'
 ];
 
-function enhanceAssert(assert) {
-	var empower = require('empower');
-	var powerAssertFormatter = require('power-assert-formatter');
-	var powerAssertRenderers = require('power-assert-renderers');
+module.exports.NON_ENHANCED_PATTERNS = [
+	't.pass([message])',
+	't.fail([message])',
```
jamestalmage: The non-enhanced patterns are a new feature we are adding to power-assert. There will be no source code transform, but we do wrap the function and automatically extract the optional message parameter for you and pass it as `event.originalMessage`. It will be `undefined` if the user does not pass it in.

lib/enhance-assert.js
jamestalmage: It's new - and it will be end up being called `wrapOnlyPatterns`. This is for functions that it does not make sense to provide the big graph for (i.e., any use of `t.throws` is likely to contain an inline function and be way to complicated for a power-assert renderer to do a good job with). By adding this though, we still wrap those and give you the assertion `events`, so you don't have to handle "enhanced" and "non-enhanced" separately. You get events for both.

lib/test.js
vdemedes: Given your answer below I understood what is `onAssertionEvent` for, but I still don't understand what would the answer to my question be.

Here, if assertion returns promise and that promise rejects, we just `_setAssertError()`, but without `powerAssertContext` like above (lines 237-242).

Or (correct me if I'm wrong) if "in that promise" there is a failed assertion, `onAssertionEvent` will be triggered again with `event.type === 'error'`. In that case, we don't need a `.catch()` block on that promise, don't we?

vdemedes:
I like this idea 👍!

lib/test.js
jamestalmage: I am not sure - it might. I need to test that. My initial thought is no, sense the only assertions that return promises is `throws` and `doesNotThrow` right? Those are `non-enhanced`. 

Definitely worth me double checking though.

vdemedes:
Oh wow, something incredible happened, AppVeyor passed, but Travis failed.

lib/test.js
vdemedes: Yes, only `throws` and `doesNotThrow` might return promises.

jamestalmage:
> Oh wow, something incredible happened, AppVeyor passed, but Travis failed.

My whole world view is shaken.

lib/test.js
jamestalmage: OK, then we are good. Those will never have `powerAssertContext`. Probably would be good to attach `originalMessage` though right?

jamestalmage:
Looks like it's because I used scoped packages, and our travis script doesn't upgrade us to NPM 3. Not sure what we should do here. Should we bump travis to use NPM 3? Then all our CI is using NPM 3, is that a problem?

jamestalmage:
@vdemedes - I just updated this to use the `empower-core` API @twada and I agreed on. It is still going to fail on travis because I am using a scoped package. I think you are pretty safe to start exploring how to use this for your TAP integration though.

jamestalmage:
@vdemedes FYI: [Some documentation](https://github.com/jamestalmage/empower-core/blob/85424dcdf2795461066389221e7fe4a9483d73fb/README.md#optionsonerror) on the recent changes to [`empower-core`](https://github.com/twada/empower-core).

@sindresorhus @vdemedes - How should I proceed w/ regard to `npm 1` not supporting scoped builds? 

 1. Do we just want to add `npm i -g npm@3` as the `before_script`? This would mask any errors that only surface with the NPM 2 directory structure.

 2. Do we want to only upgrade to `npm@2-latest` on Node `0.10` and stick with the default elsewhere?

 3. Just skip it and wait for @twada to ship those changes so we can bump versions? I'm sure @twada will get it done fast, but it would be nice to have a solution in place that allowed us to deploy hotfixes to `npm` in the future.

twada:
@jamestalmage @vdemedes @sindresorhus I just released [empower-core 0.2.0](https://github.com/twada/empower-core/blob/master/CHANGELOG.md#020-2015-12-13).

lib/enhance-assert.js
```diff
@@ -12,23 +13,40 @@ module.exports.PATTERNS = [
...
 	);
+
+	enhanced.AssertionError = opts.assert.AssertionError;
+
+	return enhanced;
+}
+
+function formatter() {
+	var powerAssertFormatter = require('power-assert-formatter');
+	var powerAssertRenderers = require('power-assert-renderers');
```
sindresorhus: Just put the `require` statements directly inside the array. They're inline anyways.

api.js
```diff
@@ -76,6 +77,13 @@ Api.prototype._handleTest = function (test) {
 	var isError = test.error.message;
 
 	if (isError) {
+		if (test.error.powerAssertContext) {
```
sindresorhus: Shouldn't this be handled more low-level, like in `test.js`? Would also be nice to put the logic in `enhanced-assert.js`.

sindresorhus:
:+1: LGTM when the dependency is updated to :arrow_up: and tests are passing.

@twada Does the changes look good to you?

twada:
@sindresorhus Looks good to me when bumped to empower-core 0.2.0

jamestalmage:
Dependency bumped :+1:

jamestalmage:
just rebased, but there weren't any conflicts. Not sure why GH was suggesting I "update branch"

vdemedes:
@jamestalmage I think it was because of lgtm.co.

sindresorhus:
Landed! :)

tumblr_mndj28lzs21qdlh1io1_250

vdemedes:
Great stuff @jamestalmage!

Issue #335, opened 2015-12-15
Fix flaky AppVeyor test


jamestalmage:
We still have [one test](https://github.com/sindresorhus/ava/blob/f03e4ba2ee2f2f4477ccf7a186d018e09edf0123/test/cli.js#L23-L65) that remains [flaky on AppVeyor](https://ci.appveyor.com/project/sindresorhus/ava/build/547/job/hhivjx4grpxlchnh#L1226).

@sotojuan - Is there any reason we can't turn this into a API test?

That one test accounts for something like 9 out of 10 false negatives on AppVeyor.

jamestalmage:
Now it is even failing on Travis: https://travis-ci.org/sindresorhus/ava/jobs/96899493#L996

sotojuan:
Should be trivial to make it an API test since the output it's testing is also inside the `api` object. I'm watching a film right now but I'll give it a go afterwards unless someone wants to fix it now.

jamestalmage:
All you.

sotojuan:
@jamestalmage Do you know if this is fixed? Guess we have to wait and see if people's tests pass.

novemberborn:
Can't recall any recent AppVeyor flakiness. Closing.

PR #336, opened 2015-12-15
remove `empower` from dependencies since it is replaced with `empower…

twada:
This PR removes `empower` from dependencies since it is replaced with new `empower-core`.

I gave LGTM to #330 but I missed this point. So sorry..

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,7 +89,6 @@
     "core-assert": "^0.1.0",
     "debug": "^2.2.0",
     "deeper": "^2.1.0",
-    "empower": "^1.1.0",
     "empower-core": "^0.2.0",
     "figures": "^1.4.0",
     "fn-name": "^2.0.0",


jamestalmage:
LGTM.
I just noticed the same thing.
How much does this affect our download size? Is worth releasing a `0.8.1`?

I am guessing the impact is actually pretty minimal (most of the same dependencies are brought in anyways between `babel-plugin-espower`, `power-assert-formatter`, and `power-assert-renderer`.

jamestalmage:
@twada - don't worry about that test. It is super flaky and not the fault of this change. See #335

twada:
> How much does this affect our download size? Is worth releasing a 0.8.1?
> I am guessing the impact is actually pretty minimal

trivial under `npm@3`, still minimal under `npm@2`.
So it's okay to ship this with other PRs.

sindresorhus:
It can wait until the next release.

twada:
@sindresorhus @jamestalmage Thanks!

PR #337, opened 2015-12-15
Move test title prefixes test to API tests

sotojuan:
This (hopefully) helps out with https://github.com/sindresorhus/ava/issues/335 by moving the test title prefixes test from the CLI tests to the API tests.

### DIFF
--- a/test/cli.js
+++ b/test/cli.js
@@ -1,6 +1,5 @@
 'use strict';
 var childProcess = require('child_process');
-var figures = require('figures');
 var test = require('tap').test;
 
 function execCli(args, cb) {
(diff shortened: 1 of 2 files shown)


test/cli.js
```diff
@@ -20,50 +19,6 @@ function execCli(args, cb) {
...
-			.replace(/[0-9] tests passed/, '')
-			.replace(new RegExp(figures.tick, 'gm'), '')
-			.replace(/^\s+/gm, '')
-			.trim();
-
-		var separator = ' ' + figures.pointerSmall + ' ';
-
-		// expected output
-		var tests = [
-			['async-await', 'async function'].join(separator),
```
jamestalmage: The whole point of this test is that the file names get prefixed. We still need to test for this whole string.

sotojuan:
Hm the `api` object doesn't have both the title and the filename in the same "place". That's something done either by the logger or CLI code. That said, I could test that (for example) the first test has the title `async function` and its index corresponds to the index of `async-await.js` in the `files` array.

sotojuan:
@jamestalmage This is a bit convoluted but it gets closer to what the CLI tests. Otherwise I am not sure how to replicate the CLI test.

jamestalmage:
@sotojuan - See [this](https://github.com/sindresorhus/ava/blob/f03e4ba2ee2f2f4477ccf7a186d018e09edf0123/api.js#L74), it really should be prefixed before you see it.

Maybe try using the event emitter?

```js
api.on('test', function (test) {
  // do the test and slice the array here
});
```

sotojuan:
Wow I feel dumb, I completely forgot about that! Give me a minute and I'll fix it.

jamestalmage:
Honestly, I'm surprised the way you were doing it initially didn't work. Seems like it would be more convenient if the API supported the way you originally had it. And honestly, as I read the code, it should have worked that way. I'm obviously missing something...

sotojuan:
I see. Either way, testing through `.on('test')` works and is much less code and easier to read.

sindresorhus:
> Seems like it would be more convenient if the API supported the way you originally had it. And honestly, as I read the code, it should have worked that way.

@jamestalmage I don't really follow, but if you think anything should be improved with the API, can you open an issue about it?

sindresorhus:
LGTM

PR #344, opened 2015-12-17
Fix fail-fast mode

vdemedes:
This PR fixes fail-fast mode. It was broken because of:

```js
props.error = props.error ? serializeError(props.error) : {};
// and
if (props.error && opts.failFast) {
```

Because of that, the `if` condition is always true and fail-fast mode was not working properly.

### DIFF
--- a/index.js
+++ b/index.js
@@ -37,16 +37,18 @@ function test(props) {
 		return;
 	}
 
+	var hasError = typeof props.error !== 'undefined';
+
 	// don't display anything if it's a passed hook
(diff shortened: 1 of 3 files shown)


jamestalmage:
LGTM. 
Do we need to add a test to protect against future regressions?

sindresorhus:
A test would be nice.

vdemedes:
There already was a test for it, but not precise enough. Updated ;)

sindresorhus:
LGTM

PR #349, opened 2015-12-22
Move Babel transforms to the main process.

jamestalmage:
This moves Babel transforms to the main process, and caches the transpiled results.

It uses `caching-transform` to provide caching. Expensive requires (like `babel`) are placed inside the `factory` function, which `caching-transform` only calls when it determines it needs to transform a file.

The path to each precompiled test is passed in the options object to `fork`:

```js
fork('/path/to/file.js', {
  precompiled: {
    '/path/to/file.js': '/path/to/cached/file.js'
  }
});
```

It uses [`require-precompiled`](https://www.npmjs.com/package/require-precompiled) to install a require hook that pulls the transformed file from the cache.

The caching directory is used by the users `package.json`, and installing in `node_modules/.cache/ava`.
If the cache directory cannot be found (i.e. if they have no `package.json`), or the user provides a `--no-cache` flag, pre-compilation still happens on the main thread - but results are published to a temp directory.

---

Speedup is dramatic. 
Using the `6.0.0` branch of got:

ava master branch

```sh
$ time ava
# npm 2
ava  85.51s user 7.19s system 592% cpu 15.650 total
# npm 3
ava  24.87s user 1.91s system 384% cpu 6.957 total
```

this branch using `--no-cache`

```
$ time ava --no-cache
# npm 2
ava --no-cache  11.19s user 1.05s system 130% cpu 9.379 total
# npm 3
ava --no-cache  7.59s user 0.75s system 134% cpu 6.217 total
```

This branch with caching enabled

```
$ time ava
# npm 2
ava  7.86s user 0.81s system 193% cpu 4.484 total
# npm 3
ava  6.64s user 0.72s system 173% cpu 4.237 total
```

A couple points of interest here: 
- This drastically reduces the penalty for sticking with `npm 2` even on your first run.
- You absolutely should upgrade to `npm 3` for the best performance.


vdemedes:
Rebase went wrong here.

jamestalmage:
> Rebase went wrong here.

It wasn't like that last night.

package.json
```diff
@@ -37,9 +37,9 @@
     "node": ">=0.10.0"
   },
   "scripts": {
-    "test": "xo && tap --coverage --reporter=spec --timeout=150 test/*.js",
-    "test-win": "tap --reporter=spec --timeout=150 test/*.js",
-    "coverage": "tap --coverage-report=lcov"
+    "pretest": "rm -rf ./node_modules/.cache",
```
sindresorhus: Use https://github.com/sindresorhus/del-cli to make it work on Windwos too.

package.json
```diff
@@ -131,8 +137,8 @@
   "config": {
     "nyc": {
       "exclude": [
-        "node_modules[/\\\\]",
-        "test[/\\\\]"
+        "node_modules/**",
+        "test/**"
```
sindresorhus: Both of these are ignored by default in `nyc` 5.

package.json
jamestalmage: OK, I tried `trash-cli` first but it fails if the directory doesn't exist. Was just about to push `rimraf`

sindresorhus: @jamestalmage Re `trash-cli`, that's a bug https://github.com/sindresorhus/trash/pull/43, but no point in having the overhead of moving it to the trash, not something you'd ever care about restoring.

api.js
```diff
@@ -8,8 +8,10 @@ var Promise = require('bluebird');
 var figures = require('figures');
 var globby = require('globby');
 var chalk = require('chalk');
+var assign = require('object-assign');
```
sindresorhus: `assign` => `objectAssign`

lib/test-transformer.js
```diff
@@ -0,0 +1,64 @@
...
+var transform = cachingTransform({
+	factory: factory,
+	cacheDir: cacheDir,
+	ext: '.js',
+	hash: function (code, filename, salt) {
+		var hash = crypto
+			.createHash('md5')
+			.update(code, 'utf8')
+			.update(filename || 'unknown file', 'utf8')
+			.update(salt || '', 'utf8')
```
sindresorhus: Just curious, is there any difference in using `.update()` multiple times rather than just concatenating the strings together before passing them to `.update()`?

package.json
```diff
@@ -121,6 +125,8 @@
   "devDependencies": {
     "coveralls": "^2.11.4",
     "hook-std": "^0.1.0",
+    "nyc": "^5.0.1",
+    "promise-delegates": "^0.1.0",
```
sindresorhus: Doesn't look like this is used anywhere.

lib/test-transformer.js
jamestalmage: I believe there is from a performance standpoint (the VM actually has to create a new concatenated string), but I believe the results would be identical.

jamestalmage:
@sindresorhus I used `rimraf` over `del-cli`, because https://github.com/sindresorhus/del-cli/issues/3

jamestalmage:
Hmm, it seems implementing caching fixed my Windows errors. I previously saw some errors that made me suspicious the problem was in `temp-write`. I'm even more suspicious now.

sindresorhus:
@jamestalmage `del-cli` now supports Node.js 0.10.

jamestalmage:
OK, cool.
I'm heading out for ~4 hours. I will apply all the suggested changes later tonight.

jamestalmage:
OK, This is ready for review.

// @vdemedes @sindresorhus

cli.js
```diff
@@ -77,7 +78,8 @@ if (cli.flags.tap) {
 var api = new Api(cli.input, {
 	failFast: cli.flags.failFast,
 	serial: cli.flags.serial,
-	require: arrify(cli.flags.require)
+	require: arrify(cli.flags.require),
+	noCache: cli.flags.cache === false
```
sindresorhus: `noCache` => `cache`. I don't like double negatives.

lib/caching-precompiler.js
```diff
@@ -0,0 +1,117 @@
...
+	});
+};
+
+CachingPrecompiler.prototype._hash = function (code, filename, salt) {
+	var hash = crypto
+		.createHash('md5')
+		.update(code, 'utf8')
+		.update(filename || 'unknown file', 'utf8')
+		.update(salt || '', 'utf8')
+		.digest('hex');
```
sindresorhus: This could use https://github.com/sindresorhus/md5-hex now.

jamestalmage:
closed in favor of #390

Issue #351, opened 2015-12-22
Performance issues after Babel 6 upgrade


vdemedes:
Since babel 6 upgrade the performance is seriously ****ed up. Using master branch and the following test file:

```js
var test = require('./');

test('wow', t => {
	t.true(true);
});
```

```
$ time ava example.js
ava example.js  4.04s user 0.39s system 101% cpu 4.344 total
```

Running it for like 10th time and getting the same results.

Would be great if someone else confirmed this.

sindresorhus:
Confirmed.

```
~/dev/ava master*
❯ time ava example.js

  ✔ wow

  1 test passed


real	3.46s
user	3.21s
sys	0.32s
```

sindresorhus:
If I run `npm dedupe` and then try again it's faster.

```
❯ time ava example.js

  ✔ wow

  1 test passed


real	1.40s
user	1.30s
sys	0.14s
````

sindresorhus:
Babel 6 is a lot more modular, so that might be where all the overhead comes from. Using npm@3 would also be faster (since it dedupes by default).

vdemedes:
Then we are stuck with initial delay on every test run. Not good.

vdemedes:
I think the solution to this is to cache babel's output. There already was some development in that direction in #189, but left incomplete.

My proposal is to cache transformed code based by md5 of original source code. That way, we don't even need to `require('babel')`, when test files don't change. When md5 changes, load babel and re-transpile source code.

This should improve performance.

jamestalmage:
> using npm 3

That is basically a requirement of Babel 6. `npm dedupe` is the only other option.

> caching

 I just finished implementing much the same for `nyc` with very good results.  It's part of my plan after I land #349

vdemedes:
Seems like #349 will take good amount of time. I can implement cache now and send a quick PR, so that we can do a release with babel 6 within next few days.

sindresorhus:
This can be closed now, right?

vdemedes:
Yep, #352 got it covered.

PR #352, opened 2015-12-22
Cache babel-transpiled code

vdemedes:
Originated from #351, inspired from #189.

This PR was created to avoid retranspiling test files on every run and even `require('babel-core')`. MD5 hash of test file gets calculated and if stored hash equals current hash, AVA loads transpiled code from cache.


vdemedes:
Compared to master branch (see results in #351), now test runs take ~0.7s, given the same test files.

jamestalmage:
I don't think we should put cached files in the home dir. Users will not be able to find and clear them. For `nyc` we settled on `node_modules/.cache/nyc`, the equivalent for use would be `node_modules/.cache/ava`. This means users will clear the cache themselves if they `trash node_modules` and reinstall (even if they don't know the cache is there). If we convince enough modules to adapt that pattern, it also makes `trash node_modules/.cache` a standard way to clear caches.

If you are storing in the home dir, then you need to take extra care salting the hash with the versions of Babel and every plugin used. That would still be a good idea in a local cache, but less critical since it's easier to clear the cache.

sindresorhus:
I like the `node_modules/.cache` convention.

sindresorhus:
// @floatdrop

vdemedes:
Good suggestion! Update is coming.

jamestalmage:
also, I wrote `caching-transform` for `nyc`. My intent was to use it here as well. Similar to `cacha`, but intended specifically for transforms.

vdemedes:
@sindresorhus @jamestalmage PR updated to store cache in `node_modules/.cache/ava`. Do you have other things in mind or we can merge it in?

lib/babel.js
```diff
@@ -70,11 +68,43 @@ if (inputSourceMap) {
 }
 
 // include test file
-var transpiled = babel.transformFileSync(testPath, options);
-sourceMapCache[testPath] = transpiled.map;
-requireFromString(transpiled.code, testPath, {
-	appendPaths: module.paths
-});
+var cachePath = hasha(testPath);
+var hashPath = hasha(testPath) + '_hash';
```
jamestalmage: `cachePath + '_hash'`
instead of invoking hasha twice

lib/babel.js
```diff
@@ -70,11 +68,43 @@ if (inputSourceMap) {
...
+
+var prevHash = cache.getSync(hashPath, {encoding: 'utf8'});
+var currHash = hasha.fromFileSync(testPath);
+
+if (prevHash === currHash) {
+	var cached = JSON.parse(cache.getSync(cachePath));
```
jamestalmage: Not sure it is even necessary to make the filename part of the hash. If two files have identical content, it doesn't matter that they are located at different points on the file system - they will encode identically right?

jamestalmage: actually scratch that. I think `power-assert` may include the filename. In the transform.

@twada?

twada: @jamestalmage Yes. `power-assert` embeds filepath into transformed file. (Is this an answer for you?)

jamestalmage: Yep. Thanks.

sindresorhus:
LGTM. @jamestalmage ?

-

Tried it on `got`. First uncached run 17 sec, and second cached run 7 sec.

jamestalmage:
The cache is currently saved in `./node_modules/ava/node_modules/.cache/ava`, whereas I think it should be in `./node_modules/.cache/ava`. That is how I did it in `nyc`, and my hope was we would all store our cache content in a single directory that was easy to delete.

I think the solution to this is to do:

```js
findUp('package.json', {cwd: path.dirname(filename)})
```

However, there is a chance they are using a global install, and do not have a `package.json` anywhere. I think we just disable caching in that situation, I do not like the idea of users having to hunt down were we are hiding the global cache.

`nyc` is starting with caching disabled by default for the time being. It is enabled with a `--cache` argument. I think we should do the opposite. Enable by default and provide a `--no-cache` argument. That could help eliminate the cache as the source of some difficult problem down the road.

We should probably salt the hash with `babel.version` as well (maybe even with the version of every preset/plugin, but that may be overkill).

vdemedes:
Why don't we just fallback to `xdg-basedir` (with `os-tmpdir` as a backup)? It is a stable approach.

Regarding cache, when do you think users would want to clear it?

jamestalmage:
> Regarding cache, when do you think users would want to clear it?

Because caches cause problems. The one CSE joke I know is about cache invalidation. I can't think of a single piece of software I use that is so absolutely confident in its caching strategy that it does not provide a means for me to clear it. My browser, my IDE, `npm`, all of these give me an option to clear the cache.

As a practical example, the first thing I can think of is `babel-plugin-xxx` publishes a regression and the result of that regression makes its way into the cache. I am sure that is just the tip of the iceberg.

Perhaps, when we can't resolve their node modules folder, we do put it with the global AVA install and just provide a `--clear-cache` flag.

sindresorhus:
> The cache is currently saved in ./node_modules/ava/node_modules/.cache/ava, whereas I think it should be in ./node_modules/.cache/ava.

Not sure about this. It would mean when the user upgrades to a new AVA version, that might change how the cache works, the existing cache is still kept around. Should at least then be `./node_modules/.cache/ava-0.8.0`.

> However, there is a chance they are using a global install, and do not have a package.json anywhere. I think we just disable caching in that situation

:+1: 

> We should probably salt the hash with babel.version

:+1: Actually, we should make the salt based on the versions of the various Babel related packages. Upgrades to any of them should invalidate the cache.

> Why don't we just fallback to xdg-basedir (with os-tmpdir as a backup)? It is a stable approach.

We now have multiple ways the cache can go wrong and when in `xdg-basedir` it's never emptied so cache errors will live forever unless the user does something manually to solve it, which I really don't want.

> Perhaps, when we can't resolve their node modules folder, we do put it with the global AVA install and just provide a --clear-cache flag.

I don't see the point of a flag if we only put the cache in `./node_modules/.cache`. A flag comes with a lot of hidden overhead.

jamestalmage:
>  Should at least then be ./node_modules/.cache/ava-0.8.0.

That makes it (only a little) harder to script removing the directory. I was going to suggest we include the AVA version in the salt. I'm fine with this approach though. I just really want it in their root `node_modules` folder so it is visible.

> Actually, we should make the salt based on the versions of the various Babel related packages

I think we are going to find that difficult and costly. Babel doesn't do that in it's own cache implementation. To actually do that right, you would need to resolve the `package.json` of all 20 (twenty!) dependencies of `babel-preset-es2015`, fetch and parse those files, and hash them all together. Same for the `stage-2` (another 5 dependencies when you walk the full graph), and again for `power-assert`.

> `--clear-cache` ... A flag comes with a lot of hidden overhead.

Agreed. That was offered as a workaround for my concerns with global installs. If we disable `caching` for global installs, then this is moot. I still think `--no-cache` has value, the cache introduces so many moving parts, it would be nice to be able to disable it and eliminate it as the source when troubleshooting a problem with users.

sindresorhus:
> I was going to suggest we include the AVA version in the salt.

:+1: AVA version in the salt would be even better.

> I think we are going to find that difficult and costly. Babel doesn't do that in it's own cache implementation.

Ok, lets not do it then.

lib/babel.js
```diff
@@ -1,8 +1,15 @@
...
+var pkgDir = require('pkg-dir').sync;
 var debug = require('debug')('ava');
+var hasha = require('hasha');
+var cacha = require('cacha');
+var path = require('path');
 
 var opts = JSON.parse(process.argv[2]);
+var testPath = opts.file;
+
+var cache = cacha(path.join(pkgDir(path.dirname(testPath)), 'node_modules', '.cache', 'ava'));
```
vdemedes: So, the cache is being stored in `node_modules/.cache/ava`, but [later down this file](https://github.com/sindresorhus/ava/pull/352/files#diff-c56ba0fef1bb510acefb4680eb31e9c6R151) versions from ava & babel-* packages are being included in the cache key.

That way, when users `rm -rf node_modules/.cache/ava`, they can be sure they removed all the cache. And they won't need to get the AVA version to remove `ava-0.8.0` (for example).

vdemedes:
So, are we good on this? Would really like to get it in before the release, so that users don't experience bad performance.

sindresorhus:
:+1:

PR #354, opened 2015-12-22
Show full stack trace for non-assert errors

### LINKED ISSUE #271: Better runtime errors
If my test throws a runtime error, it gets printed as an `AssertionError`, which swallows the stack trace. For example,
```js
var Promise = require("bluebird");
test("foo", Promise.coroutine(function*(t){
    throw new Error("asd");
}));
```
only prints
```
1. foo
  AssertionError: Promise rejected → Error: asd
```
It would be really helpful to get a stack trace here.

vdemedes:
This PR fixes #271. Now AVA shows full stack trace for non-assertion errors. Before, it wrapped them in `assert.AssertionError`, which swallowed the original stack.

### DIFF
--- a/api.js
+++ b/api.js
@@ -84,8 +84,14 @@ Api.prototype._handleTest = function (test) {
 			if (test.error.originalMessage) {
 				message = test.error.originalMessage + ' ' + message;
 			}
+
 			test.error.message = message;
 		}
+
+		if (test.error.name !== 'AssertionError') {
+			test.error.message = 'failed with "' + test.error.message + '"';
+		}
+
 		this.errors.push(test);
 	} else {
 		test.error = null;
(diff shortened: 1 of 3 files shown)


sindresorhus:
It's pretty unclear what kind of error it is now though:

```
var test = require('./');

test('wow', async t => {
	throw new Error('unicorn');
});
```

```
✖ wow unicorn

  1 test failed

  1. wow
  Error: unicorn
    at _callee$ (/Users/sindresorhus/dev/ava/example.js:4:8)
...
```

It just logs `unicorn` where you usually have `true === false`, etc.

vdemedes:
How about:

```
✖ wow failed with "unicorn"
```

sindresorhus:
:+1:

vdemedes:
Updated output:

screen shot 2015-12-25 at 2 14 34 pm

sindresorhus:
LGTM

PR #359, opened 2015-12-23
Add mini reporter

### LINKED ISSUE #18: Default reporter
I think the default reporter should be super minimal one. It should only log out detailed test info if a test fails, otherwise just `1 test passed`.

It could have a live counter, so while tests are running the line is dynamically updated with `3/12 tests passed`, and when done rewritten to be `12 tests passed`. That way it can be live, but also minimal.

Can use [log-update](https://github.com/sindresorhus/log-update).

vdemedes:
This PR fixes #18.

Replacing default reporter with a mini one:

ava

It also adds `--verbose` option, which enables old full-blown output.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -24,6 +24,7 @@ var meow = require('meow');
 var chalk = require('chalk');
 var Promise = require('bluebird');
 var verboseReporter = require('./lib/reporters/verbose');
+var miniReporter = require('./lib/reporters/mini');
 var tapReporter = require('./lib/reporters/tap');
 var Logger = require('./lib/logger');
 var Api = require('./api');
@@ -41,6 +42,7 @@ var cli = meow([
 	'  --serial     Run tests serially',
 	'  --require    Module to preload (Can be repeated)',
 	'  --tap        Generate TAP output',
+	'  --verbose    Enable verbose output',
 	'',
 	'Examples',
 	'  ava',
@@ -58,6 +60,7 @@ var cli = meow([
 	],
 	boolean: [
 		'fail-fast',
+		'verbose',
 		'serial',
 		'tap'
 	]
@@ -79,9 +82,13 @@ var api = new Api(cli.input, {
 var logger = new Logger();
 logger.api = api;
 
+logger.use(miniReporter());
+
 if (cli.flags.tap) {
 	logger.use(tapReporter());
-} else {
+}
+
+if (cli.flags.verbose) {
 	logger.use(verboseReporter());
 }
 
@@ -96,7 +103,7 @@ api.run()
 		logger.exit(api.failCount > 0 || api.rejectionCount > 0 || api.exceptionCount > 0 ? 1 : 0);
 	})
 	.catch(function (err) {
-		if (err instanceof Error) {
+		if (err.name === 'Error') {
 			console.log('  ' + chalk.red(figures.cross) + ' ' + err.message);
 		} else {
 			console.error(err.stack);
--- a/lib/reporters/verbose.js
+++ b/lib/reporters/verbose.js
@@ -3,17 +3,7 @@ var prettyMs = require('pretty-ms');
 var figures = require('figures');
 var chalk = require('chalk');
 var plur = require('plur');
-
-function beautifyStack(stack) {
-	var re = /(?:^(?! {4}at\b).{6})|(?:\((?:[A-Z]:)?(?:[\\\/](?:(?!node_modules[\\\/]ava[\\\/])[^:\\\/])+)+:\d+:\d+\))/;
-	var found = false;
-
-	return stack.split('\n').filter(function (line) {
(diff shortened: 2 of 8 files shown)


jamestalmage:
If a test is hung, how do I tell which one it is? At least with the old reporter and `--serial` you could basically figure it out. Could you maybe make this one show just the current test title? (I know that would probably only work in serial mode). Maybe it only shows in serial mode and the test has taken longer than 2 seconds (so you aren't blocking to write to stdout every 100ms when the tests are fast)?

vdemedes:
Good point. But we can only show the last test title, we don't have an information regarding currently running test.

jamestalmage:
Oh shoot. That's right. We could maybe pass up a `nextTest`  property in serial mode. We can add that later though.

vdemedes:
> We could maybe pass up a nextTest property in serial mode

I think this is a bit overkill.

sindresorhus:
I think we should keep the old output in a `--verbose` mode. (And `--debug` triggers verbose mode.

cli.js
```diff
@@ -17,6 +17,7 @@ if (debug.enabled) {
 	require('time-require');
 }
 
+var logUpdate = require('log-update');
```
sindresorhus: Can you add `logUpdate.stderr();` to hide the cursor as early as possible. Currently it doesn't hide until the tests actually starts running. Should include a comment on why it's there.

sindresorhus:
:heart: Looks really good. I much prefer this default output.

vdemedes:
I put this on hold until #363 gets merged.

vdemedes:
Feedback welcome!

lib/reporters/verbose.js
```diff
@@ -90,7 +79,7 @@ VerboseReporter.prototype.finish = function () {
 		var i = 0;
 
 		this.api.tests.forEach(function (test) {
-			if (!test.error.message) {
+			if (!test.error || !test.error.message) {
```
sindresorhus: This would be clearer as `if (!(test.error && test.error.message)) {`

cli.js
```diff
@@ -96,7 +103,7 @@ api.run()
 		logger.exit(api.failCount > 0 || api.rejectionCount > 0 || api.exceptionCount > 0 ? 1 : 0);
 	})
 	.catch(function (err) {
-		if (err instanceof Error) {
+		if (err.name === 'Error') {
```
sindresorhus: What's the reasoning behind this change? Why did we previously only show the stack for non-errors?

vdemedes: Previous code was wrong.

`err instanceof Error` is always true, because `TypeError`, `SyntaxError`, etc all inherit from `Error`. So when there was a `TypeError`, we only showed `err.message` without stack.

sindresorhus:
When all tests are finished and it logs the last `68 passed` that is moved one line up since the test title is no longer there. Would be nice if `68 passed` was always in the same place. Just log an empty line above it on the final call.

cli.js
vdemedes: And with that change, we show only `err.message` only in case of `throw new Error()` (e.g. https://github.com/sindresorhus/ava/blob/master/lib/fork.js#L50).

sindresorhus: Ok, I thought so. So why do we not want to show the stack for `Error`? Why is it special-cased?

sindresorhus:
Good move adding the test title.

vdemedes:
> Just log an empty line above it on the final call.

Just tried it, looks really bad:

screen shot 2015-12-25 at 2 07 11 pm

vdemedes:
> Good move adding the test title btw.

Credits for the idea go to @jamestalmage ;)

cli.js
vdemedes: Because we use `Error` for our internal messages (like the in the comment above), which don't need a stack and just informational. 

Errors that need stack trace are most of the time (I think) `TypeError`, `ParseError`, etc.

sindresorhus: Ok, good to know.

sindresorhus:
LGTM

jamestalmage:
> Because we use Error for our internal messages

In that case, we should create our own internal error class and use it where we explicitly want to suppress the stack trace.

sindresorhus:
> In that case, we should create our own internal error class and use it where we explicitly want to suppress the stack trace.

:+1: Same thing here: https://github.com/sindresorhus/ava/blob/4865674403c2eea40643fa8aa1d9383ea3e8eb3a/cli.js#L99

vdemedes:
Good suggestion!

sindresorhus:
Should be done in a follow-up: https://github.com/sindresorhus/ava/issues/368

sindresorhus:
:heart_eyes_cat: 

tumblr_lxx3edkjhr1qdlh1io1_400

Issue #309, opened 2015-12-06
throws doesn't count against assertion count and doesn't require error param


mattkrick:
Here's my actual code:
```
test('LocalStrategy:Signup:AlreadyExists  ', async t => {
  t.plan(1)
  await signupDB('test@test', '123123');
  t.throws(signupDB('test@test', 'XXXXXX'));
});
```
As is, it fails unless the first line is `t.plan(0)` (Assertion count does not match planned). I would assume a `throws` would count as an assertion, right?

Second, the docs show `error` is the second required param, but the assertion is successful no matter what value I put in there. I can even leave it undefined (as above) & it still passes. 

As a workaround, I'm using a try catch block that works fine, but it'd be nice to condense the code & just use throws.

jamestalmage:
How does `signupDB` throw? Does it do it synchronously? Then you need to wrap the call in a function: 

```js
t.throws(() => signupDB(...));
```

Does it return a promise that gets rejected? In that case, do **not** wrap it in a function, but **do** return the result of `t.throws`:

```js
test(t=> {
  return t.throws(signupDB(...));
});
```

mattkrick:
Ahhh, it was the `return` that got me, thanks!
Any word on the `error` param? Seems like it's optional, but if supplied, it can also take the Error's name, as well. If it's functioning correctly let me know & I'll write a little PR for the readme.

jamestalmage:
If I recall correctly, you must supply a string for an === comparison. I think I noticed that in the source a few days ago and meant to add regexp matching to my todos. Thanks for the reminder.

mattkrick:
That works for me, thanks!
I'll leave this open in case you wanna use it as a reminder, feel free to close.

jamestalmage:
Nope, I read the source wrong on that. Should support the same values for `err` that `core-assert#throws` does, with the added option of passing in a string for an === comparison

mattkrick:
Ah, that explains it then, from `core-assert`: 
`assert.throws = function(block, /*optional*/error, /*optional*/message) {...`

sindresorhus:
So what's the resolution on this? Anything we can do with the docs to make the behavior clearer?

mattkrick:
Thanks for following up on this!

In the `throws` section on the readme, it'd be nice to show an example explicitly pointing out that the statement must begin with `return`, that was the big gotcha for me since it's the only one that requires a return.

[...]

PR #360, opened 2015-12-23
Fix promise assertions

ariporad:
Hi All,

Here's the fix for #309, It has a good number of tests, but let me know if there's anything I should change.

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -172,6 +172,23 @@ Test.prototype._end = function (err) {
 Test.prototype.exit = function () {
 	var self = this;
 
+	function checkAssertLength() {
+		if (self.assertError === undefined && self.planCount !== null && self.planCount !== self.assertPromises.length) {
+			self._setAssertError(new assert.AssertionError({
+				actual: self.assertPromises.length,
+				expected: self.planCount,
+				message: 'Assertion count does not match planned',
+				operator: 'plan'
+			}));
+
+			self.assertError.stack = self.planStack;
+			return false;
+		}
+		return true;
+	}
+
+	checkAssertLength();
+
 	Promise.all(this.assertPromises)
 		.catch(function (err) {
 			self._setAssertError(err);
@@ -183,16 +200,7 @@ Test.prototype.exit = function () {
 			// stop infinite timer
 			globals.clearTimeout(self._timeout);
 
-			if (self.assertError === undefined && self.planCount !== null && self.planCount !== self.assertPromises.length) {
-				self._setAssertError(new assert.AssertionError({
-					actual: self.assertPromises.length,
-					expected: self.planCount,
-					message: 'Assertion count does not match planned',
-					operator: 'plan'
-				}));
-
-				self.assertError.stack = self.planStack;
-			}
+			checkAssertLength();
 
 			if (!self.ended) {
 				self.ended = true;
--- a/lib/test.js
+++ b/lib/test.js
@@ -25,7 +25,7 @@ function Test(title, fn) {
 
 	this.title = title || fnName(fn) || '[anonymous]';
 	this.fn = isGeneratorFn(fn) ? co.wrap(fn) : fn;
-	this.assertPromises = [];
+	this.assertions = [];
 	this.planCount = null;
 	this.duration = null;
 	this.assertError = undefined;
@@ -33,7 +33,7 @@ function Test(title, fn) {
 	Object.defineProperty(this, 'assertCount', {
 		enumerable: true,
 		get: function () {
-			return this.assertPromises.length;
+			return this.assertions.length;
 		}
 	});
 
@@ -54,7 +54,7 @@ function Test(title, fn) {
 module.exports = Test;
 
 Test.prototype._assert = function (promise) {
-	this.assertPromises.push(promise);
+	this.assertions.push(promise);
 };
 
 Test.prototype._setAssertError = function (err) {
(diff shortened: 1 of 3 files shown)


lib/test.js
```diff
@@ -165,35 +172,49 @@ Test.prototype._end = function (err) {
 Test.prototype.exit = function () {
 	var self = this;
 
-	// calculate total time spent in test
-	this.duration = globals.now() - this._timeStart;
+	function checkAssertLength() {
```
ariporad: I'm not sure if this should be here, or it should be a method, or if it should be an external function which accepts `self`. Thoughts?

lib/test.js
```diff
@@ -25,11 +25,18 @@ function Test(title, fn) {
 
 	this.title = title || fnName(fn) || '[anonymous]';
 	this.fn = isGeneratorFn(fn) ? co.wrap(fn) : fn;
-	this.assertCount = 0;
+	this.assertPromises = [];
```
jamestalmage: `assertPromises` -> `assertions`

lib/test.js
```diff
@@ -230,33 +251,23 @@ Test.prototype._publicApi = function () {
 	api.plan = this.plan.bind(this);
 
 	function skipFn() {
-		self._assert();
+		self._assert(Promise.resolve());
```
jamestalmage: `self._assert()` should work fine here. `Promise.all` converts array members to promise using `Promise.resolve` anyways.

test/test.js
```diff
@@ -2,6 +2,14 @@
 var test = require('tap').test;
 var _ava = require('../lib/test');
 
+function delay(val, time) {
```
jamestalmage: use [`delay`](https://www.npmjs.com/package/delay)

test/test.js
```diff
@@ -351,3 +359,119 @@ test('skipped assertions count towards the plan', function (t) {
 		t.end();
 	});
 });
+
+test('throws and doesNotThrow work with promises', function (t) {
```
jamestalmage: make two separate tests.

test/test.js
```diff
@@ -351,3 +359,119 @@ test('skipped assertions count towards the plan', function (t) {
...
+	}).run().then(function (a) {
+		t.ifError(a.assertError);
+		t.is(a.planCount, 2);
+		t.is(a.assertCount, 2);
+		t.is(asyncCalled, true);
+		t.end();
+	});
+});
+
+test('t.end is called when a promise passed to t.throws hasn\'t resolved yet', function (t) {
```
jamestalmage: this makes it sound like AVA will call `t.end` itself. I suggest something like:

`waits for t.throws to resolve after t.end is called`

This comment applies to multiple tests.

test/test.js
```diff
@@ -351,3 +359,119 @@ test('skipped assertions count towards the plan', function (t) {
...
+		}, 5);
+	}).run().catch(function (err) {
+		t.is(err.operator, 'plan');
+		t.is(err.actual, 3);
+		t.is(err.expected, 2);
+		t.end();
+	});
+});
+
+test('number of assertions doesn\'t t.plan when the test exits, but before all promises resolve another is added', function (t) {
```
jamestalmage: > number of assertions doesn't **match** plan

missing a word, and (minor nit) I think it reads better without the `t.`

test/test.js
```diff
@@ -2,6 +2,14 @@
 var test = require('tap').test;
 var _ava = require('../lib/test');
```
jamestalmage:
```js
```

Promise is not globally available in Node `0.10`

lib/test.js
```diff
@@ -165,35 +172,49 @@ Test.prototype._end = function (err) {
...
+		if (self.assertError === undefined && self.planCount !== null && self.planCount !== self.assertPromises.length) {
+			self._setAssertError(new assert.AssertionError({
+				actual: self.assertPromises.length,
+				expected: self.planCount,
+				message: 'Assertion count does not match planned',
+				operator: 'plan'
+			}));
+
+			self.assertError.stack = self.planStack;
+			return false;
```
jamestalmage: doesn't look like you use these return values.

jamestalmage:
@ariporad - Good work! A few minor nits, but overall very well done!

lib/test.js
ariporad: I think we should keep that here, as it makes it more obvious what we're doing, and it makes it a ***lot*** easier to reason about `this.assertions`.

jamestalmage: Agreed. I wrote this before I saw how you optimized the `onAssertion` callback. You can disregard.

ariporad:
@jamestalmage: Should be fixed now.

ariporad:
@jamestalmage: Hang on a second actually, delay doesn't like node@<4.0.0

sindresorhus:
> delay doesn't like node@<4.0.0

It's `node@<0.12`, and just put `global.Promise = require('bluebird');` at the top.

ariporad:
@sindresorhus: Yeah, I was trying to avoid that, but it's not that big a deal.

package.json
```diff
@@ -93,6 +93,7 @@
     "core-assert": "^0.1.0",
     "debug": "^2.2.0",
     "deeper": "^2.1.0",
+    "delay": "^1.2.0",
```
sindresorhus: This should be a dev dependency.

ariporad: Fixed.

lib/test.js
sindresorhus: I would make it a private instance method (prefixed with `_`). The function checks if assertion count matches plan count. I think the name could be clearer.

ariporad: Done.

lib/test.js
```diff
@@ -181,19 +180,38 @@ Test.prototype.exit = function () {
 
 		this.assertError.stack = this.planStack;
 	}
+};
 
-	if (!this.ended) {
-		this.ended = true;
+Test.prototype.exit = function () {
```
sindresorhus: remove this empty line

ariporad: @sindresorhus: Fixed

ariporad:
Ok, @jamestalmage, @sindresorhus: Everything should be fixed now.

vdemedes:
LGTM, great stuff!

jamestalmage:
Let's get https://github.com/sindresorhus/delay/pull/8 merged and published. Then use the updated function here. Once that happens, I think we are good to go.

ariporad:
Ok, I just rebased onto master.

sindresorhus:
@ariporad delay@1.3.0 is now out. Can you update tests to use it? Other than that, this looks good to land.

ariporad:
@sindresorhus: Ok. Fixed, and rebased onto master. Ready for merging.

sindresorhus:
Landed. Really dig this change @ariporad :)

PR #361, opened 2015-12-23
Travis CI cache npm_modules

fscherwi:
Cache npm packages for an faster Travis build!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,3 +7,6 @@ before_install:
   - 'npm i -g npm@latest'
 after_success:
   - '[ -z "$COVERALLS_REPO_TOKEN" ] && tap --coverage-report=text-lcov | ./node_modules/.bin/coveralls'
+cache:
+  directories:
+    - node_modules


ariporad:
I literally started working on this 30 seconds ago! Great minds think alike!

jamestalmage:
Caching `node_modules` is a bad idea. If you delete a dependency from `package.json`, but your code still uses it, you won't get a "Can't find module ..." failure because it still is available in the cached folder.

Build caching should almost never be used. The exception being truly huge assets that almost never change.

madbence:
@jamestalmage if you're working with a PR-based workflow, this won't be a problem, because cache is per PR (at least on travis), so the PR will fail before merge.

sindresorhus:
:-1: The benefit is minor as Travis is fast already and I like to test with all the latest sub-dependencies for each run. Imagine if a sub-sub-sub-dependency breaks in a patch release. I want Travis to be able to catch that.

madbence:
also the prs will be still slow, since they don't have cache on the first build

PR #362, opened 2015-12-23
Cache node_modules on appveyor and disable npm progress bar

ariporad:
Since @fscherwi beat me to Travis, this just caches `node_modules` on AppVeyor. It should speed up the build quite a bit.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,6 +10,8 @@ install:
   - ps: Install-Product node $env:nodejs_version
   - set CI=true
   - set AVA_APPVEYOR=true
+  - npm config set progress false
+  - npm config set spinner false
   - npm -g install npm@latest || (timeout 30 && npm -g install npm@latest)
   - set PATH=%APPDATA%\npm;%PATH%
   - npm install || (timeout 30 && npm install)
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,6 +3,9 @@ environment:
     - nodejs_version: '5'
     - nodejs_version: '4'
     - nodejs_version: '0.12'
+cache:
+  - node_modules                      # local npm modules
+  - '%APPDATA%\npm-cache'             # npm cache
 install:
   - ps: Install-Product node $env:nodejs_version
   - set CI=true


ariporad:
I also disabled the spinner and progress bar on npm to make the output look a lot nicer. I can enable the http logging so you can still see what's going on, if you like.

jamestalmage:
Same as my comment in #361, I think caching `node_modules` is a bad idea. It is a lot more tempting for AppVeyor since it is so slow, but probably still not worth the risks.

I'm confused by the second commit. If I review the AppVeyor output from your PR, and others like it, they seem identical. What am I missing?

ariporad:
@jamestalmage: Huh, AppVeyor must have changed it. Never mind.

Although, I think we should cache node_modules on AppVeyor, because npm on AV is a lot slower, and any dependency mess ups will be caught by travis.

jamestalmage:
> I think we should cache node_modules on AppVeyor, because npm on AV is a lot slower, and any dependency mess ups will be caught by travis.

That is a tempting thought. However, it [doesn't seem the build is actually any faster](https://ci.appveyor.com/project/sindresorhus/ava/history) between your first and second commit, should it be?

ariporad:
@jamestalmage: No. It should be faster between this branch and master.

jamestalmage:
For your first commit there wouldn't have been a cache yet, so it should have been slow. Check the [build history](https://ci.appveyor.com/project/sindresorhus/ava/history), discounting hung builds - this PR takes about the same amount of time to build.

ariporad:
Hmm... I guess never mind then. (Although that's awful strange).

jamestalmage:
> Although that's awful strange

I don't know. If I had to guess, I would bet AppVeyor proxies the `npm` repository, so installs happen very fast.

ariporad:
Yeah, Probably.

sindresorhus:
The slowest part about AppVeyor is that your builds are always queued for a long time before starting. AppVeyor probably deprioritize free plans.

PR #364, opened 2015-12-25
fix usage example in the readme

ammaroff:


sindresorhus:
Good catch. Thanks :)

Merry xmas 🎄🦄

PR #366, opened 2015-12-25
Babel feature detection

### LINKED ISSUE #348: Don't use `regenerator` in `babel-preset-es2015` for newer Node.js versions
Split out of https://github.com/sindresorhus/ava/pull/333.

Only Node.js 0.10 needs `regenerator`, but it's currently included for any Node.js version.

Someone needs to do a pull request to [`babel-preset-es2015`](https://github.com/babel/babel/tree/master/packages/babel-preset-es2015) and [auto-detect generator support](https://github.com/sindresorhus/has-generator) and conditionally include `regenerator` (preferred solution) or add an option for toggling `regenerator`.

// @thejameskyle @jamestalmage

ariporad:
This fixes #348. It uses [`babel-preset-features`](http://npm.im/babel-preset-features) to auto-detect ES6 features. Also includes tests.

Enjoy!

### DIFF
--- a/lib/babel.js
+++ b/lib/babel.js
@@ -85,7 +85,7 @@ if (prevHash === currHash) {
 		patterns: require('./enhance-assert').PATTERNS
 	});
 
-	options.presets = [require('babel-preset-stage-2'), require('babel-preset-es2015')];
+	options.presets = [require('babel-preset-stage-2'), require('babel-preset-features')];
 	options.plugins = [powerAssert, require('babel-plugin-transform-runtime')];
 
 	var transpiled = babel.transformFileSync(testPath, options);
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
     "babel-core": "^6.3.21",
     "babel-plugin-espower": "^2.0.0",
     "babel-plugin-transform-runtime": "^6.3.13",
-    "babel-preset-es2015": "^6.3.13",
+    "babel-preset-features": "^2.0.2",
     "babel-preset-stage-2": "^6.3.13",
     "babel-runtime": "^6.3.19",
     "bluebird": "^3.0.0",
--- a/test/api.js
+++ b/test/api.js
@@ -26,6 +26,17 @@ test('generators support', function (t) {
 		});
 });
 
+test('babel is only applied when needed', function (t) {
+	t.plan(1);
+
+	var api = new Api([path.join(__dirname, 'fixture/only-babel-when-needed.js')]);
+
+	api.run()
+		.then(function () {
+			t.is(api.passCount, 1);
+		});
+});
+
 test('async/await support', function (t) {
 	t.plan(1);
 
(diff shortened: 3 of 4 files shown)


ariporad:
Grr... A dependency has a `require` with an uppercase letter, which is killing travis. I submitted a PR, but the author is on the other side of the world. Give it a few hours, then it should be ready.

sindresorhus:
I like the idea, but I'm a bit wary of it. What if the native implementation has bugs and we want to use Babel for that? What is the runtime overhead of doing a bunch of feature detections? And https://github.com/hax/babel-features doesn't look very actively maintained nor used much https://www.npmjs.com/package/babel-features.

I just don't want this to cause us a lot of extra overhead with support issues, etc.

-

// @hax

ariporad:
@sindresorhus: Check the 2.x (currently in beta) branch, it was last worked on 5 days ago :wink:. It even already supports Babel 6.

I think there's a little overhead, but not a whole lot. I'm not sure how we'd go about it any other way, regardless. And I guess we'd just assume that V8 isn't going to ship something horrifically broken. The overhead is going to down once we move babel to the main thread.

sindresorhus:
> I think there's a little overhead, but not a whole lot. I'm not sure how we'd go about it any other way, 

We could create a preset for each Node.js major version  →https://github.com/sindresorhus/ava/issues/148. Albeit that won't work when we eventually add browser support → https://github.com/sindresorhus/ava/issues/24.

> And I guess we'd just assume that V8 isn't going to ship something horrifically broken.

They do all the time. That's the price for rapid evolution.

> The overhead is going to down once we move babel to the main thread.

That's a good point.

hax:
Sorry I have cold last week and not check the github notifications.

> What is the runtime overhead of doing a bunch of feature detections?

The runtime overhead of feature detection < 10ms, so I think it's ok for most applications.

>  What if the native implementation has bugs and we want to use Babel for that?

Yes they have. But Babel also have :grinning: . So we need to check it case by case. For example, whether to use `regenerator`. (node 0.12+ have generators, but missing some minor features such as `Generator.prototype.return`. `babel-preset-es2015-node-5` decide to not use `regenerator`. If there is any other case like that please let me know.)

sindresorhus:
There's also the case of many of the ES2015 features being a lot slower than the transpiled ones.

hax:
@sindresorhus 
This is due to V8 (and other implementations) seem focus on es2015 features not performance up to now. Even simple const/let could cause deoptimization.

But I never see a real case such performance degradation is significant for a webapp (browser-side) in my daily work. Maybe heavy-load Node.js apps would affected by it. Anyway they can just switch from `babel-preset-es2015-node-*` to normal `babel-preset-es2015`.

jamestalmage:
My vote is to punt on this until a later date. AVA and Babel both are currently moving at breakneck speed. I would prefer not to introduce an extra layer of complexity.

novemberborn:
Closing due to inactivity. #148 is the issue for this feature. Would love your help if you have the time @ariporad!

Issue #249, opened 2015-11-21
If specifying a directory, tests in subdirectories should be run too


rightaway:
If I do `ava test` where test is a directory, it only runs the files directly under `test`. But if I had subdirectories under it, those don't get run. I think it should run all tests in that directory recursively. Or at least there should be a command line flag like `-r` to get this behavior.

sindresorhus:
It's intentionally not recursive, as a good default and simplicity. If you need it to be recursive just specify the appropriate glob pattern. Like: `$ ava test/**`.

rightaway:
Ok. Now if I have a file `test/test1.js` and `test/subtest/test2.js` (each with one test called `runTest` in it) and run `ava test/**/*.js` (glob pattern in ZSH) on the command line, the output looks like this

```
test2 > runTest
test > test1 > runTest
```

But it should look more like

```
test1 > runTest
subtest > test2 > runTest
```

At least while nested groups aren't available (https://github.com/sindresorhus/ava/issues/222) using directories is a good way to organize structure, so the output should be in line with that.

The beginning `test` doesn't appear above since that's the 'root' test folder, but if there's no option to specify a directory and a recursive flag then I don't see how it would know that test shouldn't be printed there.

That's why I think allowing to specify a directory with a recursive flag (so it's not the default and not a breaking change) would be useful, because then ava can format the results appropriately. Because using glob patterns ava just receives a list of filenames to test and isn't aware of a 'root'.

sindresorhus:
We don't add options lightly and we try very hard to keep AVA simple. You can solve this with a glob pattern. We'll fix the output somehow.

rightaway:
A suggestion of how it could look. After seeing all the files passed in with the glob pattern it should strip the leading directories that are all in common. The way if someone passes in a pattern like `some/sub/directory/**/*.js`, and the only files that appear are `some/sub/directory/sub/path/test1.js` and `some/sub/directory/sub/path/test2.js`, then the part in common `some/sub/directory/sub/path` should not appear in the output.

sindresorhus:
Agreed, that's what I was thinking too.

Qix-:
I think it should be recursive by default.

In my case, I want to mimic my `lib/` folder with test files in order to yield more [succinct output](https://github.com/sindresorhus/ava/issues/53#issuecomment-141339499) with nested files.

Further, this is how other languages' test systems work and is analogous to those for me, personally - hence my decision to structure my tests like this.

[...]

PR #370, opened 2015-12-26
Make directories recursive.

ariporad:
Refs #249. (It's only part 1).


ariporad:
Actually, I'm just going to make them one PR.

PR #373, opened 2015-12-26
Make AVA display subdirs properly and be recursive by default.

### LINKED ISSUE #249: If specifying a directory, tests in subdirectories should be run too

ariporad:
Fixes #249.

### DIFF
--- a/api.js
+++ b/api.js
@@ -8,6 +8,7 @@ var Promise = require('bluebird');
 var figures = require('figures');
 var globby = require('globby');
 var chalk = require('chalk');
+var commondir = require('commondir');
 var resolveCwd = require('resolve-cwd');
 var fork = require('./lib/fork');
 var formatter = require('./lib/enhance-assert').formatter();
@@ -100,25 +101,6 @@ Api.prototype._handleTest = function (test) {
 	this.emit('test', test);
 };
 
-Api.prototype._findBase = function (files) {
-	this.base = files.reduce(function (base, file) {
-		file = path.relative('.', file);
-		file = file.split(path.sep);
-		if (base === false) {
-			return file;
-		}
-		return base.filter(function (part, i) {
-			return file[i].toLowerCase() === part.toLowerCase();
-		});
-	}, false).join(path.sep);
-
-	if (this.base === '' || this.base === '.') {
-		this.base = this.files[0] || 'test';
-	}
-
-	this.base += path.sep;
-};
-
 Api.prototype._prefixTitle = function (file) {
 	if (this.fileCount === 1) {
 		return '';
@@ -156,7 +138,7 @@ Api.prototype.run = function () {
 
 			self.fileCount = files.length;
 
-			self._findBase(files);
+			self.base = path.relative('.', commondir('.', files)) + path.sep;
 
 			var tests = files.map(self._runFile);
 
@@ -209,9 +191,7 @@ function handlePaths(files) {
 	if (files.length === 0) {
 		files = [
 			'test.js',
-			'**/test.js',
 			'test-*.js',
-			'**/test-*.js',
 			'test'
 		];
 	}
--- a/api.js
+++ b/api.js
@@ -31,6 +31,7 @@ function Api(files, options) {
 	this.stats = [];
 	this.tests = [];
 	this.files = files || [];
+	this.base = '';
 
 	Object.keys(Api.prototype).forEach(function (key) {
 		this[key] = this[key].bind(this);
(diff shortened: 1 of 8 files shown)


api.js
```diff
@@ -99,6 +100,25 @@ Api.prototype._handleTest = function (test) {
...
+			return file[i].toLowerCase() === part.toLowerCase();
+		});
+	}, false).join(path.sep);
+
+	if (this.base === '' || this.base === '.') {
+		this.base = this.files[0] || 'test';
+	}
+
+	this.base += path.sep;
+};
```
sindresorhus: https://github.com/substack/node-commondir

api.js
```diff
@@ -194,8 +209,10 @@ function handlePaths(files) {
 	if (files.length === 0) {
 		files = [
 			'test.js',
+			'**/test.js',
```
sindresorhus: This also matches `test.js`, so the above can be removed.

sindresorhus: Same with the below.

api.js
```diff
@@ -194,8 +209,10 @@ function handlePaths(files) {
...
 			'test-*.js',
-			'test/*.js'
+			'**/test-*.js',
+			'test'
```
sindresorhus: Actually, the change to this array is wrong.

It should match `test.js` and `test-*.js` in root, and any JS file recursively in `test` folder as long as it's not prefixed with an underscore.

readme.md
```diff
@@ -120,7 +120,7 @@ $ ava --help
   test.js test-*.js test/*.js
 ```
 
-Files in directories named `fixtures` and `helpers` are ignored, as well as files starting with `_`. This can be useful for having helpers in the same directory as your test files.
+Directories are recursive by default. Files in directories named `fixtures` and `helpers` are ignored, as well as files starting with `_`. This can be useful for having helpers in the same directory as your test files.
```
sindresorhus: This could be explained better. Not clear what it means.

test/api.js
```diff
@@ -238,6 +259,17 @@ test('absolute paths', function (t) {
 		});
 });
 
+test('recursive for directory', function (t) {
```
sindresorhus: Not sure what it's testing from the test description.

test/fixture/subdir/failing-subdir.js
```diff
@@ -0,0 +1,8 @@
+var test = require('../../../');
+
+test('subdir fail', function (t) {
+	t.fail();
+});
+
+
+
```
sindresorhus: Excessive newlines

test/fixture/subdir/failing-subdir.js
```diff
@@ -0,0 +1,8 @@
+var test = require('../../../');
+
+test('subdir fail', function (t) {
+	t.fail();
+});
```
sindresorhus: Use ES2015 syntax like the other test fixtures.

test/fixture/subdir/in-a-subdir.js
```diff
@@ -0,0 +1,5 @@
+var test = require('../../../');
+
+test('subdir', function (t) {
+	t.pass();
+});
```
sindresorhus: Use ES2015 syntax like the other test fixtures.

api.js
ariporad: Naturally, I spend 30 minutes on something to find out that there's already something that does that.

readme.md
ariporad: @sindresorhus: Sorry, I'm really terrible with wording. Would something like "Directories will be expanded recursively" be better?

sindresorhus: Not totally happy with this, but something like:

> By default, AVA run tests named `test.js` or starting with `test-` in root and any tests recursively in the `test` directory.

You also need to update the globbing pattern in the CLI help output (and the identical output in the readme).

ariporad: @sindresorhus: But that doesn't really indicate that any directories that you pass it, nor does it include the bit about fixtures and helpers or underscores.

sindresorhus: @ariporad Good point, should include that too.

ariporad: @sindresorhus: Maybe something like 

> Directories will be recursively expanded. Files in directories named `fixtures` and `helpers` are ignored, as well as files starting with `_`. This can be useful for having helpers in the same directory as your test files.

ariporad:
@sindresorhus: What's the status on this?

sindresorhus:
@ariporad At least one other maintainer need to review it.

Next time try to make your PR focused. This should probably have been two separate PRs.

readme.md
sindresorhus: :+1:

ariporad:
@sindresourhus: sorry about that, I had mentally sorted it by issue, but that doesn't make much sense outside my head. Since I implemented them sequentially, I could split them into seperate PRs, should I do that?

ariporad:
Closing this to split into multiple pull request.

PR #371, opened 2015-12-26
Use AvaError for internal errors

vdemedes:
This PR fixes #368.

Now we use `AvaError` for internal errors, that don't need a stack trace.

PR #376, opened 2015-12-27
Include actual and expected when the number of assertions doesn't match the plan.

### LINKED ISSUE #102: When assertion count is incorrect, output the expected and actual values.
I see that the actual count is recorded, but I'm not sure how to see it when running `ava`.

https://github.com/sindresorhus/ava/blob/master/lib/test.js#L150-L158

```
❯ npm test

> npm-check@4.1.2 test /Users/dgreene/projects/npm-check
> xo && ava ./test


  ✖ require Assertion count does not match planned

  1 test failed

  1. require
  Error
    at Test.plan (/Users/dgreene/projects/npm-check/node_modules/ava/lib/test.js:87:19)
    at Test.<anonymous> (/Users/dgreene/projects/npm-check/test/test.js:10:7)
    at step (/Users/dgreene/projects/npm-check/test/test.js:5:273)
    at Test.fn (/Users/dgreene/projects/npm-check/test/test.js:5:99)
```

The above is using using `master` with `power-assert` added.

Similar output using the published version

```
> npm-check@4.1.2 test /Users/dgreene/projects/npm-check
> xo && ava ./test


  ✖ require Assertion count does not match planned

  1 test failed

  1. require
  Error
    at Test.plan (/Users/dgreene/projects/npm-check/node_modules/ava/lib/test.js:59:19)
    at Object.callee$0$0$ (/Users/dgreene/projects/npm-check/test/test.js:10:7)
    at Test.callee$0$0 [as fn] (/Users/dgreene/projects/npm-check/test/test.js:32:26)
```

ariporad:
This isn't as good as it could be, as it doesn't have a nice assertion diff, because that's generated by power assert, which does all kinds of other things too that we don't want

Fixes #102.

@sindresorhus, @jamestalmage, @vdemedes: thoughs?

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -8,6 +8,7 @@ var observableToPromise = require('observable-to-promise');
 var isPromise = require('is-promise');
 var isObservable = require('is-observable');
 var assert = require('./assert');
+var AvaError = require('./ava-error');
 var enhanceAssert = require('./enhance-assert');
 var globals = require('./globals');
 
@@ -75,10 +76,6 @@ Test.prototype.plan = function (count) {
 	}
 
 	this.planCount = count;
-
-	// in case the `planCount` doesn't match `assertCount,
-	// we need the stack of this function to throw with a useful stack
-	this.planStack = new Error().stack;
 };
 
 Test.prototype.run = function () {
@@ -174,14 +171,11 @@ Test.prototype._end = function (err) {
 
 Test.prototype._checkPlanCount = function () {
 	if (this.assertError === undefined && this.planCount !== null && this.planCount !== this.assertions.length) {
-		this._setAssertError(new assert.AssertionError({
-			actual: this.assertions.length,
-			expected: this.planCount,
-			message: 'Assertion count does not match planned',
-			operator: 'plan'
-		}));
+		var planError = new AvaError('Assertion count of ' + this.assertions.length + ' does not match plan of ' + this.planCount + '.');
+		planError.actual = this.assertions.length;
+		planError.expected = this.planCount;
 
-		this.assertError.stack = this.planStack;
+		this._setAssertError(planError);
 	}
 };
 
--- a/test/test.js
+++ b/test/test.js
@@ -88,6 +88,8 @@ test('run more assertions than planned', function (t) {
 	}).run().catch(function (err) {
 		t.ok(err);
 		t.is(err.name, 'AssertionError');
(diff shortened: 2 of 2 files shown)


sindresorhus:
It's an ok workaround, but I don't think this is the correct way to solve it for real. We already supply the `actual` and `expected` to the assert error and it should handle the output, but somehow that's lost. Might be because we overwrite the `.stack`.

sindresorhus:
Could be more succinct. Something like, but better:

> Assertion count of 4 does not match planned of 5

ariporad:
@sindresorhus: It's not related to the stack. Check my comment in #102. It's a power-assert thing.

ariporad:
@sindresorhus: Message fixed.

jamestalmage:
Just wondering here. Now that #371 is merged, should we just use that for this? Is a stack trace really going to be that helpful? Especially since `test.exit()` is now always deferred via a promise?

sindresorhus:
@jamestalmage I would still want to know which assertion and test that triggers the "incorrect plan count error".

jamestalmage:
We would be setting it via `test._setAssertError`, so we will still know which test triggered that.
We currently only check the assert count when the test exits, not for every assert, so you can't know which assertion caused the error now.

(`test._setAssertError` is a bit of a misnomer now).

ariporad:
@sindresourhus, @jamestalmage: yeah, I think we probably should. I'll work on that in the am.

jamestalmage:
@ariporad - How close are you to finishing this?

ariporad:
@jamestalmage, @sindresorhus: OK, rebased onto master, and now uses `AvaError`.

sindresorhus:
Tests are failing.

ariporad:
@sindresorhus: of course they are.

ariporad:
Naturally, They're all false negatives.

@sindresorhus: Do AvaError's need `.operator`s, `.actual`s or `.expected`s?

ariporad:
@jamestalmage: ^^^^^^^^^^

sindresorhus:
@ariporad Sorry, missed this one. Can you rebase on master and fix the merge conflict?

> Do AvaError's need .operators, .actuals or .expecteds?

No

sindresorhus:
So, what should we do about this PR?

sindresorhus:
Would like to see this land, but closing for now to clean up inactive PRs. @ariporad Feel free to do a new one if you ever have time ;)

Issue #365, opened 2015-12-25
Standardize errors in API


vadimdemedes:
Currently, `errors` array of `API` instance contains exceptions, rejections and failed tests. As a consequence, items can be different from one another and we have to detect what's what in a loop.

We need to standardize the items in `errors` array, so that it's easily "forEachable" and all items have the same properties.

jamestalmage:
:+1:

sindresorhus:
:+1:

ariporad:
@sindresorhus, @jamestalmage, @vdemedes: How would you want this implemented? Should everything just get wrapped in an error? A promise? A custom object?

vdemedes:
I haven't thought about that yet, but every item in `errors` should have the same set of properties, so that we can easily output the list of errors at once. So, whether it's a failed test or uncaught exception (which does not have a test assigned to it), the error object should have the same props.

jamestalmage:
Also, `unhandledRejection`.

jamestalmage:
Copying a my suggestion [here](https://github.com/sindresorhus/ava/pull/380#issuecomment-167447244):

> Instead of separate errors and tests arrays, should it just be tests? A test could have a status attached (passed / failed), and all assertions with their respective results. Might be better than just standardizing errors
>   -- @jamestalmage

and a valid counterpoint:    

> we also need to store rejections and exceptions, which are not attached to any test and don't have test title, etc. So storing rejections and exceptions in tests doesn't sound right.
> -- @vdemedes

sotojuan:
What's the status of this?

novemberborn:
@sotojuan it's up for the taking. Maybe we need to bikeshed the standardization details, though we could do that in a PR as well if you have a good idea on where to start.

novemberborn:
While this might improve our code quality, it's not actually preventing any functionality. We've also not made progress on this refactor in over a year. I'm closing the issue, we can deal with it organically when the need arises.

PR #380, opened 2015-12-27
Display power-assert output in mini reporter

### LINKED ISSUE #372: Assertion output regression
In master (https://github.com/sindresorhus/ava/commit/945dbea40a6a500a24d54ebcc7e10dcd75647fc7), there's no longer any `power-assert` output.

```js
import test from 'ava';

test(t => {
	t.ok(false);
});
```

```
  1 failed

  1. [anonymous]
  AssertionError: false == true
    at Test.fn (/Users/sindresorhus/dev/md5-hex/test.js:4:4)
```

vdemedes:
This PR fixes #372.

Mini reporter (default one) was not displaying power-assert output. Now it does (along with a stack trace):

screen shot 2015-12-27 at 11 34 43 pm

### DIFF
--- a/lib/reporters/mini.js
+++ b/lib/reporters/mini.js
@@ -82,7 +82,7 @@ MiniReporter.prototype.finish = function () {
 	var i = 0;
 
 	if (this.failCount > 0) {
-		this.api.tests.forEach(function (test) {
+		this.api.errors.forEach(function (test) {
 			if (!test.error || !test.error.message) {
 				return;
 			}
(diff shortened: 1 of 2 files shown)


sindresorhus:
LGTM

// @ben-eb

ben-eb:
:+1:

jamestalmage:
LGTM.

Though I think this highlights some deficiencies in our API.

Instead of separate `errors` and `tests` arrays, should it just be `tests`? A `test` could have a `status` attached (passed / failed), and all assertions with their respective results. Might be better than just standardizing errors as described in #365

sindresorhus:
> Instead of separate errors and tests arrays, should it just be tests? A test could have a status attached (passed / failed), and all assertions with their respective results. Might be better than just standardizing errors as described in #365

:+1: Can you open an issue?

vdemedes:
@jamestalmage we also need to store rejections and exceptions, which are not attached to any test and don't have test title, etc. So storing rejections and exceptions in `tests` doesn't sound right.

PR #384, opened 2015-12-29
Hide annoying error print outs.

### LINKED ISSUE #339: Missing test fixture
Occurs in latest master.

screen shot 2015-12-15 at 4 47 16 pm

jamestalmage:
Fixes #339.

To get the correct file and line number from `SyntexError`s, we need to give up control of how they are logged to console (see #308 for a full discussion on why this is true).

The unfortunate side effect was that tests exercising our syntax errors

I was not sure this was worth it (it is a lot of code two fix the output from just two tests), but then I [stumbled on this](https://github.com/sindresorhus/ava/commit/5e6327de2dcdc7a63d7d53f505673dbfd7973393). There seem to be situations where [a bad plan count, will not fail the tap test suite](https://github.com/isaacs/node-tap/issues/198). Now I think this is probably worth the effort.

### DIFF
--- a/package.json
+++ b/package.json
@@ -123,6 +123,7 @@
   "devDependencies": {
     "coveralls": "^2.11.4",
     "delay": "^1.3.0",
+    "get-stream": "^1.1.0",
     "signal-exit": "^2.1.2",
     "sinon": "^1.17.2",
     "source-map-fixtures": "^0.4.0",
--- a/test/cli.js
+++ b/test/cli.js
@@ -1,6 +1,8 @@
 'use strict';
 var childProcess = require('child_process');
 var test = require('tap').test;
+global.Promise = require('bluebird');
+var getStream = require('get-stream');
 
 function execCli(args, cb) {
 	if (!Array.isArray(args)) {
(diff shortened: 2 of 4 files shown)


sindresorhus:
LGTM

test/cli.js
```diff
@@ -13,10 +15,41 @@ function execCli(args, cb) {
...
+	var processPromise = new Promise(function (resolve) {
+		var child = childProcess.spawn(process.execPath, ['../cli.js'].concat(args), {
+			cwd: __dirname,
+			env: env,
+			stdio: [null, 'pipe', 'pipe']
+		});
+
+		child.on('close', function (code, signal) {
+			if (code) {
+				var err = new Error('exited with a non-zero exit code: ' + code);
```
vdemedes: I'd use `Test worker exited with ...`

jamestalmage:
I just realized I could use `get-stream` here.

test/cli.js
```diff
@@ -1,6 +1,9 @@
 'use strict';
 var childProcess = require('child_process');
 var test = require('tap').test;
+var concatStream = require('concat-stream');
```
sindresorhus: moot now. remove.

package.json
```diff
@@ -121,8 +121,10 @@
     "update-notifier": "^0.6.0"
   },
   "devDependencies": {
+    "concat-stream": "^1.5.1",
```
sindresorhus: moot now. remove.

jamestalmage:
OK, updated to use `get-stream`, removed `concat-stream`, and updated with @vdemedes suggested verbiage.

PR #387, opened 2015-12-29
Support promises rejected with non-Errors

### LINKED ISSUE #386: Regression: 0.9.0 Not marking tests as failed
Output on `0.8.0`

screenshot 2015-12-29 07 17 28

`0.9.0` with `--verbose` flag

screenshot 2015-12-29 07 18 42

`0.9.0` with minimal reporter

screenshot 2015-12-29 07 19 04

In all cases, the test exits correctly with a non-zero error code.

All the above output is generated from this commit: https://github.com/jamestalmage/loud-rejection/commit/432b4fb3657ec4af407f4098ee80ff727e1c0d13

It uses `ava@0.8.0` by default, you will need to install `ava@0.9.0` to monitor the other outputs.

vdemedes:
Fixes #386.

Previously, AVA expected `Error`s in rejected promises and did not support other types of rejected values. This resulted in a wrong output, marking failed tests as passed.

Now, if promise rejected with non-`Error`, we wrap it into `assert.AssertionError` to ensure promises always reject with `Error`s.

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -125,6 +125,14 @@ Test.prototype.run = function () {
 				self.exit();
 			})
 			.catch(function (err) {
+				if (!(err instanceof Error)) {
+					err = new assert.AssertionError({
+						actual: err,
+						message: 'Promise rejected with "' + err + '"',
+						operator: 'promise'
+					});
+				}
+
 				self._setAssertError(err);
 				self.exit();
 			});
--- a/test/promise.js
+++ b/test/promise.js
@@ -250,3 +250,14 @@ test('reject', function (t) {
 		t.end();
 	});
 });
+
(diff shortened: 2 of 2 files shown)


lib/test.js
```diff
@@ -125,6 +125,14 @@ Test.prototype.run = function () {
 				self.exit();
 			})
 			.catch(function (err) {
+				if (!(err instanceof Error)) {
+					err = new assert.AssertionError({
+						actual: err,
+						message: 'Promise rejected → ' + err,
```
sindresorhus: Maybe:

```js
message: 'Promise rejected with "' + err + '"',
```

?

vdemedes: Cool!

PR #390, opened 2015-12-30
Transpile in main thread

jamestalmage:
Manual reimplementation of #349.

This moves Babel transforms to the main process, and caches the transpiled results.

It uses `caching-transform` to provide caching. Expensive requires (like babel) are placed inside the factory function, which caching-transform only calls when it determines it needs to transform a file.

The path to each precompiled test is passed in the options object to fork:

```js
fork('/path/to/file.js', {
  precompiled: {
    '/path/to/file.js': '/path/to/cached/file.js'
  }
});
```

It uses `require-precompiled` to install a require hook that pulls the transformed file from the cache.

The caching directory is used by the users `package.json`, and installing in `node_modules/.cache/ava`.
If the cache directory cannot be found (i.e. if they have no package.json), or the user provides a --no-cache flag, pre-compilation still happens on the main thread - but results are published to a temp directory.


jamestalmage:
Using `got` `6.0.0` branch: 


master - first run (no cache)

```
./node_modules/.bin/ava  22.74s user 1.81s system 372% cpu 6.587 total
```

master - second run (with cache)

```
./node_modules/.bin/ava  6.65s user 0.77s system 175% cpu 4.240 total
```

this branch - first run (no cache)

```
./node_modules/.bin/ava  7.78s user 0.79s system 132% cpu 6.474 total
```

this branch - second run (w /cache)

```
./node_modules/.bin/ava  6.76s user 0.77s system 178% cpu 4.226 total
```

jamestalmage:
@sindresorhus, @vdemedes, et al. - ready for review.

lib/caching-precompiler.js
```diff
@@ -0,0 +1,95 @@
...
+var fs = require('fs');
+var path = require('path');
+var md5Hex = require('md5-hex');
+var stripBom = require('strip-bom');
+
+module.exports = CachingPrecompiler;
+
+function CachingPrecompiler(cacheDir) {
+	if (!(this instanceof CachingPrecompiler)) {
+		throw new Error('CachingPrecompiler must be called with new');
```
ariporad: @jamestalmage: What we do everywhere else is just invoke ourself with `new` and return it. Maybe switch to that?:

```javascript
if (!(this instanceof CachingPrecompiler)) {
    return new CachingPrecompiler(cacheDir);
}
```

lib/caching-precompiler.js
```diff
@@ -0,0 +1,95 @@
...
+	var enhanceAssert = require('./enhance-assert');
+
+	// initialize power-assert
+	return createEspowerPlugin(babel, {
+		patterns: enhanceAssert.PATTERNS
+	});
+};
+
+CachingPrecompiler.prototype._createTransform = function () {
+	var self = this;
```
ariporad: @jamestalmage: Why is this here? There are no closures?

lib/caching-precompiler.js
```diff
@@ -0,0 +1,95 @@
...
+
+	return cachingTransform({
+		factory: this._factory.bind(this),
+		cacheDir: self.cacheDir,
```
ariporad: @jamestalmage: See above, this could just be `this`.

lib/test-worker.js
```diff
@@ -34,73 +30,40 @@ sourceMapSupport.install({
 		if (sourceMapCache[source]) {
```
ariporad: I think this should be `if (!sourceMapCache[source]) {`.

lib/test-worker.js
```diff
@@ -34,73 +30,40 @@ sourceMapSupport.install({
...
 			return {
 				url: source,
-				map: sourceMapCache[source]
+				map: fs.readFileSync(sourceMapCache[source], 'utf8')
```
ariporad: Two things:

1. It seems like this should be moved into the compiler.
2. Why are we writing maps to disk just to read them again. Wouldn't it be way faster to just not write them in the first place, and get them from memory here?

lib/test-worker.js
```diff
@@ -34,73 +30,40 @@ sourceMapSupport.install({
...
 
-	var transpiled = babel.transformFileSync(testPath, options);
+// Modules need to be able to find `babel-runtime`, which is nested in our node_modules w/ npm@2
+var nodeModulesDir = path.join(__dirname, '../node_modules');
+var oldNodeModulesPaths = module.constructor._nodeModulePaths;
+module.constructor._nodeModulePaths = function () {
+	var ret = oldNodeModulesPaths.apply(this, arguments);
+	ret.push(nodeModulesDir);
+	return ret;
+};
```
ariporad: @jamestalmage: Why not just modify `process.env.NODE_PATH` (I think that still works)? Or you could use [`app-module-path`](http://npm.im/app-module-path) and just pass in the path to node modules.

Or, you know what? This is probably fine. But the bikeshed is to be blue.

ariporad:
@jamestalmage: Looks pretty good, expect for the few things I mentioned. Also, it looks like there aren't very many tests on the `CachingPrecompiler` itself (I don't know if there needs to be).

lib/caching-precompiler.js
jamestalmage: Yeah sure. Probably worth matching style. 

That said, after doing it the way you suggest for a long time, I am starting to think it's a mistake. ES2015 classes throw when function called, so magic auto-`new`ing is not future proof. Also, it introduces an easy to overlook failure point when refactoring the function signature. I know I have forgotten to update that statement inside the if condition before.

lib/caching-precompiler.js
jamestalmage: Because the factory function used to be inlined, and I missed it. Good catch.

lib/test-worker.js
jamestalmage: Where should it go?

ariporad: Hmm... I'm not sure. I guess it's fine.

lib/test-worker.js
jamestalmage: I didn't modify this line, the only modification I made was the one discussed below (saving to disk in a different thread and loading here).

lib/caching-precompiler.js
```diff
@@ -0,0 +1,95 @@
...
+		// Extract existing source maps from the code.
+		var sourceMap = convertSourceMap.fromSource(code) || convertSourceMap.fromMapFileSource(code, path.dirname(filename));
+
+		return {
+			presets: [presetStage2, presetES2015],
+			plugins: [powerAssert, transformRuntime],
+			filename: filename,
+			sourceMaps: true,
+			ast: false,
+			inputSourceMap: sourceMap && sourceMap.toObject()
```
sindresorhus: Make sure to include the `babelrc: false` thing here.

lib/caching-precompiler.js
```diff
@@ -0,0 +1,95 @@
...
+		};
+	}
+
+	return function (code, filename, hash) {
+		code = code.toString('utf8');
```
sindresorhus: `'utf8'` is moot as it's the default

test/fork.js
```diff
@@ -33,6 +43,7 @@ test('resolves promise with tests info', function (t) {
 		});
 });
 
+/*
```
sindresorhus: ?

lib/caching-precompiler.js
```diff
@@ -0,0 +1,95 @@
...
+CachingPrecompiler.prototype._createTransform = function () {
+	var self = this;
+
+	return cachingTransform({
+		factory: this._factory.bind(this),
+		cacheDir: self.cacheDir,
+		salt: new Buffer(JSON.stringify({
+			'babel-plugin-espower': require('babel-plugin-espower/package.json').version,
+			'ava': require('../package.json').version,
+			'babel-core': require('babel-core/package.json').version
```
sindresorhus: Should probably salt with the Babel plugins too: https://github.com/sindresorhus/ava/pull/390/files#diff-b72e3e25f31d6b3c3e5f40b5991dc5f6L157

cli.js
```diff
@@ -40,6 +40,7 @@ var cli = meow([
 	'  --init       Add AVA to your project',
 	'  --fail-fast  Stop after first test failure',
 	'  --serial     Run tests serially',
+	'  --no-cache   Disable the transpiler cache',
```
sindresorhus: Can you put this at the bottom? I'm going for sorting by importance.

sindresorhus:
Woo, nice improvement on cold start! :+1: 

Generally looks good to me. Could use a test though.

@vdemedes ?

test/fork.js
jamestalmage: Oh,
I forgot.
I think this test can be dropped.
There is no longer a way to test a non-existent test file passed to `fork`, because `fork` now requires the file be precompiled. There is already a test for a non-existing files in `api.js`

sindresorhus: Ok, just drop it.

lib/caching-precompiler.js
jamestalmage: Those are just presets, and don't reflect the actual versions of the underlying plugins (they use caret ranges). While it won't break anything to salt with preset versions, it may be providing a false sense of security.

@thejameskyle - Any thoughts on this? Last time I browsed the babel source, it seemed you were only including the version of `babel-core` in the salt for the cache key.

jamestalmage:
> could use a test though.

Yeah. How far do you want me to take it? For `nyc`, we run every test twice (once with caching on, once with caching off). That was easy to pull off with `nyc` due to how we build our test suite. 

This is what I am thinking for tests:

1. Test a fixture project with caching on. Once the test completes ensure there are X number of files in the cache directory.
2. Delete the cache and test the same fixture from `1` again. Ensure the cache stays empty.
3. Duplicate some of our source maps test with caching off? I might need @novemberborn's help here since I know next to zero about how a good source map test should look.

lib/caching-precompiler.js
sindresorhus: @jamestalmage Relevant https://phabricator.babeljs.io/T6709#67067

sindresorhus:
I don't think we should run all tests with and without. Just need some integration tests to ensure it's working correctly. What you've proposed sounds fine.

lib/caching-precompiler.js
jamestalmage: I've replied to that thread: https://phabricator.babeljs.io/T6709#70444

@thejameskyle - Happy to contribute a PR to babel if you want.

thejameskyle: There's really no good way of caching based on plugins, your solution is full of holes. The current recommendation is for users to wipe out their cache when they change their config.

jamestalmage: @sindresorhus 

Maybe we should default to caching off, and just use the "compile in main thread" optimization. That appears to give us 95% of the benefit anyways.

lib/caching-precompiler.js
sindresorhus: I prefer being explicit and use `new` for classes. ES2015 seems to agree with me.

jamestalmage: OK, changed it back to the original behavior (throw if they forget to use new).

api.js
```diff
@@ -137,6 +144,11 @@ Api.prototype.run = function () {
 				return Promise.reject(new AvaError('Couldn\'t find any files to test'));
 			}
 
+			var cache = self.options.cache !== false;
```
vdemedes: I'd use `isCacheEnabled`, something more verbose & logical given the boolean variable value.

api.js
```diff
@@ -137,6 +144,11 @@ Api.prototype.run = function () {
...
+			var cacheDir = (cache && findCacheDir({name: 'ava', files: files})) || uniqueTempDir();
```
vdemedes: Maybe split this into multiple lines? Does not look clean when everything is on one.

vdemedes:
I really dig the perf improvement! I commented on a few things, but never mind. We can do a cleanup later, I'd rather have this PR merged ASAP.

api.js
jamestalmage: How about just `cacheEnabled`?

vdemedes: Or that, yeah

jamestalmage:
OK, Incorporated the last round of feedback from @vdemedes. Once CI passes again, I intend to merge.

jamestalmage:
Landed.

PR #395, opened 2015-12-31
Regression: Assertions should return promises.

jamestalmage:
This fixes that regression, and extends the promise returning behavior to all assertions.

This would be useful in `async` test functions to bail on the remaining tests if some precondition fails. Simply `await` the assertion.

See: https://github.com/sindresorhus/delay/pull/10


sindresorhus:
I'm confused when you would need this now that `t.throws`/`t.doesNotThrow` supports waiting for the promise? Can you provide an example?

PR #400, opened 2016-01-02
Rewrite all `require('babel-runtime/**)` statements to absolute paths

jamestalmage:
I think messing with the module path is a bad idea, it seems like an easy way to create hard to reproduce bugs.

Also, if a user installs their own version of `babel-runtime`, we have no guarantee it is compatible with the one needed by our tests. By rewriting to absolute paths, we can completely bypass whatever version the user installs locally.

This PR adds a babel-plugin which rewrites all `babel-runtime` requires to absolute paths.

### DIFF
--- a/lib/caching-precompiler.js
+++ b/lib/caching-precompiler.js
@@ -25,6 +25,7 @@ CachingPrecompiler.prototype._factory = function (babelConfig, cacheDir) {
 	var presetES2015 = require('babel-preset-es2015');
 	var transformRuntime = require('babel-plugin-transform-runtime');
 
+	var rewriteRuntime = this._createRewritePlugin();
 	var powerAssert = this._createEspowerPlugin(babel);
 
 	function buildOptions(filename, code) {
@@ -49,7 +50,7 @@ CachingPrecompiler.prototype._factory = function (babelConfig, cacheDir) {
 		});
 
 		options.plugins = options.plugins || [];
-		options.plugins.push(powerAssert, transformRuntime);
+		options.plugins.push(powerAssert, transformRuntime, rewriteRuntime);
 
 		return options;
 	}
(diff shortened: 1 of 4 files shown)


ariporad:
@jamestalmage: I think that it would be a much simpler if we just modified the babel-runtime plugin. This seems like something that could easily fall apart.

jamestalmage:
>  I think that it would be a much simpler if we just modified the babel-runtime plugin. 

`babel-transform-runtime` is not under our control, and I doubt the Babel team would ever accept a change that used absolute paths, the current behavior is intentional on their part.

> This seems like something that could easily fall apart.

You will need to defend that statement with clear examples of how.

sindresorhus:
> I think messing with the module path is a bad idea

:+1:

sindresorhus:
@jamestalmage Anything blocking merging this other than the merge conflict?

jamestalmage:
Only if there is a valid reason not to perform this transform. I don't see it causing any problems, and I think it solves other (i.e. they can `require('bluebird')` in `npm@2` even if they don't have `bluebird` installed).

I will look into rebasing this over the weekend.

novemberborn:
:+1:

sindresorhus:
@jamestalmage Any chance you could rebase this? This PR have been open forever and would be nice to get it resolved :)

novemberborn:
Sounds like a reasonable idea. We could then [wrap `babel-runtime` to reduce installed package size](https://github.com/sindresorhus/ava/issues/369#issuecomment-190741478) too.

jamestalmage:
rebased

sindresorhus:
LGTM

sindresorhus:
> We could then wrap babel-runtime to reduce installed package size too.

@novemberborn Could you open an issue about this?

PR #402, opened 2016-01-02
add endpoint testing example to FAQ

### LINKED ISSUE #310: Recommended pattern for testing REST API endpoints?
How would you recommend we test endpoints where we need to send http requests? Is this on the roadmap or is there another package that plugs in nicely? I know of other packages like supertest, chakram, and frisby, but nothing seems to have an API as beautiful as ava :smile:

mattkrick:
fixes #310
full example seen here: https://github.com/mattkrick/meatier/blob/master/src/server/controllers/__tests__/auth-tests.js

### DIFF
--- a/docs/recipes/endpoint-testing.md
+++ b/docs/recipes/endpoint-testing.md
@@ -1,0 +1,24 @@
+#Endpoint Testing
+
+AVA doesn't have an official assertion library for endpoints, but a great option is [`supertest-as-promised`](https://github.com/WhoopInc/supertest-as-promised).
+Since the tests run concurrently, it's a best practice to create a fresh server instance for each test because if we referenced the same instance, it could be mutated between tests. This can be accomplished with a `beforeEach` and `context`, or even more simply with a factory function:
+```
+function makeApp() {
+  const app = express();
+  app.post('/signup', signupHandler);
+  return app;
+}
+```
+
+Next, just inject your server instance into supertest. The only gotcha is to use a promise or async/await syntax instead of supertest's `end` method:
+```
+test('signup:Success', async t => {
+  t.plan(2);
+  const app = makeApp();
+  const res = await request(app)
+    .post('/signup')
+    .send({email: 'ava@rocks.com', password: '123123'})
+  t.is(res.status, 200);
+  t.is(res.body.email, 'ava@rocks.com');
+});
+```
--- a/readme.md
+++ b/readme.md
@@ -20,6 +20,7 @@ Translations: [Español](https://github.com/sindresorhus/ava-docs/blob/master/es
 - [API](#api)
 - [Assertions](#assertions)
 - [FAQ](#faq)
+- [Recipes](#recipes)
 
 
 ## Why AVA?
@@ -698,6 +699,10 @@ AVA, not Ava or ava. Pronounced [`/ˈeɪvə/` ay-və](media/pronunciation.m4a?ra
 ### Concurrency vs. parallelism
 
 Concurrency is not parallelism. It enables parallelism. [Learn more.](http://stackoverflow.com/q/1050222)
+
+## Recipes
+
+- [Endpoint testing](/docs/recipes/endpoint-testing.md)
 
 
 ## Support


readme.md
```diff
@@ -660,6 +660,30 @@ As of version `5.0.0` it uses source maps to report coverage for your actual cod
 
 ## FAQ
 
+### How do I test HTTP endpoints?
+
+AVA doesn't have an official assertion library for endpoints, but a great option is [`supertest-as-promised`](https://www.npmjs.com/package/supertest-as-promised).
+Assuming you are using express, you might set up a server for each test:
+```
+let app;
+test.beforeEach(() => {
```
jamestalmage: This should either be in a `before`, or should use `t.context` over a closure:

```js
test.beforeEach(t => {
   const app = express();
   app.post('/signup', signupHandler);
   t.context.app = app;
});

test('signup: Success', async t => {
  const res = await request(t.context.app)
    .post('/signup')
    ...
});
```

jamestalmage: or

```js
let app;
test.before(() => {
  app = express();
  app.post('signup', signupHandler);
});
```

jamestalmage: However my new preference is just to do async safe helper functions:

```js
function app() {
  const application = express();
  application.post('/signup', signupHandler);
  return application;
}

test('signup: Success', async t => {
  const res = await request(app())
    .post(...)
    ...   
});

Comment (mattkrick) [inline readme.md]: good to know, thanks! a `before` would probably suffice, but I wasn't sure if it was a "best practice" since someone could mutate `app` within a test.

if you don't mind, could you explain why `context` is needed & a `let` doesn't suffice? I'm guessing it's because the `let` won't carry through to all the child processes?

Comment (jamestalmage) [inline readme.md]: > could you explain why context is needed 

Because everything executes concurrently in AVA. Assuming you are coming from mocha, you are accustomed to this order of execution:

 - `beforeEach` for  `test 1`
 - `test 1`
 - `beforeEach` for `test 2`
 - `test 2`
 - `beforeEach` for `test 3`
 - `test 3`

With AVA, your order of operations could be:

 - `beforeEach` for `test 2`
 - `beforeEach` for  `test 1`
 - `test 1`
 - `beforeEach` for `test 3`
 - `test 2`
 - `test 3`

Note it is unlikely the order is exactly as above, the point I'm making is that you can't rely on order of execution. The only ordering you are guaranteed is that `beforeEach for test X` happens before `test X`. Where it happens in relation to other tests and their respective `beforeEach`es is not guaranteed.

If you use the `--serial` flag (`ava --serial my-test.js`), then your tests are ordered similar to mocha.

Comment (mattkrick) [inline readme.md]: ahhhh, that kinda clears it up. I was thinking that `beforeEach` was guaranteed to run immediately before the test *on that thread*. So is `beforeEach` guaranteed to run on the same thread as the test?

PS, I like your 3rd option of creating the app, I'll update the readme to reflect it

Comment (jamestalmage) [inline readme.md]: Each test file is it's own process (or thread if you want to think of it that way).

`t.context` is shared between your `beforeEach` and the appropriate `test`. So in the above example, even though `beforeEach` is executed 3 times, it gets a unique `t.context` each time that is shared with just a single `test` run. If you are using a closure to pass data between a `beforeEach` and a `test`, that is almost certainly wrong, and should use `t.context` as well.

It's perfectly acceptable to use a closure to pass data created in `t.before` (in fact `t.context` does not work in `t.before`).

**Comment by mattkrick** [comment]
@jamestalmage thanks for your thoughtful explanation! Let me know if it looks good & I'll squash the commits.

**Comment by jamestalmage** [comment]
:+1: LGTM

If you can figure out a **concise** way of communicating what I've said in this discussion, that would be a welcome PR as well. It's a point of confusion for a lot of people.

**Comment by mattkrick** [comment]
rebased & rewritten to cover our discussion, cheers

**Comment by jamestalmage** [comment]
Nice, latest changes are even better.

**Comment by sindresorhus** [comment]
I think this should be in a `docs/recipes/endpoint-testing.md` linked to from the readme. The readme is already pretty long.

File: readme.md [hunk 2]
```diff
@@ -660,6 +660,31 @@ As of version `5.0.0` it uses source maps to report coverage for your actual cod
 
 ## FAQ
 
+### How do I test HTTP endpoints?
+
+AVA doesn't have an official assertion library for endpoints, but a great option is [`supertest-as-promised`](https://www.npmjs.com/package/supertest-as-promised).
```
sindresorhus: Link to the GitHub repo, not npm page.

readme.md
```diff
@@ -660,6 +660,31 @@ As of version `5.0.0` it uses source maps to report coverage for your actual cod
...
+Since the tests run concurrently, it's a best practice to create a fresh server instance for each test because if we referenced the same instance, it could be mutated between tests. This can be accomplished with a `beforeEach` and `context`, or even more simply with a factory function:
+```
+function makeApp() {
+  const app = express();
+  app.post('/signup', signupHandler);
+  return app;
+}
+```
+
+Next, just inject your server instance into supertest. The only gotcha is to use a promise or ES7 syntax instead of supertest's `end` method:
```
sindresorhus: What does ES7 syntax mean? Be more specific.

jamestalmage: `async / await`

sindresorhus: Then I would write that instead. ES7 doesn't mean anything, it's ES2016, but async functions are not yet part of it.

jamestalmage: yep - I was recommending language.

jamestalmage:
@mattkrick - Do you have time to update this PR with the recommended changes?

mattkrick:
yep, will get a PR out today, sorry for the delay!

mattkrick:
ok, should be good to go

jamestalmage:
@mattkrick - One last question. I am not familiar enough with the supertest API to tell if the code is correct or not. Can you verify you have successfully tested IRL with code that looks similar to the example provided here?

mattkrick:
I like how you think!
Yep, it works, but don't take my word for it. Here it is in a repo: https://github.com/mattkrick/meatier/blob/master/src/server/controllers/__tests__/auth-tests.js
Feel free to give it a download & run `npm run test`.

jamestalmage:
:+1: LGTM

docs/recipes/endpoint-testing.md
```diff
@@ -0,0 +1,23 @@
+#Endpoint Testing
+AVA doesn't have an official assertion library for endpoints, but a great option is [`supertest-as-promised`](https://github.com/WhoopInc/supertest-as-promised).
```
kevva: Nitpick, but an empty line between these.

docs/recipes/endpoint-testing.md
```diff
@@ -0,0 +1,23 @@
...
+Since the tests run concurrently, it's a best practice to create a fresh server instance for each test because if we referenced the same instance, it could be mutated between tests. This can be accomplished with a `beforeEach` and `context`, or even more simply with a factory function:
+```
```
kevva: Add empty line.

docs/recipes/endpoint-testing.md
```diff
@@ -0,0 +1,23 @@
...
+function makeApp() {
+  const app = express();
+  app.post('/signup', signupHandler);
+  return app;
+}
+```
+
+Next, just inject your server instance into supertest. The only gotcha is to use a promise or async/await syntax instead of supertest's `end` method:
+```
```
kevva: Add empty line.

mattkrick: can do, PS anyone know a way to squash commits within github without doing the pull/rebase/push dance?

kevva: I don't think that's possible afaik.

sindresorhus:
Thank you @mattkrick :)

mycaule:
Could you please provide a code sample for `signupHandler`

I am trying to test a google cloud functions using the technique described in [the endpoint testing recipe](https://github.com/avajs/ava/blob/master/docs/recipes/endpoint-testing.md) but have an error when trying to follow the recipe.

`Route.post() requires a callback function but got a [object Undefined]`

PR #403, opened 2016-01-02
Make build badges more clear about their scope

MoOx:
"unix"/"window" labels seems to be more clear that just a weird additional icon for appveyor.
Also when not in a title, badges can be one per lines, since Markdown handle new lines as... nothing (unless you add 2 spaces before it, which add a br).

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -2,7 +2,10 @@
 
 > Futuristic test runner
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
Thanks, but I think the current is fine.

Issue #405, opened 2016-01-04
Add a config file.


jamestalmage:
Probably just an `ava` section in `package.json`.

The available options are getting complex enough that I think we are close to needing this (if we aren't there already).

```json
{
  "serial": true,
  "verbose": true,
  "cache": false
}
```

These would need to be merged with whatever is specified on the command line (command line taking precedence).

The goal is to be able to specify the default way you want it to behave when you simply type `ava` on the command line.

sindresorhus:
:+1: I do that with [XO](https://github.com/sindresorhus/xo#config) too. I would prefer only a section in package.json. Makes things simple.

https://github.com/sindresorhus/pkg-conf

ariporad:
Agreed. I'd also like to add a "files" section, and have it be "config.ava", vs. "ava "

sindresorhus:
> I'd also like to add a "files" section

:+1: 

> and have it be "config.ava", vs. "ava "

:-1: That's a convention nobody uses. Top-level is simpler and it will never conflict with anything anyways.

jamestalmage:
> That's a convention nobody uses

Except for https://github.com/bcoe/nyc/issues/131  :smile:

jamestalmage:
> I'd also like to add a "files" section

Yeah, I think basically any config value is fair game. I can't think of one we *shouldn't* allow them to configure this way.

vdemedes:
Great suggestion, I'm on board!

ariporad:
@sindresorhus: Except `nyc` and `comitizen` (off the top of my head).

jamestalmage:
`nyc` is falling off that list very soon: https://github.com/bcoe/nyc/pull/132

kentcdodds:
Add [`ghooks`](npm.im/ghooks) to the list of modules that use the config convention :-) But meh, it'll never conflict.

PR #407, opened 2016-01-04
Allow configuration via `package.json`

jamestalmage:
Fixes: #405

- [x] Probably needs a test or two.

- [x] Definitely needs documentation.


jamestalmage:
Gotta say, didn't think this was only going to end up being 4 LOC. Small reusable modules FTW!

vdemedes:
Damn it, was about to send the same PR! :D

screen shot 2016-01-04 at 10 37 20 am

cli.js
```diff
@@ -75,7 +79,7 @@ if (cli.flags.init) {
 	return;
 }
 
-var api = new Api(cli.input, {
+var api = new Api(cli.input.length ? cli.input : arrify(conf.files), {
```
vdemedes: How about:

```js
cli.input.concat(conf.files || [])
```

jamestalmage: because then you can't focus on a single file.

jamestalmage:
> was about to send the same PR!

I assigned myself before I started work. Sorry!

vdemedes:
Then it's totally my oversight, no worries ;)

cli.js
vdemedes: True

kevva: I don't see the difference.

```js
conf.files = 'foo.js';

cli.input.concat(conf.files);
//=> ['foo.js']

arrify(conf.files);
//=> ['foo.js']
```

kevva: Maybe you meant that `ava just-testing.js` always should take precedence if you want to quick test something and not bother having to update `package.json` :P.

jamestalmage:
Not sure the best way to test this.

I'm thinking just fixtures with tests that use `has-flag` to see what gets passed down to the child process.

Thoughts?

cli.js
vdemedes: @kevva Yeah, I think what @jamestalmage meant, was if we have:

package.json
```js
"ava": {
  "files": ["a.js", "b.js"]
}
```

and we want to test just b.js:

```
$ ava b.js
```

jamestalmage: Yep - that's what I was trying to say.

jamestalmage:
@vdemedes @sindresorhus - squashed and ready for review.

vdemedes:
Love it!

cli.js
```diff
@@ -32,6 +33,8 @@ var Api = require('./api');
 // Bluebird specific
 Promise.longStackTraces();
 
+var conf = pkgConf.sync('ava') || {};
```
sindresorhus: pkgConf already returns an empty object when not found.

readme.md
```diff
@@ -127,6 +128,28 @@ Directories are recursive by default. Files in directories named `fixtures` and
...
+    "serial": true,
+    "tap": true,
+    "verbose": true,
+    "require": ["babel-core/register", "coffee-script/register"]
+  }
+}
+```
+
+Arguments passed to the CLI will always take precedence over the configuration in `package.json`.
+
```
sindresorhus: Maybe mention the benefit of specifying config here instead of the CLI is that it will be used when you run it with `$ ava`.

readme.md
```diff
@@ -127,6 +128,28 @@ Directories are recursive by default. Files in directories named `fixtures` and
...
+All of the CLI options can be configured in the `ava` section of your `package.json`:
+
+```json
+{
+  "ava": {
+    "files": [
+      "my-test-folder/*.js",
+      "!**/not-this-file.js"
+    ],
+    "fail-fast": true,
```
sindresorhus: Should be camelCased

jamestalmage: I'd need to decimalize all the keys before feeding into `meow.options.defaults`, so... [`decamilize-keys`](https://www.npmjs.com/package/decamelize-keys)?

jamestalmage: Or file a meow PR that decimalizes defaults for you?

sindresorhus: Yeah, I think it would be useful if meow did that for us.

sindresorhus: Or you could just use objectAssign on cli.flags instead of using the minimist default thing.

sindresorhus: @jamestalmage New version of `meow` is out. Just bump to `3.7.0`.

readme.md
jamestalmage: I added it above the example.

jamestalmage:
All recommended changes made.

package.json
```diff
@@ -105,9 +105,10 @@
     "loud-rejection": "^1.2.0",
     "max-timeout": "^1.0.0",
     "md5-hex": "^1.2.0",
-    "meow": "^3.6.0",
+    "meow": "^3.7.0",
```
ariporad: This is a bit out of scope, but it shouldn't really be a problem.

ariporad:
@jamestalmage: one quick thing: require on the command line shouldn't override the package.json, it should add to it? (Sorry for no example, on mobile)

package.json
jamestalmage: Actually, it's require for this PR to work. https://github.com/sindresorhus/meow/pull/25

jamestalmage:
> require on the command line shouldn't override the package.json, it should add to it

Nah, then you can't be explicit about ordering. If we append it then you are stuck if you wanted it prepended, and visa-versa. If you have a complicated `--require` setup that is a pain to edit on the command line, just edit it directly in `package.json`.

sindresorhus:
Looks good :D

sindresorhus:
> Gotta say, didn't think this was only going to end up being 4 LOC. Small reusable modules FTW!

unicorn

ariporad:
@sindresorhus: do you know about 'npm I unicorns'?

sindresorhus:
@ariporad ?

ariporad:
@sindresorhus: try it ;). It's awesome.

kentcdodds:
Shouldn't this be part of the [`config`](https://docs.npmjs.com/files/package.json#config) object in package.json? Or am I missing something?

jamestalmage:
> Shouldn't this be part of the config object in package.json? Or am I missing something?

That got debated here: https://github.com/sindresorhus/ava/issues/405#issuecomment-168569929

The only advantage to using the config object is for build scripts, so they can easily access the config via the `npm_package_config_` prefixed environment variables. There's no advantage for AVA, because we want you to be able to call `ava` directly - not be forced to always run via `npm run` scripts.

PR #409, opened 2016-01-07
use is-equal instead of deeper for es6 set and map support

mayankchd:


mayankchd:
Failing on Node.js 0.10 as it doesn't support Set and Map

jamestalmage:
For Node 0.10, lookup how Set and Map are referenced when transformed by AVA, should be something like:

```
var Map = require('babel-runtime/lib/fn/Map');
// or 
var Map = require('core-js/lib/fn/Map');
```

sindresorhus:
I would prefer to just add support for Map/Set in `deeper`. `is-equal` brings in 3 MB of dependencies.

sindresorhus:
Closing in favor of https://github.com/othiym23/node-deeper/issues/2.

jamestalmage:
:+1: I agree. `2.3MB` is just way to big. AVA already takes just this side of eternity to download. @twada and others have put in major effort slimming down their packages and our download size. This would be a major step in the wrong direction. 

Sorry @mayankchd. Otherwise your PR looked great. It's not your fault that package is so bloated.

mayankchd:
Thanks! I like ava a lot!

PR #410, opened 2016-01-08
Add skipped test count to builtin reporters

sotojuan:
This fixes https://github.com/sindresorhus/ava/issues/408 by adding the skipped test count to the builtin reporters, in yellow color.

Let me know if I missed anything or if another color is more appropriate.

PR #412, opened 2016-01-09
Standardize colors across reporters.

jamestalmage:
PR #410 missed a color change on a single log line,
See: https://github.com/sindresorhus/ava/pull/410#issuecomment-170210415

Using this convention will make it much easier to review PR's. No need to remember what color is used where, just ensure they don't use chalk directly, and alias the color with it's intended purpose in the colors module.


cli.js
```diff
@@ -21,9 +21,9 @@ var updateNotifier = require('update-notifier');
 var figures = require('figures');
 var arrify = require('arrify');
 var meow = require('meow');
-var chalk = require('chalk');
 var Promise = require('bluebird');
 var pkgConf = require('pkg-conf');
+var colors = require('./lib/reporters/colors');
```
vdemedes: `lib/reporters` was intended to contain files of reporters only. `lib/colors.js` would be just fine as well

jamestalmage: fixed

sindresorhus:
LGTM

kevva:
:+1:

PR #414, opened 2016-01-09
Fix t.throws/doesNotThrow not outputting failure reason

therealklanni:
If a Promise is rejected with a string, `t.throws()` fails the test with "undefined undefined undefined" instead of the label of the assertion or any other identifying information, leaving the dev writing the tests unable to easily pinpoint which test has just failed.

Reproduce by writing a `t.throws()` assertion for a Promise that rejects simply using a string and have a RegExp validation rule.

Before:
```
  1 failed

  1. rejects
  undefined undefined undefined
  AssertionError: undefined undefined undefined
```

After:
```
  1 failed

  1. rejects
  Invalid due date 19791224T245051Z
  AssertionError: Invalid due date 19791224T245051Z
```

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -75,7 +75,11 @@ x.throws = function (fn, err, msg) {
 				x.throws(noop, err, msg);
 			}, function (fnErr) {
 				x.throws(function () {
-					throw fnErr;
+					if (fnErr.message) {
+						throw fnErr;
+					} else {
+						throw new Error(fnErr);
+					}
 				}, err, msg);
 			});
 	}
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -75,7 +75,7 @@ x.throws = function (fn, err, msg) {
 				x.throws(noop, err, msg);
 			}, function (fnErr) {
 				x.throws(function () {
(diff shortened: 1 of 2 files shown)


sindresorhus:
Can you add a test? So we can ensure it doesn't regress.

therealklanni:
Certainly

therealklanni:
Oh, I added it in assert.js before I saw your edit.

lib/assert.js
```diff
@@ -75,7 +75,11 @@ x.throws = function (fn, err, msg) {
 				x.throws(noop, err, msg);
 			}, function (fnErr) {
 				x.throws(function () {
-					throw fnErr;
+					if (fnErr.message) {
```
vdemedes: I think testing if `fnErr instanceof Error` would be better here.

therealklanni: Fair enough. I'll update.

sindresorhus:
Thank you @therealklanni :)

vdemedes:
Thank you!

therealklanni:
Np, guys, thanks for this great tool! :metal:

PR #418, opened 2016-01-13
Use stack-utils for nicer stack traces.

jamestalmage:
I have extracted `node-tap`s stack cleaning algorithm into it's own module, [`tapjs/stack-utils`](https://github.com/tapjs/stack-utils).

It creates much nicer stack traces than `beautify-stack.js` does. There is a bit of work left to do in the `stack-utils` module, but once these are done, this should be mergeable:

 - [x] https://github.com/tapjs/stack-utils/issues/8 - decide what should be considered `internal`
 - [x] https://github.com/tapjs/stack-utils/pull/6 - clean long stack traces

Screenshots below.

### DIFF
--- a/lib/beautify-stack.js
+++ b/lib/beautify-stack.js
@@ -1,13 +1,18 @@
 'use strict';
 
-function beautifyStack(stack) {
-	var re = /(?:^(?! {4}at\b).{6})|(?:\((?:[A-Z]:)?(?:[\\\/](?:(?!node_modules[\\\/]ava[\\\/])[^:\\\/])+)+:\d+:\d+\))/;
-	var found = false;
+var StackUtils = require('stack-utils');
+var debug = require('debug')('ava');
 
-	return stack.split('\n').filter(function (line) {
-		var relevant = re.test(line);
-		found = found || relevant;
-		return !found || relevant;
+var stackUtils = new StackUtils({
+	internals: debug.enabled ? [] : StackUtils.nodeInternals().concat([
+		/\/ava\/(?:lib\/)?[\w-]+\.js:\d+:\d+\)?$/,
+		/\/node_modules\/(?:bluebird|empower-core)\//
+	])
+});
+
+function beautifyStack(stack) {
+	return stack.split('\n')[0] + '\n' + stackUtils.clean(stack).split('\n').map(function (s) {
+		return '    ' + s;
 	}).join('\n');
 }
 
--- a/package.json
+++ b/package.json
@@ -120,6 +120,7 @@
     "serialize-error": "^1.1.0",
     "set-immediate-shim": "^1.0.1",
     "source-map-support": "^0.4.0",
+    "stack-utils": "^0.3.0",
     "strip-bom": "^2.0.0",
     "time-require": "^0.1.2",
     "unique-temp-dir": "^1.0.0",
--- a/test/reporters/mini.js
+++ b/test/reporters/mini.js
@@ -129,7 +129,7 @@ test('results with passing tests and rejections', function (t) {
 	t.is(output[2], '');
 	t.is(output[3], '  ' + chalk.red('1. Unhandled Rejection'));
 	t.match(output[4], /Error: failure/);
-	t.match(output[5], /Test\.test/);
+	t.match(output[5], /test\/reporters\/mini\.js/);
 	t.end();
 });
 
@@ -152,7 +152,7 @@ test('results with passing tests and exceptions', function (t) {
 	t.is(output[2], '');
 	t.is(output[3], '  ' + chalk.red('1. Uncaught Exception'));
 	t.match(output[4], /Error: failure/);
-	t.match(output[5], /Test\.test/);
+	t.match(output[5], /test\/reporters\/mini\.js/);
 	t.end();
 });
 
@@ -174,6 +174,6 @@ test('results with errors', function (t) {
 	t.is(output[2], '  ' + chalk.red('1. failed'));
 	t.match(output[3], /failure/);
 	t.match(output[4], /Error: failure/);
-	t.match(output[5], /Test\.test/);
+	t.match(output[5], /test\/reporters\/mini\.js/);
 	t.end();
 });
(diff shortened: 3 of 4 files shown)


jamestalmage:
Before: 

screenshot 2016-01-13 12 19 18

After: 

screenshot 2016-01-13 13 21 22

jamestalmage:
Before:

screenshot 2016-01-13 12 20 27

After:

screenshot 2016-01-13 13 16 38

jamestalmage:
Before:

screenshot 2016-01-13 13 09 57

After:

screenshot 2016-01-13 13 12 13

vdemedes:
Love love love this!

package.json
```diff
@@ -118,6 +118,7 @@
     "serialize-error": "^1.1.0",
     "set-immediate-shim": "^1.0.1",
     "source-map-support": "^0.4.0",
+    "stack-utils": "github:tapjs/stack-utils#long-stack-traces",
```
vdemedes: No need for `github:` prefix, it will work the same without it.

jamestalmage:
> Love love love this!

:smile: 

If you could review the issues and PR's on [`stack-utils`](https://github.com/tapjs/stack-utils) that would be a big help.

package.json
jamestalmage: I just did

```
npm i -S tapjs/stack-utils#long-stack-traces
```

and that is how it modified `package.json`. I want to get that PR merged before we merge this anyways.

test/fixture/return-rejected-promise.js
```diff
@@ -0,0 +1,11 @@
+import test from '../../';
+
+var Promise = require('bluebird');
+
+test(() => {
```
jamestalmage: Oops. This fixture isn't used anywhere. I just used it to create some of the above screenshots. It should be deleted.

jamestalmage:
I'm also wondering if we should do something like this:

```js
var opts = {};

// Print the whole stack if debug is enabled.
if (!debug('ava').enabled) {
  opt.internals = StackUtils.internals().concat([ 
     // ...
  ]);
}
```

vdemedes:
Probably adding those when debugging is enabled would be useful:

```diff
+		/[\\\/]ava[\\\/](?:lib[\\\/])?[\w-]+\.js:[0-9]+:[0-9]+\)?$/,
+		/[\\\/]node_modules[\\\/](?:bluebird|empower-core)[\\\/]/
```

jamestalmage:
If debugging is enabled, why not just print everything?

vdemedes:
Or that :)

package.json
sindresorhus: npm apparently adds this now, but it's not needed as github is the default for shorthands.

sindresorhus:
😁

Looks kinda weird to have it double here:

```
false == true
AssertionError: false == true
```

Maybe we can remove the AssertionError?

lib/beautify-stack.js
```diff
@@ -1,13 +1,19 @@
...
+		/[\\\/]node_modules[\\\/](?:bluebird|empower-core)[\\\/]/
+	])
+});
 
-	return stack.split('\n').filter(function (line) {
-		var relevant = re.test(line);
-		found = found || relevant;
-		return !found || relevant;
+function beautifyStack(stack) {
+	return stack.split('\n')[0] + '\n' + stackUtils.clean(stack).split('\n').map(function (s) {
```
sindresorhus: Maybe use https://github.com/sindresorhus/indent-string here

sindresorhus: Why is the first line not indented? And why do we need to manually indent it? Shouldn't `stack-utils` do that for us?

jamestalmage:
> Maybe we can remove the AssertionError?

Sure.  `stack-utils` actually chops the first line of the stack-trace (the message) off anyways. I actually manually add it back in `beautify-stacktrace.js`. The issue is the reporters only show that for assertion errors. `unhandledRejections` and `uncaughtExceptions` would be different. We would just need to rework the reporters so it's their responsibility to print the message.

lib/beautify-stack.js
```diff
@@ -1,13 +1,19 @@
...
-	var re = /(?:^(?! {4}at\b).{6})|(?:\((?:[A-Z]:)?(?:[\\\/](?:(?!node_modules[\\\/]ava[\\\/])[^:\\\/])+)+:\d+:\d+\))/;
-	var found = false;
+var StackUtils = require('stack-utils');
+
+var stackUtils = new StackUtils({
+	internals: StackUtils.nodeInternals().concat([
+		/\(internal\/[\w_-]+\.js:[0-9]+:[0-9]+\)$/,
+		/\(native\)$/,
+		/[\\\/]ava[\\\/](?:lib[\\\/])?[\w-]+\.js:[0-9]+:[0-9]+\)?$/,
+		/[\\\/]node_modules[\\\/](?:bluebird|empower-core)[\\\/]/
```
sindresorhus: Can you normalize the paths in stack traces to forward slash? That way they're consistent on all systems and with the added benefit of simpler regex here.

lib/beautify-stack.js
vdemedes: See https://github.com/tapjs/stack-utils/issues/2#issuecomment-169176569, if I'm not mistaken

sindresorhus:
> We would just need to rework the reporters so it's their responsibility to print the message.

Can you open a new issue about that? It's unrelated to this PR, just came to my mind when looking at your screenshots.

lib/beautify-stack.js
jamestalmage: Actually, `stack-utils` already does that. I'm not sure if it does it before the regexp gets applied, but it probably should.

lib/beautify-stack.js
jamestalmage: Yep. Whitespace at the beginning of any multi-line string makes the yaml output much harder to read. `stack-utils` trims whitespace from the first line.

I guess we could add an `indent` option to `stack-utils`.

ariporad:
:+1:

lib/beautify-stack.js
sindresorhus: > I guess we could add an indent option to stack-utils.

:+1: Can you open an issue on `stack-utils` so we don't forget to simplify this?

sindresorhus:
:+1:

vdemedes:
I think this feature should go into killer-feature list.

sindresorhus:
@vdemedes Go ahead ;)

kevva:
This is super nice.

PR #419, opened 2016-01-13
Add note on required import for ES2015 usage

uhoh-itsmaciek:
I was briefly confused by this until I realized I was missing this import. It's probably since I'm new to node, but so are a lot of other people.

Thanks for a great project!

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -386,7 +386,7 @@ test(t => {
 
 ### ES2015 support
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
That's not needed. We handle that for you.

vdemedes:
> Just write your tests in ES2015. No extra setup needed.

This is actually true, you don't need to import `babel-core/register` to write **tests** in ES2015. I assume your code was crashing, because you wrote *other* code (outside of tests) in ES2015, but using Node.js' version that does not support it.

uhoh-itsmaciek:
Oh interesting. Thanks for the clarification @vdemedes.

PR #432, opened 2016-01-15
Support custom reporter

SimonDegraeve:


jamestalmage:
Good idea, my only concern is this implies a stable reporter API, and I don't think we're ready to commit to that yet.

sindresorhus:
:-1: I don't want to create another silo of reporters. We have two great builtin repoters that should work for most use-cases. For anything else, create a [tap reporter](https://github.com/sindresorhus/ava#optional-tap-output).

vdemedes:
We appreciate your efforts, @SimonDegraeve, but we have to reject this PR. That's the exact reason why we worked on TAP support. So that users would be able to customize output with a dozen of already available TAP reporters.

You can check out the list of the most popular TAP reporters here: https://github.com/sindresorhus/awesome-tap#reporters

SimonDegraeve:
No problem, make sense.

yatki:
I wanted to create an html reporter like [mochawesome](https://github.com/adamgruber/mochawesome). I thought maybe i can write an transformer for converting tap output to mocawesome json structure. I searched for a solution all day but the output i get from the tap reporter is really primitive . For instance, I even couldn't find a way to get duration per test which i could easily get with a custom reporter. I'm not talking about getting executed script in the test, or it's context for attaching custom messages to output.

What is bad about letting people to write their own custom reporters? @jamestalmage mentioned a stable reporter API. It looks pretty stable to me :)

Can't you reconsider to support **proper** custom reporters?

novemberborn:
@yatki whilst I agree that the TAP output isn't as useful as we'd might like, I'm still reluctant to support custom reporters within AVA. I'd be interested in providing high-fidelity [NDJSON](http://ndjson.org/) reports, properly specified and implementable by other test runners, that custom reporters could then be build on top of. However we currently do not have the bandwidth to even begin undertaking such a project. Similarly we don't want to support more reporters within AVA.

yatki:
@novemberborn Thank you very much for the reply. Having NDJSON reports sounds a really cool. I'm really happy using AVA so far and totally thankful for your efforts. I still didn't understand why you don't want to support more reporters within AVA, but that's ok. I don't want to take more of your time, I'll go with a fork for now. 

But as far as i'm concerned supporting custom reporters seems not a big effort and this would absolutely give no harm or disadvantage to AVA. Authors of custom reporters are responsible from quality/performance of their reporters, and developers who use those custom reporters can always go back to use native AVA reporters.

But anyway. Have a great day. Thanks.

sindresorhus:
We could consider including more AVA specific info in the YAML data block in the TAP output.

yatki:
Gentlemen, you had my curiosity… but now you have my attention.

How would you like to proceed to define requirements and when would you like to start? Maybe instead of working on a solution which is gonna work only for me, I can contribute on this? Would you like to proceed with a separate issue ?

novemberborn:
> We could consider including more AVA specific info in the YAML data block in the TAP output.

@sindresorhus, we could, but to what end? Reporters need to be written specifically for AVA. We'd probably want to scope it under an `"ava"` key anyway.

---

@yatki are you referring to an NDJSON report spec?

yatki:
@novemberborn I was referring to additional TAP information actually. Because i thought It could be faster to implement. But I could be interested to contribute on NDJSON report specs too. I just don't know how data structure should look like.

novemberborn:
@yatki I'm just not sure it's worth the trouble of adding extra, custom information to the TAP output. Third-party tools need to recognize our properties, and we'd have to document them, and we have to maintain compatibility. It's a lot of trouble, and the resulting tool chain would be quite brittle.

And though the answer may be ["we need a new standard"](https://xkcd.com/927/) that too is a lot of trouble, does not guarantee interoperability, requires effort from third-party tools, requires us to maintain compatibility, etc… still, it seems like a more solid approach for the long term.

yatki:
@novemberborn I totally agree on that is not an easy task. 

However, I'd like to have more info about the tests that we are running not just to have a prettier results but also for statistics. For instance, which tests are constantly taking more time or failing more often. Also we need to integrate the process with monitoring tools to see running tests in real time because we have different projects running different tests suites and console outputs are not enough. These features are not things that I'm making up right now. We really need these and we had them with previous testing frameworks.

If there is a way to detailed test info and log it with t.log(), I'm even willing to write my own functions to put additional info to tap output and write a script to parse them (until ava have a better solution). But i couldn't find a way to do it. Or I'll go with the first idea. I'll fork the project and implement my own reporter which is going to bring a bit pain, in terms of keeping ava updated.

But to keep it short: ava report outputs have to be improved or ava should support custom reporters and let people to collect whatever information they like. Having a powerful testing framework like ava, but not having control on results is just unacceptable :)

novemberborn:
> I'll fork the project and implement my own reporter which is going to bring a bit pain, in terms of keeping ava updated.

I appreciate your needs, @yatki, but given how the reporters are currently implemented, maintaining this is currently too much of a burden. We have more important issues to resolve so we can get a 1.0 release done. It'll be good to learn from your experiences though.

> For instance, which tests are constantly taking more time or failing more often.

Failure rates you could do using the TAP reporter. Timing information is a lot trickier for the common use cases, given that tests are run concurrently.

PR #434, opened 2016-01-15
Allow files with only skipped tests

kasperisager:
Ensures that skipped tests are also counted when determining whether or not a file contains tests. My own use-case is a multi-file test suite where one or more test files contain nothing but skipped tests due to them testing functionality yet to be implemented.

### DIFF
--- a/index.js
+++ b/index.js
@@ -69,8 +69,9 @@ function exit() {
 
 globals.setImmediate(function () {
 	var numberOfTests = runner.select({type: 'test', skipped: false}).length;
+	var numberOfSkippedTests = runner.select({type: 'test', skipped: true}).length;
 
-	if (numberOfTests === 0) {
+	if (numberOfTests === 0 && numberOfSkippedTests === 0) {
 		send('no-tests', {avaRequired: true});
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -69,8 +69,9 @@ function exit() {
 
 globals.setImmediate(function () {
 	var numberOfTests = runner.select({type: 'test', skipped: false}).length;
+	var numberOfSkippedTests = runner.select({type: 'test', skipped: true}).length;
```
jamestalmage: Is there a good reason we can't just drop `skipped: false` from the line above?

kasperisager: All tests pass on my end when it's removed.

jamestalmage: So that should be the only `select` you need. @vdemedes - do we need that first value for a reporter?

vdemedes: Yeah we do - https://github.com/sindresorhus/ava/blob/master/index.js#L42

vdemedes:
Thanks @kasperisager! And great commit message btw ;)

kasperisager:
No problem!

PR #436, opened 2016-01-15
Move method chaining to an external module (`option-chain`)

jamestalmage:
See https://github.com/jamestalmage/option-chain

Feedback on the API appreciated


vdemedes:
LGTM! @sindresorhus ?

sindresorhus:
Glad we could simplify this logic out into a module :)

API looks good. I would probably not have done the `spread` option as it's moot with ES2015 and would cause a lot of boilerplate without for now, but doesn't matter.

PR #438, opened 2016-01-15
Default to the verbose reporter on CI servers. Fixes #435

### LINKED ISSUE #435: Improve default reporter on CI.
The default reporter doesn't look great if the terminal is not a `tty` or doesn't obey ansi cursor position codes: 

`node-tap` automatically defaults to the `tap` reporter for non-tty setups. That might make sense for us.

Travis does provide claim a `tty`, but it doesn't honor ansi cursor position codes. [See here](https://travis-ci.org/jamestalmage/option-chain/jobs/102683773#L758). It's not terrible, but the verbose reporter would look better.

jamestalmage:
Fixes #435

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -23,6 +23,7 @@ var arrify = require('arrify');
 var meow = require('meow');
 var Promise = require('bluebird');
 var pkgConf = require('pkg-conf');
+var isCi = require('is-ci');
 var colors = require('./lib/colors');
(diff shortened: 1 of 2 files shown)


jamestalmage:
better: https://travis-ci.org/jamestalmage/option-chain/jobs/102697434#L754

sindresorhus:
:shipit:

PR #442, opened 2016-01-16
add fly-ava plugin

pine613:
fly-ava is ava plugin for [fly](https://github.com/bucaran/fly)
https://github.com/pine613/fly-ava

Thank you


sindresorhus:
Nice! :)

PR #443, opened 2016-01-16
Add shortflags to the CLI

### LINKED ISSUE #437: Add shortflags to the CLI
https://github.com/sindresorhus/ava/issues/435#issuecomment-172090337

- `-t` => `--tap`
- `-v` => `--verbose`
- `-r` => `--require`
- `-s` => `--serial`

Example, something like `alias: { s: 'str' }` should be added below here https://github.com/sindresorhus/ava/blob/ced24c5090ad5d08eb5a2961dc0bb94d00eee445/cli.js#L72 with the shortflag mapping.

sotojuan:
This fixes #437 by adding shortflags to the CLI and documentation for them.

Let me know if there are any flags missing or if you prefer a different formatting style.


cli.js
```diff
@@ -41,13 +41,13 @@ var cli = meow([
...
-	'  --init       Add AVA to your project',
-	'  --fail-fast  Stop after first test failure',
-	'  --serial     Run tests serially',
-	'  --require    Module to preload (Can be repeated)',
-	'  --tap        Generate TAP output',
-	'  --verbose    Enable verbose output',
-	'  --no-cache   Disable the transpiler cache',
+	'  --init           Add AVA to your project',
+	'  --fail-fast      Stop after first test failure',
+	'  -s, --serial     Run tests serially',
```
vdemedes: I'd reverse flags (`--serial, -s`), so that these lines are consistent with the rest of the output.

sotojuan: Yeah, I copied the example from the `meow` readme for this, but I can see how that's better. @sindresorhus let me know which style you prefer!

vdemedes: The standard is short flag first, but it does not look very good to me. Of course this thing is way too small to argue over, so if guys prefer the other way, fine by me ;)

sindresorhus: I don't like the standard either. Let's go with `--serial, -s` which looks nicer.

jamestalmage: I'm not sure I agree. The short flag first keeps everything aligned nicely. 

Long form first means the short flags are scattered instead of in a vertical line. It also means the short flags are creating noise in the middle of what is otherwise readable text

sindresorhus: @jamestalmage That's only true when all flags have shortflags, which isn't the case here or pretty much ever.

vdemedes:
Thank you @sotojuan! Closed via https://github.com/sindresorhus/ava/commit/5d36a8045d5f124b63b0b287091f8efc0ef732a4

Issue #421, opened 2016-01-13
Run individual tests in separate processes


jamestalmage:
Right now we fork per test file, but not per individual test method. We have discussed the latter a few times in other threads, and finally decided to open an issue to discuss.

Pros:
- Truly isolated tests. No chance they interfere with each-other unless they are interacting with outside resources (servers / databases / file system).
- Fresh global state for each test. This would be a big help for polyfill developers.

Cons:
- It will almost certainly introduce a performance penalty.
- Complete isolation could mask some errors. (i.e. What if your function always blows up the second time it is run? Complete isolation might mean your test suite never runs it twice).
- It will be really hard to replicate when it comes to implementing browser support.

sindresorhus:
> Fresh global state for each test. This would be a big help for polyfill developers.

This would be huge, not only for polyfill dev, but anything that modifies global state.

> It will almost certainly introduce a performance penalty.

Sometimes it's worth it. For example in https://github.com/sindresorhus/get-stdin I had to split my tests into two files as they somehow conflicted with each other.

> It will be really hard to replicate when it comes to implementing browser support.

That won't be a goal either. We'll just run it normally there. Running files in parallel will also be gone when run in the browser.

jamestalmage:
> Running files in parallel will also be gone when run in the browser.

Not sure why that would need to be true. Just open a tab per browser.

sindresorhus:
Oh, didn't realize that was possible. Make sure the mention that in https://github.com/sindresorhus/ava/issues/24 ;)

ariporad:
@sindresorhus: You changed your profile picture!

As for the actual topic at hand, I'm not really sure either way. On one
hand, it would be nice, but on the other, there would be a *_huge_*
performance penalty. And some unexpected behavior (read: it should be
behind a flag). It kind of starts to remind me of Google Apps Script, which
isn't a good thing.

Also, then before === beforeEach? That's something else we'd have to think
about

sindresorhus:
@ariporad Test hooks would still work as we would execute the file as normal, but only run a specific test (with hooks) instead of all.

jamestalmage:
Yep, the only issue would be:

```js
before(t => setup database())
```

ariporad:
@sindresorhus: I know, but then the before hook would have to be run for
each test (because it would have to be run once per thread), therefore
making it equal to beforeEach.

sindresorhus:
Ah, that's true.

jamestalmage:
[...]

PR #455, opened 2016-01-19
[WIP] - Refactor Runner

jamestalmage:
WIP refactoring `runner.js`. One goal is to support addressable test indexes which could be used implement #421. 

The `test-collection` class makes all the test data serializable, so it can be passed to the parent thread yet (I'm not actually doing that yet).

Still really early. Just wanted some feedback.


sindresorhus:
> One goal is to support addressable test indexes which could be used implement #421.

Is there any benefit to this other than for #421? #421 should really not have any impact on how we structure the code. It's just an early discussion that could easily be abandoned.

jamestalmage:
I think it's going to end up cleaner / easier to understand - with better separation of concerns. Also has implications for using domains for logging / error tracking. This is real early stuff. Not ready for serious review. Just wanted you aware of what I was doing

sindresorhus:
Alright

jamestalmage:
closing in favor of #466

PR #461, opened 2016-01-21
Performance Improvements for files with thousands of tests

jamestalmage:
Using the [emoji-aware test-per-character branch](https://github.com/beaugunderson/emoji-aware/commit/e2c3a824ec6e28c2e86e9191fd7123cea8ac1e00) as a benchmark, and running `time ava`:

master:
```
  4363 passed

ava  22.61s user 0.54s system 109% cpu 21.093 total
```

this branch (only the `Runner` commits):
```
  4363 passed

ava  7.30s user 0.51s system 139% cpu 5.597 total
```

this branch (all the commits):
```
  4363 passed

ava  6.32s user 0.46s system 145% cpu 4.675 total
```

--- 

updated benchmark (with empower-core performance tweaks):

```
ava --verbose  5.05s user 0.32s system 138% cpu 3.872 total
```

### DIFF
--- a/test/fork.js
+++ b/test/fork.js
@@ -72,6 +72,7 @@ test('fake timers do not break duration', function (t) {
 		});
 });
 
+/*
 test('destructuring of `t` is allowed', function (t) {
 	fork(fixture('destructuring-public-api.js'))
 		.run()
@@ -81,6 +82,7 @@ test('destructuring of `t` is allowed', function (t) {
 			t.end();
 		});
 });
+*/
 
 test('babelrc is ignored', function (t) {
 	fork(fixture('babelrc/test.js'))
--- a/lib/enhance-assert.js
+++ b/lib/enhance-assert.js
@@ -30,7 +30,8 @@ function enhanceAssert(opts) {
 			onError: opts.onError,
 			onSuccess: opts.onSuccess,
 			patterns: module.exports.PATTERNS,
-			wrapOnlyPatterns: module.exports.NON_ENHANCED_PATTERNS
+			wrapOnlyPatterns: module.exports.NON_ENHANCED_PATTERNS,
+			bindReceiver: false
 		}
 	);
 
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -50,6 +50,7 @@ function Runner(opts) {
 	this.results = [];
 	this.tests = [];
 	this.testsByType = {};
+	this.hasExclusive = false;
 }
 
 util.inherits(Runner, EventEmitter);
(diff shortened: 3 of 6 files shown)


jamestalmage:
This relies on some changes some changes to `empower-core`.

jamestalmage:
Further splitting that test suite into multiple files:

```
  4363 passed

ava  10.69s user 0.83s system 453% cpu 2.539 total
```

test/fork.js
```diff
@@ -72,6 +72,7 @@ test('fake timers do not break duration', function (t) {
 		});
 });
 
+/*
```
sindresorhus: ?

sindresorhus:
:+1: Looks good! (ᵔᴥᵔ)

Too bad https://github.com/jamestalmage/ava/commit/4e23943eb9a54fb74ada303f3bfed81d7aade933 was slow. It was more readable before.

test/fork.js
jamestalmage: With this change, destructuring would no longer work. This is not allowed anymore:

```js
test({is, same} => {
  is(1 + 1, 2);
  same([1,2].slice(1), [2]);
});
```

You have to preserve the this reference:

```js
test(t => {
  t.is(1 + 1, 2);
  t.same([1,2].slice(1), [2]);
});
```

jamestalmage:
> Too bad jamestalmage@4e23943 was slow. It was more readable.

Yeah, unfortunately it is a big offender in this scenario.

jamestalmage:
Also a technically breaking change (though not one I am really concerned about):

```js
t.is(1 + 1, 2);
```

No longer returns a Promise. 

Only `t.throws` and `t.doesNotThrow` will return a promise.

sindresorhus:
> though not one I am really concerned about

Me neither.

vdemedes:
Results look promising, let's get it in!

sindresorhus:
`(つ◕౪◕)つ━☆ﾟ.*･｡ﾟ`

kentcdodds:
giphy

jamestalmage:
@kentcdodds, et. all

Just for comparison - I also created a branch converting [`emoji-aware` to mocha](https://github.com/jamestalmage/emoji-aware/tree/mocha-test-per-char)

AVA now ekes out slightly better performance:

```
ava --verbose  9.76s user   0.71s system    465% cpu    2.252 total
mocha          2.07s user   0.31s system     96% cpu    2.460 total   
```

However, it's obvious the only reason AVA is winning is that it is leveraging multiple processes. Hopefully this means there are still lots of opportunities to make AVA even faster.

kentcdodds:
Totally. If we could make ava as fast as mocha on a per-file basis then ava would scream with speed. This is making me feel really positive about the potential for ava :D

PR #464, opened 2016-01-21
Fix undefined skip count

### LINKED ISSUE #423: TAP: Skip count reports `undefined`
```
...etc...
# skip undefined
...etc...
```

What more do you need? I briefly looked at the code surrounding this, and it appears that `TapReporter#api.skipCount` is never tracked. I could be wrong. If I get time to take a deeper look into this and submit a fix, I will do so, however don't let that stop anyone from taking a crack at it sooner! :rocket: 

[Probably related](https://github.com/sindresorhus/ava/commit/a5e63c0d43bfc13b820b3d52727830c3c2354793#diff-22844eb86dced527165902d51876b484R56), also [#140](https://github.com/sindresorhus/ava/pull/410/files)

sotojuan:
This fixes #423 by only outputting the skipped test count in the TAP reporter if there are skipped tests. If there are none, it does not output anything. It also adds a test for this.

[tape](https://github.com/substack/tape/blob/master/lib/results.js#L132) does this and so does [tap](https://github.com/tapjs/node-tap/blob/master/lib/test.js#L882) (thanks to @therealklanni for the line reference for tap!).

### DIFF
--- a/lib/reporters/tap.js
+++ b/lib/reporters/tap.js
@@ -66,11 +66,14 @@ TapReporter.prototype.finish = function () {
 		'',
 		'1..' + (this.api.passCount + this.api.failCount),
 		'# tests ' + (this.api.passCount + this.api.failCount),
-		'# pass ' + this.api.passCount,
-		'# skip ' + this.api.skipCount,
-		'# fail ' + (this.api.failCount + this.api.rejectionCount + this.api.exceptionCount),
-		''
+		'# pass ' + this.api.passCount
 	];
+
+	if (this.api.skipCount > 0) {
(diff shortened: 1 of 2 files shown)


sotojuan:
The travis failures don't seem to have anything to do with my code. Not sure how to fix it.

sindresorhus:
@sotojuan I manually reran the Node.js 0.10 test. Passes now. Chalk it up to flakiness...

sindresorhus:
LGTM

therealklanni:
Looks good, thanks for knocking that out @sotojuan :facepunch: :metal:

PR #465, opened 2016-01-23
Basic watcher setup

### LINKED ISSUE #70: Watch support
$ ava --watch

Is this a plan?

tomazzaman:
My first go at continuous watching using [chokidar](https://github.com/paulmillr/chokidar). Feedback welcome.

-

Fixes #70.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -24,6 +24,7 @@ var meow = require('meow');
 var Promise = require('bluebird');
 var pkgConf = require('pkg-conf');
 var isCi = require('is-ci');
+var chokidar = require('chokidar');
 var colors = require('./lib/colors');
 var verboseReporter = require('./lib/reporters/verbose');
 var miniReporter = require('./lib/reporters/mini');
@@ -48,6 +49,7 @@ var cli = meow([
 	'  --tap, -t        Generate TAP output',
 	'  --verbose, -v    Enable verbose output',
 	'  --no-cache       Disable the transpiler cache',
+	'  --watch, -w      Run tests when files change',
 	'',
 	'Examples',
 	'  ava',
@@ -68,14 +70,16 @@ var cli = meow([
 		'fail-fast',
 		'verbose',
 		'serial',
-		'tap'
+		'tap',
+		'watch'
 	],
 	default: conf,
 	alias: {
 		t: 'tap',
 		v: 'verbose',
 		r: 'require',
-		s: 'serial'
+		s: 'serial',
+		w: 'watch'
 	}
 });
 
@@ -95,6 +99,7 @@ var api = new Api(cli.input.length ? cli.input : arrify(conf.files), {
 
 var logger = new Logger();
 logger.api = api;
+logger.watch = cli.flags.watch;
 
 if (cli.flags.tap) {
 	logger.use(tapReporter());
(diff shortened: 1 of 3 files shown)


cli.js
```diff
@@ -112,17 +117,27 @@ api.on('error', logger.unhandledError);
...
+				console.error(colors.stack(err.stack));
+			}
+
+			logger.exit(1);
 	});
+}
+
+run();
+
+var watcher = chokidar.watch(cli.input, {persistent: cli.flags.watch});
```
tomazzaman: I was thinking whether it would be more effective watch the whole `cwd` instead of just test files, but wasn't sure what kind of performance impact would that have.

jamestalmage: I would think all of cwd excluding node_modules.

tomazzaman: Done, also ignoring `.git`

cli.js
```diff
@@ -112,17 +117,32 @@ api.on('error', logger.unhandledError);
...
+
+			logger.exit(1);
 	});
+}
+
+run();
+
+var cwd = process.cwd();
+var watcher = chokidar.watch(cwd, {
+	persistent: cli.flags.watch,
```
novemberborn: I think no `chokidar` code should be run unless `cli.flags.watch` is `true`. Currently if it's `false` it means Chokidar probably won't emit changes, but I don't find the documentation entirely clear on this.

cli.js
```diff
@@ -112,17 +117,32 @@ api.on('error', logger.unhandledError);
...
+	ignored: [cwd + '/node_modules', cwd + '/.git'],
```
novemberborn: You may be able to specify the `cwd` option and then keep these patterns relative. Might not hurt excluding other VCS paths either.

cli.js
```diff
@@ -112,17 +117,32 @@ api.on('error', logger.unhandledError);
...
+	followSymlinks: false
+});
+
+watcher.on('change', function (filepath) {
+	run();
```
novemberborn: Since the tests run asynchronously, you may end up starting a new batch before the previous execution has finished. It doesn't look like `api.run()` has any logic to defend against that (which it wouldn't have needed before).

Also `filepath` is unused.

lib/logger.js
```diff
@@ -75,6 +75,10 @@ Logger.prototype.exit = function (code) {
 	process.stdout.write('');
 	process.stderr.write('');
 
+	if (this.watch) {
```
novemberborn: IMO you should place this guard in the `cli` module, the logger shouldn't have to know about watchers.

novemberborn:
Nice!

Some further feature thoughts:

* It'd be cool to only re-run a test if its test file changed.
* It'd be really cool to track the dependencies of a test and rerun it when those change (e.g. when you change a file that is being tested, only run the test that loads that file).
* Could then observe `package.json` to pick up potential `node_modules` changes.

cli.js
sindresorhus: Should also use `path.join()`

lib/logger.js
sindresorhus: :

sindresorhus: :+1:

cli.js
```diff
@@ -48,6 +49,7 @@ var cli = meow([
 	'  --tap, -t        Generate TAP output',
 	'  --verbose, -v    Enable verbose output',
 	'  --no-cache       Disable the transpiler cache',
+	'  --watch, -w      Run tests when files change',
```
sindresorhus: This should be commented out. We decided to not yet expose this. https://github.com/sindresorhus/ava/issues/70#issuecomment-173746486

cli.js
sindresorhus: :+1:

novemberborn:
Thinking through this some more, if a test file i deleted that test should no longer run, conversely when a new file is added it should be run.

sindresorhus:
Closing in favor of #502.

Sorry @tomazzaman, but that one is a lot more comprehensive.

Issue #460, opened 2016-01-20
Benchmark


sindresorhus:
Re https://twitter.com/sindresorhus/status/689851060303872000

We should set up a benchmark with various types of test suites, like one with sync tests in one file, one with multiple files and async tests. This to make sure we don't regress in performance and ability to compare ourselves to other test runners.

kentcdodds:
Thanks for taking this so seriously! :smile: I'm currently working on a repo to compare mocha and ava with a lot of tests with a lot of files. I'm no benchmark expert, but maybe this will be helpful. Will comment when I've pushed it up. Thanks!

jfmengels:
I have created this repo for the same purpose: https://github.com/jfmengels/mocha-vs-ava-performance
It's pretty small, but it makes it obvious how big the difference is between Mocha's instant feedback in watch mode, and Ava's "watch" with Nodemon, or even single runs.

jamestalmage:
> watch mode

AVA is at an extreme disadvantage in watch mode until we implement our own watch mechanism. That's certainly on the roadmap, but we've been punting on it in favor of other improvements. Not sure if it's time to tackle that yet or not.

jamestalmage:
I have confirmed [this commit](https://github.com/beaugunderson/emoji-aware/commit/e2c3a824ec6e28c2e86e9191fd7123cea8ac1e00) caused a major performance hit. I think I've identified the problem and I am working up a solution.

See twada/empower-core#7

twada:
Just released [empower-core 0.5.0](https://github.com/twada/empower-core/blob/master/CHANGELOG.md#050-2016-01-21) :tada: (Thanks @jamestalmage !!)

sindresorhus:
@jamestalmage did a superb job adding a pretty benchmark in https://github.com/sindresorhus/ava/pull/496. It's easy to add additional things to benchmark, so suggestions welcome.

sindresorhus:
Instructions here: https://github.com/sindresorhus/ava/commit/c22eabd717451dbeaeeff1520f399f5a5fd85764

Issue #222, opened 2015-11-16
Support nested groups?


novemberborn:
Currently the test files support a flat list of tests, each sharing all `before` , `beforeEach`, `afterEach` & `after` hooks. I find it useful to declare nested groups with their own hooks, while inheriting context from the parent.

For example I'm stubbing promise behavior and need to write test for if the promise is fulfilled or if it's rejected. Ideally I could do something like:

```js
test.beforeEach(t => {
  t.context.behavior = sinon.stub({ getPromise () })
  // Return a pending promise
  t.context.behavior.getPromise.returns(new Promise(() => {}))
})

// Add optional title
test.group('promise is fulfilled', () => {
  test.beforeEach(t => {
    t.context.behavior.getPromise.returns(Promise.resolve(true))
  })

  test(async () => {
    const value = await t.context.behavior.getPromise()
    t.true(value)
  })
})

test.group('promise is rejected', () => {
  // You get the idea…
})
```

Each group would maintain its own list of hooks. Hooks could be inherited from the parent group at runtime or when initializing. There would be a root group for the test file itself.

This would allow building alternative test interfaces alongside AVA.

I'd be happy to try and do a PR for this if there's interest.

---

P.S. AVA looks really intriguing! Test runners could do with some shaking up :+1:

vdemedes:
I think it has some potential as a feature request. Something like `describe` in mocha. At the moment, we'd like to merge currently open PRs and do a complete cleanup/refactoring, and then continue with feature development.

If you'd be interested in making a PR, that'd be great. I don't recommend you to do that now, because after the rewrite, your PR will have a lot of merge conflicts.

@sindresorhus what do you think?

sindresorhus:
From a quick skim this looks like the same as https://github.com/sindresorhus/ava/issues/62 just with a different name.

I'm not a big fan of this kind of complication to be honest. If you think you need nested tests or groups, you're probably better off splitting it up into multiple files (and get the benefit of parallelism) or simplifying your tests.

I'll keep this open for discussion, but even if we are going to add this, it's not going to happen right now. I want to focus our efforts on making a really good minimal core first.

novemberborn:
> From a quick skim this looks like the same as #62 just with a different name.

Ah yes, apologies for not finding that myself.

> even if we are going to add this, it's not going to happen right now. I want to focus our efforts on making a really good minimal core first.

Fair enough.

---

[...]

PR #466, opened 2016-01-23
Improve runner performance

jamestalmage:
The main premise I had while creating this: Deferring when you don't need to is expensive.

 This rewrites the runner and test to use the following semantics:

If a test never uses an async features (i.e. `cb` mode, `t.throws(promise)`, return a `promise` / `observable`) the next test is executed synchronously.

This requires a unified API between tests and the runner:

```js
// returns a "result" or promise for a result
var result = runnable.run();

// a "result" has the following properties:
result = {
  passed: true || false,
  reason: Error,
  result: "some result"
}
```

Because `run` can return either a result or a promise for a result, you frequently need to check which it is using `is-promise`. This leads to some verbose code, but it runs faster, and is fairly well encapsulated in the `sequence` and `concurrent` classes.

AVA master:
```
ava --verbose  9.38s user 0.63s system 461% cpu 2.173 total
```

This PR:
```
ava --verbose  8.46s user 0.65s system 522% cpu 1.744 total
```


jamestalmage:
All working now, using benchmarks from [jamestalmage/emoji-aware#no-assertion-plans](https://github.com/jamestalmage/emoji-aware/tree/no-assertion-plans):

Master:
```
ava --verbose  6.86s user 0.56s system 439% cpu 1.689 total
ava --verbose  6.78s user 0.55s system 438% cpu 1.669 total
ava --verbose  6.73s user 0.55s system 432% cpu 1.681 total
```

This PR:
```
ava --verbose  5.25s user 0.53s system 432% cpu 1.337 total
ava --verbose  4.97s user 0.50s system 419% cpu 1.303 total
ava --verbose  5.05s user 0.50s system 426% cpu 1.301 total
```

jamestalmage:
Using the following benchmark (no assertions - just tracking runner performance):

```js
import test from './';

	test('test' + i, t => {});
}
```

Master:
```
node ./cli.js benchmark.js --verbose  3.15s user 0.31s system 104% cpu 3.318 total
node ./cli.js benchmark.js --verbose  2.76s user 0.28s system 103% cpu 2.937 total
node ./cli.js benchmark.js --verbose  2.73s user 0.28s system 103% cpu 2.911 total
```

This PR:
```
node ./cli.js benchmark.js --verbose  1.44s user 0.25s system 107% cpu 1.573 total
node ./cli.js benchmark.js --verbose  1.40s user 0.25s system 106% cpu 1.544 total
node ./cli.js benchmark.js --verbose  1.38s user 0.24s system 108% cpu 1.499 total
```

jamestalmage:
This also seems to provide a slight performance boost for async tests:

```js
import test from './';

	test('test' + i, t => {return Promise.resolve()});
}
```

Master:

```
node ./cli.js benchmark.js --verbose  2.84s user 0.28s system 103% cpu 3.010 total
node ./cli.js benchmark.js --verbose  2.86s user 0.28s system 103% cpu 3.033 total
node ./cli.js benchmark.js --verbose  2.82s user 0.28s system 104% cpu 2.980 total
```

This PR:
```
node ./cli.js benchmark.js --verbose  2.68s user 0.28s system 103% cpu 2.843 total
node ./cli.js benchmark.js --verbose  2.69s user 0.28s system 104% cpu 2.854 total
node ./cli.js benchmark.js --verbose  2.72s user 0.27s system 104% cpu 2.880 total
```

jamestalmage:
Mixed sync/async:
```
import test from './';

test.beforeEach(t => {});
test.afterEach(t => {});

	test('test' + i, t => {return Promise.resolve()});
}
```

Master:
```
node ./cli.js benchmark.js --verbose  4.78s user 0.43s system 102% cpu 5.075 total
node ./cli.js benchmark.js --verbose  4.64s user 0.42s system 102% cpu 4.953 total
node ./cli.js benchmark.js --verbose  4.75s user 0.43s system 102% cpu 5.041 total
```

This PR:
```
node ./cli.js benchmark.js --verbose  3.56s user 0.38s system 103% cpu 3.797 total
node ./cli.js benchmark.js --verbose  3.56s user 0.38s system 103% cpu 3.791 total
node ./cli.js benchmark.js --verbose  3.62s user 0.39s system 103% cpu 3.888 total
```

lib/concurrent.js
```diff
@@ -0,0 +1,89 @@
+'use strict';
+var Promise = require('bluebird');
+var isPromise = require('is-promise');
+var BAIL_ERROR = new Error();
+
+module.exports = Concurrent;
+
+function Concurrent(tests, bail) {
+	if (!this instanceof Concurrent) {
+		throw new Error('must use `new Concurrent`');
```
sindresorhus: Should be `TypeError` to match ES2015 classes and message: `Class constructor Concurrent cannot be invoked without 'new'`

lib/concurrent.js
```diff
@@ -0,0 +1,89 @@
...
+	var ret = Promise.all(results).then(function (results) {
+		return {
+			passed: passed,
+			reason: reason,
+			result: results
+		};
+	});
+
+	if (bail) {
+		ret = ret.catch(function (e) {
```
sindresorhus: `e` => `err`

lib/sequence.js
```diff
@@ -0,0 +1,97 @@
+'use strict';
+var Promise = require('bluebird');
+var isPromise = require('is-promise');
+
+module.exports = Sequence;
+
+function Sequence(tests, bail) {
+	if (!this instanceof Sequence) {
+		throw new Error('Sequence must be called with new');
```
sindresorhus: Same as https://github.com/sindresorhus/ava/pull/466/files#r50755945

lib/test-collection.js
```diff
@@ -0,0 +1,149 @@
...
+		if (metadata.exclusive && !this.hasExclusive) {
+			this.hasExclusive = true;
+			this.serial = [];
+			this.concurrent = [];
+		}
+		(metadata.serial ? this.serial : this.concurrent).push(test);
+		return;
+	}
+	if (metadata.exclusive) {
+		throw new Error('you can\'t use only with a ' + type + ' test');
```
sindresorhus: I would quote `only` to make it clearer that it's not a normal word.

test/hooks.js
```diff
@@ -257,12 +257,12 @@ test('shared context', function (t) {
 	var runner = new Runner();
 
 	runner.before(function (a) {
-		a.is(a.context, undefined);
-		a.context = {arr: []};
+		a.is(a.context, null);
+		// a.context = {arr: []};
```
sindresorhus: ?

test/runner.js
```diff
@@ -130,7 +130,7 @@ test('methods are chainable: serial.only', function (t) {
 	t.true(runner.tests[0].metadata.exclusive);
 	t.false(runner.tests[0].metadata.skipped);
 	t.end();
-});
+});   */
```
sindresorhus: :ghost:

lib/test.js
```diff
@@ -11,12 +11,13 @@ var assert = require('./assert');
 var enhanceAssert = require('./enhance-assert');
 var globals = require('./globals');
 
-function Test(title, fn) {
+function Test(title, fn, contextRef, report) {
 	if (!(this instanceof Test)) {
-		return new Test(title, fn);
+		return new Test(title, fn, contextRef);
```
matthewbauer: this should be return new Test(title, fn, contextRef, report);

test/hooks.js
jamestalmage: Setting `t.context` in `before` or `after` will now throw an error. Previously we silently let them set it, but it was not passed to future tests.

test/runner.js
jamestalmage: oops. Yeah that should just be deleted. It's well covered in `test-collection`

lib/test.js
jamestalmage: True. (mistakes like this is why I prefer to throw instead of doing magic).

test/hooks.js
jamestalmage: I probably should add a test covering that.

matthewbauer:
Is there a way to get nested groups (#222) working in this? I'm thinking of an implementation like this:

```js
test.group(function(test) {
  test("test a", function() {
  });
  test.beforeEach("before test a but not test b", function() {
  });
});
test("test b", function() {});
```

Basically test.group just generates another runner that is added into the Runner.prototype.tests array. All of the hooks work as expected, just with the new test in context.

I was working on a PR to do this but then I saw the changes here and wanted to check what your thoughts are before moving forward.

sindresorhus:
:clap: From a quick skim this looks very promising, although a lot to take in. I like the improved separation of concerns and performance improvements, but not totally happy about the added verboseness. Would be nice if we could follow up later with ways to simplify it. I'll try to do a more thorough review tomorrow.

-

@vdemedes @ariporad @sotojuan @novemberborn I could use some help reviewing this :)

sindresorhus:
@matthewbauer Can you move your comment to https://github.com/sindresorhus/ava/issues/222 so not to derail this PR discussion? :)

matthewbauer:
:ok_hand:

jamestalmage:
> Would be nice if we could follow up later with ways to simplify it.

Agreed. Getting a good benchmarking suite (#460) would be really helpful as well.

vdemedes:
Massive applause for huge amount of work, @jamestalmage.

I have some style/cleanup-related suggestions, but I can follow up with a PR myself ;)

sindresorhus:
> I have some style/cleanup-related suggestions, but I can follow up with a PR myself ;)

Yes, let's do style stuff later. (I did a few myself, but realized it's better not to dilute the review with nitpicks).

jamestalmage:
Once merged, we should all dogfood this on existing projects before release, (maybe deploy with a `next` tag for a while as well). There are enough changes here that our regression exposure is pretty large. I've added pretty exhaustive tests, but we should still thoroughly vet it.

sindresorhus:
> Once merged, we should all dogfood this on existing projects before release

:+1: I've already tried it on `got` and `pageres` and seems to work good.

> maybe deploy with a next tag for a while as well

:+1:

sotojuan:
Currently in class but from what I skimmed it looks good :-) Awesome work!

sindresorhus:
Landed! Awesome work @jamestalmage :)

unicorn17

PR #475, opened 2016-01-26
Add a profile script

jamestalmage:
I've been working on getting [`iron-node`](http://s-a.github.io/iron-node/) working with AVA so we can do some proper profiling. I think I finally cracked it.

The first thing was to run in a single process (this means only running a single test file), and fake the response of a parent process. It logs test results to the Chrome DevTools console (and it's not very pretty), but we shouldn't really care about the test results much while profiling.

The second was working around some weirdness with `setTimeout` in `iron-node`. I'm not sure if it's an `iron-node` problem, or an `electron` one, but storing `setTimeout` the way we do in `globals.js` breaks things. The solution here was to just undo our safe copies of `setTimeout` and `clearTimeout`. This means we can't profile against anything that manipulates timers (not sure why that would ever be necessary).

### DIFF
--- a/profile.js
+++ b/profile.js
@@ -1,0 +1,102 @@
+'use strict';
+
+// iron-node does not work with forked processes
+// This cli command will run a single file in the current process.
+// Intended to be used with iron-node for profiling purposes.
+
+var path = require('path');
+var meow = require('meow');
+var Promise = require('bluebird');
+var pkgConf = require('pkg-conf');
+var arrify = require('arrify');
+var findCacheDir = require('find-cache-dir');
+var uniqueTempDir = require('unique-temp-dir');
+var EventEmitter = require('events').EventEmitter;
+var CachingPrecompiler = require('./lib/caching-precompiler');
+var globals = require('./lib/globals');
+
+// Chrome gets upset when the `this` value is non-null for these functions.
+globals.setTimeout = setTimeout.bind(null);
+globals.clearTimeout = clearTimeout.bind(null);
+
+Promise.longStackTraces();
+var conf = pkgConf.sync('ava');
+
+// Define a minimal set of options from the main CLI.
+var cli = meow([
+	'usage: iron-node node_modules/ava/profile.js [options] TEST_FILE',
+	'',
+	'Options',
+	'  --fail-fast      Stop after first test failure',
+	'  --serial, -s     Run tests serially',
+	'  --require, -r    Module to preload (Can be repeated)',
+	''
+], {
+	string: [
+		'_',
+		'require'
+	],
+	boolean: [
+		'fail-fast',
+		'verbose',
+		'serial',
+		'tap'
+	],
+	default: conf,
+	alias: {
+		r: 'require',
+		s: 'serial'
+	}
+});
+
+if (cli.input.length !== 1) {
+	throw new Error('no file');
+}
+
+var file = path.resolve(cli.input[0]);
+var cacheDir = findCacheDir({name: 'ava', files: [file]}) || uniqueTempDir();
+var opts = {
+	file: file,
+	failFast: cli.flags.failFast,
+	serial: cli.flags.serial,
+	require: arrify(cli.flags.require),
+	tty: false,
+	cacheDir: cacheDir,
+	precompiled: new CachingPrecompiler(cacheDir).generateHashForFile(file)
+};
+
+var events = new EventEmitter();
+
+// Mock the behavior of a parent process.
+process.send = function (data) {
+	if (data && data.ava) {
+		var name = data.name.replace(/^ava-/, '');
+		if (events.listenerCount(name)) {
+			events.emit(name, data.data);
+		} else {
+			console.log('UNHANDLED AVA EVENT: ', name, data.data);
+		}
+		return;
+	}
+	console.log('NON AVA EVENT: ', data);
+};
+
+events.on('test', function (data) {
+	console.log('TEST:', data.title, data.error);
+});
+
+events.on('results', function (data) {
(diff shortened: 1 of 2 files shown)


profile.js
```diff
@@ -0,0 +1,111 @@
...
+var uniqueTempDir = require('unique-temp-dir');
+var EventEmitter = require('events').EventEmitter;
+var CachingPrecompiler = require('./lib/caching-precompiler');
+var globals = require('./lib/globals');
+
+// copy original console and console.log so tests can't manipulate them.
+var originalConsole = console;
+var consoleLog = originalConsole.log;
+
+// Our usage of globals to store setTimeout/clearTimeout upsets iron-node for some reason.
```
sindresorhus: Can you open an issue on `iron-node` about this?

profile.js
```diff
@@ -0,0 +1,111 @@
...
+var arrify = require('arrify');
+var findCacheDir = require('find-cache-dir');
+var uniqueTempDir = require('unique-temp-dir');
+var EventEmitter = require('events').EventEmitter;
+var CachingPrecompiler = require('./lib/caching-precompiler');
+var globals = require('./lib/globals');
+
+// copy original console and console.log so tests can't manipulate them.
+var originalConsole = console;
+var consoleLog = originalConsole.log;
```
sindresorhus: Add these to `./lib/globals`?

sindresorhus:
Can you quickly mention how to use this in `maintaining.md`?

sindresorhus:
Can you add a `.iron-node.js` config file? With these to `true`: https://github.com/s-a/iron-node/blob/85198516e90bb9e03d45efe0bd361ad4a4fe467e/.iron-node.js#L11-L12

sindresorhus:
Tried it now. What's the workflow for profiling? The tests run on startup and finish before I manage to do anything. Is the recommended practise to put a `debugger;` statement at the top of the test file?

jamestalmage:
Sorry @sindresorhus - for some reason I never got notifications on any of these.

> What's the workflow for profiling?

Click record then reload the page (that is what I did). I did not put any debugger statements in. That's an interesting thought though.

profile.js
jamestalmage: I don't think any of our child-process code will ever call `console.log`.

profile.js
jamestalmage: done

profile.js
sindresorhus: Yeah, I know. Just thought it would be cleaner to keep all overridden globals in our `globals` file.

jamestalmage: I am just dropping the safe copies. If someone ever asks us to profile a suite that does something to console.log we can address it then.

sindresorhus:
Awesome. Excited to have an easy way to perf test AVA.

s-a:
I find this very intersting! Are you willing to share your workflow to profile your software in something like markdown documentation? I like to link from my repo to share your work with the community. Just want to say... keep in mind that results will differ from native node but it will in fact deliver insightful results.

sindresorhus:
@s-a I added a short section in https://github.com/sindresorhus/ava/blob/master/maintaining.md#profiling. Was hoping @jamestalmage would fill it out with his workflow.

s-a:
:+1:  I can not wait to read it . :) @jamestalmage Please ping me here when you finished.

mattdesl:
I'm also interested in the progress of this thread and how we can combine efforts with [devtool](https://github.com/Jam3/devtool) to provide better debugging/profiling all around.

e.g. The `setTimeout` issue was resolved in a recent commit to `devtool`, since now we are using true Node.js timers instead of the built-in browser timers.
https://github.com/s-a/iron-node/issues/85

Relevant commit:
https://github.com/Jam3/devtool/commit/11e491c2bb4540f2a2b9bf6290b3e2b82bcbba89

sindresorhus:
Relevant: https://github.com/Jam3/devtool#iron-node

@mattdesl Is there anything `devtools` is worse at than `iron-node` for our needs (profiling CPU usage)?

jamestalmage:
I just played with `devtool`, it seems great, and it already works with `profile.js`.

```
devtool node_modules/ava/profile.js test/someTest.js
```

@mattdesl @s-a, with either project, is there a way to tell Chrome DevTools to start recording a CPU profile programmatically? It would be awesome not to have to start the recording and then reload the page.

sindresorhus:
@jamestalmage `console.profile('name)` and `console.profileEnd('name')`.

mattdesl:
I think the differences will be subtle between the two tools, though my end goal is to have `devtool` support most use cases (which involves lots of tweaks for `require.main`, `process`, and all that). 

You can use `console.profile("foo")` and `console.profileEnd("foo")`. So you could programmatically add profiles for files or whatever. 

I am currently writing a blog post on all things `devtool`, you can see it here: 
http://mattdesl.svbtle.com/debugging-nodejs-in-chrome-devtools

jamestalmage:
> I am currently writing a blog post on all things devtool

Awesome - great writeup - I just added it to #489 

> console.profile('foo')

@mattdesl @s-a Feedback on #491 would be appreciated.

s-a:
@jamestalmage sorry #491 is out of my scope. I feel this could have something to do with 
> e.g. The setTimeout issue was resolved in a recent commit to devtool, since now we are using true Node.js timers instead of the built-in browser timers.

As matt metioned in his README.md devTool is in an experimental state.
> Note: This tool is still in early stages. So far it has only been tested on a couple of OSX machines. :)

However FYI just added https://github.com/s-a/iron-node/blob/master/docs/PROFILE.md. There is a lot more work todo. :) Nice weekend @all :smile:

Issue #476, opened 2016-01-27
Only run tests with title matching expression


matthewbauer:
This is something that I would do with mocha. Here's the mocha help:

```
    -g, --grep <pattern>                    only run tests matching <pattern>
```

I'm hoping to get something like that working.

sindresorhus:
I agree we should have something like this, but I'd like to discuss how we can innovate on this instead of just copying the Mocha option. I also have a feeling if we add this, people will follow up and demand `--fgrep` (only run tests containing <string>) and then `--invert` (inverts --grep and --fgrep matches). I don't want more than one flag for this.

Maybe we could use globs instead?

```
$ ava --match='fo*' --match='!bar'
```

Regex might work well too. Just putting it out there.

matthewbauer:
I mean to be honest for my use case all that's needed is just to search for a string literal in the titles. I guess I just wanted to make sure it's versatile for different use cases, although globs would be more intuitive (and match the globs in the file args).

wilmoore:
I personally think globs make more sense. Basic string literal searches only get us so far; however, constructing a valid `RegExp` in the shell feels awkward.

sindresorhus:
Can you think of any use-cases where a glob might not work?

sindresorhus:
And to be clear, `!bar` is a negated match, so we wouldn't need another flag for that.

wilmoore:
> Can you think of any use-cases where a glob might not work?

Off the top of my head, no; however, I can think of many times where mocha's grep was too limited to the point where I don't even use it so pretty much anything (especially globs) would be a vast improvement.

wilmoore:
The only other thing I'd suggest is to change `--match` to `--glob` to be more explicit about it:

```shell
$ ava --glob='fo*' --glob='!bar'
```

sindresorhus:
> I can think of many times where mocha's grep was too limited to the point where I don't even use it so pretty much anything (especially globs) would be a vast improvement.

Would be awesome if you could provide some examples. That would help make a decision.

matthewbauer:
For now, my PR is just going to do --match just because RegEx/glob matching is fairly expensive when done on large test suites. Since titles are pretty easy to change I think that should cover most if not all use cases.

sindresorhus:
@matthewbauer There's no "for now". Anything we add we'll have to support for a long time, and as I originally said, not interested in more than one flag for matching tests. We should figure out how to do it correctly. Not just a quick fix.

[...]

PR #477, opened 2016-01-27
Add `match` option

matthewbauer:
This aims to resolve issue #476. --match can be passed to ava which
means to only run tests with a title matching the provided match
string. The skipped metadata flag will be added to each test not
matching the match string.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -48,6 +48,7 @@ var cli = meow([
 	'  --tap, -t        Generate TAP output',
 	'  --verbose, -v    Enable verbose output',
 	'  --no-cache       Disable the transpiler cache',
+	'  --match, -m      Only run tests with titles matching',
 	'',
 	'Examples',
 	'  ava',
@@ -62,7 +63,8 @@ var cli = meow([
 ], {
 	string: [
 		'_',
-		'require'
+		'require',
+		'match'
 	],
 	boolean: [
 		'fail-fast',
@@ -75,7 +77,8 @@ var cli = meow([
 		t: 'tap',
 		v: 'verbose',
 		r: 'require',
-		s: 'serial'
+		s: 'serial',
+		m: 'match'
 	}
 });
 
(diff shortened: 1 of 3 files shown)


cli.js
```diff
@@ -48,6 +48,7 @@ var cli = meow([
 	'  --tap, -t        Generate TAP output',
 	'  --verbose, -v    Enable verbose output',
 	'  --no-cache       Disable the transpiler cache',
+	'  --grep, -g       Only run tests matching pattern',
```
sindresorhus: > Only run tests matching a pattern

lib/runner.js
```diff
@@ -47,6 +47,15 @@ function Runner(opts) {
 	EventEmitter.call(this);
 
 	this.options = opts || {};
+
+	if (this.options.grep instanceof RegExp) {
+		this.grep = this.options.grep;
```
sindresorhus: I don't think the API should support both regex and string. It should be converted from a string to regex in the CLI.

cli.js
```diff
@@ -48,6 +48,7 @@ var cli = meow([
 	'  --tap, -t        Generate TAP output',
 	'  --verbose, -v    Enable verbose output',
 	'  --no-cache       Disable the transpiler cache',
+	'  --match, -m      Only run tests with titles matching',
```
sindresorhus: > Only run tests with matching title

jamestalmage:
I think we've decided on slightly different semantics for all this (see the final few comments in https://github.com/sindresorhus/ava/issues/476).

Specifically, there is discussion of sudo globs (only `*` supported) and negation with a `!` prefix.

@sindresorhus - do you feel the issue is settled?

@matthewbauer - Are you willing to update this PR?

sindresorhus:
@jamestalmage Let's wait until after the weekend. Would like @vdemedes feedback on #476.

cli.js
sindresorhus: The flag docs also needs to be reflected in the CLI help output in the readme.

sindresorhus:
@matthewbauer Can you also fix the merge conflict?

Sorry about the back and forth. Glad we finally have a resolution on this.

sindresorhus:
Should it be possible to use it multiple times? I'm thinking `--match='*oo' --match='!foo'`.

We also need to document the syntax somewhere, the `*` wildcard and `!` negation.

sindresorhus:
@matthewbauer Just a friendly ping :)

sindresorhus:
Closing for lack of activity. Would be happy to reopen if this get's any updates, but I want to give other users the opportunity to work on this.

PR #480, opened 2016-01-28
Update ava version in readme package.json

zachlysobey:


sindresorhus:
Nah, if I accept this we'll have to bump it for every version. We'll bump it to 1.0.0 when we get there. Thanks though.

jamestalmage:
We could write an `npm version` script to handle swapping this out every time you publish with `np`.

sindresorhus:
@jamestalmage I think that's overdoing it. I don't see the point. It will be moot anyways when we do 1.0.0. And it's just an example. Not meant to be copy-pasted.

jamestalmage:
:+1:

PR #488, opened 2016-01-29
Fix skip count not reporting properly

### LINKED ISSUE #470: TAP: Skip count not reporting properly
Seems that after the skip count fix, AVA does not report the total skip count, even when there are `> 0` skipped tests. Additionally, skipped tests are being printed as passing tests.

We get
```
ok N - this test was skipped
```

Instead of
```
ok N - # skip this test was skipped
```

And therefore other TAP libraries (e.g. faucet) break because they count it as a passing test, not skipped.

Here's an example from one of my own libraries using `ava --tap`:

```
TAP version 13
# json-adapter › read method
ok 1 - json-adapter › read method
# json-adapter › rejects
ok 2 - json-adapter › rejects
# json-adapter › create method
ok 3 - json-adapter › create method
# create › rejects
ok 4 - create › rejects
# create › create pending task
ok 5 - create › create pending task
# create › create completed task
ok 6 - create › create completed task
# create › create waiting task
ok 7 - create › create waiting task
# create › create recurring parent task
ok 8 - create › create recurring parent task
# create › create recurring child task
ok 9 - create › create recurring child task

1..7
# tests 7
# pass 7
# fail 0
```

Test 1 and 2 above were actually skipped—notice no "# skip". Also note no "# skipped N" summary line.

This confuses faucet (and presumably anything else that would interpret the broken TAP output):

```
✓ json-adapter › read method
✓ json-adapter › rejects
✓ json-adapter › create method
✓ create › rejects
✓ create › create pending task
✓ create › create completed task
✓ create › create waiting task
✓ create › create recurring parent task
✓ create › create recurring child task
# tests 7
✓ pass 7
not ok 10 too many asserts
⨯ fail  1
```

Related: #464 

cc @sotojuan if you want to take a look, since you've been contributing in this area and you at least contributed directly to the skipped count summary.

sotojuan:
This fixes #470.

What was happening was that the API did not have a `skipCount` property, so the TAP reporter had nothing to report with. This commit adds it to the API and then updates the TAP reporter so it can use it.

[Seems to work](http://i.imgur.com/YSzK1cY.png), but could use more pairs of eyes.

### DIFF
--- a/lib/reporters/tap.js
+++ b/lib/reporters/tap.js
@@ -62,10 +62,11 @@ TapReporter.prototype.unhandledError = function (err) {
 };
 
 TapReporter.prototype.finish = function () {
+	// console.log(this.api)
 	var output = [
 		'',
-		'1..' + (this.api.passCount + this.api.failCount),
-		'# tests ' + (this.api.passCount + this.api.failCount),
+		'1..' + (this.api.passCount + this.api.failCount + this.api.skipCount),
+		'# tests ' + (this.api.passCount + this.api.failCount + this.api.skipCount),
 		'# pass ' + this.api.passCount
 	];
 
--- a/api.js
+++ b/api.js
@@ -30,6 +30,7 @@ function Api(files, options) {
 	this.rejectionCount = 0;
 	this.exceptionCount = 0;
 	this.passCount = 0;
+	this.skipCount = 0;
 	this.failCount = 0;
 	this.fileCount = 0;
 	this.testCount = 0;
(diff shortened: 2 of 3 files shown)


lib/reporters/tap.js
```diff
@@ -62,10 +62,11 @@ TapReporter.prototype.unhandledError = function (err) {
 };
 
 TapReporter.prototype.finish = function () {
+	// console.log(this.api)
```
sindresorhus: :ghost:

sindresorhus:
LGTM

jamestalmage:
:+1: - Just needs that one comment line removed.

sindresorhus:
Thank you @sotojuan :)

PR #491, opened 2016-01-30
auto start the CPU profiler

jamestalmage:
This will automatically start the DevTools profiler when running `profile.js`

### DIFF
--- a/profile.js
+++ b/profile.js
@@ -90,6 +90,7 @@ events.on('test', function (data) {
 });
 
 events.on('results', function (data) {
+	console.profileEnd();
 	console.log('RESULTS:', data.stats);
 });
 
@@ -103,4 +104,8 @@ events.on('stats', function () {
 process.argv[2] = JSON.stringify(opts);
 process.argv.length = 3;
 
-require('./lib/test-worker');
+console.profile('AVA test-worker process');
+
+setImmediate(function () {
+	require('./lib/test-worker');
+});


profile.js
```diff
@@ -103,4 +104,8 @@ events.on('stats', function () {
 process.argv[2] = JSON.stringify(opts);
 process.argv.length = 3;
 
-require('./lib/test-worker');
+console.profile('AVA test-worker process');
+
+setTimeout(function () {
```
sindresorhus: Why? And `setImmediate` instead?

jamestalmage: The profiler seemed to startup async.

jamestalmage:
`setImmediate` works.

`iron-node` works just fine with a synchronous call. With `devtool` doesn't print the "profile started" line unless I use `setImmediate`.

// @mattdesl

mattdesl:
Not sure why there would be a difference. You could try --no-node-timers to see if that is the reason setImmediate was needed.

mattdesl:
I've cloned `ava` and I'm trying to run some tests with devtool.

```sh
devtool node_modules/ava/profile.js <test-file>
```

Are there specific files to use for `<test-file>` that I should be profiling? Or are you just npm linking ava and testing it on random projects?

jamestalmage:
You can link it, or use a fixture

```
devtool profile.js test/fixtures/es2015.js
```

mattdesl:
Pulled from master. Testing with devtool@1.7.6, node@5 and npm@2.14.12, adding the `console.profile/profileEnd` works fine without `setImmediate` for the above fixture. Which test is failing?

jamestalmage:
It doesn't log the  "profiling started" line - I think that may be the only issue. But it makes me paranoid that I might be missing some profiling data

sindresorhus:
@jamestalmage Can you open an issue on `devtool` about this? It's weird that `setImmediate` is required?

PR #494, opened 2016-01-30
Return empty string instead of null for a single anonymous test

### LINKED ISSUE #487: Output `null` when tests don't have a name
I just noticed this by accident when testing [mem-map](https://github.com/SamVerschueren/mem-map) and couldn't find a similar issue.

screen shot 2016-01-29 at 08 36 36

**AVA version**: v0.11.0

The reason seems to be that I don't have a name for my test.

sotojuan:
This fixes #487 by returning an empty string instead of `null` when there's just one anonymous test. This also keeps the behavior of printing `[anonymous]` when there are more than one anonymous tests.

### DIFF
--- a/lib/reporters/verbose.js
+++ b/lib/reporters/verbose.js
@@ -27,7 +27,7 @@ VerboseReporter.prototype.test = function (test) {
 	}
 
 	if (this.api.fileCount === 1 && this.api.testCount === 1 && test.title === '[anonymous]') {
-		return '';
+		return undefined;
 	}
 
 	// display duration only over a threshold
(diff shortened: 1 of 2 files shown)


sotojuan:
Flaky Travis tests strike again?

vdemedes:
Yeah, something is wrong with 0.10 build, I just had an issue with it too. Restarted your build.

sindresorhus:
When there's no test title, instead of outputting an empty line, it should output nothing.

Right now it outputs the following:

```
❯ ava --verbose



  1 test passed


```

Notice that there are 3 empty lines above and only 2 below.

Instead it should only output 2 empty lines above:

```
❯ ava --verbose


  1 test passed


```

sotojuan:
Returning an empty string does put an extra line there, and returning nothing (just `return`) gives an XO error. I opted to return `undefined`, which seems to get the job done (no extra line or extra output).

jillesme:
Instead of `return undefined;` you can just do `return;` and it will return `undefined`.

vdemedes:
@sotojuan `return null` did not work?

sotojuan:
`null`:



`undefined`:



`''`:

PR #495, opened 2016-01-30
Clean up sequence

vdemedes:
Feedback welcome!

### DIFF
--- a/lib/sequence.js
+++ b/lib/sequence.js
@@ -1,5 +1,4 @@
 'use strict';
-var Promise = require('bluebird');
 var isPromise = require('is-promise');
 
 module.exports = Sequence;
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -141,6 +141,9 @@
   "xo": {
     "ignore": [
       "cli.js"
-    ]
+    ],
+		"rules": {
+			"no-loop-func": 1
```
vdemedes: Had to insert this rule, because it does not like https://github.com/sindresorhus/ava/blob/cleanup-sequence/lib/sequence.js#L42:

```js
for (...) {
  return fn().then(function () {
    // this anonymous function is a problem for linter
  });
}
```

jamestalmage: This will avoid that error (and be faster):
```js
var i = fromIndex;
function runAsync(ret) {
  self._addResult(ret);
  return self._run(i + 1);
}
for(; i < length; i++) {
  // ...
  if (isPromise(result)) {
    return result.then(runAsync);
  }
}
```

vdemedes: I always try to avoid functions in functions. I will need to add another `_` method.

jamestalmage:
This is awesome. Much cleaner than my mess. I need to run a few more tests, but it seems really performant too.

lib/sequence.js
```diff
@@ -9,99 +8,67 @@ function Sequence(tests, bail) {
...
-					result: results,
-					reason: result.reason
-				};
-			} else if (passed) {
-				passed = false;
-				reason = result.reason;
-			}
-		}
+	if (!fromIndex) {
+		fromIndex = 0;
```
jamestalmage: Any reason we can't just eliminate this `if` statement, and pass in `0`?

```js
return this._run(0);
```

lib/sequence.js
```diff
@@ -9,99 +8,67 @@ function Sequence(tests, bail) {
 		throw new TypeError('Class constructor Sequence cannot be invoked without \'new\'');
 	}
 
-	this.tests = tests;
-	this.bail = bail;
+	this.results = [];
+	this.passed = true;
+	this.reason = null;
+	this.tests = tests || [];
```
jamestalmage: Seems unnecessary to default to an empty array, just let it throw `this.tests is undefined` if it is misused.

vdemedes:
PR updated, implemented @jamestalmage's suggestions.

jamestalmage:
Uh oh :cry: 

```js
// many-async.js
import test from 'ava';

  test.serial('test' + i, () => {
    return new Promise(resolve => setImmediate(resolve))
  })
}
```

This PR:
```
ava --verbose many-async.js  7.15s user 0.53s system 113% cpu 6.752 total
```

Master:
```
ava --verbose many-async.js  3.76s user 0.46s system 123% cpu 3.421 total
```

vdemedes:
Oh what the hell :D

jamestalmage:
```js
// alternating-async.js

import test from 'ava';

  if (i % 2) {
    test.serial('test' + i, () => {
      return new Promise(resolve => setImmediate(resolve))
    })
  } else {
    test.serial('test' + i, () => {})
  }
}
```

This PR:
```
ava --verbose alternating-async.js  4.52s user 0.44s system 119% cpu 4.158 total
```

Master:
```
ava --verbose alternating-async.js  2.94s user 0.42s system 128% cpu 2.608 total
```

jamestalmage:
I think `Promise.each` in `bluebird` has some magic we lose out on with this implementation.

vdemedes:
Ok, I figured it out thanks to iron-node. The fix is amazingly simple - remove this:

```js
Bluebird.longStackTraces();
```

and execution time drops to 2.65s (even faster than master).

It's appeared in this PR, because there's more nesting of promises (no more `Promise.each`) than in master branch.

jamestalmage:
Hmm. Is there a way to create a promise at runtime that does not use longStackTraces, yet keep longStackTraces turned on for user code?

jamestalmage:
http://bluebirdjs.com/docs/api/promise.longstacktraces.html

> Long stack traces cannot be disabled after being enabled, and cannot be enabled after promises have alread been created

With longStackTraces disabled, I am getting:

This PR:
```
ava --verbose many-async.js  2.14s user 0.36s system 136% cpu 1.834 total
```

Master:
```
ava --verbose many-async.js  2.24s user 0.36s system 134% cpu 1.930 total
```

This PR is faster (though I'd say a tenth of a second over 10,000 tests is fairly negligible). I think the decision on what to do here ultimately comes down to how much we want to keep longStackTrace support.

I am of two minds:

1. Merge this PR, and work towards a documented `--debug` flag that would:

   - Turn on longStackTraces
   - Set stackTraceLimit to Infinity (meaning we could pick a lower number by default).
   - Set plan stack limit to Infinity
   - Reduce our list of `internals` that we have `stack-utils` filter.
   - Somehow enable the node debugger?? (maybe that's only possible with a single file and the profile scirpt? I don't know, and googling "Node debugging forked processes" is not providing answers).
 
   The current debug strategy (`DEBUG=ava`), is a bit annoying because we also spit out `ipc` and `time-require` stuff that is helpful for contributors, but not really for users. I guess we could still use the `debug` module like two different flags: (`DEBUG=ava-internals` vs just `DEBUG=ava`).

2. Stick with the implementation that we have, since it is faster over a wider array of situations, and leave longStackTraces on. This keeps things simple.

I think I would prefer `1`. Simplicity is good, but longStackTraces have a real impact on performance (for both implementations).

sindresorhus:
I would go with `1`.

vdemedes:
> The current debug strategy (DEBUG=ava), is a bit annoying because we also spit out ipc and time-require stuff that is helpful for contributors, but not really for users. I guess we could still use the debug module like two different flags: (DEBUG=ava-internals vs just DEBUG=ava).

There is no need for:

```js
require('debug')('ava-internals')
```

We can do:

 ```js
require('debug')('ava:internals')
```
and

```
$ DEBUG=ava:internals
```

to display only internal logs. Same goes for `ipc` and `time-require`:

```js
require('debug')('ava:ipc') // DEBUG=ava:ipc
require('debug')('ava:require') // DEBUG=ava:require
```

Check out docs on using wildcards for `debug` module - https://www.npmjs.com/package/debug#wildcards.

vdemedes:
Ok, since we all agree on this, merging this PR so that I can move on to next ones :)

jamestalmage:
I understand the wildcard usage. I don't think there is anything special about the `-` vs `:` separator. But, `:` is probably a better choice since that is the convention they have established.

I would just prefer `DEBUG=ava*` not include time-require stuff. So maybe just don't include the `ava` prefix for that? Just call it `time-require`? That is a bit against convention, but `time-require` is not using the `debug` module internally, so I think we can get away with it.

:+1: for finer grained debug scopes (`ava:ipc`, etc).

vdemedes:
> Just call it time-require

Agree

PR #496, opened 2016-01-31
Add a benchmark suite

jamestalmage:
Comments welcome:

Collect benchmark data for a branch:

```
$ node bench/run
```

Once you have collected data from two / three branches:

```
$ node bench/compare
```

Pretty Tables!

screenshot 2016-01-31 02 06 05

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 node_modules
 .nyc_output
 coverage
+bench/.results
--- a/bench/compare.js
+++ b/bench/compare.js
@@ -65,9 +65,9 @@ table.push(
 	[''].concat(stats.map(function (stat) {
 		return {
 			content: stat,
-		  colSpan: fileNames.length,
+			colSpan: fileNames.length,
 			hAlign: 'center'
-		}
+		};
 	})),
 	stats.reduce(function (arr) {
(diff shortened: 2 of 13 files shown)


jamestalmage:
You can now launch a subset of the suite:

```
node bench/run.js concurrent/sync.js serial/sync.js -- concurrent/sync.js -- serial/sync.js
```

Note the `--` separator. The above would be the same as benchmarking all three of the following commands.

```
ava concurrent/sync.js serial/sync.js
ava concurrent/sync.js
ava serial/sync.js
```

Also if you are benchmarking a suite that should fail, you must add the `--should-fail` flag in that group

```
node bench/run.js concurrent/sync.js -- --should-fail other/failures.js
```

The above benchmarks two commands, but expects the second one to fail.

jamestalmage:
I think this gets us a good start.

@sindresorhus @vdemedes - Any other benchmark suites that should be added?

sindresorhus:
Looks superb! Such pretty tables :)

I think it would be useful with a quick script to automate gathering data from master and the last commit and then run the tests. This would be the most common operation, so would be good to automate it, but we can do that later.

sindresorhus:
unicorn14

PR #503, opened 2016-02-02
Fix the `--serial` and `--fail-fast` flags.

### LINKED ISSUE #499: Integration test for `serial` option
We should have an integration test for the public API to make sure we don't break serial test running. It has happened before.

https://github.com/sindresorhus/ava/blob/master/test/api.js

It should ensure tests are not run concurrently.

jamestalmage:
Also fixes the test that was supposed to cover `--serial`.

Fixes #499

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,7 +10,7 @@ var log = require('./lib/logger');
 // note that test files have require('ava')
 require('./lib/test-worker').avaRequired = true;
 
-var opts = JSON.parse(process.argv[2]);
+var opts = globals.options;
 var runner = new Runner(opts);
 
 // check if the test is being run without AVA cli
--- a/index.js
+++ b/index.js
@@ -11,7 +11,10 @@ var log = require('./lib/logger');
 require('./lib/test-worker').avaRequired = true;
 
 var opts = globals.options;
-var runner = new Runner(opts);
+var runner = new Runner({
+	serial: Boolean(opts.serial),
+	bail: Boolean(opts.failFast)
+});
 
 // check if the test is being run without AVA cli
 var isForked = typeof process.send === 'function';
--- a/index.js
+++ b/index.js
@@ -12,8 +12,8 @@ require('./lib/test-worker').avaRequired = true;
 
 var opts = globals.options;
 var runner = new Runner({
-	serial: Boolean(opts.serial),
-	bail: Boolean(opts.failFast)
+	serial: opts.serial,
+	bail: opts.failFast
 });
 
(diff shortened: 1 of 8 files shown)


jamestalmage:
I am not really sure why the last commit is more performant. It would obviously be more performant to extend the prototype if we were creating many Runner instances, but we only create one per process.

Two comparisons before the last commit:

screenshot 2016-02-02 16 35 56
screenshot 2016-02-02 16 35 23

after the last commit:

screenshot 2016-02-02 16 49 16

index.js
```diff
@@ -10,8 +10,11 @@ var log = require('./lib/logger');
 // note that test files have require('ava')
 require('./lib/test-worker').avaRequired = true;
 
-var opts = JSON.parse(process.argv[2]);
-var runner = new Runner(opts);
+var opts = globals.options;
+var runner = new Runner({
+	serial: Boolean(opts.serial),
+	bail: Boolean(opts.failFast)
```
sindresorhus: Why do you need to coerce these to booleans? If it's needed it should happen at the source where they're gotten/defined.

test/runner.js
```diff
@@ -209,3 +209,111 @@ test('only test', function (t) {
 		t.end();
 	});
 });
+
+test('options.serial forces all tests to be serial', function (t) {
+	t.plan(1);
```
sindresorhus: add an empty line below `t.plan(1);`

lib/test-collection.js
```diff
@@ -121,7 +121,8 @@ function buildTestWithHooks(entry, beforeEach, afterEach, report) {
 	return arr.concat(buildHooks(afterEach, entry.title, contextRef, report));
 }
 
-TestCollection.prototype.buildPhases = function (report) {
+TestCollection.prototype.buildPhases = function (report, bail) {
+	bail = Boolean(bail);
```
sindresorhus: Same here. Can you coerce it to a boolean at the source?

sindresorhus:
:+1: Generally looks good :)

test/runner.js
jamestalmage: done

jamestalmage:
@sindresorhus - made all the recommended changes except the `global.Promise = ` one.

sindresorhus:
You missed https://github.com/sindresorhus/ava/pull/503/files#r51689060 ;)

jamestalmage:
OK. Done now.

sindresorhus:
Thanks for fixing so quickly :)

Issue #162, opened 2015-11-08
Website


sindresorhus:
This is just to gather ideas. Not going to do a website until after AVA 1.0.0, as we want to focus on AVA itself, not marketing.

sindresorhus:
Would be nice to embed the Gitter chat on the website: https://sidecar.gitter.im

vdemedes:
How about a live code playground?

sindresorhus:
@vdemedes How would it work exactly? Not sure I understand the purpose.

vdemedes:
@sindresorhus We'd have a text editor on the left side and live output (result of running ava) on the right side. That way we could demonstrate our nice stuff, like super dope assertions.

vdemedes:
Although we'd need AVA to have support for browsers.

sindresorhus:
Oh yeah, that would be cool. We'll have browser support by the time we do a website anyways.

Qix-:
I don't think a live editor will help since this is a test framework. What would you have to test? Make sure to attach everything to analytics to see what's actually being used. :+1:

sindresorhus:
@Qix- Just to give the user a feel of how nice and simple the test syntax is and what the output will look like.

sotojuan:
The main thing I can think of is lots of examples for different kinds of tests. Simple assertions, testing promises/async, testing APIs, etc. Not too many (overwhelming users is bad), but examples help newbies a lot. They don't have to be interactive, just the code and what the output would be.

kentcdodds:
When you've got the website up, if you wanna have a listing for users, you can add PayPal to that list. We're happily using it as of today :-) [PayPal.svg](https://drive.google.com/open?id=0BxZDtibcRzVWUDBHTTZhZVpDQjQ)

jamestalmage:
@kentcdodds - That's awesome news!

denar90:
Hi guys. Talking about site. It can have pages generated form docs and pages with some content, like authors, links to some resources, compare stuff etc. So I propose to build all stuff using grunt/gulp etc. Build tool will grab all content and make html from it, would it be doc file or jade or whatever. 
So site can consist of pages like
 - main page (why AVA, AVA vs other, contributors)
 - sources/plugins
 - docs
 - examples
 - changelog
What do you think guys?
Also what do you think about creating repo in AVA org where issues can be created? Eg. I can prepare some PR's and issues while core team is busy with v1. Someone from core team can from time to time look at it.

[...]

PR #506, opened 2016-02-03
Doc: Generate gh-pages base on MD files with Docpress

zckrs:
Hi,

With [Docpress](https://github.com/docpress/docpress) we can have a github pages  who based content on `docs` folder.

Temporarily you can view result here http://zckrs.github.io/ava/ pushed by [Travis build](https://travis-ci.org/zckrs/ava/jobs/106703077#L1634).

If you are interested and ready to merge I will configure for the source repo and squash commit.

**Note**:
* media moved into docs folder


sindresorhus:
Looks good, but not sure of its use-case when it's just a rendering of the readme.

zckrs:
Commits squashed

* 10ad1a2 Move media into docs/media
* 076aa09 Use "docpress"
* 3f96378 Auto build/deploy by TravisCI

***

Documentations accessible via a web application with a navigation can help some users.
And AVA can be more popular with it.

zckrs:
For deployement by TravisCI you should add env variable `GITHUB_TOKEN` in https://travis-ci.org/sindresorhus/ava/settings

sindresorhus:
> Documentations accessible via a web application with a navigation can help some users.

We already have a `Table of Contents` menu in the readme.

> And AVA can be more popular with it.

Not really a concern right now.

-

None of the arguments are very convincing.

@vdemedes @jamestalmage What do you think about adding this?

jamestalmage:
I think it looks nice. I like the Nav. I like that we get it almost for free just by generating a Readme.

@zckrs - You made a number of edits (moving things in to doc folder, etc). That I don't understand why they were necessary. I'm assuming that's required by Docpress? That would be my one big objection. I don't want to spend time trying to make my Readme acceptable to some third party post-processor.

My gut is that it's got potential, but maybe a bit premature to start focusing on this / letting it take up the maintainers time.

zckrs:
@jamestalmage yes I moved media folder into docs because Docpress is based on `docs` to resolve path.

vdemedes:
I agree with @sindresorhus point, it's a pretty wrapper for the exactly same content. I don't see how it would help AVA become more popular.

Moreover, we have a website planned when we reach near-1.0 release (#162).

@zckrs We have to pass on this PR for now, but thank you for your work and effort to contribute to AVA! Little tip, to avoid these kinds of situations, where work does not get merged, I recommend opening an issue first and see what other people think about your idea.

zckrs:
No problem
You are right I should open issue but I like coding ^^

Issue #507, opened 2016-02-04
`t.doesNotThrow` → `t.notThrows`


SamVerschueren:
When testing if a function throws an error, this can be done with `t.throws`. If you want to test that the method doesn't throw an error, you can use `t.doesNotThrow`. I always tend to use `t.notThrows` instead because it's more in line with the other assertions.

- `ok` ↔ `notOk`
- `same` ↔ `notSame`

So `doesNotThrow` doesn't feel quite right to me.

sindresorhus:
:+1: This has irked me too. `t.doesNotThrow` was initially used to be compatible with Node.js core assert, but we no longer care about that, so I think `t.notThrows` would be a good change. We can keep an undocumented alias to `t.doesNotThrow` for backwards-compatibility.

PR #508, opened 2016-02-04
`t.doesNotThrow` → `t.notThrows`

sotojuan:
This helps with #507 by renaming `t.doesNotThrow` to `t.notThrows`, updating tests and documentation, and adding a `t.doesNotThrow` alias for backwards compatibility.

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -4,6 +4,7 @@ var deepEqual = require('deeper');
 var observableToPromise = require('observable-to-promise');
 var isObservable = require('is-observable');
 var isPromise = require('is-promise');
+var util = require('util');
 var x = module.exports;
 
 Object.defineProperty(x, 'AssertionError', {value: assert.AssertionError});
@@ -122,7 +123,7 @@ x.notThrows = function (fn, msg) {
 	}
 };
 
-x.doesNotThrow = x.notThrows;
+x.doesNotThrow = util.deprecate(x.notThrows, 't.doesNotThrow is renamed to t.notThrows. The old name still works, but will be removed in AVA 1.0.0. Update your references.');
 
 x.regex = function (contents, regex, msg) {
 	test(regex.test(contents), create(regex, contents, '===', msg, x.regex));
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -110,7 +110,7 @@ x.notThrows = function (fn, msg) {
 	if (isPromise(fn)) {
 		return fn
 			.catch(function (err) {
-				x.doesNotThrow(function () {
+				x.notThrows(function () {
 					throw err;
 				}, msg);
 			});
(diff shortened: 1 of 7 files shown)


SamVerschueren:
I would remove the alias. This is a breaking change, no need to keep that stuff.

But off course the decision is up to the maintainers :).

sotojuan:
Yep, I only included it because Sindre mentioned in the issue.

SamVerschueren:
Ah missed that one :)

lib/assert.js
```diff
@@ -118,10 +118,12 @@ x.doesNotThrow = function (fn, msg) {
 	try {
 		assert.doesNotThrow(fn, msg);
 	} catch (err) {
-		test(false, create(err.actual, err.expected, err.operator, err.message, x.doesNotThrow));
+		test(false, create(err.actual, err.expected, err.operator, err.message, x.notThrows));
 	}
 };
 
+x.doesNotThrow = x.notThrows;
```
sindresorhus: Maybe we should use `util.deprecate` on this one? // @jamestalmage @vdemedes

jamestalmage: :+1: 
It would be good to start doing that to anything we think we want to drop with `1.0.0`

sindresorhus: Alright. The message could say:

> `t.doesNotThrow` is renamed to `t.notThrows`. The old name still works, but will be removed in AVA 1.0.0. Update your references.

vdemedes: Clear & informative, I like it!

lib/assert.js
```diff
@@ -118,10 +119,12 @@ x.doesNotThrow = function (fn, msg) {
 	try {
 		assert.doesNotThrow(fn, msg);
 	} catch (err) {
-		test(false, create(err.actual, err.expected, err.operator, err.message, x.doesNotThrow));
+		test(false, create(err.actual, err.expected, err.operator, err.message, x.notThrows));
 	}
 };
 
+x.doesNotThrow = util.deprecate(x.notThrows, 't.doesNotThrow is renamed to t.notThrows. The old name still works, but will be removed in AVA 1.0.0. Update your references.');
```
sindresorhus: Can you alias it first, then deprecate `x.doesNotThrow` (not `x.notThrows) later?

sindresorhus:
You also need to update the reference in `lib/enhance-assert.js`, and in test/test.js:455,.

sotojuan:
I went and fixes all references to `doesNotThrow` in the other tests, working now. I was wondering what to do with the test for the alias because it prints the deprecation notice when the tests run (unless that's not a problem).

sindresorhus:
I don't see your changes. I think you forgot to push.

> (unless that's not a problem).

That's not a problem.

sotojuan:
Oh I haven't pushed yet (been busy), was just asking! And thanks, will push it later :-)

sotojuan:
Let me know if it's ready for a squash!

sindresorhus:
Looks good. Nice work @sotojuan :)

Issue #459, opened 2016-01-20
broken stack trace if error is not an Error


jokeyrhyme:
I used [pify](https://github.com/sindresorhus/pify) on the "fs" built-in, and used the pify'ed [`fs.exists`](https://nodejs.org/dist/latest-v5.x/docs/api/fs.html#fs_fs_exists_path_callback)`()` in an ava test. This is one of those rare Node.js built-ins that invokes its callback like `callback(result)` and _not_ the conventional `callback(err, result)`.

This meant that the pify'ed version returned a rejected Promise, with a boolean value, instead of an Error value. Note: I will follow the recommendations and drop `fs.exists()`, but I'm just describing my context.

Anyhow, the stack trace produced by ava was not helpful in determining the cause here.

You can simulate my issue with the following reduced test code:

```js
const test = require('ava')
test('non-Error error', () => Promise.reject(true))
```

Here is the output that ava produces:

> TypeError: Cannot read property 'split' of null
>     at beautifyStack ([redacted]/node_modules/ava/lib/beautify-stack.js:14:62)
>     at [redacted]/node_modules/ava/lib/reporters/mini.js:107:29
>     at Array.forEach (native)
>     at MiniReporter.finish ([redacted]/node_modules/ava/lib/reporters/mini.js:95:19)
>     at Logger.finish ([redacted]/node_modules/ava/lib/logger.js:45:27)
>     at [redacted]/node_modules/ava/cli.js:112:10
> From previous event:
>     at Object.<anonymous> ([redacted]/node_modules/ava/cli.js:111:3)
>     at Module._compile (module.js:397:26)
>     at Module.replacementCompile ([redacted]/node_modules/append-transform/index.js:53:13)
>     at Module._extensions..js (module.js:404:10)
>     at Object.<anonymous> ([redacted]/node_modules/append-transform/index.js:57:4)
>     at Module.load (module.js:343:32)
>     at Function.Module._load (module.js:300:12)
>     at Function.Module.runMain (module.js:429:10)
>     at Function.runMain (/Users/ron/.node-spawn-wrap-10240-f3833035c112/node:40:10)
>     at Object.<anonymous> ([redacted]/node_modules/nyc/bin/nyc.js:19:6)
>     at Module._compile (module.js:397:26)
>     at Object.Module._extensions..js (module.js:404:10)
>     at Module.load (module.js:343:32)
>     at Function.Module._load (module.js:300:12)
>     at Function.Module.runMain (module.js:429:10)
[...]

PR #510, opened 2016-02-04
fix beautify stack split error - fixes #509

### LINKED ISSUE #509: Cannot read property 'split' of null
This might be related to #459 

### index.js

```js
module.exports = function () {
    return Promise.resolve('foo');
};
```

### test.js

```js
import test from 'ava';
import m from './';

test(t => {
    t.throws(m(), 'foo');
});
```

**Current**: `TypeError: Cannot read property 'split' of null`
**Expected**: `Missing expected exception` (or something like that)

SamVerschueren:
This tiny PR fixes #509 and #459 when `beautifyStack` returns `null`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -117,7 +117,7 @@
     "serialize-error": "^1.1.0",
     "set-immediate-shim": "^1.0.1",
     "source-map-support": "^0.4.0",
-    "stack-utils": "^0.3.0",
+    "stack-utils": "^0.4.0",
     "strip-bom": "^2.0.0",
     "time-require": "^0.1.2",
     "unique-temp-dir": "^1.0.0",


jamestalmage:
Under what conditions does `stack-utils` return null? That seems like a poor choice. Sam, would you mind taking a look and seeing if it wouldn't be better to modify `stack-utils`? Maybe it should just return an empty string instead of null?

Part of any stack-utils PR would need to include checking how changes would affect `node-tap`.

Sorry for complicating this. I'd do it myself but I'm slammed for the next few weeks

SamVerschueren:
It's the last line of the [`clean`](https://github.com/tapjs/stack-utils/blob/master/index.js#L84) function. The stack trace of the error looks like this

```
/Users/sam/Projects/playground/ava-test/node_modules/ava/lib/assert.js:78:7
From previous event:
    Test.exit (/Users/sam/Projects/playground/ava-test/node_modules/ava/lib/test.js:195:10)
    Test.run (/Users/sam/Projects/playground/ava-test/node_modules/ava/lib/test.js:142:8)
    Runner._runTest (/Users/sam/Projects/playground/ava-test/node_modules/ava/lib/runner.js:114:14)
    Runner.<anonymous> (/Users/sam/Projects/playground/ava-test/node_modules/ava/lib/runner.js:101:15)
From previous event:
    eachSeries (/Users/sam/Projects/playground/ava-test/node_modules/ava/lib/runner.js:39:17)
    Runner._runTestWithHooks (/Users/sam/Projects/playground/ava-test/node_modules/ava/lib/runner.js:91:9)
    each (/Users/sam/Projects/playground/ava-test/node_modules/ava/lib/runner.js:35:27)
    Runner._runConcurrent (/Users/sam/Projects/playground/ava-test/node_modules/ava/lib/runner.js:124:9)
    /Users/sam/Projects/playground/ava-test/node_modules/ava/lib/runner.js:182:16
From previous event:
    Runner.run (/Users/sam/Projects/playground/ava-test/node_modules/ava/lib/runner.js:181:4)
    process.<anonymous> (/Users/sam/Projects/playground/ava-test/node_modules/ava/index.js:85:10)
    process.<anonymous> (/Users/sam/Projects/playground/ava-test/node_modules/ava/lib/test-worker.js:102:10)
```

Because of the [internals](https://github.com/sindresorhus/ava/blob/master/lib/beautify-stack.js#L7) of `stack-utils` the `result` array in the clean method is empty which results in `null`.

sindresorhus:
> It's the last line of the clean function.

It should return an empty string there.

SamVerschueren:
There is actually a test that to check if it returns null if it are all internals.

https://github.com/tapjs/stack-utils/blob/master/test/test.js#L57-L60

Will create an issue and do a PR.

sindresorhus:
> There is actually a test that to check if it returns null if it are all internals.

Definitely not what *I* would expect when the method name is `.clean()`.

> Will create an issue and do a PR.

:+1:

SamVerschueren:
@jamestalmage bumped `stack-utils` and removed previous changes.

vdemedes:
Thanks Sam!

SamVerschueren:
You're welcome :)

Issue #481, opened 2016-01-29
Recipe for using a plain function instead of `beforeEach` hook


sindresorhus:
In many cases you could just use a plain function instead of a `beforeEach` hook. In the below case it's both more clear and less code.

``` js
import test from 'ava';

const setup = () => {
    doStuff();

    return {
        unicorn: fn('unicorn')
        rainbow: fn('rainbow')
    };
}

test('foo', t => {
    const {unicorn, rainbow} = setup();
    t.true(unicorn('foo'));
    t.true(rainbow('foo'));
});

test('bar', t => {
    const {unicorn, rainbow} = setup();
    t.true(unicorn('bar'));
    t.true(rainbow('bar'));
});
```

With `beforeEach` it would be:

``` js
import test from 'ava';

test.beforeEach(t => {
    doStuff();

    t.context = {
        unicorn: fn('unicorn')
        rainbow: fn('rainbow')
    };
}

test('foo', t => {
    const {unicorn, rainbow} = t.context;
    t.true(unicorn('foo'));
    t.true(rainbow('foo'));
});

test('bar', t => {
    const {unicorn, rainbow} = t.context;
    t.true(unicorn('bar'));
    t.true(rainbow('bar'));
});
```

Still, there are benefits of `beforeEach` too. They're not mutually exclusive. `beforeEach` has benefits like supporting non-promise asynchronicity. Better output when a hook fails. Void hook usage doesn't require a call in each test. And more.

jamestalmage:
The real advantage of the setup function approach is that you could have more than one, and call them based on what you need in a particular test. (If you interact with the file system, perhaps you run a few tests using `mock-fs`, and then a few that use the real file system and a temp dir or something).

There is a (very small) performance benefit to not having **any** beforeEach or afterEach tests. It's not worth worrying about unless you are dealing with hundreds (maybe even thousands) of tests in a single file though.

I have come to prefer the `setup()` approach personally. It does not require others to be familiar with AVA's `t.context`, there is no magic.

JasonRitchie:
Can I pick this up?

sindresorhus:
@JasonRitchie Of course. Go ahead :)

PR #512, opened 2016-02-05
Added recipe for plain function

jarrettmeyer:
This fixes issue #481. Adds documentation recipe for functions vs. beforeEach.

### DIFF
--- a/docs/recipes/setup-vs-beforeEach.md
+++ b/docs/recipes/setup-vs-beforeEach.md
@@ -1,10 +1,10 @@
 # Setup Functions vs. beforeEach Functions
 
-It can be stylistically beneficial to have individual setup functions called from within the test scenario, instead of a single `beforeEach` function.
+In some cases, using plain setup functions is preferred to using `beforeEach` for your tests. Since the `beforeEach` function is called for every test, using independent functions allows for easier modifications between test scenarios.
 
 ## Complex test setup
 
-In this example, we have both test setup and modifications within each test.
+In this example, we have both a `beforeEach` function, and then more modifications within each test.
 
 ```js
 test.beforeEach(t => {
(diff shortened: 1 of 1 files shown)


vdemedes:
I wouldn't focus on "stylistic" side of manually doing setup/teardown. I'd suggest focusing on actual benefits it gives to developers, which you actually mentioned at the end:

> 1. Setup functions allow for easier modifications between test runs.

If the recipe focuses on a code style, it's not really a recipe, it's just an opinion/preference. Recipes are meant to be official, ready-to-go solutions for questions/problems, that often occur while using AVA.

Everything else looks good, thanks @jarrettmeyer, looking forward to checking out and merging your updated recipe!

jarrettmeyer:
@vdemedes Your recommendation is a change to the supporting text? Are the examples good/bad/necessary?

vdemedes:
Yes, the supporting text. I think the code examples are fine. But I'd focus reader's attention to the point, that using manual setup/teardown functions gives you more flexibility. AVA's before/after hooks can't be turned off for certain tests, so here's exactly where "manual work" pays off.

The point of style, looking better and having straightforward tests is secondary, but also worth mentioning.

vdemedes:
And @jamestalmage pointed out another advantage:

> The real advantage of the setup function approach is that you could have more than one, and call them based on what you need in a particular test.

docs/recipes/setup-vs-beforeEach.md
```diff
@@ -0,0 +1,125 @@
+# Setup Functions vs. beforeEach Functions
```
sindresorhus:
```
# Setup functions vs. `beforeEach` hooks
```

sindresorhus:
I would also at the end mention some of the benefits of using `beforeEach`. From #481:

> Still, there are benefits of beforeEach too. They're not mutually exclusive. beforeEach has benefits like supporting non-promise asynchronicity. Better output when a hook fails. Void hook usage doesn't require a call in each test. And more.

There might be more I haven't thought of.

docs/recipes/setup-vs-beforeEach.md
```diff
@@ -0,0 +1,125 @@
...
+  t.true(await isValid);
+});
+
+test('authenticating with an invalid username', async t => {
+  t.context.credentials.username = 'bad_username';
+  const isValid = t.context.authenticator.authenticate(t.context.credentials);
+  t.false(await isValid);
+});
+
+test('authenticating with valid credentials', async t => {
```
forresst: Replace _'authenticating with valid credentials'_ by _'authenticating with **in**valid credentials'_ ?

docs/recipes/setup-vs-beforeEach.md
```diff
@@ -0,0 +1,125 @@
...
+  t.true(await isValid);
+});
+
+test('authenticating with an invalid username', async t => {
+  const {authenticator, credentials} = setup({username: 'bad_username'});
+  const isValid = authenticator.authenticate(credentials);
+  t.false(await isValid);
+});
+
+test('authenticating with valid credentials', async t => {
```
forresst: Replace 'authenticating with valid credentials' by 'authenticating with invalid credentials' ?

sindresorhus:
@jarrettmeyer ping :)

jamestalmage:
@jarrettmeyer Do you have time to follow up on the feedback? Would like to merge this soon.

sindresorhus:
Closing for lack of response by OP.

Issue #479, opened 2016-01-28
Mini reporter should have an empty line before the output


sindresorhus:
<img width="165" alt="screen shot 2016-01-28 at 02 27 36" src="https://cloud.githubusercontent.com/assets/170270/12654362/e223ed70-c5f3-11e5-92df-d8e4edb7dfdb.png">

Should be an empty line above `1 passed`.

Bearlock:
Is this a solved issue? I tested this with the example code provided [here](https://github.com/sindresorhus/ava#create-your-test-file) on my Mac's Iterm 2 terminal and I'm not having the same issue. Looks like this:

![screen shot 2016-01-29 at 10 04 07 am](https://cloud.githubusercontent.com/assets/6089630/12682252/abbf4762-c66f-11e5-9b51-46de24d68a26.png)

Is there something else I would have to do to recreate the issue?

sindresorhus:
@Bearlock This issue is for `master`, not the published release.

Bearlock:
@sindresorhus Ah, understood. Thanks for the clarification.

sotojuan:
This was fixed in #514! Feel free to close it.

PR #514, opened 2016-02-05
Add empty line before output to mini reporter

sotojuan:
This fixe #479 by adding a `\n` to the output of the mini reporter and updating the current tests so they pass.


vdemedes:
Thanks @sotojuan!

PR #527, opened 2016-02-08
Clarified usage example depends on `delay` library

corinna000:
Clarifies that the Usage example depends on the user installing `delay` library to run.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -72,6 +72,8 @@ Install AVA globally `$ npm install --global ava` and run `$ ava --init` (with a
 
 #### Create your test file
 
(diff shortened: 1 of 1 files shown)


vdemedes:
I don't think we need this, as it's quite obvious that it should be installed from npm and is not a core Node.js module.

@sindresorhus @jamestalmage

jamestalmage:
I agree with @vdemedes. Educating users on the basics of `npm` and `import`/`require` are a bit out of scope for AVA's docs. Thank you anyways @corinna000, we appreciate the contribution regardless!

PR #529, opened 2016-02-08
Rename notThrows to notThrow - Fixes #528

### LINKED ISSUE #528: t.notThrows - Should possibly be renamed to t.notThrow - `t.doesNotThrow` → `t.notThrows` → `t.notThrow` #508
After looking into removing the deprecation warning that (https://github.com/sindresorhus/ava/pull/508) we are now seeing when we run `npm test`. I dug into tap a little and they are using t.notThrow which the PR changed associated with this added notThrows. If you change the test to t.notThrows it can't find that function because tap doesn't have it. If you use notThrow it works fine. 

From: http://www.node-tap.org/asserts/

t.doesNotThrow(fn, message, extra)

Verify that the provided function does not throw.

If the function has a name, and the message is not provided, then the function name will be used as the message.

If the function is not provided, then this will be treated as a todo test.

Note: if an error is encountered unexpectedly, it's often better to simply throw it. The Test object will handle this as a failure.

Synonyms:** t.notThrow**

rclanan:
Fixed all references to doesNotThrow

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -102,7 +102,7 @@ x.throws = function (fn, err, msg) {
 	}
 };
 
-x.notThrows = function (fn, msg) {
+x.notThrow = function (fn, msg) {
 	if (isObservable(fn)) {
 		fn = observableToPromise(fn);
 	}
@@ -110,7 +110,7 @@ x.notThrows = function (fn, msg) {
 	if (isPromise(fn)) {
 		return fn
 			.catch(function (err) {
-				x.notThrows(function () {
+				x.notThrow(function () {
 					throw err;
 				}, msg);
 			});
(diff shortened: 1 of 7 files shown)


vdemedes:
See my comment in #528.

Also, little advice, to avoid working on things that won't get merged or accepted, I recommend discussing it in an issue first and see if people actually want this.

rclanan:
Thanks for the advice. 

I created an issue for it. I created the PR in response to the issue in case it was a "wanted" change. It's easy enough to close out the PR. As for actual work effort, not much time was lost as it was a pretty simple change. If it wasn't something that was a trivial change, I definitely wouldn't have created the PR to go with the issue without discussion first.

sindresorhus:
Thanks for the suggestion, but as commented in https://github.com/sindresorhus/ava/issues/528#issuecomment-181673612, this is not something we're interested in :)

PR #530, opened 2016-02-08
Removed redundant line from assertion error output in mini reporter

### LINKED ISSUE #422: Clean up redundant assertion error output.
Currently the following:

```js
test(t => t.ok(false))
```

Will log the following:

```
false == true
AssertionError: false == true
  ... remaining stack-trace ...
```

This is redundant.

It only happens for enhanced assertions.

I think the solution could be as simple as: `err.message + stripFirstLine(err.stack)`.

naptowncode:
Fixes #422. Removed the `AssertionError: ...` line of output.

### DIFF
--- a/lib/reporters/mini.js
+++ b/lib/reporters/mini.js
@@ -106,7 +106,7 @@ MiniReporter.prototype.finish = function () {
 			var description;
 
 			if (test.error) {
-				description = '  ' + test.error.message + '\n  ' + test.error.stack;
+				description = '  ' + test.error.message + '\n  ' + stripFirstLine(test.error.stack);
 			} else {
 				description = JSON.stringify(test);
 			}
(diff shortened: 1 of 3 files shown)


vdemedes:
Errors are not always instances of `AssertionError`, so we definitely should not just remove the first line of `err.stack`.

naptowncode:
Ah, I see. Should I leave out the `error.message` instead? The message will always be in the first line of the stack trace anyway.

sindresorhus:
> Errors are not always instances of AssertionError, so we definitely should not just remove the first line of err.stack.

@vdemedes But the message is always the first line of the stack no matter what kind of error.

sindresorhus:
@naptowncode This looks good to me. Can you fix the merge conflict?

naptowncode:
@sindresorhus Done, though AppVeyor seems to not like it.

sotojuan:
AppVeyor is very rude sometimes :cry:

sindresorhus:
Thank you @naptowncode :)

PR #531, opened 2016-02-09
Support NODE_PATH in forked process

### LINKED ISSUE #515: `NODE_PATH` does not work
We are using `NODE_PATH=.` in our app to [improve local `require()`](https://gist.github.com/branneman/8048520). AVA (version 0.11.0) has problems with it, though.

**flat-file structure**:
```.js
// foo/index.js
export const foo = 'bar'
```

```.js
// test/foo.js
import test from 'ava'

import { foo } from 'foo'

test('foo should be bar', t => {
  t.same(foo, 'bar')
})
```

```
> NODE_PATH=. ava

  1 failed

  1. foo should be bar

  t.same(foo, 'bar')
         |
         undefined

  AssertionError: undefined === 'bar'
    Test.fn (test/foo.js:6:5)
```

**nested-file structure**
```.js
// foo/bar/index.js
export const foo = 'bar'
```

```.js
// test/foo.js
import test from 'ava'

import { foo } from 'foo/bar'

test('foo should be bar', t => {
  t.same(foo, 'bar')
})
```

```
> NODE_PATH=. ava

Error: Cannot find module 'foo/bar'
    at Function.Module._resolveFilename (module.js:338:15)
    at Function.Module._load (module.js:289:25)
    at Module.require (module.js:366:17)
    at require (module.js:385:17)
    at Object.<anonymous> (/Users/inyono/Projects/flitt/test/foo.js:9:12)
    at Module._compile (module.js:435:26)
    at Object.extensions.(anonymous function) [as .js] (/Users/inyono/Projects/flitt/node_modules/ava/node_modules/require-precompiled/index.js:13:11)
    at Function.Module._load (module.js:313:12)
    at Module.require (module.js:366:17)
    at require (module.js:385:17)
    at Object.<anonymous> (/Users/inyono/Projects/flitt/node_modules/ava/lib/test-worker.js:85:1)
    at Module._compile (module.js:435:26)
    at Object.Module._extensions..js (module.js:442:10)
    at Function.Module._load (module.js:313:12)
    at Function.Module.runMain (module.js:467:10)
    at startup (node.js:136:18)
    at node.js:963:3

  ✖ test/foo.js exited with a non-zero exit code: 1
```

ingro:
This PR tries to fix #515 in a non-destructive way.

Basically it checks for the presence of `NODE_PATH` env variable before forking the process and creates an absolute path from it, so it could be used in `test-worker.js` as additional node module path.

That won't work there because `process.cwd()` points to the forked process test file.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -97,7 +97,7 @@ if (cli.flags.init) {
 var nodePaths;
 if (process.env.NODE_PATH) {
     nodePaths = process.env.NODE_PATH.split(path.delimiter).map(function (p) {
-        return path.resolve(process.cwd(), p);
+        return path.resolve(p);
     });
 } else {
     nodePaths = [];
--- a/cli.js
+++ b/cli.js
@@ -93,11 +93,20 @@ if (cli.flags.init) {
 	return;
 }
 
+var additionalPaths = [];
+if (process.env.NODE_PATH) {
+	var osSplitChar = process.platform === 'win32' ? ';' : ':';
+	process.env.NODE_PATH.split(osSplitChar).forEach(function (additionalPath) {
+		additionalPaths.push(path.resolve(opts.projectRoot, additionalPath));
+	});
+}
+
 var api = new Api(cli.input.length ? cli.input : arrify(conf.files), {
 	failFast: cli.flags.failFast,
 	serial: cli.flags.serial,
 	require: arrify(cli.flags.require),
-	cacheEnabled: cli.flags.cache !== false
+	cacheEnabled: cli.flags.cache !== false,
+	additionalPaths: additionalPaths
 });
 
 var logger = new Logger();
--- a/cli.js
+++ b/cli.js
@@ -17,7 +17,6 @@ if (debug.enabled) {
 	require('time-require');
 }
 
-var path = require('path');
 var updateNotifier = require('update-notifier');
 var figures = require('figures');
 var arrify = require('arrify');
--- a/cli.js
+++ b/cli.js
@@ -29,6 +29,7 @@ var verboseReporter = require('./lib/reporters/verbose');
 var miniReporter = require('./lib/reporters/mini');
 var tapReporter = require('./lib/reporters/tap');
 var Logger = require('./lib/logger');
+var watcher = require('./lib/watcher');
 var Api = require('./api');
 
 // Bluebird specific
@@ -48,6 +49,9 @@ var cli = meow([
 	'  --tap, -t        Generate TAP output',
 	'  --verbose, -v    Enable verbose output',
 	'  --no-cache       Disable the transpiler cache',
+	// Leave --watch and --sources undocumented until they're stable enough
+	// '  --watch, -w      Re-run tests when tests and source files change',
+	// '  --source         Pattern to match source files so tests can be re-run (Can be repeated)',
 	'',
 	'Examples',
 	'  ava',
@@ -62,20 +66,23 @@ var cli = meow([
 ], {
 	string: [
 		'_',
-		'require'
+		'require',
+		'source'
 	],
 	boolean: [
 		'fail-fast',
 		'verbose',
 		'serial',
-		'tap'
+		'tap',
+		'watch'
 	],
 	default: conf,
 	alias: {
 		t: 'tap',
 		v: 'verbose',
 		r: 'require',
-		s: 'serial'
+		s: 'serial',
+		w: 'watch'
 	}
 });
 
@@ -112,17 +119,30 @@ api.on('error', logger.unhandledError);
 api.on('stdout', logger.stdout);
 api.on('stderr', logger.stderr);
 
-api.run()
-	.then(function () {
-		logger.finish();
-		logger.exit(api.failCount > 0 || api.rejectionCount > 0 || api.exceptionCount > 0 ? 1 : 0);
-	})
-	.catch(function (err) {
+if (cli.flags.watch) {
+	try {
+		watcher.start(logger, api, arrify(cli.flags.source), process.stdin);
+	} catch (err) {
 		if (err.name === 'AvaError') {
+			// An AvaError may be thrown if chokidar is not installed. Log it nicely.
 			console.log('  ' + colors.error(figures.cross) + ' ' + err.message);
+			logger.exit(1);
 		} else {
-			console.error(colors.stack(err.stack));
+			// Rethrow so it becomes an uncaught exception.
+			throw err;
 		}
-
-		logger.exit(1);
-	});
+	}
+} else {
+	api.run()
+		.then(function () {
+			logger.finish();
+			logger.exit(api.failCount > 0 || api.rejectionCount > 0 || api.exceptionCount > 0 ? 1 : 0);
+		})
+		.catch(function (err) {
+			// Don't swallow exceptions. Note that any expected error should already
+			// have been logged.
+			setImmediate(function () {
+				throw err;
+			});
+		});
+}
--- a/cli.js
+++ b/cli.js
@@ -96,12 +96,11 @@ if (cli.flags.init) {
 
 var nodePaths;
 if (process.env.NODE_PATH) {
-    var osSplitChar = process.platform === 'win32' ? ';' : ':';
-    nodePaths = process.env.NODE_PATH.split(osSplitChar).map(function (p) {
-        return path.resolve(process.cwd(), p)
+    nodePaths = process.env.NODE_PATH.split(path.delimiter).map(function (p) {
+        return path.resolve(process.cwd(), p);
     });
 } else {
-    nodePaths = []
+    nodePaths = [];
 }
 
 var api = new Api(cli.input.length ? cli.input : arrify(conf.files), {
--- a/cli.js
+++ b/cli.js
@@ -94,21 +94,11 @@ if (cli.flags.init) {
 	return;
 }
 
-var nodePaths;
-if (process.env.NODE_PATH) {
-    nodePaths = process.env.NODE_PATH.split(path.delimiter).map(function (p) {
-        return path.resolve(p);
-    });
-} else {
-    nodePaths = [];
-}
-
 var api = new Api(cli.input.length ? cli.input : arrify(conf.files), {
 	failFast: cli.flags.failFast,
 	serial: cli.flags.serial,
 	require: arrify(cli.flags.require),
-	cacheEnabled: cli.flags.cache !== false,
-	nodePaths: nodePaths
+	cacheEnabled: cli.flags.cache !== false
 });
 
 var logger = new Logger();
--- a/cli.js
+++ b/cli.js
@@ -17,6 +17,7 @@ if (debug.enabled) {
 	require('time-require');
 }
 
+var path = require('path');
 var updateNotifier = require('update-notifier');
 var figures = require('figures');
 var arrify = require('arrify');
@@ -93,12 +94,14 @@ if (cli.flags.init) {
 	return;
 }
 
-var additionalPaths = [];
+var nodePaths;
 if (process.env.NODE_PATH) {
-	var osSplitChar = process.platform === 'win32' ? ';' : ':';
-	process.env.NODE_PATH.split(osSplitChar).forEach(function (additionalPath) {
-		additionalPaths.push(path.resolve(opts.projectRoot, additionalPath));
-	});
+    var osSplitChar = process.platform === 'win32' ? ';' : ':';
+    nodePaths = process.env.NODE_PATH.split(osSplitChar).map(function (p) {
+        return path.resolve(process.cwd(), p)
+    });
+} else {
+    nodePaths = []
 }
 
 var api = new Api(cli.input.length ? cli.input : arrify(conf.files), {
@@ -106,7 +109,7 @@ var api = new Api(cli.input.length ? cli.input : arrify(conf.files), {
 	serial: cli.flags.serial,
 	require: arrify(cli.flags.require),
 	cacheEnabled: cli.flags.cache !== false,
-	additionalPaths: additionalPaths
+	nodePaths: nodePaths
 });
 
 var logger = new Logger();
--- a/lib/fork.js
+++ b/lib/fork.js
@@ -14,7 +14,7 @@ module.exports = function (file, opts) {
 			columns: process.stdout.columns,
 			rows: process.stdout.rows
 		} : false,
-		projectRoot: process.cwd()
+		additionalPaths: opts.additionalPaths
 	}, opts);
 
 	var ps = childProcess.fork(path.join(__dirname, 'test-worker.js'), [JSON.stringify(opts)], {
--- a/lib/fork.js
+++ b/lib/fork.js
@@ -16,9 +16,25 @@ module.exports = function (file, opts) {
 		} : false
 	}, opts);
 
+	var env = process.env;
+
+	if (process.env.NODE_PATH) {
+	    var nodePath = process.env.NODE_PATH
+	        .split(path.delimiter)
+	        .map(function (p) {
+	            return path.resolve(p);
+	        })
+	        .join(path.delimiter);
+
+		env = objectAssign({
+			NODE_PATH: nodePath
+		}, env);
+	}
+
 	var ps = childProcess.fork(path.join(__dirname, 'test-worker.js'), [JSON.stringify(opts)], {
 		cwd: path.dirname(file),
-		silent: true
+		silent: true,
+		env: env
 	});
 
 	var relFile = path.relative('.', file);
--- a/lib/fork.js
+++ b/lib/fork.js
@@ -13,7 +13,8 @@ module.exports = function (file, opts) {
 		tty: process.stdout.isTTY ? {
 			columns: process.stdout.columns,
 			rows: process.stdout.rows
-		} : false
+		} : false,
+		projectRoot: process.cwd()
 	}, opts);
 
 	var ps = childProcess.fork(path.join(__dirname, 'test-worker.js'), [JSON.stringify(opts)], {
--- a/lib/fork.js
+++ b/lib/fork.js
@@ -19,12 +19,12 @@ module.exports = function (file, opts) {
 	var env = process.env;
 
 	if (process.env.NODE_PATH) {
-	    var nodePath = process.env.NODE_PATH
-	        .split(path.delimiter)
-	        .map(function (p) {
-	            return path.resolve(p);
-	        })
-	        .join(path.delimiter);
+		var nodePath = process.env.NODE_PATH
+			.split(path.delimiter)
+			.map(function (p) {
+				return path.resolve(p);
+			})
+			.join(path.delimiter);
 
 		env = objectAssign({
 			NODE_PATH: nodePath
--- a/lib/fork.js
+++ b/lib/fork.js
@@ -7,6 +7,21 @@ var debug = require('debug')('ava');
 var AvaError = require('./ava-error');
 var send = require('./send');
 
+var env = process.env;
+
+if (process.env.NODE_PATH) {
+	var nodePath = process.env.NODE_PATH
+		.split(path.delimiter)
+		.map(function (p) {
+			return path.resolve(p);
+		})
+		.join(path.delimiter);
+
+	env = objectAssign({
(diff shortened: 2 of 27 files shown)


novemberborn:
Rather than modifying `process.env.NODE_PATH` in `api.js` you should probably pass a custom value in `fork.js`: https://github.com/sindresorhus/ava/blob/bdb8b6ebaa05f1dc31cdab7c84aed7d92317b8ed/lib/fork.js#L19:L22

lib/test-worker.js
```diff
@@ -86,6 +86,9 @@ var oldNodeModulesPaths = module.constructor._nodeModulePaths;
 module.constructor._nodeModulePaths = function () {
 	var ret = oldNodeModulesPaths.apply(this, arguments);
 	ret.push(nodeModulesDir);
+	if (process.env.NODE_PATH) {
+		ret.push(process.env.NODE_PATH);
```
novemberborn: `NODE_PATH` can contain multiple values. Not sure if the env var needs to be split and concatenated with `ret` here. See <https://nodejs.org/api/modules.html#modules_loading_from_the_global_folders> for more.

ingro:
Thanks for your input @novemberborn , I made the following changes:
- removed the ugly process.env.NODE_PATH modify in `api.js`
- added an option for the forked process (`projectRoot`) which contains the original path from where ava was called;
- cycled through possible multiple values of process.env.NODE_PATH (with a simple OS check) and join their path with the one supplied by `projectRoot`.

lib/fork.js
```diff
@@ -13,7 +13,8 @@ module.exports = function (file, opts) {
 		tty: process.stdout.isTTY ? {
 			columns: process.stdout.columns,
 			rows: process.stdout.rows
-		} : false
+		} : false,
+		projectRoot: process.cwd()
```
novemberborn: This isn't necessarily the project root. Perhaps `parentWorkingDirectory`? Bit verbose though.

lib/test-worker.js
```diff
@@ -86,6 +86,12 @@ var oldNodeModulesPaths = module.constructor._nodeModulePaths;
 module.constructor._nodeModulePaths = function () {
 	var ret = oldNodeModulesPaths.apply(this, arguments);
 	ret.push(nodeModulesDir);
+	if (process.env.NODE_PATH) {
+		var osSplitChar = /^win/.test(process.platform) ? ';' : ',';
```
novemberborn: Perhaps use [`is-windows`](https://www.npmjs.com/package/is-windows). A separate package is likely preferable over inline regex fu.

---

> If the NODE_PATH environment variable is set to a colon-delimited list of absolute paths, then Node.js will search those paths for modules if they are not found elsewhere. (Note: On Windows, NODE_PATH is delimited by semicolons instead of colons.)

[(source)](https://nodejs.org/api/modules.html#modules_loading_from_the_global_folders)

The non-Windows separators are colons, not commas.

lib/test-worker.js
```diff
@@ -86,6 +86,12 @@ var oldNodeModulesPaths = module.constructor._nodeModulePaths;
...
+		process.env.NODE_PATH.split(osSplitChar).forEach(function (additionalPath) {
+			ret.push(path.join(opts.projectRoot, additionalPath));
```
novemberborn: Should use [`path.resolve`](https://nodejs.org/api/path.html#path_path_resolve_from_to) in case `additionalPath` is already absolute.

lib/test-worker.js
ingro: I didn't want to add another dependency, but if it's fine for you I think too that this would be a better solution.

novemberborn:
Perhaps resolve the additional, absolute `NODE_PATH`s in the main process, once, and then pass the array through the `opts` to the `test-worker`.

You should add a test, probably in `test/fork.js`. Note that tests are written using `node-tap`. Maybe use [`t.spawn`](http://www.node-tap.org/advanced/#t-spawn-command-arguments-options-name-extra) to spawn another test file (from the `fixtures` directory) with a `NODE_PATH` environment variable, then in that file use the `lib/fork` module to run an AVA test. You could set up `NODE_PATH` so the final test can do an `import` and have it be resolved correctly. I'm in Gitter if that wasn't clear https://gitter.im/sindresorhus/ava

lib/test-worker.js
novemberborn: I think it is, but I don't have final say ;-)

lib/fork.js
ingro: What about `parentCwd`?

novemberborn: Sure

lib/test-worker.js
ingro: Yeah you're right, I'm really bad at this.

novemberborn: No worries! This stuff is maddeningly difficult.

lib/test-worker.js
vdemedes: No need to use `is-windows`, since all of its logic is 1 line (https://github.com/jonschlinkert/is-windows/blob/master/index.js#L20):

```js
process.platform === 'win32'
```

novemberborn: Heh yea, fair enough.

ingro: So I guess I could leave the test as it is (except for the wrong separators).

novemberborn: Use `process.platform === 'win32'` though rather than a regular expression.

lib/fork.js
```diff
@@ -13,7 +13,8 @@ module.exports = function (file, opts) {
 		tty: process.stdout.isTTY ? {
 			columns: process.stdout.columns,
 			rows: process.stdout.rows
-		} : false
+		} : false,
+		additionalPaths: opts.additionalPaths
```
novemberborn: `Object.assign()` already takes care of mixing in the `opts`.

lib/test-worker.js
```diff
@@ -86,6 +86,11 @@ var oldNodeModulesPaths = module.constructor._nodeModulePaths;
 module.constructor._nodeModulePaths = function () {
 	var ret = oldNodeModulesPaths.apply(this, arguments);
 	ret.push(nodeModulesDir);
+	if (opts.additionalPaths && opts.additionalPaths.length > 0) {
+		opts.additionalPaths.forEach(function (additionalPath) {
+			ret.push(additionalPath);
+		});
+	}
```
novemberborn: Could replace with:

```js
return ret.concat(nodeModulesDir, additionalPaths);
```

cli.js
```diff
@@ -93,11 +93,20 @@ if (cli.flags.init) {
 	return;
 }
 
+var additionalPaths = [];
+if (process.env.NODE_PATH) {
+	var osSplitChar = process.platform === 'win32' ? ';' : ':';
+	process.env.NODE_PATH.split(osSplitChar).forEach(function (additionalPath) {
+		additionalPaths.push(path.resolve(opts.projectRoot, additionalPath));
```
novemberborn: `process.cwd()` rather than `opts.projectRoot` right?

Perhaps `nodePaths` as the variable / option name? And personally I'd do something like:

```js
var nodePaths
if (process.env.NODE_PATH) {
	var osSplitChar = process.platform === 'win32' ? ';' : ':';
	nodePaths = process.env.NODE_PATH.split(osSplitChar).map(function (p) {
		return path.resolve(process.cwd(), p)
	});
} else {
	nodePaths = []
}
```

ingro: Yep map is surely more elegant!

test/cli.js
```diff
@@ -15,7 +15,10 @@ function execCli(args, dirname, cb) {
 		dirname = path.join(__dirname, dirname);
 	}
 
-	var env = {};
+	var env = {
+		// This probably should be set only for the corresponding test
```
novemberborn: Yep :)

test/cli.js
```diff
@@ -110,3 +113,10 @@ test('pkg-conf: cli takes precedence', function (t) {
 		t.end();
 	});
 });
+
+test('handles NODE_PATH', function (t) {
+	execCli('fixture/node-paths.js', function (err) {
+		t.notOk(err);
+		t.end();
+	});
+});
```
novemberborn: Could you insert a linebreak at the end of the file?

test/fixture/node-paths/modules/nested/foo.js
```diff
@@ -0,0 +1,3 @@
+module.exports = function() {
+	return 'bar';
+}
```
novemberborn: Needs a semi-colon after the `}`. And could you insert a linebreak at the end of the file?

test/cli.js
```diff
@@ -15,7 +15,10 @@ function execCli(args, dirname, cb) {
 		dirname = path.join(__dirname, dirname);
 	}
 
-	var env = {};
+	var env = {
+		// This probably should be set only for the corresponding test
+		NODE_PATH: 'node-paths/modules'
```
novemberborn: Should test with multiple paths, using the correct separator based on platform.

test/cli.js
ingro: Any suggestion on how to approach this?

novemberborn: Probably by adding an `env` argument after `dirname` and before `cb`. It should be optional though, no changes to the other tests should be necessary.

Could leave `dirname` required if `env` is provided, so `execCli([], { NODE_PATH: 'foo' })` is invalid.

ingro:
I added the test and it seems like it's green now, but I surely forgot something else, let me know if something is missing!

cli.js
```diff
@@ -93,11 +94,22 @@ if (cli.flags.init) {
 	return;
 }
 
+var nodePaths;
+if (process.env.NODE_PATH) {
+    var osSplitChar = process.platform === 'win32' ? ';' : ':';
```
shinnn: Instead you can use `path.delimiter`. https://nodejs.org/api/path.html#path_path_delimiter

cli.js
```diff
@@ -93,11 +94,22 @@ if (cli.flags.init) {
...
+    nodePaths = process.env.NODE_PATH.split(osSplitChar).map(function (p) {
+        return path.resolve(process.cwd(), p)
+    });
+} else {
+    nodePaths = []
```
shinnn:
```diff
-    nodePaths = []
+    nodePaths = [];
```

cli.js
ingro: Cool! So I can get rid of all the env check stuff!

shinnn: I should have commented it earlier :)

novemberborn:
LGTM!

ingro:
Thanks @novemberborn for your patience, I hope that if I'll ever have to contribuite again it will be a smoother process, I learned a lot anyway :)

novemberborn:
> Thanks @novemberborn for your patience, I hope that if I'll ever have to contribuite again it will be a smoother process, I learned a lot anyway :)

No worries, looking forward to your next contribution!

cli.js
```diff
@@ -93,11 +94,21 @@ if (cli.flags.init) {
 	return;
 }
 
+var nodePaths;
+if (process.env.NODE_PATH) {
+    nodePaths = process.env.NODE_PATH.split(path.delimiter).map(function (p) {
+        return path.resolve(process.cwd(), p);
```
novemberborn: Just realized that `path.resolve(p)` defaults to `path.resolve(process.cwd(), p)` if `p` is relative. In other words you don't need to provide `process.cwd()` here :boom:

ingro: There's always to learn something!

lib/test-worker.js
```diff
@@ -86,6 +86,9 @@ var oldNodeModulesPaths = module.constructor._nodeModulePaths;
 module.constructor._nodeModulePaths = function () {
 	var ret = oldNodeModulesPaths.apply(this, arguments);
 	ret.push(nodeModulesDir);
+	if (opts.nodePaths && opts.nodePaths.length > 0) {
```
sindresorhus: `opts.nodePaths.length > 0` is not needed as it's handled by `concat`

test/cli.js
```diff
@@ -7,7 +7,7 @@ var getStream = require('get-stream');
 var arrify = require('arrify');
 var cliPath = path.join(__dirname, '../cli.js');
 
-function execCli(args, dirname, cb) {
+function execCli(args, dirname, env, cb) {
```
sindresorhus: Just replace `dirname` and `env` with an options object.

test/cli.js
```diff
@@ -110,3 +115,12 @@ test('pkg-conf: cli takes precedence', function (t) {
 		t.end();
 	});
 });
+
+test('handles NODE_PATH', function (t) {
+	var nodePaths = 'node-paths/modules' + path.delimiter + 'node-paths/deep/nested';
```
sindresorhus: Use `path.join()` instead

test/fixture/node-paths/deep/nested/path/bar.js
```diff
@@ -0,0 +1,4 @@
+module.exports = function() {
```
sindresorhus: `module.exports = function () {`

test/fixture/node-paths/modules/nested/foo.js
```diff
@@ -0,0 +1,4 @@
+module.exports = function() {
```
sindresorhus: `module.exports = function () {`

cli.js
```diff
@@ -93,11 +94,21 @@ if (cli.flags.init) {
...
 }
 
+var nodePaths;
+if (process.env.NODE_PATH) {
+    nodePaths = process.env.NODE_PATH.split(path.delimiter).map(function (p) {
+        return path.resolve(p);
+    });
+} else {
+    nodePaths = [];
+}
```
sindresorhus:
```js
var nodePaths = process.env.NODE_PATH ? process.env.NODE_PATH
	.split(path.delimiter)
	.map(function (p) {
        return path.resolve(p);
    }) : [];
```

sindresorhus:
Ugh, I really dread this change. Not the PR itself, but the fact of having to support `NODE_PATH`, which is deprecated and awful in so many ways...

Wouldn't it be better to just pass all environment variables from the parent to the worker? That way we wouldn't have to special-case future environment variable needs.

test/cli.js
novemberborn: @sindresorhus do you mean `path.join('node-paths', 'modules')`? `path.delimiter` is used for building the environment variable itself, which comprises multiple paths.

novemberborn:
> Wouldn't it be better to just pass all environment variables from the parent to the worker? That way we wouldn't have to special-case future environment variable needs.

The problem is that `NODE_PATH` can be relative and the tests run in a different working directory.

test/cli.js
sindresorhus: Oh, nvm, I read it as `path.sep`...

sindresorhus:
```js
if (process.env.NODE_PATH) {
	process.env.NODE_PATH = process.env.NODE_PATH
		.split(path.delimiter)
		.map(function (p) {
			return path.resolve(p);
		})
		.join(path.delimiter);
}
```

Then?

test/cli.js
novemberborn: Node probably handles any platform issues with the slashes in `node-paths/modules`.

novemberborn:
In `cli.js`? Anywhere else and it may interfere with the programatic API.

sindresorhus:
Yes, or even better, just pass this directly to the `env` option when forking the worker, without overwriting the current `process.env.NODE_PATH`.

ingro:
Well I will wait to make further changes so, while a better solution is found!

novemberborn:
@ingro I think what @sindresorhus is saying is that you can [add an `env` object in `fork.js`](https://github.com/sindresorhus/ava/blob/9c22ee4ac4384352f4c30ba7de56eac919f435ff/lib/fork.js#L21) which sets `NODE_PATH` to the corrected value. Then you don't need the changes in `test-worker.js`.

sindresorhus:
Yup, and keep in mind that the whole existing `process.env` needs to be passed to the fork thing to preserve existing behavior, so use `object-assign` to copy the object and then modify the `NODE_PATH` key.

ingro:
Well I think that's cleaner, I will look at it as soon as I have time!

ingro:
The commit has some conflicts, what's the procedure in situation like this?

sindresorhus:
@ingro You [resolve the merge conflict](https://www.google.co.th/search?q=git+resolve+merge+conflict&oq=git+resolve+merge+&aqs=chrome.0.0j69i57j0l4.5662j0j7&sourceid=chrome&es_sm=91&ie=UTF-8). (No need to open a new PR)

ingro:
It seems we are back to green!

lib/fork.js
```diff
@@ -16,9 +16,25 @@ module.exports = function (file, opts) {
...
 
+	var env = process.env;
+
+	if (process.env.NODE_PATH) {
+		var nodePath = process.env.NODE_PATH
+			.split(path.delimiter)
+			.map(function (p) {
+				return path.resolve(p);
+			})
+			.join(path.delimiter);
```
novemberborn: Could compute this outside of the function so it can be reused for subsequent forks.

lib/fork.js
```diff
@@ -7,6 +7,21 @@ var debug = require('debug')('ava');
...
+	var nodePath = process.env.NODE_PATH
+		.split(path.delimiter)
+		.map(function (p) {
+			return path.resolve(p);
+		})
+		.join(path.delimiter);
+
+	env = objectAssign({
+		NODE_PATH: nodePath
+	}, env);
```
sindresorhus: Ugh, discovered this too late. This makes no sense. Here you overwrite the modified `NODE_PATH` with the original one. This means the tests are also invalid. Tests still pass when the `.map` is commented out.

// @ingro @novemberborn

ingro:
Welp sorry, I blindly copied from one of your comment above.
That should fix the wrong assignment:
`env = objectAssign(env, { NODE_PATH: nodePath });`

Anyway tests still fails, will look into that!

ingro:
I think I found the problem with the test, it was:

```javascript
var nodePaths = 'node-paths/modules' + path.delimiter + 'node-paths/deep/nested';
```

while it should have been:

```javascript
var nodePaths = 'fixture/node-paths/modules' + path.delimiter + 'fixture/node-paths/deep/nested';
```

That's because `NODE_PATH` should be relative to the directory from which `ava` is running, not to the test itself.

I've fixed that in my fork, I should create another pull request?

novemberborn:
> I should create another pull request?

Yep! :dancers:

ingro:
Created here https://github.com/sindresorhus/ava/pull/559

hildjj:
@sindresorhus I think this regressed with commit 1f10a7cfa1b2ccc9378143c08f79727e725c5550.

novemberborn:
@hildjj that was intentional.

hildjj:
Is this the correct place to discuss whether that was the correct decision?  I don't see an issue or pull request associated with that commit, so I'll assume yes, unless you'd like me to open a new issue.

I have an (admittedly crappy) use case where being able to manipulate NODE_PATH is very handy.  It involves using the same test suite to test two different builds of the same software that have subtle differences in their dependencies.  I can figure out another way around this if I have to, but it was convenient.

novemberborn:
You could use an `ava.config.cjs` file to rewrite the `NODE_PATH` and configure it as an environment variable for AVA to use for the worker processes. We're not going to bring back specialized handling in AVA itself.

hildjj:
I'm arguing that it's not specialized handling.  ava is now modifying the environment being passed to my tests in a surprising way.  At the very least I would have expected something in the release notes and something in the docs for a change that breaks existing functionality built in to node.  The [docs](https://github.com/avajs/ava/blob/main/docs/06-configuration.md) currently say:

"environmentVariables: specifies environment variables to be made available to the tests. The environment variables defined here override the ones from process.env"

which is at least confusing at the moment, since it's not clear that the `process.env` isn't the same as the environment passed in to ava.

`ava.config.cjs` might work for me, with some logic to detect which of the cases I'm in, I suppose.

novemberborn:
No, AVA 3 no longer modifies anything. But let's continue this in a [Discussion](https://github.com/avajs/ava/discussions).

hildjj:
https://github.com/avajs/ava/discussions/2682 for linkage

PR #536, opened 2016-02-09
Move `files` argument to `api.run()`

### LINKED ISSUE #534: Consider moving files patterns from API constructor to run() method
#502 allows files patterns to be provided to `API#run()`. This will override the patterns provided in the API constructor.

For simplicity we could consider removing the `files` argument from the constructor and requiring it in `run()`. This will break the programmatic API.

Currently the watcher provides the `files` argument to `API#run()`. If provided the test titles will always be prefixed, even if only a single test is run. This behavior makes sense for the watcher but is different from normal usage. We should maintain this behavior.

sotojuan:
This fixes #534. Of course, it requires everyone to be on board with the idea of moving `files` to `api.run()` instead of the constructor. In case everyone agrees, then this should help. Let me know if I missed anything.

### DIFF
--- a/api.js
+++ b/api.js
@@ -56,7 +56,7 @@ Api.prototype._reset = function () {
 	this.stats = [];
 	this.tests = [];
 	this.base = '';
-	this.explicitTitles = false;
+	this.options.explicitTitles = false;
 };
 
 Api.prototype._runFile = function (file) {
@@ -130,7 +130,7 @@ Api.prototype._handleTest = function (test) {
 };
 
 Api.prototype._prefixTitle = function (file) {
-	if (this.fileCount === 1 && !this.explicitTitles) {
+	if (this.fileCount === 1 && !this.options.explicitTitles) {
 		return '';
 	}
 
@@ -172,7 +172,7 @@ Api.prototype.run = function (files) {
 	];
 
 	this._reset();
-	this.explicitTitles = Boolean(this.files);
+	this.options.explicitTitles = Boolean(files);
 	return handlePaths(this.files, this.excludePatterns)
 		.map(function (file) {
 			return path.resolve(file);
--- a/api.js
+++ b/api.js
@@ -18,7 +18,7 @@ var fork = require('./lib/fork');
 var formatter = require('./lib/enhance-assert').formatter();
 var CachingPrecompiler = require('./lib/caching-precompiler');
 
-function Api(files, options) {
+function Api(options) {
 	if (!(this instanceof Api)) {
 		throw new TypeError('Class constructor Api cannot be invoked without \'new\'');
 	}
@@ -27,22 +27,6 @@ function Api(files, options) {
 
 	this.options = options || {};
 	this.options.require = (this.options.require || []).map(resolveCwd);
-
-	if (!files || files.length === 0) {
-		this.files = [
-			'test.js',
-			'test-*.js',
-			'test'
-		];
-	} else {
-		this.files = files;
-	}
-
-	this.excludePatterns = [
-		'!**/node_modules/**',
-		'!**/fixtures/**',
-		'!**/helpers/**'
-	];
 
 	Object.keys(Api.prototype).forEach(function (key) {
 		this[key] = this[key].bind(this);
@@ -165,9 +149,25 @@ Api.prototype._prefixTitle = function (file) {
 Api.prototype.run = function (files) {
 	var self = this;
 
+	if (!files || files.length === 0) {
+		this.files = [
+			'test.js',
+			'test-*.js',
+			'test'
+		];
+	} else {
+		this.files = files;
+	}
+
+	this.excludePatterns = [
+		'!**/node_modules/**',
+		'!**/fixtures/**',
+		'!**/helpers/**'
+	];
+
 	this._reset();
-	this.explicitTitles = Boolean(files);
-	return handlePaths(files || this.files, this.excludePatterns)
+	this.explicitTitles = Boolean(this.files);
+	return handlePaths(this.files, this.excludePatterns)
 		.map(function (file) {
 			return path.resolve(file);
 		})
--- a/api.js
+++ b/api.js
@@ -28,17 +28,17 @@ function Api(options) {
 	this.options = options || {};
 	this.options.require = (this.options.require || []).map(resolveCwd);
 
-	Object.keys(Api.prototype).forEach(function (key) {
-		this[key] = this[key].bind(this);
-	}, this);
-
-	this._reset();
-
 	this.excludePatterns = [
 		'!**/node_modules/**',
 		'!**/fixtures/**',
 		'!**/helpers/**'
 	];
+
+	Object.keys(Api.prototype).forEach(function (key) {
+		this[key] = this[key].bind(this);
+	}, this);
+
+	this._reset();
 }
 
 util.inherits(Api, EventEmitter);
@@ -56,7 +56,6 @@ Api.prototype._reset = function () {
 	this.stats = [];
 	this.tests = [];
 	this.base = '';
-	this.options.explicitTitles = false;
 };
 
 Api.prototype._runFile = function (file) {
@@ -155,25 +154,8 @@ Api.prototype._prefixTitle = function (file) {
 Api.prototype.run = function (files) {
 	var self = this;
 
-	if (!files || files.length === 0) {
-		this.files = [
-			'test.js',
-			'test-*.js',
-			'test'
-		];
-	} else {
-		this.files = files;
-	}
-
-	this.excludePatterns = [
-		'!**/node_modules/**',
-		'!**/fixtures/**',
-		'!**/helpers/**'
-	];
-
 	this._reset();
-	this.options.explicitTitles = Boolean(files);
-	return handlePaths(this.files, this.excludePatterns)
+	return handlePaths(files, this.excludePatterns)
 		.map(function (file) {
 			return path.resolve(file);
 		})
--- a/api.js
+++ b/api.js
@@ -33,6 +33,12 @@ function Api(options) {
 	}, this);
 
 	this._reset();
+
+	this.excludePatterns = [
+		'!**/node_modules/**',
+		'!**/fixtures/**',
+		'!**/helpers/**'
+	];
 }
 
 util.inherits(Api, EventEmitter);
--- a/test/api.js
+++ b/test/api.js
@@ -77,6 +77,37 @@ test('test title prefixes — multiple files', function (t) {
 	});
 });
 
+test('test title prefixes — single file', function (t) {
+	t.plan(2);
+
+	var separator = ' ' + figures.pointerSmall + ' ';
+	var files = [
+		path.join(__dirname, 'fixture/generators.js')
+	];
+	var expected = [
+		['generator function'].join(separator)
+	];
+	var index;
+
+	var api = new Api();
+
+	api.run(files)
+		.then(function () {
+			// if all lines were removed from expected output
+			// actual output matches expected output
+			t.is(expected.length, 0);
+		});
+
+	api.on('test', function (a) {
+		index = expected.indexOf(a.title);
+
+		t.true(index >= 0);
+
+		// remove line from expected output
+		expected.splice(index, 1);
+	});
+});
+
 test('test title prefixes — single file (explicit)', function (t) {
 	t.plan(2);
 
--- a/test/api.js
+++ b/test/api.js
@@ -292,8 +292,9 @@ test('uncaught exception will throw an error', function (t) {
 test('errors can occur without messages', function (t) {
 	t.plan(2);
 
-	var api = new Api([path.join(__dirname, 'fixture/error-without-message.js')]);
-	api.run()
+	var api = new Api();
+
+	api.run([path.join(__dirname, 'fixture/error-without-message.js')])
 		.then(function () {
 			t.is(api.failCount, 1);
 			t.is(api.errors.length, 1);
--- a/test/api.js
+++ b/test/api.js
@@ -9,9 +9,9 @@ var Api = require('../api');
 test('ES2015 support', function (t) {
 	t.plan(1);
 
-	var api = new Api([path.join(__dirname, 'fixture/es2015.js')]);
+	var api = new Api();
 
-	api.run()
+	api.run([path.join(__dirname, 'fixture/es2015.js')])
 		.then(function () {
 			t.is(api.passCount, 1);
 		});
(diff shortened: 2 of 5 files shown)


novemberborn:
Sadly this breaks `--watch`. To be fair there is no actual integration test from `cli.js` with `--watch` so you wouldn't have noticed. The test for `lib/watcher` stubs all its dependencies. I'll open a ticket for adding an integration test.

`files` and `excludePatterns` were assigned so `lib/watcher.js` could reference them. You'll have to pass them to `start()` now so the watcher can run the initial tests or specific tests. This probably means defining the defaults in `cli.js` or perhaps as an export from `api.js` for programatic convenience. They no longer need to be stored on the API instance. (Looks like it'd be valid to retain the `excludePatterns` definition in the API constructor though).

`explicitTitles` is there so the watcher can force a test title prefix even when running a single test. Your change makes it so test titles are always prefixed. If we're willing to go with that we can clean up the code. Otherwise `run()` should take a second `forcePrefixTestTitles` argument the watcher can use.

sotojuan:
Adding `api.files = files;` under [here](https://github.com/sindresorhus/ava/pull/536/files?diff=split#diff-2cce40143051e25f811b56c79d619bf5R122) makes `--watch` work—just tested it on a simple file.

Not sure if that's the best solution though.

novemberborn:
> Exporting files and excludePatterns could work but it'd require a big change in a lot of files (I think), so we'd have to get permission from the powers that be

Those patterns where [hidden away inside a function](https://github.com/sindresorhus/ava/blob/158916cc9d2afbb057f10cbf761475cb1477594f/api.js#L214:L224) before `--watch` landed so it should be OK. Real question is where they should be defined such that programmatic usage of the API is still possible even if the patterns are provided when instantiating the API or running the tests.

> Again, just exploring/messing around. It "works" but the tests for watcher still fail. Will look into it this night.

Yup that should do it. Just need to provide compatible values in https://github.com/sindresorhus/ava/blob/bdb8b6ebaa05f1dc31cdab7c84aed7d92317b8ed/test/watcher.js#L23 and <https://github.com/sindresorhus/ava/blob/bdb8b6ebaa05f1dc31cdab7c84aed7d92317b8ed/test/watcher.js#L95>.

sotojuan:
In `test/watcher.js`:

```js
files = [
  'test.js',
  'test-*.js',
  'test'
];
		
excludePatterns = [
  '!**/node_modules/**',
  '!**/fixtures/**',
  '!**/helpers/**'
];

...

var start = function (sources) {
  subject.start(logger, api, files, excludePatterns, sources || [], stdin);
};
```

Under the "Chokidar is installed tests" I get `not ok test unfinished` for any I try:

```
test/watcher.js ....................................... 2/4
  chokidar is installed > watches for default source file changes, as well as test files
  not ok test unfinished: watches for default source file changes, as well as test files
    at:
      line: 128
      column: 19
      file: test/watcher.js
      function: test
    source: |
      pending.push(_t.test(name, fn));

  not ok missing plan
    results:
      ok: false
      count: 1
      pass: 1
```

novemberborn:
@sotojuan hard to say. Could you push some code so I can debug it here?

sotojuan:
Done, let me know if you can help!

novemberborn:
@sotojuan just had to declare the variables and update more `api.files` and `api.excludePatterns` references: <https://github.com/novemberborn/ava/commit/aa4e211268bff3ed386ce5c4579c0624ff3c4358>.

novemberborn:
I think you should keep the default `excludePatterns` in the API constructor. It only needs to be defined once since it's used for every run. Then the watcher can simply access it like before.

Still need to resolve whether we always want to prefix test titles (even if there's only a single test file) or only when tests are rerun via the watcher. How about making it an option on the API itself, and if `--watch` is enabled set it to `true`? That way the watcher output is consistent but regular output is still minimal, *and* we won't need a second argument to `run()` in order to differentiate.

sotojuan:
> I think you should keep the default excludePatterns in the API constructor. It only needs to be defined once since it's used for every run. Then the watcher can simply access it like before.

Sure, that makes sense.

> Still need to resolve whether we always want to prefix test titles (even if there's only a single test file) or only when tests are rerun via the watcher. How about making it an option on the API itself, and if --watch is enabled set it to true? That way the watcher output is consistent but regular output is still minimal, and we won't need a second argument to run() in order to differentiate.

Also makes sense to me, doesn't seem that hard to implement. I'm interested in what the other mantainers think though. If they're on board with it, I'll work on it tonight.

// @sindresorhus @jamestalmage @vdemedes

sindresorhus:
:+1: Sounds good. I would still want minimal output in non-watch mode.

sotojuan:
All right, I think I got it but I feel like I am missing something. Let me know!

api.js
```diff
@@ -28,27 +28,17 @@ function Api(files, options) {
...
-			'test'
-		];
-	} else {
-		this.files = files;
-	}
+	Object.keys(Api.prototype).forEach(function (key) {
+		this[key] = this[key].bind(this);
+	}, this);
+
+	this._reset();
```
novemberborn: If you look at the overall diff you'll see that these lines were moved above `this.excludePatterns`. Would be nice to leave them undisturbed.

api.js
```diff
@@ -165,9 +155,25 @@ Api.prototype._prefixTitle = function (file) {
...
+		this.files = [
+			'test.js',
+			'test-*.js',
+			'test'
+		];
+	} else {
+		this.files = files;
+	}
+
+	this.excludePatterns = [
```
novemberborn: Not needed since this is already set in the constructor.

api.js
```diff
@@ -165,9 +155,25 @@ Api.prototype._prefixTitle = function (file) {
 Api.prototype.run = function (files) {
 	var self = this;
 
+	if (!files || files.length === 0) {
+		this.files = [
```
novemberborn: Not needed since `cli.js` takes care of setting defaults. The watcher also receives those defaults, which is what it should provide when calling `run()`.

api.js
```diff
@@ -165,9 +155,25 @@ Api.prototype._prefixTitle = function (file) {
...
+		'!**/node_modules/**',
+		'!**/fixtures/**',
+		'!**/helpers/**'
+	];
+
 	this._reset();
-	this.explicitTitles = Boolean(files);
-	return handlePaths(files || this.files, this.excludePatterns)
+	this.options.explicitTitles = Boolean(files);
+	return handlePaths(this.files, this.excludePatterns)
```
novemberborn: Since `files` is provided to the watcher you no longer need to assign it to `this.files`. Just use the variable.

api.js
```diff
@@ -165,9 +155,25 @@ Api.prototype._prefixTitle = function (file) {
...
+	this.excludePatterns = [
+		'!**/node_modules/**',
+		'!**/fixtures/**',
+		'!**/helpers/**'
+	];
+
 	this._reset();
-	this.explicitTitles = Boolean(files);
-	return handlePaths(files || this.files, this.excludePatterns)
+	this.options.explicitTitles = Boolean(files);
```
novemberborn: The code shouldn't change options. `cli.js` already takes care of enabling `explicitTitles` when in watch mode.

novemberborn:
> All right, I think I got it but I feel like I am missing something. Let me know!

Getting close! The code is now doing a little bit of the old behavior, and a little bit of the new behavior. 

`cli.js` should control the default `files` patterns. `Api#run()` should assume it's called with patterns and not force defaults. That means the watcher should [call `Api#run()`](https://github.com/sindresorhus/ava/blob/7d108aa0518982152ee6ae49c13b42e91146ebe9/lib/watcher.js#L198) with the `files` argument it received from `cli.js`.

`explicitTitles` is either on or off, it doesn't need to change based on the arguments to `Api#run()`. It's configured correctly by `cli.js` based on the `--watch` flag.

api.js
sotojuan: So basically take out the whole `if` block?

sotojuan:
Okay, I'll give these a shot throughout the day. One thing I am confused about though, look at this test. Supposedly it's testing for prefixes, but it's a single file, so it fails with the changes since by default there's no explicit titles for just one file. Should I update the test so it sets the `explicitTitles` option? Or is it a problem with my changes?

novemberborn:
> Should I update the test so it sets the explicitTitles option?

Yes.

> It's the same case with a couple of the tests in api.js. I added explicitTitles: true to those that were failing and the changes you mentioned above and all tests are passing, but I'd like to know before pushing!

Push away! :)

sotojuan:
Builds and tests pass fine in my Macbook, not sure what's up with AppVeyor.

test/watcher.js
```diff
@@ -20,7 +20,7 @@ test('chokidar is not installed', function (t) {
 	});
 
 	try {
-		subject.start({}, {files: [], excludePatterns: []}, []);
+		subject.start({}, {}, [], []);
```
novemberborn: Would be nice to keep `{exludePatterns: []}` here, even though the test doesn't fail without it.

test/watcher.js
```diff
@@ -74,16 +75,11 @@ test('chokidar is installed', function (_t) {
...
+		files = [
 			'test.js',
 			'test-*.js',
 			'test'
 		];
-		api.excludePatterns = [
-			'!**/node_modules/**',
-			'!**/fixtures/**',
-			'!**/helpers/**'
-		];
```
novemberborn: I suppose technically this doesn't have to be restored before each test, but it won't do any harm either. If you do want to move it you should add it to the `api` declaration <https://github.com/sotojuan/ava/blob/1d6d84442b99407ae0c329a48c75125ad6a1e4f9/test/watcher.js#L42:L44>

test/watcher.js
```diff
@@ -207,8 +209,8 @@ test('chokidar is installed', function (_t) {
 
 	[
 		{label: 'is added', fire: add},
-		{label: 'changes', fire: change},
-		{label: 'is removed', fire: unlink}
+		{label: 'changes', fire: change, event: 'change'},
+		{label: 'is removed', fire: unlink, event: 'unlink'}
```
novemberborn: Adding `event here is unnecessary.

test/api.js
```diff
@@ -6,38 +6,38 @@ var fs = require('fs');
 var test = require('tap').test;
 var Api = require('../api');
 
-test('ES2015 support', function (t) {
-	t.plan(1);
+// test('ES2015 support', function (t) {
```
novemberborn: Why did you comment these out? They pass on my machine.

test/api.js
```diff
@@ -58,40 +58,9 @@ test('test title prefixes — multiple files', function (t) {
...
-		index = expected.indexOf(a.title);
-
-		t.true(index >= 0);
-
-		// remove line from expected output
-		expected.splice(index, 1);
-	});
-});
-
-test('test title prefixes — single file', function (t) {
```
novemberborn: This test should be preserved. It makes sure test titles aren't prefixed when testing a single file.

The `test title prefixes — single file (explicit)` will cover behavior of the `explicitTitles` option.

test/api.js
```diff
@@ -180,8 +151,9 @@ test('display filename prefixes for failed test stack traces in subdirs', functi
 test('fail-fast mode', function (t) {
 	t.plan(5);
 
-	var api = new Api([path.join(__dirname, 'fixture/fail-fast.js')], {
-		failFast: true
+	var api = new Api({
+		failFast: true,
+		explicitTitles: true
```
novemberborn: No need to set this option in this test, it's meant to test `failFast`.

novemberborn:
@sotojuan The watcher tests run assertions on a stubbed API. They pass even if the watcher is broken. https://github.com/sindresorhus/ava/commit/b8c444de57ebd97ec5aaa679f4d2fc60f5aa15e0 adds an integration test to `test/cli.js` which fails with your changes so far. I should have pointed it out when that commit landed, it's not currently in this branch.

https://github.com/novemberborn/ava/commit/651343f9d864b45aae187bad0958a908546a23f8 solves the problem.

> Builds and tests pass fine in my Macbook, not sure what's up with AppVeyor.

There's a merge conflict with `master` :(

test/api.js
sotojuan: Just me forgetting to uncomment before pushing 😑

test/api.js
sotojuan: Actually the test fails without `explicitTitles` set to `true`. The prefixes the test tests for don't show up because it's a single file.

sotojuan:
OK, well for now I fixed all the small stuff. Like I mentioned earlier, the test for `failFast` does need `explicitTitles` to be set to `true` as the output it tests for has prefixes.

I'm not very experienced with conflicts so any instructions on how to deal with that would be great!

novemberborn:
> Like I mentioned earlier, the test for failFast does need explicitTitles to be set to true as the output it tests for has prefixes.

Yes but if you check the full diff you'll notice you *added* those prefixes in this PR ;-)

test/watcher.js
```diff
@@ -91,8 +92,14 @@ test('chokidar is installed', function (_t) {
 		done();
 	});
 
+	api.excludePatterns = [
+		'!**/node_modules/**',
+		'!**/fixtures/**',
+		'!**/helpers/**'
+	];
```
novemberborn: Need to remove this now.

novemberborn:
> I'm not very experienced with conflicts so any instructions on how to deal with that would be great!

https://help.github.com/articles/resolving-a-merge-conflict-from-the-command-line/ may help.

This will pull in the new watcher test in `test/cli.js`, so don't forget about https://github.com/novemberborn/ava/commit/651343f9d864b45aae187bad0958a908546a23f8 ;-)

sotojuan:
> Yes but if you check the full diff you'll notice you added those prefixes in this PR ;-)

Haha I totally forgot I did that! It's been a while!

sotojuan:
@novemberborn When you are not busy, can you review? Thanks!

novemberborn:
> When you are not busy, can you review? Thanks!

Oh didn't notice you fixed the merge conflict. Next time ping me more explicitly when you've updated the PR :smile: 

The watcher test in `test/cli.js` is failing. The solution is in https://github.com/novemberborn/ava/commit/651343f9d864b45aae187bad0958a908546a23f8 (minus the `console.error` call I left in by mistake).

novemberborn:
LGTM otherwise.

sotojuan:
Thanks, um should I just copy what's in that commit you linked me to or is that something that will be merged?

novemberborn:
@sotojuan copy yea. It only lives within my fork :)

sotojuan:
The watcher tests pass, for but some reason an API test is failing (even though it passes on my code!).

novemberborn:
@sotojuan are you running *all* tests?

sotojuan:
I had some stuff I hadn't merged with master yet, should be good now.

sotojuan:
Phew, all green on Travis (AppVeyor will take longer), finally!

novemberborn:
Great! :shipit: :exclamation: :tada:

sindresorhus:
LGTM

vdemedes:
LGTM!

novemberborn:
Thanks @sotojuan!

sindresorhus:
Thank you @sotojuan! Good work as usual.

Issue #516, opened 2016-02-06
Improve performance of simple sync tests with long stack traces enabled


sindresorhus:
We merged https://github.com/sindresorhus/ava/pull/495, which makes the `lib/sequence.js` code a lot more readable, but with the unfortunate side-effect of [slowing down execution](https://github.com/sindresorhus/ava/pull/495#issuecomment-177342400) of running thousands of simple synchronous tests. This might sound like an edge-case, and it kinda is, but the legitimate usage of something like this are generated tests, where you e.g. programmatically generate a test for each item in a huge list, like with https://github.com/beaugunderson/emoji-aware.

We should look into whether there's any way to improve on this without having to revert the code to be less readable, like before #495.

External suggestions more than welcome.

jamestalmage:
It should be noted that it really doesn't slow things down unless bluebird long stack traces are turned on.

In #495 we discussed merging it on the condition that we would turn off long stack traces, but now we are having second thoughts on that as well.

jamestalmage:
Here is the difference with longStackTraces still on in both:

<img width="1067" alt="screenshot 2016-02-06 02 54 07" src="https://cloud.githubusercontent.com/assets/4082216/12865547/3a50f062-cc7d-11e5-9760-da37adabe7d1.png">

jamestalmage:
Note that most of these benchmarks have 10,000 tests.

jamestalmage:
Turning shortStackTraces off  #495 is generally faster, but not by much.

vdemedes:
I will follow up with another PR in coming days regarding this issue.

novemberborn:
> I will follow up with another PR in coming days regarding this issue.

That's #539 right?

vdemedes:
Yep this is the one ;)

PR #539, opened 2016-02-09
Optimize Sequence by using Promise.each

vdemedes:
Previously, Sequence was checking for Promise-type result every time, to detect when a Promise needs to be returned from .run() method, which created a deep Promise chain. This behavior resulted in a big perf loss when AVA was used in big test suites.

After this PR, when Sequence finds Promise-type result it uses Promise.each on all the remaining tests, which should fix the issue with current implementation.

Follow up PR from #495.

### DIFF
--- a/lib/sequence.js
+++ b/lib/sequence.js
@@ -1,5 +1,8 @@
 'use strict';
 var isPromise = require('is-promise');
+var AvaError = require('./ava-error');
+
+function noop() {}
 
 module.exports = Sequence;
 
@@ -17,21 +20,17 @@ function Sequence(tests, bail) {
 	this.reason = null;
 	this.tests = tests;
 	this.bail = bail || false;
+
+	// TODO(vdemedes): separate into a utility (it's being used in serveral places)
+	Object.keys(Sequence.prototype).forEach(function (key) {
+		this[key] = this[key].bind(this);
+	}, this);
 }
 
 Sequence.prototype.run = function () {
-	return this._run();
-};
-
-// run sequence items starting from specified index
-Sequence.prototype._run = function (fromIndex) {
 	var length = this.tests.length;
 
-	if (!fromIndex) {
-		fromIndex = 0;
-	}
-
-	for (var i = fromIndex; i < length; i++) {
+	for (var i = 0; i < length; i++) {
 		// if last item failed and we should bail, return results and stop
 		if (this.bail && !this.passed) {
 			return this._results();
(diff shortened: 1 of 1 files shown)


lib/ava-error.js
```diff
@@ -9,4 +9,6 @@ function AvaError(message) {
 	this.name = 'AvaError';
 }
 
+AvaError.prototype = Object.create(Error.prototype);
```
vdemedes: Needed to make `.catch(err, handler)` work. Without this change, Bluebird did not detect `AvaError` class as a class inherited from `Error` and assumed it was `handler`.

sindresorhus: :+1: Should be done regardless. Previously it wasn't a real error.

sindresorhus:
:+1: Generally looks good to me, but would like @jamestalmage to take a look before merging.

Could you share some benchmark results?

vdemedes:
screen shot 2016-02-10 at 3 32 42 pm

lib/sequence.js
```diff
@@ -39,26 +38,41 @@ Sequence.prototype._run = function (fromIndex) {
 
 		var result = this.tests[i].run();
 
+		// if a Promise returned, we don't need to check for Promises after this test
+		// so we can just use Promise.each() on the rest of the tests
```
novemberborn: `is-promise` only checks for `then()` methods though, yet the promise chain below assumes `result` is actually a Bluebird promise. That's probably guaranteed given the rest of the code but feels like a refactoring hazard.

Why not always use `Promise.resolve(result)` and the chain below? Or is that too expensive performance wise? I couldn't immediately find a "is this a Bluebird promise" function in their docs, but maybe `result instanceof Promise` will work for us?

vdemedes: @jamestalmage insisted on not creating Promises when they're not needed, because it slows down test suites with thousands of tests. I think we need to bring this discussion up again, as the code looks like (excuse me) shit, when we need to return variable async/sync result from Sequence/Concurrent#run().

novemberborn: Similar issue in <https://github.com/sindresorhus/ava/pull/542#issuecomment-189892779>.

At the risk of splitting up the discussion too much, one thing I've done in the past is to return a wrapper result from mixed synchronicity functions, either with a `promise` property or a `value` property. Then you can do `if (result.promise)` without having to do further type checking.

novemberborn:
Ah @vdemedes this is for #516?

PR #541, opened 2016-02-10
Test watcher from the CLI

### LINKED ISSUE #537: test/cli needs to test --watch
The test for `lib/watcher.js` are fully stubbed so changes to the API may break the watcher without failing any tests. Need to add an integration test to `test/cli.js`.

The test should probably trigger a file change. We can look to nodemon for [inspiration](https://github.com/remy/nodemon/blob/master/test/monitor/run.test.js#L30:L44).

Since `chokidar` is an optional dependency the test may fail on platforms where it cannot be installed. This should cause the test to be skipped, not fail.

novemberborn:
The watcher tests are fully stubbed. This test verifies it actually works from the CLI.

Fixes #537.

### DIFF
--- a/test/cli.js
+++ b/test/cli.js
@@ -5,6 +5,7 @@ var test = require('tap').test;
 global.Promise = require('bluebird');
 var getStream = require('get-stream');
 var arrify = require('arrify');
+var touch = require('touch');
 var cliPath = path.join(__dirname, '../cli.js');
 
 function execCli(args, dirname, cb) {
@@ -21,11 +22,12 @@ function execCli(args, dirname, cb) {
 		env.AVA_APPVEYOR = 1;
 	}
 
+	var child;
 	var stdout;
 	var stderr;
 
 	var processPromise = new Promise(function (resolve) {
-		var child = childProcess.spawn(process.execPath, [path.relative(dirname, cliPath)].concat(arrify(args)), {
+		child = childProcess.spawn(process.execPath, [path.relative(dirname, cliPath)].concat(arrify(args)), {
 			cwd: dirname,
 			env: env,
 			stdio: [null, 'pipe', 'pipe']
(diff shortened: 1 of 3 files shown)


test/fixture/watcher/test.js
```diff
@@ -0,0 +1,5 @@
+import test from '../../../';
+
+test('works', t => {
+	t.ok(true);
```
sindresorhus: `t.pass();`?

test/cli.js
```diff
@@ -110,3 +114,45 @@ test('pkg-conf: cli takes precedence', function (t) {
...
+		hasChokidar = true;
+	} catch (err) {}
+
+	var child = execCli(['--verbose', '--watch', 'test.js'], 'fixture/watcher', function (err, stdout) {
+		if (err && err.code === 1 && !hasChokidar) {
+			t.comment('chokidar dependency is missing, cannot test watcher');
+			t.match(stdout, 'The optional dependency chokidar failed to install and is required for --watch. Chokidar is likely not supported on your platform.');
+			t.end();
+		} else {
+			t.comment('chokidar dependency is installed, cannot test watcher behavior with missing dependency');
```
sindresorhus: This message is a bit confusing and I don't see the point of it. There's already a warning when `chokidar` is not installed. I don't see the point of commenting that it is installed.

sindresorhus:
// @vdemedes

test/cli.js
novemberborn: The test passes either way so you can't obviously tell which branch passed, hence the comment output. Though that still requires you to look at raw TAP so probably not that useful.

sindresorhus: Yes, but I'm only interested if Chokidar is not installed. Otherwise IMHO it's just noise.

novemberborn:
Updated.

sindresorhus:
:+1:

Issue #504, opened 2016-02-02
test-collection needs cleanup


jamestalmage:
Coming back just a few weeks after writing it, I find [this section](https://github.com/sindresorhus/ava/blob/master/lib/test-collection.js#L136-L162) confusing.

Maybe it just needs comments, but I would prefer to refactor into a self explanatory API if possible.

novemberborn:
Looks like this might be covered by #542?

vdemedes:
Yep!

PR #542, opened 2016-02-10
Clean up test-collection

vdemedes:
This PR cleans up lib/test-collection.js and a part of my cleaning up set of PRs.

I still don't like it a lot, but it's a step forward.

### DIFF
--- a/index.js
+++ b/index.js
@@ -68,7 +68,7 @@ function exit() {
 }
 
 globals.setImmediate(function () {
-	var numberOfTests = runner.tests.concurrent.length + runner.tests.serial.length;
+	var numberOfTests = runner.tests.tests.concurrent.length + runner.tests.tests.serial.length;
 
 	if (numberOfTests === 0) {
 		send('no-tests', {avaRequired: true});
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -96,7 +96,9 @@ Runner.prototype.run = function () {
 		testCount: 0
 	};
 
-	return Promise.resolve(this.tests.buildPhases(this._addTestResult, this._bail).run()).then(function () {
+	this.tests.on('test', this._addTestResult);
+
+	return Promise.resolve(this.tests.build(this._bail).run()).then(function () {
 		stats.passCount = stats.testCount - stats.failCount - stats.skipCount;
 	});
 };
--- a/lib/test-collection.js
+++ b/lib/test-collection.js
@@ -1,4 +1,11 @@
 'use strict';
+var EventEmitter = require('events').EventEmitter;
+var util = require('util');
+var fnName = require('fn-name');
+var Concurrent = require('./concurrent');
+var Sequence = require('./sequence');
+var Test = require('./test');
+
 module.exports = TestCollection;
 
 function TestCollection() {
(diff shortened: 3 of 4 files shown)


index.js
```diff
@@ -68,7 +68,7 @@ function exit() {
 }
 
 globals.setImmediate(function () {
-	var numberOfTests = runner.tests.concurrent.length + runner.tests.serial.length;
+	var numberOfTests = runner.tests.tests.concurrent.length + runner.tests.tests.serial.length;
```
sindresorhus: `tests.tests` looks kinda weird. Maybe the test collection should have a `.size` getter property (like a `Map`) or something?

sindresorhus:
Generally looks good, but would like @jamestalmage to review here too, as I'm not that familiar with this part of the code.

Maybe @novemberborn would be interested in reviewing too.

sindresorhus:
Does this have any perf side-effects? Maybe worth running the benchmark?

vdemedes:
Sure, will do and post the results.

index.js
novemberborn: Could use `.concurrent` and `.serial` in the collection as well, not sure it needs the `tests` container.

lib/runner.js
```diff
@@ -96,7 +96,9 @@ Runner.prototype.run = function () {
 		testCount: 0
 	};
 
-	return Promise.resolve(this.tests.buildPhases(this._addTestResult, this._bail).run()).then(function () {
+	this.tests.on('test', this._addTestResult);
+
+	return Promise.resolve(this.tests.build(this._bail).run()).then(function () {
```
novemberborn: Perhaps change `Sequence#run()` to always return a promise. At the moment it *usually* returns a promise which seems odd.

lib/test-collection.js
```diff
@@ -6,158 +13,173 @@ function TestCollection() {
...
 	if (!type) {
-		throw new Error('test type must be specified');
+		throw new Error('Test type must be specified');
+	}
+
+	if (!test.title && test.fn) {
+		test.title = fnName(test.fn);
+	}
+
+	if (test.title === 'callee$0$0') {
```
novemberborn: Perhaps restore the `// workaround for Babel giving anonymous functions a name` comment. This looks like a magical string otherwise.

sindresorhus: :+1:

lib/test-collection.js
```diff
@@ -6,158 +13,173 @@ function TestCollection() {
...
-	return arr.concat(buildHooks(afterEach, entry.title, contextRef, report));
-}
+	var beforeHooks = this._buildHooks(this.hooks.beforeEach, test.title, context);
+	var afterHooks = this._buildHooks(this.hooks.afterEach, test.title, context);
+
+	var items = [];
+
+	items.push.apply(items, beforeHooks);
+	items.push(this._buildTest(test, context));
+	items.push.apply(items, afterHooks);
```
novemberborn: How about:

```js
return [].concat(beforeHooks, this._buildTest(test, context), afterHooks)
```

index.js
vdemedes: @sindresorhus Yeah, that's what I was thinking too.

lib/runner.js
vdemedes: `Sequence#run()` and `Concurrent#run()` behavior is on purpose, @jamestalmage says it affects performance to create promises when they are not required.

lib/test-collection.js
vdemedes: Agree!

lib/test-collection.js
vdemedes: Looks way better!

lib/runner.js
novemberborn: I doubt that'll be noticeable given the cost of spawning a new process in order to run the tests.

vdemedes: Agree.

vdemedes:
PR updated, ping!

lib/test-collection.js
```diff
@@ -6,158 +13,168 @@ function TestCollection() {
...
+	}
 
-		if (metadata.exclusive && !this.hasExclusive) {
-			this.hasExclusive = true;
-			this.serial = [];
-			this.concurrent = [];
+	// a hook
+	if (type !== 'test') {
+		if (metadata.exclusive) {
+			throw new Error('"only" can not be used with a ' + type + ' test');
```
novemberborn: Change `test` to `hook`? Perhaps `cannot` over `can not`?

`"only" cannot be used with a before hook` etc.

test/test-collection.js
```diff
@@ -24,25 +24,56 @@ function mockTest(opts, title) {
 }
 
 function titles(tests) {
+	if (!tests) {
+		tests = [];
```
novemberborn: Could do `return (tests || []).map()` below.

test/test-collection.js
```diff
@@ -24,25 +24,56 @@ function mockTest(opts, title) {
...
+function removeEmptyProps(obj) {
+	if (Array.isArray(obj) && obj.length === 0) {
+		return null;
+	}
+
+	if (obj.constructor !== Object) {
+		return obj;
+	}
+
+	var cleanObj;
```
novemberborn: Should initialize to `null` so `removeEmptyProps({})` is consistent with `removeEmptyProps([])`.

novemberborn:
>> Does this have any perf side-effects? Maybe worth running the benchmark?

> Sure, will do and post the results.

Curious about those results now :) Also with regards to `Sequence#run()` not consistently returning a promise, and @vdemedes' comment in <https://github.com/sindresorhus/ava/pull/539/files#r54347365>. I'd much rather have consistent code than checking for non-promises and having a/synchronous branches.

Any thoughts on going back to `runner.tests.concurrent` and `runner.tests.serial`?

test/test-collection.js
vdemedes: This looks a bit ugly, if condition is cleaner.

novemberborn: If only we could use parameter defaults :wink:

novemberborn:
@jamestalmage once this lands should #504 be closed?

lib/test-collection.js
```diff
@@ -6,158 +13,168 @@ function TestCollection() {
...
 		}
 
-		(metadata.serial ? this.serial : this.concurrent).push(test);
+		this.hooks[type].push(test);
+		return;
+	}
 
+	// a test
```
jamestalmage: What does this comment mean?

lib/test-collection.js
jamestalmage: We certainly seem to prefer informal contractions (i.e. `can't`) in the documentation. Not sure if that should carry over to error messages.

It seems `cannot` is more widely accepted - so, at a minimum, use that.

test/test-collection.js
```diff
@@ -24,25 +24,56 @@ function mockTest(opts, title) {
...
 		return test.title;
 	});
 }
 
+function removeEmptyProps(obj) {
+	if (Array.isArray(obj) && obj.length === 0) {
+		return null;
+	}
+
+	if (obj.constructor !== Object) {
```
jamestalmage: Why do we care what the constructor is? Are we just trying to avoid cleaning primitives?

jamestalmage:
screen shot 2016-02-29 at 4 49 17 pm

lib/runner.js
jamestalmage: > I doubt that'll be noticeable given the cost of spawning a new process in order to run the tests.

It's [definitely noticeable](https://github.com/sindresorhus/ava/pull/466#issuecomment-174238246) in real test suites.

IMO, we need to stop comparing to "the cost of spawning a new process" as a reason to disqualify optimizations. That is a huge cost, nearly every optimization is going to be small in comparison at this point. There is a large subset of test suites for which AVA is demonstrably slower than the alternatives, and we are not going to close that gap without placing a high priority on optimizations.

sindresorhus:
I honestly think it's a negligible perf regression. We should optimize for what matters instead. Whether we have a 1ms or 2ms overhead per test doesn't matter much when the test itself takes 100ms to execute. I think we need a more realistic benchmark so we don't end up bikeshedding over numbers that doesn't really give us any perceived perf improvements for the common-case.

sindresorhus:
@vdemedes Can you fixup the inline feedback so this can be merged?

lib/test-collection.js
vdemedes: We don't have a `hook` type, but `beforeEach`, `afterEach`, `before` and `after`. So it's quicker to check for `type !== 'test'` than checking for all four of these.

test/test-collection.js
vdemedes: Yes.

vdemedes:
@sindresorhus All done, good to go ;)

PR #544, opened 2016-02-10
Only run tests for changed sources

### LINKED ISSUE #115: Only run tests for changed code
Many times you only change code that affects one or a few tests, but still have to suffer through the whole test suite. Would be nice to only run the affected tests instead. This would be a huge win for people with lots of tests.

This is easy on a test file level as we can just recurse the dependency tree and see which test files depend on the changed file. It gets harder for individual tests. Here we could take advantage of code coverage, diff what's changed between runs, and check which lines were affected.

###### Prior art

- http://dm.gl/2015/01/30/wallaby/
- https://github.com/Genki-S/ttnt
- http://cherrie.io

-

This is not something that is likely to come in the short term, but opening for discussion and brainstorming. Help making this happen more than welcome.

novemberborn:
Fixes #115, building on top of the `--watch` implementation.

Test workers keep track of test dependencies for all registered file extensions. These are sent back to the Api as part of the 'teardown' event, ensuring they're captured irrespective of whether tests passed.

The watcher rejects any dependencies that are not source files, matching how Chokidar watches for source file modifications. It maintains a list of source dependencies for each test file.

The watcher will find the test files that depend on modified source files and rerun those (along with any other modified test files). If any modified source files cannot be mapped to test files all tests are rerun. This is necessary because only `require()` dependencies are tracked, not file access.

Run `DEBUG=ava:watcher $(npm bin)/ava --watch` in a project to see the matching in action.

This required some extensive refactoring of the watcher code itself. Worth it in the end since the changes drop in quite nicely. This will conflict heavily with #536 so best if that lands first.

Additionally I found that test workers may receive multiple `teardown` events. This made the `emits dependencies for test files` Api test flaky since it would occur for uncaught exceptions. That's fixed in a separate commit.

### DIFF
--- a/lib/test-worker.js
+++ b/lib/test-worker.js
@@ -122,11 +122,19 @@ process.on('ava-exit', function () {
 	}, delay);
 });
 
+var tearingDown = false;
 process.on('ava-teardown', function () {
+	// ava-teardown can be sent more than once.
+	if (tearingDown) {
+		return;
+	}
+	tearingDown = true;
+
 	var rejections = loudRejection.currentlyUnhandled();
 
 	if (rejections.length === 0) {
-		return exit();
+		exit();
+		return;
 	}
 
 	rejections = rejections.map(function (rejection) {
--- a/lib/test-worker.js
+++ b/lib/test-worker.js
@@ -89,6 +89,17 @@ module.constructor._nodeModulePaths = function () {
 	return ret;
 };
 
+var dependencies = [];
+Object.keys(require.extensions).forEach(function (ext) {
+	var wrappedHandler = require.extensions[ext];
+	require.extensions[ext] = function (module, filename) {
+		if (filename !== testPath) {
+			dependencies.push(filename);
+		}
+		wrappedHandler(module, filename);
+	};
+});
+
 require(testPath);
 
 process.on('uncaughtException', function (exception) {
@@ -150,5 +161,8 @@ process.on('ava-teardown', function () {
 });
 
 function exit() {
-	send('teardown');
+	// Include dependencies in the final teardown message. This ensures the full
+	// set of dependencies is included no matter how the process exits, unless
+	// it flat out crashes.
+	send('teardown', {dependencies: dependencies});
 }
--- a/lib/watcher.js
+++ b/lib/watcher.js
@@ -2,6 +2,10 @@
 
 var AvaError = require('./ava-error');
 var debug = require('debug')('ava:watcher');
+var diff = require('arr-diff');
+var flatten = require('arr-flatten');
+var union = require('array-union');
+var uniq = require('array-uniq');
 var defaultIgnore = require('ignore-by-default').directories();
 var multimatch = require('multimatch');
 var nodePath = require('path');
@@ -33,6 +37,9 @@ function Watcher(logger, api, files, sources) {
 		}, rethrowAsync);
 	};
 
+	this.testDependencies = [];
+	this.trackTestDependencies(api, sources);
+
 	this.dirtyStates = {};
 	this.watchFiles(files, sources);
 	this.rerunAll();
@@ -54,6 +61,47 @@ Watcher.prototype.watchFiles = function (files, sources) {
 			self.debouncer.debounce();
 		}
 	});
+};
+
+Watcher.prototype.trackTestDependencies = function (api, sources) {
+	var isSource = makeSourceMatcher(sources);
+
+	var cwd = process.cwd();
+	var relative = function (absPath) {
+		return nodePath.relative(cwd, absPath);
+	};
+
+	var self = this;
+	api.on('dependencies', function (file, dependencies) {
+		var sourceDeps = dependencies.map(relative).filter(isSource);
+		self.updateTestDependencies(file, sourceDeps);
+	});
+};
+
+Watcher.prototype.removeUnlinkedTestDependencies = function (unlinkedTests) {
+	unlinkedTests.forEach(function (testFile) {
+		this.updateTestDependencies(testFile, []);
+	}, this);
+};
+
+Watcher.prototype.updateTestDependencies = function (file, sources) {
+	if (sources.length === 0) {
+		this.testDependencies = this.testDependencies.filter(function (dep) {
+			return dep.file !== file;
+		});
+		return;
+	}
+
+	var isUpdate = this.testDependencies.some(function (dep) {
+		if (dep.file === file) {
+			dep.sources = sources;
+			return true;
+		}
+	});
+
+	if (!isUpdate) {
+		this.testDependencies.push(new TestDependency(file, sources));
+	}
 };
 
 Watcher.prototype.observeStdin = function (stdin) {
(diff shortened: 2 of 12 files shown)


novemberborn:
@sindresorhus @vdemedes @jamestalmage any feedback?

sindresorhus:
@novemberborn I'll try to take a look tomorrow. Been busy IRL and James is off traveling :)

vdemedes:
I'm a bit busy this weekend, I'll take a look during the start of the week. Thanks!

novemberborn:
@sindresorhus @vdemedes ping

lib/test-worker.js
```diff
@@ -89,6 +89,22 @@ module.constructor._nodeModulePaths = function () {
...
+// TODO(novemberborn) extract into its own package
+Object.keys(require.extensions).forEach(function (ext) {
+	var old = require.extensions[ext];
+	require.extensions[ext] = function (module, filename) {
+		if (filename !== testPath) {
+			// TODO(novemberborn) this is likely to yield a huge list of dependencies,
+			// slowing down IPC. Consider removing filenames that are excluded from
+			// being sources, and not tracking dependencies at all if not in --watch
+			// mode. Additionally consider sending these in a single message when the
+			// process exits (including in error conditions).
```
sindresorhus: I think it would be a better approach to send it all at once.

lib/watcher.js
```diff
@@ -183,21 +208,38 @@ function runAfterChanges(logger, api, isTest, dirtyStates) {
...
+		if (mustRunAll || dirtyTests.length !== dirtyPaths.length) {
 			resolve(api.run());
+		} else {
+			resolve(api.run(addedOrChangedTests));
 		}
 	}).then(function () {
 		logger.finish();
+
+		// TODO(novemberborn) update dependenciesByTest. May need some more
+		// refactoring to avoid passing this many arguments to runAfterChanges(),
```
sindresorhus: :+1:

lib/watcher.js
```diff
@@ -58,6 +62,10 @@ exports.start = function (logger, api, sources, stdin) {
 
 	var busy = api.run().then(function () {
 		logger.finish();
+
+		Object.keys(api.dependencies).forEach(function (file) {
+			dependenciesByTest[file] = api.dependencies[file];
+		});
```
sindresorhus: `object-assign`?

lib/watcher.js
```diff
@@ -172,8 +180,25 @@ function makeTestMatcher(files, excludePatterns) {
...
+	};
+}
+
+function runAfterChanges(logger, api, isTest, checkTestDependencies, dirtyStates) {
 	var dirtyPaths = Object.keys(dirtyStates);
+	// All tests must be run if 'rs' is entered on stdin, in which case there
+	// may not be any dirty paths.
+	// TODO(novemberborn) consider making this explicit rather than checking
+	// dirtyPaths.length
```
sindresorhus: Not sure what making it explicit would entail, but I always prefer explicit.

sindresorhus:
The approach generally looks good.

> This tracks dependencies per test.

By `test`, I assume you mean test file?

vdemedes:
LGTM from me, will clean up later myself ;)

novemberborn:
> By test, I assume you mean test file?

Yes. Naming is hard…

> LGTM from me, will clean up later myself ;)

Let me take the feedback and improve on this :)

vdemedes:
@novemberborn it's no big deal, few things I noticed:

1) I never use .reduce()
2) https://github.com/sindresorhus/ava/pull/544/files#diff-b72e3e25f31d6b3c3e5f40b5991dc5f6R94 raises question "old what?", var name needs to be explicit
3) I'd replace `arrayUnion` with `merge`.

Most of it and others are stylistic changes, so don't worry ;)

novemberborn:
Updated! This should be good to go now.

New PR description:

>Test workers keep track of test dependencies for all registered file extensions. These are sent back to the Api as part of the 'teardown' event, ensuring they're captured irrespective of whether tests passed.
>
>The watcher rejects any dependencies that are not source files, matching how Chokidar watches for source file modifications. It maintains a list of source dependencies for each test file.
>
>The watcher will find the test files that depend on modified source files and rerun those (along with any other modified test files). If any modified source files cannot be mapped to test files all tests are rerun. This is necessary because only `require()` dependencies are tracked, not file access.
>
>Run `DEBUG=ava:watcher $(npm bin)/ava --watch` in a project to see the matching in action.
>
>This required some extensive refactoring of the watcher code itself. Worth it in the end since the changes drop in quite nicely. This will conflict heavily with #536 so best if that lands first.
>
>Additionally I found that test workers may receive multiple `teardown` events. This made the `emits dependencies for test files` Api test flaky since it would occur for uncaught exceptions. That's fixed in a separate commit.

---

@vdemedes 

> 3) I'd replace arrayUnion with merge

But @sindresorhus wrote `array-union`! :wink: If you're referring to https://www.npmjs.com/package/merge that seems to merge objects, not arrays?

vdemedes:
I meant just to change the var name, not the module :) Never mind, it's totally my preferences, not forcing them.

novemberborn:
Pushed a commit to fix Windows. Paths are insanely confusing so I've tried to simplify a few assumptions:

* test files always use platform-specific slashes
* Chokidar emits file paths using platform-specific slashes
* tracked dependencies use platform-specific slashes
* test file and source patterns always use POSIX slashes
* on Windows, file paths are converted to use POSIX slashes before matching patterns

This way the actual calls to the various matcher libraries are the same no matter the platform and we don't have to make assumptions on how the library adjusts. Just for example with `minimatch` on El Capitan:

```
>  require('minimatch').match(['nested/file.js'],'nested/*.js')
[ 'nested/file.js' ]
>  require('minimatch').match(['nested\\file.js'],'nested/*.js')
[]
>  require('minimatch').match(['nested\\file.js'],'nested\\*.js')
[]
>  require('minimatch').match(['nested/file.js'],'nested\\*.js')
[]
```

On Windows 10:

```
> require('minimatch').match(['nested/file.js'],'nested/*.js')
[ 'nested/file.js' ]
> require('minimatch').match(['nested\\file.js'],'nested/*.js')
[ 'nested\\file.js' ]
> require('minimatch').match(['nested\\file.js'],'nested\\*.js')
[ 'nested\\file.js' ]
> require('minimatch').match(['nested/file.js'],'nested\\*.js')
[ 'nested/file.js' ]
```

:confounded:

novemberborn:
@sindresorhus @vdemedes @jamestalmage rebased on `master` now.

vdemedes:
Cool, interesting idea @novemberborn! I don't think we need to squash commits here, just merge as it is.

jamestalmage:
In general LGTM.

How is thorough is coverage of new code? It just seems like a high ratio of new code to tests (gut feeling after reviewing on mobile - I could be off base).

My only concern with the require hook is that it is easily broken by bad actors. Something like `babel-detective` might be more robust (but maybe wouldn't catch as thorough a dependency graph)

sindresorhus:
Alright. I've been testing this now in many different projects and seems to be working good.

sindresorhus:
> My only concern with the require hook is that it is easily broken by bad actors.

Is there anything we could do to prevent this?

> Something like babel-detective might be more robust (but maybe wouldn't catch as thorough a dependency graph)

I think an actual dependency graph is a lot better than what a static analysis could do.

sindresorhus:
Merging. We can follow-up with any improvements/requests. Really good stuff @novemberborn :)

unicorn22

jamestalmage:
> Is there anything we could do to prevent this?

Maybe instead of a conventional hook we wrap `Module.prototype.compile`? That is a no-no for conventional hooks (not really extendable), but since our hook doesn't manipulate the args (it just tracks files) I don't think it would be a problem

> I think an actual dependency graph is a lot better than what a static analysis could do. 

I agree.

novemberborn:
@jamestalmage,

> How is thorough is coverage of new code? It just seems like a high ratio of new code to tests (gut feeling after reviewing on mobile - I could be off base).

Probably looked that way due to the refactor I did. That had little test impact though.

> My only concern with the require hook is that it is easily broken by bad actors. Something like babel-detective might be more robust (but maybe wouldn't catch as thorough a dependency graph)

Yea it's icky. If it's any consolation the require hook is added after the custom requires have loaded, so unless hooks are added afterwards we'll be able to capture any dependencies. I'll make sure to explain this in the documentation.

`babel-detective` would let us capture the direct dependencies of the test file but not transitive dependencies. That's where this feature really shines: you have tests for `Foo` and `Bar` and `Baz`, the latter two depending on `thingybob`. When you change `thingybob` you don't need to retest `Foo`.

> Maybe instead of a conventional hook we wrap Module.prototype.compile? That is a no-no for conventional hooks (not really extendable), but since our hook doesn't manipulate the args (it just tracks files) I don't think it would be a problem

I don't think that would work for native and JSON dependencies?

> Merging. We can follow-up with any improvements/requests. Really good stuff @novemberborn :)

Thanks @sindresorhus!

sindresorhus:
> Maybe instead of a conventional hook we wrap Module.prototype.compile? That is a no-no for conventional hooks (not really extendable), but since our hook doesn't manipulate the args (it just tracks files) I don't think it would be a problem

I like this idea. Could even be a separate module we consume. @novemberborn hint hint ;)

-

@novemberborn Also feel free to extract anything else into modules where it makes sense.

sindresorhus:
> I don't think that would work for native and JSON dependencies?

Yup, that's true, since they don't call `._compile`, but those are also rarely overriden, so a combination of `._compile` + hooks for JSON/native could work.

novemberborn:
> Yup, that's true, since they don't call ._compile, but those are also rarely overriden, so a combination of ._compile + hooks for JSON/native could work.

Oh that sounds good. I considered extracting it initially but then figured it wasn't quite interesting enough. Combining it with `._compile` does make it more interesting. I've added it to my personal to-do list…

PR #547, opened 2016-02-11
Clean up concurrent

vdemedes:
Benchmark:

screen shot 2016-02-11 at 2 43 12 pm

### DIFF
--- a/lib/concurrent.js
+++ b/lib/concurrent.js
@@ -1,7 +1,9 @@
 'use strict';
 var Promise = require('bluebird');
 var isPromise = require('is-promise');
-var BAIL_ERROR = new Error();
+var AvaError = require('./ava-error');
+
+function noop() {}
 
 module.exports = Concurrent;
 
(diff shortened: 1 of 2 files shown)


lib/concurrent.js
```diff
@@ -10,86 +12,85 @@ function Concurrent(tests, bail) {
 		throw new TypeError('Class constructor Concurrent cannot be invoked without \'new\'');
 	}
 
+	if (!tests) {
+		throw new Error('Concurrent items can\'t be undefined');
```
sindresorhus:
```js
if (!Array.isArray(tests)) {
  throw new TypeError('Expected an array of tests');
}
```

vdemedes: Right, this is better.

lib/concurrent.js
```diff
@@ -10,86 +12,85 @@ function Concurrent(tests, bail) {
...
+	if (result.passed === false) {
+		this.passed = false;
+
+		// only set reason once
+		if (!this.reason) {
+			this.reason = result.reason;
+		}
+
+		if (this.bail) {
+			throw new AvaError('Error in Concurrent while in bail mode');
```
sindresorhus: This error message is unclear. What is the purpose of it?

vdemedes: This error will never be shown to a user, it's just for us. Did not want to leave AvaError without a message. It's for us, like a code comment, to indicate that this error causes execution to bail.

sindresorhus: Oh ok. That makes sense.

PR #548, opened 2016-02-11
Fix test worker crash

### LINKED ISSUE #540: Crash when running test-worker without AVA CLI
Stack trace:

```js
ava/index.js:26
	log.write();
	    ^

TypeError: log.write is not a function
    at Object.<anonymous> (ava/index.js:26:6)
```

Caused by out-dated code: https://github.com/sindresorhus/ava/blob/master/index.js#L26-L27.
Should be updated to use `console.error` with `chalk`, because `Logger` was adapted for reporting tests.

sotojuan:
This fixes #540.

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,7 +6,6 @@ var beautifyStack = require('./lib/beautify-stack');
 var globals = require('./lib/globals');
 var Runner = require('./lib/runner');
 var send = require('./lib/send');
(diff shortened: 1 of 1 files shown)


sindresorhus:
LGTM. @vdemedes ?

vdemedes:
Thanks @sotojuan!

PR #555, opened 2016-02-15
Consistent error determination

### LINKED ISSUE #553: Thrown errors with no message are ignored
```
test("error", () => {
  throw new Error;
});
```

is considered a passed test. This doesn't seem like correct behaviour. Using the latest version of ava. More than happy to submit a PR if someone can point me in the right direction, thanks!

novemberborn:
The child processes determine whether the test had an error based on its `error` property not being `undefined`. However they then change this property to an empty object if there was no error.

The API determines whether a test had an error based on its (empty) error object having a `message` property. If a test did not have an error, its `error` property is set to `null`. This prevents errors without messages from being reported correctly.

Instead set `error` to `null` in the child processes and rely on that in the API.

I've added a test to validate errors without messages get reported.

Fixes #553.

Note that I did not check what happens when non-errors are thrown. Either the API code was a guard for that, or it's already handled elsewhere, or it's just really bad practice.

### DIFF
--- a/api.js
+++ b/api.js
@@ -114,9 +114,7 @@ Api.prototype._handleStats = function (stats) {
 Api.prototype._handleTest = function (test) {
 	test.title = this._prefixTitle(test.file) + test.title;
 
-	var isError = test.error.message;
-
-	if (isError) {
+	if (test.error) {
 		if (test.error.powerAssertContext) {
 			var message = formatter(test.error.powerAssertContext);
 
@@ -132,8 +130,6 @@ Api.prototype._handleTest = function (test) {
 		}
 
 		this.errors.push(test);
-	} else {
-		test.error = null;
 	}
 
 	this.emit('test', test);
--- a/index.js
+++ b/index.js
@@ -46,10 +46,13 @@ function test(props) {
 		return;
 	}
 
-	props.error = hasError ? serializeError(props.error) : {};
-
-	if (props.error.stack) {
-		props.error.stack = beautifyStack(props.error.stack);
+	if (hasError) {
+		props.error = serializeError(props.error);
+		if (props.error.stack) {
+			props.error.stack = beautifyStack(props.error.stack);
+		}
+	} else {
+		props.error = null;
 	}
 
 	send('test', props);
--- a/lib/test.js
+++ b/lib/test.js
@@ -197,11 +197,14 @@ Object.defineProperty(Test.prototype, 'end', {
 
 Test.prototype._end = function (err) {
 	if (err) {
-		this._setAssertError(new assert.AssertionError({
-			actual: err,
-			message: 'Callback called with an error → ' + err,
-			operator: 'callback'
-		}));
+		if (!(err instanceof Error)) {
+			err = new assert.AssertionError({
+				actual: err,
+				message: 'Callback called with an error: ' + inspect(err, {depth: null}),
+				operator: 'callback'
+			});
+		}
+		this._setAssertError(err);
 
 		this.exit();
 		return;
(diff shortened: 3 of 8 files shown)


vdemedes:
Looks good to me, thanks @novemberborn!

@sindresorhus ?

jokeyrhyme:
Is this related to #365 ?

sindresorhus:
Looks good.

Any chance you could handle the below too, which causes the same false-positive:

```js
test(() => {
  throw '';
});
```

I know it's a big anti-pattern, but it should not cause false-positives no matter how bad it is.

sindresorhus:
@jokeyrhyme I don't think it's directly related, but yeah, we should clean up our error handling as proposed in that issue.

novemberborn:
@sindresorhus 

> Any chance you could handle the below case too, which causes the same false-positive:
>
> ```js 
> test(() => {
>   throw '';
> });
> ```
>
> I know it's a big anti-pattern, but it should not cause false-positives no matter how bad it is.

Had a quick look but it's a little trickier. Since AVA runs in strict mode [this line](https://github.com/sindresorhus/ava/blob/69dbcf23c3ae6f80a2c3a7c926a1e185e7f2f31e/api.js#L131) throws if `test.error` is a literal. Incidentally the original check for `test.error.message` guarded against that. With this PR (as it stands now) falsy literals are rejected but truthy literals would still result in a crash.

I reckon the best approach is to wrap non-errors in [`Test#_setAssertError()`](https://github.com/sindresorhus/ava/blob/69dbcf23c3ae6f80a2c3a7c926a1e185e7f2f31e/lib/test.js#L73:L77). E.g.

```js
if (!(err instanceof Error)) {
  err = { name: 'AvaWrappedError', value: err }
}
```

And then update the reporters to unwrap and log the original error. Perhaps using [`util.inspect()`](https://nodejs.org/api/util.html#util_util_inspect_object_options).

This is a bit more involved so would appreciate any feedback first. I don't think this PR should land before this is resolved though, since the original code accidentally prevents actual AVA crashes by checking for `.message`.

sindresorhus:
Can't we just wrap thrown non-errors in an error? Something like this:

```js
if (!(err instanceof Error)) {
  err = new Error('Non-error thrown with value: ' + err);
}
```

Maybe also include the stack and test title from the test it was thrown for some context.

I don't think it's a good idea to require unwrapping and the reporters should not have to care about something like this.

Prior art: https://github.com/sindresorhus/loud-rejection/blob/b2676e557331991264def77a37ae8b94cab49f55/index.js#L9-L13

-

> Since AVA runs in strict mode this line throws if test.error is a literal.

Good thing. I didn't even know that was possible in sloppy-mode.

novemberborn:
> Can't we just wrap thrown non-errors in an error? Something like this:

Sure. It's kinda useless though if it comes back with `Non-error thrown with value: [Object object]`. Hence my `inspect()` suggestion.

>> Since AVA runs in strict mode this line throws if test.error is a literal.
> Good thing. I didn't even know that was possible in sloppy-mode.

It's a noop in sloppy mode.

sindresorhus:
> It's kinda useless though if it comes back with Non-error thrown with value: [Object object]. Hence my inspect() suggestion.

Ok, got it now. Yeah, we should could use `inspect`.

novemberborn:
@sindresorhus updated.

Followed the approach for rejected promises, which I then updated to use `inspect()` as well. And then noticed the inconsistent callback error handling so changed that too. Hopefully I understood the various cases correctly!

sindresorhus:
Thank you @novemberborn. Very happy we got this sorted out :)

Issue #549, opened 2016-02-12
Support `.only` concept for all files


kentcdodds:
Right now, if you have multiple files with multiple tests, and you place `.only` on a test in one of the files, you might expect that it only runs that test (this is how jasmine and mocha operate), but what actually happens is all the tests in the other files will run. All that `.only` does right now is prevent tests in that same file from running.

I think it could be really handy (especially during development) to only run a few tests for the specific feature you're working on.

Right now, it's not entirely unreasonable to work around this by pointing AVA directly at the file that you're working in.

As for what it should be called, honestly I think that it should be `.only` and the current `.only` functionality should be renamed to something like `.onlyInFile` or something... I actually don't see a solid use case for the current functionality so perhaps this feature request could simply be renamed to something like: `Change .only to be scoped to the entire test suite`

Thoughts? :thought_balloon:

BarryThePenguin:
I'm in the same boat as Kent.

I wrote a bit about my use case in #471

I'm not aware of a reason for the current behavior or a `.onlyInFile` method.

sindresorhus:
Agreed. This is a bug.

naptowncode:
Looks like this could be done by sending the `hasExclusive` flag along with `testCount` in the `ava-stats` message. Then if `api.run()` gets any messages with `hasExclusive: true` it could send a `runOnlyExclusive: true` flag to the `ava-run` message. Does that sound right?

novemberborn:
This landed in #556.

Issue #471, opened 2016-01-24
TAP: Should test.only() contribute to skipped count summary?


therealklanni:
I personally think it should, as this would be useful information.

Currently, if you use the `test.only()` method, it doesn't tally any other `test()` calls as skipped and thus they go unreported (as if they don't exist).

Side note: maybe even useful in the normal AVA output?

```
7 passed
2 skipped
```

BarryThePenguin:
Personally I use `test.only()` for debugging. If I use it on a single test, I only care about the result of that test.

In `karma-spec-reporter`there are config values to suppress certain results. I turn report of skipped tests off in local development to stop the output being polluted with skipped test lines. Then I rely on the CI server to give me an accurate report of skipped tests. 

```
specReporter: {
        maxLogLines: 5,         // limit number of lines logged per test 
        suppressErrorSummary: true,  // do not print error summary 
        suppressFailed: false,  // do not print information about failed tests 
        suppressPassed: false,  // do not print information about passed tests 
        suppressSkipped: true  // do not print information about skipped tests 
      },
```

In summary I guess it's the job of the individual tap reporter to handle how to interpret the output. And AVA should be giving an accurate output of test results.

therealklanni:
I agree that it's the job of the reporter to determine how to interpret the output, so yes, I think AVA should give an accurate output (i.e. report all skips).

sindresorhus:
:+1: This also catches the case of forgetting to remove an `.only()`.

@therealklanni Interested in doing a PR?

jamestalmage:
> This also catches the case of forgetting to remove an .only().

I have a really crazy idea. Should the use of `.only()` cause a non-zero exit code? Or the ability to configure it to exit with a non-zero exit code in that case?

We could probably create a linter rule that handled warning about `.only()` usage, but I think it would be so nice if AVA just handled it for you.

sindresorhus:
> Should the use of .only() cause a non-zero exit code? Or the ability to configure it to exit with a non-zero exit code in that case?

I like the idea, but need to think about possible cases where you don't want it to fail. I'm also not very interested if it has to be configurable. Trying to keep config to a minimum.

> We could probably create a linter rule that handled warning about .only() usage

Already planned regardless: https://github.com/sindresorhus/eslint-plugin-ava/issues/5

therealklanni:
@sindresorhus I can possibly take a look at this sometime this week.

If anyone else wants to give it a go before I get to it, feel free!

[...]

PR #556, opened 2016-02-15
Support .only for all files

naptowncode:
My suggested solution for #549. Needs unit tests, of course.

### DIFF
--- a/api.js
+++ b/api.js
@@ -53,6 +53,7 @@ Api.prototype._reset = function () {
 	this.failCount = 0;
 	this.fileCount = 0;
 	this.testCount = 0;
+	this.hasExclusive = false;
 	this.errors = [];
 	this.stats = [];
 	this.tests = [];
@@ -103,6 +104,15 @@ Api.prototype._handleTeardown = function (data) {
 };
 
 Api.prototype._handleStats = function (stats) {
+	if (this.hasExclusive && !stats.hasExclusive) {
+		return;
+	}
+
+	if (!this.hasExclusive && stats.hasExclusive) {
+		this.hasExclusive = true;
+		this.testCount = 0;
+	}
+
 	this.testCount += stats.testCount;
 };
 
@@ -198,9 +208,12 @@ Api.prototype.run = function (files) {
 							self.emit('ready');
 
 							var method = self.options.serial ? 'mapSeries' : 'map';
+							var options = {
+								runOnlyExclusive: self.hasExclusive
+							};
 
 							resolve(Promise[method](files, function (file, index) {
-								return tests[index].run().catch(function (err) {
+								return tests[index].run(options).catch(function (err) {
 									// The test failed catastrophically. Flag it up as an
 									// exception, then return an empty result. Other tests may
 									// continue to run.
--- a/index.js
+++ b/index.js
@@ -70,6 +70,7 @@ function exit() {
 }
 
 globals.setImmediate(function () {
+	var hasExclusive = runner.tests.hasExclusive;
 	var numberOfTests = runner.tests.tests.concurrent.length + runner.tests.tests.serial.length;
 
 	if (numberOfTests === 0) {
@@ -78,13 +79,14 @@ globals.setImmediate(function () {
 	}
 
 	send('stats', {
-		testCount: numberOfTests
+		testCount: numberOfTests,
+		hasExclusive: hasExclusive
 	});
 
 	runner.on('test', test);
 
-	process.on('ava-run', function () {
-		runner.run().then(exit);
+	process.on('ava-run', function (options) {
+		runner.run(options).then(exit);
 	});
 });
 
--- a/lib/fork.js
+++ b/lib/fork.js
@@ -134,14 +134,14 @@ module.exports = function (file, opts) {
 		isReady = true;
 	});
 
-	promise.run = function () {
+	promise.run = function (options) {
 		if (isReady) {
-			send(ps, 'run');
+			send(ps, 'run', options);
 			return promise;
 		}
 
 		ps.on('stats', function () {
-			send(ps, 'run');
+			send(ps, 'run', options);
 		});
 
 		return promise;
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -88,13 +88,17 @@ Runner.prototype._addTestResult = function (result) {
 	this.emit('test', props);
 };
 
-Runner.prototype.run = function () {
+Runner.prototype.run = function (options) {
 	var stats = this.stats = {
 		failCount: 0,
 		passCount: 0,
 		skipCount: 0,
 		testCount: 0
 	};
+
+	if (options.runOnlyExclusive && !this.tests.hasExclusive) {
+		return Promise.resolve();
+	}
 
 	this.tests.on('test', this._addTestResult);
 
--- a/test/api.js
+++ b/test/api.js
@@ -600,6 +600,26 @@ test('test file with only skipped tests does not create a failure', function (t)
 		});
 });
 
+test('test file with exclusive tests causes non-exclusive tests in other files to be ignored', function (t) {
+	t.plan(4);
+
+	var files = [
+		path.join(__dirname, 'fixture/exclusive.js'),
+		path.join(__dirname, 'fixture/exclusive-nonexclusive.js'),
+		path.join(__dirname, 'fixture/one-pass-one-fail.js')
+	];
+
+	var api = new Api();
+
+	api.run(files)
+		.then(function () {
+			t.ok(api.hasExclusive);
+			t.is(api.testCount, 2);
+			t.is(api.passCount, 2);
+			t.is(api.failCount, 0);
+		});
+});
+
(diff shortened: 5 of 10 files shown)


naptowncode:
@sindresorhus @vdemedes @jamestalmage Added unit test. Should be ready to go.

BarryThePenguin:
Is there a test for multiple exclusive tests in multiple files?

naptowncode:
The test uses three files: One file that has only an exclusive test, one file that has an exclusive test and a non-exclusive test, one file that has two non-exclusive test. Total of five tests across three files with two exclusive tests in different files.

novemberborn:
I'm wondering how this should work with `--watch` mode. Let's say we have two test files: `foo.js` and `bar.js`. The latter contains a `.only`.

Initially all test files are run and because `bar.js` contains a `.only`, only that test is actually executed.

Subsequently `foo.js` is changed. The watcher will run just `foo.js` which does not contain a `.only`. Now all tests in `foo.js` are executed.

Later a non-test file is changed and all test files are rerun. Only the `.only` test from `bar.js` is executed.

This isn't necessarily bad but it may be surprising to some. I wonder if `.only` should be sticky so that when changes are detected the watcher still only executes `.only` tests. But then if you remove the `.only` it starts executing all tests again. Similarly adding a `.only` switches back into the sticky mode.

This is a little tricky to implement though #544 requires similar per-test-file housekeeping.

Anyway not meant as a blocker to this PR landing. We can discuss in a separate issue if people think `.only` should stick across watcher runs.

api.js
```diff
@@ -108,6 +109,15 @@ Api.prototype._handleExceptions = function (data) {
 };
 
 Api.prototype._handleStats = function (stats) {
+	if (this.hasExclusive && !stats.hasExclusive) {
```
novemberborn: Could you add a comment as to why the function is returning here? It's because if we're running exclusive tests, but the test file does not include exclusive tests, then we don't want those tests to be counted, right?

api.js
```diff
@@ -108,6 +109,15 @@ Api.prototype._handleExceptions = function (data) {
...
+		return;
+	}
+
+	if (!this.hasExclusive && stats.hasExclusive) {
+		this.hasExclusive = true;
+		this.testCount = 0;
```
novemberborn: Similarly why is this being reset? Because we've already counted tests we're now not going to run, right? I appreciate `test-collection` has similar logic without comments but I reckon it'd be easier to understand for the next person if we'd explain it here.

index.js
```diff
@@ -78,12 +79,18 @@ globals.setImmediate(function () {
...
+		testCount: numberOfTests,
+		hasExclusive: hasExclusive
 	});
 
 	runner.on('test', test);
 
-	process.on('ava-run', function () {
+	process.on('ava-run', function (options) {
+		if (options.runOnlyExclusive && !hasExclusive) {
+			exit();
```
novemberborn: Maybe pass `runOnlyExclusive` to `runner.run()`? It can still create its `stats` object inside the `run()` method, but bail out if it has no exclusive tests.

test/api.js
```diff
@@ -616,6 +616,26 @@ test('test file with only skipped tests does not create a failure', function (t)
...
+test('test file with exclusive tests causes non-exclusive tests in other files to be ignored', function (t) {
+	t.plan(4);
+
+	var files = [
+		path.join(__dirname, 'fixture/exclusive.js'),
+		path.join(__dirname, 'fixture/exclusive-nonexclusive.js'),
+		path.join(__dirname, 'fixture/one-pass-one-fail.js')
+	];
+
+	var api = new Api(files);
```
novemberborn: `master` moves the `files` argument to `api.run(files)`. You'll have to rebase this branch, sorry :)

novemberborn:
Solution makes sense to me. Great job given the amount of (necessary) indirection involved in coordinating test runs :+1:

naptowncode:
I'm not actually sure how it works with `--watch`. At the very least I should add some documentation for that.

api.js
naptowncode: Correct, although I'm not super keen on just dropping this info on the floor. I just went for consistency with how non-exclusive tests are dropped within `test-collection`. I'll add a comment to that effect.

That said, I would favor changing the reporting completely as suggested by #471.

api.js
naptowncode: Yup, same logic. Again, not too keen on it but couldn't think of a more elegant solution. Suggestions welcome!

index.js
naptowncode: Good call. That's almost certainly a better approach.

novemberborn:
> I'm not actually sure how it works with --watch. At the very least I should add some documentation for that.

Watch itself isn't documented yet so don't worry about it. We can get this in and then discuss repercussions in a new issue.

api.js
novemberborn: Nah it's good enough as is, just requires some extra thinking.

naptowncode:
@novemberborn I've incorporated those tweaks. If everything looks good, merging this would make it easier to work on #471. Thanks!

sindresorhus:
LGTM. @vdemedes @jamestalmage ?

vdemedes:
LGTM from me! @naptowncode could you please resolve the merge conflict? Big thanks for your work, this is an important change!

lib/runner.js
```diff
@@ -88,14 +90,26 @@ Runner.prototype._addTestResult = function (result) {
...
 		skipCount: 0,
 		testCount: 0
 	};
 
+	return this.stats;
+};
+
+Runner.prototype.run = function (options) {
+	options = options || {};
+	this._reset();
```
novemberborn: You no longer need the `_reset()` method now that `run` is always called.

lib/fork.js
```diff
@@ -134,14 +134,16 @@ module.exports = function (file, opts) {
 		isReady = true;
 	});
 
-	promise.run = function () {
+	promise.run = function (options) {
+		options = options || {};
```
novemberborn: Given that this is an internal method we should assume it's always called with an `options` object. You'll have to update `test/fork.js` though. Maybe some other places too.

api.js
```diff
@@ -177,24 +187,30 @@ Api.prototype.run = function (files) {
...
 				tests.forEach(function (test) {
 					var counted = false;
 
-					function tryRun() {
+					function tryRun(stats) {
 						if (counted) {
 							return;
 						}
 
+						runOnlyExclusive = runOnlyExclusive || stats.hasExclusive;
```
novemberborn: I like this approach! Seems cleaner than the `this.hasExclusive` approach on top, though it can't fix the `testCount`. Maybe the test count should be incremented here as well and we can do away with `_handleStats`?

Note though that `tryRun` is also called when on `test.catch()`. You'd have to wrap it there to avoid passing the error as the `stats` object and then add a guard for `stats` being undefined.

Or maybe stick with the `_handleStats()` approach. Just not both :wink:

index.js
```diff
@@ -78,13 +79,18 @@ globals.setImmediate(function () {
...
 	runner.on('test', test);
 
-	process.on('ava-run', function () {
-		runner.run().then(exit);
+	process.on('ava-run', function (eventOptions) {
+		var runOptions = {
+			runOnlyExclusive: eventOptions.runOnlyExclusive
+		};
+
+		runner.run(runOptions).then(exit);
```
novemberborn: Just use `eventOptions` here I think. And rename it to `options`.

api.js
naptowncode: It does seem redundant but it's to avoid responsibility creep. `_handleStats()` is concerned with gathering the total `testCount` before tests run while this method is waiting for each test to emit stats before starting them all. Probably the `stats` event itself is overloaded by using it for both purposes.

I'll leave `_handleStats()` as it is but wrap the `tryRun()` on `test.catch()`.

novemberborn: Sure, but then why not use `self.hasExclusive` here?

lib/fork.js
novemberborn: Yes. Tests aside there's only one place the "run chain" is called and it's from the Api.

api.js
naptowncode: To keep to test execution concerns contained in `tryRun()` and separate from the bookkeeping concerns in `_handleStats()`. But it's not a big hangup for me. I'll go ahead and update `tryRun()` to use `self.hasExclusive`.

naptowncode:
@novemberborn @vdemedes @sindresorhus Updated and rebased. Also added a unit test that should have been there earlier for the `.runOnlyExclusive` option to `Runner.run()`.

sindresorhus:
@naptowncode Tests are failing on Travis it seems: https://travis-ci.org/sindresorhus/ava/builds/112949149

novemberborn:
@sindresorhus that's just one test in Node 0.10 though.

novemberborn:
@naptowncode LGTM!

vdemedes:
Thank you so much, @naptowncode!

novemberborn:
Please see #593 for how `.only` should interact in watch mode.

PR #557, opened 2016-02-16
improve test coverage for calling classes without the `new` keyword

Carnubak:
While looking into #161 I noticed that several tests for the constructions of objects were not checking the message from the expected exception, which caused a couple of bugs to be missed (and some lines not to be covered).

I improved the tests and corrected the bugs, and also added similar tests to all constructable classes that were making the "invoke with new" check. I also standardized the test message and the general structure, since different tests seemed to be handling it with different naming conventions for the auxiliary variables.


lib/concurrent.js
```diff
@@ -6,7 +6,7 @@ var BAIL_ERROR = new Error();
 module.exports = Concurrent;
 
 function Concurrent(tests, bail) {
-	if (!this instanceof Concurrent) {
+	if (!(this instanceof Concurrent)) {
```
sindresorhus: I'm surprised the linter didn't catch this. Would have thought ESLint would have rule for this.

sindresorhus:
Good catch! Thank you. Keep'em coming ;)

PR #564, opened 2016-02-18
Logger coverage improvement and refactor

Carnubak:
While looking to improve coverage, `Logger` seemed like the obvious choice, since I had already started on it, and it had the lowest coverage in `lib/`. The lines to be covered also looked quite simple.

After that, I noticed that there was an opportunity to refactor its behavior, since there was a hidden time dependency between `use()` and every other method in the module. Nothing worked if `use()` was not called before, so that looked like something the constructor would want to do. Moving that to the constructor showed that there was even a bit more room for improvement, as api was only used in order to set it in the current reporter, so I made that dependency explicit by taking it out of `Logger` and to each reporter.

I did try to cover the `cli.js` lines that were responsible for creating the `Logger` and right `*Reporter`, but I couldn't figure out how to test it. Please comment on the changes, and also on how one could be able to test that behavior.

Thanks.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -100,15 +100,17 @@ var api = new Api(cli.input.length ? cli.input : arrify(conf.files), {
 	cacheEnabled: cli.flags.cache !== false
 });
 
-var logger;
+var reporter;
 
 if (cli.flags.tap) {
-	logger = new Logger(tapReporter(api));
+	reporter = tapReporter(api);
 } else if (cli.flags.verbose || isCi) {
-	logger = new Logger(verboseReporter(api));
+	reporter = verboseReporter(api);
 } else {
-	logger = new Logger(miniReporter(api));
+	reporter = miniReporter(api);
 }
+
+var logger = new Logger(reporter);
 
 logger.start();
 
--- a/cli.js
+++ b/cli.js
@@ -103,13 +103,14 @@ var api = new Api(cli.input.length ? cli.input : arrify(conf.files), {
 var reporter;
 
 if (cli.flags.tap) {
-	reporter = tapReporter(api);
+	reporter = tapReporter();
 } else if (cli.flags.verbose || isCi) {
-	reporter = verboseReporter(api);
+	reporter = verboseReporter();
 } else {
-	reporter = miniReporter(api);
+	reporter = miniReporter();
 }
 
+reporter.api = api;
 var logger = new Logger(reporter);
 
 logger.start();
--- a/cli.js
+++ b/cli.js
@@ -100,15 +100,14 @@ var api = new Api(cli.input.length ? cli.input : arrify(conf.files), {
 	cacheEnabled: cli.flags.cache !== false
 });
 
-var logger = new Logger();
-logger.api = api;
+var logger;
 
 if (cli.flags.tap) {
-	logger.use(tapReporter());
+	logger = new Logger(tapReporter(), api);
 } else if (cli.flags.verbose || isCi) {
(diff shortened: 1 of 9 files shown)


cli.js
```diff
@@ -100,15 +100,14 @@ var api = new Api(cli.input.length ? cli.input : arrify(conf.files), {
 	cacheEnabled: cli.flags.cache !== false
 });
 
-var logger = new Logger();
-logger.api = api;
+var logger;
 
 if (cli.flags.tap) {
-	logger.use(tapReporter());
+	logger = new Logger(tapReporter(api));
```
vdemedes: I think this would be cleaner:

```js
var reporter;

if (cli.flags.tap) {
  reporter = new TapReporter();
}

var logger = new Logger(reporter);
```

lib/logger.js
```diff
@@ -8,15 +8,12 @@ function Logger() {
 	Object.keys(Logger.prototype).forEach(function (key) {
 		this[key] = this[key].bind(this);
 	}, this);
+
+	this.reporter = reporter;
```
vdemedes: I'd prefer to assign `api` to reporter here:

```js
this.reporter.api = this.api
```

lib/reporters/mini.js
```diff
@@ -5,14 +5,15 @@ var StringDecoder = require('string_decoder').StringDecoder;
...
+function MiniReporter(api) {
 	if (!(this instanceof MiniReporter)) {
-		return new MiniReporter();
+		return new MiniReporter(api);
 	}
 
 	this.reset();
 	this.stream = process.stderr;
 	this.stringDecoder = new StringDecoder();
+	this.api = api;
```
vdemedes: See https://github.com/sindresorhus/ava/pull/564/files#r53278132

vdemedes:
Thank you for your contribution, I commented in a few places ;) Looking forward to merge it.

cli.js
Carnubak: Agreed. This looks much cleaner. I'll update it.

lib/logger.js
Carnubak: Is there any particular reason the `Logger` needs to know about `api`? I'm not familiar with the overall design, but it seems to me that `Logger` only contains `api` in order to pass it to the different reporters, and does not use it on its own. It's a bit confusing to me.

cli.js
Carnubak: Also, if reporters are created using `new`, the check on their constructors (e.g. https://github.com/sindresorhus/ava/blob/master/lib/reporters/tap.js#L14-L16) can just throw if used without `new`, as other modules do.

Carnubak:
I made the instantiation clean up at `cli.js`, but I still would like a few clarifications on the rest.

lib/logger.js
vdemedes: Yes, `Logger` passes `api` to reporters. You can set `api` for reporters in `cli.js` too, I think.

Carnubak: How about this, in `cli.js`:

```
if (cli.flags.api) {
	reporter = tapReporter(api);
}
[...]
reporter.api = api;
logger = new Logger(reporter);
```

I think that reads more clearly, and the `Logger` is no longer manipulating the reporter behind the scenes.

vdemedes: I don't think you need `reporter.api = api` if you are passing it to `tapReporter()`.

sindresorhus:
Thank you @Carnubak :D

PR #565, opened 2016-02-18
Add `todo` test modifier

### LINKED ISSUE #563: t.todo(), like t.skip() but without a function
I'll often use `test.skip` to notify me and others that there are more tests we could write. I'd like to be able to do this without having to specify a function. Something like:

```javascript
test.skip('adds two numbers together')
test.skip('throws an error when the first argument is a 🐮') // that's for you Sindre :-)
```

Thoughts? :thought_balloon:

BarryThePenguin:
Seems too simple...

Fixes #563

### DIFF
--- a/lib/reporters/tap.js
+++ b/lib/reporters/tap.js
@@ -27,6 +27,15 @@ TapReporter.prototype.start = function () {
 TapReporter.prototype.test = function (test) {
 	var output;
 
+	var directive = '';
+	var passed = test.todo ? 'not ok' : 'ok';
+
+	if (test.todo) {
+		directive = '# TODO';
+	} else if (test.skip) {
+		directive = '# SKIP';
+	}
+
 	if (test.error) {
 		output = [
 			'# ' + test.title,
@@ -41,7 +50,7 @@ TapReporter.prototype.test = function (test) {
 	} else {
 		output = [
 			'# ' + test.title,
-			format('ok %d - %s', ++this.i, test.title)
+			format('%s %d - %s %s', passed, ++this.i, test.title, directive).trim()
 		];
 	}
 
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -5,6 +5,8 @@ var Promise = require('bluebird');
 var optionChain = require('option-chain');
 var TestCollection = require('./test-collection');
 
+function noop() {}
+
 var chainableMethods = {
 	spread: true,
 	defaults: {
@@ -12,6 +14,7 @@ var chainableMethods = {
 		serial: false,
 		exclusive: false,
 		skipped: false,
+		todo: false,
 		callback: false
 	},
 	chainableMethods: {
@@ -20,6 +23,7 @@ var chainableMethods = {
 		before: {type: 'before'},
 		after: {type: 'after'},
 		skip: {skipped: true},
+		todo: {todo: true},
 		only: {exclusive: true},
 		beforeEach: {type: 'beforeEach'},
 		afterEach: {type: 'afterEach'},
@@ -51,6 +55,16 @@ optionChain(chainableMethods, function (opts, title, fn) {
 		title = null;
 	}
 
+	if (opts.todo) {
+		fn = noop;
+
+		if (typeof title !== 'string') {
+			throw new TypeError('`todo` tests require a title');
+		}
+	} else if (opts.skipped && typeof fn !== 'function') {
+		throw new TypeError('Expected a function. Use `test.todo()` for tests without a function.');
+	}
+
 	if (this._serial) {
 		opts.serial = true;
 	}
@@ -63,10 +77,14 @@ optionChain(chainableMethods, function (opts, title, fn) {
 }, Runner.prototype);
 
 Runner.prototype._addTestResult = function (result) {
-	if (result.result.metadata.type === 'test') {
+	var test = result.result;
+
+	if (test.metadata.type === 'test') {
 		this.stats.testCount++;
 
-		if (result.result.metadata.skipped) {
+		if (test.metadata.todo) {
+			this.stats.todoCount++;
+		} else if (test.metadata.skipped) {
 			this.stats.skipCount++;
 		}
 	}
@@ -75,14 +93,13 @@ Runner.prototype._addTestResult = function (result) {
 		this.stats.failCount++;
 	}
 
-	var test = result.result;
-
 	var props = {
 		duration: test.duration,
 		title: test.title,
 		error: result.reason,
 		type: test.metadata.type,
-		skip: test.metadata.skipped
+		skip: test.metadata.skipped,
+		todo: test.metadata.todo
 	};
 
 	this.emit('test', props);
@@ -93,6 +110,7 @@ Runner.prototype.run = function (options) {
 		failCount: 0,
 		passCount: 0,
 		skipCount: 0,
+		todoCount: 0,
 		testCount: 0
 	};
 
@@ -103,6 +121,6 @@ Runner.prototype.run = function (options) {
 	this.tests.on('test', this._addTestResult);
 
 	return Promise.resolve(this.tests.build(this._bail).run()).then(function () {
-		stats.passCount = stats.testCount - stats.failCount - stats.skipCount;
+		stats.passCount = stats.testCount - stats.failCount - stats.skipCount - stats.todoCount;
 	});
 };
--- a/readme.md
+++ b/readme.md
@@ -248,12 +248,20 @@ test.only('will be run', t => {
 
 ### Skip-tests
 
-Skip-tests are shown in the output as skipped but never run.
+Skip-tests are shown in the output as skipped but never run. Skip-tests require a function.
 
 ```js
 test.skip('will not be run', t => {
 	t.fail();
 });
+```
+
+### Todo-tests
+
+Todo-tests, like skip-tests, are shown in the output but are never run. They can be useful for planning future tests. Todo-tests require only a test title.
+
+```js
+test.todo('will think about writing this later');
 ```
 
 ### Before & after hooks
@@ -478,6 +486,7 @@ AVA automatically removes unrelated lines in stack traces, allowing you to find 
 ### test.cb([title], body)
 ### test.only([title], body)
 ### test.skip([title], body)
+### test.todo(title)
 ### test.before([title], body)
 ### test.after([title], body)
 ### test.beforeEach([title], body)
--- a/test/reporters/tap.js
+++ b/test/reporters/tap.js
@@ -146,3 +146,40 @@ test('results does not show skipped tests if there are none', function (t) {
 	t.is(actualOutput, expectedOutput);
 	t.end();
 });
+
+test('todo test', function (t) {
+	var reporter = tapReporter();
+
+	var actualOutput = reporter.test({
+		title: 'should think about doing this',
+		passed: false,
+		skip: true,
+		todo: true
+	});
+
+	var expectedOutput = [
+		'# should think about doing this',
+		'not ok 1 - should think about doing this # TODO'
+	].join('\n');
+
+	t.is(actualOutput, expectedOutput);
+	t.end();
+});
+
+test('skip test', function (t) {
+	var reporter = tapReporter();
+
+	var actualOutput = reporter.test({
+		title: 'skipped',
+		passed: true,
+		skip: true
+	});
+
+	var expectedOutput = [
+		'# skipped',
+		'ok 1 - skipped # SKIP'
+	].join('\n');
+
+	t.is(actualOutput, expectedOutput);
+	t.end();
+});
--- a/test/runner.js
+++ b/test/runner.js
@@ -177,7 +177,7 @@ test('test types and titles', function (t) {
 });
 
 test('skip test', function (t) {
-	t.plan(3);
+	t.plan(5);
 
 	var runner = new Runner();
 	var arr = [];
@@ -190,9 +190,39 @@ test('skip test', function (t) {
 		arr.push('b');
 	});
 
+	t.throws(function () {
+		runner.skip('should be a todo');
+	}, {message: 'Expected a function. Use `test.todo()` for tests without a function.'});
+
 	runner.run({}).then(function () {
 		t.is(runner.stats.testCount, 2);
 		t.is(runner.stats.passCount, 1);
+		t.is(runner.stats.skipCount, 1);
+		t.same(arr, ['a']);
+		t.end();
+	});
+});
+
+test('todo test', function (t) {
+	t.plan(5);
+
+	var runner = new Runner();
+	var arr = [];
+
+	runner.test(function () {
+		arr.push('a');
+	});
+
+	runner.todo('todo');
+
+	t.throws(function () {
+		runner.todo(function () {});
+	}, {message: '`todo` tests require a title'});
+
+	runner.run({}).then(function () {
+		t.is(runner.stats.testCount, 2);
+		t.is(runner.stats.passCount, 1);
+		t.is(runner.stats.todoCount, 1);
 		t.same(arr, ['a']);
 		t.end();
 	});
--- a/api.js
+++ b/api.js
@@ -50,6 +50,7 @@ Api.prototype._reset = function () {
 	this.exceptionCount = 0;
 	this.passCount = 0;
 	this.skipCount = 0;
+	this.todoCount = 0;
 	this.failCount = 0;
 	this.fileCount = 0;
 	this.testCount = 0;
@@ -217,7 +218,7 @@ Api.prototype.run = function (files) {
 									});
 
 									return {
-										stats: {passCount: 0, skipCount: 0, failCount: 0},
+										stats: {passCount: 0, skipCount: 0, todoCount: 0, failCount: 0},
 										tests: []
 									};
 								});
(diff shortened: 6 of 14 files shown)


lib/runner.js
```diff
@@ -46,6 +46,8 @@ util.inherits(Runner, EventEmitter);
 module.exports = Runner;
 
 optionChain(chainableMethods, function (opts, title, fn) {
+	fn = fn || function () {};
```
vdemedes: I think having `function noop() {}` at the top of the file and using it here instead of anonymous function inline would be better.

vdemedes:
Thanks for this PR, @BarryThePenguin!

lib/runner.js
BarryThePenguin: Cool, can do. Was also wondering if `||` is enough, or should I explicitly be checking for `typeof fn === 'undefined'`?

vdemedes: Nah, `||` is enough here.

sindresorhus:
Can you also update the docs?

https://github.com/sindresorhus/ava#testskiptitle-body
https://github.com/sindresorhus/ava#skip-tests

sindresorhus:
This allows omitting test function for all tests, not just `.skip`, @vdemedes are we ok with that?

vdemedes:
@sindresorhus Very good point. I guess we came back to the problems discussed in #9, `test.todo()` makes more sense in that case. But for now, maybe we should restrict function-less tests only to `.skip`.

BarryThePenguin:
Mmm... in this context skip and todo are synonymous with each other. Maybe it's worth exploring?

BarryThePenguin:
Also, noting the api change, this allows for `test.skip()`. Not sure how helpful that is to people

kentcdodds:
> Also, noting the api change, this allows for test.skip(). Not sure how helpful that is to people

That makes sense. I wonder though if we should think about explicitly disallowing that to avoid people abusing it and confusing others.

BarryThePenguin:
I'd be happy to investigate the possibility of having both a `.skip()` and `.todo()` modifier.

Thoughts so far: 
- skip has optional title, requires a test function
- todo requires a test title, won't accept a test function

Would also require adding todo to the tap reporter output

kentcdodds:
You know, I would actually prefer to leave `skip` as it is and introduce a `todo` that only accepts a message. I would still like to have the `todo` in the output so it's a constant reminder. I think that's the value of coding the todo rather than just using a comment.

BarryThePenguin:
I've had a go at the `todo` modifier. I also had a go at implementing the tap output for both `todo` and `skip` directives

lib/reporters/tap.js
```diff
@@ -27,6 +27,10 @@ TapReporter.prototype.start = function () {
 TapReporter.prototype.test = function (test) {
 	var output;
 
+	var passed = test.todo ? 'not ok' : 'ok';
+	var todoDirective = test.todo ? '# TODO' : '';
+	var skipDirective = test.skipped ? '# SKIP' : '';
```
novemberborn: Perhaps do:

```
var directive = '';
if (test.todo) {
  directive = '# TODO';
} else if (test.skipped) {
  directive = '# SKIP';
}
```

Which simplifies the `todoDirective || skipDirective || ''` statement below and removes the unnecessary `|| ''` fall-through.

lib/runner.js
```diff
@@ -46,6 +50,12 @@ util.inherits(Runner, EventEmitter);
 module.exports = Runner;
 
 optionChain(chainableMethods, function (opts, title, fn) {
+	if (opts.todo && typeof title === 'function') {
+		throw new TypeError('Expected a string');
+	} else if (opts.todo) {
+		fn = noop;
+	}
```
novemberborn: I'd prefer a single `if (opts.todo) {}`, then nest the `title` check. Took me a while to figure out `fn = noop` was used when `title` was not a function :smile:

lib/runner.js
```diff
@@ -46,6 +50,12 @@ util.inherits(Runner, EventEmitter);
 module.exports = Runner;
 
 optionChain(chainableMethods, function (opts, title, fn) {
+	if (opts.todo && typeof title === 'function') {
+		throw new TypeError('Expected a string');
```
novemberborn: The error message could be more informative. Perhaps `TODO tests need a title` or something along those lines.

lib/runner.js
kentcdodds: In addition, I would make the check be `typeof title !== 'string'` :+1:

lib/runner.js
```diff
@@ -20,6 +23,7 @@ var chainableMethods = {
 		before: {type: 'before'},
 		after: {type: 'after'},
 		skip: {skipped: true},
+		todo: {skipped: true, todo: true},
```
novemberborn: It might be better to not set `skipped: true` here. That would require further clauses where `skipped` is used to determine whether the test is skipped, but it also removes exceptions where you have to check for `!todo`.

@vdemedes thoughts?

readme.md
```diff
@@ -248,12 +248,14 @@ test.only('will be run', t => {
 
 ### Skip-tests
 
-Skip-tests are shown in the output as skipped but never run.
+Skip-tests are shown in the output as skipped but never run. Skip-tests can be written without specifying the test function. This can be useful for planning future tests.
```
novemberborn: If I'm understanding things correctly this needs to be changed to reflect `test.todo` right?

test/reporters/tap.js
```diff
@@ -146,3 +146,39 @@ test('results does not show skipped tests if there are none', function (t) {
...
 	t.end();
 });
+
+test('todo test', function (t) {
+	var reporter = tapReporter();
+
+	var actualOutput = reporter.test({
+		title: 'should think about doing this',
+		passed: false,
+		todo: true
```
novemberborn: If we're sticking to having both `skipped` and `todo` in the metadata then you'll have to add `skipped: true` here.

kentcdodds:
I'm not certain I follow the results of all the changes. But I'll just say that I would recommend that `skip` require a function and the only way to avoid having to specify a function is if you use `todo`. Fewer ways to do the same thing is :+1: :D

novemberborn:
> But I'll just say that I would recommend that skip require a function and the only way to avoid having to specify a function is if you use `todo`. Fewer ways to do the same thing is :+1: 

Yes agreed. Would be good to make `test.skip('my test')` fail with a helpful error, pointing at `test.todo()`.

@BarryThePenguin changes are looking good generally, nice work. Would be good to clear up some of the logic and state management so it's easier to follow how these special test cases (pun very much intended) are handled.

lib/reporters/tap.js
BarryThePenguin: Ha, great. Will do :hankey:

lib/runner.js
BarryThePenguin: :+1:

lib/runner.js
BarryThePenguin: I can change it. Anyone else want to clarify before I do?

lib/runner.js
BarryThePenguin: Up until now I've been treating todo as a skipped test, that has failed, mainly based on the TAP spec. Kind of depends on how tightly integrated the runner implementation is with the reporter output.

readme.md
BarryThePenguin: This will need to change, yes

lib/runner.js
BarryThePenguin: FYI...

```js
var title = 'todo';
typeof typeof title !== 'string';
// === false
```

:sleeping: :zzz: :zzz:

kentcdodds: lol

BarryThePenguin:
Just went through the review.

Added Mini and Verbose reporter handling for todo + tests

vdemedes:
I like it, LGTM from me!

@sindresorhus @novemberborn @jamestalmage ?

lib/runner.js
```diff
@@ -51,6 +55,16 @@ optionChain(chainableMethods, function (opts, title, fn) {
...
 	}
 
+	if (opts.todo) {
+		fn = noop;
+
+		if (typeof title !== 'string') {
+			throw new TypeError('Expected a string');
+		}
+	} else if (opts.skipped && typeof fn !== 'function') {
+		throw new TypeError('Expected a function. Use todo for tests without a function');
```
sindresorhus:
```
throw new TypeError('Expected a function. Use `test.todo()` for tests without a function.');
```

sindresorhus:
Looks pretty good. The docs needs to be updated and some inline feedback.

lib/runner.js
```diff
@@ -51,6 +55,16 @@ optionChain(chainableMethods, function (opts, title, fn) {
 		title = null;
 	}
 
+	if (opts.todo) {
+		fn = noop;
+
+		if (typeof title !== 'string') {
+			throw new TypeError('Expected a string');
```
sindresorhus:
```
throw new TypeError('`todo` tests require a title');
```

lib/runner.js
BarryThePenguin: Any other suggestions on what to do here?

BarryThePenguin:
Updated. Happy to squash/rebase to make the commit history a little more sane :+1:

sindresorhus:
LGTM

sindresorhus:
> Happy to squash/rebase to make the commit history a little more sane

:+1:

lib/reporters/tap.js
```diff
@@ -27,6 +27,15 @@ TapReporter.prototype.start = function () {
 TapReporter.prototype.test = function (test) {
 	var output;
 
+	var directive = '';
+	var passed = test.todo ? 'not ok' : 'ok';
+
+	if (test.todo) {
+		directive = '# TODO';
+	} else if (test.skipped) {
```
novemberborn: This should be `test.skip`, from <https://github.com/BarryThePenguin/ava/blob/bb3c972cc4df4b3df0139b5d6f738cb18da0e6e7/lib/runner.js#L99>.

lib/runner.js
```diff
@@ -63,10 +77,12 @@ optionChain(chainableMethods, function (opts, title, fn) {
...
 
 Runner.prototype._addTestResult = function (result) {
-	if (result.result.metadata.type === 'test') {
+	var test = result.result;
+
+	if (test.metadata.type === 'test') {
 		this.stats.testCount++;
 
-		if (result.result.metadata.skipped) {
+		if (test.metadata.skipped && !test.metadata.todo) {
```
novemberborn: You still need to increase the `todoCount`, as well as initialize it to `0` in the `run()` method. Then in <https://github.com/sindresorhus/ava/blob/0e1886582833986b896c31859aa61d8da404e996/api.js#L233> you need to compute the total `todoCount`. These counts are mocked in the reporter tests.

novemberborn: Oh and here <https://github.com/sindresorhus/ava/blob/0e1886582833986b896c31859aa61d8da404e996/api.js#L207>.

lib/runner.js
novemberborn: I guess on the one hand todo tests should be skipped, but on the other hand they should not be counted as being skipped. I'm not even sure whether they should be counted as tests now!

Not annotating the test as `skipped` requires explicit guards in the runner where it decides whether to skip a test, but also reduces `skipped && !todo` checks where doing skip counts. Bit of a mixed bag, personally I prefer the more explicit approach.

@vdemedes @sindresorhus?

lib/reporters/tap.js
```diff
@@ -27,6 +27,15 @@ TapReporter.prototype.start = function () {
 TapReporter.prototype.test = function (test) {
 	var output;
 
+	var directive = '';
+	var passed = test.todo ? 'not ok' : 'ok';
+
+	if (test.todo) {
+		directive = '# TODO';
```
vdemedes: If I'm not mistaken, there's no such directive as `# TODO` in TAP spec. For TAP, we could treat `todo` tests just like `skip` and output them the same way.

lib/runner.js
vdemedes: > I'm not even sure whether they should be counted as tests now!

I agree with that. But for the sake of simplicity, we can just mark them as skipped tests internally, to prevent any kind of execution.

lib/reporters/tap.js
novemberborn: [There are](https://testanything.org/tap-version-13-specification.html#todo-tests).

vdemedes: In that case, great! :)

lib/runner.js
BarryThePenguin: If the todo test needs to be explicit, I'm happy to make that change

kentcdodds:
This is awesome! Thanks everyone for your work on this. Looking forward to seeing this merged :-)

sindresorhus:
Still some things to iron out I think.

```js
import test from 'ava';
test.todo('foo');
test.todo('bar');
```

Gives no output with the default reporter:

```
~/dev
❯ ava          



                    
```

And with `--verbose`:

```
~/dev
❯ ava --verbose

  - foo
  - bar

  2 tests failed


                  
```

I don't think todo tests should be failing.

-

```js
import test from 'ava';
test('foo', t => t.pass());
test.todo('bar');
```

Outputs it's passing, but no mention of the `1 todo` and the test run still exit with exit code 1 even though the output indicates differently:

```
~/dev
❯ ava

  1 passed

                  
```

lib/runner.js
novemberborn: Nah there's no clear consensus either way. It's fine as it is, just different trade-offs in different areas.

lib/runner.js
```diff
@@ -63,10 +77,14 @@ optionChain(chainableMethods, function (opts, title, fn) {
...
 Runner.prototype._addTestResult = function (result) {
-	if (result.result.metadata.type === 'test') {
+	var test = result.result;
+
+	if (test.metadata.type === 'test') {
 		this.stats.testCount++;
 
-		if (result.result.metadata.skipped) {
+		if (test.metadata.todo) {
+			this.stats.todoCount++;
```
novemberborn: You need to initialize `todoCount` to `0` in the `Runner#run()` method.

test/runner.js
```diff
@@ -190,9 +190,38 @@ test('skip test', function (t) {
...
+	runner.todo('todo');
+
+	t.throws(function () {
+		runner.todo(function () {});
+	}, {message: '`todo` tests require a title'});
+
 	runner.run().then(function () {
 		t.is(runner.stats.testCount, 2);
 		t.is(runner.stats.passCount, 1);
+		t.is(runner.stats.failCount, 1);
```
novemberborn: Need to add an assertion for the `todoCount`.

novemberborn:
@BarryThePenguin would you mind adding a test in `test/api.js` that verifies the `passCount`, `skipCount`, `todoCount` and `failCount` values? There is no explicit test for at the moment. The reporter tests use mocked values so everything seems fine until it's verified by hand.

BarryThePenguin:
@sindresorhus sure, I'll see if I can work these out

I'm starting to question my reasoning for initially marking todo tests as failed. I've been following the TAP spec and I'm not sure it's quite clear.

>These tests represent a feature to be implemented or a bug to be fixed and act as something of an executable “things to do” list. They are not expected to succeed. Should a todo test point begin succeeding, the harness should report it as a bonus. This indicates that whatever you were supposed to do has been done and you should promote this to a normal test point.

This implies that todo tests fail until they are implemented. This PR currently expects todo tests to have no implementation. Does this mean we mark them as failed? Again, I'm not sure how tightly we want AVA coupled with the TAP spec.

BarryThePenguin:
Also, thanks everyone for providing feedback and spending the time to help out. It's been a really enjoyable experiance :hamster:

sindresorhus:
The TAP description doesn't explicitly say they should fail either, though, just that they shouldn't succeed like a normal test. I think we should rather enable users to fail todo tests on the [linting step](https://github.com/sindresorhus/eslint-plugin-ava/pull/32), not in AVA.

novemberborn:
> I think we should rather enable users to fail todo tests on the linting step, not in AVA. Todo tests should be like TODO comments, informative, but not affecting.

:+1:

BarryThePenguin:
Updated.

@vdemedes @novemberborn todo is no longer skipped as well. Instead, in test-collection we check for the skipped and todo modifier.

@novemberborn I've added the `test/api.js` test

@sindresorhus I'm not quite sure about the exit code being 1, when should the exit code be 1? These are the outputs for the examples you provided above

mini reporter
```

  2 todo

```

verbose reporter
```

  - foo
  - bar

  0 tests passed
  2 tests todo

```

test + todo example

```

  1 passed  1 todo

```

lib/runner.js
BarryThePenguin: Done

sindresorhus:
@BarryThePenguin Perfect. Exactly what I was thinking :)

LGTM

vdemedes:
Looks fine to me as well, thanks @BarryThePenguin!

sindresorhus:
Landed. Thank you for tirelessly working on this one @BarryThePenguin. It turned out really good.

unicorn18

jfmengels:
Thanks @BarryThePenguin!

kentcdodds:
This is terrific. I love the level of thought for quality coming from this project. Thanks everyone (especially @BarryThePenguin)!

PR #566, opened 2016-02-18
Fix babel-register in documentation

### LINKED ISSUE #561: babel-core/register is deprecated
Or will be
`babel-core/register.js`:
```js
// TODO: eventually deprecate this console.trace("use the `babel-register` package instead of `babel-core/register`");
module.exports = require("babel-register");
```

I think better refer to `babel-register`

TrySound:
fix #561


sindresorhus:
Thanks :)

PR #570, opened 2016-02-24
Fix failed line missing from TAP output

### LINKED ISSUE #569: TAP output no longer shows the line that failed
```js
import test from 'ava';

test(t => {
	t.fail();
});
```

```
❯ ava --tap
TAP version 13
# [anonymous]
not ok 1 - Test failed via t.fail()
  ---
    operator: fail
    expected: false
    actual: false
    at: 
  ...

1..1
# tests 1
# pass 0
# fail 1
```

sotojuan:
Fixes #569.

### DIFF
--- a/lib/reporters/tap.js
+++ b/lib/reporters/tap.js
@@ -7,7 +7,7 @@ function getSourceFromStack(stack, index) {
 		.split('\n')
 		.slice(index, index + 1)
 		.join('')
-		.replace(/^\s+at /, '');
+		.replace(/^\s+ /, '');
 }
 
 function TapReporter() {
@@ -35,7 +35,7 @@ TapReporter.prototype.test = function (test) {
 			'    operator: ' + test.error.operator,
 			'    expected: ' + test.error.expected,
 			'    actual: ' + test.error.actual,
-			'    at: ' + getSourceFromStack(test.error.stack, 3),
+			'    at: ' + getSourceFromStack(test.error.stack, 1),
 			'  ...'
 		];
 	} else {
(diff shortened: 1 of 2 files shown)


lib/reporters/tap.js
```diff
@@ -7,7 +7,7 @@ function getSourceFromStack(stack, index) {
 		.split('\n')
 		.slice(index, index + 1)
 		.join('')
-		.replace(/^\s+at /, '');
```
vdemedes: why did "at" have to be removed? Because `stack` here is cleaned up by `stack-utils`?

sotojuan: I think it's because the `test` function [already has](https://github.com/sindresorhus/ava/blob/master/lib/reporters/tap.js#L61) `'at'` in it. Either way, if you add `at` to the Regex the space trimming does't work.

vdemedes: I meant, previously we were removing `at` from the "standard" stack trace:

```js
Error: damn it
  at Test.fn (file.js:4:1)
```

But now the replacing regex does not have it.

vdemedes: And you removed `at` from the stack trace in a test: https://github.com/sindresorhus/ava/pull/570/files#diff-373bdde32276db897f685ce9d49e3b2cR63. But all "standard" stack traces have `at` at the beginning. That's why I was wondering whether we work with an already cleaned up stack here (by `stack-utils`).

vdemedes: Yeah, stack is already cleaned up there, that's why `at` is no longer in it.

vdemedes:
Thank you, @sotojuan, keep the good PRs coming!

PR #571, opened 2016-02-24
Add TypeScript typings

### LINKED ISSUE #568: Add TypeScript types
Would you be interested in adding type definitions to ava? This would allow TypeScript users to easily use ava, since TypeScript can automatically find typings for node modules that have included them. JavaScript programmers that use VSCode would also benefit from better completions.

The type definitions can be added as a single file. You can find more details in the [wiki](https://github.com/Microsoft/TypeScript/wiki/Typings-for-npm-packages).

If you would merge a pull request on this, I can create a PR.

ivogabe:
I've added type definitions for TypeScript users. The exact same code from the readme can be used in TypeScript without type errors. JavaScript programmers, using VSCode, can also get better completions with these typings.

Ping me when you update the API and I will update these typings.

Fixes #568

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -131,7 +131,7 @@ export interface TestContext extends AssertContext {
 	 */
 	plan(count: number): void;
 
-	skip: TestContext;
+	skip: AssertContext;
 }
 export interface CallbackTestContext extends TestContext {
 	/**
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,141 @@
+export interface Observable {
+	subscribe(observer: (value: {}) => void): void;
+}
+
+export type Test = (t: TestContext) => Promise<void> | Iterator<any> | Observable | void;
+export type SerialTest = (t: TestContext) => void;
+export type CallbackTest = (t: CallbackTestContext) => void;
+
+export interface Runner {
+	(name: string, run: Test): void;
+	(run: Test): void;
+	skip: Runner;
+	cb: CallbackRunner;
+}
+export interface SerialRunner {
+	(name: string, run: SerialTest): void;
+	(run: SerialTest): void;
+	skip: SerialRunner;
+}
+export interface CallbackRunner {
+	(name: string, run: CallbackTest): void;
+	(run: CallbackTest): void;
+	skip: CallbackRunner;
+}
+
+export function test(name: string, run: Test): void;
+export function test(run: Test): void;
+export namespace test {
+	export const before: Runner;
+	export const after: Runner;
+	export const beforeEach: Runner;
+	export const afterEach: Runner;
+
+	export const skip: typeof test;
+	export const only: typeof test;
+
+	export function serial(name: string, run: SerialTest): void;
+	export function serial(run: SerialTest): void;
+	export function cb(name: string, run: CallbackTest): void;
+	export function cb(run: CallbackTest): void;
+}
+export namespace test.serial {
+	export const before: SerialRunner;
+	export const after: SerialRunner;
+	export const beforeEach: SerialRunner;
+	export const afterEach: SerialRunner;
+
+	export const skip: typeof test.serial;
+	export const only: typeof test.serial;
+}
+export namespace test.cb {
+	export const before: CallbackRunner;
+	export const after: CallbackRunner;
+	export const beforeEach: CallbackRunner;
+	export const afterEach: CallbackRunner;
+
+	export const skip: typeof test.cb;
+	export const only: typeof test.cb;
+}
+export default test;
+
+export type ErrorValidator
+	= (new (...args: any[]) => any)
+	| RegExp
+	| string
+	| ((error: any) => boolean);
+
+export interface AssertContext {
+	/**
+	 * Passing assertion.
+	 */
+	pass(message?: string): void;
+	/**
+	 * Failing assertion.
+	 */
+	fail(message?: string): void;
+	/**
+	 * Assert that value is truthy.
+	 */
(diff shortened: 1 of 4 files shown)


sotojuan:
This is pretty sweet, thanks! Looks great to me.

sindresorhus:
// @SamVerschueren

readme.md
```diff
@@ -363,6 +363,10 @@ test(t => {
 
 AVA comes with builtin support for ES2015 through [Babel 6](https://babeljs.io). Just write your tests in ES2015. No extra setup needed. You can use any Babel version in your project. We use our own bundled Babel with the [`es2015`](http://babeljs.io/docs/plugins/preset-es2015/) and [`stage-2`](http://babeljs.io/docs/plugins/preset-stage-2/) presets.
 
+### TypeScript support
+
+AVA includes typings for TypeScript. You have to setup transpilation yourself. When you set `module` to `commonjs`, TypeScript will automatically find the type definitions for AVA. You should set `target` to `es2015` to use Promises and async functions.
```
sindresorhus: > When you set `module` to `commonjs`,

Maybe mention where this is set? It can easily be confused as being set in the AVA config.

sindresorhus:
Does this handle `test` being chainable? E.g. `test.skip.cb()`.

sindresorhus:
You also need to explicitly include the file in the `files` field in package.json.

index.js
```diff
@@ -89,3 +89,6 @@ globals.setImmediate(function () {
 });
 
 module.exports = runner.test;
+// TypeScript imports the `default` property for
+// an ES6 default import (`import test from 'ava'`)
```
sindresorhus: ES6 => ES2015

SamVerschueren:
It might be a good idea to add the `d.ts` file to the `package.json` as described [here](https://github.com/Microsoft/TypeScript/wiki/Typings-for-npm-packages). To make sure that the typings are still found when the file changes to something else in the future (`ava.d.ts` for example).

sindresorhus:
@SamVerschueren That won't happen, but even if it did, we would just add it to package.json then. Same reason index.js is not in package.json. It's automatically resolved.

index.d.ts
```diff
@@ -0,0 +1,141 @@
...
+	/**
+	 * Assert that contents matches regex.
+	 */
+	regex(contents: string, regex: RegExp, message?: string): void;
+	/**
+	 * Assert that error is falsy.
+	 */
+	ifError(error: any, message?: string): void;
+}
+export interface TestContext extends AssertContext {
```
SamVerschueren: Not sure why this is extending from `AssertContext`.

screen shot 2016-02-26 at 08 57 56

SamVerschueren:
@sindresorhus fair enough :)

SamVerschueren:
Currently it's not possible to use the `context` object to share context between the `beforeEach`/`afterEach` and the test as describe in [this](https://github.com/sindresorhus/ava#before--after-hooks) section.

But beware! It's only available in `test`, `beforeEach` and `afterEach`. It's not available in `before` and `after`.

SamVerschueren:
@sindresorhus We should write a recipe for using TS to write tests. Will do it when this PR lands.

sindresorhus:
@SamVerschueren :+1:

ivogabe:
> Currently it's not possible to use the context object to share context between the beforeEach/afterEach and the test as describe in this section.
> 
> But beware! It's only available in test, beforeEach and afterEach. It's not available in before and after.

What do you mean exactly? Should `context` only be available in `test`, `beforeEach` and `afterEach`?

ivogabe:
`TestContext` extends `AssertContext`, so you cannot write `t.skip.skip` or `t.skip.plan` (see b6bc16790db18362ae45db3afab0b8a4caf8543b)

ivogabe:
@sindresorhus I've addressed your comments. Chaining works because of the `namespace` declarations.

SamVerschueren:
@ivogabe Didn't had my morning coffee when I mentioned to `extends AssertContext`. Realised it works as expected :).

ivogabe:
@SamVerschueren Because of your comment I realized that `skip` had the wrong type :)

package.json
```diff
@@ -33,7 +33,8 @@
   },
   "files": [
     "lib",
-    "*.js"
+    "*.js",
+	"index.d.ts"
```
sindresorhus: Wrong indent.

SamVerschueren:
@ivogabe then my comment was good for something after all :).

index.js
```diff
@@ -89,3 +89,6 @@ globals.setImmediate(function () {
 });
 
 module.exports = runner.test;
+// TypeScript imports the `default` property for
+// an ES2015 default import (`import test from 'ava'`)
```
novemberborn: Might be useful to link to a background explanation for this. I found <https://github.com/Microsoft/TypeScript/issues/2242#issuecomment-83694181>, not sure if there's official documentation for this behavior elsewhere?

tl;dr TypeScript's behavior doesn't seem without merit, and explicitly exporting a `default` here should be okay.

ivogabe: I think that's the best explanation of it. I've added it in the comment.

sindresorhus:
Looks good. Thank you @ivogabe :)

SamVerschueren:
Think the `context` property isn't supported yet. Hard to see on mobile though :).

ivogabe:
@sindresorhus Thanks for merging!
@SamVerschueren You're right about that, will send another PR.

PR #573, opened 2016-02-25
Added support for babel config under package.json/ava key (Fixes #448)

### LINKED ISSUE #448: A possible way to provide customizable babel configs using `.babelrc`/`env`
Continued from:
https://github.com/sindresorhus/ava/pull/398#issuecomment-172386983

My proposal is that we allow the following in `.babelrc`:

```js
{
  "presets": ["es2015"], // base config
  "sourceMaps": "inline",
  "env": {
    "ava" : {  // if `env/ava` is present, we disable our custom config.
      "presets": ["stage-0"],
      "plugins": [] //maybe we automatically append power-assert for them?
    },
    "production" : {
     "sourceMaps": false
    }
  }
}
```

There are a couple questions that need to be answered in order to proceed. Specifically, can we change which `env` Babel uses on a per file basis? It might be as easy as setting `process.env.BABEL_ENV` before each call to `babel.transform(..)`

spudly:
Here's an initial commit that allows package owners to specify a babel config for tests under the package.json/ava key.

This conforms to the proposal in the following comment: https://github.com/sindresorhus/ava/issues/448#issuecomment-187694472

I chose not to implement an extends key discussed in the thread because 1) it adds to the complexity, 2) I personally don't need it, and 3) how to solve that request is still somewhat unresolved. If needed, someone can add that later in another pull request.

Let me know what you think. I'm a little unsure about whether my method for obtaining babelConfig is the best one.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -35,7 +35,11 @@ var Api = require('./api');
 // Bluebird specific
 Promise.longStackTraces();
 
-var conf = pkgConf.sync('ava');
+var conf = pkgConf.sync('ava', {
+	defaults: {
+		babel: 'default'
+	}
+});
 
 var cli = meow([
 	'Usage',
--- a/cli.js
+++ b/cli.js
@@ -102,7 +102,8 @@ var api = new Api({
 	require: arrify(cli.flags.require),
 	cacheEnabled: cli.flags.cache !== false,
 	explicitTitles: cli.flags.watch,
-	match: arrify(cli.flags.match)
+	match: arrify(cli.flags.match),
+	babelConfig: conf.babel
 });
 
 var reporter;
--- a/lib/caching-precompiler.js
+++ b/lib/caching-precompiler.js
@@ -31,26 +31,27 @@ CachingPrecompiler.prototype._factory = function (babelConfig, cacheDir) {
 		// Extract existing source maps from the code.
 		var sourceMap = convertSourceMap.fromSource(code) || convertSourceMap.fromMapFileSource(code, path.dirname(filename));
 
-		var baseOptions = {
+		var options = {babelrc: false};
+
+		if (babelConfig === 'default') {
+			objectAssign(options, {presets: [presetStage2, presetES2015]});
+		} else if (babelConfig === 'inherit') {
+			objectAssign(options, {babelrc: true});
+		} else {
+			objectAssign(options, babelConfig);
+		}
+
+		objectAssign(options, {
+			inputSourceMap: sourceMap && sourceMap.toObject(),
 			filename: filename,
 			sourceMaps: true,
-			ast: false,
-			inputSourceMap: sourceMap && sourceMap.toObject(),
-			babelrc: false
-		};
+			ast: false
+		});
 
-		if (!babelConfig || babelConfig === 'default') {
-			return objectAssign({}, baseOptions, {
-				presets: [presetStage2, presetES2015],
-				plugins: [powerAssert, transformRuntime]
-			});
-		}
+		options.plugins = options.plugins || [];
+		options.plugins.push(powerAssert, transformRuntime);
 
-		if (babelConfig === 'inherit') {
-			return objectAssign({}, baseOptions, {babelrc: true});
-		}
-
-		return objectAssign({}, baseOptions, babelConfig);
+		return options;
 	}
 
 	return function (code, filename, hash) {
--- a/lib/caching-precompiler.js
+++ b/lib/caching-precompiler.js
@@ -33,7 +33,7 @@ CachingPrecompiler.prototype._factory = function (babelConfig, cacheDir) {
 
 		var options = {babelrc: false};
 
-		if (babelConfig === 'default') {
+		if (!babelConfig || babelConfig === 'default') {
 			objectAssign(options, {presets: [presetStage2, presetES2015]});
 		} else if (babelConfig === 'inherit') {
 			objectAssign(options, {babelrc: true});
--- a/lib/caching-precompiler.js
+++ b/lib/caching-precompiler.js
@@ -35,12 +35,12 @@ CachingPrecompiler.prototype._factory = function (babelConfig, cacheDir) {
 			filename: filename,
 			sourceMaps: true,
 			ast: false,
-			inputSourceMap: sourceMap && sourceMap.toObject()
+			inputSourceMap: sourceMap && sourceMap.toObject(),
+			babelrc: false
 		};
 
 		if (!babelConfig || babelConfig === 'default') {
 			return objectAssign({}, baseOptions, {
-				babelrc: false,
 				presets: [presetStage2, presetES2015],
 				plugins: [powerAssert, transformRuntime]
 			});
--- a/lib/caching-precompiler.js
+++ b/lib/caching-precompiler.js
@@ -81,7 +81,8 @@ CachingPrecompiler.prototype._createTransform = function (babelConfig) {
 		salt: new Buffer(JSON.stringify({
 			'babel-plugin-espower': require('babel-plugin-espower/package.json').version,
 			'ava': require('../package.json').version,
-			'babel-core': require('babel-core/package.json').version
+			'babel-core': require('babel-core/package.json').version,
+			'babelConfig': babelConfig
 		})),
 		ext: '.js',
 		hash: this._hash.bind(this)
--- a/lib/caching-precompiler.js
+++ b/lib/caching-precompiler.js
@@ -3,20 +3,21 @@ var path = require('path');
 var cachingTransform = require('caching-transform');
 var md5Hex = require('md5-hex');
 var stripBom = require('strip-bom');
+var objectAssign = require('object-assign');
 
 module.exports = CachingPrecompiler;
 
-function CachingPrecompiler(cacheDir) {
+function CachingPrecompiler(cacheDir, babelConfig) {
 	if (!(this instanceof CachingPrecompiler)) {
 		throw new TypeError('Class constructor CachingPrecompiler cannot be invoked without \'new\'');
 	}
 
 	this.cacheDir = cacheDir;
 	this.filenameToHash = {};
-	this.transform = this._createTransform();
+	this.transform = this._createTransform(babelConfig);
 }
 
-CachingPrecompiler.prototype._factory = function (cacheDir) {
+CachingPrecompiler.prototype._factory = function (babelConfig, cacheDir) {
 	// This factory method is only called once per process, and only as needed, to defer loading expensive dependencies.
 	var babel = require('babel-core');
 	var convertSourceMap = require('convert-source-map');
@@ -30,15 +31,26 @@ CachingPrecompiler.prototype._factory = function (cacheDir) {
 		// Extract existing source maps from the code.
 		var sourceMap = convertSourceMap.fromSource(code) || convertSourceMap.fromMapFileSource(code, path.dirname(filename));
 
-		return {
-			presets: [presetStage2, presetES2015],
-			plugins: [powerAssert, transformRuntime],
+		var baseOptions = {
 			filename: filename,
 			sourceMaps: true,
 			ast: false,
-			babelrc: false,
 			inputSourceMap: sourceMap && sourceMap.toObject()
 		};
+
+		if (!babelConfig || babelConfig === 'default') {
+			return objectAssign({}, baseOptions, {
+				babelrc: false,
+				presets: [presetStage2, presetES2015],
+				plugins: [powerAssert, transformRuntime]
+			});
+		}
+
+		if (babelConfig === 'inherit') {
+			return objectAssign({}, baseOptions, {babelrc: true});
+		}
+
+		return objectAssign({}, baseOptions, babelConfig);
 	}
 
 	return function (code, filename, hash) {
@@ -61,9 +73,9 @@ CachingPrecompiler.prototype._createEspowerPlugin = function (babel) {
 	});
 };
 
-CachingPrecompiler.prototype._createTransform = function () {
+CachingPrecompiler.prototype._createTransform = function (babelConfig) {
 	return cachingTransform({
-		factory: this._factory.bind(this),
+		factory: this._factory.bind(this, babelConfig),
 		cacheDir: this.cacheDir,
 		salt: new Buffer(JSON.stringify({
 			'babel-plugin-espower': require('babel-plugin-espower/package.json').version,
--- a/package.json
+++ b/package.json
@@ -132,6 +132,7 @@
     "update-notifier": "^0.6.0"
   },
   "devDependencies": {
+    "babel-cli": "^6.6.4",
     "cli-table2": "^0.1.9",
     "coveralls": "^2.11.4",
     "delay": "^1.3.0",
--- a/readme.md
+++ b/readme.md
@@ -154,7 +154,8 @@ All of the CLI options can be configured in the `ava` section of your `package.j
     "tap": true,
     "require": [
       "babel-register"
-    ]
+    ],
+    "babel": "inherit"
   }
 }
 ```
@@ -422,6 +423,41 @@ AVA comes with builtin support for ES2015 through [Babel 6](https://babeljs.io).
 ### TypeScript support
 
 AVA includes typings for TypeScript. You have to setup transpilation yourself. When you set `module` to `commonjs` in your `tsconfig.json` file, TypeScript will automatically find the type definitions for AVA. You should set `target` to `es2015` to use Promises and async functions.
+
+If you want to customize the babel transpiler for test files, you can do so by adding a `"babel"` key to the `"ava"` section in your package.json file.
+
+```json
+{
+	"ava": {
+		 "babel": {
+			 "presets": [
+					"es2015",
+					"stage-0",
+					"react"
+			 ]
+		 }
+	},
+}
+```
+
+In addition to specifying a custom babel config, you can also use the special `inherit` keyword. When you do this, AVA will allow tests to be transpiled using the configuration defined in your .babelrc file or in package.json/babel. This way, your test files will be transpiled using the same options as your source files, but you won't have to define the options twice.
+
+```json
+{
+	"babel": {
+		"presets": [
+			"es2015",
+			"stage-0",
+			"react"
+		]
+	},
+	"ava": {
+		"babel": "inherit",
+	},
+}
+```
+
+If you do not specify a "babel" key in your ava configuration, or if you set it to `"default"`, AVA will transpile the test files with AVA's default babel configuration.
 
 #### Transpiling Imported Modules
 
--- a/readme.md
+++ b/readme.md
@@ -424,7 +424,9 @@ AVA comes with builtin support for ES2015 through [Babel 6](https://babeljs.io).
 
 AVA includes typings for TypeScript. You have to setup transpilation yourself. When you set `module` to `commonjs` in your `tsconfig.json` file, TypeScript will automatically find the type definitions for AVA. You should set `target` to `es2015` to use Promises and async functions.
 
-If you want to customize the babel transpiler for test files, you can do so by adding a `"babel"` key to the `"ava"` section in your package.json file.
+### Babel Configuration for Test Scripts
+
+If you want to customize the babel transpiler for test files, you can do so by adding a `"babel"` key to the `ava` section in your `package.json` file.
 
 ```json
 {
@@ -440,7 +442,7 @@ If you want to customize the babel transpiler for test files, you can do so by a
 }
 ```
 
-In addition to specifying a custom babel config, you can also use the special `inherit` keyword. When you do this, AVA will allow tests to be transpiled using the configuration defined in your .babelrc file or in package.json/babel. This way, your test files will be transpiled using the same options as your source files, but you won't have to define the options twice.
+In addition to specifying a custom Babel config, you can also use the special `"inherit"` keyword. When you do this, AVA will allow tests to be transpiled using the configuration defined in your `.babelrc` file or in package.json/babel. This way, your test files will be transpiled using the same options as your source files, but you won't have to define the options twice.
 
 ```json
 {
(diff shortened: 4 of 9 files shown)


cli.js
```diff
@@ -97,7 +97,8 @@ var api = new Api(cli.input.length ? cli.input : arrify(conf.files), {
 	failFast: cli.flags.failFast,
 	serial: cli.flags.serial,
 	require: arrify(cli.flags.require),
-	cacheEnabled: cli.flags.cache !== false
+	cacheEnabled: cli.flags.cache !== false,
+  babelConfig: conf.babel,
```
sindresorhus: Wrong indentation here.

sindresorhus:
Good start :) This will need tests and docs too.

spudly:
Fixed indentation, added tests, and added docs.

readme.md
```diff
@@ -363,6 +364,41 @@ test(t => {
...
+			"react"
+		]
+	},
+	"ava": {
+		"babel": "inherit",
+	},
+}
+```
+
+If you do not specify a "babel" key in your ava configuration, or if you set it to `"default"`, AVA will transpile the test files with AVA's default babel configuration.
```
kentcdodds: Is this default configuration defined anywhere else in the docs? If not, it should be. Then this should link to that.

vdemedes:
Hey @spudly, could you rebase from master to resolve a conflict? Thanks for the great work!

spudly:
@vdemedes

It's been rebased.

vdemedes:
Great, I think it's good to go!

@sindresorhus @novemberborn @jamestalmage ?

lib/caching-precompiler.js
```diff
@@ -30,15 +31,26 @@ CachingPrecompiler.prototype._factory = function (cacheDir) {
...
 			sourceMaps: true,
 			ast: false,
-			babelrc: false,
-			inputSourceMap: sourceMap && sourceMap.toObject()
+			inputSourceMap: sourceMap && sourceMap.toObject(),
+			babelrc: false
 		};
+
+		if (!babelConfig || babelConfig === 'default') {
+			return objectAssign({}, baseOptions, {
```
novemberborn: No need for the empty object here, `baseOptions` is already fresh so it's safe to mix other properties into it.

readme.md
```diff
@@ -367,6 +368,41 @@ AVA comes with builtin support for ES2015 through [Babel 6](https://babeljs.io).
...
+					"es2015",
+					"stage-0",
+					"react"
+			 ]
+		 }
+	},
+}
+```
+
+In addition to specifying a custom babel config, you can also use the special `inherit` keyword. When you do this, AVA will allow tests to be transpiled using the configuration defined in your .babelrc file or in package.json/babel. This way, your test files will be transpiled using the same options as your source files, but you won't have to define the options twice.
```
novemberborn: `"inherit"` to be consistent with `"default"` below?

readme.md
```diff
@@ -367,6 +368,41 @@ AVA comes with builtin support for ES2015 through [Babel 6](https://babeljs.io).
 
 AVA includes typings for TypeScript. You have to setup transpilation yourself. When you set `module` to `commonjs` in your `tsconfig.json` file, TypeScript will automatically find the type definitions for AVA. You should set `target` to `es2015` to use Promises and async functions.
 
+If you want to customize the babel transpiler for test files, you can do so by adding a `"babel"` key to the `"ava"` section in your package.json file.
```
novemberborn: Some nits:

* This needs a heading.
* The README isn't entirely consistent but I'd prefer `package.json` (with the backticks). Same for `.babelrc`.
* It's named "`ava` section" (without quotes) earlier: https://github.com/spudly/ava/blob/f928959b9ad106a70fc3091a462ff3d96a60f469/readme.md#configuration
* Could you capitalize Babel throughout?

novemberborn:
Haven't looked at the tests yet, sorry. Been staring at AVA issues all day :)

Could you follow up on https://github.com/sindresorhus/ava/issues/448#issuecomment-190286370?

Looking really good though, nice work.

lib/caching-precompiler.js
```diff
@@ -30,15 +31,26 @@ CachingPrecompiler.prototype._factory = function (cacheDir) {
...
+				presets: [presetStage2, presetES2015],
+				plugins: [powerAssert, transformRuntime]
+			});
+		}
+
+		if (babelConfig === 'inherit') {
+			return objectAssign({}, baseOptions, {babelrc: true});
```
jamestalmage: As discussed, we will want to ensure `powerAssert` is added as a plugin regardless.

readme.md
jamestalmage: :+1: I don't believe it is defined anywhere, but it should be.

lib/caching-precompiler.js
```diff
@@ -30,15 +31,26 @@ CachingPrecompiler.prototype._factory = function (cacheDir) {
...
-			presets: [presetStage2, presetES2015],
-			plugins: [powerAssert, transformRuntime],
+		var baseOptions = {
 			filename: filename,
 			sourceMaps: true,
 			ast: false,
-			babelrc: false,
-			inputSourceMap: sourceMap && sourceMap.toObject()
+			inputSourceMap: sourceMap && sourceMap.toObject(),
+			babelrc: false
```
novemberborn: Unnecessary diff churn

test/caching-precompiler.js
```diff
@@ -18,9 +20,14 @@ function endsWithMap(filename) {
 	return /\.js$/.test(filename);
 }
 
+test('before', function (t) {
+	sinon.spy(babel, 'transform');
```
novemberborn: AVA's tests are run using [`tap`](http://www.node-tap.org/) which runs each test file in its own process. You don't have to worry about modifying the `babel-core` module, or about cleaning it up afterwards. Just add the spy in the module itself, outside of a test.

novemberborn:
The tests exercise `caching-compiler`, which is good. I'd like to see some integration tests in `test/cli`, though that does not have to hold up this PR. Happy to undertake that myself or help you @spudly if you're interested. It might be a bit tricky to verify the differing Babel behaviors.

---

Right now `package.json/ava/babel` allows any Babel option to be changed. To avoid footguns though I think `filename`, `sourceMaps`, `ast` and `inputSourceMap` should not be configurable.

---

@spudly thanks for the hard work on this. It's solving so many of our issues!

spudly:
Pushed a new commit & rebased. I believe I've addressed all of the concerns that were discussed, with the exception of the `babel: false` option, which I believe should be addressed in a separate pull request.

spudly:
Can anyone help explain why these 44 tests are failing?

https://travis-ci.org/sindresorhus/ava/jobs/112878667

They're unrelated to my changes - the master branch of AVA fails the same tests. These tests were passing yesterday, but today they fail. Perhaps one of the dependencies had a release that broke with semver?

wyze:
Might be related to babel 6.6 [release](https://github.com/babel/babel/releases/tag/v6.6.0) yesterday.

I noticed that your previous build used 6.5.2 while this one used 6.6.

test/caching-precompiler.js
```diff
@@ -18,9 +20,11 @@ function endsWithMap(filename) {
 	return /\.js$/.test(filename);
 }
 
+sinon.spy(babel, 'transform');
+
 test('creation with new', function (t) {
 	var tempDir = uniqueTempDir();
-	var precompiler = new CachingPrecompiler(tempDir);
+	var precompiler = new CachingPrecompiler(tempDir, null);
```
novemberborn: Should pass `'default'` here (and below).

MoOx:
Failures may be similar to https://github.com/MoOx/statinamic/issues/229 (or may not)

novemberborn:
@spudly did a fresh checkout and install, `master` still passes but this branch does not.

novemberborn:
@spudly ah! The failing tests use `api.js` directly, without setting the `babelConfig` option. Now I feel bad :cry: 

I still think using the default config in `cli.js` is the right move, but maybe add the `!babelConfig || babelConfig === 'default'` clause back in `caching-precompiler.js`. Quite bothersome passing it in the 44 tests that instantiate `Api`.

spudly:
My bad, I must have made a mistake when I checked out master and ran the tests again yesterday. Did it again this morning and the tests are passing. I'll fix this very soon.

spudly:
I'm stumped guys. If I do the following, all tests pass as expected.

```bash
git clone https://github.com/sindresorhus/ava.git
cd ava
git reset --hard 58e1c561296a409a9557ceb1c7684685404c4278
npm install
npm test
```

But if I do this, 44 tests fail:

```bash
git clone https://github.com/spudly/ava.git
cd ava
git reset --hard 58e1c561296a409a9557ceb1c7684685404c4278
npm install
npm test
```

Both repos are being reset to the same initial commit (the one before any of my changes), so there should be absolutely no difference between the two, right?

kentcdodds:
I'm running both of those series of commands now. I'll let you know what I get

kentcdodds:
Everything passed for me for both of those. Sounds like an environment issue...

kentcdodds:
But yes, this fails 44 tests for me:

```
git clone https://github.com/spudly/ava.git
cd ava
npm install
npm test
```

spudly:
my current theory is that I'm getting weird test results because of nyc cache. testing that theory now...

MoOx:
Maybe try BABEL_DISABLE_CACHE=1.

spudly:
@moOx,

I didn't know about that flag, but the following just worked for me:

1. `git clone`...
2. `cd ava`
4. `npm install`
3. [make edits]
5. `npm test`

if before step 3 I ran `npm test`, the test at step 5 would fail with 44 errors even though the source code was fixed.

Not sure at this point if it was nyc cache or babel cache but it was definitely caching the source files somewhere and screwing everything up.

novemberborn:
@spudly,

> if before step 3 I ran npm test, the test at step 5 would fail with 44 errors even though the source code was fixed.

Do you have `xo`, `nyc` and/or `tap` installed globally? Else `npm test` wouldn't even be able to cache anything. I wonder if there's some local/global mismatch, presumably in `nyc`. Which `nyc` version do you have globally?

---

Is this ready for another look?

spudly:
@novemberborn, I don't have any of those modules installed globally.

Yes, the pull request is ready.

novemberborn:
> I don't have any of those modules installed globally.

Then I'm stumped, sorry :disappointed:

test/caching-precompiler.js
```diff
@@ -48,3 +52,64 @@ test('adds files and source maps to the cache directory as needed', function (t)
...
+	t.true(babel.transform.calledOnce);
+	var options = babel.transform.firstCall.args[1];
+
+	t.true('filename' in options);
+	t.true(options.sourceMaps);
+	t.false(options.ast);
+	t.true('inputSourceMap' in options);
+	t.false(options.babelrc);
+	t.deepEqual(options.presets, ['stage-2', 'es2015']);
+	t.equal(options.plugins.length, 2);
```
novemberborn: Could you add a configured plugin on line 98, and then verify not the `plugins` length but the expected values? Would be good to test that the configured plugins are preserved and the power-assert and runtime transform plugins are appended.

Could then change the `plugins` test in the default case to match for those plugins too. Similarly the inherit case should check the plugins were added.

novemberborn:
@spudly did I mention I'm pretty excited about this landing? I'm pretty excited about this landing :)

Just one suggestion regarding testing the plugins.

Oh, and to confirm, Babel merges plugin definitions when you set `babelrc` to `true`?

test/caching-precompiler.js
spudly: I can certainly add a fake plugin, which would increase the expected plugins.length to 3. I can also verify that transformRuntime is added because I can just require() it in to get a reference to it and use === for comparison.

But I'm not sure how I can verify that powerAssert was added because it is instantiated inside of the `_factory` function, so I don't have a reference to it. That's why I only verified that lenth === 2.

Is there another way I can validate that the powerAssert plugin was added?

jamestalmage: > Is there another way I can validate that the powerAssert plugin was added?

Run a test with a bad assertion that generates powerAssert output. There is an existing test that already does that using the default config. You would just need to create a similar one that does it using custom config.

jamestalmage:
> @spudly did I mention I'm pretty excited about this landing? I'm pretty excited about this landing :)

Agreed. This should be a high priority merge. Lots of people want this feature.

kentcdodds:
> Lots of people want this feature.

[Anyone wanting to use JSX](https://github.com/kentcdodds/react-ava-workshop/blob/start/INSTRUCTIONS.md#ava) in their test files is still using `ava@0.9.2` until this lands. So thank you so much everyone for working on it and making sure that the API is awesome! :tada:

jamestalmage:
> Anyone wanting to use JSX in their test files is still using ava@0.9.2 until this lands

Sorry about that. When we made the decision to ignore `.babelrc`, it wasn't supposed to take this long to get it back in.

kentcdodds:
Heh, I know that. That's open source. If I really really needed to get this in, I could have taken a more proactive approach to making that happen. #priorities ;-) I'm just glad that there's a shared priority here!

test/caching-precompiler.js
novemberborn: Try this, after you instantiate `new CachingPrecompiler()`

```js
var powerAssert = function () {};
sinon.stub(precompiler, '_createEspowerPlugin').returns(powerAssert);
```

And then compare against the `powerAssert` variable inside your test.

spudly:
Improved unit tests to check that powerAssert and transformRuntime get appended properly.

Ready for evaluation again.

> Oh, and to confirm, Babel merges plugin definitions when you set babelrc to true?

Yes, I believe that is the case.

novemberborn:
LGTM! Great work @spudly.

jamestalmage:
>> Oh, and to confirm, Babel merges plugin definitions when you set babelrc to true?

> Yes, I believe that is the case.

Should we have an integration test confirming that?

Otherwise, lets ship this.

readme.md
```diff
@@ -367,6 +368,61 @@ AVA comes with builtin support for ES2015 through [Babel 6](https://babeljs.io).
...
+			"react"
+		]
+	},
+	"ava": {
+		"babel": "inherit",
+	},
+}
+```
+
+Note: When configuring Babel for tests manually, the espower and transform-runtime plugins will be
```
jamestalmage: Why `transform-runtime`?

kentcdodds:
FWIW, I just did this:

```
$ npm i spudly/ava#461c781fcaabe29eb29464c3d98f9e19b64a8af2
$ npm test
```

And all tests pass. Verified I had the right version

```
$ node_modules/.bin/ava --version
```

And got `0.12.0` which I think is correct.

The thing that surprised me is that I didn't need to do any configuration to make my tests work with JSX. I was under the impression I had to explicitly add this config:

```javascript
"ava": {
  "babel": "inherit"
}
```

But it looks like my tests were transpiled with my default `.babelrc` anyway. Is that correct?

Either way, it's awesome that I'll be able to use the latest AVA soon! :D

jamestalmage:
> I was under the impression I had to explicitly add this config:

Yes - you should definitely be required to explicitly configure the "inherit" property. 

@spudly - when you debug that, please be sure to add a few integration tests to verify correct behavior. Specifically, that `.babelrc` is only consulted for `inherit`.

jamestalmage:
@spudly - see https://github.com/sindresorhus/ava/commit/1b01f16d9bcd6f5d62e3becb8d60e457e7ac9690

You should be able to merge the `babel-plugin-for-integration-tests` branch into this PR without conflict.

To test out what the babel plugin does:

```sh
$ node_modules/.bin/babel --plugins ./babel-plugin-test-doubler test/fixture/fail-fast.js
```

It should be pretty easy to create integration tests from there using the API. (You should get double the number of tests when the plugin is used).

readme.md
spudly: If we don't include transform-runtime, users will have to `import 'babel-polyfill'` at the top of each test file. This has not changed in the pull request. It was like this before. This is only a documentation change.

spudly:
Apparently when ava compiles a file it caches the result (hence the name of the file, `cachingPrecompiler.js`). This cache is stored in `node_modules/.cache/ava`.

I'm pretty sure this is the issue I was having during a couple of days ago. Simply removing ./node_modules/.cache probably would have solved my problem.

@jamestalmage, is there an easy way to programmatically clear AVA's compile cache (or disable it)? I need to be able to do this so I can isolate the unit tests from each other. The issue is that if I reuse a test fixture (using `test/fixtures/es2015.js` right now), it will compile with one babel config and cache the result. Then, later, when another test is run with a different babel config, the test fails because it uses a cached version.

novemberborn:
> I'm pretty sure this is the issue I was having during a couple of days ago. Simply removing ./node_modules/.cache probably would have solved my problem.

Oh of course!

@spudly you can use the `--no-cache` flag to disable the cache.

spudly:
@kentcdodds, still looking at your issue, but is it possible that your issue is related to the compiler cache as well? I'm thinking maybe you added the react to the list of plugins, ran the compiler (cache files get created here), then removed it and expected it to fail but it worked anyway because it used the cached version the second time. Can you remove `node_modules/.cache` and try again?

novemberborn:
@spudly we'll have to include the `babelConfig` option in the salt for the caching transform, see <https://github.com/sindresorhus/ava/blob/e697629a77b511e069eb0a2991c8ee72cd09034c/lib/caching-precompiler.js#L68>.

Also there's a merge conflict now :(

spudly:
yep, fixing the merge conflict...

kentcdodds:
Success! It probably was the cache. Here's a reproducible command to show old/new behavior.

```
git clone https://github.com/kentcdodds/react-ava-workshop.git && cd react-ava-workshop && npm i && npm i spudly/ava#461c781fcaabe29eb29464c3d98f9e19b64a8af2 && npm t
```

You'll get a parser error.

Now, to the `package.json` add:

```js
{
  "ava": {
    "babel": "inherit"
  }
}
```

And rerun `npm t` and you'll get 10 passing tests :+1:

spudly:
Added babelConfig to the salt and re-enabled caching in unit tests. All working.

Should be ready to merge now, right? (assuming tests pass in travis && appveyor)

kentcdodds:
That's me, watching the the status of the AppVeyor build...

kentcdodds:
Me now that it's all passing!

spudly:
only took 22 minutes... is there a reason we're running tests in both appveyor and travis? Seems like one is enough

kentcdodds:
> is there a reason we're running tests in both appveyor and travis? Seems like one is enough

Gotta make sure windows works (right?)

kentcdodds:
> only took 22 minutes...

Yeah, it only took that to actually run, but it took much longer to start. Believe me, I've been watching :-)

spudly:
> t.threw??

This gives me a better error if the test fails. If the promise in that test is rejected, it will call `t.threw(error)` and the tests will fail with that error. If I didn't include `t.threw` to handle the rejection, the reason for tests failure would just be that there was a missing test, which is not enough information for debugging. This should probably be added to a lot of the other tests as well.

novemberborn:
> If I didn't include t.threw to handle the rejection, the reason for tests failure would just be that there was a missing test, which is not enough information for debugging. This should probably be added to a lot of the other tests as well.

AVA now requires a version of `tap` which supports promises. You should be able to return the promise and `tap` will report the failure.

package.json
```diff
@@ -131,6 +131,7 @@
     "update-notifier": "^0.6.0"
   },
   "devDependencies": {
+    "babel-cli": "^6.6.4",
```
novemberborn: Why this dependency?

novemberborn:
@spudly could you return the `api.run()` promise and remove the `t.threw` bit? And just a question about the `babel-cli` dependency. LGTM otherwise!

---

When squashing would be good to keep @jamestalmage's commit separate.

novemberborn:

spudly:
Will do. I'll make these changes either tonight or Monday morning.

spudly:
Is there a reason we have a `return` statement on [cli.js:13](https://github.com/sindresorhus/ava/blob/master/cli.js#L13)? It's causing lint errors with the latest XO config.

spudly:
Nevermind. running npm install fixed that issue. Still curious though, why are those return statements there?

spudly:
made requested changes and rebased again.

novemberborn:
> Is there a reason we have return statements on cli.js:13 and cli.js:96?

They're shorthand for exiting the CLI.

novemberborn:
@spudly what's the reason for adding the `babel-cli` dependency in `package.json`?

spudly:
@novemberborn,

That dependency was added by @jamestalmage in the `babel-plugin-for-integration-tests` branch, which is merged into my code.

vdemedes:
I guess "outside" commits shouldn't be in this PR, should they?

novemberborn:
> That dependency was added by @jamestalmage in the babel-plugin-for-integration-tests branch, which is merged into my code.

Ah!

@vdemedes are you in Gitter at the moment? Let's see about landing the integration test plugin in `master` etc.

novemberborn:
Landed in https://github.com/sindresorhus/ava/commit/a03f8264edbd99d9c6a86827106ad40dfdc42b10 yay! (Messed up the commit so it didn't close automatically… oops)

novemberborn:
Awesome work, thanks @spudly!

spudly:
Nice! Thanks for the help everyone!

sindresorhus:
Yay! Nice work @spudly :dancer: 

od3vvoq

kentcdodds:
Hooray! 🎉🎉🎉

Wildhoney:
:clap: :clap:

For those wondering, it appears as though this hasn't *yet* been published to npm &mdash; however a `npm i https://github.com/sindresorhus/ava` will suffice for the moment.

vdemedes:
giphy 7

tleunen:
Can we get a cli option for that as well? I prefer specifying it there instead of adding a config in the package.json

jamestalmage:
@tleunen

Please raise an issue.

Issue #493, opened 2016-01-30
t.throws should return the error (or a promise that resolves with the rejection)


jamestalmage:
From #492 
 
> 
```js
test(t => {
  t.plan(2);
  return shouldRejectWithFoo().catch(reason => {
    t.is(reason.message, 'HelloMessage') // Prefer t.throws if all you care about is the message
    t.is(reason.foo, 'bar');
  });
});

> test(t => {
  t.plan(2);
  try {
    shouldThrow();
  } catch (err) {
    t.is(err.message, 'HelloMessage') // Prefer t.throws if all you care about is the message
    t.is(err.foo, 'bar');
  }
});
```
> In most cases, you should prefer the `t.throws()` assertion, but this is an acceptable use since `t.throws()` only allows you to assert against the error's `message` property.

We could make that recommendation obsolete:

```js
test(async t => {
  var err = await t.throws(shouldRejectWithFoo, 'HelloMessage');
  t.is(err.foo, 'bar');
});

test(t => {
  var err = t.throws(shouldThrow, 'HelloMessage');
  t.is(err.foo, 'bar');
});
```

See also: https://github.com/sindresorhus/ava/issues/468 (it also changes the behavior of `t.throws`).

kasperlewau:
Been putting in a little bit of work into this, and I just wanted to get some clarification on the intent before finishing up and posting a PR; 

```js
test(t => {
  const err = t.throws(function () { throw new Error('foo'); });
  t.is(err.message, 'foo');
});
```

```js
test(t => {
  const err = await t.throws(function () { return Promise.reject(new Error('foo')); });
  t.is(err.message, 'foo');
});
```

Regardless of what one passes as the second/third argument to `t.throws`, you **always** want to get back the (or in the case of `string`, **an**) Error with the expected message. Is that a correct assessment or am I (miles) off? 

I must admit, I'm a little confused by your example - are you expecting a property on the given Error object of `foo` that should equal the expected message? 

--- 

I haven't taken #468 into consideration just yet - but I'd like to get cracking on that soon-*ish* as well. Babysteps.

novemberborn:
@kasperisager does this help?

```js
test(t => {
  const expected = new Error()
  const actual = t.throws(() => { throw expected })
  t.is(actual, expected)
})
```

```js
test(t => {
  const expected = new Error()
  const actual = t.throws(() => Promise.reject(expected)
  t.is(actual, expected)
})
```

[...]

PR #576, opened 2016-02-29
Return thrown error/rejected promise from t.throws

kasperlewau:
If a `throws` assertion passes, we should return the reason for said pass. That entails; 

* A thrown Error.
* A Promise rejection.
* An Observable error. 

Loosely based on the description of #493. I consider it to be loosely based on it, because of the following differences (in the Promised-land): 

```js
// Desired
test(t => {
  const expected = new Error('foo');
  const actual = t.throws(() => Promise.reject(expected));
  t.is(expected, actual); // true
})

// Implemented
test(t => {
  const expected = new Error('foo');
  const actual = t.throws(() => Promise.reject(expected));
  t.is(expected, actual.reason); //true 
})
```

--- 

I hijacked the `result.reason` property because I wasn't fully comfortable in excluding the base `result` properties out of a successful `throws` assertion, which was heavily used in `test/promise` and `test/observable`. However this can be changed quite easily, to more closely reflect the initial desires of #493. 

--- 

Haven't taken a stab at docs for the changes made yet, I'll get on top of that on my way home from work.

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -78,7 +78,7 @@ x.throws = function (fn, err, msg) {
 			.then(function () {
 				x.throws(noop, err, msg);
 			}, function (fnErr) {
-				x.throws(function () {
+				return x.throws(function () {
 					throw fnErr;
 				}, err, msg);
 			});
@@ -92,7 +92,18 @@ x.throws = function (fn, err, msg) {
 			};
 		}
 
-		assert.throws(fn, err, msg);
+		var result;
+
+		assert.throws(function () {
+			try {
+				fn();
+			} catch (error) {
+				result = error;
+				throw error;
+			}
+		}, err, msg);
+
+		return result;
 	} catch (err) {
 		test(false, create(err.actual, err.expected, err.operator, err.message, x.throws));
 	}
--- a/readme.md
+++ b/readme.md
@@ -566,6 +566,8 @@ Assert that `function` throws an error, or `promise` rejects with an error.
 
 `error` can be a constructor, regex, error message or validation function.
 
+Returns the error thrown by `function` or the rejection reason of `promise`.
+
 ### .notThrows(function|promise, [message])
 
 Assert that `function` doesn't throw an `error` or `promise` resolves.
--- a/test/assert.js
+++ b/test/assert.js
@@ -1,5 +1,6 @@
 'use strict';
 var test = require('tap').test;
+var Promise = require('bluebird');
 var assert = require('../lib/assert');
 
 test('.pass()', function (t) {
@@ -180,6 +181,26 @@ test('.throws()', function (t) {
 	});
 
 	t.end();
+});
+
+test('.throws() returns the thrown error', function (t) {
+	var expected = new Error();
+	var actual = assert.throws(function () {
+		throw expected;
+	});
+
+	t.is(actual, expected);
+
+	t.end();
+});
+
+test('.throws() returns the rejection reason of promise', function (t) {
+	var expected = new Error();
+
+	assert.throws(Promise.reject(expected)).then(function (actual) {
+		t.is(actual, expected);
+		t.end();
+	});
 });
 
 test('.notThrows()', function (t) {


lib/assert.js
```diff
@@ -89,14 +89,25 @@ x.throws = function (fn, err, msg) {
...
+			errMsg = err;
 			err = function (err) {
 				return err.message === errMsg;
 			};
 		}
 
 		assert.throws(fn, err, msg);
+
+		if (errMsg) {
+			return Error(errMsg);
```
novemberborn: `errMsg` is the expected message. The workaround above stems from #125 and was done because `assert.throws` doesn't take a string argument as the expected message, but we did want that behavior in AVA.

In other words it's not something to be returned.

lib/assert.js
```diff
@@ -89,14 +89,25 @@ x.throws = function (fn, err, msg) {
...
+		}
+
+		try {
+			fn();
```
novemberborn: This means `fn()` is invoked twice. Instead I think you need something like:

```js
var result;
assert.throws(function () {
  try {
    fn();
  } catch (error) {
    result = error;
    throw error;
  }
}, err, msg);
return result;
```

kasperlewau: Good call. I didn't consider the implications of calling `fn` outside of the call to core-assert.

readme.md
```diff
@@ -568,6 +568,8 @@ Assert that `function` throws an error or `promise` rejects.
 
 `error` can be a constructor, regex, error message or validation function.
 
+Returns the error thrown by `function` or the reason of `promise` rejection.
```
novemberborn: > Returns the error thrown by `function` or the rejection reason of `promise`.

lib/assert.js
kasperlewau: Might actually be able to kill this if-block entirely given your suggestion at https://github.com/sindresorhus/ava/pull/576/files#r54448454. I'll give it a whirl!

test/assert.js
```diff
@@ -182,6 +182,110 @@ test('.throws()', function (t) {
 	t.end();
 });
 
+test('.throws() returning Error', function (t) {
+	t.doesNotThrow(function () {
+		var assertion = assert.throws(function () {
+			throw new Error('foo');
+		});
+
+		assert.same('foo', assertion.message);
```
novemberborn: I'd prefer a strict identity check:

```js
var expected = new Error();
var actual = assert.throws(function () {
  throw expected;
});
t.equal(actual, expected);
```

Also note that you don't need the `t.doesNotThrow()` wrapper, just run a single `t.equal()` assertion. No need at all for the next test.

lib/test.js
```diff
@@ -154,8 +154,18 @@ Test.prototype.run = function () {
...
-			function () {
-				self.exit();
+			function (assertRet) {
+				var exitReason;
+
+				if (assertRet instanceof Error) {
+					exitReason = assertRet;
+				}
+
+				if (assertRet && assertRet.length) {
```
kasperlewau: I think `if (assertRet.length)` would do just fine on it's own, no need for `assertRet &&`.

lib/assert.js
novemberborn: Yes, you're returning a very different value here.

readme.md
kasperlewau: Much better. I struggled for a good couple of minutes during my commute home on how to word that. ugh. 

Thanks!

lib/assert.js
kasperlewau: It's gone!

test/assert.js
kasperlewau: > I'd prefer a strict identity check 

I'll apply that to the whole lot of these `return` tests. 

> Also note that you don't need the t.doesNotThrow() wrapper, just run a single t.equal() assertion. No need at all for the next test.

Gotcha! I'll kill off some assertions.

novemberborn:
I don't understand why you're changing `lib/test`. What am I missing?

No need to exhaustively test `.throws()`. Just need to verify it returns the thrown error. You do need a test for when a rejected promise is returned though. You shouldn't need to change the observable and promise tests.

novemberborn:
Cool getting there!

test/assert.js
kasperlewau: I just went ahead and killed off the whole bunch of the specs - just the one test now to verify `actual === expected`. Regardless of secondary arguments passed to `throws`, it should always be the same. sound good to you @novemberborn?

novemberborn: Yea. No need to pass any secondary arguments either, the existing tests don't seem to bother with that either.

kasperlewau:
> I don't understand why you're changing lib/test. What am I missing?

This might be a case of me not writing the tests as one should to get it working as expected. But when going back and reverting the changes made to `lib/test` makes the following spec break; 

```js
// test/promise.js
test('throws becomes the promise rejection', function (t) {
	var rejection;
	ava(function (a) {
		a.plan(1);

		rejection = new Error('foo');
		var promise = Promise.reject(rejection);
		return a.throws(promise);
	}).run().then(function (result) {
		t.is(result.passed, true);
		t.is(result.result.assertCount, 1);
		t.is(result.reason, rejection);
		t.end();
	});
});
``` 

[this line](https://github.com/sindresorhus/ava/pull/576/files#diff-dd2e881af3311f834199c7b515cff0c7R274) is the real deal-breaker in this specific case. Without it, I don't see how one would return the rejection reason to the callee. 

Again, it's not that far-fetched that this test does not in fact belong here and I'm just *doing it wrong.*

lib/test.js
kasperlewau: > nope. nope. nope. get a grip.

kasperlewau:
Sorry, I phased out after having my supper last night (I blame Comedy Central). I got around to cleaning some stuff out, but I'm sort of at a loss on how to handle the test case for a Promise rejection.

So far I'm at [this point](https://github.com/kasperlewau/ava/commit/380dedae108434315888e84a5c84813cf9503115); 

```js
test('.throws() becomes the rejection reason of promise', function (t) {
	var expected = new Error();

	assert.throws(Promise.reject(expected)).then(function (err) {
		assert.same(expected, err);
		t.end();
	});
});
```

But that's not *actually* testing the return value of `throws`. Still working on trying to figure that one out - I keep looking at `test/promises` and the way Promises are handled in there but, I haven't gotten it to work just yet.

kasperlewau:
Alright, I reckon we're good to go there with the latest changes made. 

* `lib/test` changes are gone. 
* `test/[promises|observables]` changes are gone. 
* Promise rejection assertion is in place in `test/assert`
  * Could potentially be cleaned up ever so slightly? 
* Code has been touched up as per @novemberborn's suggestions.

Let me know what you think, I've got an itch in my squash finger. 

--- 

I guess a brief look from @jamestalmage would be warranted as well, given he's the author of #493.

vdemedes:
To me, returning object with `reason` property is not logical. If `t.throws()` throws, everything is good and execution is continued, there's no "failure reason", because nothing failed. Why don't we just return error from `t.throws()`?

kasperlewau:
@vdemedes Sorry, I should've probably updated the OP. Like I stated before, I wasn't all too comfy with overriding the return value of `test.exit` to reflect what happened with `test.throws`. I overcomplicated the matter. 

--- 

The way it works now is this; 

```js
test(t => {
  const expected = new Error();
  const actual = t.throws(Promise.reject(expected));
  t.same(expected, actual); 
});
```

Whereas what I was doing in the beginning was something like so;

```js
test(t => {
  const expected = new Error();
  return t.throws(Promise.reject(expected));
}).then(result) {
  // result.reason === expected
}
```

I was trying to cater to the `test/promise` suite, as opposed to testing `throws` where it **should** be tested (in `test/assert`). That whole snafu is gone now however, after touching up on the parts Mark mentioned. 

I'll gladly edit the OP when I get a on the train. At work currently.

vdemedes:
Oh great, so it does return thrown Error from `t.throws()`, right? No rush, thank you!

kasperlewau:
> Oh great, so it does return thrown Error from t.throws(), right?

It does indeed! :)

test/assert.js
```diff
@@ -182,6 +183,29 @@ test('.throws()', function (t) {
 	t.end();
 });
 
+test('.throws() becomes the thrown Error', function (t) {
```
novemberborn: `returns` rather than `becomes`. Could lowercase `Error` as well.

test/assert.js
```diff
@@ -182,6 +183,29 @@ test('.throws()', function (t) {
...
+	t.end();
+});
+
+test('.throws() becomes the rejection reason of promise', function (t) {
+	var expected = new Error();
+	var exec = new Promise(function (resolve) {
+		resolve(assert.throws(Promise.reject(expected)));
+	});
+
+	exec.then(function (actual) {
```
novemberborn: This should work:

```js
assert.throws(Promise.reject(expected)).then(function (actual) {
```

`assert.throws()` already returns a promise because you pass one rather than a function. That promise fulfils with the rejection reason.

novemberborn:
@kasperlewau cool! Just some simplifications in the test. Feel free to push up a squashed commit when that's fixed and I'll merge.

test/assert.js
kasperlewau: Gosh. I really do like to [overcomplicate all the things](https://github.com/kasperlewau/ava/commit/3f24a108f516cce747191a9a788c3c337e4a66e5) :disappointed:

test/assert.js
kasperlewau: Gotcha!

kasperlewau: Would you like for me to change the subject of the Promise rejection while I'm at it? @novemberborn 

> '.throws() becomes the rejection reason of promise'
> '.throws() returns the rejection reason of promise'

vdemedes: Yeah, '.throws() returns the rejection reason of promise' is a correct version indeed.

kasperlewau:
@novemberborn Resolved, squashed, pushed. Let the CI wait commence.

test/assert.js
```diff
@@ -182,6 +183,26 @@ test('.throws()', function (t) {
 	t.end();
 });
 
+test('.throws() returns the thrown error', function (t) {
+	var expected = new Error();
+	var actual = assert.throws(function () {
+		throw expected;
+	});
+
+	assert.same(actual, expected);
```
novemberborn: Oops wait. `assert.same()` means "deep equal". We want strict equality here. You should use `assert.is(actual, expected)` instead. Though the tests here are for AVA's `assert` module. Perhaps use `t.is(actual, expected)` which is from [`tap`](http://www.node-tap.org/asserts/)'s assertion library (AVA's tests are run using `tap`).

Same below.

novemberborn:
@kasperlewau OK just the tiniest issue with the assertion. Feel free to force push again and I'll merge.

kasperlewau:
@novemberborn 

Right-o! I'm still fairly new to `ava` (and `tap` for that matter). Coming from a `chai` background so I've hit the `t.equal is not a function` several times over now :wink: 

Off topic; `same` is a *little bit far fetched* when you want to perform a deep comparison. If you think twice about it, it makes all the more sense though. imho. 

I'll get this sorted out!

kasperlewau:
Rebasing and pushing in a jiffy.

novemberborn:
Thanks @kasperlewau!

giphy

kasperlewau:
Glad I could help! : )

sindresorhus:
Thank you! Keep'em coming :)

vdemedes:
Thanks a lot!

PR #580, opened 2016-03-01
[WIP] update typescript definition file to work with ts -> es6 -> es5 pipeline

opensrcken:
I had to use something like this to be able to use ava from TS in my workflow. I am not certifying that this change will work for all flows, but wanted to start a discussion about it nonetheless.

My TS target is es6, and the output is compiled by babel to es5.

In TS, the ultimate import looks something like
```
import {test} from 'ava';
```

and the compiled es5 looks something like:
```
var ava_1 = require('ava');
ava_1.test(function (t) { ... });
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,141 +1,149 @@
-export interface Observable {
-	subscribe(observer: (value: {}) => void): void;
+declare module 'ava/types' {
+    export interface Observable {
+        subscribe(observer: (value: {}) => void): void;
+    }
+
+    export type Test = (t: TestContext) => Promise<void> | Iterator<any> | Observable | void;
+    export type SerialTest = (t: TestContext) => void;
+    export type CallbackTest = (t: CallbackTestContext) => void;
+
+    export interface Runner {
+        (name: string, run: Test): void;
+        (run: Test): void;
+        skip: Runner;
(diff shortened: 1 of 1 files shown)


sindresorhus:
@opensrcken Please read and adhere to the [contributing guidelines](../blob/master/contributing.md) :)

Give the PR a proper title and don't change the indentation.

opensrcken:
Ah, thanks for the reference.

The indentation was changed because I created new top-level modules in the file. Or were you referring to something else?

opensrcken:
Flagged as [WIP] since this is meant to spark discussion around issues others may be having with typescript.

sindresorhus:
> Or were you referring to something else?

You changed the indentation from tabs to spaces.

sindresorhus:
> import {test} from 'ava';

That should be `import test from 'ava';`

ivogabe:
:-1:, definition files for TypeScript should not use `declare module`. When you set module to commonjs (or moduleResolution to node, if you're using babel to transpile modules to commonjs), TypeScript will find these definitions

opensrcken:
Ah, I wasn't aware that TS had added auto-resolution. Excellent.

PR #582, opened 2016-03-01
[breaking] Don't convert non-Error promise rejection reason

### LINKED ISSUE #468: t.throws(promise, stringOrRegexp) should only test against errors
If you are rejecting a Promise, you should be doing so with an Error object.

If you are testing that a promise will be rejected, we currently allow this:

```js
t.throws(somePromise, /messageRegexp/);
t.throws(somePromise, 'expectedMessage');
```

That makes sense.

However. If the promise is rejected with a string:

```js
Promise.reject('someErrorMessage');
```

We magically handle detecting it's not an error and test the expectation directly against the string instead of `error.message`.

I think this is silly. We are accommodating bad practice (rejecting with non-errors), and creating an ambiguous assertion conditions:

```js
// There is no way to differentiate these using t.throws
var promiseA = Promise.reject('foo');
var promiseB = Promise.reject(new Error('foo'));
```

I propose we drop the magic string behavior. 

If they specify the second arg, we use it as an expectation against `rejectionReason.message` only.

kasperlewau:
Promises that are rejected with a non-Error are currently passing expectations as [we're converting to an Error](https://github.com/sindresorhus/ava/blob/f77ded959a0e7de6e9bf65d8c2fd6625bc59ba47/lib/assert.js#L85) in `lib/assert`. 

This PR changes the behaviour of `t.throws` so that it no longer considers the following example a pass: 
```js
test(async t => {
 await t.throws(Promise.reject('foo'), 'foo');
});
```
If the rejection reason is a non-Error, `core-assert` catches this which results in an AssertionError.

Whereas the following would pass with flying colors:
```js
test(async t => {
  await t.throws(Promise.reject(new Error('foo'), 'foo');
});
```

---

Docs for `t.throws` saw a minor addition; 
> Assert that function throws an error**,** or promise rejects **with an error.**

Fixes #468

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -79,11 +79,7 @@ x.throws = function (fn, err, msg) {
 				x.throws(noop, err, msg);
 			}, function (fnErr) {
 				x.throws(function () {
-					if (fnErr instanceof Error) {
-						throw fnErr;
-					} else {
-						throw new Error(fnErr);
-					}
+					throw fnErr;
 				}, err, msg);
 			});
 	}
--- a/readme.md
+++ b/readme.md
@@ -560,7 +560,7 @@ Assert that `value` is not deep equal to `expected`.
 
 ### .throws(function|promise, [error, [message]])
 
-Assert that `function` throws an error or `promise` rejects.
+Assert that `function` throws an error, or `promise` rejects with an error.
 
 `error` can be a constructor, regex, error message or validation function.
 
--- a/test/promise.js
+++ b/test/promise.js
@@ -210,28 +210,15 @@ test('throws with string argument will reject if message does not match', functi
 	});
 });
 
-test('handle throws with regex with string reject', function (t) {
-	ava(function (a) {
-		a.plan(1);
-
-		var promise = Promise.reject('abc');
-		return a.throws(promise, /abc/);
-	}).run().then(function (result) {
-		t.is(result.passed, true);
-		t.is(result.result.assertCount, 1);
-		t.end();
-	});
-});
-
-test('handle throws with string with string reject', function (t) {
+test('does not handle throws with string reject', function (t) {
 	ava(function (a) {
 		a.plan(1);
 
 		var promise = Promise.reject('abc');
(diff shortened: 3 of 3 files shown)


vdemedes:
I remember we had some back and forth with this decision, but atm I'd also like to merge this and strictly expect `Error` objects.

kasperlewau:
I suppose yielding a custom Error to the consumer with a recommendation to reject Promises with actual Errors would be somewhat of a middle ground, if we're in 'back and forth land' that is? 

A missing expectation AssertionError doesn't *explicitly* highlight that fact (imho), but I didn't take a good look at what the error stack yielded as it stands. A custom Error could just be extraneous information.

novemberborn:
> I remember we had some back and forth with this decision, but atm I'd also like to merge this and strictly expect Error objects.

We all seem to be in agreement in #468 so let's not restart this argument :wink:

lib/assert.js
```diff
@@ -81,8 +81,6 @@ x.throws = function (fn, err, msg) {
 				x.throws(function () {
 					if (fnErr instanceof Error) {
 						throw fnErr;
-					} else {
```
novemberborn: Remove the if-guard as well, simply `throw fnError`. This makes it equivalent to `t.throws(() => { throw 'foo' })`.

novemberborn:
@sindresorhus this will be a breaking change.

lib/assert.js
kasperlewau: Very well! I'll go ahead and amend my commit if you don't mind, and save myself the trouble of having to squash later.

novemberborn: Sure. It's going to conflict with #576 though :wink:

kasperlewau: Right. I'll get on this once  #576 is merged.

sindresorhus:
> this will be a breaking change.

True, although I really doubt (hope) nobody depends on this behavior.

sindresorhus:
Thanks @kasperlewau :)

kasperlewau:
Thank you(s)!

Issue #584, opened 2016-03-01
Unnecessary `counted` variable in Api


novemberborn:
These lines are unnecessary because `counted` is never changed from `false` to `true`.

* https://github.com/sindresorhus/ava/blob/58e1c561296a409a9557ceb1c7684685404c4278/api.js#L184
* https://github.com/sindresorhus/ava/blob/58e1c561296a409a9557ceb1c7684685404c4278/api.js#L187:L189

I'd remove them myself but there's some other stuff to get to. Writing it down here so I don't forget, but if anybody else wants to chime in please do :star2:

avivr:
I'd like to give a try

novemberborn:
@avivr cool!

PR #587, opened 2016-03-01
Remove unnecessary `counted` variable in api.js

avivr:
Remove declaration and usage for unnecessary `counted` variable in ava/api.js.
This should resolve issue #584


sotojuan:
Looks good to me :-)

novemberborn:
Thanks @avivr!

PR #588, opened 2016-03-02
[t.ds] do not provide context in before and after hooks

SamVerschueren:
The `context` fix wasn't entirely correct. It was possible to set the `context` in the `before` and `after` tests as well. So had to add some extra things in order for this to work.

// @ivogabe

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -3,8 +3,11 @@ export interface Observable {
 }
 
 export type Test = (t: TestContext) => Promise<void> | Iterator<any> | Observable | void;
+export type ContextualTest = (t: ContextualTestContext) => Promise<void> | Iterator<any> | Observable | void;
 export type SerialTest = (t: TestContext) => void;
+export type ContextualSerialTest = (t: ContextualTestContext) => void;
 export type CallbackTest = (t: CallbackTestContext) => void;
+export type ContextualCallbackTest = (t: ContextualCallbackTestContext) => void;
 
 export interface Runner {
 	(name: string, run: Test): void;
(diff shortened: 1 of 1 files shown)


ivogabe:
Looks good!

avimar:
Is there a reason why `context` is specifically NOT included in `before` and `after`? I'd think you can use it to create a fixture.. but with some dynamic data, that doesn't need a full-blown `beforeEach`.

Thanks!

SamVerschueren:
From the documentation

> Context sharing is _not_ available to before and after hooks.

Open a new issue if you want to discuss that as a feature request.

avimar:
@SamVerschueren I see you wrote the patch. You're just following spec? I assume there is a reason the spec was written that way...?
Hmm, there's each file is isolated, so I can just use globals I suppose.

SamVerschueren:
I don't think a `context` makes sense in a `before` and `after` hook to be honest.

Let's suppose (for the sake of this discussion), that we had a `context`.

```js
t.before(t => {
    t.context.foo = 'bar';
});

test('this is a test', t => {
    console.log(t.context.foo);
    //=> 'bar'
});
```

This is exactly the same as

```js
const foo = 'bar';

test('this is a test', t => {
    console.log(foo);
    //=> 'bar'
});
```

They are shared over all the tests, so are constants/variables defined in the beginning of the file.

Not sure if @jamestalmage has to say something extra on this :).

jamestalmage:
Context is used to share data between the beforeEach and a specific test. If you have five actual tests, and they all share on common beforeEach hook, that hook will be called five times: once with each context for each of the five tests.

Passing context to before doesn't make sense. It's only called once. Which context would you give it?

SamVerschueren:
@jamestalmage It would be possible to inherit the context passed in `beforeEach` from the context of `before`. But as stated before in my example, I don't think that makes sense.

avimar:
... and if you had a promise-based setup, you could either use `await` or declare the `var` and assign it during a `before` hook. Basically, if you need "global" `context`, you don't need anything special for that.

OK, thanks!

7flash:
Please remove misleading example from readme:
screenshot_2018-06-26_15-35-19

PR #592, opened 2016-03-02
Add `match` CLI option to run only matching tests

### LINKED ISSUE #476: Only run tests with title matching expression

kasperlewau:
Add a `match` option for running a specific set of tests with titles matching the given pattern. Leverages [matcher](https://github.com/sindresorhus/matcher) internally, so `!` and `*` is supported. 

```
ava --match='foo*' --match='!*bar' /test/*.js

# shorthand
ava -m='foo*' -m='!*bar' /test/*.js
```

This would run all tests inside `/test` with a title starting with `foo` and not ending with `bar`. 

The match option can also be configured in the `ava` section of `package.json`. 

```json
{
  "ava": {
    "match": [
      "foo*"
    ]
  }
}
```

Fixes #476

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -49,6 +49,7 @@ var cli = meow([
 	'  --tap, -t        Generate TAP output',
 	'  --verbose, -v    Enable verbose output',
 	'  --no-cache       Disable the transpiler cache',
+	'  --match, -m      Only run tests with matching title (Can be repeated)',
 	// Leave --watch and --sources undocumented until they're stable enough
 	// '  --watch, -w      Re-run tests when tests and source files change',
 	// '  --source         Pattern to match source files so tests can be re-run (Can be repeated)',
@@ -67,7 +68,8 @@ var cli = meow([
 	string: [
 		'_',
 		'require',
-		'source'
+		'source',
+		'match'
 	],
 	boolean: [
 		'fail-fast',
@@ -82,7 +84,8 @@ var cli = meow([
 		v: 'verbose',
 		r: 'require',
 		s: 'serial',
-		w: 'watch'
+		w: 'watch',
+		m: 'match'
 	}
 });
 
@@ -98,7 +101,8 @@ var api = new Api({
 	serial: cli.flags.serial,
 	require: arrify(cli.flags.require),
 	cacheEnabled: cli.flags.cache !== false,
-	explicitTitles: cli.flags.watch
+	explicitTitles: cli.flags.watch,
+	match: arrify(cli.flags.match)
 });
 
 var reporter;
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -3,6 +3,7 @@ var EventEmitter = require('events').EventEmitter;
 var util = require('util');
 var Promise = require('bluebird');
 var optionChain = require('option-chain');
+var matcher = require('matcher');
 var TestCollection = require('./test-collection');
 
 function noop() {}
@@ -42,6 +43,7 @@ function Runner(options) {
 	this.tests = new TestCollection();
 	this._bail = options.bail;
 	this._serial = options.serial;
+	this._match = options.match;
 
 	this._addTestResult = this._addTestResult.bind(this);
 }
@@ -67,6 +69,10 @@ optionChain(chainableMethods, function (opts, title, fn) {
 
 	if (this._serial) {
 		opts.serial = true;
+	}
+
+	if (opts.type === 'test' && (this._match && this._match.length > 0) && title !== null) {
+		opts.skipped = matcher([title], this._match).length === 0;
 	}
 
 	this.tests.add({
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -71,8 +71,8 @@ optionChain(chainableMethods, function (opts, title, fn) {
 		opts.serial = true;
 	}
 
-	if (opts.type === 'test' && (this._match && this._match.length > 0) && title !== null) {
-		opts.skipped = matcher([title], this._match).length === 0;
+	if (opts.type === 'test' && title !== null && this._match && this._match.length > 0) {
+		opts.exclusive = matcher([title], this._match).length > 0;
 	}
 
 	this.tests.add({
--- a/readme.md
+++ b/readme.md
@@ -115,6 +115,7 @@ $ ava --help
     --tap, -t        Generate TAP output
     --verbose, -v    Enable verbose output
     --no-cache       Disable the transpiler cache
+    --match, -m      Only run tests with matching title (Can be repeated)',
 
   Examples
     ava
@@ -134,6 +135,32 @@ Directories are recursive by default. Directories named `fixtures` and `helpers`
 
 When using `npm test`, you can pass positional arguments directly `npm test test2.js`, but flags needs to be passed like `npm test -- --verbose`.
 
+### The `--match` flag
+
+The `--match` flag allows you to run a subset of the tests in your suite that have a matching title. This is achieved with simple wildcard patterns. For more information, check out [sindresorhus/matcher](https://github.com/sindresorhus/matcher).
+
+```
+# match titles ending with 'foo'
+$ ava --match='*foo'
+
+# match titles starting with 'foo'
+$ ava --match='foo*'
+
+# match titles containing 'foo'
+$ ava --match='*foo*'
+
+# match titles not containing 'foo'
+$ ava --match='!*foo*'
+
+# match titles starting with 'foo' and ending with 'bar'
+$ ava --match='foo*bar'
+
+# match titles starting with 'foo' or ending with 'bar'
+$ ava --match='foo*' --match='*bar'
+```
+
+*Tests without a title will not run if* **any** *match pattern is supplied.*
+
 ## Configuration
 
 All of the CLI options can be configured in the `ava` section of your `package.json`. This allows you to modify the default behavior of the `ava` command, so you don't have to repeatedly type the same options on the command prompt.
@@ -144,6 +171,10 @@ All of the CLI options can be configured in the `ava` section of your `package.j
     "files": [
       "my-test-folder/*.js",
       "!**/not-this-file.js"
+    ],
+    "match": [
+      "*oo",
+      "!foo"
     ],
     "failFast": true,
     "tap": true,
--- a/readme.md
+++ b/readme.md
@@ -135,32 +135,6 @@ Directories are recursive by default. Directories named `fixtures` and `helpers`
 
 When using `npm test`, you can pass positional arguments directly `npm test test2.js`, but flags needs to be passed like `npm test -- --verbose`.
 
-### The `--match` flag
-
-The `--match` flag allows you to run a subset of the tests in your suite that have a matching title. This is achieved with simple wildcard patterns. For more information, check out [sindresorhus/matcher](https://github.com/sindresorhus/matcher).
-
-```
-# match titles ending with 'foo'
-$ ava --match='*foo'
-
-# match titles starting with 'foo'
-$ ava --match='foo*'
-
-# match titles containing 'foo'
-$ ava --match='*foo*'
-
-# match titles not containing 'foo'
-$ ava --match='!*foo*'
-
-# match titles starting with 'foo' and ending with 'bar'
-$ ava --match='foo*bar'
-
-# match titles starting with 'foo' or ending with 'bar'
-$ ava --match='foo*' --match='*bar'
-```
-
-*Tests without a title will not run if* **any** *match pattern is supplied.*
-
 ## Configuration
 
 All of the CLI options can be configured in the `ava` section of your `package.json`. This allows you to modify the default behavior of the `ava` command, so you don't have to repeatedly type the same options on the command prompt.
@@ -276,6 +250,49 @@ test.only('will be run', t => {
 	t.pass();
 });
 ```
+
+### Matched-tests
+
+The `--match` flag allows you to run a subset of the tests in your suite that have a matching title. This is achieved with simple wildcard patterns. For more information, check out [matcher](https://github.com/sindresorhus/matcher).
+
+```
+# match titles ending with 'foo'
+$ ava --match='*foo'
+
+# match titles starting with 'foo'
+$ ava --match='foo*'
+
+# match titles containing 'foo'
+$ ava --match='*foo*'
+
+# match titles not containing 'foo'
+$ ava --match='!*foo*'
+
+# match titles starting with 'foo' and ending with 'bar'
+$ ava --match='foo*bar'
+
+# match titles starting with 'foo' or ending with 'bar'
+$ ava --match='foo*' --match='*bar'
+```
+
+```js
+// $ ava --match='*oo'
+
+test('foo will run', t => {
+	t.pass();
+});
+
+test.only('moo will also run', t => {
+	t.pass();
+});
+
+// won't run!
+test(function () {
+	t.fail();
+});
+```
+
+Note that a match pattern takes precedence over `.only`, and *any tests without an explicit title will* **not run** *if a match pattern is supplied.*
 
 ### Skip-tests
 
--- a/test/cli.js
+++ b/test/cli.js
@@ -112,7 +112,7 @@ test('pkg-conf: pkg-overrides', function (t) {
 });
 
 test('pkg-conf: cli takes precedence', function (t) {
-	execCli(['--no-serial', '--cache', '--no-fail-fast', '--require=./required.js', 'c.js'], {dirname: 'fixture/pkg-conf/precedence'}, function (err) {
+	execCli(['--match=foo*', '--no-serial', '--cache', '--no-fail-fast', '--require=./required.js', 'c.js'], {dirname: 'fixture/pkg-conf/precedence'}, function (err) {
 		t.ifError(err);
 		t.end();
 	});
@@ -153,6 +153,13 @@ test('watcher works', function (t) {
 				killed = true;
 			}
 		}
+	});
+});
+
+test('--match works', function (t) {
+	execCli(['-m=foo', '-m=bar', '-m=!baz', '-m=t* a* f*', '-m=!t* a* n* f*', 'fixture/matcher-skip.js'], function (err) {
+		t.ifError(err);
+		t.end();
 	});
 });
 
--- a/index.js
+++ b/index.js
@@ -13,7 +13,8 @@ require('./lib/test-worker').avaRequired = true;
 var opts = globals.options;
 var runner = new Runner({
 	serial: opts.serial,
-	bail: opts.failFast
+	bail: opts.failFast,
+	match: opts.match
 });
 
 // check if the test is being run without AVA cli
--- a/package.json
+++ b/package.json
@@ -106,6 +106,7 @@
     "is-promise": "^2.1.0",
     "last-line-stream": "^1.0.0",
     "loud-rejection": "^1.2.0",
+    "matcher": "^0.1.1",
     "max-timeout": "^1.0.0",
     "md5-hex": "^1.2.0",
     "meow": "^3.7.0",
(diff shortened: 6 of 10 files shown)


vdemedes:
I :heart: it! Simple code which does its job. Thank you!

There's one little thing though, could you please add some minimal docs to the readme along with a few examples?

kasperlewau:
> There's one little thing though, could you please add some minimal docs to the readme along with a few examples?

Absolutely. I suppose somewhere in the `cli` section would make sense? Need to step away from open source for a couple of hours. Sometime tonight! :smile:

vdemedes:
Yep, in `cli` section ;) Sure, no pressure, great job!

kasperlewau:
Hm. Not sure why the build is failing, ran just fine™ on my machine. 

Node v5.7.0
Darwin v15.3.0

```
  options.match will skip tests with non-matching titles
  not ok TypeError: Cannot read property 'runOnlyExclusive' of undefined
```

vdemedes:
All versions failed, so it must be something :)

vdemedes:
Try pulling latest changes (#556 just got merged in).

kasperlewau:
Nope. Still no sauce unfortunately. It does point in the direction of #566. 

Could it be that either `match` or `runOnlyExclusive` didn't add an entry in [`runner#chainableMethods`](https://github.com/sindresorhus/ava/pull/592/files#diff-b3b53682a18f203ac8d29b0e277cad26R9)? 

Stabbing in the dark here : )

kasperlewau:
[this](https://github.com/sindresorhus/ava/pull/556/files#diff-b3b53682a18f203ac8d29b0e277cad26R99) appears to be the failing line. Somehow I'm not passing `options` to the `run` method.

lib/runner.js
```diff
@@ -38,6 +39,7 @@ function Runner(options) {
 	this.tests = new TestCollection();
 	this._bail = options.bail;
 	this._serial = options.serial;
+	this._match = options.match;
```
novemberborn: These options must be explicitly forwarded in https://github.com/sindresorhus/ava/blob/4a1bd79fdd5dd13e011750b6cb2fefc94f74b5c6/index.js#L14

lib/runner.js
```diff
@@ -55,6 +57,10 @@ optionChain(chainableMethods, function (opts, title, fn) {
 		opts.serial = true;
 	}
 
+	if (this._match) {
```
novemberborn: Once option forwarding is fixed you'll always have a value here. However you must make sure to only run the next line if `this._match` is not empty. By default it will be.

novemberborn:
@kasperlewau nice work. However it doesn't actually work end-to-end. I've left some inline comments as to why. Could you add a test to `test/cli` to exercise the feature end-to-end?

kasperlewau:
@vdemedes I've added docs for the `--match` flag (in the `cli` segment). We could probably shave off a sentence/example or two. 

I'm still on the fence regarding some wording. 
> match titles ending with 'foo' **[vs]** match titles that end with 'foo'

Let me know what you guys think.

kasperlewau:
@novemberborn Wuh-oh! I'll see what I can do about that. Cheers!

novemberborn:
> *Tests without a title will not run if* **any** *match pattern is supplied.*

Unfortunately the runner does not have access to the full test title. It's later [derived from the function name](https://github.com/sindresorhus/ava/blob/4a1bd79fdd5dd13e011750b6cb2fefc94f74b5c6/lib/test-collection.js#L45), for instance.

Maybe that title logic needs to be moved to the runner. @vdemedes?

lib/runner.js
```diff
@@ -55,6 +57,10 @@ optionChain(chainableMethods, function (opts, title, fn) {
...
+		opts.skipped = matcher([title], this._match).length === 0;
```
novemberborn: This function is also called when adding hooks, which themselves can be titled. Would be weird to skip those. I think you should be able to guard against that by checking `opts.type === 'test'`.

kasperlewau: Right, because they too are a part of `chainableMethods`. willfix!

kasperlewau:
@novemberborn 

* We're now passing through the `match` option from `api -> runner`.
* We're guarding against an empty `match` value. 
* We're guarding against non-`test` types in the 'apply-match-logic' if block. 

The one thing I haven't done is setting up an end-to-end test in `test/cli` to *ensure it all works™*. Do you have any suggestions/pointers I can take with me on that voyage? :smile:

vdemedes:
@novemberborn It can't be moved to Runner, because we display full test titles only when there are multiple files and only CLI knows how many test files we have. And plus, it would be out of Runner's scope imho.

kasperlewau:
@novemberborn @vdemedes 

I could reverse the behaviour by checking that title is not `null`, but to me as a user it'd feel weird to see tests without a title running when I've asked for a specific set of titles. 

The one time I feel a non-titled test should run, would be with either no match pattern supplied, or `--match=*`. But I don't see why one would ever want to supply such a pattern.

cli.js
```diff
@@ -49,6 +49,7 @@ var cli = meow([
 	'  --tap, -t        Generate TAP output',
 	'  --verbose, -v    Enable verbose output',
 	'  --no-cache       Disable the transpiler cache',
+	'  --match, -m      Only run tests with matching title',
```
sindresorhus: > '  --match, -m      Only run tests with matching title (Can be repeated)',

novemberborn:
@vdemedes 

> It can't be moved to Runner, because we display full test titles only when there are multiple files and only CLI knows how many test files we have.

This isn't about test title prefixes. The runner calls `TestCollection#add()` which changes the title beyond what was determined in the runner, even though the runner has access to the same arguments.

readme.md
```diff
@@ -134,6 +135,32 @@ Directories are recursive by default. Directories named `fixtures` and `helpers`
...
+$ ava --match=*foo
+
+# match titles starting with 'foo'
+$ ava --match=foo*
+
+# match titles containing 'foo'
+$ ava --match=*foo*
+
+# match titles not containing 'foo'
+$ ava --match=!*foo*
```
sindresorhus: `!*foo*` needs to be quoted in ZSH at least.

novemberborn:
> The one thing I haven't done is setting up an end-to-end test in test/cli to ensure it all works™. Do you have any suggestions/pointers I can take with me on that voyage?

See the "watcher works" test, though it does a few more things than are required here.

readme.md
sindresorhus: So does `foo*`, so probably better to quote them all?

kasperlewau: Me being the inconsiderate Bash user, I opted for no quotes. I'll single quote 'em all :)

cli.js
kasperlewau: :+1:

vdemedes:
@novemberborn Oh I see, I mixed that up. In that case, @kasperlewau's logic seems very reasonable:

> Tests without a title will not run if any match pattern is supplied.

> it'd feel weird to see tests without a title running when I've asked for a specific set of titles.

And still title manipulation feels out of runner's scope, so I'd leave it in test collection.

novemberborn:
My point is that the runner calls another method which figures out an implicit title. Those are not matched. I guess it's OK to say that match only applies to explicit titles but it shouldn't be too hard to apply it to implicit ones too.

vdemedes:
Ok, I see your point. What about merging this PR as it is and opening a new issue to extend this functionality? I see some refactoring work involved here.

kasperlewau:
Ok, I believe that's the e2e test for `--match` in the bag. Fingers crossed Travis doesn't kick me in the groin : )

sindresorhus:
I don't think it makes sense to match on implicit titles. It should IMHO only match on what people write out in their test file.

novemberborn:
> What about merging this PR as it is and opening a new issue to extend this functionality? I see some refactoring work involved here.

Yes that's fair.

> I don't think it makes sense to match on implicit titles. It should IMHO only match on what people write out in their test file.

That's fair too. Not sure we want to encourage `test(function myTestTitle (t) {` style tests anyway.

Apologies for the distraction :smiley:

kasperlewau:
I realised just now I took care of @sindresorhus's comment about repeatabilty - but I did so in the 'wrong' place :smile: 

I'll take care of that on my way to work!

novemberborn:
Should this use `exclusive` rather than `skipped`? That is, set `exclusive` for matching tests. Perhaps in the Api report an error when no tests match. That makes `--match` similar to adding `.only` in the test files.

kasperlewau:
> Should this use exclusive rather than skipped? That is, set exclusive for matching tests. 

Right off the bat that sounds better to me @novemberborn. I can't really comment further than that - I'm not familiar enough with the difference(s) between `skip` and `exclusive` when put into the context of say `--watch` or something of the sort. I bet you do though : ) 

If anything it would change the conditional statement to something slightly more pretty.

novemberborn:
> I'm not familiar enough with the difference(s) between skip and exclusive when put into the context of say --watch or something of the sort. I bet you do though : )

It'll probably be OK, but we can make whatever improvements are needed once this lands.

> What I mean to say is that the semantics of match fit inclusion better than exclusion in my mind. Albeit that's a very fine line.

Here's another angle. If I have a `.only` test and I invoke AVA using a `--match`, what is the expected behavior? IMO `--match` should win (just like CLI flags override `package.json` config). That would be easier to achieve if the match implementation changes the `exclusive` field of the test metadata. Non-matching exclusive tests need to have that field set to `false`.

kasperlewau:
> If I have a .only test and I invoke AVA using a `--match`, what is the expected behavior? IMO `--match` should win. 

Agreed. It's seldom that I pass CLI args to my test runner, and when I do pass them I expect them to be king of the hill.

I'm not so fussed about the easier way *(oh gosh, nitpicking...)* to achieve the end goal, but more so that the produced code is *simpler*. If switching to `exclusive` rather than `skip` to achieve CLI dominance would generate *simpler* code in the end, I'm all for it :+1:

novemberborn:
I think it might make it easier to reason about the side-effects, e.g. in reporter output. Both `--match` and `.only` can cause a lot of tests not to run that would have run otherwise, unlike `.skip` which only causes skipped tests not to run.

sindresorhus:
@novemberborn Does this still apply https://github.com/sindresorhus/ava/pull/592#discussion_r54730961 ?

sindresorhus:
@novemberborn Is the `exclusive` change something you want to do in this PR or can it be a follow-up?

novemberborn:
@sindresorhus,

> Does this still apply #592 (comment) ?

That's been resolved.

> Is the exclusive change something you want to do in this PR or can it be a follow-up?

I think we should do it here, shouldn't be too much work. As the PR stands now, if you have a `.only` test and use `--match` it will only be applied to `.only` tests. I think that's a bug.

sindresorhus:
> if you have a .only test and use --match it will only be applied to .only tests. I think that's a bug.

Agreed.

kasperlewau:
I've changed some stuff around to add the `exclusive` flag onto the given test if its title matches `--match`, leaving the `skipped` flag for `.skip`. 

As expected, non-matching tests no longer add to `testCount`, nor `skipCount`. That *could* open up for some confusion, as it's a fine line and a very personal opinion on whether a match pattern *excludes* or *includes*, but perhaps that is outside the scope what we want to achieve.

Should we introduce another counter to keep track of the tests that were not run due to the `--match` flag? `ignoreCount`? 

Sounds a lot like `skipCount` to me :unamused:, plus the fact that now we'd have to juggle even more words around; `skip, only, match, exclusive, ignore`. Gut feeling says that *could* be simplified, I'm just not quite sure how. 

--- 

Also, consider the following: 

`ava --match=*oo`
```js
test.skip('moo', t => {
  t.pass();
});

test('foo', t => {
  t.pass();
});
```

If `--match` takes precedence over `.only`, should it also override `.skip`?

novemberborn:
> As expected, non-matching tests no longer add to testCount, nor skipCount. That could open up for some confusion, as it's a fine line and a very personal opinion on whether a match pattern excludes or includes, but perhaps that sort of balancing act is outside the scope of what we're trying to achieve here.
> 
> Should we introduce another counter to keep track of the tests that were not run due to the --match flag? ignoreCount?
> 
> Sounds a lot like skipCount to me , plus the fact that now we'd have to juggle even more words around; skip, only, match, exclusive, ignore. Gut feeling says that could be simplified, I'm just not quite sure how.

#471 covers this. Hopefully we can solve it in one fell swoop for all cases.

> If --match takes precedence over .only, should it also override .skip?

I'd argue no. You'd use `.skip` because a test is broken. You'd use `.only` or `--match` to select specific tests.

---

> I've changed some stuff around to add the exclusive flag onto the given test if its title matches --match, leaving the skipped flag for .skip.

Doesn't look like you pushed any commits yet though? No worries if you were waiting for feedback on your questions, but figured you might have forgotten ;-)

kasperlewau:
> #471 covers this. Hopefully we can solve it in one fell swoop for all cases.

Gotcha. I mean something like that could become a mess quite quickly. Glad it's on the roadmap :) 

> I'd argue no. You'd use .skip because a test is broken. You'd use .only or --match to select specific tests.

I suppose you're right about that one. I was halfway through an explanation on why I feel it *should be so*, but halfway through I realised I ended up at the same conclusion - *no.* 

> Doesn't look like you pushed any commits yet though? No worries if you were waiting for feedback on your questions, but figured you might have forgotten ;-)

I'd say that was a 50/50. I was curious as to if it made sense from those point of views, and my net connection died on the train back home as I was about to push. Hadn't gotten around to it yet :smile:

kasperlewau:
Rebased and good to go from my end. Not sure I nailed the `exclusive` flag, so I would love to get another checkup on this PR. 

Just waiting for those CI's to finish their business.

lib/runner.js
```diff
@@ -69,6 +71,10 @@ optionChain(chainableMethods, function (opts, title, fn) {
 		opts.serial = true;
 	}
 
+	if (opts.type === 'test' && title !== null && (this._match && this._match.length > 0)) {
```
novemberborn: No need for the parentheses around the match check.

lib/runner.js
```diff
@@ -99,7 +105,8 @@ Runner.prototype._addTestResult = function (result) {
 		error: result.reason,
 		type: test.metadata.type,
 		skip: test.metadata.skipped,
-		todo: test.metadata.todo
+		todo: test.metadata.todo,
+		exclusive: test.metadata.exclusive
```
novemberborn: Why is this necessary? Presumably `.only` already works without needing this prop.

readme.md
```diff
@@ -134,6 +135,32 @@ Directories are recursive by default. Directories named `fixtures` and `helpers`
 
 When using `npm test`, you can pass positional arguments directly `npm test test2.js`, but flags needs to be passed like `npm test -- --verbose`.
 
+### The `--match` flag
```
novemberborn: I think you should move this down below the `Only-tests` section. Maybe retitle to `Running test that match a pattern`. Should mention that this is more specific than `.only`.

readme.md
```diff
@@ -134,6 +135,32 @@ Directories are recursive by default. Directories named `fixtures` and `helpers`
...
+
+The `--match` flag allows you to run a subset of the tests in your suite that have a matching title. This is achieved with simple wildcard patterns. For more information, check out [sindresorhus/matcher](https://github.com/sindresorhus/matcher).
```
novemberborn: Drop the `sindresorhus/` part from the link title.

readme.md
```diff
@@ -134,6 +135,32 @@ Directories are recursive by default. Directories named `fixtures` and `helpers`
...
+$ ava --match='!*foo*'
+
+# match titles starting with 'foo' and ending with 'bar'
+$ ava --match='foo*bar'
+
+# match titles starting with 'foo' or ending with 'bar'
+$ ava --match='foo*' --match='*bar'
+```
+
+*Tests without a title will not run if* **any** *match pattern is supplied.*
```
novemberborn: Maybe "without an explicit title".

test/cli.js
```diff
@@ -156,6 +156,13 @@ test('watcher works', function (t) {
 	});
 });
 
+test('--match works', function (t) {
+	execCli(['-m=foo', '-m=bar', '-m=!baz', '-m=t* a* f*', '-m=!t* a* n* f*', 'fixture/matcher-skip.js'], function (err) {
+		t.ifError(err);
```
novemberborn: Nice!

novemberborn:
LGTM! I'm in the midst of overhauling the readme (see #597) so don't worry too much about those changes.

lib/runner.js
kasperlewau: It's not necessary. That's me mixing up two wholly different reasons for the full test suite failing. Removing in a jiffy!

lib/runner.js
kasperlewau: Nope, not needed. This would be a case of me not killing my darlings. The amount of times I've been recommended to go and check out lisp is bordering on the ridiculous.

readme.md
kasperlewau: Sure thing!

readme.md
kasperlewau: :+1:

readme.md
kasperlewau: Yup, that's better.

test/cli.js
kasperlewau: : )

kasperlewau:
@novemberborn I agree with the changes you proposed to the Readme. Would you like for me to handle those, or defer to #597?

readme.md
kasperlewau: Hm. How's about `Matched-tests`? Too succinct? 

Thought it'd look better to keep in line with the other 'test scope' modifiers; 
`Serial-tests`, `Only-tests`, `Skip-tests`, `Todo-tests`

novemberborn:
> I agree with the changes you proposed to the Readme. Would you like for me to handle those, or defer to #597?

Seems fine as it is now. I'll give it another read within the context of the changes in #597 and adjust as necessary once this is in `master`.

Not sure why Travis failed in Node 5.

kasperlewau:
> Not sure why Travis failed in Node 5.

Beats me :confused: Can we trigger a rerun and see what's up?

novemberborn:
> Can we trigger a rerun and see what's up?

Just did. Didn't realize I had that power :muscle:

kasperlewau:
> Just did. Didn't realize I had that power :muscle:

And with that, I will calmly await any further feedback/instructions. Flex them muscles, you re-runner you! 

Aside; I was halfway past throwing my MBP right out the window in my previous PR. Not having the power to trigger a re-run had me shouting in agony, being someone who loves to push with `-f` and pull with `--rebase`. 

:beer:

sindresorhus:
> Not having the power to trigger a re-run had me shouting in agony

Just amend the latest commit without any new changes and force push and you trigger a rerun ;)

kasperlewau:
> Just amend the latest commit without any new changes and force push and you trigger a rerun ;)

See I'm *quite sure* I did just that, but it simply wouldn't rebuild on Travis/AppVeyor, which struck me as odd as I am an avid wielder of the `--force`. Nevertheless, we be green now :smile:

novemberborn:
@sindresorhus @vdemedes @jamestalmage LGTY?

novemberborn:

kasperlewau:
> Please make sure to rebase on master to fix a linting issue and run the latest XO configuration against your PR. Thanks!

Done!

sindresorhus:
LGTM2. Landed.

Superb work on this @kasperisager, and thanks for persevering through our extensive review. This turned out really good! :)

kasperlewau:
My pleasure! And thanks :) 

Sidenote; Didn't get no notification. As it turns out, there actually *could be* too many Kasper's around @novemberborn @sindresorhus :smiley:!

facepalm

PR #600, opened 2016-03-04
Fix AVA hanging when `test` is used incorrectly

### LINKED ISSUE #574: AVA hangs when `test` is used incorrectly
Did some preliminary testing and would seem that code like this causes AVA to hang:

```js
import test from 'ava'

test()
```

Incorrect invocations like these should have better user feedback too.

sotojuan:
Fixes #574.

This is just what works now, I am sure there's a better way. Stack trace is [ugly](http://i.imgur.com/YPkqfaq.png) but it's better than hanging. I also had to update some `TestCollection` test stuff so those could still pass.

Just wanted to share my approach!

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -65,6 +65,10 @@ optionChain(chainableMethods, function (opts, title, fn) {
 		throw new TypeError('Expected a function. Use `test.todo()` for tests without a function.');
 	}
 
+	if (typeof fn === 'undefined') {
+		throw new Error('Test must be specified');
+	}
+
 	if (this._serial) {
 		opts.serial = true;
 	}
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -66,7 +66,7 @@ optionChain(chainableMethods, function (opts, title, fn) {
 	}
 
 	if (typeof fn === 'undefined') {
-		throw new Error('Test must be specified');
+		throw new TypeError('Expected a function. Use `test.todo()` for tests without a function.');
 	}
 
 	if (this._serial) {
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -65,7 +65,7 @@ optionChain(chainableMethods, function (opts, title, fn) {
 		throw new TypeError('Expected a function. Use `test.todo()` for tests without a function.');
 	}
 
-	if (typeof fn === 'undefined') {
+	if (typeof fn !== 'function') {
 		throw new TypeError('Expected a function. Use `test.todo()` for tests without a function.');
 	}
 
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -61,11 +61,7 @@ optionChain(chainableMethods, function (opts, title, fn) {
 		if (typeof title !== 'string') {
 			throw new TypeError('`todo` tests require a title');
 		}
-	} else if (opts.skipped && typeof fn !== 'function') {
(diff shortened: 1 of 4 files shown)


lib/test-collection.js
```diff
@@ -68,6 +68,10 @@ TestCollection.prototype.add = function (test) {
 		return;
 	}
 
+	if (typeof test.fn === 'undefined') {
+		throw new Error('Test must be specified');
+	}
+
```
BarryThePenguin: Maybe this check would be better suited in [lib/runner.js#L52-L77](https://github.com/sindresorhus/ava/blob/e697629a77b511e069eb0a2991c8ee72cd09034c/lib/runner.js#L52-L77)

sotojuan:
Just tried it, also works, and I wouldn't have to mess with `TestCollection` tests. I'll wait for someone from the core team to confirm though.

novemberborn:
@sotojuan yes, please add the guards in the runner.

novemberborn:

sotojuan:
Um, don't think I did anything to cause the failing in 0.12.

sindresorhus:
LGTM

lib/runner.js
```diff
@@ -65,6 +65,10 @@ optionChain(chainableMethods, function (opts, title, fn) {
 		throw new TypeError('Expected a function. Use `test.todo()` for tests without a function.');
 	}
 
+	if (typeof fn === 'undefined') {
```
novemberborn: Maybe `typeof fn !== 'function'`?

lib/runner.js
```diff
@@ -65,6 +65,10 @@ optionChain(chainableMethods, function (opts, title, fn) {
...
+		throw new Error('Test must be specified');
```
novemberborn: Could use the same error as in line 65 I think:

> Expected a function. Use `test.todo()` for tests without a function.

lib/runner.js
```diff
@@ -65,6 +65,10 @@ optionChain(chainableMethods, function (opts, title, fn) {
 		throw new TypeError('Expected a function. Use `test.todo()` for tests without a function.');
 	}
 
+	if (typeof fn !== 'function') {
+		throw new TypeError('Expected a function. Use `test.todo()` for tests without a function.');
```
sindresorhus: You now have the exact same error defined twice. Put it in a variable and reuse it.

novemberborn:
@sotojuan could you please add a comment when you push changes? There are no notifications for pushes.

lib/runner.js
```diff
@@ -65,6 +65,10 @@ optionChain(chainableMethods, function (opts, title, fn) {
 		throw new TypeError('Expected a function. Use `test.todo()` for tests without a function.');
```
sindresorhus: Instead of adding a new error message, why don't we just remove `opts.skipped && ` from here? Unless I'm missing something? // @novemberborn

novemberborn: Yea that should be fine.

Would be good to have a test for this new behavior as well.

sotojuan:
@novemberborn Sorry about that, I always forget! I'll make the changes and add the tests later today :-)

sotojuan:
@novemberborn ping :baby_chick:

sindresorhus:
Thanks @sotojuan :)

PR #603, opened 2016-03-05
try to fix flaky api test

novemberborn:
This test has failed twice now, see <https://travis-ci.org/sindresorhus/ava/jobs/113902915>. I can't reproduce the error locally, perhaps there is an error that's being swallowed.

Return the promise so tap can report any errors and isn't somehow tempted to end the test early.


sindresorhus:
LGTM

PR #597, opened 2016-03-03
Readme improvements

novemberborn:
I was going to document watch mode but then realized I hadn't really read the README in a while. I figured I'd use my fresh set of eyes to try and improve it instead.

I've tried to document rationale for each change in the individual commits. I'll squash these into a single commit before merging. Happy to wait until some of the outstanding PRs are in to avoid those having to be rebased again.

Hopefully this version of the readme has some better examples and brings various nuances and hidden behaviors to the fore. Let me know what else I should change, or revert :smile: 

P.S. Apologies to the awesome translators for this massive update. You folks rock.

PR #607, opened 2016-03-06
Document watch mode

### LINKED ISSUE #532: Document --watch
> I'm thinking we add it undocumented, dogfood it, and only document it when we feel it's stable enough and we're ready to take on the addition support burden.

https://github.com/sindresorhus/ava/issues/70#issuecomment-173746486

* [ ] Document the `--watch` and `--source` CLI flags
* [ ] Add to README
* [ ] Mention `DEBUG=ava:watcher`

novemberborn:
Document watch mode. Fixes #532.

Contains a squashed #597, so just check the last commit. Will rebase once #597 has landed.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -55,7 +55,7 @@ var cli = meow([
 	'  --no-cache       Disable the transpiler cache',
 	'  --match, -m      Only run tests with matching title (Can be repeated)',
 	'  --watch, -w      Re-run tests when tests and source files change',
-	'  --source         Pattern to match source files so tests can be re-run (Can be repeated)',
+	'  --source, -S     Pattern to match source files so tests can be re-run (Can be repeated)',
 	'',
 	'Examples',
 	'  ava',
@@ -87,8 +87,9 @@ var cli = meow([
 		v: 'verbose',
 		r: 'require',
 		s: 'serial',
+		m: 'match',
 		w: 'watch',
-		m: 'match'
+		S: 'source'
 	}
 });
 
--- a/cli.js
+++ b/cli.js
@@ -54,9 +54,8 @@ var cli = meow([
 	'  --verbose, -v    Enable verbose output',
 	'  --no-cache       Disable the transpiler cache',
 	'  --match, -m      Only run tests with matching title (Can be repeated)',
-	// Leave --watch and --sources undocumented until they're stable enough
-	// '  --watch, -w      Re-run tests when tests and source files change',
-	// '  --source         Pattern to match source files so tests can be re-run (Can be repeated)',
+	'  --watch, -w      Re-run tests when tests and source files change',
+	'  --source         Pattern to match source files so tests can be re-run (Can be repeated)',
 	'',
 	'Examples',
 	'  ava',
--- a/docs/recipes/watch-mode.md
+++ b/docs/recipes/watch-mode.md
@@ -69,7 +69,7 @@ Dependency tracking works for required modules. Custom extensions and transpiler
 
 ## Manually rerunning all tests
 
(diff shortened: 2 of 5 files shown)


novemberborn:
Pushed a fixup which:

* changes `sports` to `comes with`
* links to `ignore-by-default`'s `index.js` for the ignored directories
* uses `<kbd>` to mark up the `rs` input, reworded the sentence as well
* updated example source patterns

Still to do:

* [ ] add `-S` as shorthand for `--source`
* [ ] support `r + Enter` to rerun all tests, in addition to `rs + Enter`
* [ ] resolve whether source negation patterns should disable all default ignored directories, whether those default directories can be overriden, etc (this may need to be a follow-up issue)

sindresorhus:
>  support r + Enter to rerun all tests, in addition to rs + Enter

I meant just `r`. No Enter. What really is the point of having to press Enter?

> resolve whether source negation patterns should disable all default ignored directories, whether those default directories can be overriden, etc (this may need to be a follow-up issue)

Yeah, let's not dilute this PR with that. Can you open a new issue?

novemberborn:
> I meant just r. No Enter. What really is the point of having to press Enter?

This may be a bit tricky. It requires enabling raw mode on `stdin` which then seems to intercept ^C etc. It also seemed to affect my shell after killing AVA. Happy to open a ticket though as a future improvement?

I've added `r + Enter` now. I'm inclined to keep `rs + Enter` for those users with nodemon muscle memory (or who may use AVA and nodemon in the same project). Can remove if others disagree.

> Yeah, let's not dilute this PR with that. Can you open a new issue?

#614

---

This has now been rebased against master. `-S` shorthand has been added.

sindresorhus:
> Happy to open a ticket though as a future improvement?

:+1:

> I'm inclined to keep rs + Enter for those users with nodemon muscle memory (or who may use AVA and nodemon in the same project).

Sure, that's fine, but `r+enter` should be the documented one.

novemberborn:
> Sure, that's fine, but r+enter should be the documented one.

Right now they're both documented, with `r + Enter` being first.

sindresorhus:
> Right now they're both documented, with r + Enter being first.

I don't see the point of documenting both. Your argument is that you have `rs` + Enter as muscle memory. I don't see why we would document `rs` to new users without that muscle memory and create overhead for them by making them have to choose which one to pick.

novemberborn:
Ticket: #615

novemberborn:
> I don't see the point of documenting both. Your argument is that you have rs + Enter as muscle memory. I don't see why we would document rs to new users without that muscle memory and create overhead for them by making them have to choose which one to pick.

How would users realize `rs` also works? Other than accidentally. Maybe that's good enough though?

sindresorhus:
> How would users realize rs also works? Other than accidentally. Maybe that's good enough though?

Isn't that the point? If they have to read the docs, they can just as easily use the shorter `r`+Enter version.

novemberborn:
You'd have to read the docs to even find out about r+enter though.

sindresorhus:
> You'd have to read the docs to even find out about r+enter though.

Exactly. Unless it's muscle memory, users have to read the docs anyways, so there's really no reason to document both behaviors, when `r` is shorter. I consider `rs` a graceful handling of a mistyping, not something we'd want to officially support. IMHO, we should only have [one way to do something](https://www.python.org/dev/peps/pep-0020/).

novemberborn:
OK will edit in the morning. Good to go otherwise?

sindresorhus:
:+1:

novemberborn:
Updated. Anything else?

sindresorhus:
Looks good :)

Issue #606, opened 2016-03-06
--match runs all tests when there are no matches


novemberborn:
When the new `--match` option is used and no test actually matches, all tests are run. This seems wrong. It shouldn't run any tests.

Here's a rough idea on how this could be implemented:

* In `runner.js`, if there is no match but the test is exclusive, unset the `exclusive` option ([see code](https://github.com/sindresorhus/ava/blob/94b352cc08673feac3ccbab912879bcd1865d92b/lib/runner.js#L75))
* In `api.js` when all tests are known, if the `match` option was given but there are no exclusive tests ([see code](https://github.com/sindresorhus/ava/blob/94b352cc08673feac3ccbab912879bcd1865d92b/api.js#L207)) bail out with a helpful error message ([see code](https://github.com/sindresorhus/ava/blob/94b352cc08673feac3ccbab912879bcd1865d92b/api.js#L178))

The `runner.js` change means we can rely on `hasExclusive` being true in `api.js` only if there are matching tests.

novemberborn:
@kasperlewau are you interested in fixing this? No worries if not :smile:

kasperlewau:
I'm up for getting this sorted out, although I'm not sure how much time I will be able to dedicate to the cause over the coming weeks. So I would leave the issue unassigned for now and/or assign someone else who may feel up for it! 

If I do get enough time over I'll be sure to push often, and keep this thread going - so as to help out in any way possible :).

PR #612, opened 2016-03-07
Don't run tests if --match has no matches

novemberborn:
First commit does some refactoring and fixes a rare race condition where tests are run prematurely.

Second commit implements my suggestions from #606.


vdemedes:
I think this is good to go, thanks @novemberborn!

sindresorhus:
LGTM

PR #613, opened 2016-03-07
fix #342 set child process debug port to available

### LINKED ISSUE #342: Debugger support

1. `ava --init`
2. `npm install signal-exit`
   ```javascript
#!/usr/bin/env node
'use strict';

var chalk = require('chalk');

    console.dir(data);
};

onExit(function () {
});

require('./lib/babel');
``` 
  ```javascript
import test from 'ava';

test(t => {
    debugger;
    t.same([4, 2], [1, 2]);
});
```

develar:
We decided to support AVA debug (#342) without any efforts [on your side](https://github.com/nodejs/node/pull/5025#issuecomment-177985896).

But:
* [debugger: bind to random port with --debug-port=0 #5025](https://github.com/nodejs/node/pull/5025) is not yet merged into NodeJS.
* IntelliJ JS Debugger doesn't support yet `=0`, and when will be supported, it will take time to be available for users in the generally available release.

Unfortunately, although idea of fix is very simple (`getPort().then`), change is complex because existing code adds custom context-dependent methods to Promise. And Github diff view is very unclear compared to IntelliJ IDEA.

What's done:
* if `debug-brk=` or `debug=` in the main process args, find available port and change debug arg.
* `fork` now returns object with custom methods (`on`, `run`) and `promise` field. We don't add custom methods to `promise` anymore.
  * to simplify test code (and reduce changes), method `then` added to returned object. Test-only — in production we use `promise` field.
  * `send` is not used, as far I see, but I didn't delete it. Should I do it?

Please note — as it stated [in my comment](https://github.com/sindresorhus/ava/issues/342#issuecomment-177239663), debugging will be still not possible due to another issues. Solving `sourcemap` issue is out of this PR scope.

PS: AVA is so awesome, that multi-process debug was reimplemented in the IntelliJ JS Debugger to avoid multiple debug tabs (IntelliJ Platform 146+).

### DIFF
--- a/api.js
+++ b/api.js
@@ -14,6 +14,7 @@ var resolveCwd = require('resolve-cwd');
 var uniqueTempDir = require('unique-temp-dir');
 var findCacheDir = require('find-cache-dir');
 var slash = require('slash');
+var getPort = require('get-port');
 var AvaError = require('./lib/ava-error');
 var fork = require('./lib/fork');
 var formatter = require('./lib/enhance-assert').formatter();
@@ -61,19 +62,44 @@ Api.prototype._reset = function () {
 	this.base = '';
 };
 
-Api.prototype._runFile = function (file) {
+Api.prototype._runFile = function (file, onForkStarting) {
 	var options = objectAssign({}, this.options, {
 		precompiled: this.precompiler.generateHashForFile(file)
 	});
 
-	return fork(file, options)
-		.on('teardown', this._handleTeardown)
-		.on('stats', this._handleStats)
-		.on('test', this._handleTest)
-		.on('unhandledRejections', this._handleRejections)
-		.on('uncaughtException', this._handleExceptions)
-		.on('stdout', this._handleOutput.bind(this, 'stdout'))
-		.on('stderr', this._handleOutput.bind(this, 'stderr'));
+	var execArgv = process.execArgv.slice();
+	var debugArgIndex = -1;
+	for (var i = 0; i < execArgv.length; i++) {
+		if (execArgv[i].indexOf('--debug-brk=') === 0 || execArgv[i].indexOf('--debug=') === 0) {
+			debugArgIndex = i;
+			break;
+		}
+	}
+
+	var execArgvPromise;
+	if (debugArgIndex === -1) {
+		execArgvPromise = Promise.resolve(execArgv);
+	} else {
+		execArgvPromise = getPort()
+			.then(function (port) {
+				execArgv[debugArgIndex] = '--debug-brk=' + port;
+				return execArgv;
+			});
+	}
+
+	return execArgvPromise
+		.then(function (execArgv) {
+			var result = fork(file, options, execArgv)
+				.on('teardown', this._handleTeardown)
+				.on('stats', this._handleStats)
+				.on('test', this._handleTest)
+				.on('unhandledRejections', this._handleRejections)
+				.on('uncaughtException', this._handleExceptions)
+				.on('stdout', this._handleOutput.bind(this, 'stdout'))
+				.on('stderr', this._handleOutput.bind(this, 'stderr'));
+			onForkStarting(result);
+			return result.promise;
+		}.bind(this));
 };
 
 Api.prototype._handleOutput = function (channel, data) {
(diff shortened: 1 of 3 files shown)


sindresorhus:
Tip for reviewing this, append `?w=1` to the diff URL to ignore whitespace changes: https://github.com/sindresorhus/ava/pull/613/files?w=1

api.js
```diff
@@ -18,6 +18,7 @@ var AvaError = require('./lib/ava-error');
 var fork = require('./lib/fork');
 var formatter = require('./lib/enhance-assert').formatter();
 var CachingPrecompiler = require('./lib/caching-precompiler');
+var getPort = require('get-port');
```
sindresorhus: This should be placed above the relative requires (on line 17).

api.js
```diff
@@ -61,19 +62,44 @@ Api.prototype._reset = function () {
...
-		.on('stats', this._handleStats)
-		.on('test', this._handleTest)
-		.on('unhandledRejections', this._handleRejections)
-		.on('uncaughtException', this._handleExceptions)
-		.on('stdout', this._handleOutput.bind(this, 'stdout'))
-		.on('stderr', this._handleOutput.bind(this, 'stderr'));
+	var execArgv = process.execArgv.slice();
+	var debugArgIndex = -1;
+	for (var i = 0; i < execArgv.length; i++) {
+		if (execArgv[i].indexOf('--debug-brk=') === 0 || execArgv[i].indexOf('--debug=') === 0) {
```
sindresorhus: Can't you just use the https://github.com/sindresorhus/has-flag dependency here? (we already depend on it)

api.js
```diff
@@ -191,44 +217,57 @@ Api.prototype.run = function (files) {
...
-										tests: []
-									};
+				// receive test count from all files and then run the tests
+				var statsCount = 0;
+
+				var tests = new Array(files.length);
+				files.forEach(function (file, index) {
+					self._runFile(file, function (forkManager) {
+						tests[index] = forkManager;
+						forkManager.on("stats", tryRun);
```
sindresorhus: single-quotes

api.js
develar: Fixed.

sindresorhus:
Nice! Thanks for working on this @develar :)

sindresorhus:
Can you add some tests?

api.js
develar: Fixed (hmm... linter doesn't throw error).

api.js
develar: `has-flag` returns boolean, but we need position of debug argument. In the future, es6 `includes` can be used here to avoid `=== 0`.

sindresorhus: I don't understand why you need the position? These are flag arguments, not positional arguments, and can be in any order.

sindresorhus:
> fork now returns object with custom methods (on, run) and promise field. We don't add custom methods to promise anymore.

What's the benefit of this?

api.js
develar: Because we don't append flag, but change existing. By default `child_process.fork` pass `process.execArgv` to child process as is. What we do here — we find debug argument and change its value to any free port.

develar:
>> fork now returns object with custom methods (on, run) and promise field. We don't add custom methods to promise anymore.

> What's the benefit of this?

My first approach was just add `getPort().then` And then I realized, that existing code doesn't work with nested promise — `.on is not a function`. But we cannot return promise from `fork` directly until promise of `getPort()` is not resolved — so, we return nested promise.

novemberborn:
> send is not used, as far I see, but I didn't delete it. Should I do it?

It's used as of a few hours ago :stuck_out_tongue_winking_eye: <https://github.com/sindresorhus/ava/blob/1868204c1901f45b4f66a520ef6486fdd71fe1d2/api.js#L210>

api.js
```diff
@@ -191,44 +217,57 @@ Api.prototype.run = function (files) {
...
+				// receive test count from all files and then run the tests
+				var statsCount = 0;
+
+				var tests = new Array(files.length);
+				files.forEach(function (file, index) {
+					self._runFile(file, function (forkManager) {
+						tests[index] = forkManager;
+						forkManager.on('stats', tryRun);
+					})
+					.catch(function (error) {
```
novemberborn: This is asynchronous whereas `onForkStarting` is not. Probably best to build up `tests` similarly (and synchronously) to how it was done before and leave the rest of the code as it is. (Note that I just changed this you so you'll need to rebase).

lib/fork.js
```diff
@@ -134,18 +123,31 @@ module.exports = function (file, opts) {
...
-		});
+			ps.on('stats', function () {
+				send(ps, 'run', options);
+			});
 
-		return promise;
+			return this;
+		},
+		then: function () {
+			// only to simplify test code, in production use .promise
```
novemberborn: We should just bite the bullet and use `.promise` in the tests as well.

api.js
novemberborn: Perhaps extract this into a memoizing function? Less logic in this function and no need to recompute for every file.

sindresorhus: I think memoization would be a premature optimization for this as execArv usually contains zero or one elements.

api.js
```diff
@@ -61,19 +62,44 @@ Api.prototype._reset = function () {
...
-		.on('teardown', this._handleTeardown)
-		.on('stats', this._handleStats)
-		.on('test', this._handleTest)
-		.on('unhandledRejections', this._handleRejections)
-		.on('uncaughtException', this._handleExceptions)
-		.on('stdout', this._handleOutput.bind(this, 'stdout'))
-		.on('stderr', this._handleOutput.bind(this, 'stderr'));
+	var execArgv = process.execArgv.slice();
+	var debugArgIndex = -1;
+	for (var i = 0; i < execArgv.length; i++) {
```
sindresorhus: Just use Array#forEach here.

api.js
sindresorhus: This also doesn't handle the case where the user would do `--debug-brk 213` instead of `--debug-brk=213`.

api.js
```diff
@@ -61,19 +62,44 @@ Api.prototype._reset = function () {
 	this.base = '';
 };
 
-Api.prototype._runFile = function (file) {
+Api.prototype._runFile = function (file, onForkStarting) {
```
jamestalmage: This is a weird combination of callbacks and promises. Can we clean it up to just use promises?

api.js
```diff
@@ -191,44 +217,57 @@ Api.prototype.run = function (files) {
...
+
+				var tests = new Array(files.length);
+				files.forEach(function (file, index) {
+					self._runFile(file, function (forkManager) {
+						tests[index] = forkManager;
+						forkManager.on('stats', tryRun);
+					})
+					.catch(function (error) {
+						tests[index] = {
+							run: function () {
```
jamestalmage: Why would `run` just return `this`? Shouldn't it run something?

jamestalmage:
> And then I realized, that existing code doesn't work with nested promise — .on is not a function.

You might be able to solve this with [`promise-delegates`](https://github.com/jamestalmage/promise-delegates).

novemberborn:
Hey @develar, would be great to have debug support for IntelliJ. Will you have time to follow up on the feedback here?

develar:
@novemberborn Sorry for delay, it will be addressed this week.

novemberborn:
@develar no worries, that's great! :+1:

vdemedes:
Closing due to lack of response from the author of PR.

jamestalmage:
@develar - If you ever get around to updating this PR, we would still be interested in merging.  However, be aware, there is work happening towards official Node debugger support: https://github.com/nodejs/node/issues/2546#issuecomment-189311746.

develar:
New PR — #874.

PR #618, opened 2016-03-08
remove color from test title in mini reporter

sindresorhus:
too much green when the result count is also in green

and with `todo` and `skip` tests it doesn't matter as they're shown too fast anyways

### DIFF
--- a/lib/reporters/mini.js
+++ b/lib/reporters/mini.js
@@ -70,19 +70,16 @@ MiniReporter.prototype.prefix = function (str) {
 };
 
 MiniReporter.prototype._test = function (test) {
-	var title;
+	var title = test.title;
 
 	if (test.todo) {
-		title = colors.todo('- ' + test.title);
 		this.todoCount++;
 	} else if (test.skip) {
-		title = colors.skip('- ' + test.title);
 		this.skipCount++;
 	} else if (test.error) {
 		title = colors.error(test.title);
 		this.failCount++;
 	} else {
-		title = colors.pass(test.title);
 		this.passCount++;
 	}
 
--- a/test/reporters/mini.js
+++ b/test/reporters/mini.js
@@ -120,7 +120,7 @@ test('skipped test', function (t) {
 
 	var expectedOutput = [
 		' ',
-		' ⠋ ' + chalk.yellow('- skipped'),
+		' ⠋ ' + chalk.yellow('skipped'),
 		'',
 		'   ' + chalk.yellow('1 skipped')
 	].join('\n');
@@ -140,7 +140,7 @@ test('todo test', function (t) {
 
 	var expectedOutput = [
 		' ',
-		' ⠋ ' + chalk.blue('- todo'),
+		' ⠋ ' + chalk.blue('todo'),
 		'',
 		'   ' + chalk.blue('1 todo')
 	].join('\n');


lib/reporters/mini.js
```diff
@@ -70,19 +70,16 @@ MiniReporter.prototype.prefix = function (str) {
 };
 
 MiniReporter.prototype._test = function (test) {
-	var title;
+	var title = test.title;
 
 	if (test.todo) {
-		title = colors.todo('- ' + test.title);
 		this.todoCount++;
```
sindresorhus: Why do we manually count these in the reporter? Shouldn't it be provided by the API?

novemberborn: I've wondered that too.

lib/reporters/mini.js
```diff
@@ -70,19 +70,16 @@ MiniReporter.prototype.prefix = function (str) {
 };
 
 MiniReporter.prototype._test = function (test) {
-	var title;
+	var title = test.title;
 
 	if (test.todo) {
-		title = colors.todo('- ' + test.title);
```
novemberborn: Shouldn't you keep the `- ` prefix?

sindresorhus: No point. The test is not run anyways, so it finishes faster than a human could see.

lib/reporters/mini.js
vdemedes: Agree, that should be changed.

lib/reporters/mini.js
vdemedes: If a human can not see `todo` and `skip` tests, shouldn't we also remove color from there?

vdemedes: But wait, if the next test is slow, user theoretically can actually see skip/todo tests. So I guess we should keep the prefix and color. 

Also, what about changing prefixes to these?

- if `todo` test, prefix = `To-do: `
- if `skip` test, prefix = `Skipped: `

sindresorhus: > If a human can not see todo and skip tests, shouldn't we also remove color from there?

That's what this PR does.

> But wait, if the next test is slow, user theoretically can actually see skip/todo tests. So I guess we should keep the prefix and color.

How about we just not show them all together? So if a slow test is running, the last actual test is shown, not todo/skip.

vdemedes: > That's what this PR does.

Oh my God, what an embarrassment, I got confused with the new "Files" view and read the wrong column in diff...

> How about we just not show them all together? So if a slow test is running, the last actual test is shown, not todo/skip.

Even better.

novemberborn: > How about we just not show them all together? So if a slow test is running, the last actual test is shown, not todo/skip.

Yep!

jamestalmage:
Just a thought, but what if we showed bothe the current test, and the last failed test. Or maybe just the counts and the last failed test (Since we intend to print pending tests on SIGINT - the last passed test isn't that valuable anymore). What failed is usually the most important info anyways.

lib/reporters/mini.js
jamestalmage: As I recall, the API does track this, but it's not provided until the `finish` event. The mini reporter is the only reporter that cares about the live count.

sindresorhus:
Let's continue this in https://github.com/sindresorhus/ava/issues/628.

PR #621, opened 2016-03-08
Make .only sticky in watch mode

### LINKED ISSUE #593: Make .only sticky in watch mode
#556 makes .only work across all test files. However this won't work the same when in watch mode since only a subset of the test files are run when changes occur.

Let's say we have two test files: `foo.js` and `bar.js`. The latter contains a `.only`.

Initially all test files are run and because `bar.js` contains a `.only`, only that test is actually executed.

Subsequently `foo.js` is changed. The watcher will run just `foo.js` which does not contain a `.only`. Now all tests in `foo.js` are executed.

Later a non-test file is changed and all test files are rerun. Only the `.only` test from `bar.js` is executed.

This isn't necessarily bad but it may be surprising to some. I wonder if `.only` should be sticky so that when changes are detected the watcher still only executes `.only` tests. But then if you remove the `.only` it starts executing all tests again. Similarly adding a `.only` switches back into the sticky mode.

We could build upon #544 which already performs some per-test-file bookkeeping.

Thoughts?

novemberborn:
Fixes #593.

The watcher keeps track of exclusive tests. If all test files that contained exclusive tests need to be rerun it runs them without forcing exclusive mode. This means the exclusivity is determined by the tests themselves.

If a test file, containing exclusive tests, is not one of the files being rerun, it forces exclusive mode. This ensures only exclusive tests are run in the changed files, making .only sticky.

If all test files that contained exclusive tests are removed, sticky mode is disabled. The same happens if there are no more exclusive tests after a run.

A second argument can be passed to Api#run(). If true the tests will be run in exclusive mode, regardless of whether exclusive tests are detected. The Api now remits the 'stats' event from the forks.

### DIFF
--- a/api.js
+++ b/api.js
@@ -106,6 +106,8 @@ Api.prototype._handleTeardown = function (data) {
 };
 
 Api.prototype._handleStats = function (stats) {
+	this.emit('stats', stats);
+
 	if (this.hasExclusive && !stats.hasExclusive) {
 		return;
 	}
@@ -165,10 +167,15 @@ Api.prototype._prefixTitle = function (file) {
 	return prefix;
 };
 
-Api.prototype.run = function (files) {
+Api.prototype.run = function (files, options) {
 	var self = this;
 
 	this._reset();
+
+	if (options && options.runOnlyExclusive) {
+		this.hasExclusive = true;
+	}
+
 	return handlePaths(files, this.excludePatterns)
 		.map(function (file) {
 			return path.resolve(file);
--- a/lib/watcher.js
+++ b/lib/watcher.js
@@ -39,13 +39,28 @@ function Watcher(logger, api, files, sources) {
 	this.isTest = makeTestMatcher(files, api.excludePatterns);
 	this.run = function (specificFiles) {
 		logger.reset();
-		this.busy = api.run(specificFiles || files).then(function () {
+
+		var runOnlyExclusive = false;
+		if (specificFiles) {
+			var exclusiveFiles = specificFiles.filter(function (file) {
+				return this.filesWithExclusiveTests.indexOf(file) !== -1;
+			}, this);
+
+			runOnlyExclusive = exclusiveFiles.length !== this.filesWithExclusiveTests.length;
+		}
+
+		this.busy = api.run(specificFiles || files, {
+			runOnlyExclusive: runOnlyExclusive
+		}).then(function () {
 			logger.finish();
 		}, rethrowAsync);
 	};
 
 	this.testDependencies = [];
 	this.trackTestDependencies(api, sources);
+
+	this.filesWithExclusiveTests = [];
+	this.trackExclusivity(api);
 
 	this.dirtyStates = {};
 	this.watchFiles(files, sources);
@@ -85,12 +100,6 @@ Watcher.prototype.trackTestDependencies = function (api, sources) {
 	});
 };
 
-Watcher.prototype.removeUnlinkedTestDependencies = function (unlinkedTests) {
-	unlinkedTests.forEach(function (testFile) {
-		this.updateTestDependencies(testFile, []);
-	}, this);
-};
-
 Watcher.prototype.updateTestDependencies = function (file, sources) {
 	if (sources.length === 0) {
 		this.testDependencies = this.testDependencies.filter(function (dep) {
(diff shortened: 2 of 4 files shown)


novemberborn:
Note that if you edit a file without exclusive tests, AVA will report `0 tests passed`. Let me know if this needs a more helpful message.

lib/watcher.js
```diff
@@ -113,6 +126,30 @@ Watcher.prototype.updateTestDependencies = function (file, sources) {
...
 };
 
+Watcher.prototype.trackExclusivity = function (api) {
+	var self = this;
+	api.on('stats', function (stats) {
+		self.updateExlusivity(stats.file, stats.hasExclusive);
+	});
+};
+
+Watcher.prototype.updateExlusivity = function (file, hasExclusive) {
```
vdemedes: Exlusivity -> exclusivity

lib/watcher.js
```diff
@@ -113,6 +126,30 @@ Watcher.prototype.updateTestDependencies = function (file, sources) {
...
+	var ix = this.exclusiveTests.indexOf(file);
```
vdemedes: `index` would look way better

vdemedes:
PR looks good @novemberborn, but I feel like we really have to come up with a replacement for `exclusive`-related names. `hasExclusive`, `trackExclusivity`, `updateExclusivity` sound very weird and unclear. At least to me. Let's hear what others think.

api.js
```diff
@@ -165,10 +167,14 @@ Api.prototype._prefixTitle = function (file) {
 	return prefix;
 };
 
-Api.prototype.run = function (files) {
+Api.prototype.run = function (files, forceHasExclusive) {
```
sindresorhus: I think the second argument would be better as an optional options object with an option called `forceHasExclusive`. 

http://ariya.ofilabs.com/2011/08/hall-of-api-shame-boolean-trap.html

api.js
```diff
@@ -165,10 +167,14 @@ Api.prototype._prefixTitle = function (file) {
...
 	var self = this;
 
 	this._reset();
+	if (forceHasExclusive === true) {
```
sindresorhus: Put an empty line above here.

lib/watcher.js
```diff
@@ -39,7 +39,23 @@ function Watcher(logger, api, files, sources) {
...
+		var promise;
+		if (specificFiles) {
+			var exclusiveFiles = specificFiles.filter(function (file) {
+				return this.exclusiveTests.indexOf(file) !== -1;
+			}, this);
+
+			if (exclusiveFiles.length === this.exclusiveTests.length) {
+				promise = api.run(specificFiles);
+			} else {
+				promise = api.run(specificFiles, true);
```
sindresorhus:
```js
promise = api.run(specificFiles, exclusiveFiles.length === this.exclusiveTests.length);
```

lib/watcher.js
```diff
@@ -113,6 +126,30 @@ Watcher.prototype.updateTestDependencies = function (file, sources) {
...
+	});
+};
+
+Watcher.prototype.updateExlusivity = function (file, hasExclusive) {
+	var ix = this.exclusiveTests.indexOf(file);
+
+	if (hasExclusive && ix === -1) {
+		this.exclusiveTests.push(file);
+	} else if (!hasExclusive && ix !== -1) {
+		this.exclusiveTests.splice(ix, 1);
```
sindresorhus: I don't totally get why we're splicing in here at a specific position. Maybe a comment would help.

novemberborn: We're splicing out, actually. It's just a weird API. It seems fairly clear to me that one branch adds the file, if necessary, and the other removes, if necessary.

Could use the `union()` and `diff()` helpers if that would make it clearer?

api.js
novemberborn: Sure. Opted for `runOnlyExclusive` which is the same as the option passed to the runner.

novemberborn:
> I feel like we really have to come up with a replacement for exclusive-related names. hasExclusive, trackExclusivity, updateExclusivity sound very weird and unclear. At least to me. Let's hear what others think.

English is weird :wink: 

`hasExclusive` comes from `TestCollection` and made its way into the `stats` object and Api. "Exclusivity" here refers to the updating the exclusivity of test files. I've changed some of the variable names to be more accurate, hopefully it's clearer now.

novemberborn:
Pushed fixups for the feedback. Some of the changes were slightly more involved so would like feedback before squashing :smile:

sindresorhus:
> Note that if you edit a file without exclusive tests, AVA will report 0 tests passed. Let me know if this needs a more helpful message.

Can you show an example on when this would happen exactly? More helpful messages are always welcome.

novemberborn:
> Can you show an example on when this would happen exactly? More helpful messages are always welcome.

Imagine you have two test files: `test/foo.js` and `test/bar.js`. `foo.js` contains a `.only()`. When you start the watcher it only runs that one test from `foo.js`. Now you edit `bar.js`. It runs the tests from `bar.js` in exclusive mode, but there are no such tests. It outputs `0 tests passed`.

Maybe it should report `0 exclusive tests found`, but that would entail modifying the various reporters.

It'd also help if I added debug output so you can see the tests are being run in exclusive mode.

sindresorhus:
> Maybe it should report 0 exclusive tests found, but that would entail modifying the various reporters.

> It'd also help if I added debug output so you can see the tests are being run in exclusive mode.

Agreed. Mind opening issues for this? Can be follow-up improvements.

novemberborn:
> Agreed. Mind opening issues for this? Can be follow-up improvements.

Sure. Anything else on this PR?

I'm tracking down the intermittent CI failure but that's not related to this PR.

sindresorhus:
Looks good. Land it ;)

novemberborn:
>> Maybe it should report 0 exclusive tests found, but that would entail modifying the various reporters.
> 
> It'd also help if I added debug output so you can see the tests are being run in exclusive mode.
> Agreed. Mind opening issues for this? Can be follow-up improvement

#635 #636

PR #624, opened 2016-03-09
Don't crash watch mode when tests files fail to transpile

### LINKED ISSUE #622: Watch mode crashes when test files fail to transpile
Steps to reproduce:

* Start AVA in watch mode
* Introduce a syntax error, e.g. use `test.('title', t => {})`
* AVA crashes

The error should be logged however AVA should not crash.

novemberborn:
Fixes #622.

Catch exceptions when initially running files. Don't run any tests, just report the exception.

Move test teardown into a callback for the run promise so it can be used when `--match` is used but there are no matches, and when exceptions occur while initially running files.

Teardown causes the test promise to reject, which leads to an attempt to run the tests. Add a guard to prevent this.

### DIFF
--- a/api.js
+++ b/api.js
@@ -192,8 +192,7 @@ Api.prototype.run = function (files) {
 			self.fileCount = files.length;
 			self.base = path.relative('.', commonPathPrefix(files)) + path.sep;
 
-			var tests = files.map(self._runFile);
-
+			var tests = new Array(self.fileCount);
 			return new Promise(function (resolve) {
 				function run() {
 					if (self.options.match.length > 0 && !self.hasExclusive) {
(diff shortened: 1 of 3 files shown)


api.js
```diff
@@ -192,11 +192,7 @@ Api.prototype.run = function (files) {
 			self.fileCount = files.length;
 			self.base = path.relative('.', commonPathPrefix(files)) + path.sep;
 
-			var tests = files.map(self._runFile);
-
-			// receive test count from all files and then run the tests
-			var unreportedFiles = self.fileCount;
-
+			var tests = new Array(self.fileCount);
```
vdemedes: Could you explain why is array of N (`self.fileCount`) null elements needed?

sindresorhus:
LGTM

sindresorhus:
@jamestalmage Looks good to you?

api.js
jamestalmage: It gets filled with promises for run results [here](https://github.com/sindresorhus/ava/blob/watch-and-syntax-errors/api.js#L248)

jamestalmage:
:+1:

api.js
novemberborn: Yea. I'm just preallocating it to the right size.

PR #625, opened 2016-03-09
Test chalk colors in reporter tests

### LINKED ISSUE #605: Our reporter tests don't actually test the chalk colors.
We have lines like:

```js
t.is(output, '   ' + chalk.green('1 passed'));
```

But it still passes if changed to:

```js
t.is(output, '   1 passed');
```

This is because `process.stdout.isTTY` is false in all tests. This is because `tap` forks the process and does not emulate a tty environment in the child process. That will change when we switch to self testing (AVA **does** emulate a tty in child processes if the parent process was a tty).

We should either drop the `chalk.XXX` statements from the tests or force chalk to enable colors.

sotojuan:
Fixes #605. Followed @jamestalmage's suggestions and added/modified things needed for tests to pass.


sindresorhus:
@sotojuan Looks pretty good. Can you fix the merge conflict?

sotojuan:
All right. Had to change some tests to make them pass (one of the recent changes removed the colors of some test titles, is that fine?). Not sure what's up with AppVeyor.

sindresorhus:
:+1: 

Just ignore AppVeyor. It sometimes completely fails to run for no reason...

PR #638, opened 2016-03-11
fix index.d.ts for noImplicitAny option

bouzuya:
```
$ $(npm bin)/tsc --noImplicitAny
node_modules/ava/index.d.ts(132,2): error TS7010: 'throws', which lacks return-type annotation, implicitly has an 'any' return type.
node_modules/ava/index.d.ts(136,2): error TS7010: 'notThrows', which lacks return-type annotation, implicitly has an 'any' return type.
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -129,11 +129,13 @@ export interface AssertContext {
 	 * Assert that function throws an error or promise rejects.
 	 * @param error Can be a constructor, regex, error message or validation function.
 	 */
-	throws(value: (() => void) | Promise<{}>, error?: ErrorValidator, message?: string);
+	throws(value: Promise<{}>, error?: ErrorValidator, message?: string): Promise<any>;
+	throws(value: () => void, error?: ErrorValidator, message?: string): any;
 	/**
 	 * Assert that function doesn't throw an error or promise resolves.
 	 */
-	notThrows(value: (() => void) | Promise<{}>, message?: string);
+	notThrows<U>(value: Promise<U>, message?: string): Promise<U>;
+	notThrows(value: () => void, message?: string): void;
 	/**
 	 * Assert that contents matches regex.
 	 */


index.d.ts
```diff
@@ -129,11 +129,11 @@ export interface AssertContext {
 	 * Assert that function throws an error or promise rejects.
 	 * @param error Can be a constructor, regex, error message or validation function.
 	 */
-	throws(value: (() => void) | Promise<{}>, error?: ErrorValidator, message?: string);
+	throws(value: (() => void) | Promise<{}>, error?: ErrorValidator, message?: string): void;
```
sindresorhus: `throws` return the error thrown though: https://github.com/sindresorhus/ava#throwsfunctionpromise-error-message

ivogabe:
If you change the return type of `throws` to `any`, 👍. The return type of notThrows looks correct to me.

sindresorhus:
@ivogabe There's no good way to indicate that `throws` returns an error type?

SamVerschueren:
@sindresorhus yes with `Error`. But what if someone throws a string or a number? It isn't parsed to an Error object anymore, right?

ivogabe:
Indeed, one might write `throw ""`.

sindresorhus:
Oh yeah. Even though nobody should do that...

SamVerschueren:
> Oh yeah. Even though nobody should do that...

Well, we all know how that goes ;)

bouzuya:
s/void/any/

NOTE: 

`throws` returns thrown object or PromiseLike or void.
https://github.com/sindresorhus/ava/blob/v0.13.0/lib/assert.js#L76-L85
https://github.com/sindresorhus/ava/blob/v0.13.0/lib/assert.js#L95-L106

`notThrows` returns PromiseLike or void.
https://github.com/sindresorhus/ava/blob/v0.13.0/lib/assert.js#L117-L124

document:
https://github.com/sindresorhus/ava#throwsfunctionpromise-error-message
> Returns the error thrown by function or the rejection reason of promise.
E

ivogabe:
More accurate would be:
```typescript
throws(value: Promise<{}>, error?: ErrorValidator, message?: string): Promise<any>;
throws(value: () => void, error?: ErrorValidator, message?: string): any;
notThrows<U>(value: Promise<U>, message?: string): Promise<U>;
notThrows(value: () => void, message?: string): void;
```

sindresorhus:
Thanks all :)

PR #639, opened 2016-03-11
Truncate mini reporter titles

### LINKED ISSUE #629: Mini reporter goes mad when the terminal width is smaller than the test title
To reproduce, make your terminal width narrow and try running the [`got`](https://github.com/sindresorhus/got) tests.

untitled12

sotojuan:
This fixes #629.

Here's how testing `got` looks like:

I just use Sindre's [cli-truncate](https://github.com/sindresorhus/cli-truncate) to truncate the test title if it's  longer than the CLI columns. I think truncating to the column size minus 3 looks nice, but open to change.

### DIFF
--- a/lib/reporters/mini.js
+++ b/lib/reporters/mini.js
@@ -76,9 +76,7 @@ MiniReporter.prototype.prefix = function (str) {
 };
 
 MiniReporter.prototype._test = function (test) {
-	var title = test.title;
-
-	title = cliTruncate(title, process.stdout.columns - 3);
+	var title = cliTruncate(test.title, process.stdout.columns - 3);
 
 	if (test.todo) {
 		this.todoCount++;
--- a/lib/reporters/mini.js
+++ b/lib/reporters/mini.js
@@ -78,9 +78,7 @@ MiniReporter.prototype.prefix = function (str) {
 MiniReporter.prototype._test = function (test) {
 	var title = test.title;
 
-	if (title.length > process.stdout.columns) {
-		title = cliTruncate(title, process.stdout.columns - 3);
-	}
+	title = cliTruncate(title, process.stdout.columns - 3);
 
 	if (test.todo) {
 		this.todoCount++;
--- a/lib/reporters/mini.js
+++ b/lib/reporters/mini.js
@@ -6,6 +6,7 @@ var plur = require('plur');
 var spinners = require('cli-spinners');
 var chalk = require('chalk');
 var colors = require('../colors');
+var cliTruncate = require('cli-truncate');
 
 chalk.enabled = true;
 Object.keys(colors).forEach(function (key) {
@@ -76,6 +77,10 @@ MiniReporter.prototype.prefix = function (str) {
 
 MiniReporter.prototype._test = function (test) {
 	var title = test.title;
+
+	if (title.length > process.stdout.columns) {
+		title = cliTruncate(title, process.stdout.columns - 3);
+	}
 
 	if (test.todo) {
 		this.todoCount++;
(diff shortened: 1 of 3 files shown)


lib/reporters/mini.js
```diff
@@ -77,6 +78,10 @@ MiniReporter.prototype.prefix = function (str) {
 MiniReporter.prototype._test = function (test) {
 	var title = test.title;
 
+	if (title.length > process.stdout.columns) {
```
sindresorhus: I don't think this check is needed. `cliTruncate` handles that for you anyways.

sotojuan: I used to not have the check but then for some reason the mini tests were [failing](http://i.imgur.com/zHqqEW9.png). Didn't have time to diagnose why though.

sindresorhus:
There's something weird going on. Notice how one of the test titles are not visibly truncated.

untitled12

sotojuan:
That might be because the `if` checks for just "less than". That `function gets iter count` title may just be the same "size" as the CLI. If I change the check to be like "if title is greater than CLI minus X" it should be fine. I tried doing that but for some reason that specific test title never shows for me, but other looks fine.

Alternatively, I could just get rid of the `if` altogether, though I'll have to figure out how to make the failing tests pass (unless just adding the ellipses to them works).

lib/reporters/mini.js
jamestalmage: I am guessing `process.stdout.columns === undefined` during testing. Same reason as [here](https://github.com/sindresorhus/ava/issues/605#issue-138766792).

> This is because `tap` forks the process and does not emulate a tty environment in the child process.

You could force the issue in those tests simply by setting process.stdout.columns to some large number in the tests.

jamestalmage:
Also, I would prefer that the truncation character always be in the rightmost column and try to show as much of that last word which extends past the border as possible.

i.e. when truncating to `6` characters

```
foo bar
```

should become

```
foo b…
```

instead of

```
foo…
```

It looks like this would require a PR to `cli-truncate`, or just use `slice-ansi`.

sotojuan:
Forced `process.stdout.columns` in the mini reporter tests, so no need for the `if` statement anymore.

> Also, I would prefer that the truncation character always be in the rightmost column and try to show as 
much of that last word as possible instead of just dropping it.

I'll look into this—sounds like a good idea.

lib/reporters/mini.js
```diff
@@ -77,6 +78,8 @@ MiniReporter.prototype.prefix = function (str) {
 MiniReporter.prototype._test = function (test) {
 	var title = test.title;
 
+	title = cliTruncate(title, process.stdout.columns - 3);
```
novemberborn: Should truncate `test.title` and assign the result to `var title`, rather than reassigning `title`.

sotojuan:
As per @novemberborn's suggestion, truncated `test.title` directly instead of reassigning.

test/reporters/mini.js
```diff
@@ -8,6 +8,7 @@ var beautifyStack = require('../../lib/beautify-stack');
 chalk.enabled = true;
 
 var graySpinner = chalk.gray.dim('⠋');
+process.stdout.columns = 5000;
```
vdemedes: Could you please add a comment to this?

sotojuan:
Added comment explaining the use of `process.stdout.columns = 5000;`.

sindresorhus:
Thank you @sotojuan. Good work :)

sindresorhus:
> It looks like this would require a PR to cli-truncate, or just use slice-ansi.

Switched `cli-truncate` to use `slice-ansi`.

PR #640, opened 2016-03-12
Switch from deeper to only-shallow so t.same doesn't check constructor equality

### LINKED ISSUE #630: Ignore constructor equality for t.same
## Description
t.same now uses deeper, which checks the equality of the constructors of objects. I argue that a deep equal should ignore the constructor. Here's why:
Two objects created with different instances of the constructor will fail. 
Constructor is not a key in Object.keys, not part of JSON, and not returned by tools like graphQL. That means one must recreate the result in order to append the constructor without adding it as a key (unless I'm doing it wrong).
Insanely difficult to debug and makes you pull your hair out because logging both objects returns two identical objects and you start to question your sanity when the diff returns nothing. 

## Environment

Affects all, but running on 5.7.1, OSX (Darwin 15.x)

mattkrick:
Fixes #630


vdemedes:
Thanks @mattkrick!

PR #647, opened 2016-03-15
Show error if babel config value is invalid

### LINKED ISSUE #623: Specifying a bad string for the babel config option gives unhelpful error
Using `master` and the following config in `package.json`:

```js
"ava": {
  "babel": "foo"
}
```

Produces the following error:

```
ReferenceError: [BABEL] /dev/project/test.js: Unknown option: base.0. Check out http://babeljs.io/docs/usage/options/ for more info
```

If a uses provides a string shortcut for the "babel" option, we should verify it is one of the allowed ones (currently that list is `default` and `inherit` if I'm not mistaken).

vdemedes:
This PR fixes #623.

If `babel` config value is invalid, AVA now shows an error:

screen shot 2016-03-15 at 8 57 27 pm

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -23,6 +23,7 @@ var arrify = require('arrify');
 var meow = require('meow');
 var Promise = require('bluebird');
 var pkgConf = require('pkg-conf');
+var chalk = require('chalk');
 var isCi = require('is-ci');
 var colors = require('./lib/colors');
 var verboseReporter = require('./lib/reporters/verbose');
@@ -40,6 +41,18 @@ var conf = pkgConf.sync('ava', {
 		babel: 'default'
 	}
 });
+
+// check for valid babel config shortcuts (can be either "default" or "inherit")
+var isValidShortcut = ['default', 'inherit'].indexOf(conf.babel) !== -1;
+
+if (!conf.babel || (typeof conf.babel === 'string' && !isValidShortcut)) {
+	var message = '';
+	message += 'Unexpected Babel configuration for AVA. ';
+	message += 'See ' + chalk.underline('https://github.com/sindresorhus/ava#es2015-support') + ' for allowed values.';
+
+	console.log('\n  ' + colors.error(figures.cross) + ' ' + message);
+	process.exit(1);
+}
 
 var cli = meow([
 	'Usage',
--- a/test/cli.js
+++ b/test/cli.js
@@ -4,9 +4,15 @@ var childProcess = require('child_process');
 var test = require('tap').test;
 global.Promise = require('bluebird');
 var getStream = require('get-stream');
+var figures = require('figures');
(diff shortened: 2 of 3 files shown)


vdemedes:
I'm sure there might be a better error message, so feedback is welcome!

cli.js
```diff
@@ -41,6 +42,18 @@ var conf = pkgConf.sync('ava', {
 	}
 });
 
+// check for valid babel config shortcuts (can be either "default" or "inherit")
+var isValidShortcut = ['default', 'inherit'].indexOf(conf.babel) >= 0;
+
+if (conf.babel.length > 0 && !isValidShortcut) {
```
novemberborn: Presumably this will crash if `babel` is set to `null`. I'd prefer this to be more explicit:

```js
if (!conf.babel || (typeof conf.babel !== 'object' && conf.babel !== 'default' && conf.babel !== 'inherit')) {
```

Note that `conf.babel` should default to `default` so `!conf.babel` would only occur if it's explicitly set to a falsey value.

cli.js
```diff
@@ -41,6 +42,18 @@ var conf = pkgConf.sync('ava', {
...
+	var message = '';
+	message += 'Value "' + conf.babel + '" in "babel" property of AVA config in package.json is invalid.';
+	message += ' Check out ' + chalk.underline('https://github.com/sindresorhus/ava#es2015-support') + ' for docs on how to configure Babel.';
```
novemberborn: How about:

> Unexpected Babel configuration for AVA. See https://github.com/sindresorhus/ava#es2015-support for allowed values.

Even if users don't recall where they configured Babel, the GitHub link will make that clear. I do think we should mention the "for AVA" part to make clear we're not talking about `.babelrc` stuff.

sindresorhus: :+1:

cli.js
```diff
@@ -41,6 +42,18 @@ var conf = pkgConf.sync('ava', {
 	}
 });
 
+// check for valid babel config shortcuts (can be either "default" or "inherit")
+var isValidShortcut = ['default', 'inherit'].indexOf(conf.babel) >= 0;
```
sindresorhus: `>= 0` → `!== -1`

cli.js
vdemedes: That way it will fail when `babel` config is not even present. It should only check and display this error message, when `babel` config is there and has invalid value.

vdemedes:
PR updated!

cli.js
novemberborn: > That way it will fail when babel config is not even present.

It won't, it has a default value of `"default"`.

vdemedes: > It won't, it has a default value of "default".

Oh right! Updating PR to handle this case.

PR #651, opened 2016-03-17
Ensure t.throws() returns the error

novemberborn:
PR #576 changed the `t.throws()` assertion to return the thrown error, or if asynchronous a promise for the rejection reason. Unfortunately this only worked for asynchronous errors.

The tests cover the changes in `lib/assert.js` but `t.throws()` is an *enhanced* assertion. This commit ensures any values returned from `lib/assert.js` assertions are indeed returned by the corresponding `t.` assertions.

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -277,31 +277,31 @@ function PublicApi(test) {
 }
 
 function onAssertionEvent(event) {
-	var promise = null;
-
 	if (event.assertionThrew) {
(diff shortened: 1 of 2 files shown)


jamestalmage:
LGTM

sotojuan:
Looking good here too!

sindresorhus:
👌😎🎉

Issue #171, opened 2015-11-09
Add idle timeout support


MarkTiedemann:
Just switched to AVA -- and I love it! -- but, -- and this is really a minor issue -- I also noticed that I kept reusing code like this:

```javascript
import test from 'ava';

function failAfter(time, t){
  setTimeout(() => {
    t.fail('timed out after ' + time + ' ms');
    t.end(); /* fail fast! */
  }, time);
}

async function thisShouldntTakeLong(){ /* or does it? >_< */
  return new Promise(resolve => setTimeout(resolve, 9999));
}

test('this shouldnt take long', async t => {
  failAfter(1000, t);
  await thisShouldntTakeLong();
});
```

And just to be clear, I don't like global test timeouts -- I think they're a terrible idea. 

Yet again, I think something like `t.failAfter(time)` is not too bad.

What do you guys think? :)

jamestalmage:
>  I don't like global test timeouts -- I think they're a terrible idea.

Why?

MarkTiedemann:
@jamestalmage Well, I'm just kinda annoyed of having to change the limit, for example, in Mocha again and again.

jamestalmage:
Right now AVA hangs forever if you fail to call `t.end()`. That's no good. Especially if one of those tests make it to your CI server.

One thing I am always writing (in mocha) is:

```js
if (process.env.TRAVIS) {
   this.timeout(fiveTimesWhateverItTakesLocally) 
}
```

I dislike having to wait to see my mistake during development, just to accommodate the slower CPU on my CI. It would be cool to build that in to AVA (though maybe not possible to do in a reliable way across all CI implementations out there).

Either way, I think AVA needs global and per tests timeouts with sensible defaults.

sindresorhus:
> I think AVA needs global and per tests timeouts with sensible defaults.

What are sensible defaults in your opinion?

MarkTiedemann:
@jamestalmage You changed my mind. :) I agree, in CI defaults do make sense.

jamestalmage:
Ooh. That is a hard question. :smile: 

Locally, I would prefer something really fast. `500 ms` or something. Some users might find that surprising (though it might also help to gently remind people that they should be writing small, fast tests). It may also unfairly inconvenience users with really old/slow CPU's.

alexbooker:
Hi,

I gathered some data on the default timeouts used by other popular tools. Hopefully it will aid you in reasoning about a "sensible default":

[...]

Issue #583, opened 2016-03-01
Print pending tests when interrupting AVA, or after a timeout


novemberborn:
If you interrupt AVA (Ctrl+C/SIGTERM) while tests are still pending, AVA should print those tests before exiting. Similarly it should print pending tests when `--timeout` is reached.

This should help debug which tests are taking too long.

Related to #171.

vdemedes:
Great idea!

sindresorhus:
:+1:

radiovisual:
This would be awesome. 👍

sindresorhus:
Would be really useful to use [`why-is-node-running`](https://github.com/mafintosh/why-is-node-running) to also show **why** tests are still pending.

QuotableWater7:
Just curious, has there been any movement on this issue?

cc: @novemberborn

novemberborn:
@QuotableWater7 nope. Are you interested in taking this on?

austinkelleher:
This would be tremendously useful. Is this still something that is planned @novemberborn?

novemberborn:
@austinkelleher most definitely. Would you want to give it a go?

ace-n:
Ran into this problem myself - gonna try hacking something together if y'all don't mind. 🙂

novemberborn:
This will be easier to implement now that #1722 has landed, though perhaps we should implement it in the reporters, in which case it'd be better to wait for https://github.com/avajs/ava/pull/1776.

vancouverwill:
any update on this now https://github.com/avajs/ava/pull/1722 is merged ?

novemberborn:
@vancouverwill I haven't had much time to work on #1776. I do think it should be implemented in the reporters, but I'm happy with some duplication if that unblocks.

Would you be interested in working on this?

vancouverwill:
@novemberborn hey yeah I would be interested, could you possibly give me some guidelines in terms of code style you would Iike or any goals with it? Thanks

novemberborn:
@vancouverwill we use [XO](https://www.npmjs.com/package/xo) for code style.

The reporters should receive events when tests are declared and when they finish, so it'll be possible to keep a list of pending tests (and test files, too). Then in `lib/cli` there should be a hook for SIGINT & SIGTERM to make the reporters stop listening to incoming events and shut down the workers, and then print the pending tests & test files. The same for `--timeout` handling.

At least that's the theory, it may be a bit different in practice 😄

[...]

PR #654, opened 2016-03-17
Add timeout functionality

### LINKED ISSUE #8: Timeout
Should we have some sort of timeout on the tests?

vdemedes:
This PR fixes #8 and and #171.

Timeout can be set via `--timeout` or `-T` option. Values can be either numbers or set in a human-readable form:

```
$ ava --timeout=100 # 100 milliseconds
$ ava --timeout=2s # 2 seconds
$ ava --timeout=3m # 3 minutes
```

Timer resets after each completed test. So if there was 2 seconds (for example) of inactivity (no new test results), AVA exits with the following error message:

screen shot 2016-03-17 at 11 15 35 pm

P.S. `test/fixture/long-running.js` was renamed to `test/fixture/slow-exit.js`, because the new name is more appropriate for the represented test case.


readme.md
```diff
@@ -679,6 +680,20 @@ AVA automatically removes unrelated lines in stack traces, allowing you to find
 
 <img src="media/stack-traces.png" width="300">
 
+### Global timeout
+
+A global timeout can be set via `--timeout` option.
```
sindresorhus: > A global timeout can be set via the `--timeout` option.

readme.md
```diff
@@ -679,6 +680,20 @@ AVA automatically removes unrelated lines in stack traces, allowing you to find
...
+Timeout in AVA behaves differently than in other test frameworks.
+AVA resets a timer after each test, forcing tests to quit if no new test results received within specified "timeout time".
```
sindresorhus: > AVA resets a timer after each test, forcing tests to quit if no new test results were received within the specified timeout.

readme.md
```diff
@@ -679,6 +680,20 @@ AVA automatically removes unrelated lines in stack traces, allowing you to find
...
+
+You can set timeouts in a human-readable way:
```
sindresorhus: Maybe also show that you still can set it in a millisecond way.

readme.md
```diff
@@ -679,6 +680,20 @@ AVA automatically removes unrelated lines in stack traces, allowing you to find
...
+
+```
+$ ava --timeout 10s # 10 seconds
+$ ava --timeout 2m # 2 minutes
+$ ava --timeout 100 # 100 milliseconds
```
sindresorhus: Use `=` not ` ` => `$ ava --timeout=100 # 100 milliseconds`

sindresorhus:
Shouldn't this be implemented in the API? `timeout` can be useful for other consumers too, like the grunt/gulp plugins.

cli.js
```diff
@@ -153,6 +157,27 @@ if (files.length === 0) {
 	];
 }
 
+var timeout;
+var timer;
+
+function restartTimer() {
+	clearTimeout(timer);
+
+	timer = setTimeout(onTimeout, timeout);
```
sindresorhus: Use https://github.com/sindresorhus/ava/blob/8d47119458e83d3899683ad3ea3a4c1c01b7dd49/lib/globals.js#L9 to prevent it being mocked with when people use Sinon/etc. Same with `clearTimeout`.

vdemedes:
> Shouldn't this be implemented in the API? timeout can be useful for other consumers too, like the grunt/gulp plugins.

Hm, probably, I'll push an update.

cli.js
```diff
@@ -153,6 +157,27 @@ if (files.length === 0) {
...
+}
+
+function onTimeout() {
+	logger.finish();
+	console.log('  ' + colors.error(figures.cross) + ' Exited because no new tests completed within last ' + cli.flags.timeout + ' of inactivity.');
```
novemberborn: Can remove the `of inactivity` bit I think. Should add `ms`.

novemberborn:
> Shouldn't this be implemented in the API? timeout can be useful for other consumers too, like the grunt/gulp plugins.

The watcher should follow this behavior as well.

---

Clearing and restarting timers can be expensive. Would be good to investigate the performance overhead. One simple solution may be to never clear the timer but track when the last test completed. Then the next timeout can be scheduled so it fires `timeout` ms since completion.

Timers may sometimes fire a little early. Presumably this is only an issue when a very short timeout value is used though.

cli.js
vdemedes: `ms` can't be added, because https://github.com/sindresorhus/ava/pull/654/files#diff-0730bb7c2e8f9ea2438b52e419dd86c9R689

vdemedes:
> Clearing and restarting timers can be expensive.

Could you provide any links backing this info?

cli.js
novemberborn: Ah didn't read that closely enough :smile:

novemberborn:
> Could you provide any links backing this info?

There's the infamous ["It sounds crazy, but disabling npm's progress bar yields a 2x npm install speed improvement for me"](https://twitter.com/gavinjoyce/status/691773956144119808) issue from a little while ago. See the [3.7.0](https://github.com/npm/npm/releases/tag/v3.7.0) release notes and <https://github.com/iarna/gauge/commit/a7ab9c906bb72aa0ed8996a00db2cd35a22d5992>.

Hard to say what the actual impact is on us of course.

api.js
```diff
@@ -167,6 +172,21 @@ Api.prototype._prefixTitle = function (file) {
...
 };
 
+Api.prototype._restartTimer = function () {
+	clearTimeout(this._timer);
+
+	this._timer = setTimeout(this._onTimeout, this.options.timeout);
+};
+
+Api.prototype._onTimeout = function () {
+	var message = 'Exited because no new tests completed within last ' + this.options.timeout + 'ms of inactivity';
```
sindresorhus: > completed within **the** last

sindresorhus:
@vdemedes Can you fix the merge conflict?

sindresorhus:
Also discovered a problem with the timeout handling and its test being faulty:

```js
import test from 'ava';
test.cb(t => {
	setTimeout(t.end, 10000);
});
```

```
$ ava --timeout=1s
```

This will output the error message after 10 seconds because of the `setTimeout`, not after 1s of inactivity. If you remove the `setTimeout` it will never exit.

vdemedes:
@sindresorhus Sure, will rebase tomorrow.

Will try to fix it, good catch ;)

novemberborn:
> Hard to say what the actual impact is on us of course.

Does the `bench` stuff enable us to test this? (I've never used it before)

jamestalmage:
> Does the bench stuff enable us to test this? (I've never used it before)

It should. Docs [here](https://github.com/sindresorhus/ava/blob/master/maintaining.md#benchmarking). I would love a second set of eyeballs on our benchmark tooling! :stuck_out_tongue_winking_eye:

novemberborn:
I ran the benchmarks with 8d47119458e83d3899683ad3ea3a4c1c01b7dd49 and <https://github.com/sindresorhus/ava/tree/timeout-no-clear>. For the timeout tests I added `this.options.timeout = '10s';` in `api.js` just before parsing. The results are below.

This PR seems to be the slowest overall. The `no-clear` implementation is pretty naive (it doesn't reschedule with a decreased interval) but seems to be the way to go.

```
┌────────────────────────────────────────────────┬────────────────────────────┬────────────────────────────┬────────────────────────────┬─────────────────────────────┐
│                                                │            mean            │           median           │            min             │             max             │
├────────────────────────────────────────────────┼──────────┬─────────┬───────┼──────────┬─────────┬───────┼──────────┬─────────┬───────┼──────────┬─────────┬────────┤
│ args                                           │ no-clear │ timeout │ pre   │ no-clear │ timeout │ pre   │ no-clear │ timeout │ pre   │ no-clear │ timeout │ pre    │
├────────────────────────────────────────────────┼──────────┼─────────┼───────┼──────────┼─────────┼───────┼──────────┼─────────┼───────┼──────────┼─────────┼────────┤
│ --verbose other/failures.js                    │ 1.838    │ 1.992   │ 1.813 │ 1.835    │ 1.988   │ 1.804 │ 1.826    │ 1.838   │ 1.79  │ 1.856    │ 2.195   │ 1.842  │
├────────────────────────────────────────────────┼──────────┼─────────┼───────┼──────────┼─────────┼───────┼──────────┼─────────┼───────┼──────────┼─────────┼────────┤
│ --verbose serial/alternating-sync-async.js     │ 3.227    │ 3.568   │ 3.264 │ 3.203    │ 3.569   │ 3.153 │ 3.176    │ 3.2     │ 3.123 │ 3.366    │ 3.703   │ 3.679  │
├────────────────────────────────────────────────┼──────────┼─────────┼───────┼──────────┼─────────┼───────┼──────────┼─────────┼───────┼──────────┼─────────┼────────┤
│ --verbose serial/async-immediate.js            │ 4.701    │ 5.134   │ 4.683 │ 4.661    │ 5.097   │ 4.615 │ 4.624    │ 4.691   │ 4.527 │ 4.847    │ 5.652   │ 5.028  │
├────────────────────────────────────────────────┼──────────┼─────────┼───────┼──────────┼─────────┼───────┼──────────┼─────────┼───────┼──────────┼─────────┼────────┤
│ --verbose serial/async-timeout.js              │ 5.983    │ 5.899   │ 5.948 │ 5.994    │ 6.073   │ 5.969 │ 5.888    │ 5.126   │ 5.868 │ 6.009    │ 6.33    │ 5.991  │
├────────────────────────────────────────────────┼──────────┼─────────┼───────┼──────────┼─────────┼───────┼──────────┼─────────┼───────┼──────────┼─────────┼────────┤
│ --verbose serial/sync.js                       │ 1.45     │ 1.737   │ 1.419 │ 1.437    │ 1.684   │ 1.418 │ 1.415    │ 1.59    │ 1.395 │ 1.537    │ 2.034   │ 1.443  │
├────────────────────────────────────────────────┼──────────┼─────────┼───────┼──────────┼─────────┼───────┼──────────┼─────────┼───────┼──────────┼─────────┼────────┤
│ --verbose concurrent/alternating-sync-async.js │ 2.057    │ 2.229   │ 1.945 │ 1.983    │ 2.154   │ 1.94  │ 1.973    │ 2.052   │ 1.892 │ 2.294    │ 2.468   │ 2.07   │
├────────────────────────────────────────────────┼──────────┼─────────┼───────┼──────────┼─────────┼───────┼──────────┼─────────┼───────┼──────────┼─────────┼────────┤
│ --verbose concurrent/async-immediate.js        │ 2.702    │ 3.024   │ 2.782 │ 2.701    │ 3.04    │ 2.687 │ 2.662    │ 2.831   │ 2.644 │ 2.734    │ 3.195   │ 3.341  │
├────────────────────────────────────────────────┼──────────┼─────────┼───────┼──────────┼─────────┼───────┼──────────┼─────────┼───────┼──────────┼─────────┼────────┤
│ --verbose concurrent/async-timeout.js          │ 2.694    │ 3.108   │ 2.76  │ 2.694    │ 3.082   │ 2.688 │ 2.671    │ 2.809   │ 2.64  │ 2.716    │ 3.563   │ 3.063  │
├────────────────────────────────────────────────┼──────────┼─────────┼───────┼──────────┼─────────┼───────┼──────────┼─────────┼───────┼──────────┼─────────┼────────┤
│ --verbose concurrent/sync.js                   │ 1.201    │ 1.39    │ 1.247 │ 1.201    │ 1.364   │ 1.215 │ 1.182    │ 1.335   │ 1.181 │ 1.215    │ 1.503   │ 1.408  │
├────────────────────────────────────────────────┼──────────┼─────────┼───────┼──────────┼─────────┼───────┼──────────┼─────────┼───────┼──────────┼─────────┼────────┤
│ --verbose concurrent/*.js serial/*.js          │ 9.77     │ 11.267  │ 9.824 │ 9.654    │ 11.255  │ 9.568 │ 9.579    │ 10.086  │ 9.413 │ 10.364   │ 12.382  │ 10.625 │
└────────────────────────────────────────────────┴──────────┴─────────┴───────┴──────────┴─────────┴───────┴──────────┴─────────┴───────┴──────────┴─────────┴────────┘
```

screen shot 2016-03-23 at 14 49 50

vdemedes:
@novemberborn Could you please explain the behavior and code behind that `timeout-no-clear` branch?

novemberborn:
@vdemedes timers are somewhat expensive to create. Rather than replacing an existing timer whenever a test completes I propose to let the existing timer run to completion. Then if tests have completed in the meantime we create a new timer. The delay for this new timer should be the configured timeout minus the time since the last test completed. This gets us the desired behavior without creating an unnecessary amount of timers.

This is what the `timer-no-clear` branch does, although it simply creates the new timer with the configured timeout. (The downside of that naive approach is that, worst case, it might actually take twice the timeout delay before the test run is aborted.)

vdemedes:
Anyone has ideas on how to fix this - https://github.com/sindresorhus/ava/pull/654#issuecomment-199748315? The only way I see is `process.exit()`.

> This will output the error message after 10 seconds because of the setTimeout, not after 1s of inactivity. If you remove the setTimeout it will never exit. Can you also add a test where it never exits by itself?

@sindresorhus expect that it won't exit and `.kill()` it to finish the test?

novemberborn:
@vdemedes yea, I think we'd have to explicitly kill the forks and flag any remaining tests as having timed out. Of course they could be serial async tests in which case they wouldn't even have started but that nuance is probably not relevant.

vdemedes:
@novemberborn in `timeout-no-clear`, won't timeout be different from what user specified with `--timeout`?

I pushed a new technique, ridiculously simple - debouncing. Instead of restarting timer after every test, this function is debounced with 200ms interval, so that it's not fired often. I believe it should improve things a lot.

novemberborn:
> It is simple and understandable, no black magic involved. Let me know what you think ;)

`debounce` still [creates a timer](https://github.com/component/debounce/blob/0b58e834ede3bead70c09af22375e1d62671d7b9/index.js#L30) though. This'll create fewer than your original approach but since we're effectively implementing our own kill-forks debouncer already why not go all the way?

api.js
```diff
@@ -180,6 +186,25 @@ Api.prototype._prefixTitle = function (file) {
 	return prefix;
 };
 
+Api.prototype._restartTimer = debounce(function () {
```
jamestalmage: There is no need for this function using `debounce`.

Just do the following in the constructor:

```js
if (this.options.timeout) {
  this._onTimeout = debounce(this._onTimeout, ms(this.options.timeout));
} else {
  this._onTimeout = noop;
}
```

I think that will be easier to understand (and faster) than the current implementation or the `no-clear` branch.

novemberborn: Ha of course. It implements the approach I was advocating. Missed it due to context switching :blush:

jamestalmage: Also, you shouldn't use `debounce` on a prototype method, as it will debounce across every instance and only call on the most recent.

https://github.com/component/debounce/issues/8

jamestalmage:
Even with `debounce` there is an issue with this when someone has a long `timeout` set and the tests finish. There is no way to cancel the scheduled timer.

I've raised an issue: https://github.com/component/debounce/issues/9

api.js
vdemedes: The point of the latest update is to create less timers. That's why `debounce` should be on `_restartTimer()`.

> Also, you shouldn't use debounce on a prototype method, as it will debounce across every instance and only call on the most recent.

Agree.

vdemedes:
> Even with debounce there is an issue with this when someone has a long timeout set and the tests finish. There is no way to cancel the scheduled timer.

This is not a problem, because we explicitly call `process.exit()` when the tests finish.

novemberborn:
@vdemedes I think you can use `debounce` all the way. Start it with the configured timeout and every time a test completes you call the debouncer again.

>> Even with debounce there is an issue with this when someone has a long timeout set and the tests finish. There is no way to cancel the scheduled timer.
> This is not a problem, because we explicitly call process.exit() when the tests finish.

It will be a problem with the watcher. If we can't get cancel support upstream we'd need to track some state in the debouncer callback so it doesn't abort tests from newer test runs.

vdemedes:
> It will be a problem with the watcher. If we can't get cancel support upstream we'd need to track some state in the debouncer callback so it doesn't abort tests from newer test runs.

In that case we have 3 choices:

1. Disable timeouts and merge
2. Wait for `debounce` module to support cancelling
3. Find alternative module that supports cancelling already

I'm for 3.

vdemedes:
`debounce` in lodash has a `cancel()` method, which is perfect for us - https://lodash.com/docs#debounce.

api.js
```diff
@@ -180,6 +188,25 @@ Api.prototype._prefixTitle = function (file) {
 	return prefix;
 };
 
+Api.prototype._restartTimer = function () {
+	if (this._timer) {
+		clearTimeout(this._timer);
+	}
+
+	this._timer = setTimeout(this._onTimeout, this.options.timeout);
```
jamestalmage: There is no need for `_restartTimer` or `_timer`, just use `lodash.debounce` for `onTimeout`.

api.js
novemberborn: @vdemedes bump :smile:

package.json
```diff
@@ -96,6 +96,7 @@
     "common-path-prefix": "^1.0.0",
     "convert-source-map": "^1.1.2",
     "core-assert": "^0.1.0",
+    "debounce": "^1.0.0",
```
novemberborn: This is now unused.

vdemedes:
PR updated.

I re-implemented stats calculation in Runner, because it returned zero `passCount`, `failCount`, etc when timeout was reached and forks were killed. Now Runner returns current stats (including completed tests):

screen shot 2016-04-04 at 1 55 32 pm

api.js
```diff
@@ -13,8 +13,10 @@ var commonPathPrefix = require('common-path-prefix');
 var resolveCwd = require('resolve-cwd');
 var uniqueTempDir = require('unique-temp-dir');
 var findCacheDir = require('find-cache-dir');
+var debounce = require('lodash.debounce');
```
BarryThePenguin: How does that compare to installing `lodash` and using the bellow?

```js
var debounce = require('lodash/fp/debounce');
```

vdemedes: `lodash.debounce` contains only `debounce()` function, while `lodash` module contains all of them.

sindresorhus:
LGTM

sindresorhus:
@vdemedes https://github.com/sindresorhus/ava/pull/654#discussion_r56964317

vdemedes:
Updated ;)

sindresorhus:
@novemberborn @jamestalmage :shipit:❓

jamestalmage:
Yes - ship

jamestalmage:
Why is https://github.com/sindresorhus/ava/commit/ae45752597130fb86fa678b30a24613a619540d3 in this PR?

sindresorhus:
@jamestalmage Explained in https://github.com/sindresorhus/ava/pull/654#issuecomment-205262984

jamestalmage:
:shipit:

sindresorhus:
Awesome‼️ Primo work @vdemedes

leo-happy

novemberborn:
Whoah that's a young DiCaprio :astonished: 

---

I'm confused by <https://github.com/sindresorhus/ava/pull/654/commits/ae45752597130fb86fa678b30a24613a619540d3>. These counts are [already computed in the `_addTestResult()` method](https://github.com/sindresorhus/ava/blob/e9c6cc2843e9ee6127aebb2843eafec86cb39bfc/lib/runner.js#L93:L105) in that same file. Was the change needed to get all results in case of a timeout? Then `_addTestResult()` needs to be cleaned up I think. Might not even need to expose the stats object.

vdemedes:
> Was the change needed to get all results in case of a timeout?

Yes, exactly! When test worker needed to exit before tests completed, all stats were zero. So I implemented "on-demand" calculation of stats.

@sindresorhus Thanks :)

giphy 15

novemberborn:
> Yes, exactly! When test worker needed to exit before tests completed, all stats were zero. So I implemented "on-demand" calculation of stats.

Can we remove the "on-test-complete" stats calculation then?

vdemedes:
@novemberborn Yes, definitely, I guess I forgot to remove it myself.

novemberborn:
@vdemedes see #728 :smile:

satya164:
Is there a way to see which tests exactly timed out?

novemberborn:
@satya164 not yet, #583 will help with that though.

ash0080:
Don't think --timeout should be a default feature

PR #655, opened 2016-03-18
hotfix: inline source maps

jamestalmage:
Wait to merge - this might not be the best way to fix the issue.

This fixes an `nyc` compatibility issue that [surfaced here](https://github.com/kentcdodds/react-ava-workshop/pull/5)

I am not entirely sure why source-map inlining is required (it worked fine without it in previous versions of AVA).

@novemberborn - It looks like the type of value we return from `retrieveSourceMap` inside `test-worker.js` changed from an object to a string sometime between `0.9.2` and `0.13` - I wonder if that has something to do with it.

### DIFF
--- a/lib/caching-precompiler.js
+++ b/lib/caching-precompiler.js
@@ -44,7 +44,7 @@ CachingPrecompiler.prototype._factory = function (babelConfig, cacheDir) {
 		objectAssign(options, {
 			inputSourceMap: sourceMap && sourceMap.toObject(),
 			filename: filename,
-			sourceMaps: true,
+			sourceMaps: 'both',
 			ast: false
 		});
 


novemberborn:
> It looks like the type of value we return from retrieveSourceMap inside test-worker.js changed from an object to a string sometime between 0.9.2 and 0.13 - I wonder if that has something to do with it.

Yea I think so. We should try to `JSON.parse()` the file.

> I am not entirely sure why source-map inlining is required (it worked fine without it in previous versions of AVA).

IIRC some of the source map tools support map strings and objects, but others are more picky, or something like that. It might be that the `retrieveSourceMap()` result is being disregarded and the fallback is to look at an inline source map (or perhaps `retrieveSourceMap()` is the fallback for when there is no inline source map).

novemberborn:
Oh wait got confused between test files, source files, and the source maps used by nyc.

The source map handling in AVA is so we can correct stack traces for errors coming out of the test files. Presumably this works fine even if `retrieveSourceMap()` returns a `map` string (due to the aforementioned liberal acceptance of JSON strings and objects).

I suspect nyc is trying to generate coverage for the test files but can't find the source maps. Setting it to `inline` here might be rectifying that. I'll have to do some more digging to figure out how and why it's instrumenting the test files though, that shouldn't be necessary. I'm not clear at all on how it can even access the transpiled code.

Changing the [`mapFile` extension to `.js.map`](https://github.com/sindresorhus/ava/blob/8d47119458e83d3899683ad3ea3a4c1c01b7dd49/lib/caching-precompiler.js#L61) may also help.

jamestalmage:
> Changing the mapFile extension to .js.map may also help.

It might if we weren't lying to the module system about the sources location (it's coming from the cache folder, not the modules source path). This is why we have to use a custom function for sourcemap support.

jamestalmage:
Closing in favor of #662

PR #657, opened 2016-03-18
update typescript recipe

paradox41:
This updates the recipe to use the proper 'target'. es2015 was incorrect according
to the documentation: https://github.com/Microsoft/TypeScript/wiki/Compiler-Options

### DIFF
--- a/docs/recipes/typescript.md
+++ b/docs/recipes/typescript.md
@@ -18,7 +18,7 @@ Create a [`tsconfig.json`](https://github.com/Microsoft/TypeScript/wiki/tsconfig
 {
 	"compilerOptions": {
 		"module": "commonjs",
-		"target": "es2015"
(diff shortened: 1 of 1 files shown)


SamVerschueren:
I think that doc is outdated because they changed that recently. It has to be `es2015`. Will link the docs later, on my phone now an hard to find the release notes.

SamVerschueren:
Can't seem to find the official announcement anymore. I could find this issue however

https://github.com/Microsoft/TypeScript/issues/6734

And when running `tsc -h`, this is the target description

```
Specify ECMAScript target version: 'ES3' (default), 'ES5', or 'ES2015' (experimental)
```

paradox41:
So the PR should go against TypeScript it seems. Thanks for the
clarification.

PR #658, opened 2016-03-18
Rerun tests using "r" key in watch mode

### LINKED ISSUE #615: Rerun all tests in watch mode by simply tying 'r' (no enter)
Currently, to rerun all tests in watch mode, you need to type <kbd>r</kbd> + <kbd>Enter</kbd>. It'd be even quicker if you could type <kbd>r</kbd>.

See <https://nodejs.org/api/tty.html#tty_rs_setrawmode_mode>. This seems to intercept `Ctrl+C` and otherwise mess with the shell, however. But maybe I'm doing something wrong.

Code: https://github.com/sindresorhus/ava/blob/1868204c1901f45b4f66a520ef6486fdd71fe1d2/lib/watcher.js#L116

vdemedes:
This PR fixes #615.

### DIFF
--- a/lib/watcher.js
+++ b/lib/watcher.js
@@ -8,6 +8,7 @@ var union = require('array-union');
 var uniq = require('array-uniq');
 var defaultIgnore = require('ignore-by-default').directories();
 var multimatch = require('multimatch');
+var keypress = require('keypress');
 var nodePath = require('path');
 var slash = require('slash');
 
@@ -149,11 +150,19 @@ Watcher.prototype.cleanUnlinkedTests = function (unlinkedTests) {
 Watcher.prototype.observeStdin = function (stdin) {
 	var self = this;
 
+	if (stdin.isTTY) {
+		keypress(stdin);
+		stdin.setRawMode(true);
+	}
+
 	stdin.resume();
 	stdin.setEncoding('utf8');
-	stdin.on('data', function (data) {
-		data = data.trim().toLowerCase();
-		if (data !== 'r' && data !== 'rs') {
+	stdin.on('keypress', function (char, key) {
+		if (key.ctrl && key.name === 'c') {
+			process.exit(2);
+		}
+
+		if (char !== 'r') {
 			return;
 		}
 
--- a/docs/recipes/watch-mode.md
+++ b/docs/recipes/watch-mode.md
@@ -71,7 +71,7 @@ Dependency tracking works for required modules. Custom extensions and transpiler
 
 ## Manually rerunning all tests
 
-You can quickly rerun all tests by typing <kbd>r</kbd> on the console, followed by <kbd>Enter</kbd>.
+You can quickly rerun all tests by typing <kbd>r</kbd> on the console.
 
 ## Debugging
 
--- a/package.json
+++ b/package.json
@@ -105,6 +105,7 @@
     "is-generator-fn": "^1.0.0",
     "is-observable": "^0.1.0",
     "is-promise": "^2.1.0",
+    "keypress": "^0.2.1",
     "last-line-stream": "^1.0.0",
     "loud-rejection": "^1.2.0",
     "matcher": "^0.1.1",
--- a/test/watcher.js
+++ b/test/watcher.js
@@ -113,6 +113,9 @@ group('chokidar is installed', function (beforeEach, test, group) {
 
 		stdin = new PassThrough();
 		stdin.pause();
+		stdin.on('data', function (data) {
+			stdin.emit('keypress', data, {});
+		});
 	});
 
 	var start = function (sources) {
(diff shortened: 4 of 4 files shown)


lib/watcher.js
```diff
@@ -149,11 +150,19 @@ Watcher.prototype.cleanUnlinkedTests = function (unlinkedTests) {
...
+	}
+
 	stdin.resume();
 	stdin.setEncoding('utf8');
-	stdin.on('data', function (data) {
-		data = data.trim().toLowerCase();
-		if (data !== 'r' && data !== 'rs') {
+	stdin.on('keypress', function (char, key) {
+		if (key.ctrl && key.name === 'c') {
+			process.exit();
```
sindresorhus:
```
process.exit(2);
```

?

jamestalmage: :+1: (why `2`?)

sindresorhus: > Control-C is fatal error signal 2, (130 = 128 + 2, see above)

http://tldp.org/LDP/abs/html/exitcodes.html

vdemedes: Fixed, thx!

novemberborn:
Sorry didn't have a chance earlier to play with this. I have some concerns around the keyboard interaction. For instance if you run the watcher through npm (say `npm run test:watch`), `ctrl+c` now causes npm to spit out an error log. Previously the `SIGINT` would be handled by `npm run`. One cannot use `Enter` to insert some linebreaks either, which can be useful to discern repeat output.

`ctrl+\` (`SIGQUIT`) causes a crash. `ctrl+z` is ignored.

I suppose we could switch to a `0` exit signal for the npm interaction, and even make `Enter` insert linebreaks. But there's a bunch of other shell behaviors that are broken by this. How much would we want to replicate them and is it worth the hassle?

jamestalmage:
> ctrl+\ (SIGQUIT) causes a crash. ctrl+z is ignored.

Do people actually use those?

> ctrl+c now causes npm to spit out an error log

Agreed. That is a problem. Is there a way to 1) detect if you have a parent process, and 2) send a SIGINT to it? I would rather not exit with `0` (though that may be preferable to `npm`s big old log when it sees an error.

*Aside:* It would be really nice if `npm` didn't print that huge mess of log statements when a script fails. So annoying to scroll up past that. I have a hard time believing all that "this isn't a problem with npm" language significantly reduces their support requests.

> But there's a bunch of other shell behaviors that are broken by this. How much would we want to replicate them and is it worth the hassle?

I guess that depends on what those behaviors are, and what problems we are creating for users. If it creates real problems, lets just revert to the `r+enter` behavior (one more keypress is not that big of deal). We should definitely make a decision on that before publishing this though.

---

One last thought. I believe the original `r+enter` behavior was modeled off existing libraries. I would assume they knew about `rawMode` as well. They may have evaluated this same behavior and rejected it for the reasons being discussed now. @novemberborn It might be worth doing an issue search on some of those repos to see if they discussed this.

sindresorhus:
> Do people actually use those?

Never used either personally.

>  It would be really nice if npm didn't print that huge mess of log statements when a script fails. So annoying to scroll up past that.

Couldn't agree more! The npm CLI generally has pretty awful UX.

Related issue: https://github.com/npm/npm/issues/8821 (Please comment there!)

> "this isn't a problem with npm" language significantly reduces their support requests.

The irony is that usually it *is* a problem with npm even when they output that. So the usual flow is a user opening an issue on one of my repos, I have to waste time triaging, and just send them back to the npm issue tracker...

novemberborn:
> Do people actually use those?

They can be useful when the app has trapped the `SIGINT` signal, but yea I don't really use them either. Just raising their existence and incompatibility with the restart mode in this PR.

> Agreed. That is a problem. Is there a way to 1) detect if you have a parent process, and 2) send a SIGINT to it? I would rather not exit with 0

Not sure. I don't think the exit code matters all that much when in watch mode though.

> I believe the original r+enter behavior was modeled off existing libraries. I would assume they knew about rawMode as well. They may have evaluated this same behavior and rejected it for the reasons being discussed now.

Yes, see <https://github.com/remy/nodemon/issues/162#issuecomment-17227347>. Also:

> to listen for a keypress requires changing process.stdin to use setRawMode(true) which lets me read a character at a time, but this then passes on the stdin reading method to the child process (i.e. your node script) which would cause unexpected behaviour.

Not sure how much this impacts us, it might be due to how the child processes are forked.

novemberborn:
Just learned about `SIGINFO` <https://twitter.com/robsmallshire/status/706763512903819265>, another signal broken by this PR.

(Not that it's very useful, it just outputs stuff like `load: 1.85  cmd: node 18425 waiting 0.44u 0.06s`)

vdemedes:
I guess we should rollback this PR and just rerun via <kbd>r</kbd> + <kbd>Enter</kbd>. Will be easier for everyone and additional "Enter" is not big a deal.

jamestalmage:
Argument for keeping this:

We can change this back at any point. It's not going to break builds. People are just going to have to retrain "r" vs "r + enter". For all the problems we've listed, I haven't seen one I think is a practical problem yet.

Argument for reverting:

It's the "safer" option. Guaranteed not to cause problems (however unlikely problems seem).

---

If anyone can describe a situation where this prevents you from using watch mode then let's revert. Heck, I would be convinced if just one person said "I frequently use `Ctrl + SomethingOtherThanC`".

novemberborn:
Even if we don't revert we need to fix the exit code upon `Ctrl + C`.

I'm for reverting though, seems like an uphill battle trying to solve the various edge cases when they surface.

jamestalmage:
> Even if we don't revert we need to fix the exit code upon Ctrl + C.

Oh yeah, forgot about that. I don't love the idea of exiting with `0` on Ctrl + C. It's another one of those "probably not a problem, but just feels wrong" issues.

I think I'm convinced. Let's revert. @sindresorhus @vdemedes ?

vdemedes:
I'm for reverting, one less thing to maintain. Didn't think such a simple PR would have such complications afterwards :)

sindresorhus:
Ok, let's revert to R+Enter. I don't think people will need it that often anyways. Our dependency tracking is pretty good.

PR #659, opened 2016-03-19
switch from serialize-error to clean-yaml-object

jamestalmage:
clean-yaml-object is the error serialization tool used by node-tap.

It has some nice benefits over serailze-error including better stringification of functions and buffers.

More importantly, the shared code will help keep our tap output consistent with that of node-tap.


api.js
```diff
@@ -81,10 +81,21 @@ Api.prototype._handleOutput = function (channel, data) {
 	this.emit(channel, data);
 };
 
+function normalizeError(err) {
+	if (!err || typeof err !== 'object') {
```
vdemedes: Can just be `typeof err !== 'object'`

api.js
```diff
@@ -81,10 +81,21 @@ Api.prototype._handleOutput = function (channel, data) {
...
+		err = {
+			message: err,
+			stack: err
+		};
+	}
+	return err;
```
vdemedes: Missing newline

test/fixture/throw-anonymous-function.js
```diff
@@ -2,6 +2,6 @@ import test from '../../';
 
 test('throw an uncaught exception', () => {
 	setImmediate(() => {
-		throw () => {};
+		throw function () {};
```
vdemedes: Why throwing a function here?

api.js
jamestalmage: That doesn't cover `null`.

test/fixture/throw-anonymous-function.js
jamestalmage: To verify behavior when a user does that. (Not a recommended practice)

api.js
jamestalmage: Huh?

api.js
sindresorhus: Would maybe be clearer to use something like: https://github.com/sindresorhus/is-obj

api.js
sindresorhus: Should have an empty newline above `return err;`

test/fixture/throw-anonymous-function.js
sindresorhus: Why change it from an arrow function to a function?

jamestalmage: Because the arrow function is transformed to an anonymous function by Babel. Changing it back to an arrow function would not change the assertion. I felt this way the assertion was a little easier to understand.

api.js
jamestalmage: Is there a linter rule we can use to reinforce that style?

jamestalmage: Not yet apparently: https://github.com/eslint/eslint/issues/3092

test/fixture/throw-anonymous-function.js
sindresorhus: :+1:

PR #662, opened 2016-03-21
Ensure source map can be resolved from cached code

novemberborn:
In <https://github.com/kentcdodds/react-ava-workshop/pull/5> nyc is instrumenting the test files. Due to how our caching precompiler works nyc doesn't realize it's instrumenting the transpiled code. It can't find a source map for the transpiled code causing it to crash when generating coverage reports.

Generating coverage reports for test files shouldn't be necessary (and nyc excludes them by default), but it's not something that should crash either.

This PR adds a source map file comment to the cached code. The path is relative to that of the original file, allowing nyc to resolve the source map.

It also changes the source map file extension to the more conventional `.js.map`.

I think this is a better fix than #655 since we don't have to include the source map in an inline comment. The solution might be too specific to nyc though I imagine other module-loading-intercepting-libraries would behave the same way.

### DIFF
--- a/lib/caching-precompiler.js
+++ b/lib/caching-precompiler.js
@@ -59,7 +59,7 @@ CachingPrecompiler.prototype._factory = function (babelConfig, cacheDir) {
 		code = code.toString();
 		var options = buildOptions(filename, code);
 		var result = babel.transform(code, options);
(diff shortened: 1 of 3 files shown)


jamestalmage:
LGTM

sindresorhus:
LGTM2

PR #664, opened 2016-03-21
add note about AVA mantra in contributing guidelines

kentcdodds:
See [this](https://github.com/sindresorhus/ava/pull/663) PR which inspired this addition to `contributing.md` :-)


jamestalmage:
:+1: I like this.

We have run into this a number of times (`.avarc` requests, `config/ava` vs `ava` in `package.json`, etc). It would be kinder to contributors to explain this up front.

novemberborn:
+1 for merging this as-is, we can do other clarifications as a follow-up.

novemberborn:
See also https://github.com/sindresorhus/ava/issues/175#issuecomment-199014036

PR #666, opened 2016-03-21
Fix Gitter badge on npm site.

jamestalmage:
[npmjs.com](https://www.npmjs.com/package/ava) does not correctly display the gitter badge. My guess is that it's due to the unencoded space character in the image URL (won't know for sure until we publish this change).


vdemedes:
Mongoose uses `%20` instead of `_` as separator, so this should work too and is prettier.

sindresorhus:
Can you open an issue on https://github.com/npm/marky-markdown?

PR #667, opened 2016-03-22
Print the planned and actual assertion count when incorrect.

### LINKED ISSUE #102: When assertion count is incorrect, output the expected and actual values.

jamestalmage:
Fixes #102

The assertion error already had `expected` and `actual` properties, but only the tap reporter displays those. Instead of modifying existing reporters to handle the special case of planned assertions, I took the easy way out and just dynamically generated the error message.

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -7,6 +7,7 @@ var co = require('co-with-promise');
 var observableToPromise = require('observable-to-promise');
 var isPromise = require('is-promise');
 var isObservable = require('is-observable');
+var plur = require('plur');
 var inspect = require('util').inspect;
 var assert = require('./assert');
 var enhanceAssert = require('./enhance-assert');
@@ -223,7 +224,7 @@ Test.prototype._checkPlanCount = function () {
 		this._setAssertError(new assert.AssertionError({
 			actual: this.assertions.length,
 			expected: this.planCount,
-			message: 'Assertion count does not match planned',
+			message: 'Planned for ' + this.planCount + plur(' assertion', this.planCount) + ', but got ' + this.assertions.length + '.',
 			operator: 'plan'
 		}));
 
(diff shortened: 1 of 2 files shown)


lib/test.js
```diff
@@ -11,6 +11,7 @@ var inspect = require('util').inspect;
 var assert = require('./assert');
 var enhanceAssert = require('./enhance-assert');
 var globals = require('./globals');
+var plur = require('plur');
```
sindresorhus: Should be above relative require calls. Wish we had a ESLint rule for this. (https://github.com/sindresorhus/module-requests/issues/62)

novemberborn:
:shipit:

PR #678, opened 2016-03-23
fix instructions on how to run the bench commands

novemberborn:
:rocket:

### DIFF
--- a/maintaining.md
+++ b/maintaining.md
@@ -68,13 +68,13 @@ You may also want to check out the Settings page in Dev Tools and enable one or 
 First collect benchmark data for a branch/commit:
 
 ```
-$ node bench/run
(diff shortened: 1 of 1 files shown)


maintaining.md
```diff
@@ -68,13 +68,13 @@ You may also want to check out the Settings page in Dev Tools and enable one or
 First collect benchmark data for a branch/commit:
 
 ```
-$ node bench/run
+$ node bench/run.js
```
vdemedes: It works without `.js` extension as well, why add it?

novemberborn: Huh, TBH I didn't try that because I just figured `node` wanted the exact file paths. Looks like it tries the various extensions instead. Neat!

Issue #604, opened 2016-03-05
"Empty" test files are ok on node@5 and not ok on node@4


MoOx:
I currently have a project with "empty" test files like this

```js
import test from "ava"

// files added to report accurate coverage
import "../index.js"

test("todo", () => {
  console.log("TODO: test src/PageContainer/index*")
})
```

_By empty I mean that there is not a single assertion._

This is a workaround for `nyc --all` (see https://github.com/bcoe/nyc/issues/181) to get a real coverage percentage with a project that is not fully tested (yet).

Checkout this build https://travis-ci.org/MoOx/statinamic/builds/113959817. It's working on node@5 and failing on node@4 because empty test files report " Test results were not received from ...".

I expect that AVA behave the same on node@5 and node@4. So fails on both, or pass on both.

Maybe I am doing something wrong?

MoOx:
Oh. Totally sorry, I posted this issue too quickly. It's not especially related to "empty" files.
Well it seems AVA have a more global issue with node@4 (at least on Travis, cause I cannot even reproduce my problem locally)...

Will try to investigate.

jamestalmage:
What version of npm are you using on node 4 locally?

Travis node 4 uses `npm@2`. If you've upgraded your local node 4 to use `npm@3` it could be nom that is responsible for the weirdness.

AVA should behave the same across environments, so I feel this is valid.

MoOx:
Latest default node@4 (`nvm install 4 && nvm use 4`).
I did something stupid: I just used nvm to run tests using node@4 and didn't clean node_modules. So npm@2 but with a node_modules from npm@3. Will try with a clean npm@2 install.

ben-eb:
I'm getting this error too but on 0.12.

https://github.com/postcss/postcss-selector-parser/issues/44

thangngoc89:
I edited those empty test file to use `test.todo` and AVA 0.13.0 , same old error 

https://travis-ci.org/MoOx/statinamic/jobs/115137586#L794-L800

MoOx:
Btw, if you look correctly, it's not empty files that report failure (eg:  src/configurator/__tests__/index.js do have tests).

novemberborn:
@MoOx @ben-eb @thangngoc89 could you try with `master`? I have a feeling <https://github.com/sindresorhus/ava/pull/642/commits/5435820752099fbd5eef8bfcc815e36b0042355a> may be a relevant fix.

thangngoc89:
[...]

PR #155, opened 2015-11-07
Check for SIGTERM exit code

floatdrop:
Related - https://github.com/nodejs/node-v0.x-archive/issues/8667

Closes #154.

PR #680, opened 2016-03-23
Reject test runs when forks exit with an unexpected signal

novemberborn:
As discovered in #604 child processes can be killed by outside parties. In this case there is no exit code, though a signal is present. With this PR test runs are rejected with an appropriate error message, rather than the misleading "no test results received" message.

I've also cleaned up now redundant Node 0.10 SIGTERM handling (#155) and added a test for the existing test run rejection behavior.

### DIFF
--- a/lib/fork.js
+++ b/lib/fork.js
@@ -79,7 +79,7 @@ module.exports = function (file, opts) {
 		});
 
 		ps.on('exit', function (code) {
-			if (code > 0 && code !== 143) {
+			if (code > 0) {
 				return reject(new AvaError(relFile + ' exited with a non-zero exit code: ' + code));
 			}
 
--- a/lib/fork.js
+++ b/lib/fork.js
@@ -78,9 +78,13 @@ module.exports = function (file, opts) {
 			send(ps, 'teardown');
 		});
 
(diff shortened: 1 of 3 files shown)


thangngoc89:
I tested this branch. Receiving some error : 

https://travis-ci.org/MoOx/statinamic/jobs/118010809

novemberborn:
@thangngoc89 it seems fine locally (checked out https://github.com/MoOx/statinamic/commit/f89c11636d8d423f442e5f693e6f2779e740c024). Maybe somehow strangely it did cache something? The dependency cache is based on the package version which hasn't changed in master.

novemberborn:
Oops hardcoded slashes in the assertions so the tests fail on Windows :(

thangngoc89:
>  Maybe somehow strangely it did cache something?

Yeah. I cleaned all cache. And tests are passing on node 5.

> Oops hardcoded slashes in the assertions so the tests fail on Windows :(

`path.join` can help

novemberborn:
CI passes now.

vdemedes:
LGTM ;)

sindresorhus:
LGTM

PR #683, opened 2016-03-25
show error message if literal is thrown - fixes #661

### LINKED ISSUE #661: t.throws shows `undefined undefined undefined` when a string is thrown
## Description

When using `t.throws()` with a promise that rejects with a string instead of an Error object, the assertion fails.

``` js
import test from 'ava';

async function fn(): Promise<any> {
    return Promise.reject('foo');
}

test(async t => {
    t.throws(fn(), 'foo');
});
```

screen shot 2016-03-21 at 11 53 43

When using `Promise.reject(new Error('foo'));`, it works as expected.

It's due to [these](https://github.com/sindresorhus/ava/blob/master/lib/assert.js#L88-L93) lines. I have no idea why it overrides the `err` reference with a function. I looked into `core-assert` and it doesn't seem like it works with passing in a function as the `expected` parameter. But might be missing something here.
## Environment

Node.js v4.2.1
darwin 15.3.0

SamVerschueren:
This PR fixes #661 which prints `Expected an object to be thrown.`. It's the same error message that ESLint shows when trying to throw a literal. Feel free to provide feedback of any kind.

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -94,13 +94,18 @@ x.throws = function (fn, err, msg) {
 
 		var result;
 
-		assert.throws(function () {
-			try {
-				fn();
-			} catch (error) {
-				result = error;
-				throw error;
+		try {
+			fn();
+		} catch (error) {
+			if (error && typeof error !== 'object') {
+				throw new TypeError('Expected an object to be thrown.');
 			}
+
+			result = error;
+		}
+
+		assert.throws(function () {
+			throw result;
 		}, err, msg);
 
 		return result;
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -98,7 +98,10 @@ x.throws = function (fn, err, msg) {
 			fn();
 		} catch (error) {
 			if (error && typeof error !== 'object') {
-				throw new TypeError('Expected an object to be thrown.');
+				var e = new TypeError('Error `' + error + '` is not of type `object` but of type `' + (typeof error) + '`.');
+				e.actual = typeof error;
+				e.expected = 'object';
+				throw e;
 			}
 
 			result = error;
--- a/test/test.js
+++ b/test/test.js
@@ -433,6 +433,40 @@ test('throws and notThrows work with promises', function (t) {
 	});
 });
 
+test('throws does not work with literals', function (t) {
+	var result = ava(function (a) {
+		a.plan(2);
+		a.throws(function () {
+			// eslint-disable-next-line
+			throw 'foo';
+		});
+
+		a.throws(function () {
+			// eslint-disable-next-line
+			throw 5;
+		});
+	}).run();
+
+	t.is(result.reason.message, 'Expected an object to be thrown.');
+	t.is(result.passed, false);
+	t.is(result.result.planCount, 2);
+	t.is(result.result.assertCount, 2);
+	t.end();
+});
+
+test('throws does not work with promises rejecting literals', function (t) {
+	ava(function (a) {
+		a.plan(1);
+		a.throws(Promise.reject('foo'));
+	}).run().then(function (result) {
+		t.is(result.reason.message, 'Expected an object to be thrown.');
+		t.is(result.passed, false);
+		t.is(result.result.planCount, 1);
+		t.is(result.result.assertCount, 1);
+		t.end();
+	});
+});
+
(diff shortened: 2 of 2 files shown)


sindresorhus:
Can you also add what was thrown to the message, the type and value. Might be helpful.

SamVerschueren:
Also added the types to `actual` and `expected`. Not sure if necessary as it isn't being printed in the console?

sindresorhus:
> Not sure if necessary as it isn't being printed in the console?

Not really necessary as they're only useful in an AssertionError.

sindresorhus:
LGTM

SamVerschueren:
> Not really necessary as they're only useful in an AssertionError.

But it's ok to keep it like this?

lib/assert.js
```diff
@@ -94,13 +94,21 @@ x.throws = function (fn, err, msg) {
...
-		assert.throws(function () {
-			try {
-				fn();
-			} catch (error) {
-				result = error;
-				throw error;
+		try {
+			fn();
+		} catch (error) {
+			if (error && typeof error !== 'object') {
```
SamVerschueren: Not sure if this check is good enough in every case though. Should I use something like [is-error](https://www.npmjs.com/package/is-error) instead? Because now someone can still `throw {foo: 'bar'}`.

jamestalmage: :+1: for is-error.

novemberborn:
Hi @SamVerschueren. I don't think this PR is the correct solution to #661.

First some background.

`t.throws()` essentially wraps [`assert.throws()`](https://nodejs.org/api/assert.html#assert_assert_throws_block_error_message). That method supports constructor/instance checks, regular expression matches, and validation functions. We support those checks as well as an error message comparison validation.

[These lines](https://github.com/sindresorhus/ava/blob/efa43690069b3e60526f94f0093082f0c031b79f/lib/assert.js#L88-L93) implement that error message comparison validation. `err` here is not the error thrown by `fn()`, it's how that error should be checked.

In this example the error message does not compare (`'foo'.message !== 'foo'):

```js
import test from 'ava'

async function fn() {
    return Promise.reject('foo');
}

test(async t => {
    t.throws(fn(), 'foo');
});
```

The reason you get `undefined undefined undefined` is that we're [assuming](https://github.com/sindresorhus/ava/blob/efa43690069b3e60526f94f0093082f0c031b79f/lib/assert.js#L108) a validation error like that causes an `AssertionError` to be thrown. However Node [throws the actual error](https://github.com/nodejs/node/blob/be68b68d4863f0d389cc46fdf6f1cbcd1b241d0a/lib/assert.js#L330) which is the `'foo'` string.

In other words we need to check the error coming out of `assert.throws()` and call `test()` with the right arguments.

SamVerschueren:
Hi @novemberborn. Thanks for looking into this. Can you explain why it's not a solution for #661? I read your post multiple times but can't really find an answer :). After looking into it again, I might have found a solution that's "cleaner" and might fit your approach more. I will make another PR with the new solution so we can compare and make sure I understand you correctly.

> [These lines](https://github.com/sindresorhus/ava/blob/efa43690069b3e60526f94f0093082f0c031b79f/lib/assert.js#L88-L93) implement that error message comparison validation. err here is not the error thrown by fn(), it's how that error should be checked.

Yes I know, I had it wrong in the original issue. I'm not touching those lines. I thought that was the issue but after diving deeper I noticed it's a custom compare function for `assert.throws`.

novemberborn:
> Can you explain why it's not a solution for #661?

`t.throws()` shouldn't reject literals. It should ensure *something* was thrown. If that something didn't match expectations it should correctly report that. Right now it's the reporting that's broken, not the validation.

jamestalmage:
> t.throws() shouldn't reject literals. It should ensure something was thrown.

I disagree. https://github.com/sindresorhus/ava/issues/468#issue-128347888

SamVerschueren:
I agree with @jamestalmage although you might argue if AVA should enforce this best practice...

jamestalmage:
> you might argue if AVA should enforce this best practice

I think it protects you from careless mistakes.

novemberborn:
@SamVerschueren hope it's OK if I close this for now. I reckon resolving #661 will require some more changes.

PR #684, opened 2016-03-25
(2) show error message if literal is thrown - fixes #661

### LINKED ISSUE #661: t.throws shows `undefined undefined undefined` when a string is thrown

SamVerschueren:
This is another approach (see #683) in order to resolve #661. The error messages are a little bit different now as well

> Error `foo` is not of type `[object Error]` but of type `[object String]`.

instead of 

> Error `foo` is not of type `object` but of type `string`.

To prevent the case where someone does `throw {foo: 'bar'}` which will end up in the message

> Error `{foo: 'bar'}` is not of type `object` but of type `object`.

// @novemberborn @jamestalmage @sindresorhus

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -4,6 +4,7 @@ var deepEqual = require('only-shallow');
 var observableToPromise = require('observable-to-promise');
 var isObservable = require('is-observable');
 var isPromise = require('is-promise');
+var isError = require('is-error');
 var util = require('util');
 var x = module.exports;
 
@@ -103,8 +104,21 @@ x.throws = function (fn, err, msg) {
 			}
 		}, err, msg);
 
+		if (!isError(result)) {
+			throw result;
+		}
+
 		return result;
 	} catch (err) {
+		if (!isError(err)) {
+			var actual = Object.prototype.toString.call(err);
+			var expected = '[object Error]';
+			var message = 'Error `' + err + '` is not of type `' + expected + '` but of type `' + actual + '`.';
+
+			test(false, create(actual, expected, undefined, message, x.throws));
+			return;
+		}
+
 		test(false, create(err.actual, err.expected, err.operator, err.message, x.throws));
 	}
 };
--- a/package.json
+++ b/package.json
@@ -104,6 +104,7 @@
     "globby": "^4.0.0",
     "ignore-by-default": "^1.0.0",
     "is-ci": "^1.0.7",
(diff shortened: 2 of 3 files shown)


lib/assert.js
```diff
@@ -103,8 +104,21 @@ x.throws = function (fn, err, msg) {
 			}
 		}, err, msg);
 
+		if (!isError(result)) {
+			throw result;
+		}
```
SamVerschueren: This is necessary in case a developer writes this test

```js
test(t => {
    t.throws(() => {
        throw 'foo';
    });
});
```

`assert.throws` will validate that the function indeed throws an error (a string) and thus does not jump to line 113.

novemberborn: > assert.throws will validate that the function indeed throws an error (a string) and thus does not jump to line 113.

Which is correct. Used like that you're only asserting something was thrown. You don't care what it is.

lib/assert.js
```diff
@@ -103,8 +104,21 @@ x.throws = function (fn, err, msg) {
...
+
 		return result;
 	} catch (err) {
+		if (!isError(err)) {
+			var actual = Object.prototype.toString.call(err);
+			var expected = '[object Error]';
+			var message = 'Error `' + err + '` is not of type `' + expected + '` but of type `' + actual + '`.';
```
novemberborn: The expectation depends on the original `err` argument, be it a constructor, string, regular expression or validation function.

I'm starting to think we need to reimplement `assert.throws()` here. It's too hard to format our assertion error without duplicating its internals.

novemberborn:
@SamVerschueren hope it's OK if I close this for now. I reckon resolving #661 will require some more changes.

PR #686, opened 2016-03-27
rename `t.same()` to `t.deepEqual()`

kentcdodds:
Do not merge :-)

This is a work in progress PR just to get the conversation started. There are several tests that are failing and I could use some help there I think.

Thoughts?

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -130,11 +130,11 @@ export interface AssertContext {
 	 * @param error Can be a constructor, regex, error message or validation function.
 	 */
 	 /**
- 	 * Assert that value is deep equal to expected. (DEPRECATED, use `deepEqual`)
+ 	 * DEPRECATED, use `deepEqual`. Assert that value is deep equal to expected.
  	 */
  	same<U>(value: U, expected: U, message?: string): void;
  	/**
- 	 * Assert that value is not deep equal to expected. (DEPRECATED use `notDeepEqual`)
+ 	 * DEPRECATED use `notDeepEqual`. Assert that value is not deep equal to expected.
  	 */
  	notSame<U>(value: U, expected: U, message?: string): void;
  	/**
--- a/index.d.ts
+++ b/index.d.ts
@@ -120,11 +120,11 @@ export interface AssertContext {
 	/**
 	 * Assert that value is deep equal to expected.
 	 */
-	same<U>(value: U, expected: U, message?: string): void;
+	deepEqual<U>(value: U, expected: U, message?: string): void;
 	/**
 	 * Assert that value is not deep equal to expected.
 	 */
-	notSame<U>(value: U, expected: U, message?: string): void;
+	notDeepEqual<U>(value: U, expected: U, message?: string): void;
 	/**
 	 * Assert that function throws an error or promise rejects.
 	 * @param error Can be a constructor, regex, error message or validation function.
--- a/index.d.ts
+++ b/index.d.ts
@@ -129,6 +129,18 @@ export interface AssertContext {
 	 * Assert that function throws an error or promise rejects.
 	 * @param error Can be a constructor, regex, error message or validation function.
 	 */
+	 /**
+ 	 * Assert that value is deep equal to expected. (DEPRECATED, use `deepEqual`)
+ 	 */
+ 	same<U>(value: U, expected: U, message?: string): void;
+ 	/**
+ 	 * Assert that value is not deep equal to expected. (DEPRECATED use `notDeepEqual`)
+ 	 */
+ 	notSame<U>(value: U, expected: U, message?: string): void;
+ 	/**
+ 	 * Assert that function throws an error or promise rejects.
+ 	 * @param error Can be a constructor, regex, error message or validation function.
+ 	 */
 	throws(value: Promise<{}>, error?: ErrorValidator, message?: string): Promise<any>;
 	throws(value: () => void, error?: ErrorValidator, message?: string): any;
 	/**
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -141,7 +141,6 @@ x.ifError = x.error = function (err, msg) {
 /*
  * deprecated APIs
  */
-
 x.doesNotThrow = util.deprecate(x.notThrows, getDeprecationNotice('doesNotThrow()', 'notThrows()'));
 x.same = util.deprecate(x.deepEqual, getDeprecationNotice('same()', 'deepEqual()'));
 x.notSame = util.deprecate(x.notDeepEqual, getDeprecationNotice('notSame()', 'notDeepEqual()'));
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -130,8 +130,6 @@ x.notThrows = function (fn, msg) {
 	}
 };
 
-x.doesNotThrow = util.deprecate(x.notThrows, 't.doesNotThrow is renamed to t.notThrows. The old name still works, but will be removed in AVA 1.0.0. Update your references.');
-
 x.regex = function (contents, regex, msg) {
 	test(regex.test(contents), create(regex, contents, '===', msg, x.regex));
 };
@@ -143,11 +141,11 @@ x.ifError = x.error = function (err, msg) {
 /*
  * deprecated APIs
  */
+
+x.doesNotThrow = util.deprecate(x.notThrows, getDeprecationNotice('doesNotThrow()', 'notThrows()'));
 x.same = util.deprecate(x.deepEqual, getDeprecationNotice('same()', 'deepEqual()'));
 x.notSame = util.deprecate(x.notDeepEqual, getDeprecationNotice('notSame()', 'notDeepEqual()'));
 
 function getDeprecationNotice(oldApi, newApi) {
-  return `DEPRECATION NOTICE: ${oldApi} has been renamed to ${newApi} and will eventually be removed.` +
-    ` See https://github.com/jamestalmage/ava-codemods to help upgrade your codebase automatically.`;
+	return `DEPRECATION NOTICE: ${oldApi} has been renamed to ${newApi} and will eventually be removed. See https://github.com/jamestalmage/ava-codemods to help upgrade your codebase automatically.`;
 }
-
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -60,12 +60,12 @@ x.not = function (val, expected, msg) {
 	test(val !== expected, create(val, expected, '!==', msg, x.not));
 };
 
-x.same = function (val, expected, msg) {
-	test(deepEqual(val, expected), create(val, expected, '===', msg, x.same));
+x.deepEqual = function (val, expected, msg) {
+	test(deepEqual(val, expected), create(val, expected, '===', msg, x.deepEqual));
 };
 
-x.notSame = function (val, expected, msg) {
-	test(!deepEqual(val, expected), create(val, expected, '!==', msg, x.notSame));
+x.notDeepEqual = function (val, expected, msg) {
+	test(!deepEqual(val, expected), create(val, expected, '!==', msg, x.notDeepEqual));
 };
 
 x.throws = function (fn, err, msg) {
@@ -139,3 +139,21 @@ x.regex = function (contents, regex, msg) {
 x.ifError = x.error = function (err, msg) {
 	test(!err, create(err, 'Error', '!==', msg, x.ifError));
 };
+
+/*
+ * deprecated APIs
+ */
+x.same = function (val, expected, msg) {
+	deprecationNotice('same()', 'deepEqual()');
+	return x.deepEqual(val, expected, msg);
+};
+
+x.notSame = function (val, expected, msg) {
+	deprecationNotice('notSame()', 'notDeepEqual()');
+	return x.notDeepEqual(val, expected, msg);
+};
+
+function deprecationNotice(oldApi, newApi) {
+	console.warn(`DEPRECATION NOTICE: ${oldApi} has been renamed to ${newApi} and will eventually be removed`);
+}
+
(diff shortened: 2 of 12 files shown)


lib/assert.js
```diff
@@ -139,3 +139,21 @@ x.regex = function (contents, regex, msg) {
...
+	deprecationNotice('same()', 'deepEqual()');
+	return x.deepEqual(val, expected, msg);
+};
+
+x.notSame = function (val, expected, msg) {
+	deprecationNotice('notSame()', 'notDeepEqual()');
+	return x.notDeepEqual(val, expected, msg);
+};
+
+function deprecationNotice(oldApi, newApi) {
```
kentcdodds: I expect we'll want something a little more robust :-)

kentcdodds:
And shall I just say that because the tests for this project are written in tap rather than AVA, I'm coming to appreciate AVA more and more :-) :heart:

sindresorhus:
Thanks for starting the conversation @kentcdodds :)

-

We'll definitely need a [codemod](https://github.com/sindresorhus/ava/issues/644) if we're going with this.

sindresorhus:
Not to start a bikeshedding, but is there any other word we could use that might be even clearer?

kentcdodds:
Great to explore options for the best API possible while the topic is open for debate. I have no ideas though...

vdemedes:
Here's an idea: what if `t.same` becomes `t.is`? Couldn't be clearer and shorter to type. And `t.is` gets removed in favor of `t.true(a === b)`. 

I use `t.is` only for primitive values and when I need to check that references are equal. `t.same` also handles primitive values, so the only benefit of `t.is` is checking reference equality.

What do you think?

kentcdodds:
I'd be fine with that.

jamestalmage:
I think the best criteria by which to judge is this:

If someone who has never used tap/AVA/chai/etc is *reading* the test, what gives them the best chance of understanding what the assertion does without looking at the docs. Part of this includes avoiding ambiguity where possible. If a word could mean two different types of assertions, that's a big knock against it (we won't reach perfection here, because English).

Short names should be low on the list of priorities. There is a limit at which something is too long, but a few extra characters is worth any added clarity.

kentcdodds:
> Short names should be low on the list of priorities. There is a limit at which something is too long, but a few extra characters is worth any added clarity.

Couldn't agree more.

But I still have no suggestions to a better name than `deepEqual` :-)

kasperlewau:
> Not to start a bikeshedding, but is there any other word we could use that might be even clearer?

Off of the top of my head, here's a couple of candidates (of varying degrees of *suck*); 

* `t.sameStructure`
* `t.distinctEq`
* `t.reflect`
  * `t.reflection`
  * `t.reflectionOf`
* `t.mirror`
  * `t.mirrorOf`
* `t.comparable`
  * `t.comparableTo`
* `t.represent`
  * `t.representation`
  * `t.representationOf`
* `t.isomorphism`
  * `t.isomorphicTo`

The ones that I do like to some extent would be `sameStructure` and `comparableTo`. Possibly `distinctEq` as well, even though it's somewhat of an oxymoron. Can probably bin the rest swiftly :smile: 

I reckon `sameStructure` is the winner here. I find it very hard to misinterpret, and it conveys the intent of the assertion very clearly. I don't *quite* see it trumping `deepEqual` though. Just gotta keep crossing off those candidates I suppose..

--- 

> Here's an idea: what if `t.same` becomes `t.is`? Couldn't be clearer and shorter to type. And `t.is` gets removed in favor of `t.true(a === b)`.

I agree, in parts. It's ridiculously short to type, and I think opting for `t.true(a === b)` is a good way to go so there's that going for it. But I don't think `t.is` adequately conveys the intention of comparing two distinct objects of the same structure.

I'm sorry to say I don't have any *better* suggestions, aside from the list of verbose candidates in my first segment - along with `deepEqual`.

*My dream scenario would be for the `deepEquality` assertion to die off entirely and we'd have something like `t.true(Object.equals(a,b))`*

In all honesty, I find `deepEqual` hard to beat - purely based on the fact that it's "widely adopted". It's probably something most of us are already familiar with. If not, it's pretty easy to find out with a quick Google search as it's not widely adopted outside of computer programming from what I can find. `same`, `is`, and most of my suggestions above - are and could be.

jamestalmage:
My concern with `t.is` is what happens to users who fail to update current `t.is` assertions. We leave them open to subtle bugs. Also, it doesn't pass the disambiguation test.

novemberborn:
Regardless which name we pick we should carefully document which values are considered passing and which aren't. E.g. #630 surfaced that we were requiring constructor equality.

I wonder if the "same" or "equal" part is implicit. You're testing an assertion with an actual and an expected value. `t.same()` is confusing because obviously you want things to be the same… and then what does `t.is()` mean?

Given that, how about `t.deep()`?

novemberborn:
Which would lead to `t.strict()` for `t.is()`, if we want to go that far.

kentcdodds:
> how about `t.deep()`?

I still prefer `deepEqual`. To me it's confusing without `equal`.

sindresorhus:
I'm good with `deepEqual`.

-

We should really have a codemod for this change before releasing it. Anyone interested in trying to make one (I especially encourage community members to try)? Resources: https://github.com/sindresorhus/ava/issues/644

jamestalmage:
> Anyone interested in trying to make one (I especially encourage community members to try)? 

Looping in our AST friends: @twada @jfmengels

spudly:
Whipped up a quick codemod for this: https://astexplorer.net/#/AbMVMKHZhp

I'm happy to put this in a PR. Where should I put the files? Perhaps in `/codemods`?

jfmengels:
@spudly :+1:!

I'd create a new project containing the codemods (ava-codemods?). Don't see the value of putting it in the AVA project if it's only going to be run once at the very most.
There is talk about writing other (complex) codemods too in https://github.com/sindresorhus/ava/issues/644.

jfmengels:
@spudly I improved your code a bit by providing support for `notSame` --> `notDeepEqual` (and some necessary refactoring to avoid duplicating) https://astexplorer.net/#/yXo9G9wx9m/1

spudly:
@sindresorhus, @jamestalmage, @vdemedes, @novemberborn

Opinions about where to put codemods? I was thinking in `/codemods` but @jfmengels recommended a separate repo. I'll do whatever.

spudly:
@jfmengels,

Looks good and nice refactor. Guess I forgot about `notSame`/`notDeepEqual` because I never use them...

sindresorhus:
See https://github.com/jamestalmage/ava-codemods.

lib/assert.js
```diff
@@ -139,3 +139,21 @@ x.regex = function (contents, regex, msg) {
...
+	console.warn(`DEPRECATION NOTICE: ${oldApi} has been renamed to ${newApi} and will eventually be removed`);
```
sindresorhus: We should mention the codemods here or at least a link to docs that mention it.

kentcdodds: Definitely. Would you like a more robust deprecation abstraction, or is this simple function and `console.warn` sufficient?

sindresorhus:
@kentcdodds We now have a codemod, so this could potentially go into the next release. Would you be able to finish this up and fix the failing tests?

kentcdodds:
Could definitely use some help on this to know why tests are failing. It's confusing with mix of `ava` and `assert` :-)

I definitely want to make this happen soon! :D

lib/assert.js
sindresorhus: You could use `util.deprecate`.

kentcdodds:
Updated the deprecation to use `util.deprecate` (TIL) and added a note about the codemod repo. I'll try to give the failing tests another look, but last time I know I had a bit of trouble with them.

test/assert.js
```diff
@@ -268,12 +268,46 @@ test('.same() should not mask RangeError from underlying assert', function (t) {
...
+		assert.deepEqual(a, b);
 	});
 
 	t.end();
 });
+
+test('.same() should log a deprecation notice', function (t) {
+	var calledWith;
+	var originalConsole = console.warn;
+	console.warn = function () {
```
kentcdodds: Turns out that when using `util.deprecate`, it actually logs to `console.log` which is fine, but I can't seem to get this test to work because when I reassign `console.log` it appears that my reassignment is never called. I'm not sure what's going on here... Any tips appreciated.

Alternatively, we can trust that `util.deprecate` works and not worry about writing tests for this case... ¯\\\_(ツ)_/¯

jamestalmage: Looks like deprecate calls an undocumented method [`process.emitWarning`](https://github.com/nodejs/node/blob/master/lib/internal/util.js#L18-L24).

That ends up causing a [`warning` event to be emitted on `process`](https://github.com/nodejs/node/blob/master/lib/internal/process/warning.js#L47), which seems to be [handled by writing to `console.error`](https://github.com/nodejs/node/blob/master/lib/internal/process/warning.js#L18-L24).

So try `console.error`?

Or just trust that `util.deprecate` works.

kentcdodds: > Or just trust that `util.deprecate` works.

If everyone's ok with that, I think that's what I wanna do :-)

kentcdodds:
Alrighty, this should be ready to go I think. Feedback welcome!

lib/assert.js
```diff
@@ -130,8 +130,6 @@ x.notThrows = function (fn, msg) {
 	}
 };
 
-x.doesNotThrow = util.deprecate(x.notThrows, 't.doesNotThrow is renamed to t.notThrows. The old name still works, but will be removed in AVA 1.0.0. Update your references.');
```
kentcdodds: Moved this down to the other deprecated APIs

test/assert.js
sindresorhus: Node.js core uses `util.deprecate`. So I would trust it.

lib/enhance-assert.js
```diff
@@ -8,8 +8,8 @@ module.exports.PATTERNS = [
 	't.false(value, [message])',
 	't.is(value, expected, [message])',
 	't.not(value, expected, [message])',
-	't.same(value, expected, [message])',
-	't.notSame(value, expected, [message])',
+	't.deepEqual(value, expected, [message])',
+	't.notDeepEqual(value, expected, [message])',
```
kentcdodds: This bit made things tricky for me in `test/test.js` which asserts `t.is(result.result.assertCount, 1);` and there was no real way for me to know that this is what causes `assertCount` to increment. Still not sure I follow but this fixed it so ¯\\\_(ツ)_/¯

kentcdodds:
Also, let me know if you want me to squash and clean things up with the commit message or if you're fine using GitHub's fancy new squash and merge feature :-)

lib/enhance-assert.js
jamestalmage: The `power-assert` enhancement actually has callbacks (`onSuccess`, `onFailure`) that are called with the result of each assertion. That is where we implement assertion tracking.

I think you should probably keep `t.same` and `t.notSame` in the patterns until they are officially removed. Otherwise this will break assertionCount for them. Maybe add a test that ensures the assertion count is tracked for those (which we will remove when we drop support).

jamestalmage:
We will use GH' squash and merge.

lib/assert.js
```diff
@@ -139,3 +137,14 @@ x.regex = function (contents, regex, msg) {
...
+
+/*
+ * deprecated APIs
+ */
+x.doesNotThrow = util.deprecate(x.notThrows, getDeprecationNotice('doesNotThrow()', 'notThrows()'));
+x.same = util.deprecate(x.deepEqual, getDeprecationNotice('same()', 'deepEqual()'));
+x.notSame = util.deprecate(x.notDeepEqual, getDeprecationNotice('notSame()', 'notDeepEqual()'));
+
+function getDeprecationNotice(oldApi, newApi) {
+	return `DEPRECATION NOTICE: ${oldApi} has been renamed to ${newApi} and will eventually be removed. See https://github.com/jamestalmage/ava-codemods to help upgrade your codebase automatically.`;
```
jamestalmage: We don't transpile AVA. Can't use string templates on Node 0.10

kentcdodds: Ah, good to know. Will update.

jamestalmage:
A few notes, mostly related to making sure we are deprecating `t.same` without breaking it.

Once those are cleared up, LGTM.

kentcdodds:
Updated PR to respond to @jamestalmage's comments. Thanks!

index.d.ts
```diff
@@ -120,11 +120,11 @@ export interface AssertContext {
 	/**
 	 * Assert that value is deep equal to expected.
 	 */
-	same<U>(value: U, expected: U, message?: string): void;
+	deepEqual<U>(value: U, expected: U, message?: string): void;
```
kentcdodds: I just left the original there and commented that they're deprecated. ¯\\\_(ツ)_/¯

jamestalmage: That's probably the only way to handle it. Looks like TS has a two year old issue on the topic:
https://github.com/Microsoft/TypeScript/issues/390

index.d.ts
```diff
@@ -129,6 +129,18 @@ export interface AssertContext {
 	 * Assert that function throws an error or promise rejects.
 	 * @param error Can be a constructor, regex, error message or validation function.
 	 */
+	 /**
+ 	 * Assert that value is deep equal to expected. (DEPRECATED, use `deepEqual`)
+ 	 */
+ 	same<U>(value: U, expected: U, message?: string): void;
+ 	/**
+ 	 * Assert that value is not deep equal to expected. (DEPRECATED use `notDeepEqual`)
```
jamestalmage: Move the `DEPRECATED` warning to the start?

IDE's use these comments for tooltips. So it would be good to have the warning as prominent as possible.

jamestalmage:
LGTM.

index.d.ts
kentcdodds: Done

index.d.ts
SamVerschueren: I think that's the best way to handle this situations as well. It will indeed throw a compile error because it could not find the `same` method.

novemberborn:
:shipit:!

sindresorhus:
Yay for better user UX. Thanks for nudging us towards this and doing the work @kentcdodds :)

kentcdodds:
Fantastic! Did we decided on doing truthy/falsy? I'm happy to do that work as well

sindresorhus:
Yes. We're going with that. PR super appriciated :)

PR #688, opened 2016-03-29
Fix command format in readme

alexbooker:

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -120,7 +120,7 @@ $ npm test
 ### Watch it
 
 ```console
-$ npm test -- --watch
+$ npm test --watch
 ```
 
 AVA comes with an intelligent watch mode. [Learn more in its recipe](docs/recipes/watch-mode.md).


sindresorhus:
The current command is correct. `npm test` swallows flags, so we use the `--` to indicate it should ignore the following flags.

alexbooker:
Hm. I still don't really understand why the `--` is needed. 

Would it make sense to elaborate under the ["Running tests with watch mode enabled" section](https://github.com/sindresorhus/ava/blob/master/docs/recipes/watch-mode.md#running-tests-with-watch-mode-enabled)?

sindresorhus:
http://unix.stackexchange.com/questions/11376/what-does-double-dash-mean-also-known-as-bare-double-dash

BarryThePenguin:
@alexbooker the npm docs elaborate on passing arguments to [npm run-script](https://docs.npmjs.com/cli/run-script)

alexbooker:
Thanks, both.

PR #694, opened 2016-03-30
throw error if required dependency could not be found - fixes #691

### LINKED ISSUE #691: `path must be a string` when requiring uninstall dependency
## Description

When using `babel-register` as required dependency of AVA, I got this error.

```
assert.js:93
  throw new assert.AssertionError({
        ^
AssertionError: path must be a string
    at Module.require (module.js:362:3)
    at require (module.js:380:17)
    at Array.forEach (native)
    at Object.<anonymous> (/Users/sam/Projects/test/node_modules/ava/lib/test-worker.js:48:22)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Function.Module._load (module.js:312:12)
    at Function.Module.runMain (module.js:497:10)
    at startup (node.js:119:16)
    at node.js:929:3
```

It took me a while to figure out that it was because I didn't install the `babel-register` dependency and [resolve-cwd](https://www.npmjs.com/package/resolve-cwd) returns `null` for dependencies not found.

Because of [this](https://github.com/sindresorhus/ava/blob/master/api.js#L31) line, the `require` argument results in `[null]` and this fails when trying to `require` `null` off course. It would be nice if it failed with a decent error like `Could not find dependency 'babel-register'` or whatever dependency could not be found.

## Environment

This was only the case on `v0.10.36` for me. Not sure why though but probably because it uses pre-npm@3 and thus not flattens the dependency tree. I don't really think this issue is platform dependent.

```
Node.js v0.10.36
darwin 15.3.0
```

SamVerschueren:
This PR fixes #691 by throwing an error if `resolve-cwd` returns `null`.

### DIFF
--- a/api.js
+++ b/api.js
@@ -28,8 +28,15 @@ function Api(options) {
 	EventEmitter.call(this);
 
 	this.options = options || {};
-	this.options.require = (this.options.require || []).map(resolveCwd);
 	this.options.match = this.options.match || [];
+	this.options.require = (this.options.require || []).map(function (moduleId) {
+		var ret = resolveCwd(moduleId);
+		if (ret === null) {
+			throw new Error('Could not resolve required module \'' + moduleId + '\'');
+		}
+
+		return ret;
+	});
 
 	this.excludePatterns = [
 		'!**/node_modules/**',
--- a/test/api.js
+++ b/test/api.js
@@ -533,6 +533,15 @@ test('Node.js-style --require CLI argument', function (t) {
 		});
 });
 
+test('Node.js-style --require CLI argument module not found', function (t) {
+	t.plan(1);
+
+	t.throws(function () {
+		/* eslint no-new: 0 */
+		new Api({require: ['foo-bar']});
+	}, /^Could not resolve required module 'foo-bar'$/);
+});
+
 test('power-assert support', function (t) {
 	t.plan(3);
 


jamestalmage:
I think this is better solved in resolve-cwd. Shouldn't it throw if it can't be resolved?

SamVerschueren:
From the docs:

> Unlike `require.resolve()` it returns `null` instead of throwing when the module can't be found.

Willing to do a PR to another library to make this work.

sindresorhus:
> Shouldn't it throw if it can't be resolved?

That's annoying for most use-cases when you just want to check if something exists and fallback if not. Our use-case is different as we're exposing it directly to the end-user.

> Willing to do a PR to another library to make this work.

Happy to merge a PR on `resolve-cwd` that  adds such option. PR should be on `resolve-from` and then exposed in `resolve-cwd`. Should be disabled by default though.

jamestalmage:
Rethinking my first response. This does allow us to throw a more specific error, redirecting the user to their AVA config.

SamVerschueren:
> This does allow us to throw a more specific error, redirecting the user to their AVA config.

:+1: 

And probably if we throw it somewhere down the chain, we will eventually try-catch the error to throw a custom message.

api.js
```diff
@@ -28,8 +28,14 @@ function Api(options) {
 	EventEmitter.call(this);
 
 	this.options = options || {};
-	this.options.require = (this.options.require || []).map(resolveCwd);
 	this.options.match = this.options.match || [];
+	this.options.require = (this.options.require || []).map(function (module) {
```
novemberborn: Minor nit, maybe `moduleId` rather than `module`? `module` is in the file scope as the module object. `resolve-cwd` [uses `moduleId` as well](https://www.npmjs.com/package/resolve-cwd#resolvecwdmoduleid).

novemberborn:
LGTMO!

api.js
SamVerschueren: Good one!

SamVerschueren:
Fixed

api.js
```diff
@@ -28,8 +28,14 @@ function Api(options) {
...
 
 	this.options = options || {};
-	this.options.require = (this.options.require || []).map(resolveCwd);
 	this.options.match = this.options.match || [];
+	this.options.require = (this.options.require || []).map(function (moduleId) {
+		var ret = resolveCwd(moduleId);
+		if (ret === null) {
+			throw new Error('Could not resolve required module \'' + moduleId + '\'');
+		}
+		return ret;
```
vdemedes: Missing newline here

test/api.js
```diff
@@ -533,6 +533,19 @@ test('Node.js-style --require CLI argument', function (t) {
 		});
 });
 
+test('Node.js-style --require CLI argument module not found', function (t) {
+	t.plan(1);
+
+	try {
```
vdemedes: Perhaps use `t.throws()` here?

api.js
SamVerschueren: Do you mean an empty line between `}` and `return`?

test/api.js
```diff
@@ -533,6 +533,17 @@ test('Node.js-style --require CLI argument', function (t) {
...
+test('Node.js-style --require CLI argument module not found', function (t) {
+	t.plan(1);
+
+	function create() {
+		/* eslint no-new: 0 */
+		new Api({require: ['foo-bar']});
+	}
+
+	t.throws(create, /^Could not resolve required module 'foo-bar'$/);
+});
```
SamVerschueren: Using `t.throws` here, can't get around the extra `create` function right?

api.js
vdemedes: Yeah

test/api.js
vdemedes: You could just use anonymous function.

SamVerschueren:
Processed all the feedback

test/api.js
```diff
@@ -533,6 +533,12 @@ test('Node.js-style --require CLI argument', function (t) {
 		});
 });
 
+test('Node.js-style --require CLI argument module not found', function (t) {
+	t.plan(1);
+
+	t.throws(() => new Api({require: ['foo-bar']}), /^Could not resolve required module 'foo-bar'$/);
```
vdemedes: ES6 alert! Fails on `0.12` and `0.10`.

SamVerschueren: Damnit. This is what happens when you're used to AVA :D

SamVerschueren: fixed

novemberborn:
:shipit:

SamVerschueren:
Thanks guys for the feedback and the merge! :beers:

PR #701, opened 2016-04-02
Clean up caching-precompiler

vdemedes:


lib/caching-precompiler.js
```diff
@@ -1,142 +1,180 @@
 'use strict';
-var fs = require('fs');
 var path = require('path');
+var fs = require('fs');
+var generateMapFileComment = require('convert-source-map').generateMapFileComment;
+var convertSourceMap = require('convert-source-map');
```
novemberborn: Is there any particular dependency order we're striving for?

lib/caching-precompiler.js
```diff
@@ -1,142 +1,180 @@
...
-		return options;
+module.exports = CachingPrecompiler;
+
+CachingPrecompiler.prototype._factory = function () {
+	this.init();
+
+	return this._transform;
+};
+
+CachingPrecompiler.prototype.init = function () {
```
novemberborn: Maybe inline this logic in `_factory`? `CachingPrecompiler` is really a wrapper around `caching-transform`. `init()` implies it needs to be explicitly initialized which isn't the case.

novemberborn:
Perhaps the methods in `CachingPrecompiler` could be reordered? There's a fair amount of jumping back and forth now when trying to understand the entire file.

I'm not a fan of the `Path` suffix on the variable names. I'd prefer sticking to `filename` (since Babel uses it) and `Dir` for directories.

test/caching-precompiler.js
```diff
@@ -145,6 +145,6 @@ test('uses babelConfig for babel options when babelConfig is an object', functio
 	t.true('inputSourceMap' in options);
 	t.false(options.babelrc);
 	t.same(options.presets, ['stage-2', 'es2015']);
-	t.same(options.plugins, [customPlugin, powerAssert, transformRuntime, rewrite]);
+	t.same(options.plugins, [customPlugin, transformRuntime, rewrite, powerAssert]);
```
novemberborn: Is this a bugfix or harmless?

api.js
```diff
@@ -64,8 +64,12 @@ Api.prototype._reset = function () {
 };
 
 Api.prototype._runFile = function (file) {
+	var hash = this.precompiler.precompileFile(file);
+	var precompiled = {};
+	precompiled[file] = hash;
```
novemberborn: :clap: for making this more explicit.

novemberborn:
I like the smaller functions, makes the whole thing easier to understand.

lib/caching-precompiler.js
sindresorhus: @novemberborn Yes, https://github.com/jfmengels/eslint-plugin-import-order/blob/master/docs/rules/import-order.md, which will soon be included in XO.

test/caching-precompiler.js
jamestalmage: Ideally, powerAssert would run earlier so it captures source as written, instead of as transformed.

https://gist.github.com/jamestalmage/885e1d0d732557cfb5cf854aee04fd19?ts=2

lib/caching-precompiler.js
novemberborn: Ah cool, that's the strategy I've been employing myself. I got thrown off here by the reordering of `fs` and `path`, looks like that plugin doesn't enforce any specific sorting within the categories?

vdemedes: I order them by variable length (descending).

lib/caching-precompiler.js
vdemedes: I'd prefer to rename `init` to `_init` (indicate that it's private) instead, as it's clear what containing code is about.

test/caching-precompiler.js
vdemedes: Moved back to previous order.

vdemedes:
PR updated.

vdemedes:
@sindresorhus @jamestalmage @novemberborn :shipit:❓

jamestalmage:
LGTM

novemberborn:
:shipit:

sindresorhus:
cleanup

Issue #575, opened 2016-02-28
Test-like output when test file does not exist


vdemedes:
The output is not what it should look like (latest master):

```
$ ava example.js
```

(example.js does not exist).

<img width="315" alt="screen shot 2016-02-28 at 1 46 21 pm" src="https://cloud.githubusercontent.com/assets/697676/13379285/c960aef4-de21-11e5-90d0-eee4ea62e640.png">

First of all, notice 1 extra line at the start & end of the output. And top-level CLI errors should be displayed as:

```

  ❎ Couldn't find any files to test

```

novemberborn:
This is probably due to <https://github.com/novemberborn/ava/commit/9dd03d5d2b20de48e72cb59e29573dd412201b1b> (see #502). There's some variance between the reporters, looks like the verbose one should render a cross.

Maybe this particular exception should be treated differently in the mini reporter. Or perhaps a general case can be made when there is 1 exception and no tests where found.

vdemedes:
No no, this kind of top-level output from CLI should be written without reporters, as those are fatal errors.

novemberborn:
> No no, this kind of top-level output from CLI should be written without reporters, as those are fatal errors.

Not when watching though. It's legit to start a watcher even if there are no test files yet. Or to remove the only test file while the watcher is running.

sotojuan:
What's the status of this? A simple edit to `reporters/mini.js` fixes the extra lines, but what about the cross? Also an easy fix in the reporter file.

novemberborn:
If we can agree that these are not fatal errors then I think we can fix up the reporters.

sotojuan:
Closing because this was fixed a while back in #704.

PR #704, opened 2016-04-04
Add test-like output when test file does not exist

sotojuan:
This helps with #575. Kinda.

First as @novemberborn said:

> If we can agree that these are not fatal errors then I think we can fix up the reporters.

But if we agree to do them in the reporters, I fixed up the extraneous new lines and added a cross for `AvaError` type of exceptions.

Here's how test on a nonexistent file and a regular one looks like:

screen shot 2016-04-03 at 8 08 08 pm

Also there are a lot of places I could've chosen to remove newlines from, so tell me if there are any better ones (I did this quickly to see what you guys think!).

### DIFF
--- a/lib/reporters/mini.js
+++ b/lib/reporters/mini.js
@@ -7,6 +7,7 @@ var spinners = require('cli-spinners');
 var chalk = require('chalk');
 var cliTruncate = require('cli-truncate');
 var colors = require('../colors');
+var cross = require('figures').cross;
 
 chalk.enabled = true;
 Object.keys(colors).forEach(function (key) {
@@ -131,7 +132,7 @@ MiniReporter.prototype.reportCounts = function () {
 		status += '\n   ' + colors.todo(this.todoCount, 'todo');
 	}
 
-	return status.length ? status : '\n';
+	return status.length ? status : '';
 };
 
 MiniReporter.prototype.finish = function () {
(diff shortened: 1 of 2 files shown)


lib/reporters/mini.js
```diff
@@ -131,7 +132,7 @@ MiniReporter.prototype.reportCounts = function () {
 		status += '\n   ' + colors.todo(this.todoCount, 'todo');
 	}
 
-	return status.length ? status : '\n';
+	return status.length ? status : '';
```
novemberborn: Could just return `status` now.

lib/reporters/mini.js
```diff
@@ -200,7 +201,7 @@ MiniReporter.prototype.finish = function () {
 
 MiniReporter.prototype.write = function (str) {
 	cliCursor.hide();
-	this.currentStatus = str + '\n';
+	this.currentStatus = String(str);
```
novemberborn: Why this change?

sotojuan: `xo` rule.

novemberborn: Well before it was concatenating with a newline. That's been removed so presumably you could just assign `this.currentStatus = str`?

sotojuan:
Thanks for the tips, fixed.

sindresorhus:
Thanks @sotojuan :)

Issue #673, opened 2016-03-22
A way to submit tests that should fail. 


jamestalmage:
It's really awesome when users report errors by submitting a failing test. I think it would be cool if AVA provided a way to merge those without killing the build.

I am thinking:

``` js
test.failing('this should work, but does not', t => { 
  ...
})
```

The difference from `.skip` would be:
1. It actually runs the test, it just expects the failure.
2. When it fails, it gets listed in red in the final output ("expected failure" or some such message), but the process still exits with 0. This keeps reminding you you've got a problem, but doesn't kill your CI process.
3. If the test passes, that is a failure (you've fixed the problem! hooray! remove the `.failing` flag).

novemberborn:
Should this mandate an issue link? Bit chicken-and-egg but would save you digging through git blame etc.

jamestalmage:
> Should this mandate an issue link?

That's a cool idea. Optional maybe? Not everyone uses AVA for open source projects with issue trackers (though really, that's the target audience for this feature). We could create an `eslint-plugin-ava` rule to make it mandatory.

novemberborn:
Yea maybe not make it mandatory.

How's this?

```js
test.failing('this should work, but does not', t => { 
  ...
}).see('https://github.com/sindresorhus/ava/issues/673')
```

jamestalmage:
:+1: `.see` should throw a helpful message (including the test title) if used without `.failing`.

Also, if we move ahead, we should raise issues on `eslint-plugin-ava` to enforce `use-see-with-failing` and `no-see-without-failing`.

novemberborn:
> .see should throw a helpful message (including the test title) if used without .failing.

Or it's only implemented on the `.failing` return value. We'll see.

sindresorhus:
I think this is a very cool idea, and would hopefully encourage more people to submit failing tests instead of issues. Not sure about `.see`, maybe `.issue` or `.reference`?

BarryThePenguin:
I think this comes back into `todo` and `skip` functionality too. I canrethink and come back with some suggestions

jamestalmage:
In my mind:

- `skip` - implementation that isn't run.
- `todo` - no implementation.
- `failing` - implementation that is run, but expected to fail.

novemberborn:
@sindresorhus ,

> Not sure about .see, maybe .issue or .reference?

[...]

PR #705, opened 2016-04-04
Make `test.todo(title, fn)` produce a failure

### LINKED ISSUE #697: `test.todo(title, fn)` should produce a failure
## Overview: 

My understanding of the `skip` vs `todo` vs `failing` API was:

1. `test.skip([title], fn)` - A test function is provided, but it is skipped(i.e. `fn` is never called). Common use: removing a failing test while you are in the middle of refactoring. People probably shouldn't be checking `test.skip` tests into their master branch (especially once `test.failing` is implemented). Title is optional, just like the normal `test` method.

2. `test.todo(title)` - No test function is allowed, but a title is required. Common use: documenting future tests (and maybe features) that need to be written.

3. `test.failing([title], fn)` - Proposed in #673- Optional title, test function is required and **will** be run (but is expected to fail). Common use: allow users to create PR's that reproduce the problem. Reporting bugs in a way that does not break the build.

--- 

## Issue:

Right now the API for `2` is more like: `test.todo(title, [fn])` - `title` is required, but `fn` is optional (and ignored).

sotojuan:
Fixes #697.

Let me know if a better error message helps!

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -58,7 +58,7 @@ optionChain(chainableMethods, function (opts, title, fn) {
 	}
 
 	if (opts.todo && typeof fn === 'function') {
-		throw new TypeError('`todo` tests are not allowed to have a function');
+		throw new TypeError('`todo` tests are not allowed to have a function. Use `test.skip()` for tests with a function.');
 	}
 
 	if (opts.todo) {
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -57,11 +57,11 @@ optionChain(chainableMethods, function (opts, title, fn) {
 		title = null;
 	}
 
-	if (opts.todo && typeof fn === 'function') {
-		throw new TypeError('`todo` tests are not allowed to have a function. Use `test.skip()` for tests with a function.');
-	}
-
 	if (opts.todo) {
+		if (typeof fn === 'function') {
+			throw new TypeError('`todo` tests are not allowed to have a function. Use `test.skip()` for tests with a function.');
+		}
+
 		fn = noop;
 
 		if (typeof title !== 'string') {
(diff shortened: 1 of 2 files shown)


jamestalmage:
I think the message should include instructions to "use `test.skip` instead".

When #673 lands we can amend it to say "use `test.skip` or `test.fail` instead".

Otherwise :+1:

lib/runner.js
```diff
@@ -57,6 +57,10 @@ optionChain(chainableMethods, function (opts, title, fn) {
 		title = null;
 	}
 
+	if (opts.todo && typeof fn === 'function') {
+		throw new TypeError('`todo` tests are not allowed to have a function. Use `test.skip()` for tests with a function.');
+	}
+
```
novemberborn: Should combine this with the if-block below which also applies to `opts.todo`.

novemberborn:
I wonder whether we should change "function" here (and in the other messages) to "implementation". It's currently [named `callback` in the docs](https://github.com/sindresorhus/ava#callbackt). I struggled quite  a bit with the wording there.

lib/runner.js
sotojuan: That's what I was doing before but it wasn't working because of the `fn = noop` line (which seems to be used to "override" any given function), and if I removed that other stuff broke.

novemberborn: Try setting `fn = noop` after testing `fn`.

sotojuan: Got it working, thanks!

sotojuan:
You guys decide! Let me know—either seems clear to me.

jamestalmage:
Use implementation.

sotojuan:
All right, changed to "implementation". If it looks good I can do a PR for the other messages.

novemberborn:
Thanks @sotojuan!

> If it looks good I can do a PR for the other messages.

That'd be great. Could you update the readme as well?

PR #706, opened 2016-04-04
add tests that demonstrate various ways to use babel config.

jamestalmage:
Issue #448 was kept open, in part, because we wanted to enable an `extends` option.
What we are after is actually possible using existing babel options.

A separate PR with recipes is forthcoming.

### DIFF
--- a/test/api.js
+++ b/test/api.js
@@ -5,7 +5,7 @@ var figures = require('figures');
 var rimraf = require('rimraf');
 var test = require('tap').test;
 var Api = require('../api');
-var testDoublerPlugin = require('./fixture/babel-plugin-test-doubler');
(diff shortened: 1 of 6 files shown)


test/api.js
```diff
@@ -5,7 +5,7 @@ var figures = require('figures');
 var rimraf = require('rimraf');
 var test = require('tap').test;
 var Api = require('../api');
-var testDoublerPlugin = require('./fixture/babel-plugin-test-doubler');
+var testCapitilizerPlugin = require('./fixture/babel-plugin-test-capitilizer');
```
novemberborn: `capitalizer`

novemberborn:
Just that typo. LGTM otherwise.

sindresorhus:
LGTM

Issue #619, opened 2016-03-08
Make `"babel": "inherit"` the default


kentcdodds:
## Description

I've shared AVA with a lot of people and the number one thing that trips them up is that the Babel configuration that transpiles their own code is different from the config that transpiles their test code. For most of them, if they're writing their source in ES6, they expect their tests to be able to transpile with the same config.

So my request is that we default to `"babel": "inherit"` where we look for a `.babelrc` and if it exists, use that for config.

Thoughts :thought_balloon:

novemberborn:
It'd be hard to guarantee the behavior AVA promises (async/await for instance) if it might be affected by users own code.

Maybe `ava --init` could detect a local Babel install and suggest this setting. And automatically set up `babel-register` too!

jamestalmage:
Are people usually adding stage-3 to their Babel configs? Because AVA really needs that to provide all the async/await goodness. 

Perhaps we should automatically add stage-3 to the detected `.babelrc`? If the user goes so far as to actually specify the `inherit` option then we assume they know what they are doing and only add  power-assert?

There is a tension here between doing too much magic and asking the user to write extraneous config. Always a hard decision.

kentcdodds:
> It'd be hard to guarantee the behavior AVA promises (async/await for instance) if it might be affected by users own code.

I think that people will understand that if their babel config doesn't support something, they can't use it. They're coding their ES6 in their source with this config all day every day. I think it'll be natural to continue using what they're used to using.

But if you'd rather, we could validate that the config has the properties necessary.

I like the idea of an `ava --init` but not everyone will set up AVA that way and I think it'll still be confusing for people.

> Perhaps we should automatically add stage-3 to the detected `.babelrc`

Seems like to much :sparkles: magic :sparkles: to me.

> Always a hard decision.

Agreed :-) But like I said, I hear over and over again that people have trouble understanding that the tests and source are compiled differently. It's just confusing to people. So I think we could do *something* about it. :-)

levithomason:
-1 to magic, +1 to defaults or some kind of `init`.  I ran into a hangup here as well.  The docs seem to indicate if you `"babel": "inherit"` in your `ava` config, then your tests are transpiled same as your source.  However, I also have to `"require": "babel-register"` nothing was transpiled.

So, I'd suggest the default be at least the minimal to start testing using your current source babel config:
```
"ava": {
  "babel": "inherit",
  "require": [
    "babel-register"
  ]
},
```

[...]

PR #708, opened 2016-04-04
Add a recipe for configuring `.babelrc`.

### LINKED ISSUE #448: A possible way to provide customizable babel configs using `.babelrc`/`env`

jamestalmage:
Tests verifying all these recommendations were added to the test suite in #706.

I think this is enough to close #448

### DIFF
--- a/docs/recipes/babelrc.md
+++ b/docs/recipes/babelrc.md
@@ -1,0 +1,86 @@
+# Configuring Babel
+
+There are multiple options for configuring how AVA transpiles your tests using Babel.
+
+ - [Specify a complete config in `package.json`](#specify-a-complete-config-in-packagejson)
+ - [Extend existing `.babelrc` without modification](#extend-existing-babelrc-without-modification)
+ - [Extend existing `.babelrc` with additional plugins or presets](#extend-existing-babelrc-with-additional-plugins-or-presets)
+ - [Extend an alternate config file (i.e. not `.babelrc`)](#extend-alternate-config-file)
+ - [Notes](#notes)
+
+## Specify a complete config in `package.json`
+
+The `babelrc` option defaults to `false`, meaning `.babelrc` files are not considered when transpiling tests. This means you must specify your complete config in `package.json`.
+
+```json
+{
+  "ava": {
+    "babel": {
+      "plugins": ["rewire"],
+      "presets": ["es2015"]
+    }
+  }
+}
+```
+
+## Extend existing `.babelrc` without modification
+
+Use the `"inherit"` shortcut if you want your tests transpiled the same as your sources. This will use your `.babelrc` directly (with a few additional [internal plugins](#notes)).
+
+`package.json`:
+
+```json
+{
+ "ava": {
+   "babel": "inherit"
+ }
+}
+```
+
+## Extend existing `.babelrc` with additional plugins or presets
+
+Set `babelrc` to `true`. This will use your `.babelrc` and extend it with any additional plugins specified.
+
+`package.json`:
+
+```json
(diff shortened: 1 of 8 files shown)


docs/recipes/babelrc.md
```diff
@@ -0,0 +1,87 @@
...
+```
+
+## Notes
+
+AVA *always* adds a few custom Babel plugins when transpiling your plugins. They serve a variety of functions:
+
+ * Enable `power-assert` support.
+ * Rewrite require paths internal AVA dependencies like `babel-runtime` (important if you are still using `npm@2`).
+ * Generate test metadata to determine which files should be run first (*future*).
+ * Static analysis of dependencies for precompilation (*future*).
```
novemberborn: Should we add these now, or update later?

readme.md
```diff
@@ -901,6 +901,7 @@ It's the [Andromeda galaxy](https://simple.wikipedia.org/wiki/Andromeda_galaxy).
 - [When to use `t.plan()`](docs/recipes/when-to-use-plan.md)
 - [Browser testing](docs/recipes/browser-testing.md)
 - [TypeScript](docs/recipes/typescript.md)
+- [Configuring Babel](docs/recipes/babelrc.md)
```
novemberborn: Should also add a link in the [ES2015 section](https://github.com/sindresorhus/ava/blob/3120ae6e0f2fad7244426faee93f548ec203a4e8/readme.md#es2015-support).

docs/recipes/babelrc.md
```diff
@@ -0,0 +1,87 @@
...
+      "plugins": ["custom-plugin-name"],
+      "presets": ["custom-preset"]
+    }
+  }
+}
+```
+
+The above uses `babel-test-config.json` as the base config, and extends it with the custom plugins and presets specified.
+
+## Specify a complete config in `package.json`
```
sindresorhus: I would put this at the top, since it's the default.

docs/recipes/babelrc.md
```diff
@@ -0,0 +1,87 @@
...
+
+When the `babelrc` option is `false` (the default), `.babelrc` is not read for transpiling tests. You must specify your complete config in `package.json`.
+
+```json
+{
+  "ava": {
+    "babel": {
+      "babelrc": false,
```
sindresorhus: This is not really needed, and people will just copy-paste this. So I would leave it out.

docs/recipes/babelrc.md
sindresorhus: I think it's fine to just add those like that and we can expand on them later.

docs/recipes/babelrc.md
jamestalmage: OK, If we make the change from #619 we need to revisit.

jamestalmage:
I pushed a commit addressing all the comments. It's not showing up. https://status.github.com/ is showing yellow.

PR #713, opened 2016-04-05
Refactor API

jamestalmage:
This is preliminary work on simplifying the API. The major (breaking) change is a new `test-run` event that passes an emitter for that specific test run. Said emitter is an instance of the new TestData class (might be more appropriately named TestRun). This contains test results that we were storing directly on the API instance. That really doesn't make sense anymore given that the API instance is called multiple times by Watcher. I feel this way makes a little more sense. There is definitely room for further refactoring, but this represents a big diff, and I want to get it reviewed and merged before further refactoring.


api.js
```diff
@@ -38,332 +32,156 @@ function Api(options) {
...
-	if (this.hasExclusive && !stats.hasExclusive) {
-		return;
-	}
+	var emitter = fork(file, options);
 
-	if (!this.hasExclusive && stats.hasExclusive) {
-		this.hasExclusive = true;
-		this.testCount = 0;
-	}
+	testData.listenToTestRun(emitter);
```
novemberborn: `observeFork`?

test/api.js
```diff
@@ -6,6 +6,10 @@ var rimraf = require('rimraf');
 var test = require('tap').test;
 var Api = require('../api');
 var testDoublerPlugin = require('./fixture/babel-plugin-test-doubler');
+// var only = test;
+// if (process.env.ONLY) {
+// 	test = function () {};
+// }
```
novemberborn: ?

test/watcher.js
```diff
@@ -431,7 +447,8 @@ group('chokidar is installed', function (beforeEach, test, group) {
 		t.plan(2);
 
 		files = ['foo-{bar,baz}.js'];
-		api.excludePatterns = ['!*bar*'];
+		avaFiles.defaultExcludePatterns.reset();
```
novemberborn: No need to reset, you can just use `.returns` to change the value.

test/watcher.js
```diff
@@ -102,6 +102,22 @@ group('chokidar is installed', function (beforeEach, test, group) {
 		logger.finish.reset();
 		logger.reset.reset();
 
+		avaFiles.reset();
+		avaFiles.defaultExcludePatterns.reset();
+		avaFiles.defaultIncludePatterns.reset();
```
novemberborn: https://github.com/sindresorhus/ava/pull/713/files#r58517908 notwithstanding, it is indeed good practice to reset stub state in the `beforeEach` :smile:

novemberborn:
> This contains test results that we were storing directly on the API instance. That really doesn't make sense anymore given that the API instance is called multiple times by Watcher. I feel this way makes a little more sense.

Yea that's always stuck me as strange.

Note though that the reporters reference these properties. Their tests use a stubbed `api` so they're still passing.

novemberborn:
> Said emitter is an instance of the new TestData class (might be more appropriately named TestRun).

`RunStatus`?

test/api.js
jamestalmage: Tap does not have exclusive tests. That was my hacky way of getting them

novemberborn: Neat trick!

jamestalmage:
I set reporter.api to the new testData instance with each `test-run` event.

I added a to-do comment that that should be cleaned up. I might just change it to pass `testData` as a param to reporter events. Some reporters could be stateless that way.

novemberborn:
> I might just change it to pass testData as a param to reporter events. Some reporters could be stateless that way.

:+1:

test/watcher.js
```diff
@@ -431,7 +447,8 @@ group('chokidar is installed', function (beforeEach, test, group) {
 		t.plan(2);
 
 		files = ['foo-{bar,baz}.js'];
-		api.excludePatterns = ['!*bar*'];
+		// TODO(@jamestalmage, @novemberborn): There is no way for users to actually set exclude patterns yet.
+		avaFiles.defaultExcludePatterns.returns(['!*bar*']);
```
jamestalmage: @novemberborn - Currently, there is no way for users to set the exclude patterns. Is this a feature that needs to be exposed? 

You access `api.excludePatterns` in watcher, but that's a static value that never changes (I guess an API consumer could, but we haven't published that, so ...).

Should we open an issue about this, or is `avaFiles.defaultExcludePatterns()` all you need.

novemberborn: @jamestalmage you could add exclude patterns to `files`. We should just make sure to still use the default include patterns if `files` only contains negated patterns. That's how the watcher behaves with `source` patterns.

novemberborn: https://github.com/sindresorhus/ava/issues/736

lib/run-status.js
```diff
@@ -0,0 +1,181 @@
...
+var path = require('path');
+var util = require('util');
+var chalk = require('chalk');
+var isObj = require('is-obj');
+var flatten = require('arr-flatten');
+var figures = require('figures');
+var EventEmitter = require('events').EventEmitter;
+var formatter = require('./enhance-assert').formatter();
+
+function TestData(opts) {
```
novemberborn: @jamestalmage did you mean to rename this to `RunStatus`?

jamestalmage: I did indeed

PR #714, opened 2016-04-05
Bump convert-source-map dependency, clean up duplicate require call

novemberborn:
Some follow-up to #662.

### DIFF
--- a/lib/caching-precompiler.js
+++ b/lib/caching-precompiler.js
@@ -1,7 +1,6 @@
 'use strict';
 var path = require('path');
 var fs = require('fs');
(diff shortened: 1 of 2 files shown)


lib/caching-precompiler.js
```diff
@@ -1,7 +1,6 @@
 'use strict';
 var path = require('path');
 var fs = require('fs');
-var generateMapFileComment = require('convert-source-map').generateMapFileComment;
```
vdemedes: I think it would be cleaner to remove `var convertSourceMap` instead of a shortcut.

novemberborn: `convertSourceMap` is used elsewhere in this file.

PR #715, opened 2016-04-05
watcher: fix recursive test file matcher

novemberborn:
The watcher used `dir**/*.js` as the recursive pattern when the initial pattern matched a directory. This should be `dir/**/*.js` instead. Before this fix tests were matched as sources, causing all tests to be rerun.


sindresorhus:
LGTM

PR #675, opened 2016-03-22
Ensure overridden babel config plugins array is immutable 

dcousineau:
Since `babelConfig` in the CachingPrecompiler.prototype._factory method was a JavaScript object coming from an external source, the .push was altering the array in place. objectAssign is shallow so the plugins were being mutated, ensuring that AVA's required plugins were continually appending for each file loaded and parsed.

Using .concat ensures the original plugins array is not mutated and thus doesn't collect plugins like a hoarder.

Does not solve #674 however it was noticed during an attempt to address that issue.

PR #717, opened 2016-04-05
Don't modify configured babel plugins array

novemberborn:
Reimplementation of #675 after recent changes, with an added test. Thanks to @dcousineau for finding the bug.


dcousineau:
@novemberborn thanks! I've been so busy I haven't been able to write tests, and thats a much simpler test than I would have done...

jamestalmage:
:shipit:

sindresorhus:
No worries @dcousineau :)

PR #721, opened 2016-04-05
Support more test filename conventions

kentcdodds:
This PR just updates documentation. I couldn't for the life of me find out where this magic happens. Pointers? Thanks!

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -80,7 +80,7 @@ var cli = meow([
 	'  ava --init foo.js',
 	'',
 	'Default patterns when no arguments:',
-	'test.js test-*.js test/**/*.js __tests__/**/*.js **/*.test.js'
+	'test.js test-*.js test/**/*.js **/__tests__/** **/*.test.js'
 ], {
 	string: [
 		'_',
--- a/cli.js
+++ b/cli.js
@@ -80,7 +80,7 @@ var cli = meow([
 	'  ava --init foo.js',
 	'',
 	'Default patterns when no arguments:',
-	'test.js test-*.js test/**/*.js'
+	'test.js test-*.js test/**/*.js __tests__/**/*.js **/*.test.js'
 ], {
 	string: [
 		'_',
--- a/cli.js
+++ b/cli.js
@@ -80,7 +80,7 @@ var cli = meow([
 	'  ava --init foo.js',
 	'',
 	'Default patterns when no arguments:',
-	'test.js test-*.js test/**/*.js **/__tests__/** **/*.test.js'
+	'test.js test-*.js test/**/*.js **/__tests__ **/*.test.js'
 ], {
 	string: [
 		'_',
--- a/cli.js
+++ b/cli.js
@@ -80,7 +80,7 @@ var cli = meow([
 	'  ava --init foo.js',
 	'',
 	'Default patterns when no arguments:',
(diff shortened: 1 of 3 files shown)


jamestalmage:
Wait for #713 to land. Then check `lib/ava-files`

cli.js
```diff
@@ -80,7 +80,7 @@ var cli = meow([
 	'  ava --init foo.js',
 	'',
 	'Default patterns when no arguments:',
-	'test.js test-*.js test/**/*.js'
+	'test.js test-*.js test/**/*.js __tests__/**/*.js **/*.test.js'
```
sindresorhus: I thought the convention is that `__tests__` can be anywhere? Hence why it's wrapped in underscores.

kentcdodds: Oh, is my glob wrong? Should it be `**/__tests__/**`? That's what I did for my PR to `nyc`. Forgive my ignorance with `glob` :sweat_smile:

jamestalmage: :+1: `**/__tests__/**`

kentcdodds: Updated. :+1: I've subscribed to #713. As soon as that gets merged, I'll give this another look.

jfmengels: Why not `__tests__/**/*.js`?

jfmengels:
`**/*.test.js` ! I use this all the time, would love it if it landed in the default files :heart: 
We'll need to update the [linter](https://github.com/sindresorhus/eslint-plugin-ava/blob/master/docs/rules/no-ignored-test-files.md) too.

kentcdodds:
This has been updated :+1:

lib/ava-files.js
```diff
@@ -17,7 +17,9 @@ function defaultIncludePatterns() {
 	return [
 		'test.js',
 		'test-*.js',
-		'test'
+		'test',
+		'**/__tests__/**',
```
novemberborn: Are all `**/*.js` files inside `__tests__` tests? If so I think you just need to match `**/__tests__` and AVA's built in directory recursion will take over.

kentcdodds:
I imagine that we'll want to add a test for this, but I can't seem to find existing tests for the default inclusion glob...

lib/ava-files.js
kentcdodds: :+1:

kentcdodds: updated

novemberborn:
> I imagine that we'll want to add a test for this, but I can't seem to find existing tests for the default inclusion glob...

Yea. I'm alluding to that in <https://github.com/sindresorhus/ava/issues/736>.

cli.js
```diff
@@ -80,7 +80,7 @@ var cli = meow([
 	'  ava --init foo.js',
 	'',
 	'Default patterns when no arguments:',
-	'test.js test-*.js test/**/*.js'
+	'test.js test-*.js test/**/*.js **/__tests__ **/*.test.js'
```
novemberborn: Documentation wise it should be `**/__tests__/**/*.js`, just like the `test` directory in `ava-files` is explained as `test/**/*.js`.

kentcdodds: Good point. Updated :+1:

novemberborn:
LGTM!

jfmengels:
Woohoo! This makes my day, thanks @kentcdodds!

PR #726, opened 2016-04-06
Docs: Link to French translation for babelrc recipe

forresst:
Link to french translation for new recipe : babelrc


sindresorhus:
Oh wow. You're fast! Excellent work as always @forresst :)

PR #729, opened 2016-04-06
Document issue labels

novemberborn:
Document our [issue labels](https://github.com/sindresorhus/ava/labels) in the contributing and maintaining guides.

### DIFF
--- a/maintaining.md
+++ b/maintaining.md
@@ -52,7 +52,7 @@ Please note the priority labels:
 * `low priority`: Issues we'd like to see progress on
 * `future`: Issues we're not planning on getting to anytime soon. These are the long term suggestions for which we're unlikely to accept PRs
 
-Use the `assigned` label when somebody is working on the issue, but we can't actually assign the issue to them.
+Use the `assigned` label when somebody is working on the issue so we can avoid duplicated effort.
 
 ## Profiling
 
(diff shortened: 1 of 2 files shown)


novemberborn:
IMHO we could remove the `duplicate`, `invalid` and `wontfix` labels. We've basically never used them.

maintaining.md
```diff
@@ -31,6 +31,29 @@ Read and adhere to the [Code of Conduct](code-of-conduct.md).
...
+* `performance`: Use this for performance related issues
+* `question`: Use this for issues that are in a discussion phase
+
+Please note the priority labels:
+
+* `priority`: Issues to tackle as soon as possible
+* `low priority`: Issues we'd like to see progress on
+* `future`: Issues we're not planning on getting to anytime soon. These are the long term suggestions for which we're unlikely to accept PRs
+
+Use the `assigned` label when somebody is working on the issue, but we can't actually assign the issue to them.
```
jamestalmage: I would suggest we set the `assigned` label even when a collaborator is assigned. That way they don't need to look two places.

sindresorhus:
> IMHO we could remove the duplicate, invalid and wontfix labels. We've basically never used them.

Done.

> I would suggest we set the assigned label even when a collaborator is assigned. That way they don't need to look two places.

:+1:

PR #730, opened 2016-04-06
Improve source pattern handling

### LINKED ISSUE #614: Improve source pattern handling
Watch mode lets you configure which source files should be watched. Certain directories are ignored by default, using <https://github.com/novemberborn/ignore-by-default>. Currently if you use a negation pattern in `--source` it stops ignoring those default directories. This is annoying as you'll have to repeat them manually.

It would be good to stop one of those default directories from being ignored. Then we could let you specify a negation pattern without disabling *all* default ignore patterns.

You can find the current logic at <https://github.com/sindresorhus/ava/blob/1868204c1901f45b4f66a520ef6486fdd71fe1d2/lib/watcher.js#L273:L276>.

novemberborn:
Bugfix where `..file.js` sources were treated as being outside the current working directory. But the main change of this PR is that negated source patterns are now handled much better :boom: 

Fixes #614. Allow negated source patterns to be specified without unsetting the default negation patterns.

Allow source patterns to override the default negation patterns if they start with one of the ignored directories.

### DIFF
--- a/docs/recipes/watch-mode.md
+++ b/docs/recipes/watch-mode.md
@@ -61,7 +61,9 @@ In AVA there's a distinction between *source files* and *test files*. As you can
 
 By default AVA watches for changes to the test files, `package.json`, and any other `.js` files. It'll ignore files in [certain directories](https://github.com/novemberborn/ignore-by-default/blob/master/index.js) as provided by the [`ignore-by-default`] package.
 
-You can configure patterns for the source files using the [`--source` CLI flag] or in the `ava` section of your `package.json` file. Note that if you specify a negative pattern the directories from [`ignore-by-default`] will no longer be ignored, so you may want to repeat these in your config.
+You can configure patterns for the source files using the [`--source` CLI flag] or in the `ava` section of your `package.json` file.
+
+You can specify patterns to match files in the folders that would otherwise be ignored, e.g. use `node_modules/some-dependency/*.js` to specify all `.js` files in `node_modules/some-dependency` as a source, even though normally all files in `node_modules` are ignored. Note that you need to specify an exact directory; `{bower_components,node_modules}/**/*.js` won't work.
 
 If your tests write to disk they may trigger the watcher to rerun your tests. If this occurs you will need to use the `--source` flag.
 
--- a/docs/recipes/watch-mode.md
+++ b/docs/recipes/watch-mode.md
@@ -34,7 +34,7 @@ You could also set up a special script:
 {
   "scripts": {
     "test": "ava",
-    "test:watch": "ava --watch"
+    "watch:test": "ava --watch"
   }
 }
 ```
@@ -42,7 +42,17 @@ You could also set up a special script:
 And then use:
 
 ```console
-$ npm run test:watch
+$ npm run watch:test
+```
+
+Finally you could configure AVA to *always* run in watch mode by setting the `watch` key in the [`ava` section of your `package.json`]:
+
+```json
+{
+  "ava": {
+    "watch": true
+  }
+}
 ```
 
 Please note that the TAP reporter is unavailable when using watch mode.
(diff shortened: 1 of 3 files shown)


test/watcher.js
```diff
@@ -858,17 +859,46 @@ group('chokidar is installed', function (beforeEach, test, group) {
 			});
 		});
 
-		test('ignores dependencies outside of the current working directory', function (t) {
+		test('allows default exclusion patterns to be overriden', function (t) {
```
jamestalmage: Once we land the API refactor, I would like to move all the logic of building include/exclude patterns from the config into `lib/ava-files`, and rewrite many of these tests so they are just:

```js
t.true(avaFiles.shouldWatch('some/path'));
```

novemberborn: Yea would be good to centralize the pattern matching logic.

jamestalmage:
Trying to understand what we are fixing here:

Seems like a few things:

- Proper handling of relative paths that start with `../`
- Source patterns can specify `node_modules/some_dir` even though ignore patterns still have `node_modules/**`
- Using a negation pattern in sources (`!lib/somedir/**`), will not override default ignores

novemberborn:
@jamestalmage yup, perfect score!

jamestalmage:
It just seems we have [a whole lot of code manipulating glob patterns](https://github.com/sindresorhus/ava/blob/4ff70bd5b366838e99564868570e2da440ed1392/lib/watcher.js#L278-L411).

Ignoring for a moment our intent to [refactor much of that to `ava-files`](https://github.com/sindresorhus/ava/pull/730#discussion_r58733436), is there a good reason we have made it so complicated?

The code looks good, just wondering if we need to rethink some of the design decisions that led us here.

novemberborn:
@jamestalmage different use cases I think. Chokidar needs to watch tests and sources, and exclude particular directories. But we also want it to not exclude certain directories. Then we need to see if a changed file was a source or a test which again has different logic… Let me know where it isn't clear and I can add some comments.

novemberborn:
Any further feedback aside from refactoring concerns?

docs/recipes/watch-mode.md
```diff
@@ -59,7 +59,9 @@ In AVA there's a distinction between *source files* and *test files*. As you can
 
 By default AVA watches for changes to the test files, `package.json`, and any other `.js` files. It'll ignore files in [certain directories](https://github.com/novemberborn/ignore-by-default/blob/master/index.js) as provided by the [`ignore-by-default`] package.
 
-You can configure patterns for the source files using the [`--source` CLI flag] or in the `ava` section of your `package.json` file. Note that if you specify a negative pattern the directories from [`ignore-by-default`] will no longer be ignored, so you may want to repeat these in your config.
+You can configure patterns for the source files using the [`--source` CLI flag] or in the `ava` section of your `package.json` file.
+
+You can specify patterns to match files in the folders that would otherwise be ignored, e.g. use `node_modules/some-dependency/*.js` to specify all `.js` files in `node_modules/some-dependency` as a source, even though normally all files in `node_modules` are ignored. Note that you need to specify an exact directory; `{.bower_components,node_modules}/**/*.js` won't work.
```
sindresorhus: `.bower_components` => `bower_components`

docs/recipes/watch-mode.md
```diff
@@ -59,7 +59,9 @@ In AVA there's a distinction between *source files* and *test files*. As you can
 
 By default AVA watches for changes to the test files, `package.json`, and any other `.js` files. It'll ignore files in [certain directories](https://github.com/novemberborn/ignore-by-default/blob/master/index.js) as provided by the [`ignore-by-default`] package.
 
-You can configure patterns for the source files using the [`--source` CLI flag] or in the `ava` section of your `package.json` file. Note that if you specify a negative pattern the directories from [`ignore-by-default`] will no longer be ignored, so you may want to repeat these in your config.
+You can configure patterns for the source files using the [`--source` CLI flag] or in the `ava` section of your `package.json` file.
```
sindresorhus: Maybe we should put `in the `ava` section of your `package.json` file` and mention it's the recommended way. Having the config in package.json will be important for future editor plugins and other integrations as they won't be able to read the config from the npm run script.

sindresorhus:
LGTM when the nitpick inline feedback is resolved.

docs/recipes/watch-mode.md
novemberborn: Good catch!

novemberborn:
Addressed feedback. Made some other changes to the docs, see the last commit.

sindresorhus:
@jamestalmage :shipit:?

jamestalmage:
:tada: Looks good @novemberborn

PR #731, opened 2016-04-06
Replace "callback" with "implementation" when talking about tests

sotojuan:
This was suggested in #705. Let me know if I missed anything or if there's a better word.


novemberborn:
Awesome!

PR #732, opened 2016-04-06
Disable TAP reporter in watch mode

### LINKED ISSUE #672: Disable --tap in --watch mode
You can't really do `--watch --tap`, the output stream never ends, so you only get a report for the first run, then just raw TAP output:

```
$ ava --watch --tap | mocha-tap-reporter spec


  ✓ test › create must be a function
  ✓ test › maxSize must be number
  ✓ test › it will create up to maxSize handlers as needed
  ✓ test › a new handler will not be created if one is already returned to the stack
  ✓ test › pending requests will be executed as handlers are returned
  ✓ test › it will pass through the args
ok 7 - test › create must be a function
ok 8 - test › maxSize must be number
ok 9 - test › it will create up to maxSize handlers as needed
ok 10 - test › a new handler will not be created if one is already returned to the stack
ok 11 - test › pending requests will be executed as handlers are returned
ok 12 - test › it will pass through the args
1..6
ok 13 - test › create must be a function
ok 14 - test › maxSize must be number
ok 15 - test › it will create up to maxSize handlers as needed
ok 16 - test › a new handler will not be created if one is already returned to the stack
ok 17 - test › pending requests will be executed as handlers are returned
ok 18 - test › it will pass through the args
1..6
```

---

It's probably not possible to fix this and continue to just use the bash pipe operator to pass our output to the reporter. Since our stream never ends, I don't think the reporter will ever print it's summary, nor will it know to reset itself (I could be wrong - it seems the reporters could figure this out from the `plan` line, but that is likely pretty brittle).

One option would be to incorporate `tap-mocha-reporter` as an (optional?) dependency and allow a `--tap-reporter` flag. We could then handle creating separate streams for each test run, piping them to a new reporter instance each time, and correctly ending the stream.

The final option would be "Sorry, no TAP output in watch mode". Personally, I am totally fine with this option, since I prefer our own reporters. I think the real value of TAP is realized on CI servers (where you could use it to create automated reports, etc).

---

Also note that the tap reporter doesn't reset the test id on subsequent runs (that is a bug unless we choose the "sorry" option).

novemberborn:
Fixes #672.


cli.js
```diff
@@ -129,7 +129,7 @@ var api = new Api({
 
 var reporter;
 
-if (cli.flags.tap) {
+if (cli.flags.tap && !cli.flags.watch) {
```
jamestalmage: If they do `--tap --watch` it should throw. ("watch is not allowed with tap reporter")

If they do `--watch` and their pkgConf specifies `tap:true`, then it should fall through to picking `verbose` or `mini` reporter.

novemberborn:
> If they do --tap --watch it should throw. ("watch is not allowed with tap reporter")

Oops overlooked that requirement. Done now.

Also realized the error message when watch mode is used without chokidar being installed was written to stdout. Changed to stderr now.

cli.js
```diff
@@ -116,6 +116,11 @@ if (cli.flags.init) {
 	return;
 }
 
+if (cli.flags.watch && cli.flags.tap && !conf.tap) {
```
jamestalmage: This is better. However, if they do `--tap --watch`, AND have `tap:true` in their config they won't get this error message.

The only thing I can think of is to use `hasFlag('--tap')` instead of `cli.flags.tap`.

Also, they could specify `{watch: true, tap: true}` in their config and not get this warning.

So maybe:

```js
if ((hasFlag('--tap') && hasFlag('--watch')) || (conf.watch && conf.tap)) {
  // ...
}
```

... Or I'm just overthinking it.

novemberborn: Where would `hasFlag()` come from? I didn't look too closely at `meow` but I assumed I couldn't really see whether `cli.flags.tap` came from `--tap` or `conf`. Probably not a good idea to reparse `argv` either.



> ... Or I'm just overthinking it.

`--tap` would be superfluous if it's already in the `package.json`, and there won't be an error in the `package.json` anyway, so I don't think this is a big deal.

jamestalmage: https://github.com/sindresorhus/has-flag

novemberborn: Ah. What about

> --tap would be superfluous if it's already in the package.json, and there won't be an error in the package.json anyway, so I don't think this is a big deal.

?

jamestalmage: My thought was that a user might try both flags trying to troubleshoot why tap wasn't working in watch mode. I thought an explicit error would be better than hoping they had read the docs

novemberborn: Sure, I can do that.

jamestalmage: LGTM once that's handled.

novemberborn:
Updated.

Please hold off on merging until a new release of [`has-flag`](https://github.com/sindresorhus/has-flag) is out.

sindresorhus:
> Please hold off on merging until a new release of has-flag is out.

It's out. Bump it to 2.0.0.

novemberborn:
> It's out. Bump it to 2.0.0.

Done.

sindresorhus:
Super-duper 🎉

PR #733, opened 2016-04-06
Watch mode always runs all exclusive tests

### LINKED ISSUE #635: Improve reporter output when watch mode runs test files without exclusive tests
`.only` is sticky in watch mode. If the watcher detects it in one file, and then needs to run tests from another, it only runs exclusive tests from that second file. If the second file does not contain any exclusive tests *no tests are run*. With the mini reporter the output in this case is simply "0 tests passed". This may be confusing to users.

Instead the reporters (mini, verbose and tap) should detect this scenario and output something like "no exclusive tests found".

novemberborn:
Fixes #635.

### DIFF
--- a/docs/recipes/watch-mode.md
+++ b/docs/recipes/watch-mode.md
@@ -69,6 +69,10 @@ AVA tracks which source files your test files depend on. If you change such a de
 
 Dependency tracking works for required modules. Custom extensions and transpilers are supported, provided you loaded them using the [`--require` CLI flag] and not from inside your test file. Files accessed using the `fs` module are not tracked.
 
+## Watch mode and the `.only` modifier
+
+The `.only` modifier disables watch mode's dependency tracking algorithm. When a change is made, all `.only` tests will be rerun, regardless of whether the test depends on the changed file.
+
 ## Manually rerunning all tests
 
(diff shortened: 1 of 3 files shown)


docs/recipes/watch-mode.md
```diff
@@ -69,6 +69,10 @@ AVA tracks which source files your test files depend on. If you change such a de
 
 Dependency tracking works for required modules. Custom extensions and transpilers are supported, provided you loaded them using the [`--require` CLI flag] and not from inside your test file. Files accessed using the `fs` module are not tracked.
 
+## Watch mode and the `.only` modifier
+
+You can use the [`.only` modifier] with watch mode. However now when a change is detected, *all* `.only` tests are rerun, even if they're from a test file that is unaffected by the change.
```
jamestalmage: The `.only` modifier disables watch modes test dependency tracking algorithm. When a change is made, all `.only` tests will be rerun, regardless of whether the test depends on the changed file.

jamestalmage:
One minor nit on the docs. 

Otherwise LGTM

docs/recipes/watch-mode.md
novemberborn: Yes much better, thanks.

novemberborn:
Updated.

PR #734, opened 2016-04-06
Put failing TAP test title in the second line

### LINKED ISSUE #679: TAP report prints test titles above not ok message
## Description

The TAP report prints the failing assertion rather than the title of the test after the not ok message. If I have a test file `test.js` that reads:

```
import { test } from 'ava';

test('one should equal two', t => {
  t.is(1, 2);
});
```

and I run `ava --tap`, the resulting TAP report looks like this:

```
TAP version 13
# one should equal two
not ok 1 - 1 === 2
  ---
    operator: ===
    expected: 2
    actual: 1
    at: _callee$ (test.js:4:4)
  ...

1..1
# tests 1
# pass 0
# fail 1
```

This is an issue because I am using a [Jenkins TAP Plugin](https://wiki.jenkins-ci.org/display/JENKINS/TAP+Plugin) that reports the titles of the tests incorrectly, as it interprets `1 - 1 === 2` as the name of the test. I would expect the TAP report to look like this:

```
TAP version 13
# one should equal two
not ok 1 - one should equal two
  ---
    operator: ===
    expected: 2
    actual: 1
    at: _callee$ (test.js:4:4)
  ...

1..1
# tests 1
# pass 0
# fail 1
```

If this is an issue with a dependency of ava and not with ava itself, please direct me to that project instead and I'll be happy to file an issue there.

Thanks in advance!

## Environment
Node.js v5.9.0
darwin 14.5.0

sotojuan:
Fixes #679.

### DIFF
--- a/lib/reporters/tap.js
+++ b/lib/reporters/tap.js
@@ -39,7 +39,7 @@ TapReporter.prototype.test = function (test) {
 	if (test.error) {
 		output = [
 			'# ' + test.title,
-			format('not ok %d - %s', ++this.i, test.error.message),
+			format('not ok %d - %s', ++this.i, test.title),
 			'  ---',
 			'    operator: ' + test.error.operator,
 			'    expected: ' + test.error.expected,
--- a/test/reporters/tap.js
+++ b/test/reporters/tap.js
@@ -41,7 +41,7 @@ test('failing test', function (t) {
 
 	var expectedOutput = [
 		'# failing',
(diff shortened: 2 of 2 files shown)


sotojuan:
Might have to change some stuff because the output can be redundant when used with other TAP stuff for example nyan:

screen shot 2016-04-06 at 2 23 17 pm

jamestalmage:
LGTM

PR #737, opened 2016-04-07
Improve watch logging

### LINKED ISSUE #670: Output in watch mode should be less verbose
## Description

I like the watch mode so far, but I think its verbosity could be reduced. Using the default reporter, the output is this:

screen shot 2016-03-22 at 10 51 11

A good solution to this would be to print less empty lines; but my ideal solution would be that these messages should not show at all. Instead, it would be really great if it showed simply a wait cursor, and only printed a message when a test fails.

## Environment

```
Node.js v5.4.1
darwin 15.0.0
```

novemberborn:
Fixes #670.

* When the mini and verbose reporters finish, add a timestamp after the first passed/failed/etc status line
* Ensure the watcher doesn't break the mini reporter's spinner on the first run
* Ensure the watcher restarts the mini reporter's spinner on subsequent runs
* Clear the mini reporter on subsequent runs, unless failures occurred or `r\n` was entered on stdin (because that inserts a line which can't be cleared)

Also includes some test improvements.

### DIFF
--- a/lib/reporters/verbose.js
+++ b/lib/reporters/verbose.js
@@ -1,6 +1,7 @@
 'use strict';
 var prettyMs = require('pretty-ms');
 var figures = require('figures');
+var chalk = require('chalk');
 var plur = require('plur');
 var colors = require('../colors');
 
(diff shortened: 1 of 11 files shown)


test/logger.js
```diff
@@ -45,6 +45,23 @@ test('only write if reset is supported by reporter', function (t) {
...
+	tapReporter.clear = undefined;
+	logger.clear();
+	t.end();
+});
+
+test('only write if clear is supported by reporter', function (t) {
+	var tapReporter = tap();
+	var logger = new Logger(tapReporter);
+	tapReporter.clear = undefined;
+	logger.write = t.fail;
```
jamestalmage: Are we certain `t.fail` has it's `this` value bound in `node-tap`?

jamestalmage:
Needs a rebase, but LGTM

test/logger.js
novemberborn: [Looks](https://github.com/tapjs/node-tap/blob/d5e11d0f660ac0a38b486adfb505b28a02e51fd3/lib/test.js#L118) [like it](https://github.com/tapjs/node-tap/blob/d5e11d0f660ac0a38b486adfb505b28a02e51fd3/lib/test.js#L1363)

novemberborn:
Rebased.

sindresorhus:
I like it a lot!

One thing that might make it clearer is to add one more newline between persisted results.
This looks a bit cramped:

screen shot 2016-04-11 at 22 52 17

novemberborn:
> One thing that might make it clearer is to add one more newline between persisted results.
> This looks a bit cramped:

@sindresorhus I'm not sure newlines will help. Just pushed a commit that inserts a horizontal sectioning line. No tests yet. Could you have play with that?

sindresorhus:
@novemberborn Perfect. Much clearer separation now.

lib/reporters/verbose.js
```diff
@@ -3,6 +3,7 @@ var prettyMs = require('pretty-ms');
 var figures = require('figures');
 var chalk = require('chalk');
 var plur = require('plur');
+var repeatString = require('repeat-string');
```
sindresorhus: Can you use https://github.com/sindresorhus/repeating? We already depend on modules using it, so it would mean no extra dependency.

novemberborn: Ha! I knew you'd written one! It didn't show up when searching for [string repeat](https://www.npmjs.com/search?q=string+repeat) :smile:

sindresorhus: The npm search sucks... Try http://node-modules.com.

sindresorhus:
Tests are failing.

novemberborn:
> Tests are failing.

Yea, probably due to the wip commit. Will patch that up tomorrow.

novemberborn:
Updated!

I changed the `finish()` output as well to be more consistent between the mini and verbose reporters. Each error/rejection/exception is now preceeded by two empty lines. Trailing whitespace is removed from stack traces. There's always an empty line at the end of the finish output.

This means that if the horizontal sectioning line is printed there will always be a single empty line before it, and a single empty line after it. Oh so very tidy! :stuck_out_tongue_winking_eye:

sindresorhus:
@novemberborn :shipit:

jamestalmage:
:+1: The output looks really slick now!

supercrabtree:
Thanks for all your hard work on this, what a pleasant surprise to see in the console this morning. Looks great 👍

PR #740, opened 2016-04-08
increase power-assert's maxDepth option to 3. Fixes #681

### LINKED ISSUE #681: Change default depth of power-assert rendering to 3
**Moderator UPDATE:** We've decided to change the default depth to 3. PR Welcome!

--

I'm using ava since some time, it works like a charm! Thanks.

One issue I regularly encounter is that I'm having unit tests which compare arrays containing objects. When a test fails, the nested contents of these arrays is not displayed, making it very hard to compare (I have to manually insert console.logs to output actual and expected arrays). The `power-assert` library has options like `maxDepth` (1 by default), which I would like to configure with a higher value in order to see the nested contents of my arrays.

Is there a way to pass configuration to `power-assert` via `ava`? If not, it would be great if support for this could be implemented.

iamstarkov:
just a pr to fix #681 to get more descriptive logs (found out from power-assert-js/power-assert#41)

### DIFF
--- a/lib/enhance-assert.js
+++ b/lib/enhance-assert.js
@@ -50,6 +50,7 @@ function formatter() {
 	var powerAssertRenderers = require('power-assert-renderers');
 
 	return powerAssertFormatter({
+		maxDepth: 3,
 		renderers: [
 			powerAssertRenderers.AssertionRenderer,
 			powerAssertRenderers.SuccinctRenderer


novemberborn:
LGTM!

Had a quick look and I don't think we necessarily have any tests for `enhance-assert` so this should be fine as-is.

Issue #739, opened 2016-04-07
Throw assertion does not catch class constructor errors


ejmurra:
Use case: An abstract class which implements member checking in the constructor.

Example:
```javascript

class FeedAggregator {
  constructor({feedName, feedType}) {
    this.feedName = feedName;
    this.feedType = feedType;

    // Check for methods that must be implemented
    if (this.getItems === undefined) {
      throw new TypeError("Class must define a getItems method");
    }
  }
}

class BadAggregator extends FeedAggregator {
  constructor({feedName, feedType}) {
    super({feedName, feedType});
  }
}

test("Expect FeedAggregator to throw error when subclass has no getItems method", t => {
  t.throws(new BadAggregator({feedName: 'test', feedType: 'test'}), TypeError("Class must define a getItems method"))
});
```

Expected output:
```
1 passed
```

Actual output:
```
1 failed

  1. Expect FeedAggregator to throw error when subclass has no getItems method
  failed with "Class must define a getItems method"
      BadAggregator.FeedAggregator (FeedAggregator.js:11:13)
    new BadAggregator (FeedAggregator.js:16:7)
    Test.fn (FeedAggregator.js:23:12)
```

jamestalmage:
You need to wrap any statement in a function. It's not possible for us to catch the Error otherwise.

For one liners like that, arrow functions are where it's at:

```js
test("Expect FeedAggregator to throw error when subclass has no getItems method", t => {
  t.throws(
    () => new BadAggregator({feedName: 'test', feedType: 'test'}), 
    TypeError("Class must define a getItems method")
  );
});
```

ejmurra:
You're absolutely correct. Sorry and thank you!

Issue #404, opened 2016-01-03
Document common pitfalls


jamestalmage:
There are some points that continue to confuse people (especially as it relates to our async behavior).

One in particular is using closure scope variables to try and pass data between `beforeEach` and individual tests (see #402). 

I think we need to create some documentation that addresses each of these types misconceptions so we can quickly link to them. They can be in the `readme` if they are concise, but maybe separate documents if they require lengthy explanations.

If anyone can think of other common downfalls / misconceptions / misuses - go ahead and add them to this discussion.

ariporad:
I think this is a really great idea. Also, I'd like to appreciate the relevant issue number.

sindresorhus:
Good idea!

sindresorhus:
Pitfalls:
- Using global variables and forgetting that tests are run concurrently.
- `console.log` not logging where the test result is printed (I hope we can fix https://github.com/sindresorhus/ava/issues/392, but we should document until we do)

ariporad:
Actually, What if we made this a wiki page and linked to it from the README? That way others could contribute?

sindresorhus:
> What if we made this a wiki page and linked to it from the README?

No. I want to have quality control on the docs. I don't want random strangers to be able to just freely modify the docs. Also: https://plus.google.com/+sindresorhus/posts/QSS2du26Mg4

> That way others could contribute?

They already can with a pull request.

novemberborn:
Another one: if you run AVA in Docker as part of your CI, you need to forward the appropriate environment variables (see <https://github.com/watson/is-ci/blob/master/index.js>), or configure the verbose reporter. See #751.

sotojuan:
Should we make a `common_pitfalls.md` in the documentation directory with these?

jamestalmage:
yep

sindresorhus:
👍 

-

`common_pitfalls.md` => `common-pitfalls.md`

sindresorhus:
Another pitfall I see often is trying to do an async operation in a normal test and not understand why it's not finishing.

Forgetting to return the promise:

```js
test(t => {
  fetch().then(data => {
    t.is(data, 'foo');
  });
});
```

Or should have used a callback test:


```js
test(t => {
  fetch((err, data) => {
    t.is(data, 'bar');
  });
});
```

novemberborn:
* [ ] [Cached files being used after configuration changes that AVA did not detect](https://github.com/avajs/ava/issues/872)

[...]

PR #742, opened 2016-04-09
detect improper use of t.throws

jamestalmage:
Protects against a common misuse of t.throws (Like that seen in #739).

This required the creation of a custom babel plugin.

https://github.com/jamestalmage/babel-plugin-ava-throws-helper

Sample output:

```
Improper usage of t.throws detected at /Users/jamestalmage/dev/ava/test/fixture/improper-t-throws.js (4:10).

You should wrap the following expression in a function:

  throwSync()

Like this:

  function() {
    throwSync()
  }

See https://github.com/sindresorhus/ava#throwsfunctionpromise-error-message for more details.
```

The linked documentation isn't that helpful. We should probably make a short recipe about this. (maybe a `common-mistakes` recipe? This seems like too small a topic for a whole recipe). We can update the URL when that recipe is written.

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -1,5 +1,6 @@
 'use strict';
 var path = require('path');
+var fs = require('fs');
 var inspect = require('util').inspect;
 var chalk = require('chalk');
 var isGeneratorFn = require('is-generator-fn');
@@ -11,6 +12,7 @@ var observableToPromise = require('observable-to-promise');
 var isPromise = require('is-promise');
 var isObservable = require('is-observable');
 var plur = require('plur');
+var codeFrame = require('babel-code-frame');
 var assert = require('./assert');
 var enhanceAssert = require('./enhance-assert');
 var globals = require('./globals');
@@ -72,12 +74,19 @@ Test.prototype._assert = function (promise) {
 Test.prototype._setAssertError = function (err) {
 	if (err && err._avaThrowsHelperData) {
 		var data = err._avaThrowsHelperData;
+		var frame = '';
+		try {
+			var rawLines = fs.readFileSync(data.filename, 'utf8');
+			frame = codeFrame(rawLines, data.line, data.column, {highlightCode: true});
+		} catch (e) {
+		}
 		console.error(
 			[
-				'Improper usage of t.throws detected at ' + chalk.bold.yellow('%s (%d:%d)') + '.',
-				'You should wrap the following expression in a function:',
+				'Improper usage of t.throws detected at ' + chalk.bold.yellow('%s (%d:%d)') + ':',
+				frame,
+				'The following expression:',
 				chalk.cyan('  %s'),
-				'Like this:',
+				'Should be wrapped in a function like this:',
 				chalk.cyan('  function() {\n    %s\n  }'),
 				'Visit the following URL for more details:',
 				'  ' + chalk.blue.underline('https://github.com/sindresorhus/ava#throwsfunctionpromise-error-message')
--- a/lib/test.js
+++ b/lib/test.js
@@ -1,8 +1,5 @@
 'use strict';
-var path = require('path');
-var fs = require('fs');
 var inspect = require('util').inspect;
-var chalk = require('chalk');
 var isGeneratorFn = require('is-generator-fn');
 var maxTimeout = require('max-timeout');
 var Promise = require('bluebird');
@@ -15,6 +12,7 @@ var plur = require('plur');
 var assert = require('./assert');
 var enhanceAssert = require('./enhance-assert');
 var globals = require('./globals');
+var throwsHelper = require('./throws-helper');
 
 function Test(title, fn, contextRef, report) {
 	if (!(this instanceof Test)) {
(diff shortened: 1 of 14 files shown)


lib/test.js
```diff
@@ -68,6 +68,24 @@ Test.prototype._assert = function (promise) {
...
 Test.prototype._setAssertError = function (err) {
+	var data = err._avaThrowsHelperData;
+	if (data) {
+		console.error(
+			[
+				'Improper usage of t.throws detected at %s (%d:%d).',
+				'You should wrap the following expression in a function:',
+				'  %s',
+				'Like this:',
+				'  function() {\n    %s\n  }',
```
novemberborn: Promote arrow functions?

lib/test.js
```diff
@@ -68,6 +68,24 @@ Test.prototype._assert = function (promise) {
 };
 
 Test.prototype._setAssertError = function (err) {
+	var data = err._avaThrowsHelperData;
```
novemberborn: Probably need to guard against `err` being `null` or `undefined` here?

test/caching-precompiler.js
```diff
@@ -145,7 +146,7 @@ test('uses babelConfig for babel options when babelConfig is an object', functio
 	t.true('inputSourceMap' in options);
 	t.false(options.babelrc);
 	t.same(options.presets, ['stage-2', 'es2015']);
-	t.same(options.plugins, [customPlugin, powerAssert, rewrite, transformRuntime]);
+	t.same(options.plugins, [customPlugin, powerAssert, throwsHelper, rewrite, transformRuntime]);
```
novemberborn: Without checking the `power-assert` transformations, are there cases the Babel plugin can no longer catch?

novemberborn:
> The linked documentation isn't that helpful. We should probably make a short recipe about this

:+1: 



> (maybe a common-mistakes recipe?

#404?

test/caching-precompiler.js
jamestalmage: Shouldn't be. The `babel-throws-helper` wrapper is totally transparent to the code it wraps.

lib/test.js
jamestalmage: I considered that.

But if the user is green enough that that they don't fully understand how `throw` works, I was concerned about piling on another abstraction they need to understand.

test/caching-precompiler.js
novemberborn: Sure, but `power-assert` might rewrite the `t.throws()` argument expression such that it is computed outside of `t.throws()`. Again, I don't know if it does, but if it did then you wouldn't be able to catch it.

lib/test.js
novemberborn: Yea fair enough.

vdemedes:
> /Users/jamestalmage/dev/ava/test/fixture/improper-t-throws.js (4:10)

Perhaps displaying relative path would be better and would align with our "clean stack traces" feature.

jamestalmage:
Updated with relative file paths and colors:

screenshot 2016-04-10 16 44 14

test/caching-precompiler.js
jamestalmage: Oh - Understood. Still not a problem though. `power-assert` wraps inline, just like `ava-throws-helper`.

```js
t.true(_rec._expr(_rec._capt(_rec._capt(a, 'arguments/0/left') === 'bar', 'arguments/0'), {
	content: 't.true(a === \'bar\')',
	filepath: 'test/fixture/power-assert.js',
	line: 6
}));
```

Also, we don't enhance `t.throws` with `power-assert`.

vdemedes:
I know it is an overkill, but would be incredibly cool to have syntax highlight in errors (not just this one).

test/caching-precompiler.js
twada: > we don't enhance t.throws with power-assert.

Right. `t.throws` is [listed as NON_ENHANCED_PATTERNS here](https://github.com/sindresorhus/ava/blob/84c05fe98ee5887e49262d48f1134cdf19118a7f/lib/enhance-assert.js#L23-L29). power-assert doesn't touch its arguments.

jamestalmage:
> I know it is an overkill, but would be incredibly cool to have syntax highlight in errors (not just this one).

Like this?:

screenshot 2016-04-11 01 49 24

jamestalmage:
I think maybe the middle section:

```
The following expression:
  t.throws()
```

Is kinda useless with the codeframe right above.

jamestalmage:
screenshot 2016-04-11 02 03 30

vdemedes:
Super awesome!

sindresorhus:
Marvelous. :shipit:

novemberborn:
Cool!

Should update the default plugin sections in [the readme](https://github.com/sindresorhus/ava/tree/195390ec1db90cf7d394407c846a69fbaa08806d#es2015-support) and [babelrc recipe](https://github.com/sindresorhus/ava/blob/195390ec1db90cf7d394407c846a69fbaa08806d/docs/recipes/babelrc.md#notes).

jamestalmage:
> Should update the default plugin sections in the readme and babelrc recipe.

I went with just the `babelrc` recipe. I feel like most user's don't care, so I don't want to pollute the readme with every side-effect free plugin we install. Really, we could just update the readme to say. "AVA always includes a few internal plugins regardless of configuration, but they should not impact the behavior of hour code", and then link to the Notes section of the `babelrc` recipe.

sindresorhus:
> "AVA always includes a few internal plugins regardless of configuration, but they should not impact the behavior of hour code", and then link to the Notes section of the babelrc recipe.

:+1:

jfmengels:
:tada:

sindresorhus:
🍾✨

Issue #447, opened 2016-01-17
Recipe for testing ES6 modules


vdemedes:

jamestalmage:
It's kind of covered by the code coverage recipe, but it might be good to have a simpler recipe that didn't include that.

Some stuff from [this comment](https://github.com/sindresorhus/ava/pull/444#issuecomment-172385855) would apply here as well.

jamestalmage:
See https://github.com/jamestalmage/__ava-0.8-with-babel-6 for a working example

Repository name is a misnomer now - it uses ava@0.10.0.

boutros:
It dould be great to have a recipe of how to use ava with rollup.

sotojuan:
@boutros What do you mean? Rollup is a module bundler—surely you aren't bundling AVA for production right?

boutros:
> Rollup is a module bundler—surely you aren't bundling AVA for production right?

No, not for production of course. Anyway I solved it with rollup, so my remark can be ignored. If anyone is interested here is the relevant commit: https://github.com/boutros/koblersken/commit/c00e3a098c035c7a585ab72d1dde91fd2ac47e91

It's easy to get lost in the javascript build jungle :)

novemberborn:
I think we're kinda already covering this. Of course if you use new language features that are supported by your targeted Node.js version you don't need to do anything special. If you need to transpile your source code we've covered that in the [Transpiling imported modules](https://github.com/avajs/ava/tree/4baa17038451f07ebe5d81a0ed42590e69913394#transpiling-imported-modules) section.

I'm closing this issue, but if you have ideas on how to improve the documentation please open a new issue or PR!

backspaces:
TL;DR: If you're using Rollup and no babel in your workflow, building a cjs bundle and requiring it in your tests works fine.

Details:
I've just started using ava and ran into the es6 modules problem. But because my repo uses Rollup to a cjs bundle (to be node friendly) I just require that in the test file. Ava is OK with that.

Here's an article on the "modern workflow" idea where Rollup is used to convert a bleeding edge project into any format you'd like: 
https://medium.com/@backspaces/es6-modules-part-1-migration-strategy-a48de0b7f112

Issue #577, opened 2016-02-29
Transpile source files in the main process


novemberborn:
To improve startup time (#78) we could provide a framework where source files are optimistically transpiled and cached in the main process. No require hooks would be required in the test files avoiding the Babel startup penalty. Thanks @tomazzaman for the [suggestion](https://github.com/sindresorhus/ava/issues/469).
- `--require` hooks won't do, we should support plugins and wrap them like how `caching-precompiler` wraps Babel
- The watcher needs to recompile changed source files before tests are rerun
- The test worker won't know which cache misses should have been precompiled, so tests may blow up. This would probably be due to insufficient `--source` patterns so would require useful error messaging somehow

vdemedes:
Am I confused or we already do this already via #390?

novemberborn:
#390 transpiles tests files. This is suggesting infrastructure to transpile source files in the main process as well.

vdemedes:
Yeah, but we don't transpile source files at all now. So we may start with that.

novemberborn:
Yes, that's the point of this issue. We'd have to use the approach from `caching-precompiler` to get the compiled sources into the child processes, and we have to use glob patterns to match all potential sources before tests run so they're available synchronously.

jamestalmage:
> we have to use glob patterns to match all potential sources ...

Or use [`babel-detective`](https://github.com/jamestalmage/babel-plugin-detective) which is probably more performant (the AST is being scanned already, so it should be faster than a regexp). I created it with this in mind.

Either way this won't do much good for dynamic `require`s. I did experiment with [this](https://github.com/jamestalmage/__share_transpile_across_process), which could be used as a fallback for dynamic `require` calls.

I think a first step should be supporting transpilation of helper files (files in the test directory prefixed with `_`, or in `test/helpers`). Sources would get a different Babel transform without `power-assert`, so helpers seems like an easier place to start.

jamestalmage:
For "non-Babel" transpilers (typescript, etc), we would have to create a plugin ecosystem. We would still need to finish up with Babel for powerAssert support. 

Conceivably, users may want to chain custom transforms. See [`transform-chain`](https://github.com/jamestalmage/transform-chain).

novemberborn:
This will be tackled through [RFC 001](https://github.com/avajs/ava/blob/master/docs/specs/001%20-%20Improving%20language%20support.md). See the project here: https://github.com/avajs/ava/projects/1.

StevenLiekens:
[...]

PR #743, opened 2016-04-09
Improve babelrc recipe

jamestalmage:
Better language, and more thorough linking to Babel's config documentation

I think this comes pretty close to fixing #447

### DIFF
--- a/docs/recipes/babelrc.md
+++ b/docs/recipes/babelrc.md
@@ -14,11 +14,11 @@ There are multiple options for configuring how AVA transpiles your tests using B
 
 ## AVA's default transpiler behavior
 
-By default, AVA transpiles your tests (and only your tests) using the [`es2015`] and [`stage-2`] Babel presets. This is a great option for small modules where you do not desire a build step to transpile your source before deploying to `npm`.
+By default, AVA transpiles your tests (and only your tests) using the [`es2015`](http://babeljs.io/docs/plugins/preset-es2015/) and [`stage-2`](http://babeljs.io/docs/plugins/preset-stage-2/) Babel presets. This is a great option for small modules where you do not desire a build step to transpile your source before deploying to `npm`.
 
 ## Customizing how AVA transpiles your tests
 
-You can override the default Babel configuration AVA uses for test transpilation in `package.json`. For example, the configuration below adds the Babel `rewire` plugin, and opts to only use the Babel [`stage-3`] preset (which is a subset of [`stage-2`]).
+You can override the default Babel configuration AVA uses for test transpilation in `package.json`. For example, the configuration below adds the Babel `rewire` plugin, and opts to only use the Babel [`stage-3`](http://babeljs.io/docs/plugins/preset-stage-3/) preset (which is a subset of [`stage-2`](http://babeljs.io/docs/plugins/preset-stage-2/)).
 
 ```json
 {
@@ -33,7 +33,7 @@ You can override the default Babel configuration AVA uses for test transpilation
 
 ## Transpiling Sources
 
-To transpile your sources, you will need to define a [`babel config` ][`babelrc`] in `package.json` or a `.babelrc` file. Also, you will need to tell AVA to load [`babel-register`] in every forked process, by adding it to the `require` section of your AVA config:
+To transpile your sources, you will need to define a [`babel config` ](http://babeljs.io/docs/usage/babelrc/) in `package.json` or a `.babelrc` file. Also, you will need to tell AVA to load [`babel-register`](http://babeljs.io/docs/usage/require/) in every forked process, by adding it to the `require` section of your AVA config:
 
 `package.json`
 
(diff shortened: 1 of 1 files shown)


docs/recipes/babelrc.md
```diff
@@ -4,50 +4,77 @@ Translations: [Français](https://github.com/sindresorhus/ava-docs/blob/master/f
...
+ - [Transpiling tests and sources the same way](#transpiling-tests-and-sources-the-same-way)
+ - [Extend your source transpilation configuration](#extend-your-source-transpilation-configuration)
+ - [Extend an alternate config file (i.e. not `.babelrc`)](#extend-an-alternate-config-file)
  - [Notes](#notes)
 
-## Specify a complete config in `package.json`
+## AVA's default transpiler behavior
 
-The `babelrc` option defaults to `false`, meaning `.babelrc` files are not considered when transpiling tests. This means you must specify your complete config in `package.json`.
+By default, AVA transpiles your tests (and only your tests) using Babels [`es2015`](http://babeljs.io/docs/plugins/preset-es2015/) and [`stage-2`](http://babeljs.io/docs/plugins/preset-stage-2/) presets. This is a great option for small modules where you do not desire a build step to transpile your source before deploying to `npm`.
```
sindresorhus: > By default, AVA transpiles your tests (and only your tests) using the es2015 and stage-2 Babel presets.

docs/recipes/babelrc.md
```diff
@@ -4,50 +4,77 @@ Translations: [Français](https://github.com/sindresorhus/ava-docs/blob/master/f
...
+  "ava": {
+    "require": ["babel-register"]
+  },
+  "babel": {
+    "presets": ["es2015"]
+  }
+}
+```
+
+Note that loading `babel-register` in every forked process has a non-trivial performance cost. If you have lots of test files, you may want to consider using a build step to transpile your sources *before* running your tests. This isn't ideal, since it complicates using AVA's watch mode, so we recommend using `babel-register` until the performance penalty becomes too great. Setting up a precompilation step is out of scope for this document, but we recommend you check out one of the many [build systems that support Babel](http://babeljs.io/docs/setup/). There is an open issue (#577) discussing ways we could make this experience better.
```
sindresorhus: You need to manually linkify `#577`

sindresorhus: Or maybe just linkify `open issue`. The number doesn't have much value.

docs/recipes/babelrc.md
```diff
@@ -4,50 +4,77 @@ Translations: [Français](https://github.com/sindresorhus/ava-docs/blob/master/f
...
+
+## Transpiling tests and sources the same way
+
+Using the `"inherit"` shortcut will cause your tests to be transpiled the same as your sources (as specified in your [`babelrc`](http://babeljs.io/docs/usage/babelrc/)). Ava will add a few additional [internal plugins](#notes) when transpiling your tests, but they won't affect the behavior of your test code.
```
sindresorhus: Ava => AVA

docs/recipes/babelrc.md
```diff
@@ -57,16 +84,17 @@ Set `babelrc` to `true`. This will use your `.babelrc` and extend it with any ad
 }
 ```
```
sindresorhus: Maybe include `babel` config here to show the source of what's merged for clarity?

jamestalmage: Here? This one doesn't inherit anything you mean lower in the "Extends" section?

sindresorhus: I think I meant line 64.

docs/recipes/babelrc.md
novemberborn: > This is great when testing modules that do not need to be transpiled before you can publish them to `npm`.

docs/recipes/babelrc.md
```diff
@@ -4,50 +4,77 @@ Translations: [Français](https://github.com/sindresorhus/ava-docs/blob/master/f
...
 
-## Specify a complete config in `package.json`
+## AVA's default transpiler behavior
 
-The `babelrc` option defaults to `false`, meaning `.babelrc` files are not considered when transpiling tests. This means you must specify your complete config in `package.json`.
+By default, AVA transpiles your tests (and only your tests) using Babels [`es2015`](http://babeljs.io/docs/plugins/preset-es2015/) and [`stage-2`](http://babeljs.io/docs/plugins/preset-stage-2/) presets. This is a great option for small modules where you do not desire a build step to transpile your source before deploying to `npm`.
+
+## Customizing how AVA transpiles your tests
+
+You can override the default Babel configuration AVA uses for test transpilation in `package.json`. For example, the configuration below adds the Babel `rewire` plugin, and opts to only use the Babel [`stage-3`](http://babeljs.io/docs/plugins/preset-stage-3/) preset (which is a subset of [`stage-2`]((http://babeljs.io/docs/plugins/preset-stage-2/))).
```
novemberborn: The parentheses for the `stage-2` link are messed up. Perhaps just use ``[`stage-2`]`` and define the links at the bottom?

jamestalmage:
Feedback incorporated.

forresst:
LGTM

docs/recipes/babelrc.md
sindresorhus: I think keeping the links inline, while a bit more messy, is more user-friendly as you don't have to go to the bottom to hunt down a link and then forget where you left off.

jamestalmage: The links still appear inline when rendered as HTML. No change to the rendered output. Just a bit cleaner in the raw document

sindresorhus: I mostly read raw markdown documents and having to hunt for matching links at the bottom gets annoying fast.

jamestalmage: OK - I changed them back.

novemberborn:
:shipit: 

Regarding the link syntax, I suspect it might be easier for translators if the links are defined once, in a single place. It'd be easier to change the links to point at translated content instead.

sindresorhus:
@novemberborn It's not very common to link to the exact same thing multiple times though.

sindresorhus:
🦄

Issue #724, opened 2016-04-06
Add stdev to benchmark results


jamestalmage:
Reference: https://twitter.com/trevnorris/status/712355178939199488

https://www.mathsisfun.com/data/standard-deviation-formulas.html

7373Lacym:
I would like to give this a shot.

sindresorhus:
@7373Lacym Go ahead :)

Let us know if you have any questions. We're also in https://gitter.im/sindresorhus/ava

sotojuan:
Fixed by #744.

PR #744, opened 2016-04-09
Added std dev to benchmark stats

7373Lacym:
Added standard deviation to the stats reported when benchmarking. For #724.


7373Lacym:
I couldn't find a place where we would test the changes I made.  If there is such a place let me know and I can add a few more commits

sindresorhus:
@jamestalmage Should we have some tests for the benchmarks?

jamestalmage:
Yes we should. Issue #816 opened.

PR #758, opened 2016-04-13
	watcher: remember failures from previous tests

### LINKED ISSUE #699: Watch mode should remember failures from tests it does not rerun.
## Issue

`failing-test.js`:

```js
import test from 'ava';

test(t => t.fail('oh no!'));
```

`passing-test.j2`:

```js
import test from 'ava';

test(t => t.pass('yeah!'));

// add more tests here
```

Given the above tests in watch mode, if you continue to edit `passing-test.js` (or any file that does not cause `failing-test.js` to be rerun), you will not be reminded that there are previously failing tests which were not rerun.

## Solution

I think we should track previous failures and show a message like:

```
3 passed
2 failing
2 previously failing tests were not rerun
```

The above shows 3 passing and 2 failing tests that were rerun. It also mentions 2 tests that were not rerun because the watch algorithm did not detect any changes.

## Alternate solution

Always rerun files that had failing tests regardless.

novemberborn:
Fixes #699. The watcher tracks how many failures (actual failures, but also rejections and uncaught exceptions) occurred for each test file. After each run, if there are failures from files that were not rerun, the count is added to the logger output.

### DIFF
--- a/api.js
+++ b/api.js
@@ -64,7 +64,7 @@ Api.prototype._onTimeout = function (runStatus) {
 
 	runStatus.handleExceptions({
 		exception: new AvaError(message),
-		file: null
+		file: undefined
 	});
 
 	runStatus.emit('timeout');
@@ -151,7 +151,7 @@ Api.prototype._run = function (files, _options) {
 					// continue to run.
 					runStatus.handleExceptions({
 						exception: err,
-						file: file
+						file: path.relative('.', file)
 					});
 
 					return {
@@ -197,7 +197,7 @@ Api.prototype._run = function (files, _options) {
 
 				runStatus.handleExceptions({
 					exception: err,
-					file: file
+					file: path.relative('.', file)
 				});
 
 				resolve([]);
--- a/lib/watcher.js
+++ b/lib/watcher.js
@@ -44,12 +44,10 @@ function Watcher(logger, api, files, sources) {
 
 	this.isTest = makeTestMatcher(files, AvaFiles.defaultExcludePatterns());
 
-	var isFirstRun = true;
 	this.clearLogOnNextRun = true;
+	this.runVector = 0;
 	this.run = function (specificFiles) {
-		if (isFirstRun) {
-			isFirstRun = false;
-		} else {
+		if (this.runVector > 0) {
 			var cleared = this.clearLogOnNextRun && logger.clear();
 			if (!cleared) {
 				logger.reset();
@@ -60,6 +58,8 @@ function Watcher(logger, api, files, sources) {
 			logger.reset();
 			logger.start();
 		}
+
+		var currentVector = this.runVector += 1;
 
 		var runOnlyExclusive = false;
 
@@ -82,6 +82,7 @@ function Watcher(logger, api, files, sources) {
 		this.busy = api.run(specificFiles || files, {
 			runOnlyExclusive: runOnlyExclusive
 		}).then(function (runStatus) {
+			runStatus.previousFailCount = self.sumPreviousFailures(currentVector);
 			logger.finish(runStatus);
 
 			var badCounts = runStatus.failCount + runStatus.rejectionCount + runStatus.exceptionCount;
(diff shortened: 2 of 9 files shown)


lib/watcher.js
```diff
@@ -173,10 +175,68 @@ Watcher.prototype.updateExclusivity = function (file, hasExclusiveTests) {
 	}
 };
 
+Watcher.prototype.trackFailures = function (api) {
```
jamestalmage: I don't think this tracking behavior belongs in watcher.

Let's enhance `RunStatus` to track both aggregate and per file failures, and give it the ability to track previous runs.

Here is a possible API:

```js
var testRun = api.run([...listofFiles]);

// someTime later
var previousRun = testRun;
testRun = api.run([...listOfFiles], previousRun);
```

Alternatively, maybe just create an "AggregateStatus" class:

```js
var combinedStatus = new AggregateStatus({
  previousRun: RunStatus-instance,
  currentRun: RunStatus-instance,
  files: list-of-files-that-are-being-rerun, // or not being rerun, whichever makes more sense
  deleted: list-of-deleted-tests
});
```

novemberborn: The current approach is relatively clean. We could do something like that if we're going to move more of the watcher functionality into the API maybe, or when we have a clearer idea of what programmatic interface we need to present to editors plugins.

jamestalmage: > The current approach is relatively clean

It's clean enough, but `watcher.js` as a whole is getting pretty complex. I'm just looking for ways to separate what it does into logical chunks.

api.js
```diff
@@ -64,7 +64,7 @@ Api.prototype._onTimeout = function (runStatus) {
 
 	runStatus.handleExceptions({
 		exception: new AvaError(message),
-		file: null
+		file: undefined
```
jamestalmage: ?

api.js
```diff
@@ -151,7 +151,7 @@ Api.prototype._run = function (files, _options) {
 					// continue to run.
 					runStatus.handleExceptions({
 						exception: err,
-						file: file
+						file: path.relative('.', file)
```
jamestalmage: This is more consistent with existing behavior in `fork`, etc. But maybe we should be emitting everything as absolute paths and have the reporters handle truncating to relative ones? Seems like it would be best to keep it explicit until we decide to show it to a human.

lib/watcher.js
novemberborn: Sure. Would be nice to do that later though :)

api.js
novemberborn: It was inconsistent.

api.js
novemberborn: Sure, but that should be done as a separate task. `fork` emits more of these files so I figured I'd make it consistent with that first.

jamestalmage:
LGTM.

I left a few comments, but they really require further discussion and shouldn't be holding up merging this.

I say :shipit:

lib/watcher.js
```diff
@@ -44,12 +44,10 @@ function Watcher(logger, api, files, sources) {
...
 	this.isTest = makeTestMatcher(files, AvaFiles.defaultExcludePatterns());
 
-	var isFirstRun = true;
 	this.clearLogOnNextRun = true;
+	this.runVector = 0;
 	this.run = function (specificFiles) {
-		if (isFirstRun) {
-			isFirstRun = false;
-		} else {
+		if (this.runVector > 0) {
```
kevva: `if (this.runVector)` or `if (this.runVector !== 0)`

api.js
jamestalmage: Agreed and understood.

lib/watcher.js
jamestalmage: Of those two I'd prefer `!== 0`, but I'm not sure why `> 0` is unacceptable. Because it implies `runVector` might be less than zero? Or some other reason?

lib/watcher.js
jamestalmage: Sure - as stated below - none of my comments need to hold this up.

lib/watcher.js
novemberborn: I'm trying to avoid implicit conversions. The vector is increasing so `> 0` signifies that more than `!== 0` would, but not fussed either way.

sindresorhus: `this.runVector > 0` is the best option here. I would use `!== 0` when it can also be negative.

sindresorhus:
:shipit:

PR #759, opened 2016-04-14
[WIP]: rerunnable test stats

jamestalmage:
This creates a `TestStats` object we can use to track various stats for the current run (and previous runs when in watch mode).

Critically, it's smart about tallying pass, fail, skip, and todo counts even with complicated rerun patterns.

Born from [this comment](https://github.com/sindresorhus/ava/pull/758/files#r59601446).

I haven't integrated it with anything yet. Just submitting for feedback.

### DIFF
--- a/lib/test-stats.js
+++ b/lib/test-stats.js
@@ -1,0 +1,89 @@
+'use strict';
+
+function TestStats(opts) {
+	if (!(this instanceof TestStats)) {
+		throw new TypeError('Class constructor TestStats cannot be invoked without \'new\'');
+	}
+
+	this.files = opts.files.slice(0);
+	this.errors = [];
+	this.results = Object.create(null);
+
+	this.files.forEach(function (file) {
+		this.results[file] = blankStats();
+	}, this);
+
+	if (opts.previousFiles || opts.previousStats) {
+		this.previousFiles = opts.previousFiles.slice(0);
+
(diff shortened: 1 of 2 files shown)


sindresorhus:
Hard to give feedback without seeing how it's going to be used, but I like the idea in general.

Are we still doing this?

Maybe @novemberborn can share some thoughts on it.

jamestalmage:
This is kinda what RunStatus has become. We definitely don't need the reset ability, because we create a new RunStatus for each pass.

jamestalmage:
OK, so apparently it was designed with RunStatus in mind. I will take another look this weekend, and close if I don't think it's worth doing anymore.

novemberborn:
Would be nice to draw up a desired architecture for how the API, CLI & Watcher interact with each other and how to abstract the shared pieces, including the logging. There's a fair amount of coupling at the moment, through methods that exist just to enable a slight variation of the behavior.

jamestalmage:
Yes. That sounds like a good plan. Not a big fan of writing code to design documents, but as long as we don't take them as gospel, and just use it to agree on some basic goals.

novemberborn:
> Not a big fan of writing code to design documents, but as long as we don't take them as gospel, and just use it to agree on some basic goals.

Indeed. Helps clarify direction.

jamestalmage:
I've opened #865, which basically just documents what we have now.

PR #760, opened 2016-04-14
compute cache salt using package-hash

novemberborn:
I just wrote <https://github.com/novemberborn/package-hash> which is a more powerful way of computing hashes for packages. It includes the GIt repo state which is very useful during development. Say if you're hacking on AVA and `npm link` it into a test project, now you don't need to clear the AVA cache in your test project whenever you make changes to AVA.

Feedback welcome!

### DIFF
--- a/lib/caching-precompiler.js
+++ b/lib/caching-precompiler.js
@@ -6,6 +6,7 @@ var cachingTransform = require('caching-transform');
 var objectAssign = require('object-assign');
 var stripBom = require('strip-bom');
 var md5Hex = require('md5-hex');
+var packageHash = require('package-hash');
 var enhanceAssert = require('./enhance-assert');
 
 function CachingPrecompiler(cacheDirPath, babelConfig) {
@@ -159,14 +160,11 @@ CachingPrecompiler.prototype._createEspowerPlugin = function () {
 };
 
 CachingPrecompiler.prototype._createTransform = function () {
-	var dependencies = {
-		'babel-plugin-espower': require('babel-plugin-espower/package.json').version,
-		'ava': require('../package.json').version,
-		'babel-core': require('babel-core/package.json').version,
-		'babelConfig': this.babelConfig
-	};
-
-	var salt = new Buffer(JSON.stringify(dependencies));
+	var salt = packageHash.sync([
+		require.resolve('../package.json'),
+		require.resolve('babel-core/package.json'),
+		require.resolve('babel-plugin-espower/package.json')
+	], JSON.stringify(this.babelConfig));
 
 	return cachingTransform({
 		factory: this._factory,
--- a/lib/caching-precompiler.js
+++ b/lib/caching-precompiler.js
@@ -14,7 +14,7 @@ function CachingPrecompiler(cacheDirPath, babelConfig) {
 		throw new TypeError('Class constructor CachingPrecompiler cannot be invoked without \'new\'');
 	}
 
-	this.babelConfig = babelConfig;
+	this.babelConfig = babelConfig || 'default';
 	this.cacheDirPath = cacheDirPath;
 	this.fileHashes = {};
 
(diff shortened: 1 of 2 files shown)


sindresorhus:
I like the idea in general, but [this](https://github.com/novemberborn/package-hash/blob/69a328840818d336ae13af22f9af092208fcb49b/src/index.js#L31) has me a bit worried. `execSync` is very slow.

jamestalmage:
> execSync is very slow.

`execFileSync` is better. I can't remember when it was introduced. (I wish the Node docs had a `since: version` notation).

novemberborn:
It won't compute a diff when a package is installed from npm, so it shouldn't be too bad in practice.

sindresorhus:
Might be useful if `packageHash` supports accepting an array of package paths to hash. So we could just do:

```js
packageHash.sync([
  require.resolve('../package.json'),
  require.resolve('babel-core/package.json'),
  require.resolve('babel-plugin-espower/package.json')
]);
```

Maybe even some an option with extra text to hash:

```js
var salt = packageHash.sync([
  require.resolve('../package.json'),
  require.resolve('babel-core/package.json'),
  require.resolve('babel-plugin-espower/package.json')
], {
  extra: JSON.stringify(this.babelConfig)
});
```

jamestalmage:
Answering my own question:

`execFile` and `execFileSync` were both introduced in `v0.12`, so it would be better to just use `execFileSync`.

jamestalmage:
> Might be useful if packageHash supports accepting an array of package paths to hash.

:+1: 

> Maybe even some an option with extra text to hash

Are you mixing concerns at that point?

lib/caching-precompiler.js
```diff
@@ -159,14 +160,12 @@ CachingPrecompiler.prototype._createEspowerPlugin = function () {
...
-		'babel-core': require('babel-core/package.json').version,
-		'babelConfig': this.babelConfig
-	};
-
-	var salt = new Buffer(JSON.stringify(dependencies));
+	var salt = md5Hex([
+		packageHash.sync(require.resolve('../package.json')),
+		packageHash.sync(require.resolve('babel-core/package.json')),
+		packageHash.sync(require.resolve('babel-plugin-espower/package.json')),
+		JSON.stringify(this.babelConfig)
```
jamestalmage:
```js
JSON.stringify(this.babelConfig) || ''
```

Or maybe we should allow `undefined` in `md5-hash`

sindresorhus:
> Are you mixing concerns at that point?

Yeah, probably not a good suggestion.

lib/caching-precompiler.js
sindresorhus: > Or maybe we should allow undefined in md5-hash

What would the behavior be?

jamestalmage: Allowing arrays with holes in them, for this type of situation. We could just not call `.update()` when we find a hole. Or call update with a special placeholder set of data (so that the presence and position of holes in the array affects the final hash).

Still, probably best just to fix it here

Not sure it's a good idea.

sindresorhus: Yeah, hashing is not the place you want implicit behavior.

novemberborn: @jamestalmage I could also [force the default](https://github.com/sindresorhus/ava/blob/e076b5bab3f17d6935f467108eda55fa38b90bb1/lib/caching-precompiler.js#L16) in the constructor.

novemberborn:
I'll try and use `execFileSync` to get the diff. I like the idea of taking an array of paths, and it shouldn't be too bad to allow an optional seed, e.g. `packageHash([path1, path2], seed)`.

lib/caching-precompiler.js
jamestalmage: Do that

novemberborn:
Updated.

sindresorhus:
👍

PR #761, opened 2016-04-15
Fix region dependent test failures.

jamestalmage:
This mocks time using the region dependent Date constructor, so reporter output always shows the same time regardless.

non-controversial, so I will be merging as soon as I verify Travis passes

### DIFF
--- a/test/reporters/mini.js
+++ b/test/reporters/mini.js
@@ -29,7 +29,7 @@ function miniReporter() {
 
 process.stderr.setMaxListeners(50);
 
-lolex.install(new Date('2014-11-19T00:19:12+0700').getTime(), ['Date']);
+lolex.install(new Date(2014, 11, 19, 17, 19, 12, 200).getTime(), ['Date']);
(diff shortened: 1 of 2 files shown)


novemberborn:
What issue did you encounter? The timestamp is in ISO8601 with a timezone included, and when I wrote this I was in BST not UTC so I'm surprised it would give different results.

jamestalmage:
You specify the date for `lolex` in a timezone *independent* way, but the output timezone *dependent* (it prints the local time in 24 hour format).

I was seeing 1200 hours bs the expected 1700

jamestalmage:
Weird that Travis passes. Are they BST?

novemberborn:
I'd imagine Travis is UTC. It's probably something else in your environment.

I understand your change now, it doesn't use the timezone so the time will always be correct. The original date (in UTC) was Nov 18th though :wink: 

Has anybody figured out the easter egg in the original timestamp yet?

PR #766, opened 2016-04-16
[WIP]: refactor watcher

jamestalmage:
So far I've only moved some of the path processing stuff into AvaFiles.

### DIFF
--- a/lib/ava-files.js
+++ b/lib/ava-files.js
@@ -1,6 +1,5 @@
 var fs = require('fs');
 var path = require('path');
-var nodePath = require('path');
 var Promise = require('bluebird');
 var slash = require('slash');
 var globby = require('globby');
@@ -51,13 +50,13 @@ function getDefaultIgnorePatterns() {
 	});
 }
 
-// Used on paths before they're passed to multimatch to Harmonize matching
+// Used on paths before they're passed to multimatch to harmonize matching
 // across platforms.
 var matchable = process.platform === 'win32' ? slash : function (path) {
 	return path;
 };
 
-AvaFiles.prototype.makeSourceMatcher = function makeSourceMatcher() {
+AvaFiles.prototype.makeSourceMatcher = function () {
 	var mixedPatterns = [];
 	var defaultIgnorePatterns = getDefaultIgnorePatterns();
 	var overrideDefaultIgnorePatterns = [];
@@ -110,23 +109,23 @@ AvaFiles.prototype.makeSourceMatcher = function makeSourceMatcher() {
 	};
 };
 
-AvaFiles.prototype.makeTestMatcher = function makeTestMatcher() {
+AvaFiles.prototype.makeTestMatcher = function () {
 	var excludePatterns = this.excludePatterns;
 	var initialPatterns = this.files.concat(excludePatterns);
 
-	return function (path) {
+	return function (filepath) {
 		// Like in api.js, tests must be .js files and not start with _
-		if (nodePath.extname(path) !== '.js' || nodePath.basename(path)[0] === '_') {
+		if (path.extname(filepath) !== '.js' || path.basename(filepath)[0] === '_') {
 			return false;
 		}
 
 		// Check if the entire path matches a pattern.
-		if (multimatch(matchable(path), initialPatterns).length === 1) {
+		if (multimatch(matchable(filepath), initialPatterns).length === 1) {
 			return true;
 		}
 
 		// Check if the path contains any directory components.
-		var dirname = nodePath.dirname(path);
+		var dirname = path.dirname(filepath);
 		if (dirname === '.') {
 			return false;
 		}
@@ -155,11 +154,11 @@ AvaFiles.prototype.makeTestMatcher = function makeTestMatcher() {
 
 		// See if the entire path matches any of the subpaths patterns, taking the
 		// excludePatterns into account. This mimicks the behavior in api.js
-		return multimatch(matchable(path), recursivePatterns.concat(excludePatterns)).length === 1;
+		return multimatch(matchable(filepath), recursivePatterns.concat(excludePatterns)).length === 1;
 	};
 };
 
-AvaFiles.prototype.getChokidarPatterns = function getChokidarPatterns() {
+AvaFiles.prototype.getChokidarPatterns = function () {
 	var paths = [];
 	var ignored = [];
 
--- a/lib/ava-files.js
+++ b/lib/ava-files.js
@@ -132,7 +132,7 @@ AvaFiles.prototype.makeTestMatcher = function () {
 
 		// Compute all possible subpaths. Note that the dirname is assumed to be
 		// relative to the working directory, without a leading `./`.
-		var subpaths = dirname.split(nodePath.sep).reduce(function (subpaths, component) {
+		var subpaths = dirname.split(/[\\\/]/).reduce(function (subpaths, component) {
 			var parent = subpaths[subpaths.length - 1];
 			if (parent) {
 				// Always use / to makes multimatch consistent across platforms.
--- a/lib/ava-files.js
+++ b/lib/ava-files.js
@@ -65,6 +65,7 @@ AvaFiles.prototype.makeSourceMatcher = function makeSourceMatcher() {
 	var hasPositivePattern = false;
 	this.sources.forEach(function (pattern) {
 		mixedPatterns.push(pattern);
+		// TODO: why not just pattern[0] !== '!'
 		if (!hasPositivePattern && pattern[0] !== '!') {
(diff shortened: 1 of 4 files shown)


novemberborn:
@jamestalmage sorry haven't had a chance yet to look at this.

jamestalmage:
@novemberborn - bump!

I'd like to push forward with this. There are a number of TODO comments in this PR with your name on it (quite literally). I'd like your feedback on how we should proceed.

test/watcher.js
```diff
@@ -51,75 +51,75 @@ test('chokidar is not installed', function (t) {
...
+	var Subject;
+	var runStatus;
+	var resetRunStatus;
+	var clock;
+	var chokidarEmitter;
+	var stdin;
+	var files;
 
-	var debug = sinon.spy();
+	function proxyWalker(opts) {
```
novemberborn: Walker?

lib/ava-files.js
```diff
@@ -35,12 +38,161 @@ function AvaFiles(files) {
...
 	return handlePaths(this.files, this.excludePatterns);
 };
 
+function getDefaultIgnorePatterns() {
+	return defaultIgnore.map(function (dir) {
+		return dir + '/**/*';
+	});
+}
+
+// Used on paths before they're passed to multimatch to Harmonize matching
```
novemberborn: Bad capitalization of `Harmonize` (my bad!)

test/watcher.js
```diff
@@ -566,9 +553,15 @@ group('chokidar is installed', function (beforeEach, test, group) {
 	test('initial exclude patterns override whether something is a test file', function (t) {
 		t.plan(2);
 
+		avaFiles = function (files, sources) {
+			var ret = new AvaFiles(files, sources);
+			// TODO(@jamestalmage, @novemberborn): There is no way for users to actually set exclude patterns yet.
```
novemberborn: Remind me, what are we looking to exclude here?

test/ava-files.js
```diff
@@ -0,0 +1,67 @@
...
+	notTest('foo/_foo-bar.js');
+	notTest('foo-bar.txt');
+	notTest('node_modules/foo.js');
+	notTest('fixtures/foo.js');
+	notTest('helpers/foo.js');
+	t.end();
+});
+
+test('sourceMatcher', function (t) {
+	var avaFiles = new AvaFiles(['**/foo*']);
```
novemberborn: You're specifying test patterns though?

test/ava-files.js
```diff
@@ -0,0 +1,67 @@
...
+
+	function isSource(file) {
+		t.true(matcher(file), file + ' should be a source');
+	}
+
+	function notSource(file) {
+		t.false(matcher(file), file + ' should not be a source');
+	}
+
+	// TODO: Should these actually not be considered sources since they would match as a test?
```
novemberborn: It's a trick question. Tests aren't sources, but currently the watcher deems everything that's not a test to be a source. It assumes Chokidar watches the union of tests and sources, and no more.

`isSource` is used to filter test file dependencies.

test/ava-files.js
```diff
@@ -0,0 +1,67 @@
...
+	isSource('foo-bar.js');
+	isSource('foo.js');
+	isSource('foo/blah.js');
+	isSource('bar/foo.js');
+
+	// TODO: Same as above - shouldn't these be "not sources" - they are test fixtures and helpers
```
novemberborn: Fixtures and helpers are sources, not tests.

test/ava-files.js
```diff
@@ -0,0 +1,67 @@
...
+	isSource('_foo-bar.js');
+	isSource('foo/_foo-bar.js');
+	isSource('fixtures/foo.js');
+	isSource('helpers/foo.js');
+
+	// TODO: What if a non `.js` file is read with `fs.readFileXXX` and has an impact on behavior? Shouldn't changing it trigger a re-run? (Obviously can't take advantage of smart watcher behavior).
```
novemberborn: That's out of scope for dependency detection. The watcher won't be able to determine which test file depends on that source, so it'll rerun all tests.

Funny situation though if one test file requires a source, and the other `fs.readFile`'s it. Only the first test file will be rerun.

lib/ava-files.js
```diff
@@ -65,6 +65,7 @@ AvaFiles.prototype.makeSourceMatcher = function makeSourceMatcher() {
 	var hasPositivePattern = false;
 	this.sources.forEach(function (pattern) {
 		mixedPatterns.push(pattern);
+		// TODO: why not just pattern[0] !== '!'
```
novemberborn: It saves an assignment, I suppose.

novemberborn:
@jamestalmage very cool!

test/watcher.js
jamestalmage: I dunno. We have `defaultExcludes`, but that gets passed to watcher without modification. You had a test for watcher that assumed it "could" be modified at some point. We just don't have any logic for allowing users to do that. I just copied the test.

If there's no way for a user to modify it, is there a point to a test that verifies behavior when it's modified?

test/ava-files.js
jamestalmage: So, should `isSource` be renamed to `trackedFile`? Something else?

test/ava-files.js
jamestalmage: Not sure I understand the question. (this comment shows as outdated - looks like you're commenting commit by commit?)

test/watcher.js
novemberborn: Ah, I think this is just stubbing the default exclude or something.

test/ava-files.js
novemberborn: Does it matter? The source patterns are likely to match test files, so the two sets may sometimes intersect, but depending on the config they won't always.

test/ava-files.js
novemberborn: Yea I went commit by commit. It's clear in the full diff that this tests the defaults.

test/watcher.js
vdemedes: giphy 10

lib/ava-files.js
```diff
@@ -1,9 +1,12 @@
 var fs = require('fs');
 var path = require('path');
+var nodePath = require('path');
```
sindresorhus: This is the same as the above.

lib/ava-files.js
```diff
@@ -35,12 +38,162 @@ function AvaFiles(files) {
...
 
+function getDefaultIgnorePatterns() {
+	return defaultIgnore.map(function (dir) {
+		return dir + '/**/*';
+	});
+}
+
+// Used on paths before they're passed to multimatch to Harmonize matching
+// across platforms.
+var matchable = process.platform === 'win32' ? slash : function (path) {
```
sindresorhus: I wonder if I should just make the `slash` module do this check? https://github.com/sindresorhus/slash/blob/master/index.js

lib/ava-files.js
```diff
@@ -35,12 +38,162 @@ function AvaFiles(files) {
...
+	return path;
+};
+
+AvaFiles.prototype.makeSourceMatcher = function makeSourceMatcher() {
```
sindresorhus: No point in naming the function really. Modern engines can infer it.

lib/ava-files.js
jamestalmage: yeah - it's from combining the two codebases. `nodePath` was used in watcher, because he used a `path` variable inside methods. I can clean that up.

lib/ava-files.js
novemberborn: This is an optimization because we know we're dealing with file paths, so only on win32 do we need to convert the slashes.

sindresorhus: I got that. I've seen this done before too, so was just curios if it makes sense to move the platform check into the slash module?

novemberborn: I don't think so. `slash` itself doesn't know where the string comes from, just because it's not running on `win32` doesn't mean the string doesn't contain slashes that should be replaced.

sindresorhus:
What's left for this to land?

test/ava-files.js
jamestalmage: Sorry, I didn't phrase this right.

The correct question is, what about `require('foo-bar.json')` or `require('foo-bar.coffee')`.

jamestalmage:
Assuming the rest of the AppVeyor runs pass, this should be mergeable.

novemberborn:
> The correct question is, what about require('foo-bar.json') or require('foo-bar.coffee').
> 
> Does this mean smart walking won't work with .coffee?

I assume you meant `./foo-bar.coffee`? But that should be picked up as a dependency, matchable against watched sources.

PR #767, opened 2016-04-18
Removed stray characters from CLI sample output

ingve:
The sample CLI output in the readme was presumably copied from the source in cli.js, but a single quote and comma seems to have been left behind.


sindresorhus:
Good catch :)

PR #769, opened 2016-04-18
Docs: update t.plan recipe. `Fixes #749`

### LINKED ISSUE #749: Update "good use of t.plan" recipe.
Now that we've settled on a better behavior for `t.throws`, I think two sections of the `t.plan` recipe should be updated.

1. [Promises with a `catch` block ](https://github.com/sindresorhus/ava/blob/master/docs/recipes/when-to-use-plan.md#promises-with-a-catch-block)

 > ```js
> test(t => {
>    t.plan(2);
>
>     return shouldRejectWithFoo().catch(reason => {
>         t.is(reason.message, 'Hello') // Prefer t.throws() if all you care about is the message
>         t.is(reason.foo, 'bar');
>     });
> });
> ```
>
> Here, t.plan() is used to ensure the code inside the catch block happens. In most cases, you should prefer the t.throws() assertion, but this is an acceptable use since t.throws() only allows you to assert against the error's message property.

  This is no longer relevant. You no longer need `t.plan` for this, and you should not use the catch block. Instead `await` the result of `t.throws`:

  ```js
  test(async t => {
      const reason = await t.throws(shouldRejectWithFoo());
      t.is(reason.message, 'Hello'); // you could assert the message as the second arg to `t.throws`, but I think this provides more clarity
      t.is(reason.foo, 'bar');
  });
  ```

2. [Ensuring a catch statement happens](https://github.com/sindresorhus/ava/blob/master/docs/recipes/when-to-use-plan.md#ensuring-a-catch-statement-happens)

  This can be updated in basically the same manner as `1`.

The existing examples should be moved to the "bad" section, as well as providing the "good" example of how to fix it above.

I think it's worth noting somewhere in that discussion that proper use of `t.throws` and the `await` statement flattens out the code, making it very easy to reason about.

kevmannn:
Updates the recipe surrounding proper uses of `t.plan()` due to improved `t.throws()` behavior.

https://github.com/sindresorhus/ava/issues/749

### DIFF
--- a/docs/recipes/when-to-use-plan.md
+++ b/docs/recipes/when-to-use-plan.md
@@ -48,10 +48,6 @@ test(async t => {
 });
 ```
 
-## Good uses of `t.plan()`
-
-`t.plan()` has many acceptable uses.
-
 ### Promises with a `.catch()` block
 
 ```js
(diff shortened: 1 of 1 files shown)


sotojuan:
Looks good to me!

sindresorhus:
Thank you @kevmannn :)

PR #770, opened 2016-04-18
Enhance AssertionError message - `Fixes #764`

### LINKED ISSUE #764: Show `expected` and `actual` if they exist on the AssertionError object.
The following:

``` js
import assert from 'assert';
import test from 'test';

test(() => {
  assert.equal(3, 4, 'foo');
});
```

produces this output:

```
   1 failed

  1. [anonymous]
  foo
      Test.fn (test.js:4:9)
```

If you do not specify a message, you get a more helpful `3==4` in the message. If we get an AssertionError that was not enhanced by `power-assert`, we should be displaying (at a minimum) those two properties.

Mocha's output (for reference):

```
  0 passing (9ms)
  1 failing

  1)  foo:

      AssertionError: foo
      + expected - actual

      -3
      +4
```

danilosampaio:
This adds `expected` and `actual` to AssertionError message when it's not enhanced by power-assert.
Fixes #764

#### Before
```
  1. [anonymous]
  foo
      Test.fn (test.js:5:10)
```

#### After

```
   1. [anonymous]
   
    AssertionError: foo
    + expected - actual

    -3
    +4
      Test.fn (test.js:5:10)
```


jamestalmage:
@danilosampaio - Decent start. We probably want to colorize the output a bit. Here's mocha:

screenshot 2016-04-18 19 39 52

I'm not sure what mocha uses, but we should look into that.

Also - [Unexpected.js](http://unexpected.js.org/) seems to put a lot of effort into creating beautiful error messages. It's worth looking into what they use as well.

danilosampaio:
@jamestalmage thanks for your feedback.
I'll take a look at Mocha and Unexpected.js.

danilosampaio:
`Mocha` holds a color map:
https://github.com/mochajs/mocha/blob/9ae6a85b6ba50ea415affc88456d469d6915a157/lib/reporters/base.js#L52-L72

then inserts the ansi codes into the given string:
https://github.com/mochajs/mocha/blob/9ae6a85b6ba50ea415affc88456d469d6915a157/lib/reporters/base.js#L52-L72

`Unexpectedjs` uses hexadecimal color palette that seems not to fit into AVA.

So i made the suggested changes using `chalk`. what do you think?

jamestalmage:
It's definitely better than what we have. I think there is room for further improvements (like a diff view for larger objects / strings), but we can address that in a future PR.

Can you add a test for this behavior. Probably in `test/cli.js`.

danilosampaio:
@jamestalmage sure!

danilosampaio:
Added some tests.

test/cli.js
```diff
@@ -290,3 +290,33 @@ test('handles NODE_PATH', function (t) {
...
+
+test('Show `expected` and `actual` with a message', function (t) {
+	execCli('fixture/assert-equal-with-message.js', function (err, stdout, stderr) {
+		t.ok(err);
+
+		var expectedOutput =
+			'\n    \u001b[31mAssertionError: foo\u001b[31m\n' +
+			'\u001b[32m    + expected\u001b[31m\u001b[31m - actual\u001b[31m\n\n' +
+			'\u001b[31m    -3\u001b[31m\n' +
+			'\u001b[32m    +4\u001b[31m';
```
novemberborn: We tend to repeat the `chalk.red()` etc calls in the test, it makes it easier to change things later. You may also find the [`compareLineOutput`](https://github.com/sindresorhus/ava/blob/master/test/helper/compare-line-output.js) helper useful. Here's an example: https://github.com/sindresorhus/ava/blob/7dd7055d360569b937515e907a5af22df07cbb22/test/reporters/mini.js#L268:L281

lib/run-status.js
```diff
@@ -126,6 +128,17 @@ RunStatus.prototype.handleTest = function (test) {
...
 
+RunStatus.prototype.formatAssertionError = function (error) {
+	if (error.actual && error.expected) {
+		return '\n    ' + chalk.red(error.name + ': ' + error.message) + '\n' +
+			chalk.green('    + expected') + chalk.red(' - actual') + '\n\n' +
+			chalk.red('    -' + error.actual) + '\n' +
+			chalk.green('    +' + error.expected);
+	}
+
+	return chalk.red('\n    ' + error.name + ': ' + error.message);
```
novemberborn: Could you add a test for this as well?

novemberborn:
Looking good!

danilosampaio:
@novemberborn thanks for your feedback. I'll take a look at it.

test/cli.js
danilosampaio: Done. :+1:

lib/run-status.js
danilosampaio: Done. :+1:
I assumed not necessary create a new test/run-status.js, since the added tests already check RunStatus.formatAssertionError() output. Right? :smiley:

lib/run-status.js
```diff
@@ -126,6 +128,17 @@ RunStatus.prototype.handleTest = function (test) {
 	this.emit('test', test, this);
 };
 
+RunStatus.prototype.formatAssertionError = function (error) {
```
jamestalmage: This kinda feels like the wrong place to be doing this.

Maybe in the individual reporters?

test/cli.js
```diff
@@ -290,3 +292,89 @@ test('handles NODE_PATH', function (t) {
 		t.end();
 	});
 });
+
+test('Show `expected` and `actual` without a message', function (t) {
+	execCli('fixture/assert-equal-without-message.js', function (err) {
+		t.ok(err);
+
+		var runStatus = {
```
jamestalmage: Why this? Why not just:

```js
var runStatus = new RunStatus();
```

test/cli.js
```diff
@@ -290,3 +292,89 @@ test('handles NODE_PATH', function (t) {
...
+
+		t.end();
+	});
+});
+
+test('Show `expected` and `actual` with a message', function (t) {
+	execCli('fixture/assert-equal-with-message.js', function (err) {
+		t.ok(err);
+
+		var runStatus = {
```
jamestalmage: same as above

test/cli.js
```diff
@@ -290,3 +292,89 @@ test('handles NODE_PATH', function (t) {
...
+		};
+
+		var assertionError = {
+			actual: 4,
+			expected: 3,
+			name: 'AssertionError',
+			message: 'foo'
+		};
+
+		var output = runStatus.formatAssertionError(assertionError);
```
jamestalmage: We are ignoring the output of the cli run here?

We should be comparing this to `stdout`, not generating a new output.

lib/run-status.js
danilosampaio: since other improvements will be added in a second moment (like a diff view for larger objects / strings), maybe better option is a individual reporter.

sindresorhus:
It's weird choice having the enhanced assertion as a reporter. It's not. It's a helper.

jamestalmage:
But it's strictly reporter related. I get the desire here, but if we are committed to keeping a small list of 3 reporters, giving them their own directories, while we let multiple reporter specific "helpers" proliferate in the parent directory seems silly.

The following files are all directly tied to reporting, but are cluttering up `lib`:
- `logger.js`
- `colors.js`
- `beautify-stack.js`

Add `enhanced-assertion` to that list, and we've got more reporter related files living outside the `reporters` folder than in.

Can we come up with a different folder name / structure that let's us move everything reporter related in there?

danilosampaio:
Can i go with this approach?

- Move `logger.js`, `colors.js`, `beautify-stack`, and `enhanced-assertion.js` to `lib/reporters/helper/`
- Refator references to those files

sindresorhus:
👍 

`lib/reporters/helper/` → `lib/reporters/helpers/`

jamestalmage:
This needs a rebase. 

Please ping me when it's done, and we will get this merged.

danilosampaio:
Ok, i'll do it tomorow.

danilosampaio:
@jamestalmage 
Finished the rebase. If you wish, i can squash the commits.

jamestalmage:
Nah. GitHub does that for us now. I now prefer people don't squash and let us do it from the web ui. Makes it much easier to track changes since I last reviewed.

cli.js
```diff
@@ -26,11 +26,11 @@ var Promise = require('bluebird');
 var pkgConf = require('pkg-conf');
 var isCi = require('is-ci');
 var hasFlag = require('has-flag');
-var colors = require('./lib/colors');
+var colors = require('./lib/reporters/helpers/colors');
```
jamestalmage: Meh. I forgot all the places colors is used. Not sure about having it here now.

Maybe follow the same convention we use for test helpers?

```
lib/reporters/_colors
```

IDK.

jamestalmage:
This looks good to me. The reporters folder structure still feels a bit off, but we've let that hijack a PR that was never about folder structure in the first place.

Nice work @danilosampaio.

lib/reporters/helpers/enhanced-assertion.js
```diff
@@ -0,0 +1,20 @@
+var chalk = require('chalk');
+
+function EnhancedAssertion() {
+	if (!(this instanceof EnhancedAssertion)) {
+		return new EnhancedAssertion();
```
novemberborn: This should throw: `new EnhancedAssertion()` should be used instead. See [`Hook`](https://github.com/avajs/ava/blob/4d3ed27865dc7cdfde7e651711ee4cd0646ea6e8/lib/hook.js#L8) for an example.

lib/test.js
```diff
@@ -14,6 +14,7 @@ var enhanceAssert = require('./enhance-assert');
 var globals = require('./globals');
 var throwsHelper = require('./throws-helper');
 var formatter = enhanceAssert.formatter();
+var enhancedAssertion = require('./reporters/helpers/enhanced-assertion.js')();
```
novemberborn: Could you remove the `.js` in this `require` call?

novemberborn:
I'm a little confused about what we're trying to achieve with this PR.

I've got this test file:

```js
import assert from 'assert';
import test from 'ava';

test('assert.equal with message', () => {
	assert.equal(3, 4, 'foo');
});

test('assert.equal without message', () => {
	assert.equal(3, 4);
});

test('t.deepEqual with message', t => {
	t.deepEqual(3, 4, 'foo');
});

test('t.deepEqual without message', t => {
	t.deepEqual(3, 4, 'foo');
});

test('t.true with message', t => {
	t.true(3 === 4, 'foo');
});

test('t.true without message', t => {
	t.true(3 === 4, 'foo');
});
```

Linked against this PR the output is:

```
$(npm bin)/ava

   6 failed

   1. assert.equal with message
   foo
      Test.fn (test.js:5:9)
    _combinedTickCallback (internal/process/next_tick.js:67:7)
    process._tickCallback (internal/process/next_tick.js:98:9)


   2. assert.equal without message
   3 == 4
      Test.fn (test.js:9:9)
    _combinedTickCallback (internal/process/next_tick.js:67:7)
    process._tickCallback (internal/process/next_tick.js:98:9)


   3. t.deepEqual with message

    AssertionError: foo
    + expected - actual

    -3
    +4
          +4
        Test.fn (test.js:13:4)
        _combinedTickCallback (internal/process/next_tick.js:67:7)
        process._tickCallback (internal/process/next_tick.js:98:9)


   4. t.deepEqual without message

    AssertionError: foo
    + expected - actual

    -3
    +4
          +4
        Test.fn (test.js:17:4)
        _combinedTickCallback (internal/process/next_tick.js:67:7)
        process._tickCallback (internal/process/next_tick.js:98:9)


   5. t.true with message
   foo
  t.true(3 === 4, 'foo')



      t.true(3 === 4, 'foo')
        Test.fn (test.js:21:2)
        _combinedTickCallback (internal/process/next_tick.js:67:7)
        process._tickCallback (internal/process/next_tick.js:98:9)


   6. t.true without message
   foo
  t.true(3 === 4, 'foo')



      t.true(3 === 4, 'foo')
        Test.fn (test.js:25:2)
        _combinedTickCallback (internal/process/next_tick.js:67:7)
        process._tickCallback (internal/process/next_tick.js:98:9)
```

Linked against `master`:

```
$(npm bin)/ava

   6 failed

   1. assert.equal with message
   foo
      Test.fn (test.js:5:9)
    _combinedTickCallback (internal/process/next_tick.js:67:7)
    process._tickCallback (internal/process/next_tick.js:98:9)


   2. assert.equal without message
   3 == 4
      Test.fn (test.js:9:9)
    _combinedTickCallback (internal/process/next_tick.js:67:7)
    process._tickCallback (internal/process/next_tick.js:98:9)


   3. t.deepEqual with message
   foo
      Test.fn (test.js:13:4)
    _combinedTickCallback (internal/process/next_tick.js:67:7)
    process._tickCallback (internal/process/next_tick.js:98:9)


   4. t.deepEqual without message
   foo
      Test.fn (test.js:17:4)
    _combinedTickCallback (internal/process/next_tick.js:67:7)
    process._tickCallback (internal/process/next_tick.js:98:9)


   5. t.true with message
   foo
  t.true(3 === 4, 'foo')



      t.true(3 === 4, 'foo')
        Test.fn (test.js:21:2)
        _combinedTickCallback (internal/process/next_tick.js:67:7)
        process._tickCallback (internal/process/next_tick.js:98:9)


   6. t.true without message
   foo
  t.true(3 === 4, 'foo')



      t.true(3 === 4, 'foo')
        Test.fn (test.js:25:2)
        _combinedTickCallback (internal/process/next_tick.js:67:7)
        process._tickCallback (internal/process/next_tick.js:98:9)
```

The only change seems to be for `t.deepEqual()`, and while there is a bit more detail I'm actually quite confused by the output.

Am I missing something?

lib/reporters/helpers/enhanced-assertion.js
danilosampaio: Fixed.

lib/test.js
danilosampaio: Fixed.

danilosampaio:
@novemberborn thanks for your feedback.
You're right, this is not the expected result of this PR. After [this](https://github.com/avajs/ava/pull/911#issue-159595222) PR, the `EnhancedAssertion,error` output was broken. I'm not sure if it's the best solution, but adding `equals` to `lib/assert.js`, we get the output fixed. Using your example, the output looks like:

 ```
6 failed


   1. assert.equal with message
   
    AssertionError: foo
    + expected - actual

    -3
    +4
          +4
        Test.fn (test.js:4:7)
        _combinedTickCallback (internal/process/next_tick.js:67:7)
        process._tickCallback (internal/process/next_tick.js:98:9)


   2. assert.equal without message
   
    AssertionError: 3 == 4
    + expected - actual

    -3
    +4
          +4
        Test.fn (test.js:8:7)
        _combinedTickCallback (internal/process/next_tick.js:67:7)
        process._tickCallback (internal/process/next_tick.js:98:9)


   3. t.deepEqual with message
   
    AssertionError: foo
    + expected - actual

    -3
    +4
          +4
        Test.fn (test.js:12:7)
        _combinedTickCallback (internal/process/next_tick.js:67:7)
        process._tickCallback (internal/process/next_tick.js:98:9)


   4. t.deepEqual without message
   
    AssertionError: foo
    + expected - actual

    -3
    +4
          +4
        Test.fn (test.js:16:7)
        _combinedTickCallback (internal/process/next_tick.js:67:7)
        process._tickCallback (internal/process/next_tick.js:98:9)


   5. t.true with message
   foo   
  t.true(3 === 4, 'foo')
                        
                        
  
      t.true(3 === 4, 'foo')
        Test.fn (test.js:20:5)
        _combinedTickCallback (internal/process/next_tick.js:67:7)
        process._tickCallback (internal/process/next_tick.js:98:9)


   6. t.true without message
   foo   
  t.true(3 === 4, 'foo')
                        
                        
  
      t.true(3 === 4, 'foo')
        Test.fn (test.js:24:5)
        _combinedTickCallback (internal/process/next_tick.js:67:7)
        process._tickCallback (internal/process/next_tick.js:98:9)

```

sindresorhus:
> I'm not sure if it's the best solution, but adding equals to lib/assert.js, we get the output fixed. 

Huh? Adding random things isn't a solution. We already have `t.is()` for the same.

> clarifying: i'm using your example, but calling t.equal instead of assert.equal

I've read through this whole discussion multiple times and I'm still confused about most things here including this clarification. If you can clearly explain what the issue is that's blocking this PR we might be able to help ;)

danilosampaio:
@sindresorhus feel free to discard this PR. Initially, i thought it was simple, but i'm stuck in it. Maybe, with a little more experience, i can contribute with AVA. Thanks for your help guys.

sindresorhus:
No worries :)

PR #774, opened 2016-04-23
Throwing synchronously inside a callback style test should fail it immediately.

### LINKED ISSUE #748: Throwing in callback-test is not reported and causes test to never end
```js
import test from 'ava';

test.cb(t => {
	throw new Error('unicorn');
	t.pass();
	t.end();
});
```

Reproduced with AVA 0.14.0 and latest [master](https://github.com/sindresorhus/ava/commit/3201b1b4ff80ff75f0e1c288ca7da22f92c9b814) with all reporters.

jamestalmage:
Fixes #748.

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -36,6 +36,7 @@ function Test(title, fn, contextRef, report) {
 	this.sync = true;
 	this.contextRef = contextRef;
 	this.report = report;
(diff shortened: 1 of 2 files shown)


sindresorhus:
LGTM

novemberborn:
:shipit:

PR #777, opened 2016-04-23
Use `not-so-shallow` module to improve `t.deepEqual()`

### LINKED ISSUE #772: deepEqual doesn't throw when comparing { a: false } and { a: 0 }.
## Description

It seems that deepEqual uses the module [only-shallow](https://github.com/othiym23/only-shallow). 

In their readme they state:

> Use loose equality (==) only for value types (non-objects). This is the biggest difference between only-shallow and deeper / deepest. only-shallow cares more about shape and contents than type. 

So that's probably why false and 0 are considered equal.

## Environment

- Node.js v4.2.4
- linux 3.19.0-58-generic

I cloned the current master and verified by adding the following test to test/assert.js.

```js
t.throws(function () {
  assert.deepEqual({a: 0}, {a: false});
});
```

sotojuan:
Fixes #772 by using [`not-so-shallow`](https://github.com/sotojuan/not-so-shallow) instead of `only-shallow`. `not-so-shallow` uses strict equality.

Issues with `not-so-shallow` should go in its repository—I'm having Travis errors installng `buffertools` on Node 4 and 5, but when I install it on a local machine it works.


test/assert.js
```diff
@@ -120,6 +120,10 @@ test('.not()', function (t) {
 });
 
 test('.deepEqual()', function (t) {
+	t.throws(function () {
+		assert.deepEqual({a: false}, {a: 0});
+	});
```
sindresorhus: I'd like to see some more structural tests.

sotojuan: Do you mean same values but different order?

sindresorhus:
> I'm having Travis errors installng buffertools on Node 4 and 5, but when I install it on a local machine it works.

Just drop the `buffertools` stuff from the code. Newer Node.js versions have `buffer.equals`, and for older we don't care. Would bet no one have `buffertools` installed anyways.

sotojuan:
Fixed the buffer stuff and added some more tests.

test/assert.js
sindresorhus: Sure, that, and maybe some tests from https://github.com/sotojuan/not-so-shallow Just to make sure we never regress, even if we were to replace your lib with something else any time in the future (hopefully we won't though).

mattkrick: A test to make sure we ignore constructor equality would be a good one. I
goofed up and never added that last time.

sotojuan:
Used `buffer-equals` in `not-so-shallow` and bumped the version in the PR. Will get to the tests later, lot of school work to do D:

sotojuan:
Added some random tests for arrays, functions, and objects.

vdemedes:
LGTM

test/assert.js
```diff
@@ -120,6 +120,64 @@ test('.not()', function (t) {
...
+		assert.deepEqual(x1, x2);
+	});
+
+	t.doesNotThrow(function () {
+		function Foo(a) {
+			this.a = a;
+		}
+
+		var x = new Foo(1);
+		var y = new Foo(1);
```
mattkrick:
```
function Bar(a) {
  this.a = a;
}
var y = new Bar(1);
```

test/assert.js
```diff
@@ -120,6 +120,64 @@ test('.not()', function (t) {
...
+		assert.deepEqual([1, 2, 3], [1, 2, 3]);
+	});
+
+	t.throws(function () {
+		assert.deepEqual([1, 2, 3], [1, 2, 3, 4]);
+	});
+
+	t.throws(function () {
+		var fnA = a => a;
+		var fnB = a => a;
```
novemberborn: AVA's tests aren't transpiled, so you can't use arrow functions I'm afraid.

sotojuan: Ah, missed that while copy/pasting.

sotojuan:
Fixed arrow function, added another constructor test.

vdemedes:
Tests are failing.

sotojuan:
Yep. Not sure why—can anyone help?

sindresorhus:
`let` is not available in Node.js 0.10-0.12. https://github.com/sotojuan/not-so-shallow/blob/b97d551d30f40401fd196ccf69afb74f155fcdac/index.js#L3

You also need to [test on older Node.js versions](https://github.com/sotojuan/not-so-shallow/pull/1).

sotojuan:
Yep, fixing now—I'm to used to typing `let`!

sotojuan:
`not-so-shallow` fixed for older Node versions.

test/assert.js
novemberborn: @sotojuan I think this is meant to test that `deepEqual` does not check the constructor?

test/assert.js
```diff
@@ -120,6 +120,83 @@ test('.not()', function (t) {
 });
 
 test('.deepEqual()', function (t) {
```
novemberborn: Maybe add a comment that these tests are here to detect regressions when the underlying comparison library changes?

sotojuan: Done.

sotojuan:
Added comments explaining the regression tests.

sindresorhus:
Looks good! Splendid work @sotojuan :)

highfive

flowmemo:
I am confused by ava's deepEqual and node's deepEqual.  
Before this PR was merged, ava's deepEqual is same as node's deepEqual. But now it is more like deepStrictEqual. Am I right.
Thanks.

jamestalmage:
Yes. Very similar to `deepStrictEqual`.

Issue #718, opened 2016-04-05
Implement pool for managing forks


novemberborn:
For #78 we'll end up limiting the number of concurrent forks. We should implement a pool of forks that the Api can use to run each test file.

Here's a rough sketch of a possible implementation:

``` js
import assert from 'assert'
import fork from './fork'

class Pool extends EventEmitter {
  constructor (limit) {
    this.limit = limit
    this.active = 0
  }

  fork (file, opts) {
    assert(this.active < this.limit)

    this.active++
    const promise = fork(file, opts)
    promise.catch(() => {}).then(() => {
      this.active--
      this.emit('free')
    })
    return promise
  }
}
```

novemberborn:
Currently blocked by #696.

jamestalmage:
I started [`async-task-pool`](https://github.com/jamestalmage/async-task-pool) with this in mind.

dcousineau:
We're chasing down huge memory usage in our project breaking CircleCI. With a pool we can certainly help clamp down on runaway peak usage (I'm still going to chase down other ideas).

I do notice there is The Blocker™: Would a default behavior of unlimited sized pool maintaining current behavior and adding a flag for limiting the pool with the caveat that test.only won't work be acceptable as an interim (beta) behavior?

If so would extra hands (read: my time) integrating @jamestalmage's async-task-pool be useful or has enough work been started that my time would be wasted effort?

jamestalmage:
@vdemedes brought up a good point that `async-task-pool` really is a poor substitute for Bluebird's concurrency `promise.map(handler, {concurrency: poolSize})`.

Note that would mean creating a `disposer`. http://bluebirdjs.com/docs/api/disposer.html

jamestalmage:
Sorry for the incomplete answer, got called away.

> Would a default behavior of unlimited sized pool maintaining current behavior and adding a flag for limiting the pool with the caveat that test.only won't work be acceptable as an interim (beta) behavior?

Yes.

> If so would extra hands (read: my time) integrating @jamestalmage's async-task-pool be useful

Yes, definitely. 

However, it's going to take a pretty deep dive into AVA's codebase, and it's a fairly complex issue. We're also going to have strong opinions on how it impacts the programmatic API. I'm not trying to scare you off helping (we would love it, really!). I just want you to know what you're getting into.

dcousineau:
[...]

PR #791, opened 2016-04-26
Limited Process Pools

dcousineau:
Functionality to address #718.

## Usage

~~Running `ava --pool-size=5`~~ Running `ava --concurrency=5` will switch behavior to the limited process pool code (otherwise existing 'legacy' code will always run). ~~If `--serial` is defined the limited pool code is _not_ run.~~ If `--serial` is specified with a concurrency size, the concurrency size will be overridden and forced to 1.

## Notes

I encapsulated existing functionality and left it default. We cannot do the exclusive/`.only` with the limited pools as it requires us to have loaded and understand the entire test suite to make the decision about whether there is exclusivity to respect or not.

As a _beta_ feature using the process pool limit could come with the caveat that any features related to `.only` are disabled/ignored. For our use case the pool limiting is being used to keep memory usage under circleci's limits and ideally no `.only` tests make it into source control.

### Easy Wins

The watcher worked right out of the box

### ~~Things that make me nervous~~

~~The error handling, particularly around `Couldn\'t find any matching tests`, feels a bit like I'm stumbling in the dark, they work and I'm not 100% confident I know why.~~

## Tests

Testing was done initially by wrapping almost all `api.js` tests in a generator, and running them twice: once against the old behavior and once against the new behavior

~~As a result **3 tests are currently failing.** This is because theres no real way to test exclusivity/deal with `.only`. These tests cannot pass until the fate of `.only` in relation to process pools is decided.~~

Tests related to `.only` functionality are not run against the new behavior.

## Personal Opinions

While ava could potentially statically analyze all the test files before entering in the process pool to look for exclusivity I'm concerned this is wasted effort that could be better spent deprecating .only and improving the filename globs.

Alternatively .only could be scoped to be intended to only apply to the current running file (and not to the whole testsuite) which would eliminate any need for static analysis and if a developer wanted to run only a single file they could simply use filename glob matching on a case by case basis.


dcousineau:
I suggest using the `?w=1` trick to view the diff for the test file changes. 

A link for the lazy: https://github.com/sindresorhus/ava/pull/791/files?w=1

dcousineau:
@jamestalmage ^

.gitignore
```diff
@@ -2,3 +2,4 @@ node_modules
 .nyc_output
 coverage
 bench/.results
+.idea
```
novemberborn: Please add a newline at the end of the file.

cli.js
```diff
@@ -71,6 +71,7 @@ var cli = meow([
 	'  --watch, -w      Re-run tests when tests and source files change',
 	'  --source, -S     Pattern to match source files so tests can be re-run (Can be repeated)',
 	'  --timeout, -T    Set global timeout',
+	'  --pool-size, -z  Set worker pool size (BETA)',
```
novemberborn: Perhaps we shouldn't document this option until it's solid. On the other hand that's what we tried with the watcher and without documenting it people won't find it.

novemberborn:
> If --serial is defined the limited pool code is not run.

It should use the pool but with a concurrency of `1`. Else AVA will still fork for each test file, even though it runs them serially.

I wonder if, rather than duplicating the code, we can use infinite concurrency when the pool limit is not set, and still wait for all stats unless there is limited concurrency?

dcousineau:
@novemberborn I "forked" the code because of the logic that initializes each test looking for a `.only` being incompatible with the structure of the pooled code. But I can see what happens with unlimited concurrency and see if the new code passes the 3 exclusive tests and report back.

dcousineau:
@novemberborn also concurrency of 1 makes more sense than me whining that serialMap doesn't allow concurrency :|

novemberborn:
> I "forked" the code because of the logic that initializes each test looking for a .only being incompatible with the structure of the pooled code.

Yea that makes sense. I'm proposing adding a logic branch for when concurrency is restricted, which I realize may be worse than duplicating code.

dcousineau:
> Yea that makes sense. I'm proposing adding a logic branch for when concurrency is restricted, which I realize may be worse than duplicating code.

If we either de-scope exclusive tests/`.only` to only apply to the tests in the current singular file, *or* we eliminate the feature entirely, we can drop the old logic and stick with the concurrency pool logic.

The more I think about previous suggestions of static analysis for `.only` support the more I get heartburn _but_ once thats implemented we can drop old behavior as well.

novemberborn:
Yea maybe we can live with the duplication until we sort out `.only`, especially if pool limiting is marketed as beta. (Which is odd, because AVA is still major-0, but hey). Would be good to unblock users with lots of test files.

cli.js
jamestalmage: I think it worked well with watcher. A limited pool of people used it for a while (mostly people who watch AVA closely and contribute), and we released it a short time later.

We could add a mention in the release notes and tweet about it to get more eyeballs on it, but I think leaving it undocumented gives us more freedom to make breaking changes. We are pre `1.0.0`, so it's not like breaking changes are a huge deal, but they are an annoyance. We have so many users now, I think it's a good idea to signal to them when they are playing on the bleeding edge (at least when we can).

.gitignore
sindresorhus: Don't add this. It should be in your own global gitignore, not here.

api.js
```diff
@@ -224,3 +236,85 @@ Api.prototype._run = function (files, _options) {
...
+	};
+
+	runStatus.on('timeout', function () {
+		Object.keys(tests).forEach(function (file) {
+			var fork = tests[file];
+			fork.exit();
+		});
+	});
+
+	return new Promise(function (resolve) {
```
sindresorhus: Would be nice to reduce the nesting below here ⬇️

cli.js
```diff
@@ -88,7 +89,8 @@ var cli = meow([
 		'require',
 		'timeout',
 		'source',
-		'match'
+		'match',
+		'pool-size'
```
sindresorhus: I think `concurrency` would be a better flag name than `pool-size`. It's the most common name for it that I've seen. For example: http://bluebirdjs.com/docs/api/promise.map.html#map-option-concurrency

sindresorhus:
Should we have a better default than Infinity? From experience, sometimes it's more efficient with 10 in concurrency than Infinity, since systems are bad at handling that much concurrency and get overloaded. We should maybe do some test runs and find some magic number as default.

cli.js
sindresorhus: I personally don't remember us getting any feedback on the watch trial. It was pretty much only us on the team using it. I think we should document it as `(EXPERIMENTAL)`. That way we're free to do whatever we want with it.

cli.js
dcousineau: Certainly. I was going off fabfile.org's as it was closest to my experiences. Ill change in the next push

api.js
dcousineau: I'll tinker and see what I can do

dcousineau:
@sindresorhus I'm more familiar with multi-process in Python and a quick check shows that Celery likes to default to the number of CPUs.

sindresorhus:
@dcousineau Last time I did concurrency profiling, I found the following to be optimal: 

```js
Math.max((os.cpus().length || 1) * 2, 2)
```

From: https://github.com/sindresorhus/grunt-concurrent/blob/9618f67c01e15ca447db5a36bd03bff88e40ce9d/tasks/concurrent.js#L13

dcousineau:
@sindresorhus I'm cool with that. As a part of my most recent update `Api.prototype._runLimitedPool` now uses the concurrency pool size via a function argument so I can support serial execution. I haven't bolted it in yet but the code is ready for it.

I noticed the conflicts and will have a resolution pushed soon

api.js
```diff
@@ -224,3 +236,86 @@ Api.prototype._run = function (files, _options) {
 		return runStatus;
 	});
 };
+
+Api._blankResults = {
```
sindresorhus: I think this should be a getter that returns the object. It might be an edge case, but if the consumer modifies the object, later results will also be modified. I might be missing something though and this might be a non-issue, but thought I would bring it up.

cli.js
```diff
@@ -60,17 +60,18 @@ var cli = meow([
...
+	'  --serial, -s      Run tests serially',
+	'  --require, -r     Module to preload (Can be repeated)',
+	'  --tap, -t         Generate TAP output',
+	'  --verbose, -v     Enable verbose output',
+	'  --no-cache        Disable the transpiler cache',
+	'  --match, -m       Only run tests with matching title (Can be repeated)',
+	'  --watch, -w       Re-run tests when tests and source files change',
+	'  --source, -S      Pattern to match source files so tests can be re-run (Can be repeated)',
+	'  --timeout, -T     Set global timeout',
+	'  --concurrency, -c Maximum number of child processes (EXPERIMENTAL)',
```
sindresorhus: Use two spaces to separate the flag and description columns

api.js
```diff
@@ -224,3 +236,86 @@ Api.prototype._run = function (files, _options) {
...
+			if (self.options.timeout) {
+				runStatus._restartTimer.cancel();
+			}
+
+			runStatus.processResults(results);
+			return runStatus;
+		});
+};
+
+Api.prototype._generateRunner = function (file, test, runStatus, cb) {
```
sindresorhus: Why are we using a callback `cb` here instead of a promise?

api.js
dcousineau: Everywhere that returns this object structure (but with real data) forms it themselves. On error we need to return an empty object so the `runStatus.processResults(results);` doesn't choke on an undefined so ideally no consumer gets the blank results structure. I only pulled it into its own variable because there are 3 failure cases that need to send it up stream. 

Writing this I realized I can simplify down to 2 and will be pushing a change to simplify as well as changing it to a getter Just In Case™

api.js
dcousineau: I renamed it cb but the the cb will always be the `resolve` from the promise above on line 268. 

`_generateRunner` is creating the callback for the `test.on('stats', ...)` event and this generated callback is what should eventually resolve the outer promise that represents the process in the concurrency pool. 

A good chunk of this is still a pull over from the old runner since I don't feel confident enough on the `self._runFile` logic and the fork objects to refactor so I just mimicked old behavior.

jamestalmage:
```js
Math.max((os.cpus().length || 1) * 2, 2)
```

Can be simplified to just:

```js
(os.cpus().length || 1) * 2
```

cli.js
```diff
@@ -60,17 +60,18 @@ var cli = meow([
...
+	'  --serial, -s       Run tests serially',
+	'  --require, -r      Module to preload (Can be repeated)',
+	'  --tap, -t          Generate TAP output',
+	'  --verbose, -v      Enable verbose output',
+	'  --no-cache         Disable the transpiler cache',
+	'  --match, -m        Only run tests with matching title (Can be repeated)',
+	'  --watch, -w        Re-run tests when tests and source files change',
+	'  --source, -S       Pattern to match source files so tests can be re-run (Can be repeated)',
+	'  --timeout, -T      Set global timeout',
+	'  --concurrency, -c  Maximum number of child processes (EXPERIMENTAL)',
```
sindresorhus: Maybe this?

> Maximum number of test files running at the same time (EXPERIMENTAL)

sindresorhus: Can you also sync this output into the readme? Just copy paste `node cli.js --help`.

readme.md
```diff
@@ -132,21 +132,24 @@ AVA comes with an intelligent watch mode. [Learn more in its recipe](docs/recipe
 ```console
 $ ava --help
 
+  Futuristic test runner 🚀
+
```
sindresorhus: This is intentionally left out as it just duplicates our readme and repo description.

sindresorhus:
Tests are failing

dcousineau:
@sindresorhus tests are currently failing on the 3 `.only` test cases (since the entire API test case is run against the concurrency limiting code). There are a few options:

1. For now we do not run the 3 test cases against the concurrency limit code while we decide what to do with this feature going forward.

2. We wait until we've decided to either deprecate `.only` or reduce the scope of `.only` (only to current file) or we make another decision.

I can easily implement suggestion 1. 

My long term suggestion is either deprecating or reducing the scope of `.only`.

I'll wait for your positive confirmation on how you'd prefer to proceed.

api.js
```diff
@@ -224,3 +236,87 @@ Api.prototype._run = function (files, _options) {
 		return runStatus;
 	});
 };
+
+var getBlankResults = function () {
```
novemberborn: Should also use this in `_runNoPool`.

api.js
```diff
@@ -116,6 +116,18 @@ Api.prototype._run = function (files, _options) {
 	self.precompiler = new CachingPrecompiler(cacheDir, self.options.babelConfig);
 	self.fileCount = files.length;
 
+	var overwatch;
+	if (this.options.concurrency > 0) {
+		overwatch = this._runLimitedPool(files, runStatus, self.options.serial ? 1 : this.options.concurrency);
+	} else {
+		overwatch = this._runNoPool(files, runStatus);
```
novemberborn: Maybe add a comment that `_runNoPool` is used to preserve `.only` behavior?

api.js
```diff
@@ -224,3 +236,87 @@ Api.prototype._run = function (files, _options) {
...
+
+			runStatus.processResults(results);
+			return runStatus;
+		});
+};
+
+Api.prototype._generateRunner = function (file, test, runStatus, cb) {
+	var self = this;
+	return function () {
+		self.emit('ready');
```
novemberborn: This probably shouldn't be emitted for every fork. Doesn't look like it's used for anything though.

dcousineau: Yeah I wasn't sure what it was for so I played it safe and kept it in the same area.

api.js
```diff
@@ -224,3 +236,87 @@ Api.prototype._run = function (files, _options) {
...
+		var options = {};
+		test.run(options)
+			.then(cb)
+			.catch(function (err) {
+				if (self.options.match.length > 0 && !runStatus.hasExclusive) {
+					runStatus.handleExceptions({
+						exception: new AvaError('Couldn\'t find any matching tests'),
```
novemberborn: `--match` only runs test with matching titles. Each fork will emit `hasExclusive` if it contains matching tests.

The logic here will give false warnings until a file is run that contains a matching test. Instead this exception should be handled once *all* forks have run and there were no matching tests.

api.js
```diff
@@ -224,3 +236,87 @@ Api.prototype._run = function (files, _options) {
...
+			fork.exit();
+		});
+	});
+
+	return Promise.map(files, function (file) {
+		try {
+			var test = tests[file] = self._runFile(file, runStatus);
+
+			return new Promise(function (resolve) {
+				var runner = self._generateRunner(file, test, runStatus, resolve);
```
novemberborn: Currently `_generateRunner` does a bit too much. I think once it's simplified it'll just fit here instead.

novemberborn:
> tests are currently failing on the 3 .only test cases (since the entire API test case is run against the concurrency limiting code). There are a few options:
> 
> For now we do not run the 3 test cases against the concurrency limit code while we decide what to do with this feature going forward.
> 
> We wait until we've decided to either deprecate .only or reduce the scope of .only (only to current file) or we make another decision.
> 
> I can easily implement suggestion 1.

Suggestion 1 is good for now. Our primary objective is to unblock CI runs that fail due to the number of processes being spawn. Since this is an experimental opt-in feature we don't have to worry too much about `.only` not working.

dcousineau:
@novemberborn Sounds good. I'll swing through tonight and get everything ready to go

dcousineau:
@sindresorhus @novemberborn Sorry about the delay, made the updates.

dcousineau:
I found an issue with the `using --match ...` tests where the `test/fixture/match-no-match.js` was actually failing trying to import ava instead of `../../`. I repaired the test and added a positive test case and an additional fixture.

The updated and new tests can be found here since the diff tab isn't showing them: https://github.com/dcousineau/ava/blob/process-pool/test/api.js#L920-L967

api.js
```diff
@@ -116,6 +116,19 @@ Api.prototype._run = function (files, _options) {
 	self.precompiler = new CachingPrecompiler(cacheDir, self.options.babelConfig);
 	self.fileCount = files.length;
 
+	var overwatch;
+	if (this.options.concurrency > 0) {
```
jamestalmage: This should still use `_runNoPool` if concurrency is less than the fileCount.

api.js
```diff
@@ -224,3 +229,85 @@ Api.prototype._run = function (files, _options) {
...
+						.catch(reject);
+				};
+
+				test.on('stats', runner);
+				test.on('exit', function () {
+					delete tests[file];
+				});
+				test.catch(runner);
+			}).catch(function (err) {
+				runStatus.handleExceptions({
```
jamestalmage: The contents of this and the catch block below are identical.

Hoist this function, pass it in directly here, and call it from the catch block.

jamestalmage:
GH isn't letting me comment directly on `test/api.js` because the diff is too huge.

https://github.com/dcousineau/ava/blob/process-pool/test/api.js#L920

Does this test cover the scenario where we have more files than concurrency? We would need tests for:

- No match anywhere
- No match in the first batch, match in a latter batch
- Match in the first batch, no match in a latter batch

jamestalmage:
Similar here - https://github.com/dcousineau/ava/blob/process-pool/test/api.js#L769 

Do we need a test that verifies the test counts are added up correctly across batched runs?

api.js
```diff
@@ -224,3 +229,85 @@ Api.prototype._run = function (files, _options) {
...
+			return new Promise(function (resolve, reject) {
+				var runner = function () {
+					self.emit('ready');
+					var options = {
+						// If we're looking for matches, run every single test process in exclusive-only mode
+						runOnlyExclusive: self.options.match.length > 0
+					};
+					test.run(options)
+						.then(resolve)
+						.catch(reject);
```
novemberborn: `.then(resolve, reject)`. `resolve` won't throw, so we can save the extra promise allocation.

api.js
```diff
@@ -224,3 +229,85 @@ Api.prototype._run = function (files, _options) {
...
+		} catch (err) {
+			runStatus.handleExceptions({
+				exception: err,
+				file: path.relative('.', file)
+			});
+		}
+	}, {concurrency: concurrency})
+		.then(function (results) {
+			results = results.filter(function (result) {
+				return Boolean(result); // Filter out undefined results (usually result of caught exceptions)
```
novemberborn: I'd be happy with `results.filter(Boolean)`, not sure if there's an official stance on that @sindresorhus?

api.js
```diff
@@ -224,3 +229,85 @@ Api.prototype._run = function (files, _options) {
...
+				runStatus._restartTimer.cancel();
+			}
+
+			if (self.options.match.length > 0 && !runStatus.hasExclusive) {
+				runStatus.handleExceptions({
+					exception: new AvaError('Couldn\'t find any matching tests'),
+					file: undefined
+				});
+				runStatus.processResults([]);
+				return runStatus;
```
novemberborn: Seems clearer to reassign `results = []`, rather than repeat the `processResults` call and `return` statement.

novemberborn:
@sindresorhus @jamestalmage @vdemedes what shall we do about the `ready` callback? Best I can tell it's unused so it should just be removed. With this PR it is emitted for every fork, rather than once.

api.js
jamestalmage: Meh, I prefer it the existing way. Two argument `then` is considered an anti pattern now.

api.js
sindresorhus: Sure

api.js
dcousineau: Normally I would agree but with the current behavior, no matter what if the `--concurrency` flag is used the code behind `_runLimitedPool` is used.

dcousineau:
@novemberborn looking over the older code I see how `self.emit('ready')` is emitted only once which I missed.

I can move it above the `Promise.map` easy, and the tests pass. I'll wait for a final word.

dcousineau:
@jamestalmage Dropping the concurrency pool to 1 and re-running the tests ensured they still pass. I'm checking for matches after all test files have run so it shouldn't be an issue but I can circle back and come up with a few more tests

jamestalmage:
Let's kill `ready` in a separate PR.

If you can "circle back" with a few more tests, that would be great. We want to guard against regressions.

api.js
jamestalmage: But if I add a `concurrency` option to `package.json`, I want to be able to re-enable `.only` behavior. By doing 

`ava [files ...]`

Where the number of files is less than concurrency.

dcousineau: I'm similarly worried about people using the `--concurrency` flag and seeing different behaviors based on their filename glob patterns. It's much simpler to convey in documentation "If the concurrency flag/option is present, automatic .only behavior is not supported" than a list of rules.

Additionally, a user manually running ava with a deliberate subset of tests is likely only doing so to run a few specific tests because they internally do not use `.only` (like my team) which actually cycles back up to my argument of nixing `.only` entirely since with the `--watch` flag and good file name globbing, there is no real need for `.only`.

sindresorhus:
LGTM

sindresorhus:
🎉 Woo! Thanks for your perseverance on this @dcousineau :)

jamestalmage:
Well done!

sotojuan:
Great work @dcousineau! By the way I sat in front of you at ManhattanJS yesterday :p

dcousineau:
@sotojuan Ha! Next time grab me and say hi! (I'm at all of them). Sidebar: have you considered reaching out for [EmpireJS OSS Night](http://2016.empirejs.org/#oss-night)?

sotojuan:
@dcousineau If I am still in NYC by then I'll go. Thanks!

Issue #723, opened 2016-04-05
Improve TAP Output


jamestalmage:
There are lots of open issues regarding TAP output:
- [x] #679 - TAP report prints tests titles above not ok message
- [ ] #399 - power-assert graph not shown in TAP output.
- [ ] #324 - Include individual assertions in the TAP output.
- [ ] #340 - Capture good assertion messages.
- [ ] #648 - include line number of passing tests in TAP output

Remaining todo's from https://github.com/sindresorhus/ava/pull/340#issuecomment-177008691
- [ ] Bring our tap output more inline with `node-tap`. Specifically, the yaml output for our errors. This should make the output from all the tap-mocha-reporters much nicer. For example, our errors currently have `expected` and `actual` properties, but the tap reporters prefer found and wanted. Also, we should use `stack-utils` to attach the `at` property (file name and line number of failure).
- [ ] Finish up work on `tap-emitter` and integrate into AVA. 
## 
### Notes:

We should be using the [TAP v14 draft specification](https://github.com/TestAnything/testanything.github.io/pull/36).

Note that the spec does not address what properties should be included in the [yaml blocks](https://github.com/isaacs/testanything.github.io/blob/tap14/tap-version-14-specification.md#yaml-blocks). In that case we should just emulate `node-tap` (see list item above for examples of properties we need to add).

xjamundx:
Probably related to this is the fact that `ava tests/ --tap | tap-xunit > ava-results.xml` produces some invalid XML characters.

```
<testsuite name="calculator ESC[90mESC[2m›ESC[22mESC[39m actions ESC[90mESC[2m›ESC[22mESC[39m __tests__ ESC[90mESC[2m›ESC[22mESC[39m Actions.fetchOriginAmount ESC[90mESC[2m›ESC[22mESC[39m with no changes a noop function should be returned" tests="1" failures="0" errors="0">
```

sotojuan:
@xjamundx Can you post the actual tests that are giving you the strange XML output? I just tried `tap-xunit` with some simple tests and it worked fine.

xjamundx:
@sotojuan it needs to be in a sub-directory for example

```
test/test.js
test/horse/test.js
```

~/dev/paypal/p2pnodeweb (junit) $ ./node_modules/.bin/ava test  --tap | ./node_modules/.bin/tap-xunit

The issue (from what I can tell is the little  › arrows)

```
<?xml version="1.0"?>
<testsuites>
[...]

PR #792, opened 2016-04-26
Strip ANSI escape codes in TAP test titles

sotojuan:
Fixes issue mentioned by @xjamundx in #723.

### DIFF
--- a/lib/reporters/tap.js
+++ b/lib/reporters/tap.js
@@ -1,5 +1,6 @@
 'use strict';
 var format = require('util').format;
+var stripANSI = require('strip-ansi');
 
 // Parses stack trace and extracts original function name, file name and line.
 function getSourceFromStack(stack, index) {
@@ -36,10 +37,12 @@ TapReporter.prototype.test = function (test) {
 		directive = '# SKIP';
 	}
 
+	var strippedTitle = stripANSI(test.title);
+
 	if (test.error) {
 		output = [
-			'# ' + test.title,
-			format('not ok %d - %s', ++this.i, test.title),
+			'# ' + strippedTitle,
+			format('not ok %d - %s', ++this.i, strippedTitle),
 			'  ---',
 			'    operator: ' + test.error.operator,
 			'    expected: ' + test.error.expected,
@@ -49,8 +52,8 @@ TapReporter.prototype.test = function (test) {
 		];
 	} else {
 		output = [
-			'# ' + test.title,
-			format('%s %d - %s %s', passed, ++this.i, test.title, directive).trim()
+			'# ' + strippedTitle,
+			format('%s %d - %s %s', passed, ++this.i, strippedTitle, directive).trim()
 		];
 	}
 
--- a/lib/reporters/tap.js
+++ b/lib/reporters/tap.js
@@ -1,6 +1,6 @@
 'use strict';
 var format = require('util').format;
-var stripANSI = require('strip-ansi');
+var stripAnsi = require('strip-ansi');
 
 // Parses stack trace and extracts original function name, file name and line.
 function getSourceFromStack(stack, index) {
@@ -37,12 +37,12 @@ TapReporter.prototype.test = function (test) {
 		directive = '# SKIP';
 	}
 
-	var strippedTitle = stripANSI(test.title);
+	var title = stripAnsi(test.title);
 
 	if (test.error) {
 		output = [
-			'# ' + strippedTitle,
-			format('not ok %d - %s', ++this.i, strippedTitle),
+			'# ' + title,
+			format('not ok %d - %s', ++this.i, title),
 			'  ---',
 			'    operator: ' + test.error.operator,
 			'    expected: ' + test.error.expected,
@@ -52,8 +52,8 @@ TapReporter.prototype.test = function (test) {
 		];
 	} else {
 		output = [
-			'# ' + strippedTitle,
-			format('%s %d - %s %s', passed, ++this.i, strippedTitle, directive).trim()
+			'# ' + title,
+			format('%s %d - %s %s', passed, ++this.i, title, directive).trim()
 		];
 	}
 
--- a/test/reporters/tap.js
+++ b/test/reporters/tap.js
@@ -190,7 +190,7 @@ test('reporter strips ANSI characters', function (t) {
 	  file: 'test.js'
 	};
 
-	var output = reporter.test(test)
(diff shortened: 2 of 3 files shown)


lib/reporters/tap.js
```diff
@@ -1,5 +1,6 @@
 'use strict';
 var format = require('util').format;
+var stripANSI = require('strip-ansi');
```
novemberborn: Human linting nitpick, I'd prefer `stripAnsi` actually 😉

lib/reporters/tap.js
```diff
@@ -36,10 +37,12 @@ TapReporter.prototype.test = function (test) {
 		directive = '# SKIP';
 	}
 
+	var strippedTitle = stripANSI(test.title);
```
novemberborn: `title` should suffice.

novemberborn:
Just nits, looks good. Haven't tried it though, would be good to add a test case.

sotojuan:
The nipicks are fine, but not sure about testing because I don't think ANSI escape codes have been an issue or come up until the issue that led to this PR. That said, I have tested it locally and it works, but if anyone has a good idea of a test let me know.

novemberborn:
> not sure about testing because I don't think ANSI escape codes have been an issue or come up until the issue that led to this PR

Sure, but it'd be good to have a test to protect against future regressions.

> if anyone has a good idea of a test let me know

Could add to [the TAP reporter tests](https://github.com/sindresorhus/ava/blob/master/test/reporters/tap.js), one "passing" with a title that contains escape codes, and one "failing".

sindresorhus:
Hmm. Thinking out loud. Are we attacking this from the wrong side? Maybe we should color the test title in the reporters that needs it, not remove it here?

sotojuan:
@novemberborn The ANSI codes show up when you test more than one file—is there a way to simulate that on the TAP tests?

novemberborn:
> The ANSI codes show up when you test more than one file—is there a way to simulate that on the TAP tests?

@sotojuan I don't think that matters, the test calls `reporter.test()` directly.

> Are we attacking this from the wrong side? Maybe we should color the test title in the reporters that needs it, not remove it here?

Seems sensible to color when the test data is received, rather than in different reporters. Though admittedly there's more coloring going on in the reporters themselves. Maybe all of that can be improved in a future refactor?

jamestalmage:
> Hmm. Thinking out loud. Are we attacking this from the wrong side? Maybe we should color the test title in the reporters that needs it, not remove it here?

Maybe, however, removing it here might be a good idea if TAP hates ansi codes in the test title.

I was just writing a test macro:

```js
function macro(obj, expected) {
  test(util.describe(obj, {colors: true}), t => {
    t.is(fn(obj), expected);
  });
}

macro({someObj}, 'foo');
macro({someOtherObj}, 'bar');
```

I thought the output with the verbose reporter looked awesome:

screenshot 2016-04-27 12 42 33

Maybe allowing this causes more headaches than it is worth, but I thought it was pretty cool!

sotojuan:
I would say remove it just from TAP like I am doing now—seems like TAP stuff in general doesn't really like coloring in the raw reporter.

sindresorhus:
@sotojuan Can you add `Fixes #00` (replace 00 with issue number) to the PR description so the relevant issue is closed when this is merged.

sotojuan:
Thing is this was mentioned in issue (#723).

sindresorhus:
Oh, right, then include just the issue number in the description. The more references the easier it is to navigate this at a later point.

sindresorhus:
I don't think there's much benefit of testing the title alone. `stripAnsi` is already tested and all we do is to apply it. We should rather test the whole `tap` output to ensure there's no ANSI escape codes in any of the outputted information, not just the title.

https://github.com/chalk/has-ansi

novemberborn:
> We should rather test the whole tap output to ensure there are no ANSI escape codes in any of the outputted information, not just the title.

👍

sotojuan:
Added a test—let me know if it can be improved.

test/reporters/tap.js
```diff
@@ -179,3 +180,18 @@ test('skip test', function (t) {
 	t.is(actualOutput, expectedOutput);
 	t.end();
 });
+
+test('reporter strips ANSI characters', function (t) {
+	var reporter = tapReporter();
+
+	var test = {
```
sindresorhus: Just put this inline in the `reporter.test` call.

test/reporters/tap.js
```diff
@@ -179,3 +180,18 @@ test('skip test', function (t) {
...
+		title: 'test \u001b[90m\u001b[2m›\u001b[22m\u001b[39m my test',
```
sindresorhus: Use `chalk` to colorize it. Makes it easier if we ever want to change the style.

sotojuan:
Done!

test/reporters/tap.js
```diff
@@ -179,3 +181,16 @@ test('skip test', function (t) {
 	t.is(actualOutput, expectedOutput);
 	t.end();
 });
+
+test('reporter strips ANSI characters', function (t) {
+	var reporter = tapReporter();
+
+	var output = reporter.test({
+		title: chalk.gray.dim('test \u001b[90m\u001b[2m›\u001b[22m\u001b[39m my test'),
```
sindresorhus: Should have been clearer. I meant instead of the ANSI escape codes.

```js
'test ' + chalk.gray.dim('›') + ' my test')
```

sotojuan: Ha, no problem.

test/reporters/tap.js
```diff
@@ -179,3 +181,16 @@ test('skip test', function (t) {
 	t.is(actualOutput, expectedOutput);
 	t.end();
 });
+
+test('reporter strips ANSI characters', function (t) {
+	var reporter = tapReporter();
+
+	var output = reporter.test({
+		title: chalk.gray.dim('test ' + chalk.gray.dim('›') + ' my test'),
```
sindresorhus: The outer `chalk.gray.dim(` can be removed now.

sotojuan: Ahhh, keep missing silly stuff, my bad.

sotojuan:
All right, sorry about that (got a ton of things going on right now), should be fine now I hope 😎

sindresorhus:
No worries. Looks good. 👍

Now go back to your studies. That's more important ;)

PR #800, opened 2016-04-28
Use Babel's compact mode. Fixes #790

### LINKED ISSUE #790: Use Babels Compact Mode
See babel/babel#3475.

Should we set {compact: true; comments: false} in our Babel usage? 

Seems it could improve the speed a bit.

jfmengels:
Fixes #790.
Add new options to Babel: `compact: true` and `comment: false`.

I had to fix some tests, as the whitespace in the tested output was removed. (I thought about adding a `?` to make the whitespace optional, but this is probably better I think)

I ran the bench tests, but I don't feel like there's a lot of change performance-wise (both tests run in ~5 minutes on my machine). Is this part of what the bench tests are testing?

Note that it changes the error output from
image
to
image
(though in this case, I prefer the new version :smile:)

### DIFF
--- a/lib/caching-precompiler.js
+++ b/lib/caching-precompiler.js
@@ -108,7 +108,9 @@ CachingPrecompiler.prototype._buildOptions = function (filePath, code) {
 		inputSourceMap: sourceMap,
 		filename: filePath,
 		sourceMaps: true,
-		ast: false
+		ast: false,
+		compact: true,
+		comments: false
 	});
 
 	options.plugins = (options.plugins || []).concat(this.defaultPlugins);
--- a/test/cli.js
+++ b/test/cli.js
@@ -95,7 +95,7 @@ test('timeout', function (t) {
 test('throwing a named function will report the to the console', function (t) {
 	execCli('fixture/throw-named-function.js', function (err, stdout, stderr) {
 		t.ok(err);
-		t.match(stderr, /function fooFn\(\) \{\}/);
+		t.match(stderr, /function fooFn\(\)\{\}/);
 		// TODO(jamestalmage)
 		// t.ok(/1 uncaught exception[^s]/.test(stdout));
 		t.end();
@@ -140,7 +140,7 @@ test('babel require hook only applies to the test file', function (t) {
 test('throwing a anonymous function will report the function to the console', function (t) {
 	execCli('fixture/throw-anonymous-function.js', function (err, stdout, stderr) {
 		t.ok(err);
-		t.match(stderr, /function \(\) \{\}/);
+		t.match(stderr, /function \(\)\{\}/);
 		// TODO(jamestalmage)
 		// t.ok(/1 uncaught exception[^s]/.test(stdout));
 		t.end();


jamestalmage:
You need to run the bench tests on both branches (this one and master). Then run `bench/compare`

jamestalmage:
@jfmengels - I don't think you pushed the fixed tests yet!

jfmengels:
> I don't think you pushed the fixed tests yet!

Ugh, you're right. Done.

> You need to run the bench tests on both branches (this one and master). Then run node bench/compare

Ah yes, found the section about benchmarking in the `maintaining` file.

Benchmark results on my machine:
image

jamestalmage:
Yeah, those numbers certainly aren't very enticing

jfmengels:
Re-ran it twice just to be sure
image

image

And it looks better already, especially in the first case. Don't know if this is by chance (the gain is always pretty slim), or what happened during the first run. Maybe I had done the run in this branch before I applied the changes (can't remember, sorry for the confusion...), or maybe the gains are so redicule that it willt tilt one way or another if I'm doing something else on my computer while running the benchmark test.

jamestalmage:
> or maybe the gains are so small that it willt tilt one way or another if I'm doing something else on my computer while running the benchmark test

Almost certainly. I was surprised how much just browsing GH issues on Chrome would affect benchmark results.

jamestalmage:
I guess I'm 👎 for doing this. Gains are very minimal, likely because we cache so aggressively.

nfcampos:
are you sure you're running the benchmark with babel@6.7.7? That was the version that sped up `compact` mode a lot https://github.com/babel/babel/releases/tag/v6.7.7

jfmengels:
Well, one thing to take into account is the size of the test files. They are really minuscule, and I think that each pass only transpiles the currently tested file, so I'm not sure that Babel has a chance to shine here.

Looking at it more, I don't think that the transpiled output is any different than the given input (no assertions to be changed with power-assert, no `const`, no `async`...).

I think we should run the test on something pretty big LOC-wis and that needs a lot of transpilation, in order to really mesure Babel's influence. If the results look kind of the same, I would probably agree with @jamestalmage and give it a :-1: (if the performance is confirmed by other computers).

> are you sure you're running the benchmark with babel@6.7.7?

The version of `babel-core` I find in `node_modules/babel-core` is indeed `6.7.7`.

nfcampos:
and `babel-generator` and `babel-types` versions? the release notes mention the speed up being in those two packages

jamestalmage:
> I think we should run the test on something pretty big LOC-wis and that needs a lot of transpilation, in order to really mesure Babel's influence

The issue also becomes, how much of a performance gain is worth the modified code frames for errors? I don't love that we reformat what people see on their screen. 

I think we cache babel transpilation results more aggressively than babel itself does, so even with big codebases, it is unlikely to make much of a difference except for the initial run.

sindresorhus:
Alright. Let's skip this then.

PR #804, opened 2016-05-03
add podcast episode to links

kentcdodds:
Thought people may find this chat helpful.


jfmengels:
Might be worth putting in https://github.com/sindresorhus/awesome-ava also/instead.
Maybe add a mention that it's an episode with the team behind AVA?

jamestalmage:
That guy in the plaid keeps mispronouncing the project name and totally butchers contributor names. 😈

sindresorhus:
Let's have it in both for now. @kentcdodds Can you submit to https://github.com/sindresorhus/awesome-ava too?

PR #806, opened 2016-05-03
add `after.always` and `afterEach.always` hooks - fixes #474

### LINKED ISSUE #474: Add chainable `.always` method
Implement chainable `always` method.

From [@jamestalmage's comment](https://github.com/sindresorhus/ava/issues/474#issuecomment-174607816) below:

```
What if it's always, and chainable:

test.always - even if a beforeEach fails.
test.afterEach.always - even if a test fails.
test.after.always - the OPs request.

Not sure the first one makes sense.
```

---
*Original message*

Near as I can figure, the `.after()` hook is only being ran if the tests have passed. I have a test file something like this (simplified/pseudo for clarity)

```js
// ... imports

test('create a file', ...) // passes

test('do something else', ...) // fails

// ... other tests

test.after('clean up created files', ...) // never runs
```

cgcgbcbc:
This pull request aims at #474 .
The new tests added in `test/hooks.js` passed, but these changes lead to other tests' failure.
Still working on figuring out why, comments are welcome!

### DIFF
--- a/lib/test-collection.js
+++ b/lib/test-collection.js
@@ -61,7 +61,7 @@ TestCollection.prototype.add = function (test) {
 	}
 
 	if (metadata.always && type !== 'after' && type !== 'afterEach') {
-		    throw new Error('"always" can only be used with after and afterEach hooks');
+		throw new Error('"always" can only be used with after and afterEach hooks');
 	}
 
 	// add a hook
--- a/lib/test-collection.js
+++ b/lib/test-collection.js
@@ -67,7 +67,7 @@ TestCollection.prototype.add = function (test) {
 	// add a hook
 	if (type !== 'test') {
 		if (metadata.exclusive) {
-			throw new Error('"only" cannot be used with a ' + type + ' test');
+			throw new Error('"only" cannot be used with a ' + type + ' hook');
 		}
 
 		this.hooks[type + (metadata.always ? 'Always' : '')].push(test);
--- a/lib/test-collection.js
+++ b/lib/test-collection.js
@@ -60,21 +60,17 @@ TestCollection.prototype.add = function (test) {
 		}
 	}
 
+	if (metadata.always && type !== 'after' && type !== 'afterEach') {
+		    throw new Error('"always" can only be used with after and afterEach hooks');
+	}
+
 	// add a hook
 	if (type !== 'test') {
 		if (metadata.exclusive) {
 			throw new Error('"only" cannot be used with a ' + type + ' test');
 		}
 
-		// add a always hook
-		if (metadata.always) {
-			if (type !== 'after' && type !== 'afterEach') {
-				throw new Error('"always" cannot be used with a ' + type + ' test');
-			}
-			this.hooks[type + 'Always'].push(test);
-		} else {
-			this.hooks[type].push(test);
-		}
+		this.hooks[type + (metadata.always ? 'Always' : '')].push(test);
 		return;
 	}
 
--- a/lib/test-collection.js
+++ b/lib/test-collection.js
@@ -25,7 +25,9 @@ function TestCollection() {
 		before: [],
 		beforeEach: [],
 		after: [],
-		afterEach: []
+		afterAlways: [],
+		afterEach: [],
+		afterEachAlways: []
 	};
 
 	this._emitTestResult = this._emitTestResult.bind(this);
@@ -64,7 +66,15 @@ TestCollection.prototype.add = function (test) {
 			throw new Error('"only" cannot be used with a ' + type + ' test');
 		}
 
-		this.hooks[type].push(test);
+		// add a always hook
+		if (metadata.always) {
+			if (type !== 'after' && type !== 'afterEach') {
+				throw new Error('"always" cannot be used with a ' + type + ' test');
+			}
+			this.hooks[type + 'Always'].push(test);
+		} else {
+			this.hooks[type].push(test);
+		}
 		return;
 	}
 
@@ -151,7 +161,7 @@ TestCollection.prototype._buildTest = function (test, context) {
 
 TestCollection.prototype._buildTestWithHooks = function (test) {
 	if (test.metadata.skipped) {
-		return [this._skippedTest(this._buildTest(test))];
+		return new Sequence([this._skippedTest(this._buildTest(test))], true);
 	}
 
 	var context = {context: {}};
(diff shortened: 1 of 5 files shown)


lib/test-collection.js
```diff
@@ -159,12 +169,17 @@ TestCollection.prototype._buildTestWithHooks = function (test) {
...
+		var afterAlwaysHooks = new Sequence(this._buildHooks(this.hooks.afterEachAlways, test.title, context));
+		sequence = new Sequence([sequence, afterAlwaysHooks], false);
+	}
+	return sequence;
 };
 
 TestCollection.prototype._buildTests = function (tests) {
 	return tests.map(function (test) {
-		return new Sequence(this._buildTestWithHooks(test), true);
+		return this._buildTestWithHooks(test);
```
cgcgbcbc: seems that with only the change that move the new Sequence to _buildTestWithHooks, tests will failed

test/hooks.js
```diff
@@ -223,6 +265,86 @@ test('after each with serial tests', function (t) {
...
+		throw new Error('something went wrong');
+	});
+
+	runner.run({}).then(function () {
+		t.strictDeepEqual(arr, []);
+		t.end();
+	});
+});
+
+test('after each not run if serial tests failed', function (t) {
```
jamestalmage: `after each` => `afterEach`

test/hooks.js
```diff
@@ -223,6 +265,86 @@ test('after each with serial tests', function (t) {
 	});
 });
 
+test('after each not run if concurrent tests failed', function (t) {
```
jamestalmage: `after each` => `afterEach`

jamestalmage:
We should add a test verifying what happens with `afterEach.always` if a `beforeEach` fails.

I'm not even sure what that should be!

cgcgbcbc:
@jamestalmage I've added tests verifying that `after.always`, `afterEach.always` run if `before` or `beforeEach` fails.

jamestalmage:
Just FYI, we use the new GitHub "squash and merge" feature on almost every PR, so you don't need to self squash or use commit --amend, etc. It actually makes it harder for me to review just the changes you've made in response to my comments. (Not a big deal, though).

lib/test-collection.js
```diff
@@ -64,7 +66,15 @@ TestCollection.prototype.add = function (test) {
 			throw new Error('"only" cannot be used with a ' + type + ' test');
 		}
 
-		this.hooks[type].push(test);
+		// add a always hook
+		if (metadata.always) {
+			if (type !== 'after' && type !== 'afterEach') {
+				throw new Error('"always" cannot be used with a ' + type + ' test');
```
jamestalmage: @cgcgbcbc - Can you add a few tests in `test/test-collection.js` for this throwing behavior?

lib/test-collection.js
```diff
@@ -64,7 +66,15 @@ TestCollection.prototype.add = function (test) {
...
+			}
+			this.hooks[type + 'Always'].push(test);
```
jamestalmage: Also, a few tests in `test/test-collection.js` verifying new tests are added correctly (this will involve a modification to the `serialize` function in that test).

jamestalmage:
@cgcgbcbc - You are real close. The implementation looks good to me, we just need a few more tests.

👍

cgcgbcbc:
@jamestalmage Tests added.

The `after.always` and `alfterEach.always` will always execute after `after` hook and `afterEach` hook respectively , no matter the order they are added, since they are implemented by wrapping bail sequence with non-bail one.

jamestalmage:
> The after.always and afterEach.always will always execute after after hook and afterEach hook respectively

I don't think that's a problem.

readme.md
```diff
@@ -418,9 +418,9 @@ test.todo('will think about writing this later');
 
 AVA lets you register hooks that are run before and after your tests. This allows you to run setup and/or teardown code.
 
-`test.before()` registers a hook to be run before the first test in your test file. Similarly `test.after()` registers a hook to be run after the last test.
+`test.before()` registers a hook to be run before the first test in your test file. Similarly `test.after()` registers a hook to be run after the last test. `test.after.always()` registers a hook to be run **always** after the last test, even if before hook/some tests fail.
```
novemberborn: How about:

> Use `test.after.always()` to register an after hook that is called even if tests or other hooks fail.

readme.md
```diff
@@ -418,9 +418,9 @@ test.todo('will think about writing this later');
...
 
-`test.beforeEach()` registers a hook to be run before each test in your test file. Similarly `test.afterEach()` a hook to be run after each test.
+`test.beforeEach()` registers a hook to be run before each test in your test file. Similarly `test.afterEach()` a hook to be run after each test. `test.afterEach.always()` registers a hook to be run **always** after each test, even if beforeEach hook/the test fails.
```
novemberborn: How about:

> Use `test.afterEach.always()` to register an after hook that is called even if other test hooks, or the test itself, fail.

readme.md
```diff
@@ -439,6 +439,10 @@ test.after('cleanup', t => {
 	// this runs after all tests
 });
 
+test.after.always('always cleanup', t => {
```
novemberborn: `guaranteed cleanup`?

readme.md
```diff
@@ -439,6 +439,10 @@ test.after('cleanup', t => {
...
+	// this runs always after all tests
```
novemberborn: Perhaps:

> this runs after all tests and other hooks, even if they failed

readme.md
```diff
@@ -447,6 +451,10 @@ test.afterEach(t => {
 	// this runs after each test
 });
 
+test.afterEach.always(t => {
+	// this runs always after each test
```
novemberborn: Perhaps:

> this runs after each test and other test hooks, even if they failed

test/test-collection.js
```diff
@@ -100,6 +102,14 @@ test('throws if you try to set a hook as exclusive', function (t) {
 	t.end();
 });
 
+test('throws if you try to set a before hook as alwyas', function (t) {
```
novemberborn: `always`

novemberborn:
I think you should check for `metadata.always` outside of the `type !== 'test'` condition. Currently AVA won't complain if you use say `test.only.always()`. It would also be good if the error message conveyed which combinations are valid, e.g.:

```js
if (metadata.always && type !== 'after' && type !== 'afterEach') {
	throw new Error('"always" can only be used with after and afterEach hooks');
}

if (type !== 'test') {
	if (metadata.exclusive) {
		throw new Error('"only" cannot be used with a ' + type + ' test');
	}

	this.hooks[type + (metadata.always ? 'Always' : '')].push(test);
}
```

Would be good to test this error with `test` and `beforeEach` as well.

(Maybe we should change "test" in the "only" message to "hook" too.)

---

@cgcgbcbc almost there!

I am wondering though whether this should just be the default behavior. Will discuss further in #474.

readme.md
jamestalmage: Use test.after.always() to register a hook that will always run once your tests and other hooks complete. `.always()` hooks run regardless of whether there were earlier failures, so they are ideal for cleanup tasks.

readme.md
jamestalmage: what @novemberborn said, plus:

> `.always()` hooks are ideal for cleanup tasks.

readme.md
jamestalmage: > this will always run, even if the test or earlier hooks failed.

jamestalmage:
@cgcgbcbc - All that's really left here are a few minor nits on the docs, etc.

If you can handle those, I think we're good to merge. If anything we've said is ambiguous and you can't get an answer from us right away, just use your best judgement.

cgcgbcbc:
@jamestalmage Updated according to the comments

sindresorhus:
Nice work @cgcgbcbc :)

LGTM

cgcgbcbc:
@jamestalmage should I open an issue in sindresorhus/eslint-plugin-ava about this modifier?

jamestalmage:
If one doesn't already exist, yes.

PR #811, opened 2016-05-05
Ava + Grunt - Multi level parallel test runner

subash-canapathy:
Adding a new Recipe:
Ava + Grunt - Multi level parallel test harness for end to end tests

### DIFF
--- a/docs/recipes/grunt-ava-parallel-tests.md
+++ b/docs/recipes/grunt-ava-parallel-tests.md
@@ -1,0 +1,56 @@
+Multi level parallel tests runs where the time taken is only in the order of your longest running test !
+
+##Sample Test harness
+https://github.com/subash-canapathy/grunt-ava-parallel-tests
+
+You can use this as a bootstrap test runner / harness to build automated test framework on. It is designed with webdriver/appium end to end tests in mind so that you can run your UI tests with only your grid as the limiting factor.
+## Usage
+`npm install`
+
+Run the foo bar tests in the package:
+
+`npm test`
+
+OR
+
+`grunt` - if you have grunt installed globally.
+
+### Results
+
+There are 2 test files and each of them have 4 tests in them. Three of the tests use synthetic waits with await (ES6/7 style promise sync shims). If nothing was parallel one run of all the tests in both files will take 8 seconds in total.
+
+![TestResultsWithTime](https://github.com/subash-canapathy/grunt-ava-parallel-tests/blob/master/screenshots/testResultsWithTime.png?raw=true)
+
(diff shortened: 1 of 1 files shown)


jamestalmage:
A few notes:

- I don't really understand what this recipe accomplishes, or how it's worthwhile for our users. Exactly what problem does it solve? That should probably be made clear at the top of the document (I advise you to convince us of the need for this recipe before investing more time in it though).

- I think some of the language needs to be reworked. It feels a little too flowery / sales pitchy at some points:
  -  `note how much time the tests take ...`  - It's unrealistic to expect them to actually run the code from the recipe. Show a screen shot. Or just dryly stating `this way is slower because ...` is fine to.
  - `! #mindblowing` - I'm all for exciting language and exclamation points where warranted, but let's do it while communicating something of value. This doesn't add value, a statement like `That's a whopping 300% speedup!` communicates facts while still slipping in some fun language that keeps it a light read.

All that said, your first task really is convincing us of the problem this solves, and then that this is the best way to solve it.

docs/recipes/grunt-ava-parallel-tests.md
```diff
@@ -0,0 +1,56 @@
...
+* when we get a new driver (with nanotime based ID)
+* when a test uses the driver
+* when the driver gets cleaned up.
+
+## Organizing test suites
+
+The additional parallel dimension here which is not provided by Ava comes from Grunt. The example uses a grunt plugin (from the developer who wrote Ava) `grunt-concurrent`. Its goal is to take a list of grunt tasks and be able to run them concurrently.
+To make sure we can set the environment vars and other configuration options per suite/combination we leverage `grunt-env` and some task compositions to make it work.
+
+Ava test suites are composed in `grunt-shell` blocks (until we have cli options support on `grunt-ava`)
```
sotojuan: Nitpick: It's "AVA" in all caps.

docs/recipes/grunt-ava-parallel-tests.md
```diff
@@ -0,0 +1,56 @@
+Multi level parallel tests runs where the time taken is only in the order of your longest running test !
+
+##Sample Test harness
+https://github.com/subash-canapathy/grunt-ava-parallel-tests
+
+You can use this as a bootstrap test runner / harness to build automated test framework on. It is designed with webdriver/appium end to end tests in mind so that you can run your UI tests with only your grid as the limiting factor.
+## Usage
```
sotojuan: You should have a blank like before and after Markdown headings (you do elsewhere, so this is more about consistency).

sindresorhus:
Closing for lack of response.

PR #815, opened 2016-05-07
Fix `npm link` usage on Node.js 6

### LINKED ISSUE #814: npm link not working on Node 6
To debug, I added the following statements to a number of internal AVA files:

```
console.log(__filename);
```

On Node 5, when everything works, I get this output:

```
/~/dev/ava/lib/fork.js
/~/dev/ava/lib/test-worker.js
/~/dev/ava/lib/test-worker.js
/~/dev/ava/lib/test-worker.js
/~/dev/ava/lib/test-worker.js
/~/dev/ava/index.js
/~/dev/ava/index.js
/~/dev/ava/index.js
/~/dev/ava/index.js
/~/dev/ava/index.js
```

On Node 6, where things are broken:

```
/~/dev/ava-codemods/node_modules/ava/lib/fork.js
/~/dev/ava/lib/test-worker.js
/~/dev/ava/lib/test-worker.js
/~/dev/ava/lib/test-worker.js
/~/dev/ava/lib/test-worker.js
/~/dev/ava-codemods/node_modules/ava/index.js
/~/dev/ava-codemods/node_modules/ava/index.js
/~/dev/ava-codemods/node_modules/ava/index.js
/~/dev/ava-codemods/node_modules/ava/index.js
/~/dev/ava-codemods/node_modules/ava/lib/test-worker.js
/~/dev/ava-codemods/node_modules/ava/lib/test-worker.js
/~/dev/ava-codemods/node_modules/ava/lib/test-worker.js
/~/dev/ava-codemods/node_modules/ava/index.js
/~/dev/ava-codemods/node_modules/ava/lib/test-worker.js
```

So, it appears that, in Node 6, npm linked modules think they are inside the `node_modules` folder of the project from which they are linked, instead of their actual path on the file system.

However, when we fork `test-worker.js`, it is aware of it's actual path. But that gets confused when when the user does `require('ava')` and the file paths show as nested again.

You will notice `test-worker` shows up twice as often as it should, with a deeper path the second time around.

jamestalmage:
Fixes #814, but causes all sorts of test breakage.

I can't really figure out why it breaks the tests, it seems to work fine IRL. My only guess is that being wrapped by `tap` is somehow causing the problem.

### DIFF
--- a/lib/fork.js
+++ b/lib/fork.js
@@ -31,9 +31,29 @@ module.exports = function (file, opts) {
 		} : false
 	}, opts);
 
-	var ps = childProcess.fork(path.join(__dirname, 'test-worker.js'), [JSON.stringify(opts)], {
-		cwd: path.dirname(file),
-		silent: true,
+	var execArgv = process.execArgv;
+	var cwd = path.dirname(file);
+
+	// This is straight out of node child_process
+	if (process._eval != null) { // eslint-disable-line
+		var index = execArgv.lastIndexOf(process._eval);
+		if (index > 0) {
+			// Remove the -e switch to avoid fork bombing ourselves.
+			execArgv = execArgv.slice();
+			execArgv.splice(index - 1, 2);
+		}
+	}
+
+	var args = execArgv.concat([
+		'-e',
+		'require("ava/lib/test-worker")',
+		path.join(__dirname, 'test-worker.js'),
+		JSON.stringify(opts)
+	]);
+
+	var ps = childProcess.spawn(process.execPath, args, {
+		cwd: cwd,
+		stdio: ['pipe', 'pipe', 'pipe', 'ipc'],
 		env: env
 	});
 
--- a/lib/fork.js
+++ b/lib/fork.js
@@ -31,10 +31,14 @@ module.exports = function (file, opts) {
 		} : false
 	}, opts);
 
+	// Workaround for breakage caused by https://github.com/nodejs/node/pull/5950
+	// Those changes will be reverted in https://github.com/nodejs/node/pull/6537
+
+	// Revert #815 when these changes land in Node.js
 	var execArgv = process.execArgv;
 	var cwd = path.dirname(file);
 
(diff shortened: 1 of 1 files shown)


jamestalmage:
Tests fixed

novemberborn:
What if we rewrite the `require('ava')` call in the workers to be a real path?

I'm not sure about how often Node.js releases occur, do we really need to land this? I doubt we'd want to maintain this for the early Node 6 releases. On the other hand it'll help *us* during development, so maybe that's OK.

jamestalmage:
My thought is to merge and track Node 6 progress. Once the old behavior is restored and stable, revert this out.

The code I've copied here out of `child_process` has remained pretty stable, so I don't think we're in danger of missing out on any updates in the core module.

Still, I agree, this isn't something I want to maintain long term, hence all the extra comments letting us know what and when to revert.

I don't want to wait for Node to sort this out before I can use `npm link` again.

sindresorhus:
👍

dcousineau:
When running this fix on our test suite, `nyc` drops reported coverage from **35.95%** (on v0.14.0) to **1.82%** (running HEAD from master).

Not sure if you were aware of this but you may want to delay publishing to NPM anytime soon.

jamestalmage:
@dcousineau - Good Catch!

#827 should fix your issue. NYC still will not work if you are using `npm link ava` on Node 6. But it will always work on all earlier versions of Node, and will also work on Node 6 as long as you haven't `npm link`ed AVA.

jamestalmage:
FYI: The NYC issue is because [`spawn-wrap` does not support the `-e` flag](https://github.com/tapjs/spawn-wrap/blob/master/index.js#L134-L136).

I have opened https://github.com/tapjs/spawn-wrap/issues/22, requesting that support for `-e` be added, but #827 works to solve this issue.

dcousineau:
Makes sense. I'm not using `npm link` anywhere becuase I keep forgetting its a thing :P

PR #827, opened 2016-05-12
Only fix `npm link` of AVA as necessary.

jamestalmage:
PR #815 introduced a fix for `npm link`, but it broke NYC coverage.  

Instead of always applying the "fix", this wraps it in a conditional, such that the old behavior is returned unless we detect a problematic `npm link` setup (`npm link ava` on node 6.0.0-6.?.?).


sindresorhus:
I don't think you meant to commit `ava-0.14.0.tgz`.

jamestalmage:
tarball removed.

sindresorhus:
👍

PR #829, opened 2016-05-12
fix crash in 'fake' browser env

### LINKED ISSUE #822: beautifyStack error when jsdom is imported
## Description

The following test file
```js
import test from 'ava'

import {jsdom} from 'jsdom'
global.document = jsdom('')
global.window = document.defaultView
Object.keys(document.defaultView).forEach((property) => {
  if (typeof global[property] === 'undefined') {
    global[property] = document.defaultView[property]
  }
})

test('yay', t => {
  t.is(true, false)
})
```
results in an error when processing the test failure
screen shot 2016-05-10 at 11 13 10
It's somehow related to that `Object.keys().forEach()` which copies properties from `window` to `global`. (if I comment it out it doesn't happen)
That snippet is what we suggest users do to use `enzyme` with `jsdom` https://github.com/airbnb/enzyme/blob/master/docs/guides/jsdom.md
I don't see anything wrong with it, since it doesn't replace any global properties that already existed.
So I'm not sure whether this is a bug with `ava` or not, but it at least looks like the error message wasn't printed in its entirety. 

## Environment

```
Node.js v6.0.0
darwin 15.4.0
```

nfcampos:
closes #822 

- I've added a failing test. basically just copied one of the existing source map fixtures and added the two lines that make `source-map-support` think it's a browser env. I'm not sure I added the test in the right place, do let me know if it should go somewhere else.
```js
global.window = function () {};
global.XMLHttpRequest = function () {};
```
- I've also added the option @novemberborn mentioned on https://github.com/sindresorhus/ava/issues/822#issuecomment-218703506 to `sourceMapSupport.install()` to fix the bug

### DIFF
--- a/lib/test-worker.js
+++ b/lib/test-worker.js
@@ -48,6 +48,7 @@ Promise.longStackTraces();
 var sourceMapCache = Object.create(null);
 
 sourceMapSupport.install({
+	environment: 'node',
 	handleUncaughtExceptions: false,
 	retrieveSourceMap: function (source) {
 		if (sourceMapCache[source]) {
--- a/test/api.js
+++ b/test/api.js
@@ -384,6 +384,27 @@ function generateTests(prefix, apiCreator) {
 			});
 	});
 
+	test(prefix + 'stack traces for exceptions are corrected using a source map file in what looks like a browser env', function (t) {
+		t.plan(4);
+
+		var api = apiCreator({
+			cacheEnabled: true
+		});
+
+		api.on('test-run', function (runStatus) {
+			runStatus.on('error', function (data) {
+				t.match(data.message, /Thrown by source-map-fixtures/);
+				t.match(data.stack, /^.*?Object\.t.*?as run\b.*source-map-fixtures.src.throws.js:1.*$/m);
+				t.match(data.stack, /^.*?Immediate\b.*source-map-file-browser-env.js:14.*$/m);
+			});
+		});
+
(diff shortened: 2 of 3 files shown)


novemberborn:
I don't think we need the failing test. Arguably this was due to a misconfiguration.

nfcampos:
ok, I'll remove it

vdemedes:
Thanks a lot, @nfcampos!

jamestalmage:
Hmm. The "failing test" commit passes on Travis, but not Windows.

nfcampos:
I mistakenly committed the failing test together with the fix at first :( I then rewrote the commits but I guess travis never ran the tests with just the failing test. but it does fail :)





On Fri, May 13, 2016 at 10:47 AM -0700, "James Talmage" <notifications@github.com> wrote:












Hmm. The "failing test" commit passes on Travis, but not Windows.

jamestalmage:
LGTM

nfcampos:
so should I leave the test in or not?





On Fri, May 13, 2016 at 11:00 AM -0700, "James Talmage" <notifications@github.com> wrote:












LGTM

jamestalmage:
> so should I leave the test in or not?

I missed @novemberborn's comment on the first pass. IDK. It guards against a regression, so it seems like a good idea to me.

novemberborn:
Thanks @nfcampos!

PR #831, opened 2016-05-14
[WIP] A way to submit tests that should fail.

cgcgbcbc:
see #673

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -415,6 +415,16 @@ You can use the `.todo` modifier when you're planning to write a test. Like skip
 test.todo('will think about writing this later');
 ```
 
+### Failing tests
+
+You can use the `.failing` modifier for submitting tests that are intended to be failures without killing CI process. And if the failing test does pass, it will fail, and it will remind you remove the failing modifier.
+
+```js
+test.failing('this test prove something wrong', t => {
+	t.fail();
+}); // will count as passed test
+```
+
 ### Before & after hooks
 
 AVA lets you register hooks that are run before and after your tests. This allows you to run setup and/or teardown code.
--- a/readme.md
+++ b/readme.md
@@ -417,12 +417,15 @@ test.todo('will think about writing this later');
 
 ### Failing tests
 
-You can use the `.failing` modifier for submitting tests that are intended to be failures without killing CI process. And if the failing test does pass, it will fail, and it will remind you remove the failing modifier.
+You can use the `.failing` modifier to document issues with your code that need to be fixed. Failing tests are run just like normal ones, but they are expected to fail, and will not break your build when they do. If a test marked as failing actually passes, it will be reported as an error and fail the build with a helpful message instructing you to remove the .failing modifier.
+
+This allows you to merge `.failing` tests before a fix is implemented without breaking CI. This is a great way to recognize good bug report PR's with a commit credit, even if the reporter is unable to actually fix the problem.
 
 ```js
-test.failing('this test prove something wrong', t => {
-	t.fail();
-}); // will count as passed test
+// See: github.com/user/repo/issues/1234
+test.failing('demonstrate some bug', t => {
+	t.fail(); // test will count as passed
+});
 ```
 
 ### Before & after hooks
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -16,6 +16,7 @@ var chainableMethods = {
 		exclusive: false,
 		skipped: false,
 		todo: false,
+		failing: false,
 		callback: false,
 		always: false
 	},
@@ -26,6 +27,7 @@ var chainableMethods = {
 		after: {type: 'after'},
 		skip: {skipped: true},
 		todo: {todo: true},
+		failing: {failing: true},
 		only: {exclusive: true},
 		beforeEach: {type: 'beforeEach'},
 		afterEach: {type: 'afterEach'},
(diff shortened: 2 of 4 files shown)


jfmengels:
There needs to be a test that makes the tests fail when a test passes.

jamestalmage:
Agreed, needs quite a few more tests.

- [x] failing sync test fails - result is passing
- [x] failing test.cb test fails - result is passing
- [ ] failing test with `t.throws(nonThrowingPromise)` - result is passing (test async assertions)
- [ ] failing test returns a rejected promise - result is passing
- [ ] failing sync test passes - result is a failure with helpful error message
- [ ] failing test.cb test passes - result is a failure
- [ ] failing test with `t.throws(throws)` - result is failure (test async assertions)
- [ ] failing test returns a resolved promise - result is failure

cgcgbcbc:
> failing test with t.throws(throws) - result is failure (test async assertions)

Shouldn't this be passing?

cgcgbcbc:
Just for tracking

- [ ] failing sync test fails - result is passing
- [ ] failing test.cb test fails - result is passing
- [ ] failing test with t.throws(nonThrowingPromise) - result is passing (test async assertions)
- [ ] failing test returns a rejected promise - result is passing
- [ ] failing sync test passes - result is a failure with helpful error message
- [ ] failing test.cb test passes - result is a failure
- [ ] failing test with t.throws(throws) - result is failure (test async assertions)
- [ ] failing test returns a resolved promise - result is failure

jamestalmage:
> Just for tracking

Any reason you haven't tackled the last two?

```js
// this should pass
test.failing(t => {
  t.throws(Promise.resolve('foo'));
});
// this should fail
test.failing(t => {
  t.notThrows(Promise.resolve('foo'));
});

**Comment by cgcgbcbc** [comment]
> Any reason you haven't tackled the last two?

Cause they failed...

**Comment by jamestalmage** [comment]
This should fix your failing test:

```diff
diff --git a/test/test.js b/test/test.js
index d2f3f1b..9fd23a9 100644
--- a/test/test.js
+++ b/test/test.js
@@ -676,10 +676,10 @@ test('failing test with t.throws(nonThrowingPromise) is passing', function (t) {
 
 test('failing test with t.throws(throws) is failure', function (t) {
 	ava.failing(function (a) {
-		a.throws(Promise.resolve('foo'));
+		a.notThrows(Promise.resolve('foo'));
 	}).run().then(function (result) {
 		t.is(result.passed, false);
-		t.is(result.message, failingTestHint);
+		t.is(result.reason.message, failingTestHint);
 		t.end();
 	});
 });
```

cgcgbcbc:
😅Aha, not the first time making such embarrassing mistake

jamestalmage:
Awesome!

LGTM!

jamestalmage:
@cgcgbcbc - Can you add documentation for this to the Readme?

cgcgbcbc:
@jamestalmage , do you think it's necessary to check that the failing modifier can only be used with test and throws error when used with hooks?

jamestalmage:
> do you think it's necessary to check that the failing modifier can only be used with test and throws error when used with hooks?

No, it's conceivable you could write an `afterEach` test that failed.

```js
test.failing.afterEach('always leaves X in pristine state', t => {
  t.true(isPristine(t.context.state));
});
```

That's enough of a corner case that it should probably be flagged by the linter though.

Can you open an issue in `eslint-plugin-ava` about that?

cgcgbcbc:
> Can you open an issue in eslint-plugin-ava about that?

Do you mean the no-unknown-modifiers rule?

jamestalmage:
I just did it: https://github.com/sindresorhus/eslint-plugin-ava/issues/108

readme.md
```diff
@@ -415,6 +415,16 @@ You can use the `.todo` modifier when you're planning to write a test. Like skip
 test.todo('will think about writing this later');
 ```
 
+### Failing tests
+
+You can use the `.failing` modifier for submitting tests that are intended to be failures without killing CI process. And if the failing test does pass, it will fail, and it will remind you remove the failing modifier.
```
jamestalmage: I would like to expand on this a bit: 

> You can use the `.failing` modifier to document issues with your code that need to be fixed. Failing tests are run just like normal ones, but they are expected to fail, and will not break your build when they do. If a test marked as `failing` actually passes, it will be reported as an error and fail the build with a helpful message instructing you to remove the `.failing` modifier.

> This allows you to merge `.failing` tests before a fix is implemented without breaking CI. This is a great way to recognize good bug report PR's with a commit credit, even if the reporter is unable to actually fix the problem.

readme.md
```diff
@@ -415,6 +415,16 @@ You can use the `.todo` modifier when you're planning to write a test. Like skip
...
+
+```js
+test.failing('this test prove something wrong', t => {
```
jamestalmage: People should link to the relevant issue. We should model that behavior in the docs.

Also, a bit of language/style cleanup.

```js
// See: github.com/user/repo/issues/1234
test.failing('demonstrate some bug', t => {
  t.fail(); // test will count as passed
});
```

cgcgbcbc:
@jamestalmage Thank you for your advice! I've fixed the documents.

jamestalmage:
This looks great to me!

Just need a second reviewer to merge.

// @sindresorhus @novemberborn @vdemedes @sotojuan

novemberborn:
Looks good @cgcgbcbc!

I have some nits but will follow up with a PR rather than putting you through more back-and-forth 😄

sotojuan:
Thanks @cgcgbcbc!

Issue #695, opened 2016-04-01
[Idea]: test macros


jamestalmage:
From: https://github.com/sindresorhus/ava/issues/78#issuecomment-204538386

Occasionally it's useful to create a function that calls `test` for you:

```js
// fictional module that analyzes math expressions in strings:
import calculator from 'calculator-parser';
import ava from 'ava';

function test(input, expected, only) {
  const method = only ? ava.only : ava;
  method(input + ' === ' + expected, t => {
    const actual = calculator.calculate(input);
    t.is(actual, expected);
  });
}

test('3 + 4', 7);
test('5 - 2', 3);
test('6 * 3', 18, true); // causes this to be interpreted as an exclusive test
// ...
```

Unfortunately, such helpers create a situation where it becomes basically impossible to do static analysis on the code (static analysis may be necessary to solve #78).

I think we could solve some of this with a `macro` command:

```js
import test from `ava`;

const myMacro = test.macro(
  (t, input, expected) => { // additional arguments to the macro are passed in. 
    const actual = calculator.calculate(input);
    t.is(actual, expected);
  },
  (input, expected) => input + ' === ' + expected // optional function to compute the test tile from the input.
);

myMacro('3 + 4', 7);

// macros get some of the same modifiers as a normal test function:
myMacro.skip('3 * 6', 18);
myMacro.only('9 / 3', 3);
```

jamestalmage:
@jfmengels points out another potential benefit of this in our linter: https://github.com/sindresorhus/eslint-plugin-ava/pull/56#issuecomment-204984896

jfmengels:
I love the idea for test macros, but I noticed that it will make linting some stuff harder (finding out that `.only`, `.skip` is used etc.). It's a trade-off.

jamestalmage:
> but I noticed that it will also make linting some stuff harder

I am not sure if Eslint has an equivalent to Babel's [bindings](https://github.com/thejameskyle/babel-handbook/blob/master/translations/en/plugin-handbook.md#toc-bindings). It would make linting macros pretty easy. I have not seen anything in the docs though.

jamestalmage:
@jfmengels - What if, instead of storing a reference, you had to pass a string identifier, and it became just another macro.

```js
test.macro('myMacro', (t, input, expected) => ...);

test.myMacro('input', 'expected');
test.only.myMacro('input', 'expected');
```

I think that makes static analysis a little easier (especially if you define macros in helpers).

[...]

PR #832, opened 2016-05-14
Basic macro support.

jamestalmage:
This adds basic macro support as discussed in #695.

It does not completely implement the spec outlined there, specifically:

- The macro can only specify the title using a function in `macroFn.title`. We discussed allowing `macroFn.title` to also be a string, and allowing some form of template language for extracting a title. However, using ES2015 string templates is already pretty easy, so we may just skip this.
- ~~We discussed allowing an array of macro functions.~~

Both the above proposals are found in [this comment](https://github.com/sindresorhus/ava/issues/695#issuecomment-205929738). They both enhance the implementation found in this commit, and would not break the contract. So I don't think there is anything preventing us from shipping this now.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -544,11 +544,14 @@ function macro(t, input, expected) {
 	t.is(eval(input), expected);
 }
 
-macro.title = (input, expected) => `${input} === ${expected}`;
+macro.title = (providedTitle, input, expected) => `${providedTitle} ${input} === ${expected}`.trim();
 
 test(macro, '2 + 2', 4);
 test(macro, '2 * 3', 6);
+test('providedTitle', macro, '3 * 3', 9);
 ```
+
+The `providedTitle` argument defaults to an empty string if the user does not supply a string title. This allows for easy concatenation without having to worry about `null` / `undefined`. It is worth remembering that the empty string is considered a falsy value, so you can still use `if(providedTitle)`.
 
 You can also pass arrays of macro functions:
 
--- a/readme.md
+++ b/readme.md
@@ -550,6 +550,23 @@ test(macro, '2 + 2', 4);
 test(macro, '2 * 3', 6);
 ```
 
+You can also pass arrays of macro functions:
+
+```js
+var safeEval = require('safe-eval');
+
+function evalMacro(t, input, expected) {
+	t.is(eval(input), expected);
+}
+
+function safeEvalMacro(t, input, expected) {
+	t.is(safeEval(input), expected);
+}
+
+test([evalMacro, safeEvalMacro], '2 + 2', 4);
+test([evalMacro, safeEvalMacro], '2 * 3', 6);
+```
+
 
 ### Custom assertions
 
--- a/readme.md
+++ b/readme.md
@@ -551,12 +551,12 @@ test(macro, '2 * 3', 6);
 test('providedTitle', macro, '3 * 3', 9);
 ```
 
-The `providedTitle` argument defaults to an empty string if the user does not supply a string title. This allows for easy concatenation without having to worry about `null` / `undefined`. It is worth remembering that the empty string is considered a falsy value, so you can still use `if(providedTitle)`.
+The `providedTitle` argument defaults to an empty string if the user does not supply a string title. This allows for easy concatenation without having to worry about `null` / `undefined`. It is worth remembering that the empty string is considered a falsy value, so you can still use `if(providedTitle) {...}`.
 
 You can also pass arrays of macro functions:
 
 ```js
-var safeEval = require('safe-eval');
+const safeEval = require('safe-eval');
 
 function evalMacro(t, input, expected) {
 	t.is(eval(input), expected);
@@ -570,6 +570,7 @@ test([evalMacro, safeEvalMacro], '2 + 2', 4);
 test([evalMacro, safeEvalMacro], '2 * 3', 6);
 ```
 
+We encourage you to use macros instead of building your own test generators ([here is an example](https://github.com/jamestalmage/ava-codemods/blob/47073b5b58aa6f3fb24f98757be5d3f56218d160/test/ok-to-truthy.js#L7-L9) of code that should be replaced with a macro). Macros are designed to perform static analysis of your code, which can lead to better performance, IDE integration, and linter rules.
 
 ### Custom assertions
 
--- a/readme.md
+++ b/readme.md
@@ -524,6 +524,33 @@ test.only.serial(...);
 
 This means you can temporarily add `.skip` or `.only` at the end of a test or hook definition without having to make any other changes.
 
+### Test macros
+
+Additional arguments after the test function will be passed as an array to the test function. This is useful for creating reusable test macros.
+
+```js
+function macro(t, [input, expected]) {
+  t.is(eval(input), expected);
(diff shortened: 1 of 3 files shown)


nfcampos:
The same reason that makes it worthwhile to pass arguments as an array to the test function without spreading them should make us want to give them as an array to `test()`, or is it somehow more likely we'll want in the future to introduce a third agumment to the test function than a last argument to `test()`?

readme.md
```diff
@@ -524,6 +524,33 @@ test.only.serial(...);
...
+```js
+function macro(t, [input, expected]) {
+  t.is(eval(input), expected);
+}
+
+test('2 + 2 === 4', macro, '2 + 2', 4);
+test('2 * 3 === 6', macro, '2 * 3', 6);
+```
+
+You can build the test title programatically by attaching a `title` function to the macro:
```
jfmengels: programmatically

jfmengels:
:tada: Yay, this is great @jamestalmage! Didn't know whether this was accepted, but glad it's going forward, it's a really nice feature IMO.

readme.md
```diff
@@ -524,6 +524,33 @@ test.only.serial(...);
 
 This means you can temporarily add `.skip` or `.only` at the end of a test or hook definition without having to make any other changes.
 
+### Test macros
+
+Additional arguments after the test function will be passed as an array to the test function. This is useful for creating reusable test macros.
```
jfmengels: Wouldn't it rather be "creating reusable tests"?

readme.md
```diff
@@ -524,6 +524,33 @@ test.only.serial(...);
...
+
+```js
+function macro(t, [input, expected]) {
+  t.is(eval(input), expected);
```
novemberborn: Indentation should be a tab 😉

novemberborn:
LGTM!

jamestalmage:
What about @nfcampos argument? Should it be arrays both places? Spread both places?

novemberborn:
Spread in the implementation does prevent us from adding any additional arguments, but having a third argument wouldn't be very pretty anyway.

An array in the declaration is just annoying and prone to confusion. So I guess I'm leaning towards spread in both places.

jamestalmage:
> I'm leaning towards spread in both places.

Me too, but that pretty much closes that portion if the API from further expansion. That's a big decision, so let's wait for more input.

jfmengels:
I would've preferred spread in both places, but it does get tricky if you want to add new args.

If we go for spread and want to go for arrays later, then I think that writing a codemod will be kind of tricky. (easy in most cases, but hard when the macro is defined in a separate file).

I'd go for consistency at least: either spread in both places or arrays in both places.

readme.md
jamestalmage: That may be a cleaner way of saying it, but I'm trying to introduce the term. Maybe "reusable macros", but I think it's good as is.

jamestalmage:
I added array of macros support.

jamestalmage:
Should the `macro.title` function get passed title as the first arg?

```js
function macroA() {...}
function macroB() {...}

macro.title = (title, ...remainingArgs) => title + ' - macroA';
macro.title = (title, ...remainingArgs) => title + ' - macroB';

test('groupA', [macroA, macroB], 'some', 'args');
test('groupB', [macroA, macroB], 'more', 'args');
```

results in the following test titles:

```
groupA - macroA
groupA - macroB
groupB - macroA
groupB - macroB
```

novemberborn:
> Should the macro.title function get passed title as the first arg?

I think that'd be good, but it makes it harder to write macros since it'd be optional. I'm specifically thinking about adding a space separator:

```js
macro.title = (prefix) => prefix + ' - macroA';
```

Let's assume prefix defaults to the empty string:

```js
macro.title = (prefix) => (prefix ? prefix + ' - ' : '') + 'macroA';
```

Maybe if the resulting title is always trimmed on its left so you don't have to worry about the spacing?

novemberborn:
LGTMO!

jfmengels:
> Should the macro.title function get passed title as the first arg?

:+1:

jamestalmage:
I think defaulting to the empty string is a good plan. That's still falsy. We should trim results both left and right I say.

jamestalmage:
For spacing, what if we allowed them to return an array of strings which we joined with a space character?

```js
macro.title = (title) => [title, 'macroA'];
```

Or maybe we allow this:

```js
macro.title = {prefix: 'macroA - '};
// or
macro.title = {suffix: ' - macroB'};
```

jfmengels:
I think the former is a bit unnatural. I'd say stick to a function, and seek to improve later it if it feels insufficient.

jamestalmage:
51bed36 passes title as first argument (or empty string if none provided).

I think this should be merge-able now.

@sindresorhus - Haven't heard much from you on this. Specifically, are you good with spreading the args?

readme.md
```diff
@@ -524,6 +524,53 @@ test.only.serial(...);
 
 This means you can temporarily add `.skip` or `.only` at the end of a test or hook definition without having to make any other changes.
 
+### Test macros
+
+Additional arguments passed to the test declaration will be passed to the test implementation. This is useful for creating reusable test macros.
+
+```js
+function macro(t, input, expected) {
```
sindresorhus: @jamestalmage Good thing we decided not to pass the context as the second argument 😝

readme.md
```diff
@@ -524,6 +524,53 @@ test.only.serial(...);
...
+}
+
+macro.title = (providedTitle, input, expected) => `${providedTitle} ${input} === ${expected}`.trim();
+
+test(macro, '2 + 2', 4);
+test(macro, '2 * 3', 6);
+test('providedTitle', macro, '3 * 3', 9);
+```
+
+The `providedTitle` argument defaults to an empty string if the user does not supply a string title. This allows for easy concatenation without having to worry about `null` / `undefined`. It is worth remembering that the empty string is considered a falsy value, so you can still use `if(providedTitle)`.
```
sindresorhus: `if (providedTitle) {`

readme.md
```diff
@@ -524,6 +524,53 @@ test.only.serial(...);
...
+
+You can also pass arrays of macro functions:
+
+```js
+var safeEval = require('safe-eval');
```
sindresorhus: `var` => `const`.

readme.md
```diff
@@ -524,6 +524,53 @@ test.only.serial(...);
 
 This means you can temporarily add `.skip` or `.only` at the end of a test or hook definition without having to make any other changes.
 
+### Test macros
+
+Additional arguments passed to the test declaration will be passed to the test implementation. This is useful for creating reusable test macros.
```
sindresorhus: Should we maybe mention why it's beneficial to use test macros over hand-made loops/etc? (To make static analysis possible === better perf).

sindresorhus:
:shipit:

jfmengels:
:tada: Folks, 0.15 is starting to get pretty exciting :D

readme.md
jamestalmage: Yeah, eating my words on that. Though not as badly as I've been forced to [eat these ones](https://github.com/sindresorhus/ava/issues/222#issuecomment-160546933)

PR #834, opened 2016-05-16
Protect against bad argument passed to t.throws.

jamestalmage:
If a non-function, non-promise, non-observable argument (i.e. a string)  is passed to `t.throws`, the assertion passes. This obviously isn't what we want.

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -84,6 +84,10 @@ x.throws = function (fn, err, msg) {
 			});
 	}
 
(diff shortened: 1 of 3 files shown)


jamestalmage:
This is a pretty major bug, and a straightforward fix. Should be merged ASAP.

sotojuan:
Good catch!

jamestalmage:
I didn't catch it. Reported on Gitter.

PR #835, opened 2016-05-16
Follow-up to #831.

novemberborn:
Follow-up to #831.

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -203,7 +203,7 @@ Object.defineProperty(Test.prototype, 'end', {
 			return this._end.bind(this);
 		}
 
-		throw new Error('t.end is not supported in this context. To use t.end as a callback, you must use "callback mode" via `test.cb(testName, fn)` ');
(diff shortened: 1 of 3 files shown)


lib/test.js
```diff
@@ -191,7 +191,7 @@ Test.prototype._result = function () {
 	if (this.metadata.failing) {
 		passed = !passed;
 		if (!passed) {
-			reason = new Error('Test was expected to fail, but succeeded, you should unmark the test as failing');
+			reason = new Error('Test was expected to fail, but succeeded. You should stop marking the test as failing');
```
sindresorhus: It's kinda weird only having a `.` after the first sentence.

sindresorhus: We might want an ESLint rule for this.

novemberborn: OK, reverted to the original `succeeded, you should` form.

ESLint rule?

sindresorhus: https://github.com/sindresorhus/eslint-plugin-xo/issues/26

PR #838, opened 2016-05-17
Remove timestamp from non-watch test runs

### LINKED ISSUE #775: Remove timestamp from non-watch test runs
```
  ✔ execa() (371ms)
  ✔ input can be a Stream (173ms)

  17 tests passed [15:51:49]
```

It's useful for `--watch`, but when running tests normally I don't see the value.

@novemberborn Thoughts?

sotojuan:
Fixes #775.

WIP because I need to add some tests + need more review especially for testing. Would it be better to remove the `time` from all the modified tests if I have some specific tests that do test time on watch?

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -190,7 +190,7 @@ if (cli.flags.watch) {
 } else {
 	api.run(files)
 		.then(function (runStatus) {
-			logger.finish(runStatus);
+			logger.finish(runStatus, false);
 			logger.exit(runStatus.failCount > 0 || runStatus.rejectionCount > 0 || runStatus.exceptionCount > 0 ? 1 : 0);
 		})
 		.catch(function (err) {
--- a/cli.js
+++ b/cli.js
@@ -147,7 +147,7 @@ if (cli.flags.tap && !cli.flags.watch) {
 } else if (cli.flags.verbose || isCi) {
 	reporter = verboseReporter();
 } else {
-	reporter = miniReporter();
+	reporter = miniReporter({isWatching: cli.flags.watch});
 }
 
 reporter.api = api;
@@ -190,7 +190,7 @@ if (cli.flags.watch) {
 } else {
 	api.run(files)
 		.then(function (runStatus) {
-			logger.finish(runStatus, false);
+			logger.finish(runStatus);
 			logger.exit(runStatus.failCount > 0 || runStatus.rejectionCount > 0 || runStatus.exceptionCount > 0 ? 1 : 0);
 		})
 		.catch(function (err) {
--- a/cli.js
+++ b/cli.js
@@ -147,7 +147,7 @@ if (cli.flags.tap && !cli.flags.watch) {
 } else if (cli.flags.verbose || isCi) {
 	reporter = verboseReporter();
 } else {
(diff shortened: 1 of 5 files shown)


novemberborn:
> Would it be better to remove the time from all the modified tests if I have some specific tests that do test time on watch?

Yea. Having to add the time in all those tests should've been a clue that time was being output too much.

---

Passing the boolean is a bit awkward, but I think it's OK. Ideally the logger understands test runs, so we don't have to interact with it [as precisely as we do](https://github.com/sindresorhus/ava/blob/a4541289491e7f839e7c9417fa0f01a76af42946/lib/watcher.js#L51:L59) in the watcher.

sotojuan:
Fixed tests.

novemberborn:
Can probably move [these two lines](https://github.com/sotojuan/ava/blob/55d8261ef129a2cbd072acaf19932579081b8a64/test/reporters/mini.js#L32:L33) into the new test now.

LGTMO!

jamestalmage:
We should pass an options object to the reporter constructor with static details/config like `isWatching`.

sotojuan:
Good idea—just pushed some changes to make that work.

cli.js
```diff
@@ -147,7 +147,7 @@ if (cli.flags.tap && !cli.flags.watch) {
 } else if (cli.flags.verbose || isCi) {
 	reporter = verboseReporter();
 } else {
-	reporter = miniReporter();
+	reporter = miniReporter({isWatching: cli.flags.watch});
```
jamestalmage: `isWatching` => `watching`

lib/reporters/mini.js
```diff
@@ -26,6 +26,9 @@ function MiniReporter() {
 	});
 	this.spinnerInterval = spinnerDef.interval;
 
+	this.options = options || {};
+	this.options.isWatching = this.options.isWatching || false;
```
jamestalmage: No need to do this. It's already falsy.

lib/reporters/mini.js
```diff
@@ -26,6 +26,9 @@ function MiniReporter() {
 	});
 	this.spinnerInterval = spinnerDef.interval;
 
+	this.options = options || {};
```
jamestalmage: make a save copy with `objectAssign`;

jamestalmage:
Just so we're all on the same page about reporter constructor options. IMO, I think the reporters should avoid maintaining any mutable state, especially as it relates to the test run status. We are violating that rule in a couple spots now (my fault, I was supposed to follow up after merging the PR that introduced `RunStatus`). 

I *do* think it's fine to pass immutable config values that will never change during the test run (i.e. `watching`, `basedir`, `pkgConf`, `debugging`, etc). That should keep our reporters as state free and easily testable as possible, without necessitating passing around needless config data.

sotojuan:
Pushed the changes (and thanks for the explanation, makes sense).

jamestalmage:
@sotojuan 

Move the two lines @novemberborn talked about [here](https://github.com/avajs/ava/pull/838#issuecomment-220106590) into the test that uses it.

Once you've done that (and tests pass), go ahead and merge.

PR #839, opened 2016-05-17
api#run should return runStatus even when no files are found

nfcampos:
on master running `ava blah` assuming `blah` matches no files will fail like this 
screen shot 2016-05-17 at 11 02 14

instead of like this, as it should
screen shot 2016-05-17 at 11 03 00

This happens because when no files are found `Api.prototype.run` is returning an empty array instead of `runStatus`. So it then fails in the reporter, which in its `finish` method expects its argument to be an instance of `runStatus` instead of an empty array

### DIFF
--- a/api.js
+++ b/api.js
@@ -105,7 +105,7 @@ Api.prototype._run = function (files, _options) {
 			file: undefined
 		});
 
-		return Promise.resolve([]);
+		return Promise.resolve(runStatus);
 	}
 
 	var cacheEnabled = self.options.cacheEnabled !== false;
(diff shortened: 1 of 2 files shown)


jamestalmage:
Good catch! Can you add a test to `test/cli.js` that protects against future regressions?

nfcampos:
sure

nfcampos:
I've rebased and added the test :)

jamestalmage:
LGTM

novemberborn:
Thanks @nfcampos!

PR #844, opened 2016-05-18
RunStatus#prefixTitle should remove this.base only if it occurs at the beginning of the path

nfcampos:
When `this.base` is `/` it ends up replacing the first `/`, joining the first two directories.
Before this change it  #looked like this:
screen shot 2016-05-18 at 16 36 40
(notice no separator between `backend` and `models`)
But it should look like this:
screen shot 2016-05-18 at 16 35 31
Unless I misunderstood `this.base` it should only be replaced if that's what the path begins with.
I didn't add any test because there's no test file for `run-status.js`

### DIFF
--- a/lib/run-status.js
+++ b/lib/run-status.js
@@ -143,6 +143,9 @@ RunStatus.prototype.prefixTitle = function (file) {
 		.replace(/test\-/g, '')
 		.replace(/\.js$/, '')
 		.split(path.sep)
+		.filter(function (p) {
+			return p !== '__tests__';
+		})
 		.join(separator);
 
 	if (prefix.length > 0) {
--- a/lib/run-status.js
+++ b/lib/run-status.js
@@ -134,7 +134,10 @@ RunStatus.prototype.prefixTitle = function (file) {
 	var separator = ' ' + chalk.gray.dim(figures.pointerSmall) + ' ';
 
 	var prefix = path.relative('.', file)
-		.replace(this.base, '')
+		.replace(this.base, function (match, offset) {
+			// only replace this.base if it is found at the start of the path
+			return offset === 0 ? '' : match;
+		})
 		.replace(/\.spec/, '')
 		.replace(/\.test/, '')
 		.replace(/test\-/g, '')
--- a/test/run-status.js
+++ b/test/run-status.js
@@ -1,0 +1,70 @@
+'use strict';
+var path = require('path');
+var test = require('tap').test;
+var chalk = require('chalk');
+var figures = require('figures');
+var RunStatus = require('../lib/run-status');
+
+var sep = ' ' + chalk.gray.dim(figures.pointerSmall) + ' ';
+
+test('requires new', function (t) {
+	var runStatus = RunStatus;
+	t.throws(function () {
+		runStatus({});
+	}, 'Class constructor RunStatus cannot be invoked without \'new\'');
+	t.end();
+});
+
+test('prefixTitle returns empty if prefixTitles == false', function (t) {
+	var runStatus = new RunStatus({prefixTitles: false});
+	t.is(runStatus.prefixTitle('test/run-status.js'), '');
+	t.end();
+});
+
+test('prefixTitle removes base if found at start of path', function (t) {
+	var runStatus = new RunStatus({base: 'test' + path.sep});
+	t.is(runStatus.prefixTitle('test/run-status.js'), 'run-status' + sep);
+	t.end();
+});
+
+test('prefixTitle does not remove base if found but not at start of path', function (t) {
+	var runStatus = new RunStatus({base: path.sep});
+	t.is(runStatus.prefixTitle('test/run-status.js'), 'test' + sep + 'run-status' + sep);
+	t.end();
+});
+
+test('prefixTitle removes .js extension', function (t) {
+	var runStatus = new RunStatus({base: path.sep});
+	t.is(runStatus.prefixTitle('run-status.js'), 'run-status' + sep);
+	t.end();
+});
+
+test('prefixTitle does not remove .js from middle of path', function (t) {
+	var runStatus = new RunStatus({base: path.sep});
+	t.is(runStatus.prefixTitle('run-.js-status.js'), 'run-.js-status' + sep);
+	t.end();
+});
+
+test('prefixTitle removes __tests__ from path', function (t) {
+	var runStatus = new RunStatus({base: path.sep});
+	t.is(runStatus.prefixTitle('backend/__tests__/run-status.js'), 'backend' + sep + 'run-status' + sep);
+	t.end();
+});
+
+test('prefixTitle removes .spec from path', function (t) {
+	var runStatus = new RunStatus({base: path.sep});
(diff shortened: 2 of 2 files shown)


sindresorhus:
I also noticed that we need to handle `__tests__` so it's not shown in the output.

lib/run-status.js
```diff
@@ -134,7 +134,7 @@ RunStatus.prototype.prefixTitle = function (file) {
 	var separator = ' ' + chalk.gray.dim(figures.pointerSmall) + ' ';
 
 	var prefix = path.relative('.', file)
-		.replace(this.base, '')
+		.replace(new RegExp('^' + this.base), '')
```
sindresorhus: You need to escape the value here.

https://github.com/sindresorhus/escape-string-regexp

nfcampos:
@sindresorhus it should just be removed from the path, like this?
screen shot 2016-05-18 at 16 46 25

sindresorhus:
👍

nfcampos:
Is the path separator always `/`, ie. any way this  `.replace(/__tests__\//, '')` can fail?

lib/run-status.js
nfcampos: good point, i've added that

lib/run-status.js
```diff
@@ -7,6 +7,7 @@ var isObj = require('is-obj');
 var flatten = require('arr-flatten');
 var figures = require('figures');
 var formatter = require('./enhance-assert').formatter();
+var escapeStringRegexp = require('escape-string-regexp');
```
sindresorhus: This needs to be above relative imports, so on line 9.

package.json
```diff
@@ -105,6 +105,7 @@
     "core-assert": "^0.2.0",
     "debug": "^2.2.0",
     "empower-core": "^0.5.0",
+    "escape-string-regexp": "~1.0.5",
```
sindresorhus: Use caret `^`

nfcampos: oops sorry, should there be a `.npmrc` at the root with this?
```
save-prefix='^'
```

sindresorhus: Nah, we have enough meta files.

lib/run-status.js
```diff
@@ -134,9 +135,10 @@ RunStatus.prototype.prefixTitle = function (file) {
 	var separator = ' ' + chalk.gray.dim(figures.pointerSmall) + ' ';
 
 	var prefix = path.relative('.', file)
-		.replace(this.base, '')
+		.replace(new RegExp('^' + escapeStringRegexp(this.base)), '')
```
novemberborn: How about:

```js
var prefix = path.relative('.', file);
if (prefix.indexOf(this.base) === 0) {
  prefix = prefix.slice(this.base.length);
}

prefix = prefix
  .replace(/\.spec/, '')
  // etc
```

Saves us a dependency and (to my eyes) is more obvious than the regular expression.

lib/run-status.js
```diff
@@ -134,9 +135,10 @@ RunStatus.prototype.prefixTitle = function (file) {
...
 		.replace(/\.spec/, '')
 		.replace(/\.test/, '')
+		.replace(/__tests__\//, '')
```
novemberborn: Might be better to add a filter after splitting:

```js
.split(path.sep)
.filter(function (component) {
  return component !== '__tests__';
})
.join(separator);
```

That way you don't have to worry about the slashes.

novemberborn:
> I didn't add any test because there's no test file for run-status.js

There are title prefix tests in [`test/api.js`, starting here](https://github.com/avajs/ava/blob/54e11064841c0169a88324f3daeda396d3e8889f/test/api.js#L100).

lib/run-status.js
nfcampos: @novemberborn to avoid the regex, how about this?
```js
.replace(this.base, function (match, offset, string) {
	if (offset === 0) {
		return ''
	} else {
		return match
	}
})
```

nfcampos:
I've added a test, I don't see any way of testing this through the `api` without changing the current working dir so that's what I did

lib/run-status.js
nfcampos: I've changed it to what you suggested for this

lib/run-status.js
novemberborn: I like the last one! Perhaps add a comment "Only replace base if it's at the start"?

novemberborn:
> I've added a test, I don't see any way of testing this through the api without changing the current working dir so that's what I did

That should be possible, but I don't have time right now to play with it and see how. I can't quite tell how this test verifies that `base` is only removed at the beginning of the path, either.

I suppose we should have tests for the `__test__` filtering, but there's probably no tests for any of the other mutations either.

nfcampos:
@novemberborn how about I write a unit test for the RunStatus.prototype.prefixTitle function, that should be much clearer?

novemberborn:
@nfcampos yea, sounds good!

nfcampos:
@novemberborn I've added some tests for `RunStatus.prototype.prefixTitle`. I think this is ready now

test/run-status.js
```diff
@@ -0,0 +1,70 @@
...
+});
+
+test('prefixTitle removes base if found at start of path', function (t) {
+	var runStatus = new RunStatus({base: 'test' + path.sep});
+	t.is(runStatus.prefixTitle('test/run-status.js'), 'run-status' + sep);
+	t.end();
+});
+
+test('prefixTitle does not remove base if found but not at start of path', function (t) {
+	var runStatus = new RunStatus({base: path.sep});
```
novemberborn: Should probably use a base that's a bit more than a `/`, e.g. `foo/` with `foo/bar/foo/baz.js`, which should then result in `'bar' + sep + 'foo' + sep + 'baz' + sep` (I think?).

novemberborn:
@nfcampos looking good! The base test could still be a little better, but good to go otherwise I think. @avajs/core?

test/run-status.js
nfcampos: @novemberborn `String.prototype.replace` replaces only the first occurrence of the searched string (unless you're using a regex with `g`) so with the base `foo/` the path `foo/bar/foo/baz.js` already worked correctly before this PR — since `base` is found at the beginning and replaced the second occurrence of `base` is never considered. The only `base` for which this fails is indeed `/` because it isn't found at the beginning of the path, and thus the next occurrence of `/` was replaced

novemberborn: @nfcampos yes makes sense. So then you could test with `bar/foo/baz.js` right?

`prefixTitle` replaces slashes so IMO it's not a great `base` value for this test.

nfcampos: @novemberborn I can test the base `foo/` with the path `bar/foo/baz.js` but I'll be testing a case that will never be hit in real use. unless I'm missing something the only possible value for base (defined here https://github.com/avajs/ava/blob/master/api.js#L88) that can be not found at the beginning of the path of a test is indeed `/`

novemberborn: Sorry I'm being silly. You're testing the exact case you set out to fix. 👍 💯

jamestalmage:
Thanks @nfcampos

PR #846, opened 2016-05-20
[typescript] added missing todo definition

andischerer:
Hey guys,
first of all thanks for this awesome testrunner.

I've added the missing todo definition to the index.d.ts.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -57,6 +57,7 @@ export namespace test {
 	export function serial(run: ContextualSerialTest): void;
 	export function cb(name: string, run: ContextualCallbackTest): void;
 	export function cb(run: ContextualCallbackTest): void;
+	export function todo(name: string): void;
 }
 export namespace test.serial {
 	export const before: SerialRunner;


jamestalmage:
@SamVerschueren

sindresorhus:
We also need to add `test.failing`. Anything else that needs updating?

sindresorhus:
// @ivogabe

jamestalmage:
It would be good to land this as part of the upcoming release as well.

ivogabe:
Does `todo` allow chaining? This definition doesn't allow it.

jamestalmage:
Yes. The only checks against `todo` are [here](https://github.com/avajs/ava/blob/master/lib/runner.js#L112-L121).
  - It can't contain an implementation.
  - It must have a title

SamVerschueren:
Seems ok, not much to it :).

jamestalmage:
@SamVerschueren Does it need something additional to enable chaining?

SamVerschueren:
Yes, this will only allow you to write `test.todo('message')`. But `todo` doesn't allow chaining right?

jamestalmage:
I can't really think of a reason to chain:

```
test.serial.todo
test.after.todo
```

Maybe we should just disallow chaining of `todo` methods.

jamestalmage:
Merging. We can debate what to do about `todo` chaining in #847.

sindresorhus:
@ivogabe @SamVerschueren Could either of you add `test.failing`?

https://github.com/avajs/ava#failing-tests

SamVerschueren:
Yeah sure.

jamestalmage:
While you are at it, I think there's an error when it comes to using `test.serial`. 

`test.serial` tests are allowed to return a promise, or have callbacks.

`test.serial.cb` is totally allowed, as is `test.serial(promiseReturning)`.

The point of `test.serial`, is that it makes the tests run independently without contention. `.serial` tests are run first, and one at a time, whereas the default is to run with max concurrency. They are allowed to be async, via `.cb` or returning a Promise/Observable.

SamVerschueren:
I think you're right (not tested yet). It looks like `test.serial.cb` is not defined.

PR #856, opened 2016-05-23
Avoid running the same file multiple times

jamestalmage:
When a glob pattern matches a directory, we launch another glob search, scanning for all `.js` files in that directory (deeply). This causes a problem if the glob pattern was already going to catch those files anyways (patterns that end in `/**` match directories and files) - we end up running those files multiple times.

With this directory structure:

```
test/
  foo/
    bar/
      baz.js
      quz.js
```

And this glob pattern: `test/**`

We currently end up running each file **3 times**!! This is because:
 1. The original glob pattern matches the files
 2. The original glob pattern matches the `test/foo` directory, so we search for all files within that directory, and all subdirectories.
 3. The original glob pattern matches the `test/foo/bar` directory, so we search for all files within that directory.

This adds an optimization that drops `3`, by recognizing that we are already doing a deep directory search of a parent (from step `2`).

It also performs a filter on the final list of files returned guaranteeing there are no duplicates.


lib/ava-files.js
```diff
@@ -197,9 +197,33 @@ function handlePaths(files, excludePatterns) {
 	// convert pinkie-promise to Bluebird promise
 	files = Promise.resolve(globby(files.concat(excludePatterns)));
 
+	var searchedParents = Object.create(null);
+	var foundFiles = Object.create(null);
```
sindresorhus: This would have been the perfect place to use ES2015 Set if we could...

sindresorhus:
Ouch, glad you caught this. Changes looks good :)

PR #858, opened 2016-05-23
document always hook with fail fast flag

### LINKED ISSUE #820: `failFast` causes process to exit before `always.afterEach` hooks complete
## Description

If I try and do something asynchronous (or just slow?) from an `always.afterEach()` function while the `failFast` flag is set in my `package.json` the process won't wait for it to complete when the test fails. My specific case if is that I'm trying to do some database cleanup after each test and even though the `afterEach` hook gets called it doesn't seem to be given a chance to finish before the process exits. 

I've tried to distill this down to a simple example:

```javascript
import test from 'ava'
    
test.afterEach.always('CLEANUP', async t => {
  console.log('START', t.title)
    
  await new Promise(resolve => setTimeout(resolve, 5000))
  
  console.log('END', t.title)
})
    
test('Test', async t => {
  console.log('START', t.title)
  t.plan(1)
    
  t.is(1, 2)
  console.log('END', t.title)
})
```

This is the output I get:

    START Test
    END Test
      ✖ Test 1 === 2
    START CLEANUP for Test
    
      1 test failed [10:42:18]
    
    
      1. Test
      AssertionError: 1 === 2
        _callee2$ (cleanup-test.js:15:5)
        Test.fn (cleanup-test.js:11:1)

I would expect to see `END CLEANUP` printed

## Environment

Node.js v5.11.0
darwin 15.3.0

(Originally asked on SO: From http://stackoverflow.com/questions/37118762/ava-aftereach-callback-not-completing)

cgcgbcbc:
resolve #820

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -436,7 +436,8 @@ AVA lets you register hooks that are run before and after your tests. This allow
 
 `test.beforeEach()` registers a hook to be run before each test in your test file. Similarly `test.afterEach()` a hook to be run after each test. Use `test.afterEach.always()` to register an after hook that is called even if other test hooks, or the test itself, fail. `.always()` hooks are ideal for cleanup tasks.
 
-**Note**: if `--fail-fast` flag is provided, ava will stop after first test failure and `.always` hook will **NOT** run.
+**Note**: If the `--fail-fast` flag is specified, AVA will stop after the first test failure and the `.always` hook will **not** run.
+
 
 Like `test()` these methods take an optional title and a callback function. The title is shown if your hook fails to execute. The callback is called with an [execution object](#t).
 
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -436,6 +436,8 @@ AVA lets you register hooks that are run before and after your tests. This allow
 
 `test.beforeEach()` registers a hook to be run before each test in your test file. Similarly `test.afterEach()` a hook to be run after each test. Use `test.afterEach.always()` to register an after hook that is called even if other test hooks, or the test itself, fail. `.always()` hooks are ideal for cleanup tasks.
 
+**Note**: if `--fail-fast` flag is provided, ava will stop after first test failure and `.always` hook will **NOT** run.
```
```
**Note**: If the `--fail-fast` flag is specified, AVA will stop after the first test failure and the `.always` hook will **not** run.
```

novemberborn:
Thanks @cgcgbcbc!

PR #862, opened 2016-05-23
Fix "Test files must be run with the AVA CLI warning".

jamestalmage:
We used to have a nice warning, telling people to use the CLI if they ran `node path/to/testfile.js`. It was broken by requiring `test-worker` to early. `test-worker` tries `JSON.parse(process.argv[2])`, which will likely throw. Preventing the helpful message.

This adds a fix, and a regression test.


sotojuan:
👍🏼

PR #863, opened 2016-05-24
Properly resolve files from `package.json` dir.

jamestalmage:
Currently, if cwd is a subdir of the project, and you just run `ava`, it can fail to resolve the correct set of test files. This is fixed by changing the `cwd` argument passed to `globby`. However, it seems to me that if you are in a subdirectory, and pass glob patterns on the command line, you probably want those resolved from `cwd`, not magically resolved from some parent directory.


jamestalmage:
This still needs to be fixed for watch mode. I've tried to figure out how to do it / how to test it, and come to the conclusion that I want to punt that to @novemberborn!

jamestalmage:
I think this gets even harder for watch mode because of `--sources`.

If they are `cd`ed into a child directory and use the `--sources` flag, that should probably be resolved from `cwd`, right? But if they don't specify `--sources`, then we probably want to use whatever is in `package.json`, and resolve from that directory? That's getting complicated.

Should we eliminate the `--sources` flag? I have never used it, and I don't see why you really need it. Doesn't this qualify as a static config item that should live in `package.json`?

jamestalmage:
Should we merge this and fix watcher in another PR? This is really just a bugfix.

novemberborn:
Probably won't have time for the next few days to look into this. There's no rush to get this out though?

👍  on removing the `--sources` flag. Having those be relative to the `package.json` seems sensible. As does overriding `files` through the CLI, and those being relative to the actual CWD.

jamestalmage:
No rush, other than avoiding a rebate.

jamestalmage:
I would like to merge this.

We can look into making this work for `watcher` later.

jamestalmage:
closing in favor of #925 - which extracts `ava-files` to an external module (which already has these changes)

PR #865, opened 2016-05-24
[WIP - Discuss]: Document current API design.

jamestalmage:
Just putting together a list of all the options we pass around, so we can start thinking about how to clean things up.

I think I see some easy wins already.

TODO's:

- [x] document the options passed down to the child process via `fork.js` (currently everything passed to the API gets serialized - but I want to develop a list of what is actually used, so we know what is getting passed around).

- [ ] document the various events we're passing around:

  - [x] events between child and fork.js

  - [ ] events between fork and api

  - [ ] events that get passed to loggers

  - [ ] ???

If anybody wants to take a peice of this, just chime in.

### DIFF
--- a/api-design.md
+++ b/api-design.md
@@ -60,3 +60,64 @@ new Watcher(logger, api, files, sources);
 - **files** - Glob patterns to watch
 
 - **files** - Glob patterns to restrict source watching. Otherwise any change to a file that is not a source causes a watcher rerun.
+
+
+# Events
+
+## IPC Events
+
+- :arrow_up: `no-tests` - `{avaRequired: boolean}`
+
+- :arrow_up: `uncaughtException` - `{exception: serializedError}`
+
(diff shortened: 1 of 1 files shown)


novemberborn:
@avajs/core how interested in this are we? Clearly we've ignored this PR for 4 months now 😉  These kinds of documents are also hard to keep up to date. I'd rather we capture this stuff more clearly in the source code.

sotojuan:
Agreed. At the very least we may do this on 1.0.

jamestalmage:
The point was never really to merge, but to wrap my head around all the events getting passed around, and find ways to simplify it all.

Documenting in the source is fine, but not that helpful for a high level view of how everything glues together. Event transmission and reception happen all over the place.

I'm fine with closing it.

sindresorhus:
I also think we should simplify the event transmission.

PR #866, opened 2016-05-24
add test.failing and add cb to test.serial

SamVerschueren:
Added the declaration for [test.failing](https://github.com/avajs/ava#failing-tests) and fixed `test.serial`. It now allows you to write `test.serial.cb`.

I'm not sure if this is possible?
- `test.cb.serial`
- `test.cb.failing`

But I don't think it is and it is not allowed at the moment.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -55,6 +55,8 @@ export namespace test {
 
 	export function serial(name: string, run: ContextualSerialTest): void;
 	export function serial(run: ContextualSerialTest): void;
+	export function failing(name: string, run: ContextualCallbackTest): void;
+	export function failing(run: ContextualCallbackTest): void;
 	export function cb(name: string, run: ContextualCallbackTest): void;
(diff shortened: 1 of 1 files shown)


sindresorhus:
> I'm not sure if this is possible?

It is. Order doesn't matter. The modifiers just add options to the test. Imagine `test.cb.serial` as just `test({type: 'cb', serial: true})`.

SamVerschueren:
Aha, thanks for the clarification. In this case, other stuff has to be fixed as well.

jamestalmage:
Merging this since I'm cutting a release. 

Typescript users have been living with the stricter ordering for a while now without complaint. This at least allows them to use the new AVA features.

@SamVerschueren If you can follow up with another PR that allows the less strict ordering, that would be great.

SamVerschueren:
@jamestalmage no problem. Will try to add it today.

PR #869, opened 2016-05-25
bump dependencies.

jamestalmage:
arr-diff@3.0.0 causes failures, so skipping for now.


jamestalmage:
@novemberborn ⬆️ - `arr-diff` is only used in watcher. I looked at the API docs for `arr-diff@2.0.0`, and they seem identical to `arr-diff@3.0.0`.

Looks like there was a major rewrite: https://github.com/jonschlinkert/arr-diff/commit/bbf8aefaf161c4fa9dc526b9ca18d3bb8f19d147, so maybe our tests are finding some corner case.

You mind taking a look and figuring out what's gone wrong here?

jamestalmage:
@sindresorhus - We are using `strip-bom@2.0.0` on a Buffer. It looks like `strip-bom@3.0.0` has split the buffer functionality into `strip-bom-buf`. Unfortunately, `strip-bom-buf` uses syntax unsupported on Node `0.12`.

I'm just sticking with the older version for this release, but do we want to do that going forward? Should we file a PR with `strip-bom-buf` using ES5 syntax?

jamestalmage:
Looks like `update-notifier` is the only dependency that didn't break us.

novemberborn:
@jamestalmage do we really need `strip-bom`?

sindresorhus:
@novemberborn Node.js also strip the BOM when you `require` something. I think we need to it match that behavior and ensure we never fail because of a BOM in a file.

PR #873, opened 2016-05-25
Add after.always afterEach.always runners to index.d.ts

rhysd:
`after.always` and `afterEach.always` runners were added to ava in 0.15.  I added them to `index.d.ts`.  I already confirmed that below code can be compiled with new `index.d.ts`.

```typescript
import test from 'ava';

test.afterEach(t => {
    console.log('BEFORE');
});

test.afterEach.always(t => {
    console.log('BEFORE ALWAYS');
});

test.after(t => {
    console.log('AFTER');
});

test.after.always(t => {
    console.log('AFTER ALWAYS');
});
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -15,11 +15,17 @@ export interface Runner {
 	skip: Runner;
 	cb: CallbackRunner;
 }
+export interface AfterRunner extends Runner {
+	always: Runner;
+}
 export interface ContextualRunner {
 	(name: string, run: ContextualTest): void;
 	(run: ContextualTest): void;
 	skip: ContextualRunner;
 	cb: ContextualCallbackRunner;
+}
+export interface ContextualAfterRunner extends ContextualRunner {
+	always: ContextualRunner;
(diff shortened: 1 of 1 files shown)


sindresorhus:
@SamVerschueren @ivogabe :shipit:?

SamVerschueren:
LGTM, except if `test.always.after` should work as well, that will not work. But I have to take a look if swapping keywords can be solved elegantly so I'm ok with a merge and I will solve it in a separate PR together with the others.

ivogabe:
👍, adding `test.always.after` should not be to hard but that can be done later too.

jamestalmage:
Chain ordering will never matter. It uses[`option-chain`](https://github.com/avajs/option-chain) under the hood.

SamVerschueren:
Good to know :)

rhysd:
I think this PR can be merged, right?

jamestalmage:
Thanks @rhysd.

PR #874, opened 2016-05-25
set child process debug port to an available port - fixes #342

### LINKED ISSUE #342: Debugger support

develar:
Continue #613

More robust and maintainable approach was found. Just compute required data before main promise for all files. So, it is simple to add new function in to the existing promise chain. Yeah, I spend a lot of time trying to fix last failed test and decided to rework solution.

> This also doesn't handle the case where the user would do --debug-brk 213 instead of --debug-brk=213.
> https://github.com/avajs/ava/pull/613#discussion_r55246479

This syntax is not supported because node supports only `--debug=`, but not `--debug <port number>`.

```
develar-work:~ develar$ node --debug 12334 foo.js
Debugger listening on port 5858
module.js:442
    throw err;
    ^

Error: Cannot find module '/Users/develar/12334'
```

> Can you add some tests?
> https://github.com/avajs/ava/pull/613#issuecomment-193336125

Tests added. Initially as heavy-weight tests, but I got weird error `Test files must be run with the AVA CLI` if I run with `nyc` (passed without `nyc`). So, I decided to write unit tests.

### DIFF
--- a/test/api.js
+++ b/test/api.js
@@ -1023,16 +1023,20 @@ function generateTests(prefix, apiCreator) {
 	});
 }
 
-function generatePassDebugTests(execArgv) {
+function generatePassDebugTests(execArgv, expectedInspectIndex) {
 	test('pass ' + execArgv.join(' ') + ' to fork', function (t) {
-		t.plan(3);
+		t.plan(expectedInspectIndex === -1 ? 3 : 2);
 
 		var api = new Api({testOnlyExecArgv: execArgv});
 		return api.computeForkExecArgs(['foo.js'])
 			.then(function (result) {
 				t.true(result.length === 1);
-				t.true(result[0].length === 1);
-				t.true(/--debug=\d+/.test(result[0][0]));
+				if (expectedInspectIndex === -1) {
+					t.true(result[0].length === 1);
+					t.true(/--debug=\d+/.test(result[0][0]));
+				} else {
+					t.true(/--inspect=\d+/.test(result[0][expectedInspectIndex]));
+				}
 			});
 	});
 }
@@ -1049,8 +1053,18 @@ function generatePassDebugIntegrationTests(execArgv) {
 	});
 }
 
-generatePassDebugTests(['--debug=0']);
-generatePassDebugTests(['--debug']);
+generatePassDebugTests(['--debug=0'], -1);
+generatePassDebugTests(['--debug'], -1);
 
+generatePassDebugTests(['--inspect=0'], 0);
+generatePassDebugTests(['--inspect'], 0);
+
+generatePassDebugTests(['--inspect=0', '--debug-brk'], 0);
+generatePassDebugTests(['--inspect', '--debug-brk'], 0);
+
+generatePassDebugTests(['--debug-brk', '--inspect=0'], 1);
+generatePassDebugTests(['--debug-brk', '--inspect'], 1);
+
+// --inspect cannot be tested because released node doesn't support it
 generatePassDebugIntegrationTests(['--debug=0']);
-
+generatePassDebugIntegrationTests(['--debug']);
--- a/test/api.js
+++ b/test/api.js
@@ -1022,3 +1022,20 @@ function generateTests(prefix, apiCreator) {
 		]);
 	});
 }
+
+function generatePassDebugTests(execArgv) {
+	test('pass ' + execArgv.join(' ') + ' to fork', function (t) {
+		t.plan(3);
+
+		var api = new Api({testOnlyExecArgv: execArgv});
+		return api.computeForkExecArgs(['foo.js'])
+			.then(function (result) {
+				t.true(result.length === 1);
+				t.true(result[0].length === 1);
+				t.true(/--debug=\d+/.test(result[0][0]));
+			});
+	});
+}
+
+generatePassDebugTests(['--debug=0']);
+generatePassDebugTests(['--debug']);
--- a/test/api.js
+++ b/test/api.js
@@ -1037,5 +1037,20 @@ function generatePassDebugTests(execArgv) {
 	});
 }
 
+function generatePassDebugIntegrationTests(execArgv) {
+	test('pass ' + execArgv.join(' ') + ' to fork', function (t) {
+		t.plan(1);
+
+		var api = new Api({testOnlyExecArgv: execArgv});
+		return api.run([path.join(__dirname, 'fixture/debug-arg.js')])
+			.then(function (result) {
+				t.is(result.passCount, 1);
+			});
+	});
+}
+
 generatePassDebugTests(['--debug=0']);
 generatePassDebugTests(['--debug']);
+
+generatePassDebugIntegrationTests(['--debug=0']);
+
--- a/api.js
+++ b/api.js
@@ -137,13 +137,26 @@ Api.prototype._run = function (files, _options) {
 Api.prototype.computeForkExecArgs = function (files) {
 	var execArgv = this.options.testOnlyExecArgv || process.execArgv;
 	var debugArgIndex = -1;
+
+	// --debug-brk is used in addition to --inspect to break on first line and wait
 	execArgv.some(function (arg, index) {
-		if (arg === '--debug' || arg === '--debug-brk' || arg.indexOf('--debug-brk=') === 0 || arg.indexOf('--debug=') === 0) {
+		if (arg === '--inspect' || arg.indexOf('--inspect=') === 0) {
 			debugArgIndex = index;
 			return true;
 		}
 		return false;
 	});
+
+	var isInspect = debugArgIndex !== -1;
+	if (!isInspect) {
+		execArgv.some(function (arg, index) {
+			if (arg === '--debug' || arg === '--debug-brk' || arg.indexOf('--debug-brk=') === 0 || arg.indexOf('--debug=') === 0) {
+				debugArgIndex = index;
+				return true;
+			}
+			return false;
+		});
+	}
 
 	if (debugArgIndex === -1) {
 		return Promise.resolve([]);
@@ -153,7 +166,7 @@ Api.prototype.computeForkExecArgs = function (files) {
 		.then(function (ports) {
 			return ports.map(function (port) {
 				var forkExecArgv = execArgv.slice();
-				var flagName = '--debug';
+				var flagName = isInspect ? '--inspect' : '--debug';
 				var oldValue = forkExecArgv[debugArgIndex];
 				if (oldValue.indexOf('brk') > 0) {
 					flagName += '-brk';
--- a/api.js
+++ b/api.js
@@ -11,6 +11,7 @@ var findCacheDir = require('find-cache-dir');
 var debounce = require('lodash.debounce');
 var ms = require('ms');
 var AvaFiles = require('ava-files');
+var getPort = require('get-port');
 var AvaError = require('./lib/ava-error');
 var fork = require('./lib/fork');
 var CachingPrecompiler = require('./lib/caching-precompiler');
@@ -46,7 +47,7 @@ function Api(options) {
 util.inherits(Api, EventEmitter);
 module.exports = Api;
 
-Api.prototype._runFile = function (file, runStatus) {
+Api.prototype._runFile = function (file, runStatus, execArgv) {
 	var hash = this.precompiler.precompileFile(file);
 	var precompiled = {};
 	precompiled[file] = hash;
@@ -55,7 +56,7 @@ Api.prototype._runFile = function (file, runStatus) {
 		precompiled: precompiled
 	});
 
-	var emitter = fork(file, options);
+	var emitter = fork(file, options, execArgv);
 
 	runStatus.observeFork(emitter);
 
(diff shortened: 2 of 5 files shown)


jamestalmage:
In addition to unit tests, you should have a few integration tests. We can help figure out why NYC causes a failure.

develar:
>  you should have a few integration tests

Integration test added. Maybe on CI it will be passed, but on my machine I got "Test files must be run with the AVA CLI:" from child stderr if use `nyc`.

develar:
CI failed " not ok AvaError: test/fixture/debug-arg.js exited with a non-zero exit code: 1"  Hmm... it works for me — I use fork of ava in my project and can debug tests in IntelliJ IDEA. 

Something is wrong with nyc, but I don't see how it it is connected. 

If you will have no clue, I will investigate what's wrong  with nyc.

jamestalmage:
As I recall, in linux, ports below some number (1024?) are "privileged", and need some special permissions to run. Try bumping your tests to use something higher.

develar:
> Try bumping your tests to use something higher.

Port here doesn't matter — for child we use random free port.

jamestalmage:
Then what does `--debug=0` mean?

jamestalmage:
The build runs fine locally on OSX. It's either Linux or Travis specific.

Travis could be blocking this somehow.

develar:
> The build runs fine locally on OSX.

I use OS X and it is failed with nyc.

> Then what does --debug=0 mean?

Args of the main process. In the tests we don't start main process, we start child only (if I am not wrong). Purpose of this fix is — replace port in the main process args to another free port.

Ok, I will investigate what's wrong with nyc.

develar:
AppVeyor build succeeded only because `npm run test-win` — in the `test-win` nyc is not called.

jamestalmage:
Ahhh. Yes. This is almost certainly a spawnwrap bug then.

jamestalmage:
See: https://github.com/jamestalmage/__nyc-ipc-argv-bug

develar:
Sorry for delay, I will investigate it soon. Currently I am working to support new v8 inspector in the IntelliJ IDEA. `--inspect` has the same problem — no way to ask node to use any port.

jamestalmage:
@develar  - The NYC bug is fixed. Any chance of finalizing this soon?

develar:
Rebased — tests passed.
`--inspect` and combination `--inspect` + `--debug-brk` supported.

test/api.js
```diff
@@ -1022,3 +1022,49 @@ function generateTests(prefix, apiCreator) {
 		]);
 	});
 }
+
+function generatePassDebugTests(execArgv, expectedInspectIndex) {
+	test('pass ' + execArgv.join(' ') + ' to fork', function (t) {
+		t.plan(expectedInspectIndex === -1 ? 3 : 2);
+
+		var api = new Api({testOnlyExecArgv: execArgv});
```
novemberborn: This should stub `process.execArgv` rather than needing a test-only option.

novemberborn:
This looks pretty good @develar!

I wonder if there's anybody who'd like to install this PR locally to debug their tests?

test/api.js
develar: > This should stub

Not quite understand, do you mean that we should replace `process.execArgv` during test? I think, it is not acceptable to touch global state.

novemberborn: > do you mean that we should replace process.execArgv during test?

Yes.

> I think, it is not acceptable to touch global state.

True, but tests are special.

Something I've done in other projects is to have a `process` module which exports the global `process`, and then stubbing that through Proxyquire. But let's not do that unless we really have to.

develar: > True, but tests are special.

I think, some day AVA tests will be also executed in parallel, so, it is not wise to сhange properties in the global `process` object.

novemberborn: > I think, some day AVA tests will be also executed in parallel, so, it is not wise to сhange properties in the global process object.

We can deal with it then.

jamestalmage:
Hmm.

I kinda think we're going about this the wrong way. Instead of parsing `process.execArgv`, let's use `process.argv`:

```
ava --debug=<PORT>
```

If a user is debugging their tests, they aren't going to care about attaching a debugger to the parent process.

Using the debug flag would imply `--concurrency=1`, so multiple processes don't compete for the same port.

Then the debug client can connect to a single port, no parsing `stderr` for port numbers (which seems unreliable IMO).

IMO, If you really want to do it this way (find random open ports and print to stderr for parsing). Then that should be implemented as a separate tool using `spawn-wrap` (that is what `nyc` uses under the hood to give us coverage across many processes). That allow much wider reuse.

test/api.js
jamestalmage: Even when we do self test with AVA, the parallel processes will have their own `process.execArgv`. Test isolation FTW!

develar:
@jamestalmage Why? User just run debug in the IDE. Or run debug using standard node args. That's all. Nothing more. No need to reinvent the wheel and introduce new tool-specific API. User just use standard tools and way to debug.

JetBrains IDE — we reimplement multiproces debug specially for AVA. It just works. No multiple debug tabs anymore, it is like chrome workers (you can select "thread" using combo box if need).  You just hit "debug" and it works (in context runner for AVA not yet implemented, but it is another task).

Chrome using `--inspect` — copy url and debug. Yes, it is not seamless as in the IDE, but... IDE is IDE. I think, you told exactly about it — simplify debug using some Ava-specific tool.

But, at first, I think, we should support debug using standard tools and way.

jamestalmage:
> we reimplement multiprocess debug specially for AVA

Is that available for us to take a look at? It sounds kinda awesome. Having multiple processes running at a time sounds cool, but I'm not sure how useful it is. Maybe you could debug two test-files that interleave disk operations in some way with clever breakpoints? IDK.

--

Either way, I don't think we want them to use `execArgv`, especially with `--debug-brk. Using execArgv with `--debug-brk` causes it to pause the main process. So you need to manually continue it each time before child processes are launched.

```
node --debug node_modules/.bin/ava tests/foo.js

vs.

ava --debug test/foo.js
```

The second is more convenient.

See #929 - It works really well with `node-inspector` right now, and the CLI command `node debug localhost:<PORT>` works OK (you need to manually reconnect between tests by typing `run`). I tried this PR out locally, and I didn't find it nearly as convenient using `node-inspector` as #929. That might be me misunderstanding how this supposed to work (some basic instructions on how I should use this with `node-inspector` or similar would be good).

develar:
> Is that available for us to take a look at?

Yes, these changes available since 2016.1 release. You can try 2016.2 EAP — [IDEA](https://confluence.jetbrains.com/display/IDEADEV/IDEA+2016.2+EAP) or [WebStorm](https://confluence.jetbrains.com/display/WI/WebStorm+EAP). `--inspect` not yet supported for child process (it is in the unreleased 2016.3 (164 branch)).

Video: https://youtu.be/C75UwuZXI98

> Having multiple processes running at a time sounds cool, but I'm not sure how useful it is.

I use `--match` if I want to debug test. I run `npm run test` in terminal and then debug failed tests in the IDE.

> Using execArgv with --debug-brk causes it to pause the main process.
> So you need to manually continue it each time before child processes are launched.

IDE uses this break to set breakpoints, find sourcemaps and then continues debug automatically.

develar:
#929 is about how to simplify AVA debug in the CLI using special ava specific tests. This PR is to make AVA debug working using standard ways (i.e. IDE vendor should not explicitly support AVA). Strictly speaking, this PR is a workaround of https://github.com/nodejs/node/pull/5025

test/api.js
develar: > the parallel processes will have their own process.execArgv

But not tests in the same file ;)

Ok, up to you, I removed option and replace `process.execArgv`.

develar:
> Maybe you could debug two test-files that interleave disk operations in some way with clever breakpoints? IDK.

You can use conditional breakpoints.

test/api.js
jamestalmage: > But not tests in the same file ;)

True. Though you can always just do `test.serial` for the few that would interfere.

jamestalmage:
I've gotten this working. Other maintainers should give it a spin as well.


I have downloaded Webstorm EAP2016.2, and used it with this PR. I chose to just use `./cli.js` to run two fixtures from AVA's own test suite.

Here were my steps:

1. Download and install WebStorm EAP2016.2

2. Added breakpoints to `test/fixture/es2015.js` and `test/fixture/async-await` (you can add wherever you want - just change your config below). You can set breakpoints by adding a `debugger;` line or clicking in the gutter of the IDE.

3. Click "Edit Configurations" from the drop down in the middle top:

  screenshot 2016-06-20 19 03 29

4. Click the `+` icon in that dialog to create a new task, and select a Node.js task.

5. Configure task settings:

  screenshot 2016-06-20 18 34 31

6. Run the task using the debug icon (looks like a beetle near the top - where you clicked "Edit Configurations") .

  I got to this screen:
  screenshot 2016-06-20 18 39 39

It works pretty similar to DevTools debugger, but you can use the drop down to select the running process (the above screenshot is showing debugger from pid 40360).

develar:
BTW, AVA developers can get opensource WebStorm licence for free.

jamestalmage:
> BTW, AVA developers can get open source WebStorm/other IDE license for free.

Yep. The maintainers are already signed up.

develar:
@jamestalmage Checked, cool, you already have open source licence. Feel free to ping me directly in case of any JetBrains IDE issue.

develar:
Now my turn to ping ;)

sindresorhus:
@develar Doesn't https://github.com/avajs/ava/pull/929 supersede this? Or are both needed? It's a bit unclear how they relate.

develar:
@sindresorhus This PR doesn't add ava-specific implementation — just fix nodejs "bug". I.e. IDE can debug AVA using standard non-specific way. #929 only for users, who want to use terminal to debug (simplification).

develar:
Branch rebased to fix conflicts.

develar:
Conflicts resolved again :)

sindresorhus:
This looks good to me.

-

Note to self: Open issue about removing this eventually when the Node.js version we target has the fix (https://github.com/nodejs/node/pull/5025).

develar:
@sindresorhus Sorry, forget to mention — https://github.com/nodejs/node/pull/5025 only about `--debug-port=0`, but currently no issue about `--inspect` — please see https://github.com/nodejs/node/pull/5025#issuecomment-225323248 "it should be possible to extend this to --inspect after this PR lands." Currently, `--inspect` should be used to achieve the best user debug experience (yeah, nodejs 7 is required).

sindresorhus:
Landed. Finally :)

Thanks for doing the hard work here @develar and persevering through this long review.

korsmakolnikov:
This doesn't work for me. I'm getting an 'Error: listen EADDRINUSE :::39903' exception.

develar:
@korsmakolnikov Please specify IDE version, do you use ava from sources?

KeKs0r:
@develar It does not work for me. I am Using Webstorm
```
WebStorm 2016.2
Build #WS-162.1121.31, built on July 9, 2016
JRE: 1.8.0_76-release-b216 x86_64
JVM: OpenJDK 64-Bit Server VM by JetBrains s.r.o
```
And Ava from master, but my debugger just never stops. I was trying with single file debugging (without match):
image

Not sure why the debugger never stops. It also does not terminate.

develar:
@KeKs0r Please use [WebStorm 2016.3 EAP](https://confluence.jetbrains.com/display/PhpStorm/PhpStorm+Early+Access+Program)

mightyiam:
Is this documented anywhere, please? Does it need to be, or will it simply work with anything that debugs node, like Chrome DevTools?

develar:
@mightyiam https://github.com/avajs/ava/blob/master/docs/recipes/debugging-with-webstorm.md

mightyiam:
How likely is this to ever work in Chrome DevTools?

sindresorhus:
@mightyiam It's planned with https://github.com/avajs/ava/pull/929, but that PR stalled.

Workaround for now. Put a `debugger;` statement somewhere in your code and run:

```
$ node --inspect node_modules/ava/profile.js test.js
```

Where `node_modules/ava/profile.js` is the relative path to that file.

develar:
@mightyiam In addition to `--inspect` you need to pass `--debug-brk` So, Chrome will stop on first line.

> This will output a link that you can open in Chrome. After opening that link, the Chrome Developer Tools will be displayed, and a breakpoint will be set at the first line of the Jest CLI script (this is done simply to give you time to open the developer tools and to prevent Jest from executing before you have time to do so). Click the button that looks like a "play" button in the upper right hand side of the screen to continue execution. When Jest executes the test that contains the debugger statement, execution will pause and you can examine the current scope and call stack.

Docs from Jest is true and correct for AVA as well. So, you can debug in Chrome DevTools also.

mightyiam:
Thanks!

richchurcher:
Just because I couldn't find this anywhere else: when using VS Code as a debugger, you can coerce AVA to behave by using @sindresorhus 's suggestion as part of a `launch.json` config:

```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "type": "node2",
            "request": "launch",
            "name": "Launch Program",
            "program": "${workspaceRoot}/tests/index.test.js",
            "cwd": "${workspaceRoot}",
            "sourceMaps": true,
            "runtimeArgs": [
                "${workspaceRoot}/node_modules/ava/profile.js"
            ]
        }
    ]
}
```

Hardware-Hacks:
Worth mentioning that this setup worked for me but the one bit I had missing was that I forgot to install the webstorm debug helper in chrome https://chrome.google.com/webstore/detail/jetbrains-ide-support/hmhgeddbohgjknpmjagkdomcpobmllji?hl=en-GB

dohomi:
@develar @mightyiam did you guys managed to test AVA tests in Webstorm with the current 2017.2? I only receive an error that the port is already in use. I use the configuration mentioned in this document: https://github.com/avajs/ava/blob/master/docs/recipes/debugging-with-webstorm.md

melisoner2006:
I was also having problem with debugging ava unit tests with Webstorm 2017.2. It didn't stop at the break points. Adding `inspect-brk` to node options fixed the problem. [Here](https://stackoverflow.com/questions/45494136/webstorm-doesnt-stop-at-breakpoint-when-debugging-ava-tests)'s my SO question & answer. Thanks @develar  and @mightyiam  for mentioning the node options.

novemberborn:
@melisoner2006 would you be interested in opening a PR to fix the recipe?

melisoner2006:
@novemberborn I'd love to. I'll get on it now.

Gregoyle:
@dohomi Did you ever get this working? I'm having a hard time getting the WS debugger to work. It could be related to the fact that our project is using babel. @melisoner2006 Was your project ES6 or something else?

melisoner2006:
@Gregoyle Yes, my project was ES6.

PR #876, opened 2016-05-25
fix(cli): Remove default files from CLI

kentcdodds:
WIP: Not sure where to put tests/how to test this.


kentcdodds:
I hope this could be patch-released soonish because the feature was advertized as part of the most recent release but does not actually work :-/

jamestalmage:
Yes. I want to land this ASAP (i.e. add some tests then let's publish).

Here is the test from `ava-files` I was thinking you should replicate: https://github.com/avajs/ava/blob/master/test/ava-files.js#L99-L106

Also, see [these tests](https://github.com/avajs/ava/pull/863/files#diff-583d616f4b01bf1fa153e71c0cef7de7R179). Those launch a new process for the CLI (which in turn launches test processes), so they aren't fast tests. We could do one or two just to verify it's all glued together correctly, but we should build up a more thorough set of faster tests that fully exercise `ava-files` directly.

For now, let's just land one of each and ship a patch release.

kentcdodds:
I updated with one test. I think it's pretty solid. Let me know if you want an integration test with the CLI as well.

jamestalmage:
Linter errors!

kentcdodds:
How embarrassing!

kentcdodds:
Fixed!

test/ava-files.js
```diff
@@ -104,3 +104,29 @@ test('findFiles - does not return duplicates of the same file', function (t) {
 		t.end();
 	});
 });
+
+test('findFiles - finds the correct files by default', function (t) {
+	var avaFiles = new AvaFiles(['**/ava-files/default-patterns/**']);
```
jamestalmage: Argh. I'm realizing this is actually kinda hard to test.

This will actually find every file in the `default-patterns` folder.

Instead of this argument, can you just `process.chdir()` to the `default-patterns` folder, and drop the constructor argument?

You are technically overriding the default here.

#863 introduce `cwd` and `resolveTestsFrom` options, that would make this doable without changing the directory.

jamestalmage: Also, you should sprinkle in a few files that should NOT be found (and ensure they are not).

kentcdodds:
I'm not sure what's wrong with the windows build. It does look my test is causing the issue, but I'm not sure why...

test/ava-files.js
```diff
@@ -104,3 +104,29 @@ test('findFiles - does not return duplicates of the same file', function (t) {
...
+		'sub/directory/bar.test.js',
+		'test-foo.js',
+		'test.js',
+		'test/baz.js'
+	];
+
+	avaFiles.findTestFiles().then(function (files) {
+		var allFilesFound = filesToFind.every(function (fileToFind) {
+			return files.some(function (file) {
+				return endsWith(file, fileToFind);
```
jamestalmage: Use [`slash`](https://github.com/sindresorhus/slash) to convert the results from `\` results, to `/` results (I think).

test/ava-files.js
kentcdodds: I can do that. Unfortunately I've run out of time. Someone can feel free to branch off of my branch and make the improvements in another PR if it needs to be fixed sooner than I can get to it.

test/ava-files.js
kentcdodds: :+1:

test/ava-files.js
jamestalmage: Sorry Kent, 
I see why you did it this way - It's exactly like the test I told you to copy. I'm updating your PR now. Expect a new PR and release soon.

kentcdodds: Thanks. It's family time for me now. Sorry I didn't quite get to it.

jamestalmage:
Closing in favor of #878 

Thanks @kentcdodds

PR #880, opened 2016-05-26
[WIP] Refactor reporters

vdemedes:
This PR removes Logger and refactors reporters.

### Reason

Logger is basically a proxy for reporters and does not do anything useful at all. In this PR, reporters are refactored to subscribe only to events they support. That eliminates the need for a Logger.

`Logger#exit()` was also moved back to `cli.js`.

### Updated behavior

Each reporter has `init()` method, which accepts a `RunStatus`. This method is responsible for cleaning up previous listeners, attaching new ones and possibly resetting (when in watch mode).

### Todo

- [x] Refactor mini reporter
- [x] Make watch mode work with new reporters
- [ ] Update watcher tests


lib/reporters/tap.js
```diff
@@ -16,11 +16,24 @@ function TapReporter() {
...
+TapReporter.prototype.init = function (status) {
+	this.status = status;
+	this.write('TAP version 13');
+
+	status.on('test', this.test);
+	status.on('error', this.unhandledError);
+	status.on('finish', this.finish);
+};
+
 TapReporter.prototype.start = function () {
```
nfcampos: forgot to remove this one

lib/reporters/tap.js
```diff
@@ -75,24 +88,25 @@ TapReporter.prototype.unhandledError = function (err) {
...
 	}
 
-	output.push('# fail ' + (runStatus.failCount + runStatus.rejectionCount + runStatus.exceptionCount), '');
+	output.push('# fail ' + (status.failCount + status.rejectionCount + status.exceptionCount), '');
 
-	return output.join('\n');
+	this.write(output.join('\n'));
 };
 
 TapReporter.prototype.write = function (str) {
```
nfcampos: why do we need TapReporter.prototype.stdout and TapReporter.prototype.stderr if they're never used here?

jamestalmage:
Hmm. I really liked the Logger facade. It makes testing a lot easier. Now you have to employ a workaround, replacing the `write` method.

What does doing this get us?

I was thinking of some functionality we should actually move IN to logger. Specifically the `explicitPrefixes` option we use to manipulate whether we prefix titles with the file name.

Unless you've got a really good reason for doing so, I don't think we should get rid of it.

vdemedes:
> I really liked the Logger facade. It makes testing a lot easier.

I wouldn't say tests changed much and it became "a lot" harder to test reporters. It's just as easy to test their output.

> What does doing this get us?

Less code in the AVA repository, especially unneeded code. As I mentioned in the first post, Logger did not do anything useful, except checking for method existence and falsy return value. It's an unnecessary layer. Reporters, that listen to API events are much more "javascript", than previous objective-c delegates-like code.

> I was thinking of some functionality we should actually move IN to logger. Specifically the explicitPrefixes option we use to manipulate whether we prefix titles with the file name.

This sounds more like a helper function, rather than a method belonging to Logger - a class, that displays output.

lib/reporters/tap.js
vdemedes: True, should be removed.

lib/reporters/tap.js
vdemedes: Thanks!

jamestalmage:
Except this change adds 100 lines of code, without any change in functionality, or additional tests.

vdemedes:
I meant less modules.

vdemedes:
While working on browser support, I discovered another argument in favor of this PR: reporters can have different output destinations. Not all reporters just write to stderr.

With browser support, there will be a reporter, that renders DOM elements. If we keep a centralized Logger, it will have to take care of detecting browser environment and rendering DOM elements.

This PR definitely brings more flexibility and makes reporting future-proof.

jamestalmage:
Wouldn't you just follow the same pattern we do now with other reporters?

Build the message in the given event specific function, then render it to the output destination in `write`?

novemberborn:
I like this @vdemedes! The watcher work required very tight coupling with the logger and specific reporters. Abstracting that to an event interface might restore some sanity.

jamestalmage:
What if we do something like:

```js
function reporterBind(reporter, emitter, events) {
  events.forEach(function (event) {
    emitter.on(event, function () {
      var output = reporter[event].apply(reporter, arguments);
      reporter.write(output);
    });
  });
}
```

Then within tap reporter (for instance):

```js
TapReporter.prototype.init = function (status) {
    reporterBind(this, status, ['test', 'error', 'finish']);
};
```

That allows the individual methods to still return strings (for easy testing), but also allows them to maintain control of which events they register for, and allows the elimination of Logger. Make sense?

vdemedes:
Does not make sense to me, because it's basically a wrapper over `emitter.on`. Reporters can just use `.on()` to listen to events they expect, why provide a function to do the exact same task? As I mentioned before, tests did not become any harder after this PR's changes.

vdemedes:
I'd really like this PR to go in. @jamestalmage is -1 on this, @novemberborn is +1.

@sindresorhus @sotojuan we'd love your feedback on this.

jamestalmage:
Still has a WIP title, btw

vdemedes:
@jamestalmage I don't see the point in going forward, when we don't know whether this is going in or not.

jamestalmage:
So, I looked into it more, and I have changed my mind. I think I'm 👍 now. (Sorry for all the noise @vdemedes). 

A few additional thoughts:
- Should we pass `process.stdout`, and `process.stderr` in as constructor arguments? You'd mentioned being able to redirect output in the browser. Also might make tests easier.
- Should we be unregistering all our listeners in watch mode? Theoretically they should never emit another event after a run finishes, so maybe it's no big deal. I don't think there are any garbage collection implications from not unregistering.

cli.js
```diff
@@ -194,3 +183,15 @@ if (cli.flags.watch) {
 			});
 		});
 }
+
+function exit (code) {
```
sindresorhus: Put this at the top. Better for it to be defined before it's used.

sindresorhus:
Huge 👍

vdemedes:
PR updated with a refactored mini reporter and passing tests for it. Code comments on this PR were fixed as well. Watch mode also works, but tests weren't adapted to the latest changes.

@novemberborn I'm going on vacation today, so I won't be able to take care of watcher tests. Could you please pick this up from now on?

sindresorhus:
Closing this temporarily as Vadim doesn't have time to finish it up right now. We can reopen at some later point. We definitely want it.

Issue #870, opened 2016-05-25
Type definition tests


jamestalmage:
It would be good to have a set of typescript fixtures that exercised the API.
The test would just be running the typescript compiler and ensuring it didn't complain about the syntax.

That seems to be how they do it for [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped).

@SamVerschueren @ivogabe - That sound like a good idea?

SamVerschueren:
Yes, I was thinking the same yesterday "we should have tests for the type definitions". That would definitely be an improvement to somewhat validate them as the definitions getting more complex and more features are added.

jamestalmage:
When we add new API features we can use `.failing` to mark them until a TypeScript contributor updates the definition...

Except we don't test with AVA 😭 .

sindresorhus:
@jamestalmage We can use skipped tests.

unional:
Chiming in. Just shape test is not enough:
https://github.com/typings/generator-typings#about-writing-tests-for-typings

jamestalmage:
> Just shape test is not enough:

It would almost certainly be better than the nothing we have now though!

unional:
> It would almost certainly be better than the nothing we have now though!

:smile: of course.

unional:
Oh, by the way, if you lobby on this... :smile:
https://github.com/Microsoft/TypeScript/issues/7661

ivogabe:
This seems to be related too: Microsoft/TypeScript#9281

sindresorhus:
We also need this for a forthcoming Flow type definition: https://github.com/avajs/ava/pull/1007

novemberborn:
We have… something, now. Closing since it's not clear to me whether there are any deficiencies in our current approach.

PR #884, opened 2016-05-27
Auto generate TypeScript definition to allow chaining

### LINKED ISSUE #871: Update typescript `.d.ts` file to any chaining order.
From #866. The `.d.ts` file restricts which order you can chain methods. It can be loosened to allow any ordering.

// @SamVerschueren @ivogabe

ivogabe:
I've implemented a generator that will create the TS definitions, with full support for chaining. Writing these type definitions by hand would be a lot of work or inaccurate. See the first comment in `types/generator.js` for details on how these definitions are created.

Fixes #871. Tests for this would be useful too, see #870.

@jamestalmage @SamVerschueren What do you think?


novemberborn:
Haven't really looked at the generate script, but I think it'd be fine for it to use ES2015 (and thus require Node 6). It's something we'd have to run locally and then check in the results.

types/generate.js
```diff
@@ -0,0 +1,114 @@
...
+// `after.cb` is fully written, and `cb.after` is emitted as an alias
+// using `typeof after.cb`.
+
+var path = require('path');
+var fs = require('fs');
+
+var base = fs.readFileSync(path.join(__dirname, 'base.d.ts')).toString();
+
+// All suported function names
+var allParts = ['serial', 'before', 'after', 'skip', 'todo', 'failing', 'only', 'beforeEach', 'afterEach', 'cb', 'always'];
```
SamVerschueren: It would be nice if it was exported from within the library https://github.com/avajs/ava/blob/master/lib/runner.js#L11. Then if something changes, the `generate` script shouldn't be adjusted.

But then again, should libraries export stuff because of "documentation" purposes.

types/generate.js
```diff
@@ -0,0 +1,114 @@
...
+fs.writeFileSync(path.join(__dirname, 'generated.d.ts'), output);
+
+// Generates type definitions, for the specified prefix
+// The prefix is an array of function names
+function generatePrefixed(prefix) {
+	var output = 'export namespace ' + ['test'].concat(prefix).join('.') + ' {\n';
+	
+	var children = '';
+	var empty = true;
+	for (var i = 0; i < allParts.length; i++) {
```
SamVerschueren: I would use `allParts.forEach()` here.

types/generate.js
```diff
@@ -0,0 +1,114 @@
...
+			} else {
+				var type = testType(parts);
+				output += '\t' + writeFunction(part, 'name: string, implementation: ' + type);
+				output += '\t' + writeFunction(part, 'implementation: ' + type);
+			}
+		}
+		
+		children += generatePrefixed(parts);
+	}
+	if (empty) {
```
SamVerschueren: I would set the `output` variable at the top to an empty string.

```js
if (output === '') {
    return children;
}

return 'export namespace ' + ['test'].concat(prefix).join('.') + ' {\n' + output + children;
```

And drop the `empty` variable

types/generate.js
```diff
@@ -0,0 +1,114 @@
...
+		output = '';
+	} else {
+		output += '}\n';
+	}
+	return output + children;
+}
+function writeFunction(name, args) {
```
SamVerschueren: Newline above this one

types/generate.js
```diff
@@ -0,0 +1,114 @@
...
+	return 'export function ' + name + '(' + args + '): void;\n';
+}
+
+function verify(parts, asPrefix) {
+	var has = arrayHas(parts);
+	if (has('only') + has('skip') + has('todo') > 1) return false;
```
SamVerschueren: This is quite weird. You add up booleans? Why not `has('only') || has('skip') ...`?

ivogabe: This line checks that at least two of them are true, and that's not very easy to do with `||`.

types/generate.js
ivogabe: 👍

types/generate.js
ivogabe: As @novemberborn suggested, it would be even better to use a `for of` loop here

types/generate.js
ivogabe: Sounds like a good idea, it can be really annoying that the type definitions are not up to date.

ivogabe:
I've addressed most of the comments and fixed the linting issues. Can you take another look?

types/generate.js
SamVerschueren: Although it works, it feels kinda hacky. I think I'd rather prefer something like

```js
has('only', 'skip', 'todo').filter(Boolean).length
```

Or something like that.

types/generate.js
```diff
@@ -0,0 +1,122 @@
+// TypeScript definitions are generated here.
```
SamVerschueren: 'use strict'

lib/runner.js
```diff
@@ -198,3 +198,6 @@ Runner.prototype.run = function (options) {
 
 	return Promise.resolve(this.tests.build(this._bail).run()).then(this._buildStats);
 };
+
+
+Runner.chainableMethods = chainableMethods.chainableMethods;
```
sindresorhus: Should be underscored as it's not something we'd like people to use.

types/generate.js
```diff
@@ -0,0 +1,122 @@
...
+// `after.cb` is fully written, and `cb.after` is emitted as an alias
+// using `typeof after.cb`.
+
+const path = require('path');
+const fs = require('fs');
+const runner = require('../lib/runner');
+
+const arrayHas = parts => part => parts.includes(part);
+
+const base = fs.readFileSync(path.join(__dirname, 'base.d.ts')).toString();
```
sindresorhus:
```js
const base = fs.readFileSync(path.join(__dirname, 'base.d.ts'), 'utf8');
```

types/generate.js
```diff
@@ -0,0 +1,122 @@
...
+	for (const part of allParts) {
+		const parts = [...prefix, part];
+
+		if (prefix.includes(part) || !verify(parts, true)) {
+			// Function already in prefix or not allowed here
+			continue;
+		}
+
+		// Check that `part` is a valid function name.
+		// `always` is a valid prefix, for instance of `always.after`,
```
sindresorhus: @jamestalmage I this correct? I thought any of the modifiers could be the function.

sindresorhus:
Can you call the file `generate.js` => `make.js` instead and add a npm run script for it, like `"make-ts": "node types/make.js"`.

-

And Travis is failing.

SamVerschueren:
I would make the file executable. './make.js'

lib/runner.js
```diff
@@ -198,3 +198,5 @@ Runner.prototype.run = function (options) {
 
 	return Promise.resolve(this.tests.build(this._bail).run()).then(this._buildStats);
 };
+
+Runner._chainableMethods = chainableMethods.chainableMethods;
```
jamestalmage: Should we wrap these options in a generator function? So we aren't exporting a mutable copy?

types/make.js
```diff
@@ -0,0 +1,124 @@
+'use strict';
```
jamestalmage: Shebang line and make it executable?

Does that even work on Windows?

SamVerschueren: 👍 on this. I believe it will work as `package-cli` stuff also works...

types/generate.js
jamestalmage: Anything can be a function. Mix and match however you want.

`always.after` and `after.always` are the same thing.

You could technically do `always.always.always.after` - it would still be treated the same way.

The only thing disallowed is certain combinations:

Only one of `todo`, `skip`, `failing` should be allowed in the chain.

`always` only combines with the `after(Each)` hooks, etc.

Technically, the JavaScript allows any combination (you will never get a "not a function" TypeError) - we just validate the metadata to make sure the combinations make sense.

lib/runner.js
sindresorhus: @jamestalmage Doesn't matter. It's for internal use only.

types/make.js
sindresorhus: Haven't tried, but I don't think it will work. The reason you can use binaries in dependencies is that when npm install binaries it creates a "shim" on Windows that makes it behave like Unix.

jamestalmage: Yeah that was my concern. Let's just leave it as is.

jamestalmage:
Are we good here? This all looks good to me. @SamVerschueren - should I merge?

SamVerschueren:
I did not test the output though but generally everything looks good. If you want me to test it more thoroughly, I'm ok with that. If you want to merge, that's ok as well. We can always finetune in follow up commits.

Great work @ivogabe!

jamestalmage:
Should we be storing `generated.d.ts` in the repo?

Why not add it to `.gitignore` and create it in a prepublish script?

SamVerschueren:
Good point. Don't see any downside of doing that.

ivogabe:
I usually don't add generated files to a repo, though I had two reasons now: it's still possible to install AVA from GitHub, and you can easily see what changed when you update the generator script.

jamestalmage:
I'd prefer ignoring it if possible.

It would be awesome if npm had a "installed from GitHub" hook

SamVerschueren:
Can it be executed on `postinstall` maybe? This would overwrite the `generated.d.ts` shipped with `npm` though.

ivogabe:
I've added it to `.gitignore`. `postinstall` won't work, since the generator script requires NodeJS 6. I think that TS users already know that it's usually not possible to install a package from GitHub. Do you agree?

jamestalmage:
>  I think that TS users already know that it's usually not possible to install a package from GitHub.

I say we start with it this way, and see if there are lots of complaints.

package.json
```diff
@@ -33,18 +33,20 @@
...
+  "typings": "types/generated.d.ts",
   "engines": {
     "node": ">=0.10.0"
   },
   "scripts": {
     "test": "xo && nyc --cache --reporter=lcov --reporter=text tap --no-cov --timeout=150 test/*.js test/reporters/*.js",
     "test-win": "tap --no-cov --reporter=classic --timeout=150 test/*.js test/reporters/*.js",
-    "visual": "node test/visual/run-visual-tests.js"
+    "visual": "node test/visual/run-visual-tests.js",
+    "make-ts": "node types/make"
```
jamestalmage: This should probably be a `prepublish` script. Otherwise I am going to forget this step when publishing.

jamestalmage:
I think that script generation should happen in a `prepublish` script. Otherwise, I think we are good to go.

ivogabe:
@jamestalmage I would agree, but I'm afraid that the package cannot be published anymore with that change... The generation script uses ES2015, thus needs Node 6, and `npm publish` is broken on Node 6.

SamVerschueren:
@ivogabe I can't find features that can only be ran on Node 6 (spread operators for instance). Node 4 supports the majority of those features as well.

ivogabe:
That makes life easier then. I've added it in 54a8324f334ec0c74787e5078a0d0fea33cefffd.

jamestalmage:
You need to polyfill `.includes`. I see no reason not to just `require('babel-polyfill')` for this.

types/make.js
```diff
@@ -0,0 +1,125 @@
...
+// Writing these definitions by hand is hard. Because of chaining,
+// the number of combinations grows fast (2^n). To reduce this number,
+// illegal combinations are filtered out in `verify`.
+// The order of the options is not important. We could generate full
+// definitions for each possible order, but that would give a very big
+// output. Instead, we write an alias for different orders. For instance,
+// `after.cb` is fully written, and `cb.after` is emitted as an alias
+// using `typeof after.cb`.
+
+const path = require('path');
```
jamestalmage: `const => var` throughout. It's not supported on Node `0.12` or `0.10`.

Hopefully, that's the last of it. I would like to get this merged.

sindresorhus: Same with arrow functions, `let`, etc, used here. Would be much easier to just `require('babel/register')` this script.

sindresorhus:
@ivogabe ping :) => https://github.com/avajs/ava/pull/884#discussion_r67598376

ivogabe:
Sorry, I've been too busy lately and forgot about this. I've added `babel/register`, I thought it was easiest to do that in `package.json`, otherwise I would need to add another file that required babel and the generation script.

package.json
```diff
@@ -33,18 +33,21 @@
...
   "engines": {
     "node": ">=0.10.0"
   },
   "scripts": {
     "test": "xo && nyc --cache --reporter=lcov --reporter=text tap --no-cov --timeout=150 test/*.js test/reporters/*.js",
     "test-win": "tap --no-cov --reporter=classic --timeout=150 test/*.js test/reporters/*.js",
-    "visual": "node test/visual/run-visual-tests.js"
+    "visual": "node test/visual/run-visual-tests.js",
+    "prepublish": "npm run make-ts",
+    "make-ts": "node -e \"require('babel-register')({ presets: 'babel-preset-es2015' }); require('./types/make');\""
```
sindresorhus: Actually, we could just use `babel-node` here instead of `node`. We just need to add `babel-cli` as a devDependency. Then we could drop this boilerplate.

https://babeljs.io/docs/usage/cli/#babel-node

ivogabe:
Does `babel-node` require more configuration? It still fails on node 0.10 and 0.12. The tests are failing on node 4, has anyone an idea why?

sindresorhus:
@ivogabe Unfortunately yes, since Babel 6, everything needs config.

Replace it with:

```
"make-ts": "babel-node --presets=babel-preset-es2015 --plugins=transform-runtime types/make.js"
```

sindresorhus:
Awesome! Thanks for working on this @ivogabe :)

celebration

SamVerschueren:
Nice work @ivogabe !

ivogabe:
🎉

PR #886, opened 2016-05-28
Fixed typo: remove unnecessary `--` from  'Watch it' section

ismnoiet:
Remove unnecessary `--` from  **Watch it** section (readme file).

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -120,7 +120,7 @@ $ npm test
 ### Watch it
 
 ```console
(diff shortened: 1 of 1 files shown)


sindresorhus:
That's not a typo. You need it as `npm test` swallows flags.

http://unix.stackexchange.com/questions/11376/what-does-double-dash-mean-also-known-as-bare-double-dash

Issue #533, opened 2016-02-09
Investigate instantiating new API and reporter in --watch


novemberborn:
#502 resets the API and logger/reporter between runs. Creating fresh instances may be preferable. This would require some sort of code sharing between `cli.js` and `watcher.js` though, since the former sets up the instances and has access to the required CLI flags.

One option would be to pass a callback to `watcher.start()` allowing the watcher to create new API and logger/reporter instances, but leaving the definition in `cli.js`.

novemberborn:
@jamestalmage given your refactoring work what's your take on this?

jamestalmage:
I actually was thinking we could move watcher support into the API itself. That was kind of the thinking behind the `test-run` event. Also, in my refactor branch `lib/ava-files` could probably incorporate some of the file-path processing in `watcher.js` (I'm thinking methods like `makeTestMatcher` and `buildChokidarPatterns`, etc).

I'm thinking of adapting the run method a little:

```js
// new testRun is returned immediately, it may not start receiving events until the current run finishes.

var testRun = api.run(files, {
  runOnlyExclusive: boolean,
  stopCurrentRun: boolean
});

testRun.on('test', ...)

testRun.result.then(function (result) {

});
```

sindresorhus:
I like that idea. Ideally the CLI should just be a tiny wrapper around the API, with CLI only stuff.

jamestalmage:
> Ideally the CLI should just be a tiny wrapper around the API, with CLI only stuff.

Agreed. Our CLI is pretty huge. A lot of it is options processing (parsing the command line and merging with `package.json`). Part of the idea behind #700 is to reduce the amount of options processing API consumers would need to do.

novemberborn:
> I actually was thinking we could move watcher support into the API itself.

The watcher is fairly big though. I wonder if we could make the watcher instantiate its Api. We couldn't before due to how the loggers were coupled.

jamestalmage:
I didn't mean "move all of `watcher.js` into `api.js`". Just expose a way to pass options `{watch: true}` to the API and receive `test-run` events as files change.

novemberborn:
@jamestalmage that sound cool.

novemberborn:
The lack of this refactor hasn't been an issue. Closing this, but I expect we'll do something like this eventually as needs dictate.

PR #887, opened 2016-05-28
[WIP] Browser support

vdemedes:
Check out the commit messages, I tried to make them as descriptive as possible.

Implementation and usage details, vision: https://github.com/avajs/ava/issues/24#issuecomment-222367426.

**Coming soon.**


test/watcher.js
```diff
@@ -115,15 +115,20 @@ group('chokidar is installed', function (beforeEach, test, group) {
...
 		files = [
 			'test.js',
 			'test-*.js',
 			'test'
 		];
 
+		avaFiles = AvaFiles;
+
+		avaFiles.prototype.findTestFilesSync = function () {
+			return this.files;
```
vdemedes: Because API expects resolved paths, `findTestFilesSync` needs to be stubbed to return "input" files without changes, in order to easily check them in tests. Otherwise, `findTestFilesSync` would return empty array.

lib/ava-files.js
```diff
@@ -257,6 +266,67 @@ function handlePaths(files, excludePatterns, globOptions) {
 		});
 }
 
+function handlePathsSync(files, excludePatterns, globOptions) {
```
vdemedes: Async and sync versions of `handlePath` need to be split out into functions to DRY it up.

jamestalmage:
@vdemedes - mind sharing where you are going with this? There are lots of different ways to attack browser support, and we probably should all agree on the best way.

For instance, in my mind `api.js` was never going to run browser side. We would have done that with an alternate to `fork.js` that launched browsers and then received events via `socket.io`.

novemberborn:
I agree with @jamestalmage here. Let's continue the discussion in <https://github.com/avajs/ava/issues/24#issuecomment-222362675>?

vdemedes:
I pushed browser support work I've done, so that you can get a glance.

To test it out:

1. Generate tests.

```
$ ava --browser
```

2. Start http server to serve current directory (I use `serve`):

```
$ serve
```

3. Open `http://localhost:3000` in your browser.
4. Open console.

vdemedes:
Implementation and usage details: https://github.com/avajs/ava/issues/24#issuecomment-222367426.

test/api.js
```diff
@@ -565,66 +554,6 @@ function generateTests(prefix, apiCreator) {
...
-			runStatus.on('error', function (err) {
-				t.is(err.name, 'AvaError');
-				t.match(err.message, /Couldn't find any files to test/);
-			});
-		});
-
-		return api.run([path.join(__dirname, 'fixture/ignored-dirs/fixtures/test.js')]);
-	});
-
-	test(prefix + 'test file in helpers is ignored', function (t) {
```
jamestalmage: Lots of tests being deleted here. Many of which guard specific regressions. If the intent is to reimplement later, I would prefer skipping instead:

```js
test('tilte', { skip: true }, function (t) {
  ...
});

**Comment by jamestalmage** [comment]
Some thoughts after an initial review:

I don't think we should try to run the API in the browser. Instead, we should just be reimplementing `lib/send.js` to send messages via `socket.io` - the API would continue to run in Node.

There is also already a whole lot to review for a single PR. I would like to see an incremental approach.

I think the first step should be creating a build/bundle tool that packages a single test with browser safe versions of all the child process dependencies. I don't think `socket.io` interop would even need to be tackled in that first batch, we could just `console.log` messages sent to the parent (`profile.js` already contains polyfills for `process.send` that does just that).

File: test/api.js (hunk 3 above)
Comment (vdemedes) [inline test/api.js]: Those tests are not deleted, they moved to `ava-files.js` tests, which covers these cases.

**Comment by vdemedes** [comment]
> I don't think we should try to run the API in the browser.

Why? I had no issue with doing that.

File: test/api.js [hunk 4]
```diff
@@ -479,18 +480,6 @@ function generateTests(prefix, apiCreator) {
 			});
 	});
 
-	test(prefix + 'search directories recursively for files', function (t) {
```
jamestalmage: I do not see this test replicated anywhere.

test/api.js
```diff
@@ -565,66 +554,6 @@ function generateTests(prefix, apiCreator) {
 		return api.run([path.join(__dirname, 'fixture/immediate-3-exit.js')]);
 	});
 
-	test(prefix + 'testing nonexistent files causes an AvaError to be emitted', function (t) {
```
jamestalmage: I do not see this test replicated.

test/api.js
```diff
@@ -565,66 +554,6 @@ function generateTests(prefix, apiCreator) {
...
-			runStatus.on('error', function (err) {
-				t.is(err.name, 'AvaError');
-				t.match(err.message, /Couldn't find any files to test/);
-			});
-		});
-
-		return api.run([path.join(__dirname, 'fixture/broken.js')]);
-	});
-
-	test(prefix + 'test file in node_modules is ignored', function (t) {
```
jamestalmage: I see now you combined the following three tests into a single one. I would have preferred they stay separate, so each behavior is verified individually, but no biggie.

jamestalmage:
>> I don't think we should try to run the API in the browser.
> Why? 

We have efforts underway to bring watcher into the API, and simplify `cli.js` so it's a simple wrapper around it.

The less complexity we push into the browser, the better.

> I had no issue with doing that.

It's not so much a matter of whether or not it's possible, but rather if it is a good idea. It's definitely adding complexity.

I think it would be better if we started trying to make a browser fork that behaves like a `child_process` fork, replacing node's child process ipc, with `socket.io` stuff.

vdemedes:
> We have efforts underway to bring watcher into the API

Is there a discussion/issue/pr that I missed?

> It's not so much a matter of whether or not it's possible, but rather if it is a good idea. It's definitely adding complexity.

I don't see how it adds more complexity. I'm not modifying API in any way, it's good to work in the browser 1:1, unchanged. As I mentioned previously, I was trying to reuse as much as possible. I don't want to rewrite AVA core for browsers. Instead, the plan was to replace node-dependent parts with browser-compatible ones. What makes it a bad idea? Do you have specific arguments against doing this?

test/api.js
vdemedes: Each behavior remains verified, because there are all those should-be-ignored directories in the fixture, that `AvaFiles` tests use. If at least one directory was not ignored, test would fail.

jamestalmage:
> Is there a discussion/issue/pr that I missed?

https://github.com/avajs/ava/issues/533#issuecomment-205899266
https://github.com/avajs/ava/pull/759#issuecomment-220843047
https://github.com/avajs/ava/pull/865

> I don't want to rewrite AVA core for browsers

Not suggesting that. I'm suggesting you rewrite less than you have.

>  Instead, the plan was to replace node-dependent parts with browser-compatible ones

I do not disagree with this. I just think we could get there simpler if we sent stuff over the wire to the console instead of trying to do stuff in the browser.

niieani:
What's the status on this?

novemberborn:
Closing this since it's been languishing. Let's hash out our desired implementation in #24.

PR #888, opened 2016-05-30
add recipe for testing interactive console apps

arve0:
This is an example of how ava can be used to test interactive console applications. Helper functions has been put in the module [tty-test-helper](https://github.com/arve0/tty-test-helper), as suggested by @sindresorhus. tty-test-helper is currently in alpha stage, so things might change over there.

A repo with babel setup etc can be [found here](https://github.com/arve0/test_terminal_apps).

Any feedback is welcome.

### DIFF
--- a/docs/recipes/interactive-console-apps.md
+++ b/docs/recipes/interactive-console-apps.md
@@ -1,0 +1,103 @@
+# Testing interactive terminal applications
+
+## Using tty-test-helper
+
+Install the [tty-test-helper](https://github.com/arve0/tty-test-helper).
+```sh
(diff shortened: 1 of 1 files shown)


jamestalmage:
See: https://github.com/arve0/tty-test-helper/issues/1

Testing `tty` directly is fraught. I think most people would be better served if there was an easy way to test question flow in [`Inquirer.js`](https://github.com/SBoudrias/Inquirer.js/), or other similar library.

arve0:
Thanks for the feedback, closing.

PR #895, opened 2016-06-01
Add support for test macros in Typescript definitions

ChristianMurphy:
This brings support for test macros in the typescript definition.
It adds a macro interface and overloads the test function definition.

Based off https://github.com/avajs/ava#test-macros


jamestalmage:
See #884. I think that's probably the direction we are headed.

// @ivogabe

PR #896, opened 2016-06-02
simplify send.

jamestalmage:
Replaces `lib/send` with `send-to-parent`.

We already had special wrapping `send` when the communication direction was parent to child, so I just inlined a bit of logic in `fork.js` (basically a 1 liner). While there is a little bit of duplication there, doing so makes it clearer what's happening.

`send-to-parent` encapsulates "sending data to the parent", including a safety check to make sure the parent exists.

---
### original description:

Splits `lib/send` into two pieces:

- `wrap-send` wraps an IPC implementations. It reformats two arguments (eventName, data) into a single data object, and sends via the supplied send callback.

- `send-to-parent` wraps `process.send` using the above wrapper function. It also provides the "must be used via CLI" message.

`wrap-send` is only used by `fork.js` and the newly introduced `send-to-parent`.

`send-to-parent` can be used anywhere in the child process to send data to the parent (currently `index.js`, and `test-worker.js`).


vdemedes:
I'm struggling to see how this PR simplifies things. `send` function was already simple and straightforward. Now we need to use `send-to-parent` function when we need to send data to a parent process, and `wrap-send` + `function it returns` to send data **in a same way**, but to a child process?

This looks more like making things complex. What was the reason behind this PR, what issue does it fix?

lib/fork.js
```diff
@@ -86,7 +86,7 @@ module.exports = function (file, opts) {
 		ps.on('results', function (data) {
 			results = data;
 			data.tests = testResults;
-			send(ps, 'teardown');
+			send('teardown');
```
jamestalmage: simpler usage in `fork.js`

index.js
```diff
@@ -13,18 +11,6 @@ var runner = new Runner({
...
-var isForked = typeof process.send === 'function';
-
-if (!isForked) {
-	var fp = path.relative('.', process.argv[1]);
-
-	console.log();
-	console.error('Test files must be run with the AVA CLI:\n\n    ' + chalk.grey.dim('$') + ' ' + chalk.cyan('ava ' + fp) + '\n');
-
-	process.exit(1); // eslint-disable-line
-}
```
jamestalmage: removes this platform dependent check/error message into it's own file.

lib/test-worker.js
```diff
@@ -62,7 +62,7 @@ sourceMapSupport.install({
 
 var loudRejection = require('loud-rejection/api')(process); // eslint-disable-line
 var serializeError = require('./serialize-error');
-var send = require('./send');
```
jamestalmage: no additional complication in child process.

lib/fork.js
```diff
@@ -50,14 +50,14 @@ module.exports = function (file, opts) {
 	var relFile = path.relative('.', file);
 
 	var exiting = false;
-	var send = function (ps, name, data) {
```
jamestalmage: the only place you could make the argument things are "more complex" anywhere is this line, and I don't think it's that bad

lib/fork.js
vdemedes: I don't agree. Removal of one argument does not make things any simpler in this case. I'd prefer to leave it to make it more clear. It speaks for itself, "send ps a teardown event". 

While `send('teardown')` leaves me (at least) wondering where is this getting sent, what is the context.

If you are worried about consistency, we can compromise by using `send(process, 'some-event')` in the child processes to send event to a parent process.

index.js
vdemedes: We'd better move it to its own utility file. `wrap-send.js` does not sound like a good place for it either. I'm requiring a module to wrap a function, but under the hood it checks if the process was forked?

lib/test-worker.js
vdemedes: Could you elaborate? I wouldn't call a line, which consists of variable declaration and a require statement, a complication.

lib/fork.js
vdemedes: https://github.com/avajs/ava/pull/896#issuecomment-223197857

lib/test-worker.js
jamestalmage: Yes - that's my point, it's not adding any complexity here.

index.js
jamestalmage: it's not in `wrap-send`, it's in `send-to-parent` (it's own module - as requested), If there isn't a parent, it complains.

It seems entirely sensible that `require('send-to-parent')` would throw if you were no parent.

lib/fork.js
jamestalmage: Really, we could do away with `wrap-send` entirely, and just inline in `send-to-parent` and `fork.js`. It's hardly worth it at this point.

index.js
vdemedes: The argument stays the same (just another file name, I chose the wrong one). I meant own module for a check if the process is forked.

vdemedes:
@jamestalmage We got away from my original question with this discussion in the diff. Could you please address it and explain?

index.js
jamestalmage: if the module is called `send-to-parent`, I think it's reasonable to check for it there.

jamestalmage:
Look again. I think the wrapper was making it too complicated.

vdemedes:
Better, but I still don't see the value. Now we have to maintain the same message structure in **two** places to avoid bugs: https://github.com/avajs/ava/pull/896/files#diff-a48173105195b7a6481a3ad85b45144eR57 and https://github.com/avajs/ava/pull/896/files#diff-0e75bb55e5299c7ac7e253725935a6c4R18.

I think we should just resort to enforcing `send(where, what)` everywhere:

- in the fork.js: `send(ps, event)`
- in the test-worker.js: `send(process, event)`

novemberborn:
Looks good. The old send had a weird overload allowing you to pass in the `ps` object. Directly calling `ps.send()` in `fork.js` helps us get rid of that. Moving the fork guard into `send-to-parent` seems sensible.

We're already making minor assumptions around the message structure in various places. Doing it that way is easier than trying to hide it all behind an abstraction.

vdemedes:
I still don't see the value in these changes. I'll try to describe why step by step.

##### 1. One message structure, two functions.

Previously, we called `send('event name', { key: value })` and we knew the message is going to be in the "ava format". To change it, we edited `lib/send.js` and done. Now we have 2 functions in 2 files that do exact same thing and we need to edit 2 files to change the message structure (try not to forget about that one).

##### 2. One task, two functions.

This basically repeats the above point. For one task (sending data to another process) we now have 2 functions in 2 different places: `send` in `lib/fork.js` and `send` in `lib/send-to-parent`.

##### 3.
> The old send had a weird overload allowing you to pass in the ps object.

Why is it weird? It looks logical to me: `send(childPs, 'event name', data)`, which means `send this child process an event with that data`.

##### 4. IMO, fork guard does not belong in `lib/send-to-parent`.

To test `lib/send-to-parent` we have to set a fake function to `process.send` to pass the fork guard. I think the fork guard belongs to `lib/test-worker`, where it was residing previously.

---

To address the new PR description:

> We already had special wrapping send when the communication direction was parent to child, so I just inlined a bit of logic in fork.js (basically a 1 liner).

Could easily be:

```js
ps.send = function (event, data) {
  if (!exiting) {
    send(ps, event, data);
  }
};
```

which is just as clear (if not more) and avoids duplication.

@jamestalmage I'm arguing against PR's changes, but I've never heard an original intention or purpose of this PR. To bring up an unanswered older question: What was the reason behind this PR, what issue does it fix?

sindresorhus:
Closing in favor of https://github.com/avajs/ava/pull/942.

Issue #890, opened 2016-05-31
Document reporter behavior when run in CI environment.


danielhusar:
Hey guys,

I have set verbose to false which works great on my machine:

```
> NODE_ENV=test ava app/**/*.test.{js,jsx}
97 passed
```

However same tests on CI have always verbose logging enabled:
<img width="481" alt="screen shot 2016-05-30 at 21 46 03" src="https://cloud.githubusercontent.com/assets/1857751/15663436/f396a59e-26af-11e6-94b4-8db07ed18a7b.png">

Is this indented ?
Tests run on codeship.

BarryThePenguin:
I know it's intended, using `is-ci`, as shown in [cli.js L147-L149](https://github.com/avajs/ava/blob/ab3418fe62e1bcf2514b127ee340574027c507a3/cli.js#L147-L149)

#435 explains the history

danielhusar:
👍 would be probably nice to add it to documentation

jamestalmage:
Anyone want to do a documentation PR? It should just be adding a single short sentence. Just "the verbose reporter is always used in CI environments unless tap is specified"

sindresorhus:
Documented in https://github.com/avajs/ava/pull/955.

PR #898, opened 2016-06-02
Update docs for always verbose behavior

7373Lacym:
The docs now show that in CI environments the verbose reporter is used unless --tap is specified for #890

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -750,7 +750,6 @@ $ ava --tap | tap-nyan
 
 Please note that the TAP reporter is unavailable when using [watch mode](#watch-it).
 
-
 ### Clean stack traces
 
 AVA automatically removes unrelated lines in stack traces, allowing you to find the source of an error much faster.
--- a/readme.md
+++ b/readme.md
@@ -975,7 +975,7 @@ In contrast AVA is highly opinionated and runs tests concurrently, with a separa
 
 ### How can I use custom reporters?
 
-AVA supports the TAP format and thus is compatible with any [TAP reporter](https://github.com/sindresorhus/awesome-tap#reporters). Use the [`--tap` flag](#optional-tap-output) to enable TAP output. Please note, the verbose reporter is always used in CI environments unless [`--tap`](#optional-tap-output) is specified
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -975,7 +975,7 @@ In contrast AVA is highly opinionated and runs tests concurrently, with a separa
 
 ### How can I use custom reporters?
 
-AVA supports the TAP format and thus is compatible with any [TAP reporter](https://github.com/sindresorhus/awesome-tap#reporters). Use the [`--tap` flag](#optional-tap-output) to enable TAP output.
+AVA supports the TAP format and thus is compatible with any [TAP reporter](https://github.com/sindresorhus/awesome-tap#reporters). Use the [`--tap` flag](#optional-tap-output) to enable TAP output. Please note, the verbose reporter is always used in CI environments unless [`--tap`](#optional-tap-output) is specified
```
jamestalmage: `Please note, the verbose reporter` => `The verbose reporter`

7373Lacym:
Updated!  Thanks for the feedback!

jamestalmage:
LGTM!

sindresorhus:
Is this really the best place to put that information? People not wanting to use custom reporters might also want to know about the behavior. I would put it in the CLI section: https://github.com/avajs/ava#cli

jamestalmage:
I almost left this same comment, but couldn't come up with a better place to put it.

Thinking about it more, It definitely shouldn't go here - if users are interested in custom reporters, then they definitely don't care about this behavior at all.

I think we should create a whole new "reporters" section that encompasses this.

- It should show screenshots from both the verbose and mini reporter, and discuss how to set up TAP reporters in more detail.

- We could explain this behavior there.

- We could eliminate this FAQ

- We could move the [clean stack traces section](https://github.com/avajs/ava#clean-stack-traces) in there.

sindresorhus:
@jamestalmage Agreed

jamestalmage:
@7373Lacym Are you able to update this PR as suggested above? No worries if the answer is no, someone else can take what you've done so far and run with it.

7373Lacym:
Sorry Guys,  internship started so I cannot get to this at the moment.  If it's still around in a few weeks I will pick it back up.

MK-61:
Hello. Can I take this and continue, please?

sindresorhus:
Closing in favor of #955.

Issue #881, opened 2016-05-26
ES Bind Operator Throws SyntaxError When Test Failing


ghost:
## Description

I was trying to use the ES bind operator in my test files.
If I transpile the code in the Babel REPL and try it with the resulting ES5 code all is working fine so I think it must be something with AVAs or power-asserts internal transpilation.
After that I tried to make a reproduction and noticed that the SyntaxError is only there if the test fails.

### Test Source

```js
import test from 'ava'
import validate from './validate.js'

test('should be a function', t => {
    t.is(typeof validate, 'function')
})

test('should wrap a value in a validation object', t => {
    t.deepEqual(true::validate('foo'), {
        valid: true,
        value: true,
        name: 'bar'
    }, 'boolean')
})
```

`validate.js`:
```js
function validate(name) {
    return {
        name,
        valid: true,
        value: this
    }
}

export default validate
```

### Error Message & Stack Trace

```
> ava


 ⠏ required › should be a function

   1 passed
   1 failed/Users/konstantinpschera/Development/Projects/feed/validation/node_modules/acorn/dist/acorn.js:945
  throw err;
  ^

SyntaxError: Unexpected token (1:16)
    at Parser.pp.raise (/Users/konstantinpschera/Development/Projects/feed/validation/node_modules/acorn/dist/acorn.js:943:13)
    at Parser.pp.unexpected (/Users/konstantinpschera/Development/Projects/feed/validation/node_modules/acorn/dist/acorn.js:1503:8)
    at Parser.pp.expect (/Users/konstantinpschera/Development/Projects/feed/validation/node_modules/acorn/dist/acorn.js:1497:26)
    at Parser.pp.parseExprList (/Users/konstantinpschera/Development/Projects/feed/validation/node_modules/acorn/dist/acorn.js:643:12)
    at Parser.pp.parseSubscripts (/Users/konstantinpschera/Development/Projects/feed/validation/node_modules/acorn/dist/acorn.js:240:29)
npm ERR! Test failed.  See above for more details.
```

### Config

`package.json`:

```json
{
  "ava": {
    "require": [
      "babel-register"
    ],
    "failFast": true,
    "babel": "inherit"
  }
}
```

`.babelrc`:
```json
{
    "presets": ["es2015", "stage-0"]
}
```

### Command-Line Arguments

```
ava
```

## Relevant Links

- https://github.com/tim-mcgee/ava-es-bind

## Environment

```
[...]

PR #903, opened 2016-06-04
update to new power-assert-runtime

### LINKED ISSUE #693: Feature Request: Better error messaging for acorn parsing errors when testing React
If JSX code is inlined into a test assertion (e.g. using enzyme's .contains, etc) acorn from the power-assert library fails hard (fails hard enough to kill the `--watch` watcher):

```javascript
test("some test", t => {
    //...
    const wrapper = shallow(<Lesson {...props} />);
    t.true(wrapper.contains(<div className="page-container is-kiosk-mode" />));
});
```

```
/Users/dcousineau/Projects/frontend/node_modules/acorn/dist/acorn.js:945
  throw err;
  ^

SyntaxError: Unexpected token (1:24)
    at Parser.pp.raise (/Users/dcousineau/Projects/frontend/node_modules/acorn/dist/acorn.js:943:13)
    at Parser.pp.unexpected (/Users/dcousineau/Projects/frontend/node_modules/acorn/dist/acorn.js:1503:8)
...
```

The solution to fix this error is a very simple:

```javascript
test("some test", t => {
    //...
    const wrapper = shallow(<Lesson {...props} />);
    const actual = wrapper.contains(<div className="page-container is-kiosk-mode" />);
    t.true(actual);
});
```

It would be nice if AVA could expect and catch this error, prevent it from hard exiting the process, and possibly display messaging like "Unparseable (possibly JSX) code found in your test assertion. Try assigning the code to a variable first then running the assert against the variable".

nfcampos:
- should close #693, #881 and similar issues about experimental syntax breaking on assertion failure.
- it also appears to be faster than before, but not by much
- i've added two tests that assert that using JSX and `{...o}` on an assertion works

maybe @twada can check whether I correctly configured the various `power-assert` things


twada:
Thanks @nfcampos! This is exactly what I wanted to do today 😃 

[babel-plugin-espower 2.2.0](https://github.com/power-assert-js/babel-plugin-espower/pull/13) and the new [power-assert-runtime](https://github.com/twada/power-assert-runtime) comes with many architectural improvements with less dependencies. I hope AVA guys will like it.

jamestalmage:
I'm sure will!

@twada, has the runtime dependency graph been reduced? Should we move rendering back into the child processes - we've had lots of problems trying to serialize large objects for rendering in the main thread.

twada:
> has the runtime dependency graph been reduced?

Yeah, reducing dependencies is one of the goals of new power-assert-runtime. There's no extra parser anymore. Many ponyfills are consolidated into single core-js.

I also recommend @nfcampos to update [empower-core to 0.6.1](https://github.com/twada/empower-core/blob/master/CHANGELOG.md#061-2016-06-05) to reduce dependencies much more.


> Should we move rendering back into the child processes

Yes. Now we can render power-assert output in subprocesses.

jamestalmage:
@nfcampos. Can you bump the dependency as mentioned? Otherwise this looks good to me.

We can tackle moving rendering to the child process in another PR

nfcampos:
done

twada:
Thanks @nfcampos, looks good to me! 👍

jamestalmage:
Thanks @nfcampos! Want to work on moving the renderer into the child process?

The idea is to go back to the way `power-assert` behaves in other contexts - replacing the error message on the assertion.

nfcampos:
@jamestalmage yes, I'll do that, if I understand correctly it's something like using https://github.com/power-assert-js/empower instead of `empower-core` in `enhance-assert.js` and stop using the formatter in `run-status.js`?

twada:
@nfcampos One thing to notice is that `empower` module is now a thin wrapper around `empower-core` for power-assert internal compatibility and there is no reason to use `empower` in AVA. It's better to keep using `empower-core` directly.

nfcampos:
@twada yeah you're right

jimthedev:
@nfcampos So, quick question about the ava docs and this PR, specifically in [the React recipe for JSX helpers](https://github.com/avajs/ava/blob/master/docs/recipes/react.md#using-jsx-helpers) there the following note:

> Note that you have to use variables like actual and expected because power-assert doesn't handle JSX correctly.

The text links to [this](https://github.com/power-assert-js/power-assert/issues/34) issue which although open (documentation still pending) states that it will be available in ava as soon as this PR (#903) is merged. Since #903 has been merged, do we need to remove that note from the React recipe?

nfcampos:
yes that note should be removed, that should now work

PR #907, opened 2016-06-08
Fix TypeScript recipe.

jamestalmage:
Following the old example didn't work for me. It appears the new recommended way to install `tsc` is just to install `typescript`.


sindresorhus:
// @SamVerschueren

SamVerschueren:
Good catch James, you're right.

PR #911, opened 2016-06-10
move power assert error message formatting to child processes

nfcampos:
- removed power assert formatter usage from `run-status.js`
- onError function passed to `enhanceAssert` in `test.js` now replaces message on `Error` object if there is `powerAssertContext` and does not attach `powerAssertContext` to the error anymore.
- removed assertion from test in `test/api.js` that assumed errors had `powerAssertContext`.


jamestalmage:
Thanks!

twada:
@nfcampos Great! 🎉

Issue #917, opened 2016-06-13
`after.always` doesn't run if there are uncaught exceptions


arve0:
Hi! This is the second time I read the documentation at README.md and find that the feature I'm using is not yet released (after some debugging). I'm not sure what it was last time, but this time it is `after.always` found in [commits since release](https://github.com/avajs/ava/compare/v0.15.2...master).

Any thoughts about stabilizing/syncing the docs with the release cycle?

sindresorhus:
`after.always` was in the 0.15.0 release. If you're referring to https://github.com/avajs/ava/commit/2e984380c4df2bfe5a0af2ec37a970f4001c9f53, it's just updating the TypeScript definition.

> Any thoughts about stabilizing/syncing the docs with the release cycle?

New features always come with docs. That's a requirement.

arve0:
Hm. I'll recreate my issue, give me some seconds.

arve0:
OK, this is actually a bug. This code will reproduce:

```js
var test = require('ava');
var r = require('request-promise');

test.before(() => console.log('before'));

test('failing test', (t) =>
  r('http://httpbin.org/status/400')
    .catch(err => t.true(err.statusCode === 200)));

test('another failing test', (t) =>
  r({
    uri: 'http://httpbin.org/status/400',
    method: 'POST',
    body: { should: 'be string' } // this will throw
  }).catch(err => t.true(err.statusCode === 200)));

test.after.always(() => console.log('after'));
```

Seems to be a race condition. `after` runs if running test serially or if one of the tests is omitted.

https://asciinema.org/a/8z74ozt2e1tahjlqlx3mnj774

arve0:
Throws from a place ava is unable to catch? Might not really an issue, as this is misuse of `request-promise`, as `body` should be a string or `json` should be set to true.

arve0:
Simpler test case:

```js
var test = require('ava');

test.before(() => console.log('before'));

test.cb('another failing test', (t) => {
  setTimeout(() => {
    t.true(true);
    throw new Error('catch me');
  }, 1);
});

test.after.always(() => console.log('after'));
```

Is this expected?

jamestalmage:
We shut the process down immediately if an uncaught exception happens, so `after.always` isn't guaranteed to fire in that case.

I guess we could trigger `after.always` in the case of uncaught exceptions.

arve0:
Nice when using `--watch`, cleaning state in `after` and unintentionally writing something wrong in your test.

[...]

PR #918, opened 2016-06-14
readme: add note about uncaught exceptions

arve0:
fixes issue #917

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -435,7 +435,7 @@ test.failing('demonstrate some bug', t => {
 
 AVA lets you register hooks that are run before and after your tests. This allows you to run setup and/or teardown code.
 
-`test.before()` registers a hook to be run before the first test in your test file. Similarly `test.after()` registers a hook to be run after the last test. Use `test.after.always()` to register a hook that will **always** run once your tests and other hooks complete. `.always()` hooks run regardless of whether there were earlier failures, so they are ideal for cleanup tasks.
+`test.before()` registers a hook to be run before the first test in your test file. Similarly `test.after()` registers a hook to be run after the last test. Use `test.after.always()` to register a hook that will **always** run once your tests and other hooks complete. `.always()` hooks run regardless of whether there were earlier failures, so they are ideal for cleanup tasks. There are two exceptions to this however. If you use `--fail-fast` AVA will stop testing as soon as a failure occurs, and it won't run any hooks including the `.always()` hooks. Uncaught exceptions will crash your tests, possibly preventing `.always()` hooks from running.
 
 `test.beforeEach()` registers a hook to be run before each test in your test file. Similarly `test.afterEach()` a hook to be run after each test. Use `test.afterEach.always()` to register an after hook that is called even if other test hooks, or the test itself, fail. `.always()` hooks are ideal for cleanup tasks.
 


jamestalmage:
I'm not sure this is the correct behavior. We may just want to run `after.always` on uncaught exceptions

novemberborn:
> I'm not sure this is the correct behavior. We may just want to run after.always on uncaught exceptions

Uncaught exceptions cause the worker to be torn down. That seems reasonable to me. The behavior is somewhat similar to `--bail`, which we've previously decided does not cause `after.always` to run.

readme.md
```diff
@@ -433,7 +433,7 @@ test.failing('demonstrate some bug', t => {
 
 AVA lets you register hooks that are run before and after your tests. This allows you to run setup and/or teardown code.
 
-`test.before()` registers a hook to be run before the first test in your test file. Similarly `test.after()` registers a hook to be run after the last test. Use `test.after.always()` to register a hook that will **always** run once your tests and other hooks complete. `.always()` hooks run regardless of whether there were earlier failures, so they are ideal for cleanup tasks.
+`test.before()` registers a hook to be run before the first test in your test file. Similarly `test.after()` registers a hook to be run after the last test. Use `test.after.always()` to register a hook that will **always** run once your tests and other hooks complete. `.always()` hooks run regardless of whether there were earlier failures, so they are ideal for cleanup tasks. **Note**: In the case of uncaught exceptions, `test.after.always()` will not be called.
```
novemberborn: I would rephrase this though:

> … they are ideal for cleanup tasks. There are two exceptions to this however. If you use `--fail-fast` AVA will stop testing as soon as a failure occurs, and it won't run any hooks including the `.always()` hooks. Uncaught exceptions will crash your tests, possibly preventing `.always()` hooks from running.

arve0: Uncaught exceptions will crash your tests, ~~possibly~~ preventing .always() hooks from running?

arve0: I'm not familiar with the internals, but are there cases where uncaught exceptions will not prevent `.after` to run?

novemberborn: Yes, if they occur after `.after` has run. With asynchronous tests it's a race between a test completing and an uncaught exception occurring.

arve0:
Updated to new wording.

novemberborn:
LGTM 😉

sindresorhus:
LGTM, because it reflects reality.

Though, I'm not totally happy about `.always` not actually being always, but I think the earlier arguments were that you'd want the tests and fixtures to be in a consistent state so you can debug. Personally, I find it a bit annoying having to manually cleanup the leftovers. There's also the argument that cleanup should happen on startup, but then you get a dirty git repo. Feels like there's something we could do to ease this for the user.

jamestalmage:
> There's also the argument that cleanup should happen on startup, but then you get a dirty git repo.

I usually get around that with `.gitignore`.

>  Feels like there's something we could do to ease this for the user.

What if we allowed a `.cleanup` modifier that ran before and after the tests?

Or what if we allowed an `and` modifier:

```js
test.before.and.after(fn)
test.beforeEach.and.after.always(fn)
```

sindresorhus:
@jamestalmage Can you move the bottom part to a new issue?

PR #922, opened 2016-06-17
Update version in readme.md

rohitpaulk:
I installed this as per the instructions, and was confused when half of the documented API methods weren't working. Checked the releases page and realized that I was on an older version :)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -73,7 +73,7 @@ $ ava --init
     "test": "ava"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


sindresorhus:
It's just an example. You shouldn't copy-paste it...

Read the instruction:

> Install AVA globally run with `--init` to add AVA to your `package.json`

`--init` adds the latest version for you to package.json.

PR #924, opened 2016-06-18
Extract babel config generation into it's own submodule

jamestalmage:
This adds `lib/babel-config.js`, it is used to:

1. Validate the `babel` option in `package.json` (i.e. make sure it's either `"default"`, `"inherit"`, or an `Object`). We were doing this in `cli.js`, which was definitely not the right place to do it.

2. Build a babel config which is passed to `babel.transform`. This was previously done in `caching-precompiler`. That didn't make much sense either, `caching-precompiler` should be dedicated to handling the caching, not options manipulation.

### DIFF
--- a/lib/babel-config.js
+++ b/lib/babel-config.js
@@ -85,7 +85,7 @@ var defaultPlugins = lazy(function () {
 });
 
 function build(babelConfig, filePath, code) {
-	validate(babelConfig);
+	babelConfig = validate(babelConfig);
 
 	var options;
 
--- a/lib/babel-config.js
+++ b/lib/babel-config.js
@@ -12,7 +12,7 @@ function validate(conf) {
 	}
 
 	// check for valid babel config shortcuts (can be either "default" or "inherit")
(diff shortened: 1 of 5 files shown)


lib/babel-config.js
```diff
@@ -0,0 +1,146 @@
...
+var objectAssign = require('object-assign');
+var colors = require('./colors');
+
+function validate(conf) {
+	if (conf === undefined || conf === null) {
+		conf = 'default';
+	}
+
+	// check for valid babel config shortcuts (can be either "default" or "inherit")
+	var isValidShortcut = ['default', 'inherit'].indexOf(conf) !== -1;
```
novemberborn: This always seemed weird to me… but now that the variable is shorter why not `conf === 'default' || conf ===' inherit'`?

novemberborn:
Typo in the `test/bable-config.js` filename.

LGTMO!

sindresorhus:
LGTM

PR #925, opened 2016-06-19
extract ava-files to it's own module

jamestalmage:
This will allow other tools (i.e. atom-ava, linter, etc) to have access to the file selection logic.

The external module includes the updates from #863, so I have closed that.


novemberborn:
Is the watcher still broken with this? (Didn't check why CI failed.)

Do we want to remove the `--sources` flag as part of this? (https://github.com/avajs/ava/pull/863#issuecomment-221465954)

api.js
```diff
@@ -23,7 +23,9 @@ function Api(options) {
 
 	EventEmitter.call(this);
 
-	this.options = options || {};
+	this.options = objectAssign({}, options);
+	this.options.cwd = this.options.cwd || process.cwd();
+	this.options.resolveTestsFrom = this.options.resolveTestsFrom || this.options.cwd;
 	this.options.match = this.options.match || [];
```
sindresorhus: The above 3 options can just be put as defaults in the objectAssign call.

jamestalmage:
> Is the watcher still broken with this? (Didn't check why CI failed.)

The CI failure was unrelated (npm install checksum errors), I restarted and it passed. Neither this or #863 "broke" watcher. But it doesn't improve it either. With this you can do `ava some-file.js` from a subdir and it works. With watcher, you will still need to run it from the root directory of the package for things to work. 

It is a little more complicated to make watcher work in a subdir. As agreed, the `sources` config in `package.json` should be evaluated from the basedir regardless. If you `ava -w some-file.js` from the `test` subdir, you still want to watch for sources from the basedir, but the pattern you passed on the command line needs to be converted to be relative to the subdir.  If we **don't** eliminate the `--sources` flag, that makes watching from a subdir even more complicated (I think).

novemberborn:
I've opened #936 and #937 for follow-up discussion.

PR #926, opened 2016-06-19
extract ava-assert to it's own module

jamestalmage:
This extracts our core assertions to it's own module.

While doing so I found a few bugs in our tests:

- We were testing circular references using a class that we created with an arrow function. Arrow functions have their `this` reference bound to the surrounding block, so we actually weren't testing circular references at all.
- We were missing a few patterns in `enhanceAssert`. The new module includes a test that ensures our exposed API and our patterns match.


sindresorhus:
Why? What benefit does it give us from being modularized? Is it going to be used elsewhere? It actually used to be its own module, but I moved it into the AVA repo as it's highly specific to our needs and having it in the same repo makes it easier to maintain. There are more flexible assert libs out there for anyone wanting this.

jamestalmage:
I was thinking it would allow a means for users of other test runners to gradually transition to AVA asserts before switching to full blown AVA.

sindresorhus:
I don't really see why anyone would do that. They could just switch fully with the `--serial` flag initially. It would also be incompatible with our forthcoming automatic migration scripts. Modularization is amazing, but there should be some clear benefits for doing so. Now it seems it's just introducing extra coordinating for any assert change. I guess we can land this since it's already done, but would have been nice to have had a discussion about this beforehand.

novemberborn:
Wait @sindresorhus is arguing *against* small modules?!

> Now it seems it's just introducing extra coordinating for any assert change.

Perhaps assertions are a little too tightly coupled to AVA itself for them to be in their own module.

sindresorhus:
@novemberborn Lol. I'm all for small modules when it makes sense and it can be reused for other things. Unclear in this case whether it's worth the overhead.

jamestalmage:
We could always do a monorepo!! 😉

sindresorhus:
disgusted

jamestalmage:
I don't really think we are going to add that much overhead with this. Changing our assertions API should be a big deal anyways.

This could also open the door for pluggable/swappable assertions, which is something I was thinking about working on. We could wait to land this until it's proven necessary (though that would mean `ava-assert` might drift out of date - and might need to be redone completely if it takes to long for me to get around to pluggable assertions).

> I guess we can land this since it's already done

That is definitely not a good enough reason to land it. Let's not get in the habit of doing things for that reason.

sindresorhus:
> This could also open the door for pluggable/swappable assertions

I don't really see how this PR makes that easier. Can you elaborate? We could easily do swappable asserts with the assert.js being in a file in this repo. Doesn't really make a difference for that.

> That is definitely not a good enough reason to land it.

Read my statement in context. I meant that since you all seem to feel differently and it's already done, I could cave.

jamestalmage:
It don't know that it would necessarily make it easier. But it would force us to write our our asserts as a proper plugin, dogfooding our own plugin API, and it could serve as a model for how to write your own assertion plugin.

jamestalmage:
> Read my statement in context

Yeah, I just wanted to be clear that my feelings wouldn't be hurt over the "wasted" effort. (It wouldn't really be wasted - I found some bugs regardless).

jamestalmage:
Oh - I also discovered we weren't directly testing assertions that act on Observables. We were doing it in `test/observable.js`, but that was rather indirect via `lib/test`. (just trying to make sure I capture all the improvements I made while I still remember, if we decide to abandon this and backport)

novemberborn:
> Read my statement in context. I meant that since you all seem to feel differently and it's already done, I could cave.

I hadn't thought about it to the same extend. I'm in favor of leaving this in AVA now.

sindresorhus:
> But it would force us to write our our asserts as a proper plugin, dogfooding our own plugin API

We could still do that when it's here though. I think that's a premature concern as we're not currently working on a plugin API. We have no idea how such plugin API would look like, not even if a custom assertion lib would replace the built-in or be an addition.

-

Let's keep it in AVA core for now. We can reconsider when doing the plugin API if we see enough benefit of doing so.

sindresorhus:
Opened an issue for fixes to bring back into AVA core: https://github.com/avajs/ava/issues/944

PR #927, opened 2016-06-19
Fix babel settings for jsx in test files

jairtrejo:
# Challenge
When using JSX in test code, you need to enable the react preset, even if you already do so as part of your general babel config.

The current docs use JSX in the test, but have no babel presets in the `ava` config.

# Solution
Setting the `babel` option to `"inherit"` (to pick up the babel config set in the same file) does the trick.


jamestalmage:
Good catch! Thanks!

Issue #131, opened 2015-11-05
JSPM support?


dananichev:
Is there any JSPM support built-in? 

Thanks!

sindresorhus:
I don't understand the question. JSPM support in what form? Can you elaborate?

dananichev:
@sindresorhus at least as a module loader. I mean, AVA is obviously supports ES2015 syntax, but JSPM includes SystemJS loader. Will AVA resolve modules correctly?

sindresorhus:
We just run the test file through Babel.js. Whatever Babel supports, we support.

This might help: https://babeljs.io/docs/setup/#jspm

dananichev:
@sindresorhus 
I suppose it's not that easy.
The problem is: you need to load somehow SystemJS config (generated by JSPM) before running the tests.

Example of the project
https://github.com/dananichev/ava-jspm-example

```
npm test
```
results in
```
/src/helpers.js:1
(function (exports, require, module, __filename, __dirname) { import * as utils from 'underscore.string';
                                                              ^^^^^^

SyntaxError: Unexpected token import
```

Am i missing something?

eirikb:
Hi @dananichev 

Your `import..`issue can be solved this way:

1. Update ava to latest version.
2. Add `"ava": { "require": [ "babel-core/register" ]` to your _package.json_-file.
3. Replace your _.babel.rc_-file content with `{"presets": ["es2015"]}`.

You will still have problems with modules not loading from _jspm_packages_.
I don't know how to solve this properly, one super-ugly-don't-do-this-hack is to override `require` to check _jspm_packages_-folder first (since the layout is a bit different other resolvers might not handle it).
Here is a working example (with the horrible hack): https://gist.github.com/eirikb/d8d9a20673166c02bf62

Note that jspm is not actually used when testing.

dananichev:
@eirikb Thanks. But i think i will stick with Jasmine/Mocha/etc for now. Anyhow AVA seems pretty intriguing and i'm planning to wait for a full JSPM support and so on.

salfield:
This code below works fine for me. Slightly annoying boilerplate to add to each test file, but works.

```
import test from 'ava';
import jspm from 'jspm';
jspm.setPackagePath('../../../');
let moduleName;

test.before(t => {
  return jspm.import('path/to/module')
    .then((module) => {
      moduleName = module;
      return
    })
});

test(t => {
  console.log(moduleName);
  /* run a test here */
});
```

[...]

PR #941, opened 2016-06-27
JSPM and SystemJS recipe.

skorlir:
A recipe for using AVA with JSPM, per discussion on #131. It requires the import of an external library I wrote for the purpose of encapsulating the loader shim so it can be maintained and updated. Here's a link to that library: [ava-jspm-loader](https://github.com/skorlir/ava-jspm-loader).

### DIFF
--- a/docs/recipes/jspm-systemjs.md
+++ b/docs/recipes/jspm-systemjs.md
@@ -1,0 +1,68 @@
+# JSPM + SystemJS for ES6
+
+It requires a special loader helper to correctly resolve `import`s of JSPM
+Packages when using AVA.
+
+## Setup
+
+This recipe has only been tested with JSPM v0.17.0-beta.22, but it should work
+with any version of JSPM v0.17 and may work with v0.16.
+
+### Babel
+
+Configure your .babelrc to work with AVA if you have not already. NOTE: You can keep
+additional configuration in your JSPM config files to override these settings during
+bundling and building.
+
+```json
+{
+  "presets": [ "es2015", "stage-2" ]
+}
+```
+
+You can find more information about setting up Babel with AVA in the [babelrc recipe](babelrc.md).
+
+### JSPM Loader Helper
+
+You will need to install the AVA JSPM Loader as a dev dependency.
+
+`npm install --save-dev ava-jspm-loader`
+
+You will also need to update your AVA config in package.json to use the jspm
+loader.
+
+```json
+{
+  "ava": {
+    "require": [
+      "babel-register",
+      "ava-jspm-loader"
+    ]
+  }
+}
+```
+
+NOTE: if you use async/await in your source code (not in your test code), you
+will need to install "babel-polyfill" from NPM and add it to your require array.
+
+### Example test file
+
+Note that you will need to use `System.import` paths for all of your project
+files. So, if you named your project `app` and you want to import your `main.js`
+into a test file, you will need to `import main from 'app/main'`.
+
+```js
+import test from 'ava'
+
+import main      from 'app/main'     // maps to your JSPM config for "app/main.js"
+import BigNumber from 'bignumber.js' // in jspm_packages
+
+function fn () {
+  return Promise.resolve(new BigNumber("1234567890.123456789"))
+}
+
+test('example test', async t => {
+  t.is((await fn()).toString(), "1234567890.123456789")
+})
+```
+
--- a/docs/recipes/jspm-systemjs.md
+++ b/docs/recipes/jspm-systemjs.md
@@ -1,7 +1,7 @@
-# JSPM + SystemJS for ES6
+# JSPM + SystemJS for ES2015
 
 It requires a special loader helper to correctly resolve `import`s of JSPM
-Packages when using AVA.
+packages when using AVA.
 
 ## Setup
 
@@ -10,13 +10,13 @@ with any version of JSPM v0.17 and may work with v0.16.
 
 ### Babel
 
-Configure your .babelrc to work with AVA if you have not already. NOTE: You can keep
-additional configuration in your JSPM config files to override these settings during
-bundling and building.
+Configure your .babelrc to work with AVA if you have not already. NOTE: You can
+keep additional configuration in your JSPM config files to override these
+settings during bundling and building.
 
 ```json
 {
-  "presets": [ "es2015", "stage-2" ]
+  "presets": ["es2015", "stage-2"]
 }
 ```
 
@@ -24,11 +24,12 @@ You can find more information about setting up Babel with AVA in the [babelrc re
 
 ### JSPM Loader Helper
 
-You will need to install the AVA JSPM Loader as a dev dependency.
+You will need to install the [AVA JSPM loader](https://github.com/skorlir/ava-jspm-loader)
+as a dev dependency.
 
 `npm install --save-dev ava-jspm-loader`
 
-You will also need to update your AVA config in package.json to use the jspm
+You will also need to update your AVA config in package.json to use the JSPM
 loader.
 
 ```json
@@ -42,8 +43,8 @@ loader.
 }
 ```
 
-NOTE: if you use async/await in your source code (not in your test code), you
-will need to install "babel-polyfill" from NPM and add it to your require array.
+NOTE: If you use async/await in your source code (not in your test code), you
+will need to install "babel-polyfill" from npm and add it to your require array.
 
 ### Example test file
 
@@ -52,17 +53,15 @@ files. So, if you named your project `app` and you want to import your `main.js`
 into a test file, you will need to `import main from 'app/main'`.
 
 ```js
-import test from 'ava'
+import test from 'ava';
+import main from 'app/main';     // maps to your JSPM config for "app/main.js"
+import BigNumber from 'bignumber.js'; // in jspm_packages
 
-import main      from 'app/main'     // maps to your JSPM config for "app/main.js"
-import BigNumber from 'bignumber.js' // in jspm_packages
-
-function fn () {
-  return Promise.resolve(new BigNumber("1234567890.123456789"))
+function fn() {
+	return Promise.resolve(new BigNumber('1234567890.123456789'));
 }
 
 test('example test', async t => {
-  t.is((await fn()).toString(), "1234567890.123456789")
-})
+	t.is((await fn()).toString(), '1234567890.123456789');
+});
 ```
-
--- a/readme.md
+++ b/readme.md
@@ -1004,6 +1004,7 @@ It's the [Andromeda galaxy](https://simple.wikipedia.org/wiki/Andromeda_galaxy).
 - [TypeScript](docs/recipes/typescript.md)
 - [Configuring Babel](docs/recipes/babelrc.md)
 - [Testing React components](docs/recipes/react.md)
+- [JSPM and SystemJS](docs/recipes/jspm-systemjs.md)
 
 ## Support
 


dak:
This appears to work for normalizing mappings, but not for paths, although the `ava-jspm-loader` also appears to only focus on paths.

This may be because I'm only using JSPM to bundle dependencies and not the entire app.

I wonder if we may need a better synchronous API from JSPM to properly let it handle loading modules using its configs.

skorlir:
@dak 
When you say, "not for paths", what do you mean?

Currently, if you have a System config mapping like:

```
System.config({
  browserConfig: {
    "paths": {
      "app/": "/src/"
    }
  }
})
```

And then you try to import `app/main.js`, this will resolve correctly to `/src/main.js`. That's typical SystemJS behaviour. If you aren't mapping your source files in your SystemJS config, it won't be able to find your source files that way. Of course, you could probably still do something like `import main from '../src/main'`, but that's less ideal.

Why are you only bundling part of your project using SystemJS?

Can you give a concrete example of a configuration that does not work correctly?

dak:
I'm only bundling part of it because I'm hosting over HTTP/2 and the app is quite large, so the benefit of reducing the number of requests is near zero due to multiplexing (just the extra headers) compared with being able to return just the relevant templates/scripts for a given page.
It's a supported use case by systemjs builder: https://github.com/systemjs/builder#example---third-party-dependency-bundles

I'm also using JSPM 0.17, which may? be part of the discrepancy, although I suspect it's more than that.

Stripped down, my config is basically the same as above:
```js
paths: {
    "~/": "src/app/"
}
```
but I'll get errors like: `Error: Cannot find module '~/path/to/module'`, which suggest the path normalization never occurred.

Let me try to make a simple repo to demo the issue.

skorlir:
[This part](https://github.com/skorlir/ava-jspm-loader/blob/master/jspm-loader.js#L16) of ava-jspm-loader should account for that path mapping, so it strikes me as odd that this isn't working. Let me know if you get a demo repo up!

dak:
@skorlir See: https://github.com/dak/jspm-ava-test

skorlir:
@dak I made a pull request into your example project that fixes the testing configuration.

Your project was just misconfigured. I've spent a bunch of time on JSPM and AVA configuration already this week, so I feel your pain. You have to get it just right.

skorlir:
Actually, there is a bug there - you pointed it out but I didn't see it.

The jspm-ava-loader doesn't handle nested paths. Just a sec and I'll push a fix,

skorlir:
It's fixed now and working correctly. The recipe is unchanged, but ava-jspm-loader has been patched to handle directory nesting.

dak:
@skorlir I think that did the trick. :+1:

MeoMix:
@skorlir When I try using your repo I encounter an error:

> Error: Cannot find module '/C:/Users/Meo/GitHub/jspm-ava-test/jspm_packages/github/jspm/nodelibs-fs@0.2.0-alpha'

None of the node-libs seem to be loading properly.

If I work around it by not applying `System.normalizeSync` to nodelibs then I encounter another error:

> Error: Cannot find module '/C:/Users/Meo/GitHub/jspm-ava-test/src/app/test/test.js'

Suspect you guys aren't testing this against Windows paths?

dak:
@skorlir @MeoMix I just noticed something similar actually... you need to change the order and prefer NPM modules over JSPM modules. It broke when I tried combining it with nyc for the same reason @MeoMix identified.

Reversing the loading order (with a try/catch) resolved the issue.

MeoMix:
@dak Yeah, that appears to work, too -- although using try/catch as part of normal control flow is generally frowned upon.

Additionally, there are still issues with Windows paths. I need to have the following:

`jspmUri = jspmUri.replace('///', '//');` for it to work properly as Windows looks like `file:///C:/...` which results in the prefix'ed slash which causes lookup failure.

dak:
Rather than trying to duplicate how JSPM is loading modules, it would be nice if JSPM exposed one or two methods to handle this. /me nudges @guybedford
I'm sure we're missing other edge cases still.

MeoMix:
Well, I'm pretty sure that's what JSPM does internally. It's a bit of an odd scenario to interface into JSPM, but then go back to loading through NPM after using it.

I have to wonder if we can just do something similar to how I've handled it with the JSPM css loader: https://github.com/MeoMix/jspm-loader-css/blob/master/src/abstractLoader.js#L58

but I'm not sure atm if they just seem similar or if it's actually the same problem being solved

skorlir:
The Node errors are interesting. I don't get those... I will have to look into it further. You are probably right that we should just try to load from node_modules first, but I don't like it. I also suspect this will break module loading error messages if we are not careful. 

I can fix the Windows paths, but I am confused why a Windows path would even matter? If it isn't a path in System.map (which those paths are platform agnostic) the function in question should return false. I think any absolute path, Windows or otherwise, will return false - which was the behavior I expected. It could very well be the case that it is possible to have a leading slash in a map path, but that strikes me as unlikely since everything is baseUrl relative.

I agree however: that function is an area for improvement. The nested paths patch made it messy and it is likely there are more edge cases there. 

@dak I think the intent is that you'll just use jspm.import (which itself is just an interface into SystemJS). Unfortunately there's no synchronous import - that's what I would have wanted to use really. I tried to think of a way to "synchify" it but it made my head hurt. 

@meomix I can't use the exact fix you are using for the css loader partly because I can't do any async work in the Module._load shim. I am not sure I should manually prepend baseUrl either - I'd rather get normalization working. Was there something else there you wanted me to look at?

skorlir:
@dak can you give a sample setup where `nyc` or some other node module breaks?

dak:
@skorlir See: https://github.com/dak/jspm-ava-test again
You'll need nyc installed globally, I didn't add it to the package (or npm install it manually). The run `npm test`.

dak:
@skorlir So I think I've found an even better solution by making use of Ava's support for async functions.  We kept trying to get around how JSPM loads modules asynchronously instead of just embracing it.

The `ava-jspm-loader` would simply be:
```js
const path = require('path');
const jspm = require('jspm');

const pjsonLoc = require('find-pkg').sync(process.cwd());
const pjsonDir = path.parse(pjsonLoc).dir;

jspm.setPackagePath(pjsonDir);

global.System = jspm.Loader();
```

No funky hacks.

Then, you'd write tests using async functions:

import test from 'ava';

test('Test name', async function (assert) {
    const module = await System.import('jspm/import/path');

    let actual = module.returnTrue();
    let expected = true;

    assert.is(actual, expected, 'returnTrue() should return true.');
});
```

**Comment by skorlir** [comment]
Interesting - you must be bundling your code before testing if that works
for you?

**Comment by dak** [comment]
@skorlir It does appear to need to run through Babel first... would it work with babel-register if you included the systemjs plugin for babel too?

**Comment by skorlir** [comment]
Can you test modules in your source code separately this way? Can you test
"internal" modules that aren't exposed as part of the API?

If you have to bundle first, I'm guessing not.

If so, though, this would be worth looking into further.

**Comment by dak** [comment]
Yes, I build them but don't bundle them, so they're still individual files.
See: https://github.com/dak/jspm-ava-test
After you run `npm test` take a look at the `dev` folder it creates.

File: docs/recipes/jspm-systemjs.md [hunk 1]
```diff
@@ -0,0 +1,68 @@
+# JSPM + SystemJS for ES6
```
sindresorhus: ES6 => ES2015

docs/recipes/jspm-systemjs.md
```diff
@@ -0,0 +1,68 @@
...
+
+### Babel
+
+Configure your .babelrc to work with AVA if you have not already. NOTE: You can keep
+additional configuration in your JSPM config files to override these settings during
+bundling and building.
+
+```json
+{
+  "presets": [ "es2015", "stage-2" ]
```
sindresorhus:
```json
"presets": ["es2015", "stage-2"]
```

docs/recipes/jspm-systemjs.md
```diff
@@ -0,0 +1,68 @@
...
+
+You can find more information about setting up Babel with AVA in the [babelrc recipe](babelrc.md).
+
+### JSPM Loader Helper
+
+You will need to install the AVA JSPM Loader as a dev dependency.
+
+`npm install --save-dev ava-jspm-loader`
+
+You will also need to update your AVA config in package.json to use the jspm
```
sindresorhus: jspm => JSPM

docs/recipes/jspm-systemjs.md
```diff
@@ -0,0 +1,68 @@
...
+{
+  "presets": [ "es2015", "stage-2" ]
+}
+```
+
+You can find more information about setting up Babel with AVA in the [babelrc recipe](babelrc.md).
+
+### JSPM Loader Helper
+
+You will need to install the AVA JSPM Loader as a dev dependency.
```
sindresorhus: Loader => loader

docs/recipes/jspm-systemjs.md
```diff
@@ -0,0 +1,68 @@
...
+    "require": [
+      "babel-register",
+      "ava-jspm-loader"
+    ]
+  }
+}
+```
+
+NOTE: if you use async/await in your source code (not in your test code), you
+will need to install "babel-polyfill" from NPM and add it to your require array.
```
sindresorhus: NPM => npm

docs/recipes/jspm-systemjs.md
```diff
@@ -0,0 +1,68 @@
...
+  "ava": {
+    "require": [
+      "babel-register",
+      "ava-jspm-loader"
+    ]
+  }
+}
+```
+
+NOTE: if you use async/await in your source code (not in your test code), you
```
sindresorhus: if => If

docs/recipes/jspm-systemjs.md
```diff
@@ -0,0 +1,68 @@
...
+
+### Example test file
+
+Note that you will need to use `System.import` paths for all of your project
+files. So, if you named your project `app` and you want to import your `main.js`
+into a test file, you will need to `import main from 'app/main'`.
+
+```js
+import test from 'ava'
+
```
sindresorhus: No need for empty line here.

docs/recipes/jspm-systemjs.md
```diff
@@ -0,0 +1,68 @@
...
+import main      from 'app/main'     // maps to your JSPM config for "app/main.js"
```
sindresorhus: Don't align the `from`.

docs/recipes/jspm-systemjs.md
```diff
@@ -0,0 +1,68 @@
...
+import BigNumber from 'bignumber.js' // in jspm_packages
+
+function fn () {
```
sindresorhus:
```js
function fn() {
```

docs/recipes/jspm-systemjs.md
```diff
@@ -0,0 +1,68 @@
...
+  return Promise.resolve(new BigNumber("1234567890.123456789"))
+}
+
+test('example test', async t => {
+  t.is((await fn()).toString(), "1234567890.123456789")
+})
```
sindresorhus: single-quotes, tab indent, and semicolons

applies to all the code examples

sindresorhus:
Can you add it to the reamde: https://github.com/avajs/ava#recipes

docs/recipes/jspm-systemjs.md
```diff
@@ -0,0 +1,68 @@
+# JSPM + SystemJS for ES6
+
+It requires a special loader helper to correctly resolve `import`s of JSPM
```
sindresorhus: Packages => packages

skorlir:
@dak released a patch of ava-jspm-loader today that was (tangentially) related to `nyc` and may actually apply to your earlier issue

@sindresorhus is there anything else I should do for this PR? I made your requested changes.

kdex:
@skorlir @dak [sync](https://www.npmjs.com/package/sync) might be a way to wrap `jspm.import` in a synchronous way so people could import their jspm dependencies using the `import` syntax.

sindresorhus:
Looks good! :)

Thanks @skorlir and @dak.

PR #942, opened 2016-06-28
Support `karma-ava`

jamestalmage:
This does a few things:

 - Moves anything used in child processes that the browser can't replicate into `lib/process-adapter.js` (so that browser implementations can implement that missing functionality).

 - Separates out `prefix-title` so `karma-ava` can use it.

It enables use of [`karma-ava`](https://github.com/avajs/karma-ava), which is very preliminary right now, but does work:

screenshot 2016-06-27 21 38 04

--

This also includes two commits from #896


lib/process-adapter.js
```diff
@@ -0,0 +1,110 @@
...
+			if (filename !== testPath) {
+				dependencies.push(filename);
+			}
+			wrappedHandler(module, filename);
+		};
+	});
+};
+
+module.exports.startTests = function () {
+	console.log('Do not use processAdapter.startTests() in Node. Only browsers');
```
sindresorhus: Why? What is it for? I can't see it being called here at all.

`Node` => `Node.js`

lib/run-status.js
```diff
@@ -1,11 +1,11 @@
 'use strict';
 var EventEmitter = require('events').EventEmitter;
-var path = require('path');
 var util = require('util');
 var chalk = require('chalk');
 var isObj = require('is-obj');
 var flatten = require('arr-flatten');
 var figures = require('figures');
+var prefixTitle = require('./prefix-title');
```
sindresorhus: Great! This should have been done regardless of browser support :)

vdemedes:
@jamestalmage we are still not settled on #896, would be cool to finalize discussion on that one before including it here.

lib/process-adapter.js
```diff
@@ -0,0 +1,110 @@
...
+
+module.exports.send = function (name, data) {
+	process.send({
+		name: 'ava-' + name,
+		data: data,
+		ava: true
+	});
+};
+
+module.exports.on = process.on.bind(process);
```
vdemedes: can just be `exports.on`

lib/process-adapter.js
```diff
@@ -0,0 +1,110 @@
...
+};
+
+module.exports.installDependencyTracking = function (dependencies, testPath) {
+	Object.keys(require.extensions).forEach(function (ext) {
+		var wrappedHandler = require.extensions[ext];
+		require.extensions[ext] = function (module, filename) {
+			if (filename !== testPath) {
+				dependencies.push(filename);
+			}
+			wrappedHandler(module, filename);
```
vdemedes: Missing newline.

lib/process-adapter.js
jamestalmage: Yeah. That can be dropped. A mirror existed in [`browser-adapter`](https://github.com/avajs/karma-ava/blob/master/lib/browser-adapter.js) at one point, but that was early on. I just forgot to remove this.

jamestalmage:
>  we are still not settled on #896, would be cool to finalize discussion on that one before including it here.

This just solidifies the need for #896. `send-to-parent.js` which was debated there, is now just merged with `process-adapter.js`. We don't want to check `process.send` and throw the "Must use AVA CLI" error in the browser, so the child-process check belongs in `process-adapter.js`.

sindresorhus:
Exciting! :) Looks good to me.

@novemberborn @sotojuan ?

sindresorhus:
I fixed the merge conflicts and landed this to unblock other work. Great work James :)

PR #945, opened 2016-06-30
Transpile in the main process

### LINKED ISSUE #577: Transpile source files in the main process

jamestalmage:
Fixes #577 

Use the `--precompile` flag to apply transpilation to all file dependencies.

The idea would be to drop `--require babel-register` and use the `--precompile` flag instead.

It's pretty naive right now:
- The `shouldTranspile` method just prevents transpiling `node_modules`, there's probably more logic that needs to be incorporated from Babels `only` and `ignore` options.
- It caches using the hash of the transpilation _output_. This means the transpiler needs to run every time (and consequently `babel-core` gets loaded every time). We could speed things up more by trying to create a cache key from the input and the babel config, but that means resolving and parsing said config.
- It can't handle dynamic requires, and using them will currently cause things to blow up (if the dynamic require needs transpilation). This could be improved by creating our own version of `babel-register` that does not load up `babel-core` except for dynamic requires (so you only pay the price when you actually use a dynamic require).
- The test suite is very basic right now. Just a couple integration tests that verify it has an impact. This could be improved. Specifically, it probably needs some tests with dependencies in `node_modules`, and circular dependencies. It could also stand to have some true unit tests instead of just integration tests.
## 

I would like it if a few people gave it a try to see if it improved performance at all. Especially people with lots of test files that are using `--require babel-register`. (paging @kentcdodds).


kentcdodds:
Wahoo! Thanks for working on this. I'll give it a look in an hour or two and report back.

jamestalmage:
So, some anecdotal numbers from a small project with just five tests:

```
ava --precompile              2.03s user  0.29s system  181% cpu  1.280 total
ava --require=babel-register  4.18s user  0.65s system  370% cpu  1.304 total
```

Total execution time is a smidge faster, but CPU time is much reduced (looks to be by about 50%).

jamestalmage:
Wrote a copy script to turn my 7 test files into 70 (230 tests):

```
ava --require=babel-register  48.46s user  7.36s system  725% cpu  7.695 total

ava --precompile              16.34s user  2.41s system  577% cpu  3.250 total
```

nfcampos:
either I'm not understanding how it works or it doesn't seem to work for me, when removing `--require babel-register` and replacing it with `-precompile` I'm getting this, which suggests that file is not being run through `babel`
screenshot 2016-06-30 22 39 21

jamestalmage:
Hmm. Try it without the concurrency flag.

jamestalmage:
Make sure to clear the cache: 

```
rm -rf node_modules/.cache
```

nfcampos:
clearing the cache didn't help

kentcdodds:
Mine was interesting... So I just remembered that one of my tests has dynamic requires (so we instrument all our code for coverage), so I noticed the output indicating that which was cool. The tests went from 41 seconds to 19 seconds which is awesome. But I ended up with 5 test failures due to "Unexpected reserved word" and another test failure that's kind of interesting:

```
   6. Uncaught Exception
   TypeError: Expected `fromDir` and `moduleId` to be a string
    at module.exports (/Users/kdodds/Developer/paypal/p2pnodeweb/node_modules/resolve-from/index.js:7:9)
    at /Users/kdodds/Developer/paypal/p2pnodeweb/node_modules/ava/lib/precompiler.js:68:11
    at Array.map (native)
    at Precompiler.normalizeDependencies (/Users/kdodds/Developer/paypal/p2pnodeweb/node_modules/ava/lib/precompiler.js:67:4)
    at Precompiler.createHash (/Users/kdodds/Developer/paypal/p2pnodeweb/node_modules/ava/lib/precompiler.js:86:26)
    at Api._runFile (/Users/kdodds/Developer/paypal/p2pnodeweb/node_modules/ava/api.js:56:39)
    at Promise.map.concurrency (/Users/kdodds/Developer/paypal/p2pnodeweb/node_modules/ava/api.js:283:34)
    at tryCatcher (/Users/kdodds/Developer/paypal/p2pnodeweb/node_modules/ava/node_modules/bluebird/js/release/util.js:16:23)
...
```

I did clear the cache (note, with no cache it actually takes 26 seconds, but subsequent runs are 19ish seconds). I do have the concurrency option enabled.

So far these results are promising though!

nfcampos:
so, it appears that some test files do run correctly, though not all, so it looks like you're finding the dependencies of some test files and missing the dependencies of others

jamestalmage:
For dynamic requires, we probably need to give you a `--force-precompile=pattern` option. It won't be smart about what to precompile in that case, but just do everything that matches the force pattern.

Try with higher concurrency. You should be able to handle more now that `babel-register` isn't hogging memory in each thread.

I will look into that other error later tonight

jamestalmage:
Are any of these open source projects?

kentcdodds:
Sorry, not mine

nfcampos:
mine isn't either

nfcampos:
if any part of the ava or babel config or the file structure helps just ask

jamestalmage:
Probably file structure and the list of imports for each.

nfcampos:
[deps.json.zip](https://github.com/avajs/ava/files/343217/deps.json.zip)
this is a json map of files (sources and test files, test files are those that have \__tests__ in the path) to array of dependencies (each dependency is an array `[string_in_the_import_statement, absolute_path]`). does this help?

lib/precompiler.js
```diff
@@ -0,0 +1,111 @@
...
+
+Precompiler.prototype.buildEntry = function (filename) {
+	if (this._cache[filename]) {
+		return;
+	}
+
+	var entry = this._cache[filename] = {};
+
+	var result = babelCore().transformFileSync(filename, {
+		plugins: [detective]
```
geowarin: I think if you don't merge this with the babelConfig already defined, this will implicitly be merged with the `.babelrc`.
Which is totally cool if the babel option is set to "inherit". Otherwise, metadata generation might fail.

```js
var config = objectAssign({}, this.babelConfig, {
   plugins: this.babelConfig.plugins.concat(detective)
})
```

geowarin: See [this repo](https://github.com/geowarin/repro-ava-precompile) failing with [this config](https://github.com/geowarin/repro-ava-precompile/blob/master/package.json#L26-34)

jamestalmage: > this will implicitly be merged with the .babelrc.

Yes. That's the idea. That is only precompiling your sources. We already precompile your tests in `caching-precompiler`.

jamestalmage: @geowarin - see https://github.com/geowarin/repro-ava-precompile/issues/1 to fix yours.

jamestalmage:
@nfcampos - What's up with importing `.css` files? Are they normally processed by `babel-register`?

nfcampos:
@jamestalmage it's for using `css-modules`, they're dealt with by https://github.com/css-modules/css-modules-require-hook

jamestalmage:
So `babel-register` is completely uninvolved? The PR as it stands only precompiles `.js` files that are not in `node_modules`.

nfcampos:
prior to this PR I had this array in the `require` entry in the `ava` config
```js
"require": [
  "babel-register",
  "babel-polyfill",
  "css-modules-require-hook/preset"
]
```
to try out this PR I removed `babel-register` from that array and left the other two.
running `ava --precompile` after removing it resulted in some of the source files being transpiled and some not being transpiled

jamestalmage:
@nfcampos - do you know which files?

jamestalmage:
@nfcampos 
Also, from your `.json` file. Why is `ava` resolving to `/Users/nuno/cp/ava/index.js` and not something in `node_modules`?

nfcampos:
@jamestalmage because it's linked with `npm link ava`?

jamestalmage:
Ah. I see.

nfcampos:
@jamestalmage alright, I've figured out what the problem is
```js
// abc.js
export const abc = 1 
```
```js
// def.js
export {abc} from './abc' 
```
```js
// pass.js
import test from 'ava'

import {abc} from './abc'

test('abc', t => t.pass())
```
```js
// fail.js
import test from 'ava'

import {abc} from './def'

test('abc', t => t.pass())
```
`pass.js` works, `fail.js` doesn't, for some reason you're not picking up on re-exports as dependencies

jamestalmage:
> for some reason you're not picking up on re-exports as dependencies

Ah yes. I'm definitely not!

jamestalmage:
[This](https://github.com/avajs/babel-plugin-detective/pull/5) should fix the re-exports problem.

jamestalmage:
@nfcampos - New version of `babel-plugin-detective` published, and updated here. Should fix the re-exports problem.

nfcampos:
@jamestalmage it work now, thanks
time with `babel-register`
```
real	0m48.368s
user	2m28.899s
sys	0m15.420s
```
time with `--precompile`
```
real	0m32.534s
user	1m27.033s
sys	0m10.058s
```

jamestalmage:
@kentcdodds - Your stack trace is weird. Are you calling the AVA CLI within an AVA test? Based on the stacktrace, I would expect it was throwing in the top level process. But that line `6. Uncaught Exception`, makes it look like a failure in an isolated test child process.

Either way, I've wrapped the offending `resolveFrom` call in a `try/catch` and added some debugging output. If you could rerun and repost the stacktrace, that would be awesome.

jamestalmage:
FYI: I've opened PR https://github.com/babel/babel/pull/3564, that I hope gives us the best of both worlds: main file transpilation where possible, and deferring to `babel-register` for tests that have dynamic requires (those tests with dynamic requires would take the same performance hit they currently do, but hopefully that's only a portion of your test suite).

sindresorhus:
@jamestalmage Is this PR blocked by https://github.com/babel/babel/pull/3564?

jamestalmage:
No, but we don't have a good solution for dynamic requires without it.

jamestalmage:
Also, that PR is just a refactor. It doesn't introduce added functionality to Babel. If it ever gets merged, it could be used to write our own version of `babel-register` that integrates with main process precompilation or a way to make babel-register not load up tons of dependencies until needed.

We could merge this, and work towards improving Babel separately

STRML:
In terms of caching etc., [babel-loader](https://github.com/babel/babel-loader) does a really good job of that & reading your .babelrc, babel versions, etc., to make a pretty great fast cache that is busted properly every time I've used it.

jamestalmage:
Added a commit documenting this and `--concurrency` (`docs/experimental-features.md`).

docs/experimental-features.md
```diff
@@ -0,0 +1,36 @@
+The following features are considered "Experimental", in that their exact behavior is not yet considered locked. In general, we want to keep these features around long term (they are unlikely to disappear completely), but significant breaking changes are fair game until we are confident in the implementation. For core features, we generally try to accompany any breaking changes with codemods to ease the pain of upgrading, but users should not expect the same for experimental features.
```
sindresorhus: Can you add a heading `# Experimental Features`?

docs/experimental-features.md
```diff
@@ -0,0 +1,36 @@
...
+
+Our recommendation is to go ahead and use these features *if* they provide significant benefits to your build.
+
+# Limited Concurrency
+
+Enabled via a CLI flag (`--concurrency=5`) or via the `ava` section of `package.json` (`concurrency:5`).
```
sindresorhus: `concurrency:5` => `"concurrency": 5`

to make it copy-pasteable.

docs/experimental-features.md
```diff
@@ -0,0 +1,36 @@
...
+
+This will limit the number of test files spawned concurrently during the test run. The default is unlimited, and will cause all test files to be spawned at once. This can overload some systems if there are many, many test files. Especially on CI machines. Use the concurrency config to limit the number of processes. You may need to experiment to determine the optimal number of processes.
+
+## Caveats
+
+Use of `test.only` does not work correctly when concurrency is limited. We think this is a solvable problem.
```
sindresorhus: Can you linkify the text here to the relevant issue?

sindresorhus: I know it's linked below, but would be useful to linkify it here too.

docs/experimental-features.md
```diff
@@ -0,0 +1,36 @@
...
+
+We believe limited concurrency is valuable enough to eventually become the default. We would love your help determining what the default concurrency limit should be (See [#966](https://github.com/avajs/ava/issues/966)).
+
+
+# Precompiling Sources
+
+AVA already precompiles your *test* files in the main process. This removes the need for loading `babel-register` in every child process just to compile the test. This has significant performance benefits. Unfortunately, this does not help users who want to write their project *source* files using the latest language features, as they often still need to use the `--require=babel-register` flag.
+
+Enabled via a CLI flag (`--precompile`) or via the `ava` section of `packages.json` (`precompile:true`), this feature will cause AVA to precompile all required sources via Babel in the main process before launching the child process. If you use this feature, you **must** remove `--require=babel-register` from your config to see any benefits.
```
sindresorhus: `precompile:true` => `"precompile":true`

docs/experimental-features.md
```diff
@@ -0,0 +1,36 @@
...
+Use of `test.only` does not work correctly when concurrency is limited. We think this is a solvable problem.
+
+## Stability
+
+Fixing the `test.only` bug is the main barrier to this becoming a stable core feature. It's proven invaluable to many users, so we aren't likely to remove it unless some significantly better solution is discovered.
+
+We believe limited concurrency is valuable enough to eventually become the default. We would love your help determining what the default concurrency limit should be (See [#966](https://github.com/avajs/ava/issues/966)).
+
+
+# Precompiling Sources
```
sindresorhus: Worth adding a note in the Babel recipe about this and linking here?

docs/experimental-features.md
```diff
@@ -0,0 +1,36 @@
...
+
+AVA already precompiles your *test* files in the main process. This removes the need for loading `babel-register` in every child process just to compile the test. This has significant performance benefits. Unfortunately, this does not help users who want to write their project *source* files using the latest language features, as they often still need to use the `--require=babel-register` flag.
+
+Enabled via a CLI flag (`--precompile`) or via the `ava` section of `packages.json` (`precompile:true`), this feature will cause AVA to precompile all required sources via Babel in the main process before launching the child process. If you use this feature, you **must** remove `--require=babel-register` from your config to see any benefits.
+
+## Caveats
+
+It does not work with relative requires. You must use strings in your require statements, expressions will not work. This is very similar to the restrictions imposed by packagers like `browserify`. If you need to make relative requires, you must still use `babel-register`, so this is unlikely to be helpful.
+
+It only works with Babel 6. We still need to find a solution for typescript users.
```
sindresorhus: typescript => TypeScript

sindresorhus:
I've commented on some nitpick, but generally looks great!

kentcdodds:
I think that I've found why some of my tests are broken. This apparently doesn't transpile files that are required with [`proxyquire`](http://npm.im/proxyquire). This totally makes sense. Do you have any thoughts about what I could do to get around this? Or do you suppose this could be taken into account somehow?

dananichev:
@jamestalmage Does it work with basic setup of ```nyc```?

E.g.:
```
  "nyc": {
    "lines": 100,
    "statements": 100,
    "functions": 100,
    "branches": 100,
    "reporter": [
      "lcov",
      "text",
      "html"
    ],
    "include": [
      "src/**/*.js"
    ]
  }
```
And **w/o** ```babel-plugin-istanbul``` or ```babel-plugin-__coverage__```.

Because with ```babel-require``` ```nyc``` produces coverage report but not with ```--precompile```

dananichev:
@jamestalmage nvm. Decided to remove ```babel-plugin-rewire``` and switch back to ```babel-plugin-istanbul``` (due to incompatibility between them). Works fine with ```--precompile```.

michaelgilley:
Hi all! Great work on the project. My team seeing really slow test times along with thrashed cpus when testing > 30 test files in a React app. This is especially the case for those using a VM (Ubuntu on PC). Any updates on this feature or ideas how to mitigate this issue? Thanks!!

jfairbank:
Was going to pop in here too. I've run comparisons between `"require": ["babel-register"]` and `"precompile": true`, and this feature consistently halves our test suite run time, so I'm eager to see it merged.

@michaelgilley Have you tried the experimental `--concurrency` option for CPU thrashing? Tweaking the number of concurrent tests stopped the processes from overtaking my cores. I took some measurements too, and it didn't seem to slow down our test suite either.

michaelgilley:
@jfairbank Yes, have tried that but as noted in #966 it appears to affect the process differently based on what's available including what other things are running in the background. @nickjvm has had issues running ava in his VM. It's even caused active Chrome tabs to crash! It seems like the sort of thing that ought to be baked into ava - to programmatically check what concurrency level it should be running. When running ava in npm scripts and often behind nyc for coverage it becomes awkward for a team to figure out how to individually configure ava to run correctly on their machines.

michaelgilley:
What's the status on this PR? It seems pretty stale. We really need something like this to be done. With > 100 test files this really becomes essential. Any idea when Ava will be updated to support transpiling once, in the main process before tests are run?

For now we are setting concurrency to 6 but that means that 400 tests are taking at least several minutes to run.

sotojuan:
Agreed with above but IIRC James is quite busy right now and probably not able to work on this.

JaKXz:
@sindresorhus would it be possible to get the documentation fixes merged in a separate PR so the `--precompile` flag can be shipped?

novemberborn:
I need to write up my thoughts but I'm doing some experiments with speeding up how we bootstrap Babel. That may make it feasible to load Babel in the test workers again, and even allow us to build in source compilation.

michaelgilley:
Might I suggest taking a look at how Jest does this? We ended up switching and it's much, much faster. It may be prudent to take a page from their book. 😄

vdemedes:
@michaelgilley I believe Jest does isolation via `vm` module, which indeed provides a sandbox for the tests. While it's an interesting solution, we are aiming for 100% isolation using forked `node` processes. As mentioned up the thread, @novemberborn has alternative speed-up implementation in the labs and there's also this one by @jamestalmage. Performance is one of our top concerns, we are constantly looking on how to speed your tests up.

novemberborn:
Closing in favor of [RFC001](https://github.com/avajs/ava/blob/master/docs/specs/001%20-%20Improving%20language%20support.md).

PR #946, opened 2016-07-01
Docs: Fix link to issue 751 and add link to french translation

forresst:
* Fix link to issue 751
* Add link to french translation


sindresorhus:
Oh wow. That translation arrived fast! :)

PR #951, opened 2016-07-06
Improve failure output of the default mini reporter

### LINKED ISSUE #891: Improve failure output
With the following:

``` js
'use strict';
module.exports = function () {
    throw new Error('foo');
};
```

``` js
import test from 'ava';
import fn from './';

test('unicorn', t => {
    t.true(fn());
});

test('rainbow', t => {
    t.fail();
});
```

I get the following using the default reporter:

screen shot 2016-05-31 at 14 11 11

The current output is a bit messy and ambiguous. For example the stack trace indent is different for the two lines.

I've done a quick mockup of what I'd like it to look like:

screen shot 2016-05-31 at 14 18 29

Note that I reduced the indent of `2 failed` from 3 to 2 spaces. And I changed `failed with "foo"` to `Error: foo` to make it clear it failed with an error and what type of error. Could be a ReferenceError for example which would be lost in the previous output.

Happy to improve upon it. I want our failure output to be as readable as possible.

Here's what I used to generate the mockup:

``` js

console.log(`
  ${chalk.red('2 failed')}

  unicorn
  ${chalk.red('Error: foo')}
   ${chalk.dim('module.exports (index.js:3:8)')}
   ${chalk.dim('Test.fn (test.js:5:9)')}

  rainbow
  ${chalk.red('Test failed via t.fail()')}
   ${chalk.dim('Test.fn (test.js:9:4)')}`);
```
## 

Thoughts?

Is there a way we could improve the `Test.fn` name in the stack trace?

Should we place the filename:line:column first? Usually the function name isn't very useful, so it's usually the line:column you care about.

sotojuan:
Fixes #891.

Doing these during my lunches at work.

- [x] Reduce indentation of test headers
- [x] Test title color change and cleanup
- [x] `failed with x` to `Error: x`
- [x] Reduce indentation of test content
- [x] Error stack color change and cleanup


sotojuan:
@avajs/core Almost done with this but let me know if I can take care of any nitpicks—might have to pull down my branch to check it out but will try to post screenshots tomorrow.

vdemedes:
@sotojuan Good stuff!

sotojuan:
Thanks! Las thing is the error stack's first line having a different color/indentation than the rest. Currently it's all one big string—what's a good strategy here @sindresorhus? Manually take the first line out?

sindresorhus:
Yes

sotojuan:
All right, getting close to being done. Below is a screenshot (not sure why the stacks got longer from when I started):

screen shot 2016-08-08 at 3 39 21 pm

Probably a lot of nitpicks to do but let me know how it looks :-)

sindresorhus:
> not sure why the stacks got longer from when I started):

See: https://github.com/tapjs/stack-utils/issues/14#issuecomment-238026000

sindresorhus:
Looking good :)

The test output should be vertically aligned on the left with `2 failed`.

sotojuan:
Should be aligned now!

sindresorhus:
@sotojuan I think you forgot to push some commits.

This is what I'm seeing:

screen shot 2016-08-11 at 15 32 01

It's not aligned and the stack should not be red.

sindresorhus:
Also notice from the original issue:

>  And I changed `failed with "foo"` to `Error: foo` to make it clear it failed with an error and what type of error. Could be a ReferenceError for example which would be lost in the previous output.

Currently, if I throw a `TypeError` in the test, it will still show `Error: foo`, but it should show `TypeError: foo`.

sotojuan:
That's weird, I guess I did. I'll check when I get to work and look at the rest of the issues—thanks.

sindresorhus:
And the merge conflict needs fixing ;)

sotojuan:
All right, should be better now.

sindresorhus:
Using the code in the original issue, this is what I'm seeing:

screen shot 2016-08-13 at 13 46 02

Notice how the stack trace is indented consistently, and `undefined` shows up twice.

-

Also think we might need to indent the stack trace two spaces instead of 1. Right now it looks accidentally unaligned.

sotojuan:
Huh, weird. On my computer it looks a bit different (no power assert). I'll take care of the spacing in the stack trace first as that's simple and investigate.

sotojuan:
Here's what I am seeing (after indenting stack one more space). Not sure why ours differ so much—I merged latest master, re-linked, and reinstalled Node modules.

screen shot 2016-08-13 at 6 56 52 pm

sindresorhus:
Hmm, must have been me doing something weird.

Anyways, tried it fresh now, and with:

```
import test from 'ava';

const fn = () => 'foo';

test('unicorn', t => {
    t.is(fn(), 'bar');
});
```

I get:

screen shot 2016-08-14 at 01 15 39

Notice how the `foo` is printed twice and the stack is indented way too much.

sotojuan:
Ah, I got it—happens on power assert stuff. Will get it fixed up!

sotojuan:
This is what I have now:

screen shot 2016-08-14 at 9 54 32 am

Let me know if it looks good so I can take care of the tests and push!

vdemedes:
Shouldn't we erase stack trace lines with `next_tick.js`? Also, looks like the newline is missing at the end of output.

sotojuan:
Good catch on the newline—I think I accidentally removed it earlier. About the stacks, they just randomly got long while I was working on this so I'm not sure when or why it happened 😅 . Is `next-tick` something on master that we're using?

sindresorhus:
@vdemedes Already answered: https://github.com/avajs/ava/pull/951#issuecomment-238353808

sindresorhus:
Super work @sotojuan :)

PR #961, opened 2016-07-12
prevent nested tests #948

### LINKED ISSUE #948: Nested test functions hide errors
## Description

I was just reviewing a PR where a colleague wrote tests in a nested way.

While the sub-tests never execute / show up in the test output they also did not create an error.

I would expect such code as the example below to at least throw an error.

### Test Source

```js
import test from 'ava';

test('outer', (t) => {

  test('innter', (t) => {
    t.fail();
    throw new Error('ERR');
  });

});
```

### Console Output

```sh
=> ./node_modules/.bin/ava -v avaTest.js

  ✔ outer

  1 test passed [16:35:53]
```

### Config

No config in package.json

### Command-Line Arguments

```
/ava -v avaTest.js
```

## Relevant Links

- If your project is public, link to the repo so we can investigate directly.
- **BONUS POINTS:** Create a [minimal reproduction](http://stackoverflow.com/help/mcve) and upload it to GitHub. This will get you the fastest support.

## Environment

```
Node.js v5.9.0
darwin 15.5.0
ava 0.15.2
npm 3.8.6
```

asafigan:
This is my first pull request so any feedback is welcome.

fixes #948 

Added a test in the runner test file which expects an error to be thrown when test is called in a test. I don't know if this is the right file to put this test, but it seem like a good place since my solution only changed the runner.js file.

My solution was to add a new property to runner called 'hasBegunRunning' which is set to true in the run method. The chaining function checks this property and throws an error when it is true.

I don't know if this is a good solution but it is simple and it works. Also, I don't know if I need more tests for this pull request.

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -63,7 +63,8 @@ optionChain(chainableMethods, function (opts, args) {
 	var macroArgIndex;
 
 	if (this.hasBegunRunning) {
-		throw new Error('Cannot have nested tests or hooks');
+		throw new Error('All tests and hooks must be declared synchronously in your ' +
+		'test file, and cannot be nested within other tests or hooks.');
 	}
 
 	if (typeof args[0] === 'string') {
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -46,6 +46,7 @@ function Runner(options) {
 
 	this.results = [];
 	this.tests = new TestCollection();
+	this.hasBegunRunning = false;
 	this._bail = options.bail;
 	this._serial = options.serial;
 	this._match = options.match || [];
@@ -60,6 +61,10 @@ optionChain(chainableMethods, function (opts, args) {
 	var title;
 	var fn;
 	var macroArgIndex;
+
+	if (this.hasBegunRunning) {
+		throw new Error('Cannot have nested tests or hooks');
+	}
 
 	if (typeof args[0] === 'string') {
 		title = args[0];
@@ -195,6 +200,8 @@ Runner.prototype.run = function (options) {
 	}
 
 	this.tests.on('test', this._addTestResult);
+
+	this.hasBegunRunning = true;
 
 	return Promise.resolve(this.tests.build(this._bail).run()).then(this._buildStats);
 };
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -46,7 +46,7 @@ function Runner(options) {
 
 	this.results = [];
 	this.tests = new TestCollection();
-	this.hasBegunRunning = false;
+	this.hasStarted = false;
 	this._bail = options.bail;
 	this._serial = options.serial;
 	this._match = options.match || [];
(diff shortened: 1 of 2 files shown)


lib/runner.js
```diff
@@ -61,6 +62,10 @@ optionChain(chainableMethods, function (opts, args) {
 	var fn;
 	var macroArgIndex;
 
+	if (this.hasBegunRunning) {
+		throw new Error('Cannot have nested tests or hooks');
```
jamestalmage: This could also happen with:

```js
test(async t => {
  // stuff
});

setImmediate(() => {
  test(t => {
    // not allowed!
  });
});
```

The message should include something about declaring tests async.

> All tests and hooks must be declared synchronously in your test file, and cannot be nested within other tests.

jamestalmage:
> I don't know if I need more tests for this pull request.

Add a second one that tests what I talked about [here](https://github.com/avajs/ava/pull/961#discussion_r70534537), and I think you are covered.

```js
// test
var r = new Runner();

r.test(() => Promise.resolve());
r.run()

t.throws(() => {
  r.test(() => {});
}, {message: ...})
```

> This is my first pull request

One of the best first PR's I have seen. Very good job. 🎉

asafigan:
also I didn't know what type of error to throw

asafigan:
Ok, I will add that test and also tweak the error message and the second test

asafigan:
Ok I'm gone with those changes. Let me know if there is anything else I need to change.

sotojuan:
👍 great job!

asafigan:
thanks!

jamestalmage:
LGTM

lib/runner.js
```diff
@@ -46,6 +46,7 @@ function Runner(options) {
 
 	this.results = [];
 	this.tests = new TestCollection();
+	this.hasBegunRunning = false;
```
sindresorhus: I would use `isRunning` instead.

asafigan: I didn't name it is running because it isn't reset to true after the tests are done running.

asafigan: but I don't like `hasBegunRunning` either

sindresorhus: `hasStarted`?

asafigan: I like that better

sindresorhus: Let's go with that then.

asafigan:
Ok, I changed the property name.

asafigan:
Also should we mention in the docs somewhere that you shouldn't have embedded tests or hooks. And also that we have no alternative to describe.

jamestalmage:
Thanks @asafigan!

> Also should we mention in the docs somewhere that you shouldn't have embedded tests or hooks. And also that we have no alternative to describe.

That probably belongs in a "transitioning from mocha/tape" recipe. Not sure it belongs in the core docs.

asafigan:
woot! first contribution! 🎉
Thanks for all of the help on this

sotojuan:
Awesome work @asafigan!

sindresorhus:
Super first PR. Thanks @asafigan! Keep 'em coming ;)

68747470733a2f2f6d656469612e67697068792e636f6d2f6d656469612f695054546a4574313969676e652f67697068792e676966

PR #967, opened 2016-07-16
improve test coverage by adding more tests

dananichev:
#161 tests update.
Also, removed unreachable (?) code from lib/reporters/mini.js


sindresorhus:
Just wanted to say that we appreciate the pull request :) This will have to wait for a little bit as we would like to get https://github.com/avajs/ava/pull/880 landed first.

crazyhuggins:
@sindresorhus Any timeframe regarding when #880 are going to be merged? Seems stall atm.

sindresorhus:
@dananichev #880 is not happening now after all, so we can go forward with this :)

test/beautify-stack.js
```diff
@@ -0,0 +1,36 @@
...
+		'at processImmediate [as _immediateCallback] (timers.js:367:17)\n' +
+		'From previous event:\n' +
+		'	at Object.<anonymous> (/ava/foo.js:24:3)\n' +
+		'at Module._compile (module.js:460:26)\n' +
+		'at Object.Module._extensions..js (module.js:478:10)\n' +
+		'at Module.load (module.js:355:32)\n' +
+		'at Function.Module._load (module.js:310:12)\n' +
+		'at Function.Module.runMain (module.js:501:10)\n' +
+		'at startup (node.js:129:16)\n' +
+		'at node.js:814:3'
```
sindresorhus: It's too fragile to embed the full stack trace, it will change between Node.js versions and AVA versions.

test/beautify-stack.js
```diff
@@ -0,0 +1,36 @@
...
+
+var test = require('tap').test;
+var beautifyStack = require('../lib/beautify-stack');
+
+test('should return empty string w/o any arguments', function (t) {
+	t.is(beautifyStack(), '');
+	t.end();
+});
+
+test('should return full stack', function (t) {
```
sindresorhus: Why should it return full stack?

test/beautify-stack.js
```diff
@@ -0,0 +1,36 @@
...
+		'at Object.Module._extensions..js (module.js:478:10)\n' +
+		'at Module.load (module.js:355:32)\n' +
+		'at Function.Module._load (module.js:310:12)\n' +
+		'at Function.Module.runMain (module.js:501:10)\n' +
+		'at startup (node.js:129:16)\n' +
+		'at node.js:814:3'
+	);
+
+	t.true(result.search('ava/cli.js') !== -1);
+
```
sindresorhus: Should also test to make sure it correctly strips off moot stack info.

test/beautify-stack.js
```diff
@@ -0,0 +1,36 @@
+process.env.DEBUG = 'ava';
+
+var test = require('tap').test;
+var beautifyStack = require('../lib/beautify-stack');
+
+test('should return empty string w/o any arguments', function (t) {
```
sindresorhus: `w/o` => `without`

test/run-status.js
```diff
@@ -68,3 +68,9 @@ test('prefixTitle removes test- from path', function (t) {
 	t.is(runStatus.prefixTitle('backend/test-run-status.js'), 'backend' + sep + 'run-status' + sep);
 	t.end();
 });
+
+test('should successfully init w/o any options provided', function (t) {
```
sindresorhus: `w/o` => `without`

and drop the `should`

> successfully initializes without any options provided

test/reporters/verbose.js
```diff
@@ -424,6 +425,28 @@ test('full-width line when sectioning', function (t) {
 	t.end();
 });
 
+test('write should call console.error', function (t) {
+	var reporter = verboseReporter();
+	var consoleError = console.error;
+	console.error = function () {};
```
sindresorhus: I think sinon handles overriding console.error and restoring.

test/reporters/mini.js
```diff
@@ -455,3 +456,92 @@ test('results with watching enabled', function (t) {
...
+	var expectedOutput = [
+		'\n   ' + colors.error('1 exception'),
+		'\n\n   ' + colors.error('1. Uncaught Exception'),
+		'   ' + colors.stack(JSON.stringify({error: err1})) + '\n'
+	].join('\n');
+	t.is(actualOutput, expectedOutput);
+	t.end();
+});
+
+test('should return empty string', function (t) {
```
sindresorhus: Description needs to be clearer about when it should return empty string. Drop the `should` prefix.

test/reporters/mini.js
```diff
@@ -455,3 +456,92 @@ test('results with watching enabled', function (t) {
...
+	var actualOutput = reporter.finish(runStatus);
+	var expectedOutput = [
+		'\n   ' + colors.error('1 rejection'),
+		''
+	].join('\n');
+	t.is(actualOutput, expectedOutput);
+	t.end();
+});
+
+test('should return description based on error itself if no stack available', function (t) {
```
sindresorhus: `should return` => `returns`

test/reporters/mini.js
```diff
@@ -455,3 +456,92 @@ test('results with watching enabled', function (t) {
...
+	reporter.passCount = 0;
+	reporter.exceptionCount = 0;
+	var err = new Error('failure one');
+	err.type = 'exception';
+	reporter.unhandledError(err);
+	t.is(reporter.exceptionCount, 1);
+	t.end();
+});
+
+test('should silently handle errors w/o body', function (t) {
```
sindresorhus: `w/o` => `without`

test/profile.js
```diff
@@ -19,6 +19,15 @@ function run(files) {
 	return execa(process.execPath, [profileScript].concat(arrify(files).map(fixture)), {cwd: cwd});
 }
 
+test('should run profiler and throw an error when invoked w/o files to run', function (t) {
```
sindresorhus: `w/o` => `without`

sindresorhus:
Nice work @dananichev :)

test/reporters/mini.js
dananichev: @sindresorhus the only thing i know about ```MiniReporter.clear()``` that it used in ```Logger.clear()``` which used in ```Watcher``` to separate different test runs. Changed test title to ```returns empty string (used in watcher in order to separate different test runs)``` but i'm not sure about it. Am i missing something?

dananichev:
@sindresorhus sorry for my Perfect English skills (c) :)

test/beautify-stack.js
```diff
@@ -0,0 +1,44 @@
...
+		}
+	});
+
+	var result = beautify(
+		'Error: TypeError\n' +
+		'at null._onTimeout (/ava/cli.js:27:11)\n' +
+		'at Stub.listOnTimeout (timers.js:119:15)\n'
+	);
+
+	t.true(result.search('ava/cli.js') !== -1);
```
sindresorhus: This should use `indexOf` instead. You're actually not using a regex in the search() method.

test/reporters/tap.js
```diff
@@ -194,3 +195,29 @@ test('reporter strips ANSI characters', function (t) {
...
+	var consoleLog = console.log;
+	console.log = function () {};
+	var spy = sinon.spy(console, 'log');
+
+	reporter.write('result');
+
+	t.true(spy.called);
+	t.end();
+	console.log.restore();
+	console.log = consoleLog;
```
sindresorhus: I don't think this is needed. You're already restoring it above.

sindresorhus: Probably need to switch to `stub` here too.

test/reporters/verbose.js
```diff
@@ -424,6 +425,25 @@ test('full-width line when sectioning', function (t) {
...
+});
+
+test('reporter.stdout and reporter.stderr both use process.stderr.write', function (t) {
+	var reporter = verboseReporter();
+	var stub = sinon.stub(process.stderr, 'write');
+	reporter.stdout('result');
+	reporter.stderr('result');
+	t.is(stub.callCount, 2);
+	t.end();
+	process.stderr.write.restore();
```
sindresorhus: This should be moved before the `t.end()` call.

sindresorhus: Same in other places.

test/beautify-stack.js
```diff
@@ -0,0 +1,44 @@
...
+		}
+	});
+
+	var result = beautify(
+		'Error: TypeError\n' +
+		'at null._onTimeout (/ava/cli.js:27:11)\n' +
+		'at Stub.listOnTimeout (timers.js:119:15)\n'
+	);
+
+	t.true(result.indexOf('ava/cli.js') >= 0);
```
sindresorhus: `>= 0` => `!== -1`

test/beautify-stack.js
```diff
@@ -0,0 +1,44 @@
...
+	t.end();
+});
+
+test('strips ava internals and dependencies from stack trace with debug disabled', function (t) {
+	var result = beautifyStack(
+		'Error: TypeError\n' +
+		'at null._onTimeout (/ava/cli.js:27:11)\n' +
```
sindresorhus: I think `/ava/cli.js:27:11` should be `ava/cli.js:27:11`

test/cli.js
```diff
@@ -326,3 +337,34 @@ test('should warn ava is required without the cli', function (t) {
 		t.end();
 	});
 });
+
+test('prefers local version of ava', function (t) {
+	t.plan(1);
+
+	var stubModulePath = path.join(__dirname) + '/fixture/empty';
```
sindresorhus:
```js
var stubModulePath = path.join(__dirname, 'fixture/empty');
```

test/profile.js
```diff
@@ -19,6 +19,15 @@ function run(files) {
 	return execa(process.execPath, [profileScript].concat(arrify(files).map(fixture)), {cwd: cwd});
 }
 
+test('should run profiler and throw an error when invoked without files to run', function (t) {
```
sindresorhus: > runs the profiler and throws an error when invoked without files to run

test/profile.js
```diff
@@ -19,6 +19,15 @@ function run(files) {
...
+	t.plan(1);
+	run()
+		.catch(function (err) {
+			t.ok(err.message.search(/Specify a test file/g));
```
sindresorhus:
```js
t.ok(/Specify a test file/.test(err.message));
```

Also note, no need for the `g` flag.

sindresorhus:
Added some more nitpick feedback inline. Otherwise looks good :)

test/profile.js
dananichev: It is some kind of bad habit. I mean using ```g``` in most of regex-expressions.

test/beautify-stack.js
dananichev: How about ```node_modules/ava/cli.js:27:11```?

sindresorhus: Yes, that probably what it usually is printed as.

dananichev: I mean due to 
``/\/ava\/(?:lib\/)?[\w-]+\.js:\d+:\d+\)?$/``
and 
``/\/node_modules\/(?:bluebird|empower-core|(?:ava\/node_modules\/)?(?:babel-runtime|core-js))\//``
it should be either 
``/node_modules/ava/node_modules/babel-runtime/``
or 
``/ava/cli.js:27:11``
or 
``node_modules/ava/cli.js:27:11`` (because it will work with first regex)

sindresorhus: > node_modules/ava/cli.js:27:11

This

dananichev: Okay

sindresorhus:
The Windows build is failing: https://ci.appveyor.com/project/ava/ava/build/170

dananichev:
@sindresorhus it seems the problem is somewhat related to https://github.com/sindresorhus/execa/issues/35. At least simptoms are similar:
```
[Error: spawn undefined ENOENT]
  errno: 'ENOENT',
  code: 'ENOENT',
  syscall: 'spawn undefined',
...
```

I'm trying to locate the problem behind this. But my VMs are pretty much slow (dunno why though). I will continue to investigate this further from home (with native Windows).

dananichev:
@sindresorhus i hope you are fine with ```err.stderr``` in ```test/profile.js```?

test/profile.js
```diff
@@ -23,7 +23,7 @@ test('runs the profiler and throws an error when invoked without files to run',
 	t.plan(1);
 	run()
 		.catch(function (err) {
-			t.ok(/Specify a test file/.test(err.message));
+			t.ok(/Specify a test file/g.test(err.stderr));
```
sindresorhus: There's no point in having the `g` flag here.

sindresorhus: Fixed it for you when merging.

sindresorhus:
🎉 Looks great now. Thank you @dananichev :)

test/profile.js
dananichev: One of this CMD+Z's was too much i think.

PR #975, opened 2016-07-21
support test macros in typescript definitions

ChristianMurphy:
This brings support for [test macros](https://github.com/avajs/ava#test-macros) in the typescript.
It adds a macro interface, overloads the test function definition, and overloads all combinations of test modifiers.

Similar to #895 but built on top of #884

### DIFF
--- a/types/base.d.ts
+++ b/types/base.d.ts
@@ -124,8 +124,9 @@ export interface Macro<I, E, T> {
 	(t: T, input: I, expected: E): void;
 	title? (providedTitle: string, input: I, expected: E): string;
 }
+export type Macros<I, E, T> = Macro<I, E, T> | Macro<I, E, T>[];
 
 export function test(name: string, run: ContextualTest): void;
 export function test(run: ContextualTest): void;
-export function test<I, E> (run: Macro<I, E, ContextualTestContext> | Macro<I, E, ContextualTestContext>[], input: I, expected: E): void;
-export function test<I, E> (name: string, run: Macro<I, E, ContextualTestContext> | Macro<I, E, ContextualTestContext>[], input: I, expected: E): void;
+export function test<I, E> (name: string, run: Macros<I, E, ContextualTestContext>, input: I, expected: E): void;
+export function test<I, E> (run: Macros<I, E, ContextualTestContext>, input: I, expected: E): void;
--- a/types/base.d.ts
+++ b/types/base.d.ts
@@ -121,12 +121,12 @@ export interface ContextualCallbackTestContext extends CallbackTestContext {
 }
 
 export interface Macro<T, I, E> {
-	(t: T, input: I, expected: E): void;
-	title? (providedTitle: string, input: I, expected: E): string;
+	(t: T, input?: I, expected?: E): void;
+	title? (providedTitle: string, input?: I, expected?: E): string;
 }
 export type Macros<T, I, E> = Macro<T, I, E> | Macro<T, I, E>[];
 
 export function test(name: string, run: ContextualTest): void;
 export function test(run: ContextualTest): void;
-export function test<I, E> (name: string, run: Macros<ContextualTestContext, I, E>, input: I, expected: E): void;
-export function test<I, E> (run: Macros<ContextualTestContext, I, E>, input: I, expected: E): void;
+export function test<I, E> (name: string, run: Macros<ContextualTestContext, I, E>, input?: I, expected?: E): void;
+export function test<I, E> (run: Macros<ContextualTestContext, I, E>, input?: I, expected?: E): void;
--- a/types/base.d.ts
+++ b/types/base.d.ts
@@ -120,13 +120,13 @@ export interface ContextualCallbackTestContext extends CallbackTestContext {
 	context: any;
 }
 
-export interface Macro<T, I, E> {
-	(t: T, input: I, expected: E): void;
-	title? (providedTitle: string, input: I, expected: E): string;
+export interface Macro<T> {
+	(t: T, ...args: any[]): void;
+	title? (providedTitle: string, ...args: any[]): string;
 }
-export type Macros<T, I, E> = Macro<T, I, E> | Macro<T, I, E>[];
+export type Macros<T> = Macro<T> | Macro<T>[];
 
 export function test(name: string, run: ContextualTest): void;
 export function test(run: ContextualTest): void;
-export function test<I, E> (name: string, run: Macros<ContextualTestContext, I, E>, input?: I, expected?: E): void;
-export function test<I, E> (run: Macros<ContextualTestContext, I, E>, input?: I, expected?: E): void;
+export function test(name: string, run: Macros<ContextualTestContext>, ...args: any[]): void;
+export function test(run: Macros<ContextualTestContext>, ...args: any[]): void;
--- a/types/base.d.ts
+++ b/types/base.d.ts
@@ -120,13 +120,13 @@ export interface ContextualCallbackTestContext extends CallbackTestContext {
 	context: any;
 }
 
-export interface Macro<I, E, T> {
+export interface Macro<T, I, E> {
 	(t: T, input: I, expected: E): void;
 	title? (providedTitle: string, input: I, expected: E): string;
 }
-export type Macros<I, E, T> = Macro<I, E, T> | Macro<I, E, T>[];
+export type Macros<T, I, E> = Macro<T, I, E> | Macro<T, I, E>[];
 
 export function test(name: string, run: ContextualTest): void;
 export function test(run: ContextualTest): void;
-export function test<I, E> (name: string, run: Macros<I, E, ContextualTestContext>, input: I, expected: E): void;
-export function test<I, E> (run: Macros<I, E, ContextualTestContext>, input: I, expected: E): void;
+export function test<I, E> (name: string, run: Macros<ContextualTestContext, I, E>, input: I, expected: E): void;
+export function test<I, E> (run: Macros<ContextualTestContext, I, E>, input: I, expected: E): void;
--- a/types/base.d.ts
+++ b/types/base.d.ts
@@ -121,8 +121,8 @@ export interface ContextualCallbackTestContext extends CallbackTestContext {
 }
 
 export interface Macro<T, I, E> {
-	(t: T, input?: I, expected?: E): void;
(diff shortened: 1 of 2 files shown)


sindresorhus:
// @ivogabe @SamVerschueren

SamVerschueren:
@ChristianMurphy I'll try looking into it this evening or tomorrow. If not, don't hesitate to ping me back in a couple of days, quite busy lately :).

types/make.js
```diff
@@ -51,6 +51,8 @@ function generatePrefixed(prefix) {
 				output += '\t' + writeFunction(part, 'name: string', 'void');
 			} else {
 				const type = testType(parts);
+				output += '\t' + writeFunction(part + '<I, E>', 'implementation: Macro<I, E, ' + type + 'Context>, input: I, expected: E');
```
ivogabe: This doesn't allow multiple macro's (like in the `test` function in `base.d.ts`. I would define a type alias, after the definition of Macro in `base.d.ts`:
```typescript
type Macros<I, E, T> = Macro<I, E, T> | Macro<I, E, T>[];
```

ivogabe: Can you put the macro declarations after the original declarations? Editors will then first show the macro-less declaration, and that is probably the most used one.

ChristianMurphy: Done and Done

ivogabe:
Looks good to me now. @SamVerschueren have you taken a look at this?

ChristianMurphy:
// @SamVerschueren

types/base.d.ts
```diff
@@ -120,5 +120,13 @@ export interface ContextualCallbackTestContext extends CallbackTestContext {
 	context: any;
 }
 
+export interface Macro<I, E, T> {
```
SamVerschueren: Actually I would put the `T` argument at the first position. Doesn't it make more sense? Because the macro also provides the testcontext as the first argument.

```ts
export interface<T, I, E>
```

Same for `Macros`

// @ivogabe

SamVerschueren:
Sorry for the late reply guys. Only have one comment. If that's answered/addressed, seems good to merge.

types/base.d.ts
ChristianMurphy: Updated with T as first type param

types/base.d.ts
```diff
@@ -120,5 +120,13 @@ export interface ContextualCallbackTestContext extends CallbackTestContext {
 	context: any;
 }
 
+export interface Macro<T, I, E> {
+	(t: T, input: I, expected: E): void;
+	title? (providedTitle: string, input: I, expected: E): string;
```
SamVerschueren: Last small remark. Should we call this `title: string` instead of `providedTitle: string`? Autocompletion tools show the param names and `title: string` might be better.

// @ivogabe @sindresorhus

types/base.d.ts
```diff
@@ -120,5 +120,13 @@ export interface ContextualCallbackTestContext extends CallbackTestContext {
...
+}
+export type Macros<T, I, E> = Macro<T, I, E> | Macro<T, I, E>[];
+
 export function test(name: string, run: ContextualTest): void;
 export function test(run: ContextualTest): void;
+export function test<I, E> (name: string, run: Macros<ContextualTestContext, I, E>, input: I, expected: E): void;
+export function test<I, E> (run: Macros<ContextualTestContext, I, E>, input: I, expected: E): void;
```
novemberborn: Zero or more arguments can be provided to a macro. Rather than specifying `input` and `expected` I think this should just be `...args: any[]` (not 100% certain on the syntax).

SamVerschueren: @novemberborn Can you provide an example on when to use 0 or 1 argument?

I believe the syntax would then become

`run?: Macros<ContextualTestContext, I, E>, input?: I, expected?: E`

Or is `test` just passing all arguments to the macro?

```js
function macro(t, foo, bar, baz, unicorn) {
    t.is(unicorn, 'unicorn');
}

test('2 + 2 === 4', macro, 'foo', 'bar', 'baz', 'unicorn');
```

novemberborn: > Or is test just passing all arguments to the macro?

It is.

SamVerschueren: That makes sense.

ivogabe: That would require Microsoft/TypeScript#5453. That's not implemented yet, so for now these are the options:

- Use `...args: any[]` - no type checking at all
- Support only two arguments (current)
- Support max two arguments, and mark them optional (`?`) - this means that the amount of arguments isn't checked; a macro with 2 arguments can be run without arguments, and the other way around
- Support zero, one or two arguments

Using `...args: any[]` means that TypeScript doesn't check anything at all, so that's not a good idea in my opinion. The last option means that three interfaces are needed (`Macro0`, `Macro1`, `Macro2`), and 3 x 2 = 6 function overloads for Macros. So that's not ideal, but it might be the best solution. (That option could be extended to any amount of arguments of course. Since the definition is already generated, it wouldn't be very hard to add.)

types/base.d.ts
ivogabe: Renaming it to `title` might be confusing, since you'd then have a function and an argument both named `title`. I did a quick test and autocompletion doesn't show the argument name when implementing it like this:
```typescript
export interface Macro<T, I, E> {
	(t: T, input: I, expected: E): void;
	title? (providedTitle: string, input: I, expected: E): string;
}

const macro: Macro<...> = () => {};
macro.title = (/* cursor */) => "";

```

ChristianMurphy: > Should we call this `title: string` instead of `providedTitle: string`? 

I would agree with @ivogabe I would lean against renaming, it could be confusing. Additionally `providedTitle` is how this is documented in the readme.

> I did a quick test and autocompletion doesn't show the argument name when implementing it like this

I've noticed autocomplete is not suggesting as well, though the type checker is validating correctly.
I'm open to suggestions on how to improve the typing to get autocompletion here.

types/base.d.ts
sindresorhus: > Support max two arguments, and mark them optional (?) - this means that the amount of arguments isn't checked; a macro with 2 arguments can be run without arguments, and the other way around

I think this would be the best solution. At least it would then typecheck the first two arguments. We should open an issue to improve this when https://github.com/Microsoft/TypeScript/issues/5453 is fixed.

ChristianMurphy: Updated to make `input` and `expected` optional

types/base.d.ts
```diff
@@ -121,12 +121,12 @@ export interface ContextualCallbackTestContext extends CallbackTestContext {
 }
 
 export interface Macro<T, I, E> {
-	(t: T, input: I, expected: E): void;
-	title? (providedTitle: string, input: I, expected: E): string;
+	(t: T, input?: I, expected?: E): void;
```
ivogabe: These should not be marked optional here, since that would require a user to make them optional as well. Making them optional in `test` and the generator script is sufficient. (A function with less arguments is assignable to a function type with more arguments.)

ChristianMurphy: Good catch, updated

sindresorhus:
Is this good to merge now?

SamVerschueren:
I'm not entirely sure to be honest. It doesn't feel 100% right to me as a macro could actually be anything.

This implementation was based upon the example in the readme.

```js
function macro(t, input, expected) {
    t.is(eval(input), expected);
}

test('2 + 2 === 4', macro, '2 + 2', 4);
test('2 * 3 === 6', macro, '2 * 3', 6);
```

But what if I want to do this

```js
function macro(t, obj) {
    for (const key of Object.keys(obj)) {
        t.is(eval(obj[key]), key);
    }
}

test('2 + 2 === 4', macro, {4: '2 + 2'});
test('2 * 3 === 6', macro, {6: '2 * 3'});
```

This doesn't follow the `input, expected` order as defined by the typescript definition. And because you can actually pass as many arguments to the macro function as you want, I would suggest adding a couple extra like this

```js
export function test<I, E> (name: string, run: Macros<ContextualTestContext>, arg1?: any): void;
export function test<I, E> (name: string, run: Macros<ContextualTestContext>, arg1?: any, arg2?: any): void;
export function test<I, E> (name: string, run: Macros<ContextualTestContext>, arg1?: any, arg2?: any, arg3?: any): void;
export function test<I, E> (name: string, run: Macros<ContextualTestContext>, arg1?: any, arg2?: any, arg3?: any, arg4?: any): void;
export function test<I, E> (name: string, run: Macros<ContextualTestContext>, arg1?: any, arg2?: any, arg3?: any, arg4?: any, arg5?: any): void;
```

The current implementation forces the user to use it like the example provided in the readme. But macros are much more powerful then that.

sindresorhus:
`test('2 + 2 === 4', macro, {4: '2 + 2'});` should definitely be supported.

ivogabe:
@SamVerschueren Then all parameters are still typed as `any`, so that doesn't give any type safety. Using `...args: any[]` would then even be better, as it allows an arbitrary amount of arguments. Your example will work. When you give `t` and `obj` type annotations, you will get this macro: `Macro<ContextualTestContext, { [ key: number ]: string }, {}>`. The last position isn't used and defaults to `{}`. So, macros with less than two arguments work; macros with more than two parameters do not work.

-----
Maybe a bit off-topic, but I think that it might be better to define macros based on partial application/currying like this: (using the same example)
```typescript
const macro = (obj: { [key: number]: string }) => (t: ContextualTestContext) => {
    for (const key of Object.keys(obj)) {
        t.is(eval(obj[key]), key);
    }
}
test('2 + 2 === 4', macro({4: '2 + 2'}));
test('2 * 3 === 6', macro({6: '2 * 3'}));
```
That doesn't require a special overload for macros, and can thus be typed correctly. It might be an idea to type the macro overload with `...args: any[]` and advise TypeScript users to use currying like in the previous example.

sindresorhus:
Alright, let's go with `...args: any[]` then. We can improve it when https://github.com/Microsoft/TypeScript/issues/5453 is fixed.

> but I think that it might be better to define macros based on partial application/currying like this: (using the same example)

👍 Could be added to the [TypeScript recipe](https://github.com/avajs/ava/blob/master/docs/recipes/typescript.md).

ChristianMurphy:
Updated with `...args: any[]`

sindresorhus:
Thank you @ChristianMurphy for this excellent contribution, and to @ivogabe and @SamVerschueren for reviewing :)

PR #980, opened 2016-07-24
Improve metadata checks

### LINKED ISSUE #847: Improve metadata checks
We currently validate the test `metadata` property in two different places:
- `todo` is checked [here](https://github.com/avajs/ava/blob/7d0f42050d8310d2ca5b1c38b5eb51d3fbb80888/lib/runner.js#L112-L121)
- `always` and `only` are checked [here](https://github.com/avajs/ava/blob/master/lib/test-collection.js#L63-L71)
- We currently allow `failing` on any test or hook, but as discussed [here](https://github.com/avajs/eslint-plugin-ava/issues/108), we probably only want to allow that on tests, not hooks.
- I'm not sure we prevent `skip.only` anywhere, but we probably should.

This should probably all be extracted into `lib/validate-test.js`, with something like this:

``` js
// validate-test.js
module.exports = function(title, fn, metadata) {
  // throws? Then we need to wrap with a try/catch (which is slow).
  // returns a String with the rejection reason, returning null means it's good?
};
```

This would simplify our tests for these validations as well (they could be unit tests for this validation module, instead of integration tests that require spinning up forks).

asafigan:
Fixes #847

validate-test.js checks the following:
- `always`  should only be used with `after` or `afterEach`
- `todo` may not have an implementation, but must have a title
- `todo` should not be combined with any other modifiers. It's simply for documentation. So no `skipped`, no `failing`, no `afterEach`, etc.
- `skip.only` should be a failure
- `failing` shouldn't be used with hooks
- `failing` shouldn't be `skipped`
- `only` shouldn't be used with hooks

All unit tests are in test/validate-test.js. Integration tests are in test/runner.js. Moved some tests from test/test-collection.js to test/runner.js.


vdemedes:
Good work, @asafigan! Sorry for not taking a look sooner. Let's hear one more LGTM and merge it in.

@sindresorhus @jamestalmage @novemberborn @sotojuan @jfmengels looks good to you?

lib/validate-test.js
```diff
@@ -0,0 +1,54 @@
...
+		}
+	}
+
+	if (metadata.skipped) {
+		if (metadata.exclusive) {
+			return '`only` tests cannot be skipped';
+		}
+
+		if (metadata.failing) {
+			return '`only` tests cannot be failing';
```
jfmengels: I'm not well aware of how this works, but the message talks about `only` tests, not `skipped` tests. Are tests skipped due to the presence of an only marked as skipped, or is the message inconsistent with what is checked?

Not sure why failing tests should not be skipped/only-ed by the way.

lib/validate-test.js
```diff
@@ -0,0 +1,54 @@
...
+			return '`todo` tests are not allowed to have an implementation. Use ' +
+			'`test.skip()` for tests with an implementation.';
+		}
+
+		if (typeof title !== 'string') {
+			return '`todo` tests require a title';
+		}
+
+		if (metadata.skipped || metadata.failing || metadata.exclusive) {
+			return '`todo` tests are just for documentation and cannot be used with skip, only, or failing';
```
jfmengels: `skip, only, or failing` --> `skip, only or failing` (remove the last comma)?

lib/validate-test.js
```diff
@@ -0,0 +1,54 @@
...
+		}
+
+		if (metadata.failing) {
+			return '`failing` is only for tests and cannot be used with hooks';
+		}
+
+		if (metadata.todo) {
+			return '`todo` is only for documentation of future tests and cannot be used with hooks';
+		}
+	}
```
jfmengels: Is the fact that hooks are "skipped" handled here?

lib/validate-test.js
vdemedes: I think just the message is wrong and should be fixed.

> Not sure why failing tests should not be skipped/only-ed by the way.

I agree, `failing` tests should be allowed to `skip` and `only`.

asafigan:
Thanks for the feed back guys. I'm going to be busy for awhile, but I will try to look into everything.

lib/validate-test.js
sotojuan: Keep the comma :-) I like Oxford Commas.

sotojuan:
Few nitpicks but overall LGTM. Should be merged by the next push!

vdemedes:
Hey @asafigan, any chance you have some free time to complete this? :)

asafigan:
Sorry I have been so slow on this. I will do it this weekend.

vdemedes:
@asafigan No worries, thanks!

lib/validate-test.js
asafigan: In the issue, it was discussed that skipping hooks should be allowed. Do you think we shouldn't skip them?

asafigan: @jfmengels I don't know what you meant here.

jfmengels: I can't remember why I wrote that either :sweat_smile:

asafigan:
@vdemedes unless there is any other feedback, I believe it ready to merge

vdemedes:
@asafigan Thanks for finishing this up! I'm a little busy now, I'll try to review ASAP, unless someone from the team beats me to it.

lib/validate-test.js
novemberborn: For consistency, should use backticks around `skip`, `only` and `failing`.

sindresorhus:
> Also, @sindresorhus, do we have rules around when error messages end in a period? Looking at the deletions we haven't been consistent about this before, and perhaps we don't care, but wanted to raise it regardless.

See: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/26

sindresorhus:
@asafigan Any chance you would be able to address @novemberborn's feedback so this can be merged?

sindresorhus:
ping @asafigan

asafigan:
So, I need to rewrite the tests in `test/runner.js`. What about reformatting the error messages? I couldn't tell if there was a discussion with that.

sindresorhus:
> So, I need to rewrite the tests in test/runner.js.

Yes, so they're not tied to the internal implementation. Just make them like the existing ones you already have there: https://github.com/avajs/ava/pull/980/files#diff-a0b87ab613d07bba77930609ec360b3dR323

> What about reformatting the error messages? I couldn't tell if there was a discussion with that.

What you have now is fine.

novemberborn:
@asafigan how's it going? Wondering if you're still interested in completing this? Thanks :)

asafigan:
I'm getting major issues when I try to test now.

asafigan:
I pull from master and solved the merge conflicts but now there are a bunch of errors when I run npm test or npm run test-win

asafigan:
I went to my master branch and pulled from upstream master and I get the same issues

lib/runner.js
```diff
@@ -4,6 +4,7 @@ const Promise = require('bluebird');
 const optionChain = require('option-chain');
 const matcher = require('matcher');
 const TestCollection = require('./test-collection');
+const validateTest = require('./test-colection');
```
sindresorhus: You have a typo. You're also importing the same thing twice.

asafigan:
oh yeah I noticed that

asafigan:
the real issue was that I forgot to rerun npm install

asafigan:
I had to fix some other things, too.
Sorry, it took me so long to circle back on this.

asafigan:
Will do

test/runner.js
```diff
@@ -234,7 +234,7 @@ test('skip test', t => {
 
 	t.throws(() => {
 		runner.skip('should be a todo');
-	}, {message: 'Expected an implementation. Use `test.todo()` for tests without an implementation.'});
+	}, TypeError, 'Expected an implementation. Use `test.todo()` for tests without an implementation.');
```
sindresorhus: This is incorrect. You're no longer testing the error message. The third argument is just an assert failure message.

You could do:

```js
t.throws(() => {
	runner.todo('todo', () => {});
}, new TypeError('`todo` tests are not allowed to have an implementation. Use `test.skip()` for tests with an implementation.'));
```

asafigan: thanks for catching that

sindresorhus:
Thank you @asafigan. Excited to finally land this :)

PR #985, opened 2016-08-02
Invoke `resolveCwd` from package.json directory within `require`

### LINKED ISSUE #936: Ensure watch mode works when AVA is called outside the package root
[Issuehunt badges](https://issuehunt.io/r/avajs/ava/issues/936)

Given the following project layout:

```
/package.json
/lib/foo.js
/test/foo.js
```

Watch mode works as expected when invoked like this:

```
cd /
$(npm bin)/ava -w test/foo.js
```

But not if invoked like:

```
cd /test
$(npm bin)/ava -w foo.js
```

See also #863.

> There is a $40.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/avajs/ava/issues/936).

ltk:
Resolves #936

A variable `process.cwd()` was problematic for the invocation of `resolveCwd` on https://github.com/avajs/ava/compare/master...ltk:ltk/fix-936?expand=1#diff-2d5bbda346ebb54b9745d10f2fb5b1b6R37 when using the CLI from outside of the package.json directory.

This PR normalizes the cwd to the pkgDir for every invocation.

### DIFF
--- a/api.js
+++ b/api.js
@@ -24,16 +24,21 @@ function Api(options) {
 	EventEmitter.call(this);
 
 	this.options = objectAssign({
-		cwd: process.cwd(),
 		resolveTestsFrom: process.cwd(),
 		match: []
 	}, options);
 
 	this.options.require = (this.options.require || []).map(function (moduleId) {
+		var originalCwd = process.cwd();
+		if (options.pkgDir) {
+			process.chdir(options.pkgDir);
+		}
+
 		var ret = resolveCwd(moduleId);
 		if (ret === null) {
 			throw new Error('Could not resolve required module \'' + moduleId + '\'');
 		}
+		process.chdir(originalCwd);
 
 		return ret;
 	});
--- a/cli.js
+++ b/cli.js
@@ -130,6 +130,7 @@ var api = new Api({
 	explicitTitles: cli.flags.watch,
 	match: arrify(cli.flags.match),
 	babelConfig: conf.babel,
+	pkgDir: pkgDir,
 	resolveTestsFrom: cli.input.length === 0 ? pkgDir : process.cwd(),
 	timeout: cli.flags.timeout,
 	concurrency: cli.flags.concurrency ? parseInt(cli.flags.concurrency, 10) : 0


api.js
```diff
@@ -24,16 +24,21 @@ function Api(options) {
...
 	this.options = objectAssign({
-		cwd: process.cwd(),
 		resolveTestsFrom: process.cwd(),
 		match: []
 	}, options);
 
 	this.options.require = (this.options.require || []).map(function (moduleId) {
+		var originalCwd = process.cwd();
+		if (options.pkgDir) {
+			process.chdir(options.pkgDir);
```
sindresorhus: Using `process.chdir` is almost always a bad idea. It changes the CWD globally and can cause race issues and all kinds of problems.

api.js
```diff
@@ -24,16 +24,21 @@ function Api(options) {
...
+		}
+
 		var ret = resolveCwd(moduleId);
```
sindresorhus: Should maybe switch to https://github.com/sindresorhus/resolve-from here and pass in the custom `cwd`.

cli.js
```diff
@@ -130,6 +130,7 @@ var api = new Api({
 	explicitTitles: cli.flags.watch,
 	match: arrify(cli.flags.match),
 	babelConfig: conf.babel,
+	pkgDir: pkgDir,
```
sindresorhus: Name it `cwd`?

sindresorhus:
@novemberborn would be the best one to review this.

cli.js
novemberborn: `pkgDir` seems good to me. IMO `cwd` should equal `process.cwd()`.

api.js
novemberborn: 👍 . No need to actually change the working directory.

ltk:
Yeah, I'll update this later today. Thanks for the suggestion!

sindresorhus:
@ltk Friendly ping :)

sindresorhus:
Closing for lack of response.

PR #993, opened 2016-08-10
Fix broken CLI on Windows

### LINKED ISSUE #984: CLI not working on Windows
The CLI is not working for me on Windows, and the issue seems to be with this code:

https://github.com/avajs/ava/blob/f9fe8f242dd58d02ab1110f677e82172652d9d3d/cli.js#L11-L15
```js
if (localCLI && localCLI !== __filename) {
	debug('Using local install of AVA');
	require(localCLI);
	return;
}
```

The problem is with the use of `!==` to compare `localCLI` and `__filename`.  In my case, the two variables hold the following values:

`localCLI: c:\workspace\...\node_modules\ava\cli.js`
`__filename: C:\workspace\...\node_modules\ava\cli.js`

As you can see, the drive letter `c:` is lowercase in `localCLI` and uppercase in `__filename`, thus failing the `!==` check.

Using `!==` to compare paths doesn't seem to be a particularly good idea since there are numerous different ways the exact same path can be encoded.  Not sure what the best practice is for comparing paths in node.js is.  Perhaps [`path.relative()`](https://nodejs.org/api/path.html#path_path_relative_from_to) could be used?

vdemedes:
This PR fixes #984.

Thanks to @eugirdor for reporting the issue and for the fix.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -8,7 +8,10 @@ var debug = require('debug')('ava');
 var resolveCwd = require('resolve-cwd');
 var localCLI = resolveCwd('ava/cli');
 
-if (localCLI && localCLI !== __filename) {
+// Use path.relative() to detect local AVA installation,
+// because __filename's case is inconsistent on Windows
+// see https://github.com/nodejs/node/issues/6624
+if (localCLI && path.relative(localCLI, __filename) !== '') {
 	debug('Using local install of AVA');
 	require(localCLI);
 	return;


cli.js
```diff
@@ -8,7 +8,7 @@ var debug = require('debug')('ava');
 var resolveCwd = require('resolve-cwd');
 var localCLI = resolveCwd('ava/cli');
 
-if (localCLI && localCLI !== __filename) {
+if (localCLI && path.relative(localCLI, __filename) !== '') {
```
sindresorhus: Can you add a comment referencing https://github.com/nodejs/node/issues/6624 for posterity. Otherwise someone will come by and try to simplify the code without understanding what it's solving.

vdemedes: Sure.

vdemedes:
PR updated.

sindresorhus:
👌

PR #996, opened 2016-08-11
Wrapped function for Async

DenisCarriere:
Error was detected by Typescript typings and documented by Ava:

https://github.com/avajs/ava/blob/master/docs/recipes/typescript.md

```
[ts] 
Supplied parameters do not match any signature of call target.
[ts] 
',' expected.
[ts] 
Parameter 't' implicitly has an 'any' type.
(parameter) t: any
```

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -107,7 +107,7 @@ test('foo', t => {
 	t.pass();
 });
 
-test('bar', async t => {
+test('bar', async (t) => {
 	const bar = Promise.resolve('bar');
 
 	t.is(await bar, 'bar');


sindresorhus:
The code in the readme is valid JavaScript, so that's either a problem with the type definition or TypeScript.

DenisCarriere:
Thanks for the feedback, it does sounds like an issue with the Typescript.

Cheers,

sindresorhus:
// @ivogabe

ivogabe:
See Microsoft/TypeScript#8262, fixed in TS2.0

Issue #676, opened 2016-03-22
Display test results in AnyBar


vadimdemedes:
Idea born out of #671.

Links:
- https://github.com/tonsky/AnyBar
- https://github.com/sindresorhus/anybar

jamestalmage:
Interesting. 

The downside is that it is not cross platform. I often develop with Webstorm full screen (navbar not visible). Also, you can't send text (i.e. the failing test title, etc).

cryptoquick:
Could it be made an option? It sounds hella cool!

novemberborn:
How would we implement this? I wonder if we should have a plugin architecture with a documented internal API or something, then these things could be build by others.

sindresorhus:
Sounds nice, but I think it's outside the scope of AVA core. Would be better as a recipe or something.

I just added an additional binary to [`anybar-cli`](https://github.com/sindresorhus/anybar-cli) that makes it so you can just do:

```console
$ anybar-exit ava
```

And have AnyBar turn green/red depending on the exit code.

Do you think that would be a good compromise? Also looking for feedback on how the binary works.

https://github.com/sindresorhus/anybar-cli/commit/747a9562d8e5c99701f6f92382444ea4f03a3f8f

vdemedes:
I think solutions/suggestions from both @novemberborn and @sindresorhus are valid. Also, big +1 on plugin system, @novemberborn!

sindresorhus:
I honestly think a plugin system is way too early and not something we should focus on at the moment. When we open up that floodgate there's no way back. I think IMHO it's rather something we should explore post-1.0.0.

vdemedes:
Sure, I think it's the right thing to do, there's too much stuff on our hands at the moment. We need a "future" label for that kind of stuff.

sindresorhus:
> We need a "future" label for that kind of stuff.

:+1: → https://github.com/sindresorhus/ava/labels/future

PR #1002, opened 2016-08-14
Notifier Reporter

### LINKED ISSUE #671: Growl / Notifications reporter
One thing I still miss from karma is the Growl reporter. When doing TDD I liked to background the actual terminal window and just get the red/green notifications in the corner of my screen. I would only surface the terminal window when something unexpected happened (I wrote a test I expected to fail, and it didn't.  Or, I wrote code to make my latest failing test go green, and it didn't).

Some ideas on how I think messages should look:

**Passing Tests**

> :white_check_mark: 20 tests passed

Keep it simple, no need to list todo's/skipped

**Failing Tests**

> :x: 3 tests failed

Or, maybe show the title if there is only one failing test

> :x: failed test: "ensure it does stuff"

Or, maybe always the title of the first failing test.

> :x: "ensure it does stuff" and 2 other tests failed.

**Internal Errors**

> :warning: There was some problem.

This would be for "no tests received", etc.

---

Also, it might be interesting to make Notifications clickable (I think you can do that), and launch an app according to the `EDITOR` environment variable. We do something similar in XO.

BarryThePenguin:
This is me experimenting so if it's too soon to include this in AVA that's all good :+1: 

New media is used by node-notifier. It is square but looks like it has a bit too much padding

Notify Reporter passes objects instead of strings

Multiple reporters in the `Logger` is the main bit I'm unsure about. At the moment every `reporter.write` gets called with the result of each reporter method call

`reporter.api` didn't look like this was being used... Can include it in another PR if need be

Fixes #671


vdemedes:
AVA does not have a "public" concept of reporters, they are only for internal use and each of them has a specific use case. If you are interested in creating a custom output, create a TAP reporter and execute AVA with `--tap` flag.

As mentioned in the original issue and #676, we might introduce a plugin system, but that definitely won't happen in the near future.

Sad to see your time wasted, that's why I encourage you to discuss upcoming pull requests beforehand in the dedicated issue, to prevent stuff like this. Thank you for understanding ;)

BarryThePenguin:
Sure, totally understand. It was fun learning :grinning:

PR #1007, opened 2016-08-17
Add Flow type definition

### LINKED ISSUE #986: Flow type definition
We already have a great [TypeScript definition](https://github.com/avajs/ava/tree/master/types). We should have one for [Flow](https://flowtype.org/) too! Both of these type checkers are awesome and we shouldn't play favorites. This would also be interesting comparison wise.

Anyone familiar with Flow willing to contribute this?

// @thejameskyle

-

The TypeScript definition is actually partly generated (https://github.com/avajs/ava/pull/884) because it was too tedious to define the chainable test API manually. For example: `test.cb.serial.skip()`. The properties can be in any order. I wonder if that is needed for Flow too.

thejameskyle:
Resolves #986

I wanted to test Flow against one of AVA's test files but it looks like the test suite isn't self-hosted and there doesn't appear to be any integration-like tests for the public api.

I would suggest doing something along those lines and testing both Flow and TypeScript against it, otherwise you'll end up with out of date type definitions like what has already happened with the TS definition and `.always`

### DIFF
--- a/index.js.flow
+++ b/index.js.flow
@@ -5,79 +5,79 @@
  */
 
 type PromiseLike<R> = {
-  then<U>(
-    onFulfill?: (value: R) => Promise<U> | U,
-    onReject?: (error: any) => Promise<U> | U
-  ): Promise<U>;
+	then<U>(
+		onFulfill?: (value: R) => Promise<U> | U,
+		onReject?: (error: any) => Promise<U> | U
+	): Promise<U>;
 }
 
 type ObservableLike = {
-  subscribe(observer: (value: {}) => void): void;
+	subscribe(observer: (value: {}) => void): void;
 };
 
 type SpecialReturnTypes =
-  | PromiseLike<any>
-  | Iterator<any>
-  | ObservableLike;
+	| PromiseLike<any>
+	| Iterator<any>
+	| ObservableLike;
 
 type Constructor = Class<{
-  constructor(...args: Array<any>): any
+	constructor(...args: Array<any>): any
 }>;
 
 type ErrorValidator =
-  | Constructor
-  | RegExp
-  | string
-  | ((error: any) => boolean);
+	| Constructor
+	| RegExp
+	| string
+	| ((error: any) => boolean);
 
 /**
  * Asertion Types
  */
 
 type AssertContext = {
-  // Passing assertion.
-  pass(message?: string): void;
-  // Failing assertion.
-  fail(message?: string): void;
-  // Assert that value is truthy.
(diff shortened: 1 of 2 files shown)


thejameskyle:
Btw the reason for this file's location and naming is so that Flow doesn't require any configuration in order to use the type definitions for AVA. `npm install ava` + `require('ava')` and you're done.

sindresorhus:
Awesome!

You also need to add this to the `files` array in package.json

index.js.flow
```diff
@@ -0,0 +1,175 @@
...
+/**
+ * Misc Setup Types
+ */
+
+declare type PromiseLike<R> = {
+  then<U>(
+    onFulfill?: (value: R) => Promise<U> | U,
+    onReject?: (error: any) => Promise<U> | U
+  ): Promise<U>;
+}
```
sindresorhus: Is there any plan to have this builtin in Flow? Kinda weird having to define it manually in every type definition.

sindresorhus:
Sidenote: Someone really needs to submit Flow syntax highlighting to the syntax highlighter GitHub uses.

sindresorhus:
@thejameskyle Do you know anyone familiar with Flow that could review?

thejameskyle:
@jeffmo @marudor @ryyppy could one of you review this?

ryyppy:
Will have a look! It's very similar to tape's API anyways

ryyppy:
@thejameskyle @sindresorhus Looking good! 

Seriously, I really tried to break the flow definitions... instead it lead me through API... as a second thought, it really is very different to `tape`, especially without any nested tests.

Other than that: Is there a reason why you don't put this in `flow-typed` instead?
I thought it would be problematic to track declaration files in the maintainers repositories, because of breaking flow versions etc.?

ryyppy:
NOTE: I found one case which wasn't really supported by flow:

```
test.after.always('foo', t => {
  t.is //<-- from here flow cannot tell me what's `t`
});
```

thejameskyle:
Thanks for pointing that out @ryyppy, this actually uncovers a really interesting problem with intersection types that needs to be fixed.

For now I've worked around the issue with a bit more repetitive of a type definition. Once we fix the issue I can revert the previous change.

sindresorhus:
@thejameskyle Is this good to merge?

thejameskyle:
Yeah, this is good for now. But should update with no repetition once https://github.com/facebook/flow/issues/2358 is fixed.

sindresorhus:
Yay \o/ Thank you @thejameskyle :)

PR #1008, opened 2016-08-17
fix context in beforeEach and afterEach hooks - fixes #1004

### LINKED ISSUE #1004: TypeScript: Can't compile code accessing `context` in `beforeEach`
## Description

I tried ava v0.16.0 in my project and it causes some compilation error with tsc v2.0.0.  When accessing to `context` in `beforeEach` or `afterEach` hook, 'context is not a property of `t`' error occurred.  I found below code in `types/make.js` so this may be intentional.

```js
if (!has('beforeEach') && !has('afterEach')) {
	type = 'Contextual' + type;
}
```

But I also found some test codes actually access to `context` in `beforeEach` hook.  [Sample code in README](https://github.com/avajs/ava/blob/adc823ecc9747b43d96da70d39cb7040997f8e85/readme.md#before--after-hooks) also does.  I couldn't find why above `!has('beforeEach') && !has('afterEach')` guard was added from any issues or commit histories.

### Test Source

Please save below code as `foo.ts`.

```typescript
import test from 'ava';

test.beforeEach(t => {
    t.context.foo = 42;
});
```

Compilation error by `tsc` command.

```
foo.ts(4, 7): error TS2339: Property 'context' does not exist on type 'TestContext'.
```

### Config

Copy the relevant section from `package.json`: N/A

### Command-Line Arguments

```
tsc --target es6 --module commonjs foo.ts
```

## Relevant Links

- If your project is public, link to the repo so we can investigate directly: https://github.com/rhysd/YourFukurou

## Environment

- OS X 10.11.5
- Node.js v6.2.1
- npm v3.10.4
- ava v0.16.0
- tsc v2.0.0

SamVerschueren:
This small fix fixes #1004. The statement was wrong as it provided a context to `before` and `after` but not to `beforeEach` and `afterEach`.

// @ivogabe

### DIFF
--- a/types/make.js
+++ b/types/make.js
@@ -128,7 +128,7 @@ function testType(parts) {
 		type = 'Callback' + type;
 	}
 
-	if (!has('beforeEach') && !has('afterEach')) {
+	if (!has('before') && !has('after')) {
 		type = 'Contextual' + type;
 	}
 


ivogabe:
👍

PR #1009, opened 2016-08-17
WebStorm recipe

### LINKED ISSUE #998: Recipe for debugging AVA tests with WebStorm
https://github.com/avajs/ava/pull/874 makes it possible. See  https://github.com/avajs/ava/pull/874#issuecomment-227298029 for how to use it.

-

Our current recipes: https://github.com/avajs/ava/tree/master/docs/recipes

prigara:
Fixes #998

### DIFF
--- a/docs/recipes/debugging-with-webstorm.md
+++ b/docs/recipes/debugging-with-webstorm.md
@@ -1,0 +1,19 @@
+# Debugging AVA tests with WebStorm
+
+Starting with version 2016.2, [WebStorm](https://www.jetbrains.com/webstorm/) and other JetBrains IDEs (IntelliJ IDEA Ultimate, PhpStorm, PyCharm Professional, and RubyMine with installed Node.js plugin) allow you to debug AVA tests. 
+
+## Setup
+
+Set breakpoints in the tests you would like to debug.
+
+Add a new *Node.js Run/Debug configuration*: select `Edit Configurations...` from the drop-down list on the top right, then click `+` and select *Node.js*.
+
+In the `JavaScript file` field specify the path to AVA in the project' `node_modules` folder: `node_modules/.bin/ava` on OS X and Linux or `node_modules/.bin/ava.cmd` on Windows.
+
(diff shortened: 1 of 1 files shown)


prigara:
Debugging in WebStorm works as described only with #874

docs/recipes/debugging-with-webstorm.md
```diff
@@ -0,0 +1,19 @@
+# Debugging AVA tests with WebStorm
+
+Starting with version 2016.2, [WebStorm](https://www.jetbrains.com/webstorm/) and other JetBrains IDEs (IntelliJ IDEA Ultimate, PhpStorm, PyCharm Professional, and RubyMine with installed Node.js plugin) allow you to debug AVA tests. 
+
+## Setup
+
+Add a new *Node.js Run/Debug configuration*: select `Edit Configurations...` from the drop-down list on the top right, then click `+` and select *Node.js*.
+
+In the `JavaScript file` field specify the path to AVA in the project's `node_modules` folder: `node_modules/.bin/ava` on OS X and Linux or `node_modules/.bin/ava.cmd` on Windows.
```
sindresorhus: @develar Unrelated, but would be nice if the user could just specify the binary name, like `ava`, instead of the full path, like they can in `npm run script`. Pretty easy to do. You can see how it's done here: https://github.com/sindresorhus/npm-run-path/blob/master/index.js

sindresorhus:
Thank you @prigara :)

docs/recipes/debugging-with-webstorm.md
develar: @sindresorhus Will be addressed as https://youtrack.jetbrains.com/issue/WEB-19540

cc @segrey

machineghost: Thanks so much for providing this recipe! By the way, in my version of Ava the path to the Ava executable was in node_modules/ava/cli.js.

Also (for anyone else coming after me) if your Node root isn't your Webstorm project root you'll need to update the working directory to point to the Node root.

PR #1014, opened 2016-08-20
Upgrade tap and ensure asynchronous tests return promises 

### LINKED ISSUE #935: Upgrade tap, ensure asynchronous tests return promises
With `tap@5.8.0`, if a test returns a promise `tap` now ensures the next test isn't launched until that promise is settled (https://github.com/tapjs/node-tap/pull/249).

We should upgrade to `^5.8.0` and review our unit tests, ensuring that correct promises are returned for every asynchronous test (unless `t.end()` is used). This should help with race conditions in the tests and failures being attributed to the wrong test.

sotojuan:
Fixes #935.

Added `return` when a test finished with a promise and didn't use `t.end`.

There's probably stuff I missed, so let me know!

### DIFF
--- a/test/concurrent.js
+++ b/test/concurrent.js
@@ -331,7 +331,7 @@ test('all sync - end failure - bail', function (t) {
 });
 
 test('all async - no failure - no bail', function (t) {
-	new Concurrent(
+	return new Concurrent(
 		[
 			passAsync('a'),
 			passAsync('b'),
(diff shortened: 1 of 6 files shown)


test/concurrent.js
```diff
@@ -362,7 +362,7 @@ test('all async - no failure - no bail', function (t) {
 });
 
 test('all async - no failure - bail', function (t) {
-	new Concurrent(
+	return new Concurrent(
```
sindresorhus: There are a bunch of `t.end();` that needs to be removed. Like https://github.com/avajs/ava/pull/1014/files#diff-adeeb9de91f4da36bd570d73ef0d19a1R391

sotojuan: Ah, knew I'd miss something! Thanks.

sotojuan:
Updated!

sindresorhus:
Looks great! 👌

PR #1015, opened 2016-08-21
fix typo in readme

zeke:
Watch -> Match


sindresorhus:
Good catch. Thanks Zeke :)

PR #1024, opened 2016-08-29
Add option to disable power-assert - fixes #1017

### LINKED ISSUE #1017: Disable power assert
## Description

As amazing as [`power-assert`](https://npmjs.com/package/power-assert) is, it's constantly causing my team problems. Often it results in reeeeally weird behavior (for example, sometimes suddenly AVA thinks we have infinite tests and runs them like it's got a mind of its own until we kill the process). Often the issues involve JSX, enzyme wrappers, objects with circular references, and upcoming JS features.

Many of the problems that have come up with power-assert in the past have been fixed which is awesome, but it's still just too much magic for my team to stay productive and still causes us a great deal of grief. I love it when it works, but I just don't think that we can keep using it. We're seriously considering changing frameworks because of this (and some perf issues which I know are being worked on).

### Proposal

Expose a config to disable power-assert:

```json
{
  "ava": {
    "powerAssert": false
  }
}
```

Then [this](https://github.com/avajs/ava/blob/81ffd8db1fd68c24f4d54ceeef4397a56539002d/lib/babel-config.js#L78-L85):

```javascript
var defaultPlugins = lazy(function () {
	return [
		espowerPlugin(),
		require('babel-plugin-ava-throws-helper'),
		rewritePlugin(),
		require('babel-plugin-transform-runtime')
	];
});

function build(babelConfig, filePath, code) {
	// ...
	objectAssign(options, {
		inputSourceMap: sourceMap,
		filename: filePath,
		sourceMaps: true,
		ast: false
	});
	// ...
	options.plugins = (options.plugins || []).concat(defaultPlugins());
	// ...
}
```

Could be changed to this:

```javascript
var defaultPlugins = lazy(function (options) {
	return [
		options.powerAssert ? espowerPlugin() : null,
		require('babel-plugin-ava-throws-helper'),
		rewritePlugin(),
		require('babel-plugin-transform-runtime')
	].filter(plugin => !!plugin); // filter out null plugins
});

function build(babelConfig, filePath, code) {
	// ...
	objectAssign(options, {
		inputSourceMap: sourceMap,
		filename: filePath,
		sourceMaps: true,
		ast: false,
		powerAssert: true
	});
	// ...
	options.plugins = (options.plugins || []).concat(defaultPlugins(options));
	// ...
}
```

Here's a `diff`:

```diff
- var defaultPlugins = lazy(function () {
+ var defaultPlugins = lazy(function (options) {
	return [
- 		espowerPlugin()
+ 		options.powerAssert ? espowerPlugin() : null,
		require('babel-plugin-ava-throws-helper'),
		rewritePlugin(),
		require('babel-plugin-transform-runtime')
-   	]
+   	].filter(plugin => !!plugin); // filter out null plugins
});

function build(babelConfig, filePath, code) {
	// ...
	objectAssign(options, {
		inputSourceMap: sourceMap,
		filename: filePath,
		sourceMaps: true,
- 		ast: false
+ 		ast: false,
+ 		powerAssert: true
	});
	// ...
- 	options.plugins = (options.plugins || []).concat(defaultPlugins());
+ 	options.plugins = (options.plugins || []).concat(defaultPlugins(options));
	// ...
}
```

Without actually testing this, it seems to me like it would work. Thoughts? 💭

jfmengels:
Add option to disable power-assert - fixes #1017

Let me know if I've missed anything or if there are better ways to handle this.

cc @kentcdodds


cli.js
```diff
@@ -64,6 +64,7 @@ var cli = meow([
 	'  --tap, -t          Generate TAP output',
 	'  --verbose, -v      Enable verbose output',
 	'  --no-cache         Disable the transpiler cache',
+	'  --no-power-assert  Disable the power-assert feature',
```
sindresorhus: > Disable Power Assert

test/babel-config.js
```diff
@@ -50,7 +50,7 @@ test('uses babelConfig for babel options when babelConfig is an object', functio
 	var fixturePath = fixture('es2015.js');
 	var fixtureSource = fs.readFileSync(fixturePath, 'utf8');
 
-	var options = babelConfigHelper.build(babelConfig, fixturePath, fixtureSource);
+	var options = babelConfigHelper.build(babelConfig, undefined, fixturePath, fixtureSource);
```
kentcdodds: I'd rather give this `undefined` a variable name (like `var disablePowerAssert = undefined`) so people looking at this later know what it's for.

test/babel-config.js
```diff
@@ -89,3 +89,26 @@ test('should reuse existing source maps', function (t) {
...
+
+	var babelConfig = {
+		presets: ['stage-2', 'es2015'],
+		plugins: [customPlugin]
+	};
+
+	var fixturePath = fixture('es2015.js');
+	var fixtureSource = fs.readFileSync(fixturePath, 'utf8');
+
+	var options = babelConfigHelper.build(babelConfig, true, fixturePath, fixtureSource);
```
kentcdodds: Similarly here: `var disablePowerAssert = true`

kentcdodds:
Looks awesome!

cli.js
```diff
@@ -130,6 +131,7 @@ var api = new Api({
 	serial: cli.flags.serial,
 	require: arrify(cli.flags.require),
 	cacheEnabled: cli.flags.cache !== false,
+	disablePowerAssert: cli.flags.powerAssert === false,
```
sindresorhus: Double-negatives are usually bad, better to make it `powerAssert: cli.flags.powerAssert !== false,`

test/babel-config.js
sindresorhus: Might be worth making it an options object instead of arguments.

cli.js
kentcdodds: agreed. So on the CLI it'd be:

> ava --power-assert=false

Or

> ava --power-assert false

?

jfmengels: The CLI handler seems to handle `no-` well, to it sets `powerAssert` to false when usng `no-power-assert`. I think we should keep this.

sindresorhus: Yeah, the CLI part should be kept as-is. `--no` is the best way to handle negative flags.

jfmengels:
I addressed all comments, except https://github.com/avajs/ava/pull/1024#discussion_r76693217 (let me know if you want to me to do it), and renamed disablePowerAsssert to powerAssert throughout my changes in the codebase.

kentcdodds:
LGTM. Thanks for working on this @jfmengels!

jfmengels:
The tests work on my end on Node v6.4.0. Any hints to make those failing tests pass? :confused:

sindresorhus:
@jfmengels Tests pass on 6.4.0, but not on the just released 6.5.0. So it seems something changed there that broke some assumptions in the tests. Unrelated to your changes though, as it breaks in master too.

sindresorhus:
Looks great! Congrats on your first pull request as a core team member :D

leo-celebrate

kentcdodds:
Great work @jfmengels! Thanks for working on that!

PR #1025, opened 2016-08-30
Add `always` to TS definition

ivogabe:
The difficulty was that `test.always` should be disallowed, but `test.always.after` should work. So `test.always` should not have a call signature, but it should have members. I wrote some code for that, but that wasn't working. (We need tests for these definitions.) This PR fixes that.

// @SamVerschueren @sindresorhus


sotojuan:
Some of the failing tests are fixed by https://github.com/avajs/ava/commit/20ab39de046e527dec7b369f375d6c8e5fd4f5e1

novemberborn:
@ivogabe any chance you could rebase this? Changes look good otherwise.

ivogabe:
Yes, rebased

sindresorhus:
Thank you! :)

PR #1028, opened 2016-08-31
Fix tests in Node 6.5.0 by handling `fn` test name

### LINKED ISSUE #1027: Fix tests on Node.js 6.5.0
From https://github.com/avajs/ava/pull/1024#issuecomment-243609090. They passed on 6.4.0, but something changed.

sotojuan:
Attempts to fix #1027—feel free to close/discuss a better solution.

Node 6.5.0 uses a newer version of the v8 engine that is a bit smarter in figuring out the name of a function. Therefore, some tests were breaking because we presumed `test.name` to be a falsy value, and now it's `'fn'`. This PR handles `'fn'` the same way it handles a falsy function name value. While it fixes the breaking tests, it may not be the best choice in the long term.

### DIFF
--- a/lib/test-collection.js
+++ b/lib/test-collection.js
@@ -52,7 +52,7 @@ TestCollection.prototype.add = function (test) {
 		test.title = null;
 	}
 
-	if (!test.title || test.title === 'fn') {
+	if (!test.title) {
 		if (type === 'test') {
 			test.title = '[anonymous]';
 		} else {
--- a/lib/test-collection.js
+++ b/lib/test-collection.js
@@ -52,7 +52,7 @@ TestCollection.prototype.add = function (test) {
 		test.title = null;
 	}
 
-	if (!test.title) {
+	if (!test.title || test.title === 'fn') {
 		if (type === 'test') {
 			test.title = '[anonymous]';
 		} else {
--- a/package.json
+++ b/package.json
@@ -176,6 +176,7 @@
     "pify": "^2.3.0",
     "proxyquire": "^1.7.4",
     "rimraf": "^2.5.0",
+    "semver": "^5.3.0",
     "signal-exit": "^3.0.0",
     "sinon": "^1.17.2",
     "source-map-fixtures": "^2.1.0",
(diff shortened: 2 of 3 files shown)


lib/test-collection.js
```diff
@@ -52,7 +52,7 @@ TestCollection.prototype.add = function (test) {
 		test.title = null;
 	}
 
-	if (!test.title) {
+	if (!test.title || test.title === 'fn') {
```
sindresorhus: What if the user names their test or function `fn`?

sotojuan: Yeah, that's the issue. I'm trying to think of a way that we can handle both falsy values and the test's function names but that also won't affect anyone else :/

sotojuan:
@sindresorhus, what do you think is the best way to go forward with this?

kevva:
> If nothing else, we can just check for Node versions...

If the purpose is to fix the test, I would just check the Node version in the test and assert accordingly.

kevva:
I think this was added in V8 `5.1.39`. These links might be worth reading:

https://bugs.chromium.org/p/v8/issues/detail?id=3699
https://bugs.chromium.org/p/chromium/issues/detail?id=588803

sotojuan:
Yeah it depends on the purpose—I feel like it's just to fix the test right? So it'd make sense to rollback the code changes and maybe have a test for Node versions before and after the change?

kevva:
Yup, sounds good to me. Especially since this is a feature in ES2015 and not something we want to circumvent.

sotojuan:
All right, so it's kinda hacky and it introduces a new dependency, but it works. Let me know what you think!

fearphage:
Why not feature test it instead? 

```js
var supportsFunctionNames = (function testing() {}).name === 'testing';
```

This has the benefit of working in every environment consistently. Imagine if this feature were to be broken in v6.6 or removed in the future, the tests would always know how to do the right thing.

sotojuan:
@fearphage Great idea. I've updated the code to do this with the existing `noop` function in the file.

sotojuan:
I'm gonna go ahead and merge this since it's dev-only and the latest code changes have no negative or breaking effects.

PR #1035, opened 2016-09-08
Implemented t.finally, which runs after test is finished

rainydio:
So I've implemented `t.finally`. It's somewhat similar to test.afterEach.always, but it's easier to use and more flexible. For example:

```js
test("mock, subscribe, create files, etc", async t => {
  sinon.spy(console, "log");
  t.finally(() => console.log.restore());

  // might throw for whatever reason
  const connection = await connectSomewhere();
  t.finally(async () => await connection.disconnect());

  const server = mockAxios();
  server.onGet("/definitions", 200, []);
  server.onGet("/our/data", 200, [ { hello: "world" } ]);
  t.finally(() => server.restore());
  
  const data = await connection.query();
  t.is(data.length, 1); // throwing is totally expected

  storage.__Rewire__("fs", fakeFs);
  t.finally(() => storage.__ResetDependency__("fs"));
  
  const file = await store.write(data);
  sinon.aseert.called(fakeFs.write);
});
```

Finalizers are called in reverse order. They can fail test, transition it into async state, etc. Callback can be synchronous, return Promise, or Observable. Also `t.finally.cb` passes object into callback with a single function `end`.

I didn't updated base.d.ts, is it valid?

Closer to real world example when working with mobx

```js
test('using mobx', async t => {
  sinon.spy(store.local.preferences, "save");
  sinon.spy(store.local.preferences, "load");
  t.finally(() => store.local.preferences.save.restore());
  t.finally(() => store.local.preferences.load.restore());
  
  const user = store.users.byId(3);
  const dispose = user.autorun();
  t.finally(dispose)

  t.true(store.local.preferences.load.calledWith(3));
  await store.local.preferences.load.returnValues[0];
  t.is(user.preferences.color, "#ccc");

  mobx.runInAction(() => {
    user.preferences.color = "#fff";
  });

  t.true(store.local.preferences.save.calledWith(
    3, { color: "#fff" }
  ));
  await store.local.preferences.save.returnValues[0];
});
```

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -3,6 +3,7 @@ var inspect = require('util').inspect;
 var isGeneratorFn = require('is-generator-fn');
 var maxTimeout = require('max-timeout');
 var Promise = require('bluebird');
+var optionChain = require('option-chain');
 var fnName = require('fn-name');
 var co = require('co-with-promise');
 var observableToPromise = require('observable-to-promise');
@@ -46,6 +47,10 @@ function Test(title, fn, contextRef, report) {
 	// store the time point before test execution
 	// to calculate the total time spent in test
 	this._timeStart = null;
+
+	this._finalizing = false;
+	this._finalizers = [];
+	this._finalized = null;
 
 	// workaround for Babel giving anonymous functions a name
 	if (this.title === 'callee$0$0') {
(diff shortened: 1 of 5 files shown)


vdemedes:
How is this different from `t.afterEach.always`?

rainydio:
Well, yes:
* For specific test
* Initialization and disposal are nearby in code.
* When you need to dispose external resources which become available during test execution.
* No need to put data into context, and take it back. Just pass the callback.

Another example
```js
test("new repo fields", async t => {
  const user = await api.users.create({ username: "test" });
  t.finally(() => api.users.delete(user));

  const repo = await api.repos.create(user, { name: "my-repo" });
  t.finally(() => api.repos.remove(repo));
});

test.beforeEach(async t => {
  t.context.user = await api.users.create({ username: "test" });
  t.context.repo = await api.repos.create(user, { name: "my-repo" });
  // ^-- what if exception is raised here
})

test.afterEach.always(async t => {
  await api.repos.remove(t.context.repo); // then this fails
  await api.users.remove(t.context.user); // this one isn't freed
});
```

If you have dangling resources, those start affecting other tests. Like *Attempted to wrap load which is already wrapped* etc. So I find myself wrapping everything into `try { } finally { }`. And that's all fine, but resource allocation and disposing are away from each other, create nesting, and require variable outside of scope. So I start grouping them together and hitting the same problem.

This is an attempt to solve it.

vdemedes:
Not sure I want to see this in the core, as there's a solution for this problem already in there - `.afterEach.always`. This looks like an API sugar to me. Let's see what other people think.

---

I strongly advise you to check out contribution guidelines before submitting pull requests. We don't want to waste people's time by rejecting pull requests, that haven't been discussed before or just don't fit in. That's why it's always best to discuss your changes/ideas in the issue first, so we can all agree on the solution before doing any actual work.

rainydio:
No problem with rejecting that.

novemberborn:
This seems like a problem that `try/finally` already solves, variable scoping aside, or indeed something that can be solved with before/after hooks. The extra syntax seems unnecessary.

Props on the extensive test coverage though!

PR #1038, opened 2016-09-14
Add test run duration

fearphage:
## Changes

This does 2 new things:
- Send test run duration to reporters
- Show test duration in mini reporter (I added a utility to calculate this)

I added 2 properties to the `runStatus`:
- `testRunStart`
- `testRunElapsed`
## Duration Utility
- `hrtimeToDuration` - converts `process.hrtime()` to duration
- `httimeToDuration.duration` - converts milliseconds to duration
## Mini Reporter

In the mini reporter, I convert the `hrtime` back to a human-readable duration to show the end user. This could be added to any reporter, but I started with just this one.
## Notes

Currently I'm calculating the elapsed time in two places (`cli` and `watcher`). The best place to calculate this was in the logger (`Logger.prototype.finish`). However the lgoger seems to be logic-less and a complete pass-through so I didn't want to add functionality there.

### DIFF
--- a/lib/duration.js
+++ b/lib/duration.js
@@ -1,0 +1,39 @@
+'use strict';
+
+var second = 1000;
+var minute = 60 * second;
+var hour = 60 * minute;
+var day = 24 * hour;
+
+function round(value) {
+	return Number(Math.round(value * 1000) / 1000);
+}
+
+function duration(ms) {
+	if (ms >= day) {
+		return round(ms / day) + 'd';
+	}
+
+	if (ms >= hour) {
+		return round(ms / hour) + 'h';
+	}
+
+	if (ms >= minute) {
+		return round(ms / minute) + 'm';
+	}
+
+	if (ms >= second) {
+		return round(ms / second) + 's';
+	}
+
+	return round(ms) + 'ms';
+}
+
+function hrtimeToDuration(hrtime) {
+	return duration(((hrtime[0] * 1e6) + (hrtime[1] / 1e3)) / 1e3);
+}
+
+module.exports = exports = hrtimeToDuration;
+
+// exposed for testing purposes
+exports.duration = duration;
--- a/lib/duration.js
+++ b/lib/duration.js
@@ -29,11 +29,4 @@ function duration(ms) {
 	return round(ms) + 'ms';
 }
 
-function hrtimeToDuration(hrtime) {
-	return duration(((hrtime[0] * 1e6) + (hrtime[1] / 1e3)) / 1e3);
-}
-
-module.exports = exports = hrtimeToDuration;
-
-// exposed for testing purposes
-exports.duration = duration;
+module.exports = duration;
--- a/lib/watcher.js
+++ b/lib/watcher.js
@@ -61,6 +61,8 @@ function Watcher(logger, api, files, sources) {
 		this.busy = api.run(specificFiles || files, {
 			runOnlyExclusive: runOnlyExclusive
 		}).then(function (runStatus) {
+			runStatus.testRunElapsed = runStatus.testRunStart && process.hrtime(runStatus.testRunStart);
+
 			runStatus.previousFailCount = self.sumPreviousFailures(currentVector);
 			logger.finish(runStatus);
 
--- a/lib/watcher.js
+++ b/lib/watcher.js
@@ -61,7 +61,7 @@ function Watcher(logger, api, files, sources) {
 		this.busy = api.run(specificFiles || files, {
 			runOnlyExclusive: runOnlyExclusive
 		}).then(function (runStatus) {
-			runStatus.testRunElapsed = runStatus.testRunStart && process.hrtime(runStatus.testRunStart);
+			runStatus.testRunElapsed = runStatus.testRunStart && Date.now() - runStatus.testRunStart;
 
 			runStatus.previousFailCount = self.sumPreviousFailures(currentVector);
 			logger.finish(runStatus);
--- a/cli.js
+++ b/cli.js
@@ -157,7 +157,7 @@ logger.start();
 
 api.on('test-run', function (runStatus) {
 	reporter.api = runStatus;
-	runStatus.testRunStart = process.hrtime();
+	runStatus.testRunStart = Date.now();
 	runStatus.on('test', logger.test);
 	runStatus.on('error', logger.unhandledError);
 
@@ -184,7 +184,7 @@ if (cli.flags.watch) {
 } else {
 	api.run(files)
 		.then(function (runStatus) {
-			runStatus.testRunElapsed = runStatus.testRunStart && process.hrtime(runStatus.testRunStart);
+			runStatus.testRunElapsed = runStatus.testRunStart && Date.now() - runStatus.testRunStart;
 			logger.finish(runStatus);
 			logger.exit(runStatus.failCount > 0 || runStatus.rejectionCount > 0 || runStatus.exceptionCount > 0 ? 1 : 0);
 		})
--- a/cli.js
+++ b/cli.js
@@ -157,6 +157,7 @@ logger.start();
 
 api.on('test-run', function (runStatus) {
 	reporter.api = runStatus;
+	runStatus.testRunStart = process.hrtime();
 	runStatus.on('test', logger.test);
 	runStatus.on('error', logger.unhandledError);
 
@@ -183,6 +184,7 @@ if (cli.flags.watch) {
 } else {
 	api.run(files)
 		.then(function (runStatus) {
+			runStatus.testRunElapsed = runStatus.testRunStart && process.hrtime(runStatus.testRunStart);
 			logger.finish(runStatus);
 			logger.exit(runStatus.failCount > 0 || runStatus.rejectionCount > 0 || runStatus.exceptionCount > 0 ? 1 : 0);
 		})
--- a/lib/reporters/mini.js
+++ b/lib/reporters/mini.js
@@ -10,6 +10,7 @@ var cross = require('figures').cross;
 var repeating = require('repeating');
 var objectAssign = require('object-assign');
 var colors = require('../colors');
+var duration = require('../duration');
 
 chalk.enabled = true;
 Object.keys(colors).forEach(function (key) {
@@ -121,7 +122,7 @@ MiniReporter.prototype.unhandledError = function (err) {
 	}
 };
 
-MiniReporter.prototype.reportCounts = function (time) {
+MiniReporter.prototype.reportCounts = function (time, testRunElapsed) {
 	var lines = [
 		this.passCount > 0 ? '\n  ' + colors.pass(this.passCount, 'passed') : '',
 		this.knownFailureCount > 0 ? '\n  ' + colors.error(this.knownFailureCount, plur('known failure', this.knownFailureCount)) : '',
@@ -130,8 +131,14 @@ MiniReporter.prototype.reportCounts = function (time) {
 		this.todoCount > 0 ? '\n  ' + colors.todo(this.todoCount, 'todo') : ''
 	].filter(Boolean);
 
-	if (time && lines.length > 0) {
-		lines[0] += ' ' + time;
+	if (lines.length > 0) {
+		if (time) {
+			lines[0] += ' ' + time;
+		}
+
+		if (testRunElapsed) {
+			lines[0] += ' (' + duration(testRunElapsed) + ')';
+		}
 	}
 
 	return lines.join('');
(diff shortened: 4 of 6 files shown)


fearphage:
Restarting tests.

fearphage:
Test failures are due to a change in Node v6.5 as outlined in #1028.

sotojuan:
That PR has been merged! PS can I have some screenshots of this in action?

fearphage:
Picture and text format:

![screenshot](http://dl.dropbox.com/u/2400/scrot/selection_2016-09-17_162309.png)

```shell
➜  ava git:(add-test-run-duration) node cli.js --watch test/fixture/watcher/tap-in-conf/test.js test/fixture/watcher/tap-in-conf                                                                                                                               

  1 passed [16:21:22] (598.783ms)
  1 failed

  test › fails
  Test failed via t.fail()
    Test.fn (test.js:8:4)
    _combinedTickCallback (internal/process/next_tick.js:67:7)
    process._tickCallback (internal/process/next_tick.js:98:9)


───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

  1 passed [16:21:22] (235.709ms)
  1 failed

  test › fails
  Test failed via t.fail()
    Test.fn (test.js:8:4)
    _combinedTickCallback (internal/process/next_tick.js:67:7)
    process._tickCallback (internal/process/next_tick.js:98:9)


───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

  1 passed [16:21:32] (209.869ms)


^C%                                                                                                                                                                                                                                                            ➜  ava git:(add-test-run-duration) node cli.js test/fixture/pkg-conf/defaults                                                                                                                                                                                  

  1 passed (199.979ms)

➜  ava git:(add-test-run-duration) node cli.js test/fixture/long-running.js          

  2 passed (5.202s)
```

sindresorhus:
@fearphage Why do you need this? What problem does it solve? Why not just `$ time ava` when you want to know the total test run time?

fearphage:
> What problem does it solve?

It shows you the total duration of a test run. 

> Why not just `$ time ava` when you want to know the total test run time?

* How do you use that with `--watch` exactly? 
* `time` works horribly in windows (read: not at all).

Another problem with `time` is it includes the timings of the scaffolding and setup for `ava` itself as well as any tear down or waiting for file descriptors to close or whatever the case may be. I've tried to make this time measurement as close to the tests themselves running and finishing.

sindresorhus:
> It shows you the total duration of a test run.

That's the solution, not the problem. What will you do with the knowledge of the time it takes to run the test suite? Why do you need it on every run? I'm just trying to figure out why this is essential.

fearphage:
> That's the solution, not the problem.

The problem then is prior to this PR, I had no idea how long all the tests took to run together. There was information that is easily captured that was not being presented to me.

> What will you do with the knowledge of the time it takes to run the test suite?

I don't fully understand what you're asking. I will do the same thing I do with the knowledge of the time it takes to run individual tests. That's currently a feature in ava today right now. Do you have a problem with that functionality as well? What's the difference between individual test times and suite run times?

> Why do you need it on every run?

It's beneficial to be able to see at a glance the length of time a test suite takes to run. 

I have a few questions for you. Why would you _not_ want to see how long your tests take to run? What is it detracting from your code, tests, and/or life?

novemberborn:
> I will do the same thing I do with the knowledge of the time it takes to run individual tests. That's currently a feature in ava today right now.

It's been a summer without much AVA work for me so I might have forgotten, but IIRC we don't show individual test duration, do we?

I would be interested in surfacing which tests are exceedingly slow, in order to help users optimize their test suits. I'm not convinced outputting raw numbers is a sufficient solution, especially given the variability in host machine activity, process scheduling, and test concurrency.

sotojuan:
We show individual test time and the actual time all tests finished in verbose mode.

novemberborn:
I don't see individual test time in my verbose output.

sindresorhus:
> I don't fully understand what you're asking. I will do the same thing I do with the knowledge of the time it takes to run individual tests. That's currently a feature in ava today right now. Do you have a problem with that functionality as well? What's the difference between individual test times and suite run times?

That feature is there to expose slow tests. We only show it when the tests crosses a set threshold, which currently is 100ms. That's different from showing the total time. You can't do much about the total time unless you know the time of the individual tests.

> I have a few questions for you. Why would you not want to see how long your tests take to run? What is it detracting from your code, tests, and/or life?

Because everything we add to the output creates noise and distracts from the other information. We are very conservative about adding things. The default AVA output is intentionally minimal.

sindresorhus:
> I don't see individual test time in my verbose output.

Only for slow tests. >100ms.

lib/watcher.js
```diff
@@ -61,6 +61,8 @@ function Watcher(logger, api, files, sources) {
 		this.busy = api.run(specificFiles || files, {
 			runOnlyExclusive: runOnlyExclusive
 		}).then(function (runStatus) {
+			runStatus.testRunElapsed = runStatus.testRunStart && process.hrtime(runStatus.testRunStart);
```
sindresorhus: No point in using `process.hrtime()`. `Date.now()` is more than accurate enough for this purpose.

fearphage: Is this suggestion due to `process` not being supported in the browser or is there another reason? I'm just asking, because I'm curious.

My rationale for choosing `hrtime` over `Date` was providing the most amount of information to the reporters. One can always get milliseconds from nanoseconds, but the reverse is not true.

novemberborn:
> I feel like can do something with the deltas though. If before PR X, tests ran in 1 minute faster, that's a big deal.

That's true. @avajs/core ?

I don't think there's much value in tracking test duration when in watch mode, especially since the number of test files run can vary greatly.

lib/watcher.js
fearphage: Ping?

novemberborn: I think @sindresorhus is saying that `Date.now()` is accurate enough. I think that's true for this use case, no point optimizing test runs down to the nanosecond, and there is more cognitive overhead with `hrtime`.

That said I don't think there's much value in tracking test duration when in watch mode, especially since the number of test files run can vary greatly.

fearphage: Updated.

> That said I don't think there's much value in tracking test duration when in watch mode, especially since the number of test files run can vary greatly.

In that case, it's not very useful. When you're doing TDD or just updating a single test in general, it makes a lot more sense.

lib/duration.js
```diff
@@ -0,0 +1,32 @@
...
+	}
+
+	if (ms >= second) {
+		return round(ms / second) + 's';
+	}
+
+	return round(ms) + 'ms';
+}
+
+module.exports = duration;
```
sindresorhus: This could just be [`pretty-ms`](https://github.com/sindresorhus/pretty-ms) instead.

sindresorhus:
I'm open to adding the test time to the verbose and TAP reporter, but I'm hesitant to add it to the mini-reporter. The mini-reporter is meant to be as simple as possible, and I personally don't see much value in the test time. And I agree with @novemberborn, no point in having it in watch mode.

fearphage:
> no point in having it in watch mode

What about adding an option for it? It's inconvenient to have to leave watch mode just to see how long the tests are taking.

What's the point of printing the timestamp in watch mode? Is it so you can see that things are changing at a glance? If that is the sole purpose, could this replace the timestamp? It's more useful in terms of testing than showing the time.

sindresorhus:
> What about adding an option for it? It's inconvenient to have to leave watch mode just to see how long the tests are taking.

@novemberborn already commented why showing total test time in watch mode is not useful.

> What's the point of printing the timestamp in watch mode? Is it so you can see that things are changing at a glance? If that is the sole purpose, could this replace the timestamp? It's more useful in terms of testing than showing the time.

Yes, it's to see if something has changed, and no, total test time could not replace it for the same reason as above and the fact that the test time could be the same as previous and you would have to remember the previous test time to be sure it's different.

fearphage:
> @novemberborn already commented why showing total test time in watch mode is not useful.

Unless I'm missing it, @novemberborn only said s/he didn't find it useful. S/he did not elaborate on the why. Much like novermberborn, I have an opinion too. I believe it is useful. That's the reason I added it and the reason I expressed how inconvenient it would be to have to leave watch mode to figure out how long the tests were taking to run. I'm trying to see if there's a compromise that would allow it to remain in some form or another.

> the fact that the test time could be the same as previous  

This is a possibility. The [lack of precision in the timing](https://github.com/avajs/ava/pull/1038#discussion_r79607332) adds to that factor. Hitting the same millisecond is definitely possible. The same nanosecond is less probable.

> and you would have to remember the previous test time to be sure it's different.

How does this differ from printing the timestamp? By printing the time, you have to remember the time it displayed last time you saw it or know what time it is now including seconds. Neither one of them let you know the tests just ran without using your memory or external factors.

sindresorhus:
> Unless I'm missing it, @novemberborn only said s/he didn't find it useful.

No, he said why it's objectively useless:

> especially since the number of test files run can vary greatly.

-

> This is a possibility. The lack of precision in the timing adds to that factor. Hitting the same millisecond is definitely possible. The same nanosecond is less probable.

This has nothing to do with precision. Even if we used nanosecond precision we would only show a prettified output to the user. Let's say a test takes 1.4243242 seconds to run the first time and 1.45435353 the second time, we would only show 1.4 second for each.

fearphage:
> objectively useless

I have a use for this "objectively useless" thing.

> Even if we used nanosecond precision we would only show a prettified output to the user. Let's say a test takes 1.4243242 seconds to run the first time and 1.45435353 the second time, we would only show 1.4 second for each.

I had it setup for 3 decimal places so the output in your examples would be `1.424s` and `1.454s` respectively I believe.

sindresorhus:
> I have a use for this "objectively useless" thing.

Either you're not understanding how it's useless or you have a use-case you haven't being able to explain, but I feel this discussion is going nowhere. I would recommend focusing on what we're willing to do and not push this anymore unless you can actually come up with some convincing arguments not already exhausted ;)

> I had it setup for 3 decimal places so the output in your examples would be 1.424s and 1.454s respectively I believe.

`pretty-ms` converts it to `1.4`, which is what we want.

fearphage:
> pretty-ms converts it to 1.4, which is what we want.

> the fact that the test time could be the same as previous

> This has nothing to do with precision.

`pretty-ms` in its current configuration increases the likelihood of repeated output. You cited repetitious output as a problem. You also said this has nothing to do with precision. However as you increase the precision, repeated output becomes less likely. So it *directly* addresses one of the problems that you cited. I don't understand why you're saying it has nothing to do with this.

> Either you're not understanding how it's useless or you have a use-case you haven't being able to explain

Really?! Since I already [explained the use case](https://github.com/avajs/ava/pull/1038#issuecomment-248392914) a few weeks ago, I guess it's the former in this false dichotomy. Are you going to elaborate on the uselessness of my use case? Help me understand how useless it really is.

sindresorhus:
> pretty-ms in its current configuration increases the likelihood of repeated output. You cited repetitious output as a problem. You also said this has nothing to do with precision. However as you increase the precision, repeated output becomes less likely. So it directly addresses one of the problems that you cited. I don't understand why you're saying it has nothing to do with this.

Already said this, but I guess I can repeat myself again, we want it to be `1.4`, not `1.424s`. I do realize more precise output would make one of my arguments less relevant, but that's irrelevant as it's not something we want.

> Really?! Since I already explained the use case a few weeks ago, I guess it's the former in this false dichotomy. Are you going to elaborate on the uselessness of my use case? Help me understand how useless it really is.

You want the total test run time so you can use the deltas to see if it became faster or slower, but watch mode only runs tests that changed or had their source change, so the delta is not constant, making it useless for this use-case, as the time will vary greatly depending on what tests/sources changed.

novemberborn:
@fearphage given how watch mode functions I'm sceptical whether it'll actually deliver the information as reliably as you're expecting it to. I'm also wary it encourages micro-optimization of test runs.

Again, we're happy to have this in regular output since large performance regressions will surface that way, but at this point we don't want this for watch mode.

fearphage:
> I'm sceptical whether it'll actually deliver the information as reliably as you're expecting it to

Understandably so. I've been using it currently and it's delivering the expected results thus far.

Does ava support external reporters currently?

sindresorhus:
> Does ava support external reporters currently?

`$ ava --tap` and you can use any [TAP-reporter](https://github.com/avajs/ava#tap-reporter), but it does not apply during watch mode, neither will it ever.

Issue #567, opened 2016-02-19
Asynchronous errors do not fail tests


ORESoftware:
Hey all

Just wondering if this is expected output

```js
var test = require('ava').test;

test.beforeEach(t => {
   console.log('before each');
});

test(t => {
     setTimeout(function(){
       throw new Error('123');
     });
});

test(t => {
     setImmediate(function(){
       throw new Error('456');
     });
});

test(t => {
     process.nextTick(function(){
       throw new Error('789');
     });
});
```

output:

```
  3 passed
  3 exceptions

  1. Uncaught Exception
  Error: 789
    

  2. Uncaught Exception
  Error: 456
    

  3. Uncaught Exception
  Error: 123
```

shouldn't it be 0 passed, 3 exceptions?

novemberborn:
@ORESoftware some relevant fixes were made in `0.12`. Which version are you using?

sindresorhus:
I'm getting the same with AVA 0.12. Definitely a bug.

```
before each
before each
before each

  3 passed
  3 exceptions

  1. Uncaught Exception
  Error: 789
    test.js:21:14
    

  2. Uncaught Exception
  Error: 456
    Immediate._onImmediate (test.js:15:14)
    

  3. Uncaught Exception
  Error: 123
    [object Object]._onTimeout (test.js:9:14)
```

sindresorhus:
> var test = require('ava').test;

Btw, can be just `var test = require('ava');`

vdemedes:
I don't think we can catch errors on next tick, because at that time test function has completed. And we also can't know where this uncaught error came from. To sum up, I'm not sure we can do anything about it, without using domains.

novemberborn:
@vdemedes makes sense.

Presumably using `t.plan()` would prevent the test from passing (or indeed ending).

ORESoftware:
yeah just fyi

the same thing will happen if you replace

setTimeout/setImmediate/process.nextTick with any other standard async
functions such as

fs.exists('foo', function(err){
    throw new Error();
});

it's all about throwing an error inside an function called asynchronously.

the only real solution that I know of is the domain core module. but the
problem is ES6 promises do not support domains because domains are pending
deprecation. My strong belief is that without domains you won't be able to
solve this problem, so my only recommendation is to try to get Node.js TC
to backtrack on domains. Bluebird library supports domains, but not ES6
promises fyi.

Without domains you will just have to have an uncaughtException handler but
[...]

PR #1045, opened 2016-09-21
Document more async pitfalls

novemberborn:
General update of the async section and document takeaways from #567.


sotojuan:
LGTM.

Issue #882, opened 2016-05-26
The browser testing recipe should be more generic


lukechilds:
I don't mean in regards to the React example as mentioned here: https://github.com/avajs/ava/issues/773 but rather the way the environment is set up. Only `window`, `document` and `navigator` are exposed globally by the helper which means any other global variables available in a browser environment will only be accessible through `window`. This is fine for writing tests but lots of browser libs and their dependencies will expect everything to be global.

For example if one of the browser dependencies was doing something like: `if(domNode instanceof HTMLDivElement)` then `HTMLDivElement` would be undefined so you'd have to go in to the helper and add:

```js
global.HTMLDivElement = window.HTMLDivElement;
```

This could quickly get out of hand. I'm using a simple module ([lukechilds/node-browser-environment](https://github.com/lukechilds/node-browser-environment)) that by default adds everything in `window` to `global` so everything just works. If you want to limit to certain properties you can pass an array of required properties.

So the helper would just be:

```js
require('node-browser-environment')();
```

and you have everything on window available globally.

If you wanted to limit it to certain properties it'd be:

```js
require('node-browser-environment')(['window', 'document', 'navigator']);
```

Would you accept a PR to use this instead? Seems like it would be more frictionless.

lukechilds:
Thanks for that example. That just sticks everything on to global though, I still think it would be beneficial to be able to provide an array of properties to be used. That way if your browser lib has no deps you can safely require what you need and be less likely to experience issues with node dev dependencies that may get confused by browser globals (like in that issue). Plus I'd rather not create a gazillion global variables if I don't have to. 😆 

There wouldn't be any extra stuff to install compared to the current recipe. Currently it's:

```shell
$ npm install --save-dev jsdom
```

`test/helpers/setup-browser-env.js`

```js
global.document = require('jsdom').jsdom('<body></body>');
global.window = document.defaultView;
global.navigator = window.navigator;
```

And I'm proposing to change it to:

```shell
$ npm install --save-dev node-browser-environment
```

`test/helpers/setup-browser-env.js`

```js
require('node-browser-environment')();
```

or if you know exactly what you need:
```js
require('node-browser-environment')(['window', 'document', 'navigator']);
```

If you'd rather I just added [this](https://github.com/lukechilds/node-browser-environment/blob/master/src/index.js) to the recipe I'm happy to do this but I thought this seemed like a good [Lego block](https://github.com/sindresorhus/ama/issues/10#issuecomment-117766328)

[...]

PR #1054, opened 2016-09-26
Replace `jsdom` with `browser-env` in browser recipe

lukechilds:
Fixes https://github.com/avajs/ava/issues/882

### DIFF
--- a/docs/recipes/browser-testing.md
+++ b/docs/recipes/browser-testing.md
@@ -7,29 +7,35 @@ An example of this is React, at least if you want to use ReactDOM.render and sim
 
 This recipe works for any library that needs a mocked browser environment.
 
-## Install jsdom
+## Install browser-env
 
-Install [jsdom](https://github.com/tmpvar/jsdom).
+Install [browser-env](https://github.com/lukechilds/browser-env).
 
-> A JavaScript implementation of the WHATWG DOM and HTML standards, for use with node.js
+> Simulates a global browser environment using jsdom.
 
 ```
-$ npm install --save-dev jsdom
+$ npm install --save-dev browser-env
 ```
 
-## Setup jsdom
+## Setup browser-env
 
 Create a helper file and place it in the `test/helpers` folder. This ensures AVA does not treat it as a test.
 
 `test/helpers/setup-browser-env.js`:
 
 ```js
-global.document = require('jsdom').jsdom('<body></body>');
-global.window = document.defaultView;
-global.navigator = window.navigator;
+import browserEnv from 'browser-env';
+browserEnv();
 ```
 
-## Configure tests to use jsdom
+By default `browser-env` will add all global browser variables to the Node.js global scope creating a full browser environment. This should have good compatibility with most front-end libraries, however it's generally not a good idea to create lots of global variables if you don't need to. If you know exactly which browser variables you need you can pass an array of required properties.
+
+```js
+import browserEnv from 'browser-env';
+browserEnv(['window', 'document', 'navigator']);
+```
+
+## Configure tests to use browser-env
 
 Configure AVA to `require` the helper before every test file.
 
@@ -47,7 +53,7 @@ Configure AVA to `require` the helper before every test file.
 
 ## Enjoy!
 
-Write your tests and enjoy a mocked window object.
+Write your tests and enjoy a mocked browser environment.
 
 `test/my.react.test.js`:
 
--- a/docs/recipes/browser-testing.md
+++ b/docs/recipes/browser-testing.md
@@ -55,6 +55,19 @@ Configure AVA to `require` the helper before every test file.
 
 Write your tests and enjoy a mocked browser environment.
 
+`test/my.dom.test.js`:
+
+```js
+import test from 'ava';
+
+test('Insert to DOM', t => {
+  const div = document.createElement('div');
+  document.body.appendChild(div);
+
+  t.is(document.querySelector('div'), div);
+});
+```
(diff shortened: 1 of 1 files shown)


docs/recipes/browser-testing.md
```diff
@@ -7,29 +7,33 @@ An example of this is React, at least if you want to use ReactDOM.render and sim
...
 
-Install [jsdom](https://github.com/tmpvar/jsdom).
+Install [node-browser-environment](https://github.com/lukechilds/node-browser-environment).
 
-> A JavaScript implementation of the WHATWG DOM and HTML standards, for use with node.js
+> Simulates a global browser environment using jsdom.
 
 ```
-$ npm install --save-dev jsdom
+$ npm install --save-dev node-browser-environment
```
sindresorhus: That's a very verbose module name. Why not simply `browser-env`? Node is already implied.

docs/recipes/browser-testing.md
```diff
@@ -7,29 +7,33 @@ An example of this is React, at least if you want to use ReactDOM.render and sim
...
 
 Create a helper file and place it in the `test/helpers` folder. This ensures AVA does not treat it as a test.
 
 `test/helpers/setup-browser-env.js`:
 
 ```js
-global.document = require('jsdom').jsdom('<body></body>');
-global.window = document.defaultView;
-global.navigator = window.navigator;
+require('node-browser-environment')();
```
sindresorhus: Should use `import`, not `require`.

docs/recipes/browser-testing.md
```diff
@@ -7,29 +7,33 @@ An example of this is React, at least if you want to use ReactDOM.render and sim
...
 ```
 
-## Configure tests to use jsdom
+By default `node-browser-environment` will add all global browser variables to node's global scope creating a full browser environment. This should have good compatibility with most front end libraries, however it's generally not a good idea to create loads of global variables if you don't need to. If you know exactly which browser variables you need you can pass an array of required properties.
+
+```js
+require('node-browser-environment')(['window', 'document', 'navigator']);
```
sindresorhus: Use `import`

docs/recipes/browser-testing.md
```diff
@@ -7,29 +7,33 @@ An example of this is React, at least if you want to use ReactDOM.render and sim
...
 
 ```js
-global.document = require('jsdom').jsdom('<body></body>');
-global.window = document.defaultView;
-global.navigator = window.navigator;
+require('node-browser-environment')();
 ```
 
-## Configure tests to use jsdom
+By default `node-browser-environment` will add all global browser variables to node's global scope creating a full browser environment. This should have good compatibility with most front end libraries, however it's generally not a good idea to create loads of global variables if you don't need to. If you know exactly which browser variables you need you can pass an array of required properties.
```
sindresorhus: node => Node.js

front end => front-end

loads of => lots of

sindresorhus:
@lukechilds Are you committed to keeping this module up to date with `jsdom`?

lukechilds:
@sindresorhus Yep, I've pinned the version number for `jsdom` to stop any `jsdom` breaking changes breaking AVA tests and I've got Greenkeeper set up to notify me of `jsdom` updates. If you look through the commit history you'll see I normally update this within 24 hours.

I've got no intention of abandoning this module but I'm happy to add you or anyone else on the AVA team as a maintainer if you want.

Working on your requested changes now 👍

lukechilds:
@sindresorhus Done requested changes 🙂

docs/recipes/browser-testing.md
sindresorhus: It's pretty easy. You just bump major on the existing one and deprecate with a message saying it's been renamed. We bump major first so the deprecation notice doesn't annoy existing users until they bump a version. I would personally do it as that module is going to be used a lot and `browser-environment` would look nicer in the import statement, but your choice ;)

sindresorhus: Actually, https://github.com/CamShaft/browser-env is deprecated, so you could just request the module name.

https://docs.npmjs.com/misc/disputes

lukechilds: Good spot on the deprecation, yeah I'll rename it. If you wanna hold off on merging this for now I'll update you when it's sorted 👍

Quick Q: Why bump major first? Wouldn't it be better to notify existing users? If they've installed with the default `^` range they'll be expecting to get SemVer minor/patch updates automatically.

sindresorhus: @lukechilds Not very relevant in this situation, as it's a top-level module, so maybe not necessary, but when your module is deep in a dependency tree, people will start opening issues on every module that depends on it somewhere in the dependency tree about updating, which is very annoying for the maintainers and a lot of wasted time. Users think warnings are critical.

lukechilds: @sindresorhus `node-browser-environment` is dead, long live `browser-env`! 🎉

lukechilds:
I'm thinking something along the lines of:

`test/my.dom.test.js`:

```js
import test from 'ava';

test('Insert to DOM', t => {
  const div = document.createElement('div');
  document.body.appendChild(div);

  t.is(document.querySelector('div'), div);
});
```

It could be as well as the current React test or instead of. What do you think?

sindresorhus:
Good idea. Should be in addition to the React one.

lukechilds:
@sindresorhus Done 👍

sindresorhus:
Thanks :)

PR #1061, opened 2016-10-01
Clean up API

vdemedes:
Starting a series of "cleaning up" PRs to make AVA's codebase cleaner and friendlier to everyone. These PRs don't change any functionality and don't modify tests on purpose, those are just code improvements.

This PR removes lots of nesting, nested functions (`tryRun()`, `run()`, `runner()`) and confusing code. Also, it brings little improvements in the naming, options objects and makes functions shorter by extracting their parts into individual functions.

### DIFF
--- a/api.js
+++ b/api.js
@@ -2,21 +2,47 @@
 var EventEmitter = require('events').EventEmitter;
 var path = require('path');
 var util = require('util');
-var Promise = require('bluebird');
-var objectAssign = require('object-assign');
 var commonPathPrefix = require('common-path-prefix');
-var resolveCwd = require('resolve-cwd');
 var uniqueTempDir = require('unique-temp-dir');
 var findCacheDir = require('find-cache-dir');
+var objectAssign = require('object-assign');
+var resolveCwd = require('resolve-cwd');
 var debounce = require('lodash.debounce');
-var ms = require('ms');
 var AvaFiles = require('ava-files');
 var autoBind = require('auto-bind');
+var Promise = require('bluebird');
 var getPort = require('get-port');
-var AvaError = require('./lib/ava-error');
-var fork = require('./lib/fork');
+var arrify = require('arrify');
+var ms = require('ms');
 var CachingPrecompiler = require('./lib/caching-precompiler');
 var RunStatus = require('./lib/run-status');
+var AvaError = require('./lib/ava-error');
+var fork = require('./lib/fork');
+
+function resolveModules(modules) {
+	return arrify(modules).map(function (name) {
+		var modulePath = resolveCwd(name);
+		if (modulePath === null) {
+			throw new Error('Could not resolve required module \'' + name + '\'');
+		}
+
+		return modulePath;
+	});
+}
+
+function getBlankResults() {
+	return {
+		stats: {
+			knownFailureCount: 0,
+			testCount: 0,
+			passCount: 0,
+			skipCount: 0,
+			todoCount: 0,
+			failCount: 0
+		},
+		tests: []
+	};
+}
 
 function Api(options) {
 	if (!(this instanceof Api)) {
@@ -24,6 +50,7 @@ function Api(options) {
 	}
 
 	EventEmitter.call(this);
+	autoBind(this);
 
 	this.options = objectAssign({
 		cwd: process.cwd(),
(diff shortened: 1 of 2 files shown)


vdemedes:
PR is ready for review!

api.js
```diff
@@ -56,309 +60,310 @@ Api.prototype._runFile = function (file, runStatus, execArgv) {
...
 	return emitter;
 };
 
+Api.prototype.run = function (files, options) {
+	var self = this;
+
+	var findOptions = {
+		cwd: this.options.resolveTestsFrom,
+		files: files
+	};
```
sindresorhus: I don't really see how this improves anything though. It was already clear what the options are for.

api.js
```diff
@@ -56,309 +60,310 @@ Api.prototype._runFile = function (file, runStatus, execArgv) {
...
 };
 
-Api.prototype._run = function (files, _options) {
-	var self = this;
+Api.prototype._run = function (files, options) {
+	if (!options) {
+		options = {};
+	}
+
+	var prefixTitles = this.options.explicitTitles || files.length > 1;
```
sindresorhus: I don't see the point of extracting this either. It has the same title as the option name, so I don't see how this makes it any clearer.

api.js
```diff
@@ -56,309 +60,310 @@ Api.prototype._runFile = function (file, runStatus, execArgv) {
...
+	var isPowerAssertEnabled = this.options.powerAssert !== false;
+	this.precompiler = new CachingPrecompiler(cacheDir, this.options.babelConfig, isPowerAssertEnabled);
 };
 
-Api.prototype._run = function (files, _options) {
-	var self = this;
+Api.prototype._run = function (files, options) {
+	if (!options) {
+		options = {};
+	}
```
sindresorhus:
```js
options = options || {};
```

lib/fork.js
```diff
@@ -153,7 +153,14 @@ module.exports = function (file, opts, execArgv) {
 		return promise;
 	};
 
+	var isExitInitiated = false;
```
sindresorhus: What change made this necessary? Maybe worth using [`onetime`](https://github.com/sindresorhus/onetime) here?

api.js
```diff
@@ -56,309 +60,310 @@ Api.prototype._runFile = function (file, runStatus, execArgv) {
...
+				self._handleError(runStatus, err);
+			}
+
+			runStatus.processResults(results);
+
+			return runStatus;
 		});
 };
 
 function getBlankResults() {
```
sindresorhus: Can you move this to the the top? Better to have it defined before use.

api.js
```diff
@@ -56,309 +60,310 @@ Api.prototype._runFile = function (file, runStatus, execArgv) {
...
+	var prefixTitles = this.options.explicitTitles || files.length > 1;
+	var basePath = path.relative('.', commonPathPrefix(files)) + path.sep;
+
 	var runStatus = new RunStatus({
-		prefixTitles: this.options.explicitTitles || files.length > 1,
-		runOnlyExclusive: _options && _options.runOnlyExclusive,
-		base: path.relative('.', commonPathPrefix(files)) + path.sep
+		runOnlyExclusive: options.runOnlyExclusive,
+		prefixTitles: prefixTitles,
+		base: basePath
```
sindresorhus: If `base` is unclear, I think we instead should rename it in `RunStatus`.

sindresorhus:
Would be good if @jamestalmage could review, as he's the one that wrote most of the code in `api.js`.

api.js
vdemedes: The point of this is not the name (I named it `findOptions`, because `options` is already defined there), but extracting options to reduce nesting. I can move it back, not insisting on this one.

lib/fork.js
vdemedes: Unfortunately, I forgot... I remember it had something to do with this https://github.com/avajs/ava/blob/ea25e802bce67b195d5f0fc331291bb02f16b866/api.js#L305. Sure, I'll use `onetime`.

api.js
vdemedes: Yes, it'd rename it to `basePath`, but in a separate PR dedicated to cleaning up `RunStatus`.

lib/fork.js
vdemedes: I removed this completely and tests ran successfully, I guess there's no need for it.

vdemedes:
@sindresorhus addressed all your feedback, thanks!

sotojuan:
LGTM as well :-) Good work!

sindresorhus:
Alright. Landed. Good stuff @vdemedes :)

vdemedes:
Thank guys!

PR #1063, opened 2016-10-04
Switch to lodash.isEqual for deep quality check

### LINKED ISSUE #1055: Consider switching deep equal module
We currently use: https://github.com/sotojuan/not-so-shallow

I know this is something like the third time, but someone pointed out our current module [has a lot of issues](https://github.com/sotojuan/not-so-shallow/issues), and I'm personally not interested or have time to fix those, nor maintain a deep equal module.

I'd like to look into whether we could use [`lodash.isEqual`](https://lodash.com/docs/4.16.2#isEqual).

// @jdalton

vdemedes:
Fixes #1055.

- Switch to `lodash.isEqual()` for `t.deepEqual()`.
- Move tests from `not-so-shallow` to `test/assert.js` to ensure regression-free migration (no duplicates).


package.json
```diff
@@ -131,6 +131,7 @@
     "last-line-stream": "^1.0.0",
     "lodash.debounce": "^4.0.3",
     "lodash.difference": "^4.3.0",
+    "lodash.isequal": "^4.4.0",
     "loud-rejection": "^1.2.0",
     "matcher": "^0.1.1",
     "max-timeout": "^1.0.0",
```
sindresorhus: You also need to remove the `no-so-shallow` dependency here.

vdemedes: Oh, I thought I removed it.

vdemedes: Done.

sotojuan:
LGTM.

PR #1068, opened 2016-10-07
Reduce transpilation on Node.js >=4

sindresorhus:
Better initial startup performance and improved stack traces. I mostly did this for better stack traces. Related to https://github.com/avajs/ava/issues/148. Long-term we'll want to use [`babel-preset-env`](https://github.com/babel/babel-preset-env), but it's too immature at the moment. I didn't include a specific preset for Node.js 6 as it didn't improve performance and will be solved with the eventual use of `babel-preset-env`.

You can see that the initial cold run improved by more than a second after this change.

### Without cache

```
~/dev/got master
❯ time ava

  85 passed

ava --no-local  8.98s user 0.86s system 108% cpu 9.060 total
```

### With cache

```
~/dev/got master
❯ time ava

  85 passed

ava --no-local  5.06s user 0.62s system 101% cpu 5.621 total
```

### Without cache

```
~/dev/got master
❯ time ava               

  85 passed

ava --no-local  7.29s user 0.81s system 104% cpu 7.783 total
```

### With cache

```
~/dev/got master
❯ time ava

  85 passed

ava --no-local  5.18s user 0.63s system 106% cpu 5.476 total
```

-

### Before

```
_callee3$ (error.js:40:5)
    tryCatch (/Users/sindresorhus/dev/ava/node_modules/regenerator-runtime/runtime.js:62:40)
    GeneratorFunctionPrototype.invoke [as _invoke] (/Users/sindresorhus/dev/ava/node_modules/regenerator-runtime/runtime.js:336:22)
    GeneratorFunctionPrototype.prototype.(anonymous function) [as throw] (/Users/sindresorhus/dev/ava/node_modules/regenerator-runtime/runtime.js:95:21)
```

### After

```
Test.<anonymous> (error.js:40:5)
```

-

Didn't include any test as the code is straightforward and manually tested to work, and we have other tests already confirming it's working.


sindresorhus:
@novemberborn Does the caching take into account `process.version`?

novemberborn:
@sindresorhus 

> Does the caching take into account `process.version`?

No. Would have to [add that in `caching-precompiler.js`](https://github.com/avajs/ava/blob/16b7e538d2061dbab363087be4babd42e4549aec/lib/caching-precompiler.js#L79).

sindresorhus:
@novemberborn Done. I went with only salting per major Node.js version. Don't want the cache to expire just because the user did a security upgrade.

sindresorhus:
Thanks for reviewing :)

Issue #1069, opened 2016-10-07
Remove --require CLI flag


novemberborn:
Per discussion in #1048 we'll be removing the `--require` CLI flag. You'll still be able to configure this via the `package.json` config.

Issue #1048, opened 2016-09-23
Thoughts on CLI flags


novemberborn:
Came across [this comment](https://github.com/avajs/ava/issues/782#issuecomment-217615776) from @jamestalmage:

> My feeling on options via `cli.js`: We should only add those where users would conceivably want to temporarily override the contents of their `package.json` config. Allowing different reporters on CI servers, running in `--serial` mode occasionally to debug, these are good examples.

I quite like this! Maybe it should be our "official" stance.

I think that'd mean `--require` has to go. On the other hand the `--source` flag may have to stay (as opposed to #937). We shouldn't have `--extension` either (#631).

Thoughts?

sindresorhus:
@novemberborn I agree, but what other flags would we remove?

> On the other hand the --source flag may have to stay

Why?

novemberborn:
> other flags would we remove?

Just `require` I think. Ideas for more?

>> On the other hand the --source flag may have to stay

> Why?

To restrict watch mode on the fly to a more specific pattern.

jamestalmage:
> To restrict watch mode on the fly to a more specific pattern.

That seems unlikely to be truly useful. Why would you want to avoid rerunning when certain sources change, but not others?

vdemedes:
I think `--require` and `--source` are the only ones that can safely go away. All the other ones are quite useful, don't think there's more room for removal.

novemberborn:
Great! I've updated the relevant issues and opened #1069 to remove the `--require` flag.

Issue #937, opened 2016-06-26
Remove --sources CLI flag


novemberborn:
With #925 we've changed how test files are resolved. If file patterns are passed as CLI arguments they're resolved relative to the current working directory. Otherwise file patterns in `package.json`, or indeed default patterns, are resolved relative to the the directory that contains `package.json`.

Our management of `source` patterns is similarly affected. This may or may not be the reason why watch mode doesn't work correctly when AVA is called outside of the project root (see #936). However we should reconsider how `sources` works.

From @jamestalmage in https://github.com/avajs/ava/pull/863#issuecomment-221465954:

> If they are `cd`ed into a child directory and use the `--sources` flag, that should probably be resolved from `cwd`, right? But if they don't specify `--sources`, then we probably want to use whatever is in `package.json`, and resolve from that directory? That's getting complicated.
> 
> Should we eliminate the `--sources` flag? I have never used it, and I don't see why you really need it. Doesn't this qualify as a static config item that should live in `package.json`?

I'm 👍  on removing the `sources` flag. Patterns must then be specified in the `package.json`, which hopefully makes them easier to manage. Both for AVA users and maintainers.

sindresorhus:
Let's do it.

novemberborn:
Somewhat blocked by #1048.

sotojuan:
Are we still doing this? Making a list of things to take care of this weekend!

novemberborn:
@sotojuan yes.

LasaleFamine:
Hi guys! 
Can I take this one? Is there something more I should know?

novemberborn:
@LasaleFamine go for it! 👍

LasaleFamine:
Just to be sure, I can see a `--source` flag, not a `--sources`. Is right?

Another thing I can see is that the `cli.flags.source` is used within the `Watcher` constructor: 

```javascript
// lib/cli.js
if (cli.flags.watch) {
		try {
			const watcher = new Watcher(logger, api, files, arrify(cli.flags.source));
			watcher.observeStdin(process.stdin);
		} catch (err) {

...
```

We need to get that `source` prop from the `pacakge.json`?

novemberborn:
> Just to be sure, I can see a --source flag, not a --sources. Is right?

Yes.

> We need to get that source prop from the pacakge.json?

I think you can get it from `conf.source`. See #1046 for context.

PR #1070, opened 2016-10-07
Remove --require CLI option (completes #1069)

jacobkahn:
This completes #1069 as part of #1048 - removes `--require`/`-r` from the CLI interface, corrects documentation where relevant, modifies CLI tests accordingly.

Not sure if it makes sense to add a deprecation notice or any clarifications to docs - these changes remove mentions of the flag and defer to `package.json` configuration guidelines.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -60,7 +60,6 @@ var cli = meow([
 	'  --init             Add AVA to your project',
 	'  --fail-fast        Stop after first test failure',
 	'  --serial, -s       Run tests serially',
-	'  --require, -r      Module to preload (Can be repeated)',
 	'  --tap, -t          Generate TAP output',
 	'  --verbose, -v      Enable verbose output',
 	'  --no-cache         Disable the transpiler cache',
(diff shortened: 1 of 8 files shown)


vdemedes:
I'm sure many people use `--require babel-register` or similar, I think deprecation message pointing to configuration in package.json is a must have.

cli.js
```diff
@@ -129,7 +126,6 @@ if (
 var api = new Api({
 	failFast: cli.flags.failFast,
 	serial: cli.flags.serial,
-	require: arrify(cli.flags.require),
```
novemberborn: Before, `cli.flags.require` would default to the config from `package.json`, through [this line](https://github.com/jacobkahn/ava/blob/f94b58df23202749691e3e82305486efd6a9854d/cli.js#L98). `Api` still needs the `require` option, so this should now be `require: arrify(conf.require),`.

novemberborn:
And wow @jacobkahn that was quick!

jacobkahn:
Re-added that arg into API - thanks for catching that. Right now the deprecation message mentions the docs, then exits. Do we think this behavior is alright?

novemberborn:
🎊🎉 Thanks @jacobkahn!

If you're up for another challenge, #937 is fairly similar 😄

Issue #32, opened 2015-09-01
Should `process.cwd()` be changed to the directory of the currently running test?


sindresorhus:
Currently the cwd is the directory you execute `ava` in, but `ava` can be executed from any upper level in the project and will search down for test files, so the cwd is pretty arbitrary and useless. I want to make the cwd actually useful.

There are two options:

1.  Change cwd to root of the project. (This is what gulp/grunt does)
2.  Change cwd to the directory of the currently running test file. (This is what I personally prefer)

The reason I prefer the latter is that we *run* the test files hence IMHO the cwd is where we run it. It's like we would do: `cd tests && node test.js`.

-

Currently you have to do this to be able to load a read a test relative file:

```js
var path = require('path');
var fs = require('fs');
fs.readFileSync(path.join(__dirname, 'foo.json'));
```

If we changed `process.cwd()` to the currently running test directory we could instead just do:

```js
var fs = require('fs');
fs.readFileSync('foo.json');
```

-

Thoughts? Anything I'm missing?

sindresorhus:
@Qix- @kevva @vdemedes @mdibaiee @mattdesl @arthurvr

Qix-:
It sounds good, but it might make migrating from Mocha a bit of a pain. New projects will benefit from it; ones coming from Mocha won't.

sindresorhus:
@Qix- Why? It will make no difference if you've already hard-coded your paths with `__dirname`.

Qix-:
True. Those that haven't should be anyway.

vdemedes:
As for me, I think overriding core libraries/methods is always a bad idea. You never know what and how many things rely on that method.

Also, this is not that big of an issue to include test directory in paths.

sindresorhus:
Not going to override anything. This about using [`process.chdir()`](https://nodejs.org/api/process.html#process_process_chdir_directory) to set the cwd.

vdemedes:
Oh damn, what a shame :D Then I am all for it! :+1: Also tired of `join(__dirname, 'fixtures', 'shit')`

sindresorhus:
Currently the cwd is the directory you execute `ava` in, but `ava` can be executed from any upper level in the project and will search down for test files, so the cwd is pretty arbitrary and useless. I want to make the cwd actually useful.

There are two options:

1.  Change cwd to root of the project. (This is what gulp/grunt does)
2.  Change cwd to the directory of the currently running test file. (This is what I personally prefer)

[...]

Issue #1041, opened 2016-09-21
TypeError: Path must be a string. Received null


con-ssc:
## Description

ava nor ava --init nor ava --version is working.
The project has no config at all and no node_modules.
ava is installed globaly

### Error Message & Stack Trace

```
d:\repositories\git\crmintern\Testing_Ground>ava --version                                                                                                                             
path.js:7                                                                                                                                                                              
    throw new TypeError('Path must be a string. Received ' + inspect(path));                                                                                                           
    ^                                                                                                                                                                                  
                                                                                                                                                                                       
TypeError: Path must be a string. Received null                                                                                                                                        
    at assertPath (path.js:7:11)                                                                                                                                                       
    at Object.dirname (path.js:697:5)                                                                                                                                                  
    at Object.<anonymous> (C:\Users\stefan.schneider\AppData\Roaming\npm\node_modules\ava\cli.js:43:19)                                                                                
    at Module._compile (module.js:556:32)                                                                                                                                              
    at Object.Module._extensions..js (module.js:565:10)                                                                                                                                
    at Module.load (module.js:473:32)                                                                                                                                                  
    at tryModuleLoad (module.js:432:12)                                                                                                                                                
    at Function.Module._load (module.js:424:3)                                                                                                                                         
    at Module.runMain (module.js:590:10)                                                                                                                                               
    at run (bootstrap_node.js:394:7)                                                                                                                                                   
    at startup (bootstrap_node.js:149:9)                                                                                                                                               
    at bootstrap_node.js:509:3  
```

### Command-Line Arguments

Copy your npm build scripts or the `ava` command used:

```
ava --version
```

## Environment

Tell us which operating system you are using, as well as which versions of Node.js, npm, and AVA. Run the following to get it quickly:

```
node: v6.6.0
ava: v0.16.0
npm: 3.10.8
VSCode Version: Code 1.5.2 (66f37fd2a99eb9d628dd374d81d78835b410c39b, 2016-09-12T13:30:03.456Z)
OS Version: Windows_NT ia32 10.0.10586
```

novemberborn:
That's [this line](https://github.com/avajs/ava/blob/v0.16.0/cli.js#L43):

```js
var pkgDir = path.dirname(pkgConf.filepath(conf));
```

That changed in #925. Looks like AVA now won't start without a `package.json` present. We use that file (or rather the directory it's in) to figure out various paths. I suggest that without it we just use the current working directory.

con-ssc:
[...]

Issue #611, opened 2016-03-07
Remove defaults for private code


novemberborn:
AVA exposes two public interfaces: the CLI and the programmatic API. All other code is private. Currently a lot of the private code has defaults, e.g. `Runner` [can be instantiated without `options`](https://github.com/sindresorhus/ava/blob/eaca9f082fc36add3c905afad3eeca7f0f10cfcb/lib/runner.js#L39).

This tends to happen in tests and causes confusion where code has to be [extra defensive](https://github.com/sindresorhus/ava/blob/eaca9f082fc36add3c905afad3eeca7f0f10cfcb/lib/runner.js#L74) against scenarios that wouldn't occur if defaults were set at the public interface level.

Allowing defaults throughout the code makes it hard to reason about expected values. I'm proposing we  remove defaults from internal code and make our tests more explicit.

Thoughts?

vdemedes:
I think we shouldn't do it, because those libraries/components should be stable on their own. They should not depend on other code passing these arguments correctly.

> Allowing defaults throughout the code makes it hard to reason about expected values.

We can improve our checking of value types, using functions like `Array.isArray()`, so that it's more clear what we're expecting.

novemberborn:
> I think we shouldn't do it, because those libraries/components should be stable on their own

Sure, they'd just require valid options to be passed.

novemberborn:
@avajs/core any thoughts on this?

sindresorhus:
👍 Sounds good.

novemberborn:
We'll just do this when we're touching code. No need to keep the issue open.

PR #1074, opened 2016-10-13
Change the current working directory of tests to be the same directory as package.json

alathon:
### Description

Changes fork to spawn the process based on the (existing) value of ```options.resolveTestsFrom```, which is the package directory by default. Fixes, amongst other things, problems with babel-root-import, which does not work with the current behavior (as it expects process.cwd() to be the package.json folder).

Related to #32 and several other issues that stem from the same problem.

I'm not 100% sure that I'm meant to be piggybacking options.resolveTestsFrom for this purpose - but it seems like the appropriate option. There were no tests of resolveTestsFrom at all, so I'm not 100% sure what the option is meant to be doing, if not this.


sindresorhus:
You also need to update this paragraph in the readme:

> Test files are run from their current directory, so process.cwd() is always the same as __dirname. You can just use relative paths instead of doing path.join(__dirname, 'relative/path').

alathon:
@sindresorhus I've updated the readme to reflect the change. However, now I'm slightly in doubt as to whether resolveTestsFrom should be implemented in cli.js as it is.

I'm confused by why api.js:140 (https://github.com/alathon/ava/blob/use-resolveTestsFrom-in-fork/cli.js#L140) is implemented like that. Can you elaborate on why resolveTestsFrom gets set to process.cwd() if there is non-flag input, and why resolveTestsFrom wasn't made a CLI flag to begin with?

readme.md
```diff
@@ -247,7 +247,7 @@ If you're unable to use promises or observables, you may enable "callback mode"
 
 You must define all tests synchronously. They can't be defined inside `setTimeout`, `setImmediate`, etc.
 
-Test files are run from their current directory, so [`process.cwd()`](https://nodejs.org/api/process.html#process_process_cwd) is always the same as [`__dirname`](https://nodejs.org/api/globals.html#globals_dirname). You can just use relative paths instead of doing `path.join(__dirname, 'relative/path')`.
+The process that runs a test file has a [`process.cwd()`](https://nodejs.org/api/process.html#process_process_cwd) that is set to the directory of the package.json file AVA finds. In most projects this will be the root folder of the project. To override this behavior and run tests from a *different* folder, you can set the `resolveTestsFrom` option to a relative path, which will be used instead of the package file directory.
```
sindresorhus: Drop the following: `To override this behavior and run tests from a *different* folder, you can set the `resolveTestsFrom` option to a relative path, which will be used instead of the package file directory.`

It doesn't make sense here as it's for the API, not the consumer. I don't think we want it documented at all as the API is not stable.

alathon: Good point. Removed that part again. Is it relevant to discuss my question above about resolveTestsFrom in a separate issue, or is Gitter more appropriate for that?

alathon:
@novemberborn mentions that resolveTestsFrom is used to resolve command-line paths to files. That means I'm misusing it here, and ironically you'd end up with the same problematic behavior, if you ran ava from a sub-folder.

I've gone ahead and added a package option called `pkgDir` instead, which does the same thing, and changed the tests accordingly.

test/api.js
```diff
@@ -307,12 +307,23 @@ function generateTests(prefix, apiCreator) {
 			});
 	});
 
-	test(prefix + 'change process.cwd() to a test\'s directory', function (t) {
+	test(prefix + 'run from package.json folder by default', function (t) {
```
novemberborn: This test doesn't do what you think it does.

The `apiCreator` executor doesn't actually set the `pkgDir` option when it creates the `api`. This means that the child process is forked with an `undefined` `cwd` option, so it ends up using `process.cwd()`. Because we run tests from the project root that happens to be the expected directory, but that's just luck.

Instead let's find the correct `pkgDir` value in the test itself, and provide it when creating the API [on this line](https://github.com/alathon/ava/blob/2a65c1295588ebdf36dcc833484c34f3ed956add/test/api.js#L20) and [this one](https://github.com/alathon/ava/blob/2a65c1295588ebdf36dcc833484c34f3ed956add/test/api.js#L65).

The next test is good (since it tests a non-default value). Just make sure not to accidentally break it with the aforementioned changes.

readme.md
```diff
@@ -247,7 +247,7 @@ If you're unable to use promises or observables, you may enable "callback mode"
 
 You must define all tests synchronously. They can't be defined inside `setTimeout`, `setImmediate`, etc.
 
-Test files are run from their current directory, so [`process.cwd()`](https://nodejs.org/api/process.html#process_process_cwd) is always the same as [`__dirname`](https://nodejs.org/api/globals.html#globals_dirname). You can just use relative paths instead of doing `path.join(__dirname, 'relative/path')`.
+The process that runs a test file has a [`process.cwd()`](https://nodejs.org/api/process.html#process_process_cwd) that is set to the directory of the package.json file AVA finds. In most projects this will be the root folder of the project.
```
novemberborn: How about simply:

> AVA tries to run test files with their current working directory set to the directory that contains your `package.json` file. 

(Please mark the `package.json` bit as code.)

novemberborn:
If you rebase against `master` you should be able to make CI pass.

test/api.js
alathon: Ahh yeah, I see what you're saying. The fact that its even possible to create an Api() leading to an undefined pkgDir being given to fork is probably not a good idea. What about doing the following:

1. Set pkgDir in the same way we do now, but in api.js instead, around https://github.com/alathon/ava/blob/2a65c1295588ebdf36dcc833484c34f3ed956add/api.js#L55

2. Remove the option from cli.js (making it a flag you can set through the CLI is a decent idea, but probably a separate PR)

This would mean the test would work as expected, and test what it says - that if you just create an Api(), then it'll run from the package.json folder by default.

cli.js
```diff
@@ -139,6 +139,7 @@ var api = new Api({
 	babelConfig: conf.babel,
 	resolveTestsFrom: cli.input.length === 0 ? pkgDir : process.cwd(),
 	timeout: cli.flags.timeout,
+	pkgDir,
```
sindresorhus: You can't use object literal shorthand. We still support older Node.js versions.

test/api.js
alathon: I changed things to reflect what I suggest here. The commit history is a bit of a mess now though, due to the rebase from master happening before a pull/push. My bad ^^

novemberborn:
> Ahh yeah, I see what you're saying. The fact that its even possible to create an Api() leading to an undefined pkgDir being given to fork is probably not a good idea.

Yea that's kinda what <https://github.com/avajs/ava/issues/611> is about: making the interfaces more explicit.

>What about doing the following:
> 
> Set pkgDir in the same way we do now, but in api.js instead, around https://github.com/alathon/ava/blob/2a65c1295588ebdf36dcc833484c34f3ed956add/api.js#L55

I'd rather we keep passing it in, test issues aside. For instance <https://github.com/avajs/ava/issues/1041> relates to this PR as well, and we should just solve that in `cli.js`.

Could you go back to the previous implementation and fix the test instead? I'm actually OK with leaving the option implicit for the other places where tests create API instances.

> I changed things to reflect what I suggest here. The commit history is a bit of a mess now though, due to the rebase from master happening before a pull/push. My bad ^^

No worries, we'll just squash and merge it.

alathon:
I've reverted to the previous version where `pkgDir` is specified in cli.js.

test/api.js
```diff
@@ -18,6 +21,7 @@ test('must be called with new', function (t) {
 generateTests('Without Pool: ', function (options) {
 	options = options || {};
 	options.powerAssert = true;
+	options.pkgDir = options.pkgDir || path.dirname(pkgConf.filepath(conf));
```
novemberborn: Could you recompute this on line 11? It's run every time a test needs to create an API which seems unnecessary.

novemberborn:
You might be about to push the commit, in which case, apologies, but don't forget to fix the whitespace in the README 😉

alathon:
The Travis CI failures seem to be related to upgrading xo, am I right in this @sindresorhus ?

sindresorhus:
@alathon Fixed in https://github.com/avajs/ava/commit/6165a233c5f35bdf3538103f4a6cdcac4a560d6d.

cli.js
```diff
@@ -57,18 +57,18 @@ var cli = meow([
...
+	'  --serial, -s			Run tests serially',
+	'  --tap, -t				Generate TAP output',
+	'  --verbose, -v			Enable verbose output',
+	'  --no-cache				Disable the transpiler cache',
+	'  --no-power-assert			Disable Power Assert',
+	'  --match, -m				Only run tests with matching title (Can be repeated)',
+	'  --watch, -w				Re-run tests when tests and source files change',
+	'  --source, -S			Pattern to match source files so tests can be re-run (Can be repeated)',
+	'  --timeout, -T			Set global timeout',
+	'  --concurrency, -c			Maximum number of test files running at the same time (EXPERIMENTAL)',
```
novemberborn: @alathon still seeing a diff here.

alathon: Yes, there is a diff. But it lines up correctly in the terminal, when I execute it now. I'm not sure how to un-diff just that particular part of the file, although I know its possible through git.

novemberborn: Well it renders the same so I'm not too bothered I suppose. @sindresorhus?

alathon:
Branch updated with changes from upstream master now (to fix `xo` errors and make it easier to squash/rebase).

cli.js
sindresorhus: They're different as you changed from tabs to spaces. They should remain spaces.

sindresorhus: I'll fix it on merge, but try not to do unrelated changes when doing a PR ;)

You could have uncommitted that part btw. Would be easiest to use a GUI git app for that. I use [GitUp.app](http://gitup.co/).

sindresorhus:
Thanks @alathon :)

sindresorhus:
@novemberborn This fixes #32, right?

PR #1075, opened 2016-10-14
Update readme.md

alathon:
The readme isn't very explicit about whether `t.context` can be used to build up and share *between* tests, or not. I had to test this out manually to convince myself this wasn't the case, so maybe the readme should reflect it?

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -549,7 +549,7 @@ test(t => {
 });
 ```
 
-By default `t.context` is an object but you can reassign it:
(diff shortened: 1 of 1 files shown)


novemberborn:
Thanks @alathon!

sindresorhus:
`Update readme.md` is not a very good or descriptive commit title ;)

novemberborn:
@sindresorhus sorry :)

PR #1076, opened 2016-10-14
Speed up CI

novemberborn:
Inspired by <https://twitter.com/kentcdodds/status/786282951210262528> I looked into caching npm and `node_modules`, which led to some more improvements:

* Stop testing Node.js v5
* In Travis, only install latest npm if the Node.js version is older than v6 (though I suppose when v4 ships we may want to change that again, or actually force it to install v3)
* Cache npm cache and `node_modules` in both Travis and Appveyor
* Clean up Appveyor config a little

On Travis the first, non-cached run took 5m59s. The second run took 4m15s. I think it saves on nyc instrumentation too since that's cached in `node_modules`. On Appveyor the time went from 13m3s to 9m30s (note that Appveyor doesn't run builds in parallel for us).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ cache:
     - $HOME/.npm
     - node_modules
 before_install:
-  - 'node -e "process.exit(Number(process.version.match(/^v(\d+)/)[1])>=6?0:1)" || npm install -g npm@latest'
+  - 'node -e "process.exit(Number(process.version.match(/^v(\d+)/)[1])>=6?0:1)" || npm install -g npm@3.x-latest'
 before_script:
   - npm prune
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,11 @@ node_js:
   - '4'
   - '0.12'
   - '0.10'
+cache:
+  directories:
+    - $HOME/.npm
+    - node_modules
 before_install:
-  - 'node -e"process.exit(Number(process.version.match(/^v(\d+)/)[1])>=6?0:1)" || npm install -g npm@latest'
+  - 'node -e "process.exit(Number(process.version.match(/^v(\d+)/)[1])>=6?0:1)" || npm install -g npm@latest'
 after_success:
   - '[ -z "$COVERALLS_REPO_TOKEN" ] && tap --coverage-report=text-lcov | ./node_modules/.bin/coveralls'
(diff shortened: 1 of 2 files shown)


vdemedes:
Great stuff!

jamestalmage:
I definitely wouldn't cache `node_modules`. If deps are removed from `package.json`, they won't be cleared. This introduces initial state to the build, meaning CI no longer verifies it will install from scratch.

I am not sure about `$HOME/.npm`. If caching it does mask any problems, it seems certain those problems will be with `npm`.

jamestalmage:
The speedup is drastic enough on Appveyor that it's probably worth it there. I'd avoid it on Travis though.

novemberborn:
> I definitely wouldn't cache node_modules. If deps are removed from package.json, they won't be cleared. This introduces initial state to the build, meaning CI no longer verifies it will install from scratch.

npm still verifies all declared dependencies are present though. We could end up deleting a dependency that we still use, yes, but we have a similar problem with deduped dependencies.

Perhaps we could run `npm prune` after installation?

.travis.yml
```diff
@@ -1,11 +1,14 @@
...
   - '4'
   - '0.12'
   - '0.10'
+cache:
+  directories:
+    - $HOME/.npm
+    - node_modules
 before_install:
-  - 'npm install -g npm@latest'
+  - 'node -e "process.exit(Number(process.version.match(/^v(\d+)/)[1])>=6?0:1)" || npm install -g npm@latest'
```
vdemedes: Good catch.

sindresorhus:
> We could end up deleting a dependency that we still use, yes, but we have a similar problem with deduped dependencies.

The latest XO prevents this through the [`import/no-unresolved`](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-unresolved.md) rule.

novemberborn:
> The latest XO prevents this through the import/no-unresolved rule.

That should help. Worth pruning then? Are you OK with caching in Travis given #1080?

sindresorhus:
I'm still sceptical to caching the `node_modules` folder, but I'm open to trying it out and see how it works out. As long as we use [`np`](https://github.com/sindresorhus/np) when publishing, we're certain we'll at least never publish a broken version. And yes, let's do pruning.

Issue #1049, opened 2016-09-23
External macros are not transformed with power-assert


MethodGrab:
## Description

External AVA macros are not transformed with power-assert.

### Expected
Macros in the same file as the tests (internal macros) should have the same behaviour as those imported from an external file.  
Both tests should output the same error:

```js
test( 'internalMacro: 2 + 2 === 4', internalMacro, '2 + 2', 8 );
test( 'externalMacro: 2 + 2 === 4', externalMacro, '2 + 2', 8 );
```

### Actual
External macros are not transformed with power-assert and so display standard errors.

### Test Source

Full example: https://github.com/MethodGrab/ava-power-assert-external-macro-test-case

test.js:
```js
import test from 'ava';

import { externalMacro } from './macros';

function internalMacro( t, input, expected ) {
	console.log( 'running internal macro...' );
	t.is( eval( input ), expected );
}

// this will output a power-assert error
test( 'internalMacro: 2 + 2 === 4', internalMacro, '2 + 2', 8 );

// this will output a standard assert error
test( 'externalMacro: 2 + 2 === 4', externalMacro, '2 + 2', 8 );
```

macros.js:
```js
module.exports.externalMacro = function( t, input, expected ) {
	console.log( 'running external macro...' );
	t.is( eval( input ), expected );
}
```

### Error Message & Stack Trace

```
running internal macro...
  × internalMacro: 2 + 2 === 4
  t.is(eval(input), expected)
       |    |       |
       4    "2 + 2" 8

running external macro...
  × externalMacro: 2 + 2 === 4 4 === 8

  2 tests failed [23:18:50]


  1. internalMacro: 2 + 2 === 4
  AssertionError:
    4    "2 + 2" 8
        Test.internalMacro (test.js:7:4)
        _combinedTickCallback (internal/process/next_tick.js:67:7)
        process._tickCallback (internal/process/next_tick.js:98:9)


  2. externalMacro: 2 + 2 === 4
  AssertionError: 4 === 8
    Test.module.exports.externalMacro (macros.js:3:4)
    _combinedTickCallback (internal/process/next_tick.js:67:7)
    process._tickCallback (internal/process/next_tick.js:98:9)
```

### Config

```json
{
  "ava": {}
}
```

### Command-Line Arguments

```
ava --verbose test.js
```

## Relevant Links

- https://github.com/MethodGrab/ava-power-assert-external-macro-test-case

## Environment

- Node.js v6.3.0
- win32 6.1.7601
[...]

PR #1078, opened 2016-10-15
Precompile helpers

### LINKED ISSUE #720: Transpile helper files
From #577:

> [we] should be supporting transpilation of helper files (files in the test directory prefixed with `_`, or in `test/helpers`).

This would be a great feature. Currently it can be pretty frustrating to write helpers in ES5 when you can write tests in ES2016!

It wouldn't be too much effort to change the Api implementation so it finds helper files and precompiles them. I think if they're then added to [the `precompiled` option](https://github.com/sindresorhus/ava/blob/c4ec8e7e5965e5546d33ecf4fb00d9f4a8c9740c/api.js#L86) the forks will pick it up automatically.

vadimdemedes:
Fixes #720 and potentially #1049.

Lack of this feature has been bugging me for a long time, glad I've finally found a time to fix it. It's still WIP, need to:
- [x] Submit a PR to `ava-files` to support `findHelpers()` (currently named as `findTestDependencies()`).
- [x] Fix tests.
- [x] Write new tests.

As for #1049 (external macros are not precompiled), we could add a special glob for that, like `**/macros/**` to precompile those files as well.

### DIFF
--- a/api.js
+++ b/api.js
@@ -67,7 +67,7 @@ module.exports = Api;
 
 Api.prototype._runFile = function (file, runStatus, execArgv) {
 	var hash = this.precompiler.precompileFile(file);
-	var precompiled = {};
+	var precompiled = objectAssign({}, this._precompiledHelpers);
 	var resolvedfpath = fs.realpathSync(file);
 	precompiled[resolvedfpath] = hash;
 
@@ -139,7 +139,22 @@ Api.prototype._setupPrecompiler = function (files) {
 	});
 };
 
+Api.prototype._precompileHelpers = function () {
+	var self = this;
+
+	this._precompiledHelpers = {};
+
+	return new AvaFiles({cwd: this.options.resolveTestsFrom})
+		.findTestHelpers()
+		.map(function (file) { // eslint-disable-line array-callback-return
+			var hash = self.precompiler.precompileFile(file);
+			self._precompiledHelpers[file] = hash;
+		});
+};
+
 Api.prototype._run = function (files, options) {
+	var self = this;
+
 	options = options || {};
 
 	var runStatus = new RunStatus({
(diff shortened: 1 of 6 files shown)


novemberborn:
Good start! This compiles the helpers as if they were test files, right? That seems correct to me.

Let's see where that gets us before doing anything else for macros.

vdemedes:
> This compiles the helpers as if they were test files, right? That seems correct to me.

Yep, exactly.

> Let's see where that gets us before doing anything else for macros.

Yes for sure, that is unrelated to this PR.

api.js
```diff
@@ -130,7 +130,22 @@ Api.prototype._setupPrecompiler = function (files) {
...
 };
 
+Api.prototype._precompileHelpers = function () {
+	var self = this;
+
+	this._precompiledHelpers = {};
+
+	return new AvaFiles({cwd: this.options.resolveTestsFrom})
+		.findTestDependencies()
+		.map(function (file) { // eslint-disable-line array-callback-return
```
sindresorhus: Why use `.map` if you're not actually going to map anything?

test/api.js
```diff
@@ -705,9 +705,16 @@ function generateTests(prefix, apiCreator) {
 	test(prefix + 'caching is enabled by default', function (t) {
 		t.plan(3);
 		rimraf.sync(path.join(__dirname, 'fixture/caching/node_modules'));
+
+		var prevCwd = process.cwd();
+		process.chdir(path.join(__dirname, 'fixture/caching'));
```
novemberborn: Why this change?

novemberborn:
> Do we foresee people wanting to disable this behavior?

Specifically transpiling fixtures seems fraught.

vdemedes:
> Specifically transpiling fixtures seems fraught.

Could you elaborate on this? I don't quite understand the reasoning.

novemberborn:
Helpers should contain logic to make it easier to write test. Fixtures contain all sorts of files that relate to your test. If you're testing a Babel plugin you may write `.js` fixtures that you're using to exercise your tests. Those should not be transpiled.

test/api.js
novemberborn: The `pkgDir` and `resolveTestsFrom` directories can be passed to `Api` though. That's how its meant to be used, [implicit defaults](https://github.com/avajs/ava/issues/611) aside. We shouldn't need to change the working directory of the `tap` process.

sindresorhus:
@vdemedes What's your use-case for transpiling fixtures? I'm leaning towards @novemberborn's argument.

vdemedes:
My thinking was that everything that resides under `test/` belongs to AVA, so it should take care of those files and provide its features (power-assert, ES6, etc) to them as well.

However, @novemberborn's reasoning about fixtures relation to tests does make sense to me now. I will correct this PR and the one in `ava-files` as well.

test/api.js
vdemedes: Thanks for the tip!

Snugug:
w/r/t #1049, I've just installed this branch and ran against an external macro, and it did not appear to resolve the issue that `powerAssert` is not applied to those tests

vadimdemedes:
@Snugug only files with names starting with `_` and located in `helpers/` folders get transpiled in this PR. In the issue you mentioned the files are located in `macros/` folder, right?

sindresorhus:
Code looks good to me, but needs to be added to the readme.

vadimdemedes:
Added a small note to readme: https://github.com/avajs/ava/pull/1078/files#diff-0730bb7c2e8f9ea2438b52e419dd86c9R738.

sindresorhus:
Yay! This one is going to be super useful 💃

PR #1080, opened 2016-10-16
Upgrade XO

novemberborn:
The latest `xo` no longer supports Node.js v0.10 and v0.12. Don't run xo on those versions.

Upgrade `xo` and fix linting issues.

Mess up `cli.js` to avoid top-level returns. `xo` cannot parse the file due to <https://github.com/sindresorhus/xo/issues/51>.


novemberborn:
The reason for upgrading, rather than leaving `xo` pinned at `0.16.0` is that when I'm editing using [`atom-linter-xo`](https://github.com/sindresorhus/atom-linter-xo) I still get all the errors from `0.17.0`. This is annoying to me and possibly confusing to contributors.

cli.js
```diff
@@ -13,186 +13,192 @@ var localCLI = resolveCwd('ava/cli');
 // see https://github.com/nodejs/node/issues/6624
 if (localCLI && path.relative(localCLI, __filename) !== '') {
 	debug('Using local install of AVA');
-	require(localCLI);
-	return;
+	require(localCLI); // eslint-disable-line import/no-dynamic-require
+} else {
+	load();
 }
```
sindresorhus: Instead of nesting the below in a function, maybe we should split it out into a new file and just have this file as a conditional loader CLI?

```js
if (localCLI && path.relative(localCLI, __filename) !== '') {
	debug('Using local install of AVA');
	require(localCLI); // eslint-disable-line import/no-dynamic-require
} else {
	require('./cli');
}
```

And rename this file to `_cli.js` or something.

test/visual/run-visual-tests.js
```diff
@@ -1,5 +1,5 @@
 'use strict';
-require('loud-rejection');
+require('loud-rejection'); // eslint-disable-line import/no-unassigned-import
```
sindresorhus: Actually, just noticed. This does nothing. It should either be executed as a function or imported as `loud-rejection/register`.

sindresorhus:
> The reason for upgrading, rather than leaving xo pinned at 0.16.0 is that when I'm editing using atom-linter-xo I still get all the errors from 0.17.0. This is annoying to me and possibly confusing to contributors.

https://github.com/sindresorhus/atom-linter-xo/issues/41

test/visual/run-visual-tests.js
novemberborn: Is it even necessary?

sindresorhus: Yes, in case of rejected promises, since we don't manually catch them all. This is easier.

sindresorhus:
@novemberborn Looks good. Can you fix the merge conflict and merge this?

novemberborn:
@sindresorhus fixed conflict. Also added a commit for the `loud-rejection` thing.

PR #1084, opened 2016-10-18
Add test for ava-throws-helper hoisting await and yield

### LINKED ISSUE #978: Plans to transpile the source to ES2015?
Plans to transpile the source to ES2015? Babel are already used as an dependency and also part of the core.

### LINKED ISSUE #987: t.throws() causes Node syntax error if first argument is function call with await arguments
## Description

I have a test that breaks after updating ava to version ~0.15.2 that works when using ava@0.14.0. This has been tested on node 6.3.1 and 4.2.4 amongst others. and seems related to a missing "use strict" after transpiling ES6 modules (wild guess there). I've looked at the commits between the two versions but haven't been able to identify what might cause the problem, but I haven't tested individual commits due to lack of time.
### Test Source, error message, and configuration

You can see the relevant files and error message here: https://gist.github.com/nesukun/f8cb2eef133548224d21bfd8a3ae5aa4

As a side note, using AVA's default babel config does not solve or alter the result. Hope you don't mind having everything in a gist, as I prepared that for the gitter channel before opening the issue.
## Environment

Tested mainly using the versions of node specified above under macOS 10.11.5 and inside the docker default node image

novemberborn:
Needs https://github.com/avajs/babel-plugin-ava-throws-helper/pull/6 so CI should fail until that is shipped. We should bump the minimum version in `package.json`.

Fixes #987.

### DIFF
--- a/test/cli.js
+++ b/test/cli.js
@@ -123,6 +123,14 @@ test('improper use of t.throws from within an async callback will be reported to
 	});
 });
 
+test('improper use of t.throws (with an awaited argument) will be reported to the console', function (t) {
+	execCli('fixture/improper-t-throws-await-argument.js', function (err, stdout, stderr) {
+		t.ok(err);
+		t.match(stderr, /Improper usage of t\.throws detected at .*improper-t-throws-await-argument.js \(5:10\)/);
+		t.end();
+	});
+});
+
 test('babel require hook only applies to the test file', function (t) {
 	t.plan(3);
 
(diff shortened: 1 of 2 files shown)


sindresorhus:
I think you linked to the wrong issue. Can you correct that?

novemberborn:
> I think you linked to the wrong issue. Can you correct that?

Oops. Fixed.

PR #1085, opened 2016-10-18
Ensure test files load correct AVA installation

### LINKED ISSUE #643: Error when running tests specifying an absolute path
Hello!,

I'm using the version 0.13.0 with the following environment: Node.js v5.6.0, darwin 15.3.0.

When I execute ava in the same folder than its contents the operation gets completed without problem. However, if I try to do the same specifying an absolute path outside the folder that holds the file I get:

> ava Repositories/x/y.js 
> 1 exception
> 1. No tests found in Repositories/nodetests/test1.js, make sure to import "ava" at the top of your test file

Any idea?. Thanks!,

novemberborn:
The main process sets the AVA_PATH environment variable to the absolute
path of the index.js file. Workers are loaded with this variable
present. When test files require the AVA module (assuming it's a version
containing this commit of course), the index.js file redirects to the
one used by the worker if necessary by comparing AVA_PATH.

The redirect required most of index.js to be moved into a separate
module (lib/main.js).

Fixes #643.


vdemedes:
LGTM!

sindresorhus:
🙌❤️

PR #1096, opened 2016-10-27
Add docs for test parameter need to be named `t`

### LINKED ISSUE #1031: Docs: Really, use `t` for assertions
I just tried to covert a project that's tested with [tape](https://github.com/substack/tape) over to using AVA instead. The tests all follow this kind of format:

``` js
test('function-a', (assert) => {
  assert.plan(1);

  assert.is(
    functionA(),
    resultB,
    'function a should return result b'
  );
});
```

They seem to run just fine under AVA, but the assertion messages don't show up when I create a failing test case, just the test messages.

Is there a way to report on the assertion messages? Or is the best practice just to wrap every one of these assertions in its own `test` block? (In which case, why do the assertions accept optional message parameters?

Am I missing something stupid?

gurpreetatwal:
- Reworded the statement in the readme regarding the first parameter needing to be `t` to make it stronger and explain why its required.
- Add a note in common-pitfalls regarding the same issue
- Make the indentation style consistent in the code samples 

Closes #1031

### DIFF
--- a/docs/common-pitfalls.md
+++ b/docs/common-pitfalls.md
@@ -63,6 +63,16 @@ test(async t => {
 
 AVA [can't trace uncaught exceptions](https://github.com/avajs/ava/issues/214) back to the test that triggered them. Callback-taking functions may lead to uncaught exceptions that can then be hard to debug. Consider promisifying and using `async`/`await`, as in the above example. This should allow AVA to catch the exception and attribute it to the correct test.
 
+### Why is the power-assert information not shown?
+
+Ensure that the first parameter passed into your test is named `t`. This is a byproduct of the way power-assert works. It uses a pattern matching scheme that makes it easier for implementors to wrap any assertion library with power-assert goodness without having to understand the ES AST at all. See [#1031](https://github.com/avajs/ava/issues/1031) for more details.
+
+```js
+test(t => {
+	t.is(1, 1);
+}
+```
+
 ---
 
 Is your problem not listed here? Submit a pull request or comment on [this issue](https://github.com/avajs/ava/issues/404).
--- a/docs/common-pitfalls.md
+++ b/docs/common-pitfalls.md
@@ -63,14 +63,13 @@ test(async t => {
 
 AVA [can't trace uncaught exceptions](https://github.com/avajs/ava/issues/214) back to the test that triggered them. Callback-taking functions may lead to uncaught exceptions that can then be hard to debug. Consider promisifying and using `async`/`await`, as in the above example. This should allow AVA to catch the exception and attribute it to the correct test.
 
-### Why is the power-assert information not shown?
-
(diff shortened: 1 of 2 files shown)


docs/common-pitfalls.md
```diff
@@ -63,6 +63,16 @@ test(async t => {
 
 AVA [can't trace uncaught exceptions](https://github.com/avajs/ava/issues/214) back to the test that triggered them. Callback-taking functions may lead to uncaught exceptions that can then be hard to debug. Consider promisifying and using `async`/`await`, as in the above example. This should allow AVA to catch the exception and attribute it to the correct test.
 
+### Why is the power-assert information not shown?
+
+Ensure that the first parameter passed into your test is named `t`. This is a byproduct of the way power-assert works. It uses a pattern matching scheme that makes it easier for implementors to wrap any assertion library with power-assert goodness without having to understand the ES AST at all. See [#1031](https://github.com/avajs/ava/issues/1031) for more details.
```
vdemedes: > This is a byproduct of the way power-assert works.

I'd rather say that "This is a requirement by power-assert.".

vdemedes:
Good job explaining the what & why!

docs/common-pitfalls.md
novemberborn: Could you change this to:

> ### Why are the enhanced assertion messages not shown?
>
> Ensure that the first parameter passed into your test is named `t`. This is a requirement of how AVA uses [`power-assert`](https://github.com/power-assert-js/power-assert), the library that provides the enhanced messages. works.

I see that this paragraph originally comes from @jamestalmage's comment in https://github.com/avajs/ava/issues/1031#issuecomment-244474530. It's quite technical though. I don't think we need that kind of detail here.

docs/common-pitfalls.md
```diff
@@ -63,6 +63,16 @@ test(async t => {
...
+
+```js
+test(t => {
+	t.is(1, 1);
+}
```
novemberborn: Missing `);` at the end here.

readme.md
```diff
@@ -74,13 +74,13 @@ Your `package.json` will then look like this:
...
-  "devDependencies": {
-    "ava": "^0.15.0"
-  }
+	"name": "awesome-package",
+	"scripts": {
+		"test": "ava"
+	},
+	"devDependencies": {
+		"ava": "^0.15.0"
+	}
```
novemberborn: Could you revert these to their original two space indentation? It's how `package.json` files are formatted.

docs/common-pitfalls.md
gurpreetatwal: I fully agree, that was just some blind copy and pasting from the comment on my part. Should I still keep the link to the issue for anyone who wants to know more?

novemberborn: Nah, the issue still requires a lot of careful reading and background to fully understand why it is this way.

sindresorhus:
Thanks :)

PR #1098, opened 2016-10-29
Add `--no-color` flag - fixes #843

### LINKED ISSUE #843: Switch to disable color output to console

thinkimlazy:
Disables color output with `--no-color` option for CLI and 
`"noColor": true` for package.json
Resolves #843

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -8,6 +8,7 @@ var Promise = require('bluebird');
 var pkgConf = require('pkg-conf');
 var isCi = require('is-ci');
 var hasFlag = require('has-flag');
+var chalk = require('chalk');
 var Api = require('../api');
 var colors = require('./colors');
 var verboseReporter = require('./reporters/verbose');
@@ -41,6 +42,7 @@ exports.run = function () {
 		'  --source, -S       Pattern to match source files so tests can be re-run (Can be repeated)',
 		'  --timeout, -T      Set global timeout',
 		'  --concurrency, -c  Maximum number of test files running at the same time (EXPERIMENTAL)',
+		'  --no-color         Disable color output',
 		'',
 		'Examples',
 		'  ava',
@@ -85,6 +87,16 @@ exports.run = function () {
 	if (cli.flags.init) {
 		require('ava-init')();
 		return;
+	}
+
+	if (
+		hasFlag('--no-color') ||
+		conf.noColor
+	) {
+		chalk.enabled = false;
+		Object.keys(colors).forEach(function (key) {
+			colors[key].enabled = false;
+		});
 	}
 
 	if (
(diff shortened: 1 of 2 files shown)


novemberborn:
@thinkimlazy thanks for your PR.

Reading #843 I can't determine whether @sindresorhus wanted this fixed in `chalk`, or whether he wanted a flag in AVA. Ideally this just works without needing additional options.

@sindresorhus?

thinkimlazy:
@novemberborn Hey, thats not working ideally as you want because of this code in verbose/mini reporters:
```
chalk.enabled = true;
Object.keys(colors).forEach(function (key) {
	colors[key].enabled = true;
});
```

So when `chalk` initialized in reporters, `--no-color` really disables colors, but this loop enables them. So there is no need to fix anything in `chalk`. 
We can either delete this loops in reporters, then `--no-color` just works.

snypelife:
Killing those `forEach`'s in the reporters seems to work nicely, as well as clean things up a bit. But I  noticed that env vars like `CI` or `TEAMCITY_VERSION` aren't getting picked up from https://github.com/chalk/supports-color/blob/master/index.js.

snypelife:
@novemberborn from my perspective of that conversation in the issue, it looks like both a fix in the supports-color module as well as ava were sought after (ava being the flag needs to be passed down and supports-color needing teamcity <9 support).

lib/cli.js
```diff
@@ -88,6 +90,16 @@ exports.run = function () {
 	}
 
 	if (
+		hasFlag('--no-color') ||
+		conf.noColor
```
sindresorhus: This is incorrect. It should be:

```js
cli.flags.color === false
```

lib/cli.js
```diff
@@ -88,6 +90,16 @@ exports.run = function () {
...
+	) {
+		chalk.enabled = false;
+		Object.keys(colors).forEach(function (key) {
+			colors[key].enabled = false;
+		});
```
sindresorhus: I think `chalk.enabled = false` should be enough.

sindresorhus: Actually. None of this should be needed as `chalk` already supports the `--no-color` flag.

sindresorhus: What actually needs to be done is to pass the flag down to the forked process.

readme.md
```diff
@@ -228,7 +229,8 @@ All of the CLI options can be configured in the `ava` section of your `package.j
     "require": [
       "babel-register"
     ],
-    "babel": "inherit"
+    "babel": "inherit",
+    "noColor": true
```
```
"color": false
```

sindresorhus:
@thinkimlazy Why did you close?

thinkimlazy:
@sindresorhus sry, I messed up  
I reopened [PR ](https://github.com/avajs/ava/pull/1104)

sindresorhus:
@thinkimlazy You never have to open a new PR. That just creates noise in the issue tracker. You can always reset the state of an existing PR and start over if you really messed up.

thinkimlazy:
@sindresorhus I can't reopen this one because I recreated my branch. Again sorry for spam

sindresorhus:
No worries :)

PR #1104, opened 2016-11-05
Pass `--no-color` flag down to the forked process - fixes #843

### LINKED ISSUE #843: Switch to disable color output to console

thinkimlazy:
Sorry that I accidentally deleted last one.

Also I want to notice that just `chalk.enabled = true` not working, because functions from `lib/colors.js` has own `enabled` property. So we need that loop
```
Object.keys(colors).forEach(function (key) {
	colors[key].enabled = true;
});
```

### DIFF
--- a/lib/reporters/mini.js
+++ b/lib/reporters/mini.js
@@ -11,11 +11,6 @@ var repeating = require('repeating');
 var objectAssign = require('object-assign');
 var colors = require('../colors');
 
-chalk.enabled = true;
-Object.keys(colors).forEach(function (key) {
-	colors[key].enabled = true;
-});
-
 function MiniReporter(options) {
 	if (!(this instanceof MiniReporter)) {
 		return new MiniReporter(options);
--- a/test/cli.js
+++ b/test/cli.js
@@ -6,7 +6,6 @@ var test = require('tap').test;
 global.Promise = require('bluebird');
 var getStream = require('get-stream');
 var figures = require('figures');
-var arrify = require('arrify');
 var chalk = require('chalk');
 var mkdirp = require('mkdirp');
 var touch = require('touch');
@@ -20,6 +19,10 @@ var cliPath = path.join(__dirname, '../cli.js');
 // for some reason chalk is disabled by default
 chalk.enabled = true;
 var colors = require('../lib/colors');
+
+Object.keys(colors).forEach(function (key) {
+	colors[key].enabled = true;
+});
 
 function execCli(args, opts, cb) {
 	var dirname;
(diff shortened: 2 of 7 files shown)


test/cli.js
```diff
@@ -6,7 +6,6 @@ var test = require('tap').test;
 global.Promise = require('bluebird');
 var getStream = require('get-stream');
 var figures = require('figures');
-var arrify = require('arrify');
```
thinkimlazy: We don't need `arrify` because `Array.prototype.concat` already can concat single values
[Array.prototype.concat docs](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/concat)

lib/reporters/mini.js
```diff
@@ -11,11 +11,6 @@ var repeating = require('repeating');
 var objectAssign = require('object-assign');
 var colors = require('../colors');
 
-chalk.enabled = true;
-Object.keys(colors).forEach(function (key) {
-	colors[key].enabled = true;
-});
-
```
thinkimlazy: This turn on colors no matter what.

test/cli.js
```diff
@@ -21,6 +20,10 @@ var cliPath = path.join(__dirname, '../cli.js');
 chalk.enabled = true;
 var colors = require('../lib/colors');
 
+Object.keys(colors).forEach(function (key) {
+	colors[key].enabled = true;
+});
+
```
thinkimlazy: We need to turn on colors no matter what while testing. Also in case that Istanbul turning them off

test/cli.js
```diff
@@ -43,7 +46,7 @@ function execCli(args, opts, cb) {
 	var stderr;
 
 	var processPromise = new Promise(function (resolve) {
-		child = childProcess.spawn(process.execPath, [path.relative(dirname, cliPath)].concat(arrify(args)), {
+		child = childProcess.spawn(process.execPath, [path.relative(dirname, cliPath)].concat(args, '--color'), {
```
thinkimlazy: We need to turn on colors, since `Istanbul` [disables colors at this condition](https://github.com/chalk/supports-color/blob/master/index.js#L41)

sindresorhus:
It also needs to pass the `--no-color` flag down to the forked process. If I add `console.log(require('chalk').red('colorzzzz'));` in a test file, it will still log colors even with `$ ava --no-colors`. I would expect that to prevent all colors.

thinkimlazy:
Actually colors in test files disabled by default. 
```
import test from 'ava'
import chalk from 'chalk'

test('kek', t => {
    t.false(chalk.enabled); // passes
})
```

sindresorhus:
Not for me:

```
❯ ava

  1 failed

  no colors here

  t.false(chalk.enabled)
                |
                true

    Test.fn (test.js:5:5)
    internal/child_process.js:721:12
```

vadimdemedes:
Hey @thinkimlazy, happy holidays! Will you have some free time soon to finish this up?

thinkimlazy:
@vadimdemedes Hello, I'm quite busy right now so I'm not sure that I can fix this any time soon.

sindresorhus:
Closing for lack of activity and to free it up for someone else interested in working on this.

PR #1113, opened 2016-11-17
Add snapshot assertion

lithin:
This PR introduces snapshot testing into ava, using the jest-snapshost package that has been recently rewritten to allow usage in other runners than Jest. The reason behind it is that in our new project, we'd love to use Ava as it's absolutely brilliant. However, snapshot testing is a requirement for our project, and so we need to introduce it in one way or another.

- [x] Introduce jest snaphost in a spike to test viability
- [x] Allow updating of snapshots
- [x] Produce filename and snapshot path dynamically
- [x] Enhance reporting
- [x] Unit testing
- [x] Write documentation
- [x] Hook up code coverage
- [x] Check this works with con-current tests
- [x] Use correct test name for snapshots
- [x] ~~CRITICAL DEPENDENCY: submit pull request to jest-snapshots for minor changes that made this PR easier (see https://github.com/facebook/jest/pull/2130) - uses npm link locally atm~~ I removed the new version for jest snapshots as a dependency for now - the owner is currently on holiday for a week :x
- [x] ~~Check what "expand" option in snapshot state does~~ Turns out it expands the diff which we want by default
- [x] Enable making two snapshots in one test
- [x] Add snapshots into test/assertion planning

All ideas, suggestions, comments are very welcome!


sindresorhus:
Makes sense to me. Lots of users need this.

lib/assert.js
```diff
@@ -151,6 +153,23 @@ x.ifError = function (err, msg) {
 	test(!err, create(err, 'Error', '!==', msg, x.ifError));
 };
 
+x.snap = function (tree, filename, snapshotPath) {
```
sindresorhus: `snap` => `snapshot`

sindresorhus:
> Produce filename and snapshot path dynamically

I would look at how Jest handles this.

> Enhance reporting

@vdemedes is working on a new output which includes diffing, so this could simply use `t.is()` comparison of the snapshot, and diffing will be supported soon.

lithin:
@sindresorhus thanks for such quick feedback!

### Filepath & directory
I totally agree that we should look at how jest handles this sort of stuff. Is there any way in Ava to see what filename/path the current test is in? 

### Reporting
As for reporting, Jest snapshots generate a message for you:

screen shot 2016-11-18 at 9 00 31 am

I'm not sure how easy it would be to change this?

### Snapshots directory
At the moment, Jest saves snapshots into the same folder as the current test is in. We were wondering with my colleagues if it wouldn't be better to create a `__snapshots__` folder in the root folder. What do you think?

lithin:
Here are a few updates on my progress:

- We decided to follow jest and have `__snapshots__` directory next to the test files
- Checked code coverage with `nyc` works as expected
- Reporting has been slightly improved - please check the screenshot

screen shot 2016-11-18 at 2 16 55 pm

lithin:
@sindresorhus I think I've finished - could you please review?

lithin:
Builds are failing because jest-snapshot is using `const` in build files which isn't compatible with node 0.12. Is there any chance of upgrading node on travis?

sindresorhus:
> Builds are failing because jest-snapshot is using const in build files which isn't compatible with node 0.12. Is there any chance of upgrading node on travis?

Just ignore that for now. We're dropping Node.js 0.12 support very soon.

lithin:
@sindresorhus Does that mean we need to wait for the upgrade to be made?

We just caught one bug so I did a quick fix :) It's again ready for review.

lib/snapshot-state.js
```diff
@@ -0,0 +1,27 @@
+var jestSnapshot = require('jest-snapshot');
+var path = require('path');
+var globals = require('./globals');
+
+var x = module.exports;
+
+x.get = function (initializeState, globalsOptions) {
+  if (!this.state) {
+    // set defaults - this allows tests to mock deps easily
+  	var options = globalsOptions || globals.options;
```
lpil: This file has mixed tabs and spaces.

lib/snapshot-state.js
```diff
@@ -0,0 +1,27 @@
+var jestSnapshot = require('jest-snapshot');
+var path = require('path');
+var globals = require('./globals');
+
+var x = module.exports;
```
lpil: Why the `var x`?

test/snapshot-state.js
```diff
@@ -0,0 +1,45 @@
...
+
+	t.plan(3);
+
+	t.doesNotThrow(function () {
+		var result = snapshotState.get(stateStub, {
+			file: 'hello/world.test.js',
+			updateSnapshots: false
+		});
+
+    t.is(result, 'state');
```
lpil: Mixed tabs and spaces in this file.

lib/snapshot-state.js
lithin: thanks will fix!

test/snapshot-state.js
lithin: ...and here too :D

sindresorhus:
> Does that mean we need to wait for the upgrade to be made?

No. I just upgraded a big part of the code base, but made sure not to touch anything changed by this pull request. Feel free to start using ES2015 features. Otherwise I'll do it after merge, no problem.

lib/cli.js
```diff
@@ -31,18 +31,19 @@ exports.run = function () {
...
+		'  --tap, -t          		Generate TAP output',
+		'  --verbose, -v      		Enable verbose output',
+		'  --no-cache         		Disable the transpiler cache',
+		'  --no-power-assert  		Disable Power Assert',
+		'  --match, -m        		Only run tests with matching title (Can be repeated)',
+		'  --watch, -w        		Re-run tests when tests and source files change',
+		'  --source, -S       		Pattern to match source files so tests can be re-run (Can be repeated)',
+		'  --timeout, -T      		Set global timeout',
+		'  --concurrency, -c  		Maximum number of test files running at the same time (EXPERIMENTAL)',
+		'  --update-snapshots, -u Update snapshots',
```
sindresorhus: The description are unaligned now. Same with the one in the readme.

lib/snapshot-state.js
```diff
@@ -0,0 +1,27 @@
+var jestSnapshot = require('jest-snapshot');
```
sindresorhus: Put a `'use strict';` at the top, above this.

readme.md
```diff
@@ -909,6 +910,48 @@ Assert that `contents` does not match `regex`.
 
 Assert that `error` is falsy.
 
+## Snapshot testing
+
+Snapshot testing comes as another kind of assertion and uses [jest-snapshot](https://facebook.github.io/jest/blog/2016/07/27/jest-14.html) under the hood.
+
+When used with react, it looks very similar to Jest:
```
sindresorhus: react => React

readme.md
```diff
@@ -909,6 +910,48 @@ Assert that `contents` does not match `regex`.
...
+```
+
+```js
+// your test
+import test from 'ava';
+import render from 'react-test-renderer';
+
+import HelloWorld from './';
+
+test('HelloWorld component', (t) => {
```
sindresorhus: `(t)` => `t`

readme.md
```diff
@@ -909,6 +910,48 @@ Assert that `contents` does not match `regex`.
...
+  const tree = render.create(<HelloWorld />).toJSON();
+  t.snapshot(tree);
+});
+```
+
+The first time you run this test, a snapshot file will be created in `__snapshots__` folder looking something like this:
+
+```
```
```
```js
```

readme.md
```diff
@@ -909,6 +910,48 @@ Assert that `contents` does not match `regex`.
...
+exports[`HelloWorld component 1`] = `
+<h1>
+  Hello World...!
+</h1>
+`;
+```
+
+Every time you run this test afterwards, it'll check if the component render has changed. If it did, it'll fail the test. Then you'll have the choice to check your code - and if the change was intentional, you can use `--update-snapshots` (or `-u`) flag to update the snapshots into their new version.
```
sindresorhus: it'll => it will

sindresorhus: you can use => you can use the

lib/snapshot-state.js
```diff
@@ -0,0 +1,27 @@
...
+var path = require('path');
+var globals = require('./globals');
+
+var x = module.exports;
+
+x.get = function (initializeState, globalsOptions) {
```
sindresorhus: `x.get` => `exports.get` and drop the `module.exports`

sindresorhus:
@lithin This looks very good :) I'll play with it some more tomorrow and do a more thorough review.

Does snapshot testing only work with JSX or can you use it with other things too?

You also need to add an entry to the Assertion section: https://github.com/avajs/ava#assertions

Can you fix the lint issues reported by Travis: https://travis-ci.org/avajs/ava/jobs/177985206 ?

lib/assert.js
```diff
@@ -151,6 +153,29 @@ x.ifError = function (err, msg) {
...
+	};
+
+	var result = toMatchSnapshot.call(context, tree);
+
+	var messagePrefix = 'Please check your code or --update-snapshots\n\n';
+	var message = typeof result.message === 'function' && (messagePrefix + result.message());
+
+	state.save();
+
+	test(result.pass, create(result, false, 'snap', message, x.snap));
```
vdemedes: `'snap'` => `'snapshot'`
`x.snap` => `x.snapshot`

readme.md
```diff
@@ -909,6 +910,48 @@ Assert that `contents` does not match `regex`.
...
+
+That might look like this:
+
+`npm t -- -u`
```
vdemedes: I'd skip the npm script and give an example with `ava` binary:

```
$ ava --update-snapshots
```

test/assert.js
```diff
@@ -495,3 +496,58 @@ test('.deepEqual() should not mask RangeError from underlying assert', function
 
 	t.end();
 });
+
+test('snapshot makes a snapshot using a library and global options', function (t) {
+	var saveSpy = sinon.spy();
+	var state = {
```
vdemedes: Can be on the same line:

```js
var state = {save: saveSpy};
```

test/assert.js
```diff
@@ -495,3 +496,58 @@ test('.deepEqual() should not mask RangeError from underlying assert', function
...
+		save: saveSpy
+	};
+	var stateGetter = sinon.stub().returns(state);
+	var matchStub = sinon.stub().returns({
```
vdemedes: Can also be on the same line as well.

test/assert.js
```diff
@@ -495,3 +496,58 @@ test('.deepEqual() should not mask RangeError from underlying assert', function
...
+
+	delete assert.title;
+
+	t.end();
+});
+
+test('if snapshot fails, prints a message', function (t) {
+	var saveSpy = sinon.spy();
+	var state = {
+		save: saveSpy
```
vdemedes: Can be on the same line too.

test/snapshot-state.js
```diff
@@ -0,0 +1,45 @@
+var test = require('tap').test;
```
vdemedes: `'use strict';` is missing here.

vdemedes:
Really well done, @lithin! Nicely organized and clean PR, thanks a lot for working on this!

lib/assert.js
```diff
@@ -151,6 +153,29 @@ x.ifError = function (err, msg) {
 	test(!err, create(err, 'Error', '!==', msg, x.ifError));
 };
 
+x.snapshot = function (tree, match, snapshotStateGetter) {
```
sindresorhus: We should not expose these internal arguments `match` and `snapshotStateGetter` to the outside. Instead you could expose an internal base method.

```js
x._snapshot = function (tree, match, snapshotStateGetter) {

...

x.snapshot = function (tree) {
```

sindresorhus: It should also have an optional `message` argument like the other assertions.

readme.md
```diff
@@ -909,6 +910,48 @@ Assert that `contents` does not match `regex`.
 
 Assert that `error` is falsy.
 
+## Snapshot testing
+
```
sindresorhus: It should also include how to handle the snapshots. Should they be committed or gitignored.

readme.md
lithin: This one isn't js actually - it's just a `snap` file

test/snapshot-state.js
lithin: 👍

lithin:
Thanks @vdemedes and @sindresorhus for the feedback - hopefully I didn't miss anything :)

lithin:
@sindresorhus it seems one of my comments got lost in all the changes - how do you propose to use the x.snapshot vs x._snapshot? I wasn't quite sure how to implement that.

lib/snapshot-state.js
lithin: Had to return this back to make it work with `.state`

sindresorhus:
I was thinking something like this:

```js
x._snapshot = function (tree, optionalMessage, match, snapshotStateGetter) {
	// actual implementation here
};

x.snapshot = function (tree, optionalMessage) {
	x._snapshot(tree, optionalMessage);
};
```

That way the user (which uses `x.snapshot`) can never accidentally supply those internal arguments, and we can still easily test it using `x._snapshot`.

lib/assert.js
```diff
@@ -169,11 +169,11 @@ x.snapshot = function (tree, match, snapshotStateGetter) {
 	var result = toMatchSnapshot.call(context, tree);
 
 	var messagePrefix = 'Please check your code or --update-snapshots\n\n';
-	var message = typeof result.message === 'function' && (messagePrefix + result.message());
+	var message = optionalMessage || (typeof result.message === 'function' && (messagePrefix + result.message()));
```
sindresorhus: The message argument should be in addition to the default output, not replace it. Probably best to prepend before `messagePrefix`.

lib/enhance-assert.js
```diff
@@ -18,7 +18,8 @@ module.exports.PATTERNS = [
 	't.ok(value, [message])',
 	't.notOk(value, [message])',
 	't.same(value, expected, [message])',
-	't.notSame(value, expected, [message])'
+	't.notSame(value, expected, [message])',
+	't.snapshot(contents, [message])'
```
sindresorhus: Should be added to the below `NON_ENHANCED_PATTERNS` as this enhancement replaces the output, which we don't want here.

lib/assert.js
```diff
@@ -151,6 +153,29 @@ x.ifError = function (err, msg) {
...
+	var context = {
+		dontThrow: function () {},
+		currentTestName: this.title,
+		snapshotState: state
+	};
+
+	var result = toMatchSnapshot.call(context, tree);
+
+	var messagePrefix = 'Please check your code or --update-snapshots\n\n';
+	var message = optionalMessage || (typeof result.message === 'function' && (messagePrefix + result.message()));
```
sindresorhus: The message argument should be in addition to the default output, not replace it. Probably best to prepend before `messagePrefix`.

sindresorhus:
I noticed if create a snapshot, rename the test, and run `ava -u`, the previous test is still in the snapshot file. Is this intentional?

---

I'm not getting the colorized diff output like you have:

screen shot 2016-11-25 at 14 15 23

---

The generated output should also be indentated so it's aligned with the test title in the output. You can use [`indent-string`](https://github.com/sindresorhus/indent-string) for this.

lib/assert.js
```diff
@@ -151,6 +153,29 @@ x.ifError = function (err, msg) {
 	test(!err, create(err, 'Error', '!==', msg, x.ifError));
 };
 
+x.snapshot = function (tree, optionalMessage, match, snapshotStateGetter) {
+	// set defaults - this allows tests to mock deps easily
+	var toMatchSnapshot = match || jestSnapshot.toMatchSnapshot;
+	var getState = snapshotStateGetter || snapshotState.get;
```
sheepsteak: @lithin this needs to be `snapshotState.get.bind(snapshotState)` or `snapshotState.get.bind(this)` as it uses `this` internally and I keep getting the error: `Error: Cannot read property 'state' of undefined`.

lithin: This has been fixed

lithin:
@sindresorhus All fixed :)

As for the report output, I'm not sure what happened there :x I found out that no matter how I use chalk from within jest or ava, it wouldn't colour anything. I thought it might be cause by the reporter but now I realised it used to work... Very confusing! Any ideas what this might be caused by?

I haven't seen anywhere that the test would appear twice after updating the snapshot. It could happen if you've changed the name of the test though?

sindresorhus:
The indentation is not correct. As commented earlier, it should be vertically aligned with the test title. So you need to add 2 space indent. It seems the first line in the output is already indented by the assertion output, so we just indent the following text. Like this:

```js
var message = 'Please check your code or --update-snapshots\n\n';

if (optionalMessage) {
	message += indentString(optionalMessage, 2);
}

if (typeof result.message === 'function') {
	message += indentString(result.message(), 2);
}

state.save();

test(result.pass, create(result, false, 'snapshot', message, x.snap));
```

sindresorhus:
> Any ideas what this might be caused by?

Yes. The error is colored red by [our reporter](https://github.com/avajs/ava/blob/abd947e86fc6d80ffd8113d04f59a03bfb943879/lib/reporters/mini.js#L196). I'm more confused by how you actually got a diff color there in the first place. But we can ignore it for this pull request, as we're very close to a rewrite of the reporter output, where this can be handled.

sindresorhus:
> I haven't seen anywhere that the test would appear twice after updating the snapshot. It could happen if you've changed the name of the test though?

Yes, that's what I wrote. I renamed the test and then ran `ava -u`. Shouldn't that remove old tests?

lithin:
> The error is colored red by our reporter. I'm more confused by how you actually got a diff color there in the first place.

I'm really confused about that as well. The colouring coming from jest diff used to work fine. Now, even if I remove the `test(... message ...)` in the snapshot assertion and just do `console.log(message)`, the message isn't coloured.

> Shouldn't that remove old tests?

Jest identifies tests by their name - if you rename a test, it's as if you were running a new one. It has no way of telling if a test of another name should be identified with the renamed one. In other words, I don't think it can remove old tests from snapshots - it has to be done manually.

sindresorhus:
> Jest identifies tests by their name - if you rename a test, it's as if you were running a new one. It has no way of telling if a test of another name should be identified with the renamed one. In other words, I don't think it can remove old tests from snapshots - it has to be done manually.

Yes, that's desirable between runs, but when running `$ ava --update-snapshots`, I expected it to update the snapshot to the reflect the current state of the tests file, removing non-existing/renamed tests. Is there any benefit to keeping non-existing tests around after updating the snapshot? I don't really see it, as we already have git, but I might be missing something fundamental about how it works.

lithin:
@sindresorhus I think you're absolutely right :) Might be a good suggestion to the jest repo but I'm afraid I can't change that through my implementation of it.

sindresorhus:
@lithin Alright. I'll open an issue on Jest when I have some time :) This looks ready to me, but the tests are failing on Windows: https://ci.appveyor.com/project/ava/ava/build/1.0.462/job/jnx6609ubg58gfcq Mind taking a look?

lithin:
Tests passed! :)

sindresorhus:
Yay! Super excited about this. Thank you so much for working hard on this @lithin. We really appreciate it. ✨🎉🦄🚀🌸

lpil:
Great work, thanks @lithin & @sindresorhus

vadimdemedes:
This is amazing, thank you, Anna!

giphy 8

P.S. I post this gif only on really cool stuff :)

lithin:
Hahaha thank you @vadimdemedes ! It was a pleasure to contribute to a project like Ava :)

Issue #1109, opened 2016-11-15
Support to Run TypeScript test file(.ts) directly without Precompiling(tsc)


huan:
[<img alt="Issuehunt badges" src="https://issuehunt.io/r/avajs/ava/issues/1109/badge.svg" />](https://issuehunt.io/r/avajs/ava/issues/1109)

## Description

AVA now support ES6 by using Babel, run directly without any precompiling, this is very convenience. But for TypeScript, AVA has to precompile ts files by run `tsc`.

I use TypeScript in my project wechaty, run TS code directly by `ts-node` like this: `ts-node demo.ts`.  This is a convenience way for developing.

If we can run TypeScript file directly by AVA, just like `ava test.ts`, that will be wonderful.

About this topic, there has already an issue #631 on it. I saw @sindresorhus agree that use TypeScript without precompiling is a better way.

> @niieani You can already use TypeScript by precompiling it. I know it's not optimal, but it's not like anyone has done a pull request fixing this. https://github.com/avajs/ava/issues/631#issuecomment-250368210

I love AVA, because I'm a TAP fan, and AVA is based on TAP. So I tried to hack AVA to support TypeScript directly. After some study & work, I finished it finally, and in the past month, I was using it in my own project and it worked as my expected.

I'll share my idea here, and send a Pull Request later. Hope we can support TypeScript without precompiling soon.

![ava-typescript](https://cloud.githubusercontent.com/assets/1361891/20292192/98b13bb6-ab26-11e6-8ea2-f833031da6d7.gif)

### About the Hack

I added a `--ext ts` argument parameter to CLI, then pass `ext` to AVA API. If ext is `ts` then I use TypeScript Compiler to replace the Babel Compiler, because TypeScript is the superset of ES6/7.

1. Add CLI Arg: `--ext`/`-e`

    In order to support TypeScript without precompiling, first, we need to add a CLI args to support that. I use the name `--ext` and the alias `-e` for this, like `ava -e ts`. Set the ext automatically is considered, but I think maybe it should be done later. If set ext automatically, then we can just run TS tests by `ava test.ts`

1. Get Test Files from `ava-files`

    AVA use a submodule `ava-files` to get all test files list. It had to be modified to support the `ext` parameter. I use `exts` in the ava-files module and it is an `Array` type, that's for the future support of multiple file extensions.

1. new API options: `ext`

    in API call, we get file extension name from `options.ext`, pass it to `AvaFiles`/`CachingPrecompiler`/`Watcher`.

[...]

PR #1122, opened 2016-11-21
Add support for TypeScript without precompiling

zixia:
Follow #1109 

Notice that this PR has to work together with https://github.com/avajs/ava-files/pull/9

So unit testing in CI will expect failure before we merge&update `AvaFiles` NPM.

In my development environment with `npm link ava-files`, all unit tests had passed.


package.json
```diff
@@ -184,13 +184,16 @@
...
+    "typescript": "^2.0.3",
     "xo": "^0.17.0",
     "zen-observable": "^0.3.0"
   },
   "xo": {
     "rules": {
       "import/newline-after-import": "off",
-      "no-use-extend-native/no-use-extend-native": "off"
+      "no-use-extend-native/no-use-extend-native": "off",
+      "no-useless-escape": "off"
```
sindresorhus: Fix the issue instead.

test/fixture/typescript/simple.ts
```diff
@@ -0,0 +1,6 @@
+import { test } from 'ava'
```
sindresorhus:
```js
import {test} from 'ava';
```

And use semi-colons.

package.json
```diff
@@ -89,7 +89,7 @@
     "arr-flatten": "^1.0.1",
     "array-union": "^1.0.1",
     "array-uniq": "^1.0.2",
-    "arrify": "^1.0.0",
+    "arrify": "^1.0.1",
```
sindresorhus: Don't do unrelated changes.

package.json
```diff
@@ -154,7 +154,7 @@
     "resolve-cwd": "^1.0.0",
     "semver": "^5.3.0",
     "set-immediate-shim": "^1.0.1",
-    "source-map-support": "^0.4.0",
+    "source-map-support": "^0.4.5",
```
sindresorhus: Why?

lib/ext/ts.js
```diff
@@ -0,0 +1,116 @@
+'use strict';
+/**
+ * AVA File Extention for typescript
```
sindresorhus: Extention => Extension

typescript => TypeScript

lib/ext/ts.js
```diff
@@ -0,0 +1,116 @@
...
+ * Author: https://github.com/zixia
+ */
+var fs 					= require('fs');
```
sindresorhus: 🐐

lib/ext/ts.js
```diff
@@ -0,0 +1,116 @@
...
+
+var typescript;
+try {
+	typescript = require('typescript');
+} catch (err) {
+	console.log('AVA error: `--extensions ts` require typescript installed. Please run `npm install typescript` first.');
```
sindresorhus:
```
AVA error: `--extensions=ts` require TypeScript to be installed.
```

lib/ext/ts.js
```diff
@@ -0,0 +1,116 @@
...
+	throw err;
+}
+var tsnode; // = require('ts-node'); XXX: maybe useful in the future, because tsnode do more than typescritp for our case.
```
sindresorhus: typo in comment

lib/test-worker.js
```diff
@@ -30,6 +30,20 @@ process.installPrecompilerHook();
...
 
+if (opts.extensions) {
+	opts.extensions.forEach(function (ext) {
+		switch (ext) {
+			case 'ts':
+				var extTs = require('./ext/ts');
+				extTs.register();
+				break;
+			default:
+				// not supported ext?
```
sindresorhus: It should throw an useful error here.

lib/watcher.js
```diff
@@ -58,9 +59,13 @@ function Watcher(logger, api, files, sources) {
 		}
 
 		var self = this;
-		this.busy = api.run(specificFiles || files, {
+		var options = {
 			runOnlyExclusive: runOnlyExclusive
-		}).then(function (runStatus) {
+		};
+		if (extensions) {
```
sindresorhus: This is an internal API. We can just make passing the `extensions` argument required and drop this if-statement.

lib/cli.js
```diff
@@ -100,6 +103,11 @@ exports.run = function () {
 		throw new Error(colors.error(figures.cross) + ' The --require and -r flags are deprecated. Requirements should be configured in package.json - see documentation.');
 	}
 
+	if (cli.flags.extension) {
+		cli.flags.extension = arrify(cli.flags.extension);
+		cli.flags.extension.push('js');
```
sindresorhus: As we discussed in https://github.com/avajs/ava-files/pull/9#issuecomment-261962080, the option should override `js`.

lib/cli.js
```diff
@@ -43,6 +43,7 @@ exports.run = function () {
 		'  --source, -S       Pattern to match source files so tests can be re-run (Can be repeated)',
 		'  --timeout, -T      Set global timeout',
 		'  --concurrency, -c  Maximum number of test files running at the same time (EXPERIMENTAL)',
+		'  --extension, -e    Set test file extention. `ts` is for Typescript',
```
sindresorhus:
```
Test file extension, for example `ts` for TypeScript (Can be repeated)
```

lib/caching-precompiler.js
```diff
@@ -48,8 +49,32 @@ CachingPrecompiler.prototype._init = function () {
...
+		switch (fileExt) {
+			case 'ts':
+				var extTs = require('./ext/ts');
+				result = extTs.compile(code, filePath);
+				break;
+
+			// here for other exts in the future
+
+			default:
+				// not supported(yet)
```
sindresorhus: It should throw an useful error here.

package.json
zixia: I just run an `npm install arrify source-map-support --save` by mistake

Restored old version.

package.json
zixia: Ok. fixed.

but this part looks to be a little weird:

```js
  var isBabelPath = /^babel-runtime[\\/]?/.test(path.node.value);
```

Ref: http://stackoverflow.com/a/40315390/1123955

And a question: How did the AVA pass the CI test? I just clone the AVA repo, run `npm install && npm t`, XO will throw `no-useless-escape` error.

lib/ext/ts.js
zixia: fixed

lib/ext/ts.js
zixia: :goat:

lib/ext/ts.js
zixia: done

lib/ext/ts.js
zixia: fixed

lib/watcher.js
zixia: done.

lib/caching-precompiler.js
zixia: done.

throws the same error as the above.

package.json
sindresorhus: Yeah, that's weird, but just fix it.

sindresorhus:
I don't think you pushed your changes.

Merge conflict also needs fixing.

zixia:
@sindresorhus Yes, I'm still fighting with the unit tests.

Because of the new `extensions` optoins parameter, some unit tests broken.

Will push after I fixed them.

zixia:
Just finished all the unit tests with the latest code in this PR branch.

`ava-files` has updated PR to latest version as well.

Please let me know if I missed anything. :)

lib/cli.js
jedmao: Couldn't you just do this?
```js
cli.flags.extension = arrify(cli.flags.extension || 'js');
```

zixia: @jedmao great idea, thanks!

.gitignore
```diff
@@ -3,3 +3,4 @@ node_modules
 coverage
 bench/.results
 types/generated.d.ts
+yarn.lock
```
vdemedes: Extra change.

lib/caching-precompiler.js
```diff
@@ -47,8 +48,33 @@ CachingPrecompiler.prototype._init = function () {
...
+
+	if (isCustomExt) {
+		switch (fileExt) {
+			case 'ts':
+				var extTs = require('./ext/ts');
+				result = extTs.transpile(code, filePath);
+				break;
+
+			/**
+			 * here for other extensions in the future
```
vdemedes: Comment should be removed.

lib/caching-precompiler.js
```diff
@@ -47,8 +48,33 @@ CachingPrecompiler.prototype._init = function () {
...
-	var result = this.babel.transform(code, options);
+	var fileExt = path.extname(filePath).replace(/^\./, '');
+	var isCustomExt = this.extensions.filter(function (ext) {
+		return (ext !== 'js' && ext === fileExt);
+	}).length > 0;
+
+	var result;
+
+	if (isCustomExt) {
+		switch (fileExt) {
```
vdemedes: I'm not a fan of switch statements, simple if here could do just as well. @sindresorhus ?

lib/cli.js
```diff
@@ -43,6 +43,7 @@ exports.run = function () {
 		'  --source, -S       Pattern to match source files so tests can be re-run (Can be repeated)',
 		'  --timeout, -T      Set global timeout',
 		'  --concurrency, -c  Maximum number of test files running at the same time (EXPERIMENTAL)',
+		'  --extension, -e    Test file extension, for example `ts` for TypeScript (Can be repeated)',
```
vdemedes: I think example here is extra. Or it could at least be shortened to:

```
Test file extension (e.g. `coffee`, `ts`)
```

lib/cli.js
```diff
@@ -100,6 +103,8 @@ exports.run = function () {
 		throw new Error(colors.error(figures.cross) + ' The --require and -r flags are deprecated. Requirements should be configured in package.json - see documentation.');
 	}
 
+	var extensionList = [].concat(cli.flags.extension || 'js');
```
vdemedes: `extensions` might be a better name.

lib/ext/ts.js
```diff
@@ -0,0 +1,111 @@
+'use strict';
+/**
```
vdemedes: Comment should be removed.

lib/ext/ts.js
```diff
@@ -0,0 +1,111 @@
...
+var fs = require('fs');
+
+var typescript;
+try {
+	typescript = require('typescript');
+} catch (err) {
+	console.log('AVA error: `--extensions=ts` require TypeScript to be installed.');
+	throw err;
+}
+var tsnode; // = require('ts-node'); XXX: maybe useful in the future, because ts-node do more than typescript for our case.
```
vdemedes: Comment should be removed.

vdemedes:
Hey @zixia, thanks for all this hard work you've done!

I was reviewing the code, but found this file https://github.com/avajs/ava/pull/1122/files#diff-f9bf36e65f4c439ede43fea80500f031 to have a lot of comments. Is it still work in progress?

.gitignore
zixia: deleted.

lib/caching-precompiler.js
zixia: removed.

lib/cli.js
zixia: shorted.

lib/cli.js
zixia: I use `extensions` at first. I try to make the name more clear by changed it to `extensionList` .

However, I had followed your suggestion, renamed it back to `extensions`.

lib/ext/ts.js
zixia: removed.

lib/ext/ts.js
zixia: removed.

zixia:
Hey @vdemedes , 

I had just followed all your comments. 

About the comments in `lib/ext/ts.js`, they are just comments. I also had made this file clearer with new code.

Thanks for your reviews.

lib/cli.js
sindresorhus: I agree it could be shortened, but not like that.

Should be:

```
Test file extension (e.g. `ts`) (Can be repeated)
```

sindresorhus:
https://github.com/avajs/ava/pull/1122#discussion_r89947713

lib/caching-precompiler.js
sindresorhus: @vdemedes Agreed. I usually don't use switch unless there are lots of same if's.

lib/ext/ts.js
```diff
@@ -0,0 +1,89 @@
+'use strict';
+var fs = require('fs');
+
+var typescript;
+try {
+	typescript = require('typescript');
+} catch (err) {
+	console.log('AVA error: `--extensions=ts` require TypeScript to be installed.');
```
sindresorhus: Use console.error

`AVA error:` => `AVA:`

lib/test-worker.js
```diff
@@ -30,6 +30,23 @@ process.installPrecompilerHook();
...
+			case 'ts':
+				var extTs = require('./ext/ts');
+				extTs.register();
+				break;
+			case 'js':
+				// We already have babel for ES6/7
+				break;
+
+			default:
+				throw new Error('`--extension ' + ext + '` is not supported by AVA.');
```
sindresorhus: `--extension '` => `--extension='`

test/api.js
```diff
@@ -56,6 +56,29 @@ test('Without Pool: test files can be forced to run in exclusive mode', function
...
+test('TypeScript supported by set `extensions`', function (t) {
+	var api = new Api({extensions: ['ts']});
+	return api.run(
+		[path.join(__dirname, 'fixture/typescript/simple.ts')]
+	).then(function (result) {
+		t.is(result.testCount, 1);
+		t.is(result.passCount, 1);
+		t.is(result.failCount, 0);
+	});
+});
```
sindresorhus: Needs a test to ensure it support reading user TS config file.

lib/watcher.js
```diff
@@ -59,7 +60,8 @@ function Watcher(logger, api, files, sources) {
 
 		var self = this;
 		this.busy = api.run(specificFiles || files, {
-			runOnlyExclusive: runOnlyExclusive
+			runOnlyExclusive: runOnlyExclusive,
+			extensions: extensions
```
sindresorhus: Needs a watcher test for TypeScript.

lib/ext/ts.js
```diff
@@ -0,0 +1,89 @@
...
+		compilerOptions: typescriptOptions,
+		reportDiagnostics: true
+	});
+	return {
+		code: result.outputText,
+		map: result.sourceMapText
+	};
+}
+
+var extTs = {
```
sindresorhus: Just export this directly.

lib/ext/ts.js
```diff
@@ -0,0 +1,89 @@
...
+			compilerOptions: typescriptOptions,
+			reportDiagnostics: true
+		});
+		module._compile(result.outputText, fileName);
+	};
+
+	require('source-map-support').install({
+		hookRequire: true
+	});
+}
```
sindresorhus: Doesn't `ts-node` already do this?

lib/ext/ts.js
```diff
@@ -0,0 +1,89 @@
...
+	sourceMap: true,
+	inlineSources: true,
+	inlineSourceMap: true
+};
+
+/**
+ *
+ * Ts-Node version
+ *
+ */
```
sindresorhus: Remove comment.

lib/cli.js
zixia: done.

lib/ext/ts.js
zixia: done.

lib/test-worker.js
zixia: done.

lib/ext/ts.js
zixia: done.

lib/ext/ts.js
zixia: removed

lib/ext/ts.js
zixia: yes and no.

ts-node will do this if we use transpile function from ts-node. However, the current version of lib/ext/ts.js is not using the ts-node by default, it use the raw typescript module, so we have to do this by ourselves.

I leave the ts-node in lib/ext/ts.js is because at the beginning I'm not sure which one is better, so I implement both, but at last I decided to use pure typescript by default:

```js
module.exports = module.exports.default = {
  register: typescriptRegister,
  transpile: typescriptTranspile,
```

Do you think we should use ts-node, or I need to get rid of the additional ts-node code in lib/ext/ts.js?

lib/caching-precompiler.js
```diff
@@ -47,8 +48,29 @@ CachingPrecompiler.prototype._init = function () {
 CachingPrecompiler.prototype._transform = function (code, filePath, hash) {
 	code = code.toString();
 
-	var options = babelConfigHelper.build(this.babelConfig, this.powerAssert, filePath, code);
-	var result = this.babel.transform(code, options);
+	var fileExt = path.extname(filePath).replace(/^\./, '');
+	var isCustomExt = this.extensions.filter(function (ext) {
+		return (ext !== 'js' && ext === fileExt);
+	}).length > 0;
```
novemberborn: This could be simplified to:

```js
var isCustomExt = fileExt !== 'js' && this.extensions.indexOf(fileExt) > -1
```

lib/caching-precompiler.js
```diff
@@ -47,8 +48,29 @@ CachingPrecompiler.prototype._init = function () {
...
+				var extTs = require('./ext/ts');
+				result = extTs.transpile(code, filePath);
+				break;
+
+			default:
+				throw new Error('`--extension ' + fileExt + '` is not supported by AVA.');
+		}
+	}
+
+	if (!result) {
```
novemberborn: Should just be an `else`, assuming the body for `if (isCustomExt)` assigns `result` or throws.

lib/ext/ts.js
```diff
@@ -0,0 +1,83 @@
+'use strict';
+var fs = require('fs');
+
+var typescript;
+try {
+	typescript = require('typescript');
+} catch (err) {
+	console.error('AVA: `--extensions=ts` require TypeScript to be installed.');
+	throw err;
```
novemberborn: Rather than logging and then rethrowing the original error construct a new error like in https://github.com/avajs/ava/blob/0f0400158ec4192fc32ccc603877d0b088302662/lib/cli.js#L99 and throw that. It'll be logged correctly.

lib/ext/ts.js
novemberborn: Looks like `ts-node` is closer to `babel-node` than `babel-core` (and `ts-node/register` is equivalent to `babel-core/register`). IMO this module should just use `typescript` to transpile the code, and no other functionality is required.

For consistency we should not automatically transpile source files, at least until we have a better answer for the `.js` use case.

lib/caching-precompiler.js
```diff
@@ -47,8 +48,29 @@ CachingPrecompiler.prototype._init = function () {
 CachingPrecompiler.prototype._transform = function (code, filePath, hash) {
 	code = code.toString();
 
-	var options = babelConfigHelper.build(this.babelConfig, this.powerAssert, filePath, code);
-	var result = this.babel.transform(code, options);
+	var fileExt = path.extname(filePath).replace(/^\./, '');
```
zixia: done.

lib/caching-precompiler.js
zixia: done.

lib/caching-precompiler.js
zixia: done.

lib/ext/ts.js
zixia: done.

zixia:
@novemberborn I love ES2015 too. Had just followed your suggestions and resolved all the conflicts.

lib/caching-precompiler.js
novemberborn: @zixia could you change this to an `if` statement?

lib/caching-precompiler.js
```diff
@@ -47,8 +48,25 @@ CachingPrecompiler.prototype._init = function () {
...
+
+	if (isCustomExt) {
+		switch (fileExt) {
+			case 'ts':
+				var extTs = require('./ext/ts');
+				result = extTs.transpile(code, filePath);
+				break;
+
+			default:
+				throw new Error('`--extension ' + fileExt + '` is not supported by AVA.');
```
novemberborn: Let's check for allowed extensions in [`lib/cli` instead](https://github.com/avajs/ava/blob/0f0400158ec4192fc32ccc603877d0b088302662/lib/cli.js#L104).

lib/ext/ts.js
novemberborn: @zixia per the above discusison, could you simplify this module to just the `transpile` function?

lib/ext/ts.js
```diff
@@ -0,0 +1,85 @@
...
+		compilerOptions: typescriptOptions,
+		reportDiagnostics: true
+	});
+	return {
+		code: result.outputText,
+		map: result.sourceMapText
+	};
+}
+
+module.exports = module.exports.default = {
```
novemberborn: Just use `exports.transpile = typescriptTranspile`.

lib/test-worker.js
```diff
@@ -30,6 +30,23 @@ process.installPrecompilerHook();
...
+				break;
+			case 'js':
+				// We already have babel for ES6/7
+				break;
+
+			default:
+				throw new Error('AVA: `--extension=' + ext + '` is not supported by AVA.');
+		}
+	});
+}
```
novemberborn: Please remove these changes. Right now I'd prefer our TypeScript support is consistent with our Babel support, where dependencies are not automatically transpiled.

I know that's not ideal but we can revisit it later.

lib/caching-precompiler.js
zixia: ok. done.

lib/ext/ts.js
zixia: Hi @novemberborn ,

I had simplified this module, removed all the ts-node parts, cleaned the code, but not just the transpile.

The `register()` function has to be kept because if we run test written in TypeScript, the lib/test-worker.js must has the ability to transpile TypeScript, which is enabled by `extTs.register()` at here: https://github.com/zixia/ava/blob/typescript/lib/test-worker.js#L38

lib/ext/ts.js
zixia: I think we need to keep `register()` with `tranpile()`, as I described here: https://github.com/avajs/ava/pull/1122#discussion_r91015553

If there has any alternative way to do this, please let me know, thanks.

lib/caching-precompiler.js
jedmao: What about `.tsx` for TypeScript/React?

jedmao: Could we do the following to cover both `.ts` and `.tsx` files?
```js
if (/^tsx?$/.test(fileExt))
```

zixia: I use `.ts` only and I do not know anything about react(I use angular). 

Could TypeScript transpile .tsx by default?

lib/test-worker.js
zixia: I don't think the TypeScript support would still work after we removed those changes. Because if no `extTs.register()`, then test-worker.js will not work with .ts test files.

This TypeScript support will only be enabled when we specify `--extension ts` arg with AVA, which is the case we need to automatically transpile all TypeScript file with extension: `.ts`

And about the dependencies, I think for TypeScript is different as for JavaScript, because:
1. There are almost no TypeScript dependencies in NPM(node_modules)
1. The `.ts` extension is not the default JavaScript file extension, so it's almost only in our development environment.

Please correct me if I'm wrong, or missed anything. Thanks.

zixia:
Thank you guys for all the replies of this PR. 

Please let me know if I missed any important replies because this conversation is growing fast. ;-p

novemberborn:
@zixia could you explain why you think `register` is needed in the workers? The `CachingPrecompiler` already compiles the test files, and the workers load the resulting code. On-the-fly compilation would only apply to files required by the test file. I'm arguing that, though not ideal, it's better to be consistent with our Babel implementation and not automatically transpile such dependencies. Thus, `register` would not be needed.

> Where are the allowed extensions in cli.js? Did you mean we should save all supported extensions to an array in cli.js, then reference it in other js files?

There are none, currently, since `.js` is hardcoded in a few places. I'm suggesting you check for `.ts` after the `hasFlag` tests, before `API` is instantiated. [Around here](https://github.com/avajs/ava/blob/0f0400158ec4192fc32ccc603877d0b088302662/lib/cli.js#L104).

Zooming out a bit, `CachingPrecompiler` was really written for Babel. It uses the Babel config to [salt the cache](https://github.com/zixia/ava/blob/41043e74219974a70e48c0ad0e8b88b0839fd9c3/lib/caching-precompiler.js#L88) and it [automatically loads Babel](https://github.com/zixia/ava/blob/41043e74219974a70e48c0ad0e8b88b0839fd9c3/lib/caching-precompiler.js#L43) when it needs to compile something. Your proposed changes still cause the Babel code paths to execute and don't allow the TypeScript compiler to be configured. I think that's OK at this stage but we'll have to solve those issues at a later stage. @avajs/core thoughts?

---

@jedmao perhaps we should revisit `.tsx` once this has landed? There's a parallel with `.jsx` which currently doesn't work either. Maybe we can treat these as aliases and just assume the existing `.js` / `.ts` setup supports JSX. It's complicated though because JSX doesn't necessarily mean React.

zixia:
@novemberborn got it.

The reason I think `register` is needed in the workers is because if I do not do this, my test will not work anymore. 

I believe that's because my tests have to import pure TypeScript module file: there is all kind of TypeScript files in my source tree, and I never compile them, just `import` them in my tests.

Will check them and reply you with the details later.

novemberborn:
@zixia:

> The reason I think register is needed in the workers is because if I do not do this, my test will not work anymore.
> 
> I believe that's because my tests have to import pure TypeScript module file: there is all kind of TypeScript files in my source tree, and I never compile them, just import them in my tests.
> 
> Will check them and reply you with the details later.

Right, that's the same issue we have with our Babel support: source files are not automatically transpiled so users need to add `babel-core/register` to AVA's `require` configuration.

This is by no means ideal, but automatically including `babel-core/register` has a measurable impact on performance for those users who do not need it. I imagine the same applies to TypeScript.

I've been playing with performance optimizations in this area (e.g. https://github.com/avajs/ava/pull/1082). Really I'd like to have users opt-in to a particular behavior for their tests and source files, and then have everything be automatic.

For now though I'm suggesting we *do not* automatically load TypeScript when running tests. The behavior would be the same as for users relying on `babel-core/register`.

This PR already introduces many subtle changes to AVA's behavior that should be factored out. That's a larger task and I think this PR provides enough value as it stands today, so we should try and land it. But not with automatic TypeScript loading.

zixia:
Hi @novemberborn ,

Got your idea, and thanks for the explanations.

> There are none, currently, since .js is hardcoded in a few places. I'm suggesting you check for .ts after the hasFlag tests, before API is instantiated. Around here.

done.

> Right, that's the same issue we have with our Babel support: source files are not automatically transpiled so users need to add babel-core/register to AVA's require 

done.

I had removed `extTs.register()` part from lib/test-worker.js, and add the following json part to package.json, TypeScript work without any problem again.

```json
  "ava": {
    "require": [
      "ts-node/register"
    ]
  },
```

For more about TypeScript register, here's some discussions about it: https://github.com/Microsoft/TypeScript/issues/1823

Have a good weekend!

zixia:
I found a strange unit test failure in test/cli.js at here: https://github.com/zixia/ava/blob/d7bfa76fb708b6d918f1c821df0417d79e96cde4/test/cli.js#L383

We run `cli.js` in `cwd`, however, we pass `resolveTestsFrom` to AvaFiles at [here](https://github.com/zixia/ava/blob/d7bfa76fb708b6d918f1c821df0417d79e96cde4/api.js#L87), which is `/tmp/` instead of `/tmp/xxxx/`, as the result, test fail.

I have no idea of it because I did not touch this part this week and everything is ok before. Will check it later.

asvetliakov:
Any news about PR ? I'd like to adopt ava in my project, but separate TS precompilation (and as i understand, the ava now just doesn't work with file extensions rather than .js) is huge stopper for me.

zixia:
@asvetliakov Before this PR be merged, you can try this new feature by clone the source with `npm link`:

```shell
$ git clone https://github.com/zixia/ava-files.git
$ cd ava-files
$ npm install && npm link
$ cd ..

$ git clone https://github.com/zixia/ava.git
$ cd ava
$ git checkout typescript
$ npm install & npm link ava-files && npm link
```

Then you can use ava with TypeScripe in your project like:

```shell
$ cd my-project
$ npm link ava
$ ./node_modules/.bin/ava --extension=ts
```

Please let me know if there's any issue with your project if you have time to try it.

Good luck! :)

novemberborn:
Hi @zixia, I have some concerns with the direction we're taking here.

As already mentioned it's not clear how this would work for source files without running into the same performance issues we already see with Babel. The TypeScript test files won't get enhanced assertions or protection against improper `t.throws()` usage either. It's hard to communicate this to users when the way TypeScript support is enabled is to specify an `extensions` option.

I've been doing a lot of thinking on how we can better solve these problems. Not only for TypeScript but for Babel-based projects as well. I've started writing an RFC with my specific proposal but I'm not ready to share it yet. I'm currently on holiday, and the weather is awfully nice (except when there's ⛈ like right now), so I can't commit to any particular publication date.

I didn't want to leave you hanging though. As I see it right now I don't think we can land this pull request. I'm really sorry about that, since you've put in a lot of work already.

I hope you'll give my proposal a chance. It should enable us to make AVA's TypeScript support really awesome.

Thanks.

zixia:
@novemberborn Understand, and please let me know when you made any progress on your RFC. :)

Have a nice holiday!

novemberborn:
@zixia see #1159.

novemberborn:
We'll be taking a different direction with regards to TypeScript support, see [RFC001](https://github.com/avajs/ava/blob/master/docs/specs/001%20-%20Improving%20language%20support.md). @zixia I'll let you know when the Babel side of that has stabilized sufficiently for you (or somebody else) to take on the TypeScript implementation.

zixia:
@novemberborn No problem, looking forward hearing from you.

Issue #736, opened 2016-04-07
Harmonize test file & source pattern handling


novemberborn:
Handling of test file patterns should be harmonized with how source patterns are handled (#614):
- Use the default patterns if `files` is empty, _or_ if it only contains negated patterns
- Apply specified negated patterns together with the default exclusion patterns
- Allow specified patterns to override the default exclusion patterns

The default exclusion patterns for source files are relative to the current working (`node_modules/**/*`) however the test file exclusion patterns can apply anywhere (`**/node_modules/**`). I guess that makes sense, though I'm not sure how to override the default exclusion patterns. Maybe that's not necessary for test files?

Should we change the source exclusion patterns? Or do those make more sense relative to the current working directory?

---

Implementation wise I think we can move the file selection tests in `test/api.js` to a new `test/ava-files.js` test file. We can set up a fixture directory to exercise the globbing behavior.

jamestalmage:
> The default exclusion patterns for source files are relative to the current working (`node_modules/**/*`) however the test file exclusion patterns can apply anywhere (`**/node_modules/**`)

I think the source file pattern should be made to match the test file pattern.

> I guess that makes sense, though I'm not sure how to override the default exclusion patterns. Maybe that's not necessary for test files?

Aren't you doing exactly that (overriding the defaults) in #614?

novemberborn:
> I think the source file pattern should be made to match the test file pattern.

Yea maybe. [Here's the list of source excludes](https://github.com/novemberborn/ignore-by-default/blob/master/index.js). `coverage` may be tricky.

>> I guess that makes sense, though I'm not sure how to override the default exclusion patterns. Maybe that's not necessary for test files?

> Aren't you doing exactly that (overriding the defaults) in #614?

Yes, but that's based on the path prefix. So `node_modules/whatever` maps to `node_modules/**/*`. That's trickier when the exclude patterns are recursive (`**/node_modules/**`). Now every path component needs to be checked. Though I guess that answers my question :)

Maybe we should move all this manipulation code into an external library.

jamestalmage:
> `coverage` may be tricky.

Why?

novemberborn:
@jamestalmage it's a fairly generic name when ignoring deeply nested directories. I doubt anybody would put legitimate code in say `.sass-cache`.

novemberborn:
> Allow specified patterns to override the default exclusion patterns

[...]

PR #1136, opened 2016-12-01
fails running test in node_modules

mastilver:
I've got a weird use case where I want to run test files inside `node_modules`: I want to share test files between several modules

Is it something that you want to handle?

### DIFF
--- a/test/cli.js
+++ b/test/cli.js
@@ -400,3 +400,10 @@ test('workers ensure test files load the same version of ava', t => {
 		t.end();
 	});
 });
+
+test('run test in node_modules', t => {
+	execCli([], {dirname: 'fixture/test-in-node-modules'}, err => {
+		t.ifError(err);
+		t.end();
+	});
+});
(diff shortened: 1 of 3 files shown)


novemberborn:
Does specifying a `./node_modules/shared-tests` as a test pattern work? (I'm not sure, and I don't have a dev environment on the laptop I'm tying this from.)

Does our [macro support](https://github.com/avajs/ava#test-macros) help your use case?

What kind of tests are you sharing?

mastilver:
For this package: https://github.com/mastilver/nosqwal/tree/master/packages/nosqwal-memory
inside `package.json`, I set: `ava.files` to `["./node_modules/nosqwal-test/test"]`

I believe it's an intended behavior given this test:  https://github.com/avajs/ava/blob/58a2e6803f74fa29d601e019b22f6a0b945c3544/test/api.js#L595

I would be happy this #1137 being solved :)
I guess you can close this one

novemberborn:
@mastilver ah, that traces back to https://github.com/avajs/ava/issues/226. https://github.com/avajs/ava/issues/736 discusses further changes to those patterns. IMO your setup should work, exotic though it is.

Closing in favor of #736.

PR #1140, opened 2016-12-02
fix vadimdemedes github link

SamVerschueren:
Noticed that @vadimdemedes changed his github name :).


vadimdemedes:
Thank you, Sam :)

PR #1145, opened 2016-12-06
Support symlinked test files - fixes #1143

### LINKED ISSUE #1143: AVA fails when running test files which are actually symlinks
Per #1137, if a test pattern matches a symlink, AVA fails to run that test file.

AVA precompiles the test file in the main process. That works fine, even with symlinks. It then communicates a hash for the precompilation result to the worker process. When the test file is required the worker loads it from the precompilation cache, using the hash received from the main process.

The problem is that the hash mapping is for the symlink path, but the worker sees the _resolved_ path instead.

This is where the worker looks up the hash: https://github.com/avajs/ava/blob/0f0400158ec4192fc32ccc603877d0b088302662/lib/process-adapter.js#L85

This is where the `precompiled` hash mapping is prepared: https://github.com/avajs/ava/blob/0f0400158ec4192fc32ccc603877d0b088302662/api.js#L70

Probably the best fix is to use [`fs.realpathSync`](https://nodejs.org/api/fs.html#fs_fs_realpathsync_path_options) when preparing the hash mapping in `api.js`.

rnkdev:
Fixes #1143 

Following from the bug thread, I modified api.js hashmap preparation so that it uses resolved path as the key. I also added one test to ensure that ava can now be ran with linked test file.


rnkdev:
Hmm it seems that it fails the test for Windows. I'm not sure how to handle the symlink in it.

rnkdev:
After some research, I think what I could do is I can add check to the test not to be run (or to be automatically passed) if the environment is detected to be Windows. But I'm not sure if it's the right approach considering Windows has no symlink support.

rnkdev:
Thanks for adding that support for Windows!

On the other hand, there is a new problem, when I added new direct symlink to a test file in the test, it fails to recognise that file exist. 

It throws `AvaError: Couldn't file any files to test`strangely. Normally you would think if it can resolve a symlink to a directory, it should resolve symlink to a file as well. 

The following commit shows the problem. It stops before even calling precompilation and `Api._runFile`

rnkdev:
So I copied the same tests to the ava master branch. Surprisingly it failed all of the tests including the one that points to directory. 

In a sense this fix fixes problem #1143 as the user is trying to do a symlink to a directory containing test files. But then it also uncovers that direct symlink is somehow can't be resolved. (Maybe it has something to do with how files are searched in ava-files)

novemberborn:
> On the other hand, there is a new problem, when I added new direct symlink to a test file in the test, it fails to recognise that file exist.

You need to make sure it has the `.js` extension.

Let's keep the test for symlinked directories, but also add one for symlinked files.

rnkdev:
Hi I just added the last commit with the requested change. (My bad for not researching that .js part) Is there anything else I should do to get the PR accepted? Thanks.

sindresorhus:
Looks good to me too. Nice work @rnkdev :)

mastilver:
@rnkdev Thank you so much for your work :+1:

PR #1154, opened 2016-12-12
Magic assert

### LINKED ISSUE #1155: Magic assert
AVA team is extremely excited to announce the new & shiny AVA feature. We call it "Magic assert".

We could try and describe `magic-assert` and all its features, but it's faster and better to let you see what it is all about right away.

### Strings

```js
test('strings', t => {
  const actual = 'this is so cool';
  const expected = 'this is so bad';

  t.is(actual, expected);
});
```

### Different value types

```js
test('different types', t => {
  const actual = 'this is so cool';
  const expected = {msg:'this is so bad'};

  t.is(actual, expected);
});
```

### Objects and arrays

```js
test('objects', t => {
  const actual = {
    a: 1,
    b: {
      c: 2
    }
  };

  const expected = {
    a: 2,
    b: {
      c: 1
    }
  };

  t.deepEqual(actual, expected);
});
```

### t.true/t.false assertions

These kind of assertions obviously can't have diffs, because they're being compared to `true`/`false`. Having "actual: true, expected: false" output wouldn't be much of a help, so instead we display the value of each statement in the assertion.

```js
test('nested', t => {
  const actual = {a:[1]};
  
  t.true(actual.a[0]);
});
```

---

Now that we got you all excited (hopefully), few words about what it actually is. Magic assert is not an assertion library, but a feature that *handles* assertion errors and is responsible for displaying them in a clean yet informative way. It replaces `power-assert` output, although it still uses its core for some cases.

Magic assert adds code excerpts pointing to the source of the error and clean colorful diffs to each error report. You can also spot there's now more spacing between each error, which makes it easier to go through all the errors.

Your tests don't need to be modified in any way to take support `magic-assert`'s output. If you use other assertion library like expect/chai or other, you are in luck too - you will get the same magical output!

**There's one more thing...**

### React support

Magic assert PR also lands two new assertion methods to improve React support: `t.jsxEquals()` and `t.jsxNotEquals()`. And as you may have expected, it provides a nice output of the "React tree" as well!

```js
test('react', t => {
  const actual = <div className="body">Hello</div>;
  const expected = <div className="header">Hello</div>;

  t.jsxEqual(actual, expected);
});
```

screen shot 2016-12-12 at 11 19 57 pm

I'm not sure whether this change should be in this or separate PR - feedback welcome here too.

### Credits

Magic assert was made possible by these amazing modules:

- [power-assert](https://www.npmjs.com/package/power-assert) by Takuto Wada (@twada).
- [diff](https://www.npmjs.com/package/diff) by Kevin Decker (@kpdecker).
- [pretty-format](https://www.npmjs.com/package/pretty-format) by James Kyle (@thejameskyle).

Huge thanks!

---

We think it is a big step forward towards more comfortable debugging of failed tests and developer happiness in general.

You can try Magic assert by using PR https://github.com/avajs/ava/pull/1154 today!

We'd love to hear your feedback and ideas on how to improve the output of Magic assert! Thank you!

vadimdemedes:
Fixes #1155.

Todo:

- [x] Test `t.jsxEqual()` and `t.jsxNotEqual()`
- [x] Test `lib/code-excerpt.js`
- [x] Test `lib/serialize-error.js`
- [x] Test `lib/format-assert-error.js`
- [x] Update mini reporter
- [x] Update verbose reporter
- [x] Update mini reporter tests
- [x] Update verbose reporter tests
- [x] Skip output for `t.fail()`
- [x] Make snapshots compatible
- [x] Resolve TODOs in code
- [x] Insert `err.message` in the output
- [x] Make error output identical in mini & verbose
- [x] Clean up changed code
- [x] Resolve https://github.com/avajs/ava/issues/1175
- [x] Resolve https://github.com/avajs/ava/pull/1154#issuecomment-273297034
- [x] Fix https://github.com/avajs/ava/pull/1154#issuecomment-274571949
- [x] Fix https://github.com/avajs/ava/pull/1154#issuecomment-274570069
- [x] Remove all `NOTE`s
- [x] Explain https://github.com/avajs/ava/pull/1154#discussion-diff-97379313R189
- [x] Remove `t.jsxEqual` and open a new issue describing what it's for
- [x] Transfer my fork of `pretty-format`
- [x] Trim code excerpts
- [x] Fix code excerpt extraction from non-test files when `babel-register` is involved (line number doesn't match the actual one)


lib/code-excerpt.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
+
+const fs = require('fs');
+const codeExcerpt = require('code-excerpt');
+const repeating = require('repeating');
+const truncate = require('cli-truncate');
+const chalk = require('chalk');
+
+function formatLineNumber(line, maxLines) {
+	return repeating(' ', String(maxLines).length - String(line).length) + line;
```
sindresorhus: You can use `String#repeat()` now.

package.json
```diff
@@ -149,10 +152,11 @@
     "package-hash": "^1.1.0",
     "pkg-conf": "^2.0.0",
     "plur": "^2.0.0",
-    "power-assert-context-formatter": "^1.0.4",
-    "power-assert-renderer-assertion": "^1.0.1",
-    "power-assert-renderer-succinct": "^1.0.1",
+    "pretty-format": "github:vadimdemedes/pretty-format#colors",
     "pretty-ms": "^2.0.0",
+    "react": "^15.4.0",
```
sindresorhus: Mistake?

sindresorhus:
Can you comment on what's left?

sindresorhus:
I noticed `$ ava` and `$ ava --verbose` has some inconsistent output. We should really try to unify the output and share logic between them. The output below `1 failed` and `1 test failed [02:03:57]` should be identical:

screen shot 2016-12-17 at 02 03 44

screen shot 2016-12-17 at 02 03 57

sindresorhus:
For `t.fail()`, I think we should make an exception and not show the Actual/Expected output. It makes no sense and the code excerpt is clear enough.

screen shot 2016-12-17 at 02 06 33

package.json
```diff
@@ -149,10 +152,11 @@
     "package-hash": "^1.1.0",
     "pkg-conf": "^2.0.0",
     "plur": "^2.0.0",
-    "power-assert-context-formatter": "^1.0.4",
-    "power-assert-renderer-assertion": "^1.0.1",
-    "power-assert-renderer-succinct": "^1.0.1",
+    "pretty-format": "github:vadimdemedes/pretty-format#colors",
```
cpojer: fyi pretty-format 18 is out :)

lib/format-assert-error.js
```diff
@@ -0,0 +1,73 @@
...
+
+		return `Difference:\n\n${msg}\n`;
+	}
+
+	return [
+		'Actual:\n',
+		`${indentString(err.actual, 2)}\n`,
+		'Expected:\n',
+		`${indentString(err.expected, 2)}\n`
+	].join('\n');
```
novemberborn: Could use a single template string.

lib/format-assert-error.js
```diff
@@ -0,0 +1,73 @@
...
+		return statements
+			.map(statement => `${statement[0]}\n${chalk.grey('=>')} ${statement[1]}`)
+			.join('\n\n') + '\n';
+	}
+
+	if ((err.actualType === 'object' || err.actualType === 'array') && err.actualType === err.expectedType) {
+		const patch = diff.createPatch('string', err.actual, err.expected);
+		const msg = patch
+			.split('\n')
+			.splice(4)
```
novemberborn: Should use `.slice(4)`.

lib/cli.js
```diff
@@ -112,7 +114,7 @@ exports.run = function () {
 		explicitTitles: cli.flags.watch,
 		match: arrify(cli.flags.match),
 		babelConfig: babelConfig.validate(conf.babel),
-		resolveTestsFrom: cli.input.length === 0 ? pkgDir : process.cwd(),
+		resolveTestsFrom: resolveTestsFrom,
```
novemberborn: Can just specify `resolveTestsFrom,` if you like.

package.json
vadimdemedes: This is a fork with support for colorizing the output, from what I can see the latest release doesn't have smth similar.

vadimdemedes:
> Can you comment on what's left?

From what I can remember, mostly tests. I also wanted to check if t.jsxEquals works equally great with Preact.

> The output below 1 failed and 1 test failed [02:03:57] should be identical

I think this can be solved right now outside of this PR, as this change would be unrelated.

> For t.fail(), I think we should make an exception and not show the Actual/Expected output. It makes no sense and the code excerpt is clear enough.

Sure, good catch.

package.json
vadimdemedes: Ah, need to check that.

vadimdemedes:
Added a todo list, so that you guys know what's going on. I squash commits right away, so there wasn't a quick way to track progress.

lib/format-assert-error.js
vadimdemedes: Tried that, but didn't work out, because the resulting string contains code's indentation:

```
    Actual:
    ...
    Expected:
    ...
```

Am I doing smth wrong here?

package.json
vadimdemedes: React is needed by `react-element-to-jsx-string` module. I guess I could try submitting a PR (if possible) to avoid the dependency on React.

lib/format-assert-error.js
sindresorhus: No, that's a common issue when using template strings like that. You either need to just remove the indent, which looks weird, or use something like https://github.com/sindresorhus/redent In this case, I don't think it's worth it.

sindresorhus:
> I squash commits right away, so there wasn't a quick way to track progress.

Can you not squash? It makes it impossible to review what changed. Just add commits and we'll squash at the end when merging.

vadimdemedes:
Good point.

sindresorhus:
@vadimdemedes Note this issue when using this PR: https://github.com/avajs/ava/issues/1176 (I've confirmed it)

lib/assert.js
```diff
@@ -151,6 +154,16 @@ x.ifError = (err, msg) => {
 	test(!err, create(err, 'Error', '!==', msg, x.ifError));
 };
 
+x.jsxEqual = (val, expected, msg) => {
+	const treeEquals = reactElementToJSXString(val) === reactElementToJSXString(expected);
+	test(treeEquals, create(val, expected, '===', msg, x.jsxEqual));
+};
+
+x.jsxNotEqual = (val, expected, msg) => {
```
sindresorhus: Shouldn't this be `notJsxEqual`?

package.json
sindresorhus: 👍

lib/assert.js
vadimdemedes: I was changing it back and forth, probably should be that way indeed. I didn't like capitalization of `jsx` is different in `jsxEqual` and `notJsxEqual`.

sindresorhus: I agree about it looking uglier, but better to be consistent and always have `not` first. In a perfect world we would use snakecase in JS...

vadimdemedes:
What output should we display for these tests? https://github.com/avajs/ava/blob/master/test/fixture/throw-named-function.js https://github.com/avajs/ava/blob/master/test/fixture/throw-anonymous-function.js

sindresorhus:
@vadimdemedes For reference, currently they output:

```
❯ ava

  1 passed
  1 exception

  Uncaught Exception
  function fooFn() {}
```

```
❯ ava

  1 passed
  1 exception

  Uncaught Exception
  () => {}
```

I don't think we can do much better than that, as we have no information when people throw non-errors. We could maybe better indicate that it's a non-error. For example:

```
❯ ava

  1 passed
  1 exception

  Uncaught Exception
  Threw non-error: function fooFn() {}
```

mightyiam:
"Threw non-error" seems sweet.

vadimdemedes:
Oh my, big milestone for magic-assert - first green CI build!

vadimdemedes:
Before I forget, I better note this here: when this is released, users will need to update their snapshots. This is required, because this PR changes the way snapshots are stored. Previously, snapshots stored the complete JSX tree, now they store JSON representation of that tree. Thanks to that, AVA can render beautiful diffs between your actual JSX tree and the one stored in the snapshot.

lib/cli.js
```diff
@@ -103,7 +103,9 @@ exports.run = () => {
 		throw new Error(colors.error(figures.cross) + ' The --require and -r flags are deprecated. Requirements should be configured in package.json - see documentation.');
 	}
 
-	const api = new Api({
+	var resolveTestsFrom = cli.input.length === 0 ? pkgDir : process.cwd();
+
+	var api = new Api({
```
vadimdemedes: `var` => `const`

vadimdemedes:
Regarding https://github.com/avajs/ava/pull/1154/files#diff-9c6c3cd6ffcef4f57ac6e0c793b31b85R337, I don't think there's a need anymore to generate error message for `t.deepEqual()`. They aren't helpful either way.

ccorcos:
:shipit: !!

sindresorhus:
@vadimdemedes It's not entirely clear what you're asking. What error message? Why is it not helpful?

lib/enhance-assert.js
```diff
@@ -1,69 +1,82 @@
...
 	't.truthy(value, [message])',
 	't.falsy(value, [message])',
 	't.true(value, [message])',
 	't.false(value, [message])',
-	't.is(value, expected, [message])',
-	't.not(value, expected, [message])',
-	't.deepEqual(value, expected, [message])',
-	't.notDeepEqual(value, expected, [message])',
-	't.regex(contents, regex, [message])',
-	't.notRegex(contents, regex, [message])',
```
novemberborn: You'll have to update https://github.com/avajs/babel-preset-transform-test-files/blob/master/espower-patterns.json as well. Which is a breaking change 😝

lib/format-assert-error.js
novemberborn: Nudge.

novemberborn:
I wonder if some of the new files in `lib` can be placed in a subdirectory instead? `lib` is becoming quite hard to scan.

lib/enhance-assert.js
vadimdemedes: Cool, will do!

vadimdemedes:
@novemberborn Agree. Could we do it in a separate beginner-friendly PR instead?

vadimdemedes:
@sindresorhus I'm referring to `err.message` that gets auto-generated by `AssertionError` when `t.deepEqual()` fails:

```js
t.deepEqual(actual, expected);
// if no message is provided as a 3rd argument,
// a default message will be auto-generated and will look like:
// actual === expected
```

See https://github.com/avajs/ava/pull/1154/files#diff-9c6c3cd6ffcef4f57ac6e0c793b31b85L340 for how we tested it before (note 2nd argument of `t.throws()`).

novemberborn:
>> I wonder if some of the new files in lib can be placed in a subdirectory instead? lib is becoming quite hard to scan.

> @novemberborn Agree. Could we do it in a separate beginner-friendly PR instead?

@vadimdemedes I mean, it's not a blocker, but it's probably easier for you to organize then for somebody who comes along later.

sindresorhus:
> Regarding https://github.com/avajs/ava/pull/1154/files#diff-9c6c3cd6ffcef4f57ac6e0c793b31b85R337, I don't think there's a need anymore to generate error message for t.deepEqual(). They aren't helpful either way. That's why I want to remove those NOTEs and keep the tests without asserting error messages, are you guys ok with this?

👍 No need now that we have good diffing.

> I mean, it's not a blocker, but it's probably easier for you to organize then for somebody who comes along later.

Regardless of the outcome, it could be a separate PR, not here.

> Document jsxEqual & jsxNotEqual (Added by Sindre)

@vadimdemedes What's left other than this?

package.json
sindresorhus: @vadimdemedes What's the intention here? Upstream those changes? We cannot depend on a git branch, so we'll either have to upstream or you need to publish your fork to npm.

novemberborn: If we're forking, we could publish to `@ava/pretty-format`.

jfmengels:
Just saw this pass by: https://github.com/facebook/jest/issues/2673, which I think would be a cool feature. (We could implement that later)

package.json
vadimdemedes: I will publish it as `@ava/pretty-format` and resubmit my syntax highlighting proposal to facebook/jest (original at https://github.com/thejameskyle/pretty-format/issues/57).

vadimdemedes: Will move my fork over to our org too.

vadimdemedes:
>> I wonder if some of the new files in lib can be placed in a subdirectory instead? lib is becoming quite hard to scan.
>
> @vadimdemedes I mean, it's not a blocker, but it's probably easier for you to organize then for somebody who comes along later.

@novemberborn Agree, but I just want to get it out ASAP. I'd like to postpone this to another PR (can assign this to myself) and focus on blocking issues for this PR, due to lack of a lot of free time.

vadimdemedes:
@jfmengels good suggestion! I need to try that out with this PR and see if it works already. It might work, since we use char by char diffing for strings.

@sindresorhus Aside from documentation, nothing, I think. I'd like to take another look on output of `t.true()`/`t.false()`, but it can be postponed until later. I need to limit the number of statements in the output and see how we can prevent big arrays/objects displayed in full.

As for docs, if anyone has some free time, I'd really really appreciate the help on updating the readme and documenting `t.jsxEqual()`.

sindresorhus:
> Aside from documentation, nothing, I think. I'd like to take another look on output of t.true()/t.false(), but it can be postponed until later. I need to limit the number of statements in the output and see how we can prevent big arrays/objects displayed in full.

Yes. Should we postponed. We need to land this asap. Can you open a new issue about it so we don't forget?

> As for docs, if anyone has some free time, I'd really really appreciate the help on updating the readme and documenting t.jsxEqual().

Sure. Happy to :)

vadimdemedes:
> Yes. Should be postponed. We need to land this PR asap. Can you open a new issue about it so we don't forget?

Sure, in a minute!

> Sure. Happy to :)

Thanks a lot!

lib/assert.js
```diff
@@ -150,6 +152,14 @@ x.ifError = (err, msg) => {
 	test(!err, create(err, 'Error', '!==', msg, x.ifError));
 };
 
+x.jsxEqual = (val, expected, msg) => {
+	test(deepEqual(val, expected), create(val, expected, '===', msg, x.jsxEqual));
+};
+
+x.notJsxEqual = (val, expected, msg) => {
+	test(!deepEqual(val, expected), create(val, expected, '!==', msg, x.notJsxEqual));
+};
```
sindresorhus: When looking into documenting these, I realized they're exactly the same as `t.deepEqual()`. So what is the point of these?

lib/assert.js
```diff
@@ -163,21 +173,26 @@ x._snapshot = function (tree, optionalMessage, match, snapshotStateGetter) {
...
+		message += '\n\n' + indentString(optionalMessage, 2);
 	}
 
 	state.save();
 
-	test(result.pass, create(result, false, 'snapshot', message, x.snap));
+	let expected;
+
+	if (result.expected) {
+		// JSON in a snapshot is surrounded with `"`
```
sindresorhus: Why?

test/api.js
```diff
@@ -303,14 +303,15 @@ function generateTests(prefix, apiCreator) {
...
+		t.plan(1);
 
 		const api = apiCreator();
 
 		return api.run([path.join(__dirname, 'fixture/circular-reference-on-assertion.js')])
 			.then(result => {
 				t.is(result.failCount, 1);
-				t.match(result.errors[0].error.message, /'c'.*?'d'/);
+				// NOTE: error's message is empty due to lib/assert.js:22
+				// t.match(result.errors[0].error.message, /'c'.*?'d'/);
```
sindresorhus: Can you remove these then? And in the other places.

sindresorhus:
Snapshot testing seems to be broken:

```js
import test from 'ava';

test('foo', t => {
	t.snapshot({foo: 'bar'});
});
```

I run AVA and then change `'bar'` to `'bar2'`, I get:

```
  1 failed

  foo
  test.js:4

   3: test('foo', t => {
   4:   t.snapshot({foo: 'bar2'});
   5:

  Actual:

    Object {
      foo: "bar2",
    }

  Expected:

    undefined

  Please check your code or --update-snapshots

  Test.fn (test.js:4:4)
```

`Expected` should not be undefined.

lib/assert.js
vadimdemedes: I don't think there's any now.

I do have one idea for these, but not sure if that can be implemented to be stable enough. We (or community) could write a babel plugin for react/preact/inferno, which detects `t.jsxEqual` and convert:

```jsx
t.jsxEqual(<This/>, <That/>);
```

into:

```jsx
t.jsxEqual(renderer.create(<This/>).toJSON(), renderer.create(<That/>).toJSON());
```

to save the user from typing that manually or creating shortcuts (as mentioned in #1175).

lib/assert.js
vadimdemedes: This issue is described in https://github.com/developit/preact-render-to-string/issues/16, where user tried to save arbitrary string (not JSX) as a snapshot.

test/api.js
vadimdemedes: Sure, now that we agreed it's not needed, I will remove all the `NOTE`s.

sindresorhus:
`t.deepEqual()` doesn't display very useful output:

```js
import test from 'ava';

test('foo', t => {
	t.deepEqual({
		foo: {
			bar: 3
		}
	}, {
		foo: {
			bar: 4
		}
	});
});
```

Results in:

```
  1 failed

  foo
  test.js:4

   3: test('foo', t => {
   4:   t.deepEqual({
   5:     foo: {

  {foo:{bar:3}}
  => Object {
    foo: Object {
      bar: 3,
    },
  }

  {bar:3}
  => Object {
    bar: 3,
  }

  sdfs

  Test.fn (test.js:4:4)
```

I don't see the change to `4` anywhere there.

vadimdemedes:
> t.deepEqual() doesn't display very useful output:

Hm, this is the output for `t.true`/`t.false`. Weird how it got displayed for `t.deepEqual()`, will check that.

lib/assert.js
sindresorhus: @vadimdemedes Can you include the link in the comment. Usually comments are great for explaining "why".

lib/assert.js
sindresorhus: Ok. I like that idea. Can you remove the `jsxEqual` stuff for now and open a new issue about `jsxEqual`? We can add it back when we have a better story for it. It was pretty unrelated to this pull request anyways.

lib/assert.js
vadimdemedes: Sure, will do.

lib/assert.js
vadimdemedes: Yeah, agree.

gajus:
I am so excited about this. Cannot wait to see it out. :-)

package.json
gajus: How I did not know about `pretty-format`. :-(

Seems like I reinvented the wheel https://www.npmjs.com/package/prettyprint.

vadimdemedes: Haha yeah, it took me a while to find out about it too.

vadimdemedes:
Updated PR description with new todo items.

@sindresorhus I removed "Document t.jsxEqual" todo added by you, since we're removing it.

vadimdemedes:
Ok, I figured out why `t.deepEqual()` was missing the expected "Difference" output and was showing statements instead. In https://github.com/avajs/babel-preset-transform-test-files `t.deepEqual()` is still contained in enhanced patterns config, that's why power assert context is created for them. If it's removed from `@ava/babel-preset-transform-test-files`, output goes back to normal. Will create a new branch there and fix it.

vadimdemedes:
Attention everyone, magic-assert just crossed off all points from the todo list and is very short from landing in master! I would very appreciate, if you tested it out in some real-world projects and report any issues here. Thank you!

jfmengels:
The error output is a bit odd when using `t.throws`.

image

1. There is a comparison of actual and expected, which may make sense when comparing the thrown errors, but here there is one missing. I think it would be clearer by not having this comparison

2. Missing expected exception has two dots at the end. I feel like this one has always been like this, so maybe for another PR. That said, I never found this error very comprehensive: Does `t.throws` expect me to pass an exception type/message, or did it run and the function simply did not throw? Happy to create a different issue to discuss this.

vadimdemedes:
@jfmengels Great catch, I'll disable assertion output for t.throws. And yes, dots were there before, will create a beginner-friendly issue.

package.json
novemberborn: Need to remove this line.

vadimdemedes:
@novemberborn @jfmengels PR updated, thanks! No more output for `t.throws`/`t.doesNotThrow`.

vadimdemedes:
Added a new todo:

> Fix code excerpt extraction from non-test files when babel-register is involved (line number doesn't match the actual one)

novemberborn:
> Fix code excerpt extraction from non-test files when babel-register is involved (line number doesn't match the actual one)

Is this a source maps issue? I've always assumed that `babel-core/register` causes conflicts in `source-map-support`. We'll be rolling our own solution with [RFC 001](https://github.com/avajs/ava/blob/master/docs/specs/001%20-%20Improving%20language%20support.md) so perhaps this isn't worth worrying about.

sindresorhus:
> Is this a source maps issue? I've always assumed that babel-core/register causes conflicts in source-map-support. We'll be rolling our own solution with RFC 001 so perhaps this isn't worth worrying about.

@vadimdemedes Can you open an issue about this and mention that this will be handled by RFC1?

vadimdemedes:
> Is this a source maps issue?

@novemberborn Yes, I think so, because it points to a compiled file, not the original.

Cool, I'll open an issue, so that we don't forget about this. Until then, I'll add a change, that prevents AVA from crashing, when there's no code excerpt available.

vadimdemedes:
Done, AVA doesn't crash anymore when code excerpt points to the wrong location and is broken.

sindresorhus:
This is absolutely amazing @vadimdemedes! So good in fact, I almost want test failures now 😝

vadimdemedes:
Thank you, thank you, thank you! I'm sooo excited! Indeed, can't wait for the first bug report :D

Issue #707, opened 2016-04-04
Normalize relative paths in babel.extends, plugins, and presets to package.json directory


jamestalmage:
Probably best handled along with #700.

Assuming following config in `package.json`:

``` js
{
  "ava": {
    "babel": {
       "extends": "./some-config.json"
     }
  }
}
```

It will work as long as `process.cwd()` is the same as `package.json`.

`cli.js` should normalize the path to `package.json` (`api.js` does not know anything about `package.json`).

This should be part of the config normalizing discussed in #700

novemberborn:
This will be done as part of [RFC001](https://github.com/avajs/ava/blob/master/docs/specs/001%20-%20Improving%20language%20support.md).

Issue #631, opened 2016-03-10
Allow tests files to have any extension (i.e. `.jsx`, `.ts`)


dustinsanders:
With the latest release(0.13) and the ability to support jsx; I would like the ability to write my test that contains jsx with a jsx extension. However ava is [filtering](https://github.com/sindresorhus/ava/blob/master/api.js#L320) only for files with .js extensions. I'm a little confused as to why ava allows for a glob pattern in the cli but then only filters down to .js files.

I can make a PR if this is something worthwhile to the maintainers.

---

**See conclusion here: https://github.com/avajs/ava/issues/631#issuecomment-248659780**

sotojuan:
Personally I'd argue that there's no point in using `.jsx` files in general, but that's for another debate. This would only require changing one line, so I am for it.

jamestalmage:
Extensions provide hints to the IDE as to which language features should be enabled.

My question is, do we just make a special exception for `.jsx`, or parse glob strings for extensions?

```
ava test/*.{js,jsx}
```

Would enable `js` and `jsx`

dustinsanders:
I think ideally ava would accept any extension, as long as there is a register hook to get the file to vanilla javascript. I believe this would also fix issue #521.

jamestalmage:
>I think ideally ava would accept any extension, as long as there is a register hook to get the file to vanilla javascript.

I like that thinking. The only difficulty is that we can't know the list of registered extensions in the parent process, we process require hooks in the child.

novemberborn:
I think we can support `.jsx` specifically. We could even add the Babel transform for `.jsx` files (though perhaps not depend on it).

Accepting any extension is trickier. We might have to swap out the compiler. #577 discusses this with regards to source files but the same would apply here. I suspect we might solve that with an explicit extension registry, not just accepting whatever is matched by the glob patterns.

jamestalmage:
Why is matching the glob string problematic? If they explicitly define an extension just trust the user will add the require hook themselves (perhaps a helpful message if they don't).

sindresorhus:
> If they explicitly define an extension just trust the user will add the require hook themselves (perhaps a helpful message if they don't).

:+1:

novemberborn:
> If they explicitly define an extension just trust the user will add the require hook themselves

Yes, but this is the part we're currently missing. You can't swap out the Babel precompiler. Adding `.jsx` support is easier because it uses Babel.

vdemedes:
[...]

PR #1159, opened 2016-12-20
RFC 001: Improving language support

novemberborn:
An [RFC](http://blog.npmjs.org/post/153881413635/some-notes-on-rfcs) with a proposal for improving AVA's support for Babel, React and TypeScript projects.

We'll leave this open as a PR for a while to gather feedback. If accepted this incorporates / supersedes #707, #1082, #945, #631 and #1122.

### DIFF
--- a/docs/specs/001 - Improving language support.md
+++ b/docs/specs/001 - Improving language support.md
@@ -42,20 +42,20 @@ AVA's other transforms, such as `babel-plugin-espower` and `babel-plugin-ava-thr
 
 The above assumes AVA is used with regular JavaScript projects that do not require compilation. Many users though already have a Babel pipeline in place and wish to use AVA without having to precompile their source files.
 
-At it's simplest, setting `"babel": true` in the AVA configuration enables AVA's support for Babel projects. Test and helper files are compiled as per the above, but source files are now automatically compiled as well.
+At its simplest, setting `"babel": true` in the AVA configuration enables AVA's support for Babel projects. Test and helper files are compiled as per the above, but source files are now automatically compiled as well.
 
 AVA looks at either the project's `package.json` or `.babelrc` files for the Babel options used to compile source files (ideally we can extract a proper Babel configuration object from these two locations). This is a simplification of Babel's actual configuration management, which searches for the options file that is closest to the file being compiled. Looking at these two specific files allows AVA to use cached compilation results without even having to load Babel, while still recompiling source files if options change.
 
 AVA's handling of Babel projects can be further configured by passing an options object instead of `true`:
 
 * `compileSources: true | false`: defaulting to `true`, determines whether sources are compiled.
-* `extensions: "js" | ["js", "jsx", ...]`: defaulting to `"js"`, specifies the allowed file extensions. This expandn the default test file patterns.
+* `extensions: "js" | ["js", "jsx", ...]`: defaulting to `"js"`, specifies the allowed file extensions. This expands the default test file patterns.
 * `sourceOptions: {}`: specify the [Babel options] used to compile source files. In this context `babelrc: true` causes options to be merged with those found in either the project's `package.json` or `.babelrc` files. `babelrc` defaults to `true`.
 * `testOptions: {}`: specify the [Babel options] used to compile test and helper files. If provided this completely disables the default Babel configuration AVA uses to compile test and helper files. Like with `sourceOptions`, `babelrc` defaults to `true`. Set `presets: ["ava"]` to apply AVA's transforms.
 
 `sourceOptions` can be used to extend a shared Babel configuration so that the source files can be loaded in AVA tests. For instance users may [rely on webpack to resolve ES2015 module syntax at build time, but still need to apply `babel-plugin-transform-es2015-modules-commonjs` for sources to work in AVA][source options reason].
 
-`sourceOptions` and `testOptions` may specify `ignore` and `only` values. These are only used to determine whether the file needs compilation. They do not impact test file selection or source watching.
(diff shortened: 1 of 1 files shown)


zixia:
Great, hope we can add support to those languages soon.

As my view, if we could get rid of the limitation in ava-files, ie: do not restrict the file extension, just let users define them... 

Then we can use `ts-node` or `babel-node` to run ava with the specific language support, without any more code design/modification, and everyone will be happy. :P

novemberborn:
> As my view, if we could get rid of the limitation in ava-files, ie: do not restrict the file extension, just let users define them...

One of the patterns we support is one that matches a directory. AVA automatically expands that to find all `.js` files within. We couldn't do that if any extension is allowed, yet not having to add `**/*.js` to your patterns is quite nice.

> Then we can use ts-node or babel-node to run ava with the specific language support, without any more code design/modification, and everyone will be happy. :P

While tempting, I think the performance cost will be too high. It'll probably also stop us from doing things like #789.

docs/specs/001 - Improving language support.md
```diff
@@ -0,0 +1,94 @@
...
+* `sourceOptions: {}`: specify the [Babel options] used to compile source files. In this context `babelrc: true` causes options to be merged with those found in either the project's `package.json` or `.babelrc` files. `babelrc` defaults to `true`.
+* `testOptions: {}`: specify the [Babel options] used to compile test and helper files. If provided this completely disables the default Babel configuration AVA uses to compile test and helper files. Like with `sourceOptions`, `babelrc` defaults to `true`. Set `presets: ["ava"]` to apply AVA's transforms.
+
+`sourceOptions` can be used to extend a shared Babel configuration so that the source files can be loaded in AVA tests. For instance users may [rely on webpack to resolve ES2015 module syntax at build time, but still need to apply `babel-plugin-transform-es2015-modules-commonjs` for sources to work in AVA][source options reason].
+
+`sourceOptions` and `testOptions` may specify `ignore` and `only` values. These are only used to determine whether the file needs compilation. They do not impact test file selection or source watching.
```
sindresorhus: It's not immediately clear that `ignore` and `only` refers to Babel options.

docs/specs/001 - Improving language support.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Compilation
+
+Based on this [proof of concept](https://github.com/avajs/ava/pull/1082) Babel compilation is moved into the test workers. If source files are to be compiled AVA will load its own require hook, rather than relying on `babel-core/register`.
```
sindresorhus: 👍

docs/specs/001 - Improving language support.md
```diff
@@ -0,0 +1,94 @@
...
+
+The above assumes AVA is used with regular JavaScript projects that do not require compilation. Many users though already have a Babel pipeline in place and wish to use AVA without having to precompile their source files.
+
+At it's simplest, setting `"babel": true` in the AVA configuration enables AVA's support for Babel projects. Test and helper files are compiled as per the above, but source files are now automatically compiled as well.
+
+AVA looks at either the project's `package.json` or `.babelrc` files for the Babel options used to compile source files (ideally we can extract a proper Babel configuration object from these two locations). This is a simplification of Babel's actual configuration management, which searches for the options file that is closest to the file being compiled. Looking at these two specific files allows AVA to use cached compilation results without even having to load Babel, while still recompiling source files if options change.
+
+AVA's handling of Babel projects can be further configured by passing an options object instead of `true`:
+
+* `compileSources: true | false`: defaulting to `true`, determines whether sources are compiled.
```
sindresorhus: So does this mean that compilation of sources are now true by default without any config?

novemberborn: Only if you set `babel`. So out of the box the behavior stays exactly the same as it is now (with helper compilation).

sindresorhus: Good :)

sindresorhus:
> But how will TS compilation be made? Will AVA have to ship with the TypeScript build tools? Or can we expect it as an optional peer dependency or something?

We will not bundle TypeScript. Users will have to add TypeScript as a dependency. We'll throw a nice error if they forgot to add the dependency.

novemberborn:
We might not even include the TypeScript support module (I think that's what I proposed here, but on mobile so don't feel like checking right now).

vadimdemedes:
Great work, @novemberborn! Thorough and to the point!

I wonder if we could come up with a better name for `babelrc: true` option, since it pulls config not only from `.babelrc`, but from `babel` prop in package.json.

novemberborn:
> I wonder if we could come up with a better name for babelrc: true option, since it pulls config not only from .babelrc, but from babel prop in package.json.

Yea I hear ya. It's a (no longer documented, as far as I can tell) option in the transform APIs though. It refers both to `.babelrc` and the `babel` stanza in `package.json`. (In the CLI there is a [negation flag: `--no-babelrc`](http://babeljs.io/docs/usage/cli/#ignoring-babelrc).)

What's tricky is that the `sourceOptions` and `testOptions` objects are supposed to be passed to Babel as-is. I'll do some more research and get in touch with Babel people for what the preferred setting is.

novemberborn:
> I'll do some more research and get in touch with Babel people for what the preferred setting is.

Though I haven't spoken with anybody, it seems pretty clear to me that it's `babelrc`. I've opened https://github.com/babel/babel/pull/5101 and https://github.com/babel/babel.github.io/pull/1134 to clarify Babel's documentation.

Anything else or is this good to go?

sindresorhus:
:shipit:

novemberborn:
I shall commence the project planning! 💥

PR #1160, opened 2016-12-21
Print out warning in verbose & mini reporter when `--fail-fast` is enabled

### LINKED ISSUE #1134: Print a reminder when `--fail-fast` is on
When a run stops due to a fail and when `--fail-fast` was provided, I'd like there to be a bold notice somewhere, perhaps at the bottom of the output, reminding me that `--fail-fast` is on, so that I will never mistakenly think that I have only one failing test.

ThomasBem:
Hi there,

This PR fixes #1134. If --fail-fast is enabled it will print out the following message at the end of the report for both mini and verbose.

> `--fail-fast` is on. Any number of tests may have been skipped

**Mini:**
screen shot 2016-12-21 at 14 42 39

**Verbose:**
screen shot 2016-12-21 at 14 44 03

This works for now, but i´m not sure if the way I have done it is the best. 

I have been thinking that maybe instead of going through all the results in run-status looking for a stat element that has failFastEnabled=true. 

It might be better to send a event or something if the Runner is created with the fail-fast option enabled. Then run-status could pickup on this and just set:
 this.failFastEnabled = true

Let me know what you guys think!

Also i appreciate the support in the issue section helping me out with this :)

### DIFF
--- a/api.js
+++ b/api.js
@@ -145,7 +145,8 @@ Api.prototype._run = function (files, options) {
 	var runStatus = new RunStatus({
 		runOnlyExclusive: options.runOnlyExclusive,
 		prefixTitles: this.options.explicitTitles || files.length > 1,
-		base: path.relative('.', commonPathPrefix(files)) + path.sep
+		base: path.relative('.', commonPathPrefix(files)) + path.sep,
+		failFast: this.options.failFast
 	});
 
 	this.emit('test-run', runStatus, files);
--- a/lib/colors.js
+++ b/lib/colors.js
@@ -9,5 +9,6 @@ module.exports = {
 	pass: chalk.green,
 	duration: chalk.gray.dim,
 	errorStack: chalk.gray,
-	stack: chalk.red
+	stack: chalk.red,
+	failFast: chalk.magenta
 };
--- a/lib/reporters/mini.js
+++ b/lib/reporters/mini.js
@@ -216,6 +216,10 @@ MiniReporter.prototype.finish = function (runStatus) {
 		});
 	}
 
+	if (runStatus.failFastEnabled === true) {
+		status += '\n\n  ' + colors.failFast('`--fail-fast` is on. Any number of tests may have been skipped');
+	}
+
 	return status + '\n\n';
 };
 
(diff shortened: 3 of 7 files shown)


novemberborn:
You should pass the `bail` option when `RunStatus` is instantiated: https://github.com/avajs/ava/blob/033d4dcdcbdadbf665c740ff450c2a775a8373dc/api.js#L145:L149

It'd be nice though if this wouldn't report when it knows no other tests didn't run. Which brings us back to the test counting logic.

ThomasBem:
Thats a much better solution than what i did. Thanks, I'll try and make that change.

You mentioned in the issue that it was probably hard or at least not a guarantee that we could  determine the total amount of test to be run by the time failure occurred? So probably out of my league to fix that at this time.

But maybe open that as a separate issue / feature request for after we can get this initial functionality in place?

ThomasBem:
Updated PR to reflect change suggested by @novemberborn. FailFast is now passed into RunStatus.

api.js
```diff
@@ -145,7 +145,8 @@ Api.prototype._run = function (files, options) {
 	var runStatus = new RunStatus({
 		runOnlyExclusive: options.runOnlyExclusive,
 		prefixTitles: this.options.explicitTitles || files.length > 1,
-		base: path.relative('.', commonPathPrefix(files)) + path.sep
+		base: path.relative('.', commonPathPrefix(files)) + path.sep,
+		failFast: options.failFast
```
sindresorhus: Use `this.options.failFast` instead and you don't have to pass it on https://github.com/avajs/ava/pull/1160/files#diff-f8849ae9f0614ba9464fdd14c2d98dc0R163 as it's already passed in the API constructor.

ThomasBem:
Updated PR with your suggestions. Thanks for the assist 👍

novemberborn:
> It'd be nice though if this wouldn't report when it knows no other tests didn't run. Which brings us back to the test counting logic.

@sindresorhus any thoughts on this? Happy to land this PR as-is and then do this as a follow-up.

sindresorhus:
Looks great. Thanks for this excellent contribution @ThomasBem :)

@novemberborn Can you open that follow-up issue you talked about?

novemberborn:
Thanks @ThomasBem!

Opened #1171 and #1172 for follow-up.

PR #1164, opened 2016-12-23
Fix Flow type ambiguity

### LINKED ISSUE #1114: ava 0.17.0 with FlowType 0.35.0 results in type-ambiguity errors
## Description

Trying to update a project to use ava 0.17.0

For what seem like pretty standard-looking ava tests to me (as far as I know), FlowType is getting all up in my face about some ambiguity relating to the new types that ava declares. I am obviously quite scandalised by all this. Oh my!

### Test Source

This is the test code in question:

```js
test('A, A->B, B->C, BC->end', (t) => {
  const tasks = [
    { fn: noop, id: 'a', label: 'a', requires: [ 'B', 'C' ] },
    { fn: noop, id: 'b', label: 'b', provides: [ 'C' ], requires: [ 'B' ] },
    { fn: noop, id: 'c', label: 'c', provides: [ 'B' ], requires: [ 'A' ] },
    { fn: noop, id: 'd', label: 'd', provides: [ 'A' ] }
  ]
  t.notThrows(() => tasks.sort(compareByMetadata))
  const ids = tasks.map((task) => task.id)
  t.deepEqual(ids, [ 'd', 'c', 'b', 'a' ])
})
```

When I run `flow_check`, I get 23 errors that are all very similar, like this:

```
test/sort-tasks.js:63
 63: test('A, A->B, B->C, BC->end', (t) => {
     ^ function call. Could not decide which case to select
201: declare module.exports: {
                             ^ callable object type. See: node_modules/ava/index.js.flow:201
  Case 2 may work:
  203: 	(name: string, run: ContextualTest): void;
       	^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ function type. See: node_modules/ava/index.js.flow:203
  But if it doesn't, case 4 looks promising too:
  205: 	(name: string, run: Macros<ContextualTestContext>, ...args: Array<any>): void;
       	^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ function type. See: node_modules/ava/index.js.flow:205
  Please provide additional annotation(s) to determine whether case 2 works (or consider merging it with case 4):
   63: test('A, A->B, B->C, BC->end', (t) => {
                                            ^ return
```

### Config

```json
{
  "ava": {}
}
```

### Command-Line Arguments

```
nyc ava
```

## Relevant Links

- https://github.com/jokeyrhyme/node-init.js

## Environment

```
Node.js v7.1.0
darwin 16.3.0
ava 0.17.0
npm 4.0.2
```

leebyron:
Fixes #1114

As reported in #1114, flow reports ambiguity when defining tests:

```js
test('my test', t => {})
```

Flow will report that it could apply either the `Test` type or the `Macro` type. The `Test` type is an obvious fit, but `Macro` also applies since the `...args` resolves to the empty array, and the optional `title?` attribute happens to be not provided.

My suggested fix is to merge the `Test` and `Macro` types together since `Macro` is really a superset of the `Test` type, and we don't lose any type safety by just expecting to always get Macros.

I tested this locally on a flow-typed repository and confirmed that the type ambiguity errors are solved.

### DIFF
--- a/index.js.flow
+++ b/index.js.flow
@@ -93,26 +93,23 @@ type ContextualTestContext         = TestContext         & { context: any; };
 type ContextualCallbackTestContext = CallbackTestContext & { context: any; };
 
 /**
- * Test Types
+ * Test Implementations
  */
 
-type Test                   = (t: TestContext)                   => SpecialReturnTypes | void;
-type CallbackTest           = (t: CallbackTestContext)           => void;
-type ContextualTest         = (t: ContextualTestContext)         => SpecialReturnTypes | void;
-type ContextualCallbackTest = (t: ContextualCallbackTestContext) => void;
-
-/**
- * Macro Types
- */
-
-type Macro<T> = {
-	(t: T, ...args: Array<any>): void;
+type TestFunction<T, R> = {
+	(t: T, ...args: Array<any>): R;
 	title?: (providedTitle: string, ...args: Array<any>) => string;
 };
 
-type Macros<T> =
-	| Macro<T>
-	| Array<Macro<T>>;
+type TestImplementation<T, R> =
+	| TestFunction<T, R>
+	| Array<TestFunction<T, R>>;
+
+type Test                   = TestImplementation<TestContext, SpecialReturnTypes | void>;
+type CallbackTest           = TestImplementation<CallbackTestContext, void>;
+type ContextualTest         = TestImplementation<ContextualTestContext, SpecialReturnTypes | void>;
+type ContextualCallbackTest = TestImplementation<ContextualCallbackTestContext, void>;
+
 
 /**
  * Method Types
@@ -121,8 +118,6 @@ type Macros<T> =
 type TestMethod = {
 	(              implementation: Test): void;
 	(name: string, implementation: Test): void;
-	(              implementation: Macros<TestContext>, ...args: Array<any>): void;
-	(name: string, implementation: Macros<TestContext>, ...args: Array<any>): void;
 
 	serial     : TestMethod;
 	before     : TestMethod;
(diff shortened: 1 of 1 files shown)


bookercodes:
Thanks for this, @leebyron! 

Can't wait for this to be merged 🎆. In the meantime, is there a convinent way to apply this patch, do you know?

bookercodes:
Hm. I tried to apply this locally but [ran into some trouble](https://github.com/facebook/flow/issues/3078). If anyone has any ideas, please let me know! Thanks!

novemberborn:
> My suggested fix is to merge the Test and Macro types together since Macro is really a superset of the Test type, and we don't lose any type safety by just expecting to always get Macros.

I haven't looked at (these, or any, really) Flow definitions much. From the `test()` method's perspective the `t => {}` function is a _test implementation_. "Macro" is more a way of explaining how to reuse test implementations, AVA doesn't care. Perhaps the types can be renamed `Implementation` rather than `TestOrMacro`?

Besides that, if Flow-users could try out this PR that'd be great. I don't really know how to test it myself.

leebyron:
Great suggestion - I cleaned up the names of the types.

Anyone who would like to test this out locally, here's how I'm doing it:

* Clone AVA with this PR applied (https://github.com/leebyron/ava/tree/fix-flow-types) locally.
* From the directory you cloned to, run `npm link` to have npm use that as the source for `ava`.
* From your project that suffers from #1114, run `npm link ava`, then run `flow check` - you should no longer see flow complaining about AVA internals, and only see errors about your own code.
* (`npm unlink`)

bookercodes:
@leebyron Ah, I guess you could say `npm link` was the missing... _link_. Gosh, I'm hilarious!

Jokes aside, I managed to apply your fix locally and things work smoothly in my project, thanks!

sindresorhus:
Thanks for helping us out with this one @leebyron. We really appreciate it. No one on the team has much experience with Flow.

leebyron:
Thanks for the merge. Feel free to shout my way if you ever need help with anything related to type definitions.

PR #1165, opened 2016-12-25
switch from babel-preset-es2015-node4 to babel-preset-env

mastilver:

### DIFF
--- a/docs/recipes/babelrc.md
+++ b/docs/recipes/babelrc.md
@@ -14,7 +14,7 @@ There are multiple options for configuring how AVA transpiles your tests using B
 
 ## AVA's default transpiler behavior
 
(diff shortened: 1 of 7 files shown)


sindresorhus:
`babel-preset-env` is a good idea, but I'm not sure it's mature enough yet. Also the main reason we recently switched to `eslint-plugin-es2015-node4` (https://github.com/avajs/ava/commit/5158ac8bb0d0a79acde12c1bf6dfc1bbf32fd71e) by default was to get rid of lots of heavy Babel plugins to reduce our install time. This plugin brings them back.

mastilver:
Cool, I understand
I'm closing it then ;)

sindresorhus:
I would recommend opening an issue on `babel-preset-env` about doing something about all the dependencies. Maybe do a lightweight version of it that only targets Node.js 4 and higher.

PR #1168, opened 2017-01-03
Signature of assertions true/false for flow/TS

sebald:
## Description

The signature for the two assertions `t.true()` and `t.false()` requires the first parameter to be a `boolean` ([Link to relevant lines](https://github.com/avajs/ava/blob/8816faf6bee675213c5bfe32f628a0a683708484/types/base.d.ts#L43-L50)). This makes it cumbersome to tests if a `function` returns `true` or another non-`boolean` value.

Using `t.truthy()` or `t.falsy()` instead doesn't entirely solve this, if the function must return a `boolean` and not just something ... truthy/falsy. The below example illustrates the issue.

> This patch changes the signature of the two assertions `t.true()` and `t.false()` to `any`/`mixed` in order to allow a more convenient way to test for the exact outcome.

### Test Source

```ts
  const isNumber = x => typeof x === 'number' ? true: Error('Not a number');

  t.true(isNumber(6)); // Not working
  t.true(<boolean>isNumber(6)); // Working, even though output might not be a boolean.
```

```
  Argument of type 'true | Error' is not assignable to parameter of type 'boolean'.
  Type 'Error' is not assignable to type 'boolean'.
```

### DIFF
--- a/index.js.flow
+++ b/index.js.flow
@@ -48,9 +48,9 @@ type AssertContext = {
 	// DEPRECATED, use `falsy`. Assert that value is falsy.
 	notOk(value: mixed, message?: string): void;
 	// Assert that value is true.
-	true(value: boolean, message?: string): void;
+	true(value: mixed, message?: string): void;
 	// Assert that value is false.
-	false(value: boolean, message?: string): void;
+	false(value: mixed, message?: string): void;
 	// Assert that value is equal to expected.
 	is<U>(value: U, expected: U, message?: string): void;
 	// Assert that value is not equal to expected.
--- a/types/base.d.ts
+++ b/types/base.d.ts
@@ -43,11 +43,11 @@ export interface AssertContext {
 	/**
 	 * Assert that value is true.
 	 */
-	true(value: boolean, message?: string): void;
+	true(value: any, message?: string): void;
 	/**
 	 * Assert that value is false.
 	 */
-	false(value: boolean, message?: string): void;
+	false(value: any, message?: string): void;
 	/**
 	 * Assert that value is equal to expected.
 	 */


sindresorhus:
Makes sense to me.

@leebyron Any opinion?

novemberborn:
I can't say I understand the reasoning for this. `t.true` and `t.false` are assertions for what is essentially a type check. If you use a typed language then this is verified at compile time, which pretty much makes the assertions unnecessary (other than to make sure the code doesn't crash).

In your example you *return* an `Error` if the argument is invalid. That seems odd, why not throw instead?

leebyron:
This makes sense to me. The strong type check assumes I'm testing values of type boolean, but that's not a safe assumption. What the type signature should assume is that t.true() will be provided either "true" or "anything that isn't true", that's not the same as false. The combination of the two values then isn't "boolean" but instead "mixed" (the supertype of all types). 

I don't think it's AVA's place to make judgements on the APIs its testing in this way. Maybe this example is odd, but it exists and would like to be tested. A more common example I see a lot is nullable-boolean where a value is typically true | false but also possibly null | undefined. That would fail the type check as well but seems like a reasonable thing to test.

sebald:
> I can't say I understand the reasoning for this. `t.true` and `t.false` are assertions for what is essentially a type check.

Maybe the given example is a little bit too much contrived, but I am *not* testing the type. Rather I am testing if the functions returns the correct outcome. I could also use `t.is(isNumber(6), true)` instead, but using `t.true` is easier to read and shorter to write.

Basically, the current type definitions limit the way you can test, if the thing under tests has (or returns) a union type. The suggested PR gives the power back to the user 😄 

Furthermore: Writing `t.true(<boolean>isNumber(6));` is more error-prone, because some people might think that the returned value is cast to a boolean (which is actually not happening on runtime, the cast is only there to satisfy TypeScript).

novemberborn:
Great, thanks @sebald!

PR #1170, opened 2017-01-06
Supertest ^2.0.0 supports Promises by default

goto-bus-stop:
Hi! :wave: The `endpoint-testing` recipe mentions using SuperTest-as-promised, but it's deprecated since SuperTest now supports Promises itself.

See https://github.com/visionmedia/supertest/blob/master/History.md#200--2016-07-29

Thanks for y'all's work on Ava, it's great :tada:


novemberborn:
Thanks @goto-bus-stop!

PR #1174, opened 2017-01-08
Ensure assertion message is a string - Fixes #1125

### LINKED ISSUE #1125: Check assertion message is a string
(See https://github.com/avajs/ava/issues/1125#issuecomment-264707430, @novemberborn)

---

`node@6.8, ava@0.17.0`

When I run test, sometimes I get this error:

```bash
TypeError: test.error.message.split is not a function
    at /Users/creeper/work/projects/ysprite/node_modules/ava/lib/reporters/mini.js:181:43
    at Array.forEach (native)
    at MiniReporter.finish (/Users/creeper/work/projects/ysprite/node_modules/ava/lib/reporters/mini.js:173:20)
    at Logger.finish (/Users/creeper/work/projects/ysprite/node_modules/ava/lib/logger.js:49:27)
    at /Users/creeper/work/projects/ysprite/node_modules/ava/lib/cli.js:161:12
From previous event:
    at Object.exports.run (/Users/creeper/work/projects/ysprite/node_modules/ava/lib/cli.js:160:5)
    at Object.<anonymous> (/Users/creeper/work/projects/ysprite/node_modules/ava/cli.js:23:24)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.runMain (module.js:604:10)
    at bootstrap_node.js:509:3
```

And I found this maybe caused by:

```js
test('message', t => {
  return aPromise.then(() => {
    // error occured 
  })
})
```

leewaygroups:
This PR is intended to [fix #1125 ](https://github.com/avajs/ava/issues/1125). 

1.  The approach taken is to ensure/force assertion message to string.
2. This is handled in a single helper function `stringMsg` and
3. `stringMsg` is invoked only in assertion methods it is needed (in my opinion)

**Note: If msg is an object and non Error instance, the msg object is stringified using JSON.stringify.**

All comments/reviews are welcome.

Cheers!!

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -35,7 +35,7 @@ function stringMsg(msg) {
 }
 
 x.pass = msg => {
-	test(true, create(true, true, 'pass', msg, x.pass));
+	test(true, create(true, true, 'pass', stringMsg(msg), x.pass));
 };
 
 x.fail = msg => {
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -18,6 +18,7 @@ function create(val, expected, operator, msg, fn) {
 	if (msg !== undefined && typeof msg !== 'string') {
 		throw new TypeError('Assertion message must be a string');
 	}
+
 	return {
 		actual: val,
 		expected,
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -15,13 +15,16 @@ const noop = () => {};
 Object.defineProperty(x, 'AssertionError', {value: assert.AssertionError});
 
 function create(val, expected, operator, msg, fn) {
-	return {
-		actual: val,
-		expected,
-		message: msg,
-		operator,
-		stackStartFunction: fn
-	};
+	if (msg === undefined || typeof msg === 'string'){
+		return {
+			actual: val,
+			expected,
+			message: msg,
+			operator,
+			stackStartFunction: fn
+		};
+	}
+	throw new TypeError('assertion message must be a string');
 }
 
 function test(ok, opts) {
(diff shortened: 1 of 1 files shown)


lib/assert.js
```diff
@@ -30,45 +30,49 @@ function test(ok, opts) {
 	}
 }
 
+function stringMsg(msg) {
+	return msg instanceof Error ? msg.message : typeof msg === "string" ? msg : msg ? JSON.stringify(msg) : null;
+}
+
 x.pass = msg => {
-	test(true, create(true, true, 'pass', msg, x.pass));
+	test(true, create(true, true, 'pass', stringMsg(msg), x.pass));
```
sindresorhus: Instead of wrapping every `msg`, you can just add the validation to the `create` function.

lib/assert.js
```diff
@@ -30,45 +30,49 @@ function test(ok, opts) {
 	}
 }
 
+function stringMsg(msg) {
+	return msg instanceof Error ? msg.message : typeof msg === "string" ? msg : msg ? JSON.stringify(msg) : null;
```
sindresorhus: It should fail if `msg` is anything other than a string, not gracefully handle it.

leewaygroups: Following your comment, I can think of two possible ways to go about this.
Throw an exception when msg is anything other than

1.  A string or 
2. A string or Error object (since for errro type,  the message propoerty value can be used).

lib/assert.js
leewaygroups: @sindresorhus : Great catch! I wonder why I didn't see that.

lib/assert.js
sindresorhus: 1. Message should always be a string

lib/assert.js
```diff
@@ -15,13 +15,16 @@ const noop = () => {};
...
+	if (msg === undefined || typeof msg === 'string') {
+		return {
+			actual: val,
+			expected,
+			message: msg,
+			operator,
+			stackStartFunction: fn
+		};
+	}
+	throw new TypeError('assertion message must be a string');
```
sindresorhus: Would be better to invert the check and put the throw in the if-statement. Also: `assertion` => `Assertion`

leewaygroups: I can check if msg exists and not string then throw an error in within the if-statement:

`
if(msg && typeof msg !== 'string'){
  throw error 
}
` 

But that logic is flawed because 0 will slip through uncaught. Also, inverting the logic as suggested seem flawed as well since an OR condition is short-circuited once first test is truthy.

I'll look more closely.

Cheers!

sindresorhus: I simply meant:

```js
if (msg !== undefined && typeof msg !== 'string') {
	throw ...
}

return ...
```

sindresorhus:
This looks good now, but needs a test.

leewaygroups:
@novemberborn Yeah, some delays from me there. Sorry about that. I'll check-in the test in shortly.

novemberborn:
Hi @leewaygroups, just checking in 😄

sindresorhus:
Closing for lack of activity.

PR #1177, opened 2017-01-11
Print out warning in verbose & mini reporter when .only() tests are used

### LINKED ISSUE #1135: Show clear indicator in the output when `.only()` is used
It's easy to forget and end up not running the rest of the tests before pushing changes.

---

Similar to https://github.com/avajs/ava/issues/1134.

ThomasBem:
This PR fixes #1135. If all test files have at least one .only test in them, it will print out the following message at the end of the report for both mini and verbose.

> The .only() modifier is used in some tests. All other tests have been skipped.

If there are test files without .only in them, it will print out the following.

> The .only() modifier is used in some tests. At least', remainingTests, 'tests were not run.

remainingTests will then be the number of tests in the test files that don´t have any .only tests.

It currently looks like this:

mini

verbose

Feedback appreciated :)

### DIFF
--- a/lib/reporters/mini.js
+++ b/lib/reporters/mini.js
@@ -216,7 +216,11 @@ MiniReporter.prototype.finish = function (runStatus) {
 	}
 
 	if (runStatus.failFastEnabled === true) {
-		status += '\n\n  ' + colors.failFast('`--fail-fast` is on. Any number of tests may have been skipped');
+		status += '\n\n  ' + colors.information('`--fail-fast` is on. Any number of tests may have been skipped');
+	}
+
+	if (runStatus.hasExclusive === true && runStatus.remainingCount > 0) {
+		status += '\n\n  ' + colors.information('The .only() modifier is used in some tests.', runStatus.remainingCount, plur('test', runStatus.remainingCount), plur('was', 'were', runStatus.remainingCount), 'not run.');
 	}
 
 	return status + '\n\n';
--- a/lib/reporters/verbose.js
+++ b/lib/reporters/verbose.js
@@ -115,7 +115,11 @@ VerboseReporter.prototype.finish = function (runStatus) {
 	}
 
 	if (runStatus.failFastEnabled === true) {
-		output += '\n\n\n  ' + colors.failFast('`--fail-fast` is on. Any number of tests may have been skipped');
+		output += '\n\n\n  ' + colors.information('`--fail-fast` is on. Any number of tests may have been skipped');
+	}
+
+	if (runStatus.hasExclusive === true && runStatus.remainingCount > 0) {
+		output += '\n\n\n  ' + colors.information('The .only() modifier is used in some tests.', runStatus.remainingCount, plur('test', runStatus.remainingCount), plur('was', 'were', runStatus.remainingCount), 'not run.');
 	}
 
 	return output + '\n';
--- a/lib/colors.js
+++ b/lib/colors.js
@@ -10,5 +10,5 @@ module.exports = {
 	duration: chalk.gray.dim,
 	errorStack: chalk.gray,
 	stack: chalk.red,
-	failFast: chalk.magenta
+	information: chalk.magenta
 };
--- a/lib/main.js
+++ b/lib/main.js
@@ -55,7 +55,7 @@ function exit() {
 
 globals.setImmediate(() => {
 	const hasExclusive = runner.tests.hasExclusive;
-	const numberOfTests = runner.tests.tests.concurrent.length + runner.tests.tests.serial.length;
+	const numberOfTests = runner.tests.testCount;
 
 	if (numberOfTests === 0) {
 		send('no-tests', {avaRequired: true});
--- a/lib/run-status.js
+++ b/lib/run-status.js
@@ -45,6 +45,7 @@ class RunStatus extends EventEmitter {
 		this.failCount = 0;
 		this.fileCount = 0;
 		this.testCount = 0;
+		this.remainingCount = 0;
 		this.previousFailCount = 0;
 		this.knownFailures = [];
 		this.errors = [];
@@ -89,13 +90,8 @@ class RunStatus extends EventEmitter {
 	handleStats(stats) {
 		this.emit('stats', stats, this);
 
-		if (this.hasExclusive && !stats.hasExclusive) {
-			return;
-		}
-
-		if (!this.hasExclusive && stats.hasExclusive) {
+		if (stats.hasExclusive) {
 			this.hasExclusive = true;
-			this.testCount = 0;
 		}
 
 		this.testCount += stats.testCount;
@@ -139,6 +135,7 @@ class RunStatus extends EventEmitter {
 		this.skipCount = sum(this.stats, 'skipCount');
 		this.todoCount = sum(this.stats, 'todoCount');
 		this.failCount = sum(this.stats, 'failCount');
+		this.remainingCount = this.testCount - this.passCount - this.failCount - this.skipCount - this.todoCount - this.knownFailureCount;
 	}
 }
 
--- a/lib/test-collection.js
+++ b/lib/test-collection.js
@@ -10,6 +10,7 @@ class TestCollection extends EventEmitter {
 		super();
 
 		this.hasExclusive = false;
+		this.testCount = 0;
 
 		this.tests = {
 			concurrent: [],
@@ -65,6 +66,8 @@ class TestCollection extends EventEmitter {
 			this.hooks[type + (metadata.always ? 'Always' : '')].push(test);
 			return;
 		}
+
+		this.testCount++;
 
 		// Add `.only()` tests if `.only()` was used previously
 		if (this.hasExclusive && !metadata.exclusive) {
--- a/test/api.js
+++ b/test/api.js
@@ -33,7 +33,7 @@ test('Without Pool: test file with exclusive tests causes non-exclusive tests in
 	return api.run(files)
 		.then(result => {
 			t.ok(result.hasExclusive);
-			t.is(result.testCount, 2);
+			t.is(result.testCount, 5);
 			t.is(result.passCount, 2);
 			t.is(result.failCount, 0);
 		});
@@ -48,7 +48,7 @@ test('Without Pool: test files can be forced to run in exclusive mode', t => {
 		{runOnlyExclusive: true}
 	).then(result => {
 		t.ok(result.hasExclusive);
-		t.is(result.testCount, 0);
+		t.is(result.testCount, 1);
 		t.is(result.passCount, 0);
 		t.is(result.failCount, 0);
 	});
--- a/test/reporters/mini.js
+++ b/test/reporters/mini.js
@@ -424,7 +424,7 @@ test('results when fail-fast is enabled', function (t) {
 	compareLineOutput(t, output, [
 		'',
 		'',
-		'  ' + colors.failFast('`--fail-fast` is on. Any number of tests may have been skipped')
+		'  ' + colors.information('`--fail-fast` is on. Any number of tests may have been skipped')
 	]);
 	t.end();
 });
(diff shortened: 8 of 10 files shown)


mightyiam:
> If there are test files without .only in them, it will print out the following.

> > The .only() modifier is used in some tests. At least', remainingTests, 'tests were not run.

Are you sure you got this part correct? I don't see the sense in it.

ThomasBem:
You mean you think the implementation is wrong? Or do you not think there is any value in showing the remaining count?

What is happening in the case of the screenshots is that there are two test files being run, test.js and test2.js. There is one .only() test in test.js, and none in test2.js.

So now you are at least notified of all the tests in test2.js that has not been run because of the .only() tests.

If there was at least one .only in test2.js also. It would just report "all other tests have been skipped".

mightyiam:
I get it now. The first quote is the current message (result of your previous PR). I did not understand this, at first.

BTW, why do you put quotes around the `-v`?

lib/reporters/mini.js
```diff
@@ -216,7 +216,16 @@ MiniReporter.prototype.finish = function (runStatus) {
...
 
 	if (runStatus.failFastEnabled === true) {
-		status += '\n\n  ' + colors.failFast('`--fail-fast` is on. Any number of tests may have been skipped');
+		status += '\n\n  ' + colors.information('`--fail-fast` is on. Any number of tests may have been skipped');
+	}
+
+	if (runStatus.hasExclusive === true) {
+		var remainingTests = runStatus.testCount - (runStatus.passCount + runStatus.failCount);
+		if (remainingTests > 0) {
+			status += '\n\n  ' + colors.information('The .only() modifier is used in some tests. At least', remainingTests, 'tests were not run.');
```
mightyiam: Use native JavaScript template string instead of some library's API, I think.

lib/reporters/mini.js
```diff
@@ -216,7 +216,16 @@ MiniReporter.prototype.finish = function (runStatus) {
...
+		} else {
+			status += '\n\n  ' + colors.information('The .only() modifier is used in some tests. All other tests have been skipped.');
```
mightyiam: DRY first sentence.

lib/reporters/verbose.js
```diff
@@ -115,7 +115,16 @@ VerboseReporter.prototype.finish = function (runStatus) {
...
 
 	if (runStatus.failFastEnabled === true) {
-		output += '\n\n\n  ' + colors.failFast('`--fail-fast` is on. Any number of tests may have been skipped');
+		output += '\n\n\n  ' + colors.information('`--fail-fast` is on. Any number of tests may have been skipped');
+	}
+
+	if (runStatus.hasExclusive === true) {
+		var remainingTests = runStatus.testCount - (runStatus.passCount + runStatus.failCount);
+		if (remainingTests > 0) {
+			output += '\n\n\n  ' + colors.information('The .only() modifier is used in some tests. At least', remainingTests, 'tests were not run.');
```
mightyiam: You have some shared strings here. Shared with the mini reporter.

lib/reporters/verbose.js
```diff
@@ -115,7 +115,16 @@ VerboseReporter.prototype.finish = function (runStatus) {
...
+		} else {
+			output += '\n\n\n  ' + colors.information('The .only() modifier is used in some tests. All other tests have been skipped.');
```
mightyiam: Shared string.

lib/reporters/verbose.js
mightyiam: And template string.

ThomasBem:
Thanks for the review! 👍 

I am not familiar with template strings, but I think the implementation I have done with regards to  string formatting is in-line / same styles as what was already in place in the reporters.

I personally don´t think there is a problem with having the same output strings repeated across the two different reporters. They are two separate pieces that in theory could output different things. It also makes it easy to read what the reporter outputs without having to check some global configuration property.

But I am up for making changes if more people think its a better solution :)

@mightyiam Could you provide an example of how my two lines of output would look using template strings?

lib/reporters/mini.js
novemberborn: Meh, it still needs to be wrapped with `colors.information()`, so this combination seems fine to me.

Previously we were targeting Node.js 0.10, which meant that we couldn't use template strings. Now we can, but it's no big deal if new code follows the pattern already used in the file.

lib/reporters/verbose.js
novemberborn: > You have some shared strings here. Shared with the mini reporter.

I find that to be the kind of repetition that is too bothersome to dry up.

lib/reporters/mini.js
```diff
@@ -216,7 +216,16 @@ MiniReporter.prototype.finish = function (runStatus) {
 	}
 
 	if (runStatus.failFastEnabled === true) {
-		status += '\n\n  ' + colors.failFast('`--fail-fast` is on. Any number of tests may have been skipped');
+		status += '\n\n  ' + colors.information('`--fail-fast` is on. Any number of tests may have been skipped');
+	}
+
+	if (runStatus.hasExclusive === true) {
+		var remainingTests = runStatus.testCount - (runStatus.passCount + runStatus.failCount);
```
novemberborn: This should be computed in `lib/run-status.js`, which @jarlehansen is working on in #1179. Note that as I commented there (https://github.com/avajs/ava/pull/1179#discussion_r95819806) you need to include other test states in this count, see the highlighted lines in https://github.com/avajs/ava/blob/033d4dcdcbdadbf665c740ff450c2a775a8373dc/lib/run-status.js#L41:L45.

ThomasBem: Will sync up with @jarlehansen regarding this and update PR.

ThomasBem:
@novemberborn updated PR with the changes to remainingCount in run-status.js and added a test for it like you suggested to @jarlehansen in his PR. Had a quick chat with him and he would get on later today to take a look.

So you guys can review my suggestion and see what you think :)

jarlehansen:
Thanks @ThomasBem, this improves on what I created in my pull request. When this is merged I will update #1179

novemberborn:
@ThomasBem @jarlehansen thanks for your collaboration 💯 

I found one edge case. Tests in the file with `.only()` are not counted. This is because the test count [is based on the scheduled tests](https://github.com/avajs/ava/blob/ce42fcb7741039eeb26f0a167b27fd5923936a1b/lib/main.js#L58) (`main.js`), and [non-exclusive tests are discarded](https://github.com/avajs/ava/blob/ce42fcb7741039eeb26f0a167b27fd5923936a1b/lib/test-collection.js#L69:L78) (`test-collection.js`). This is why, before this PR, [`testCount` was reset](https://github.com/avajs/ava/blob/ce42fcb7741039eeb26f0a167b27fd5923936a1b/lib/run-status.js#L98) (`run-status.js`).

I think we should count *all* tests in `test-collection.js` and use that value in `main.js`.

ThomasBem:
Yeah, I knew about that behaviour and I tried explaining it in my original PR text. But guess I did not get that through properly, or at least i should have mentioned it more specific :)

> If all test files have at least one .only test in them, it will print out the following message at the end of the report for both mini and verbose.

> The .only() modifier is used in some tests. All other tests have been skipped.

 Thats the reason its split into two. Because when .only is present it did not count all the other tests.

Did not feel comfortable with changing some of the deeper behaviour of how tests where added and removed from collections.

But if you think its a good idea to change the behaviour to count all the tests even if there is .only tests present. Then i guess we can simplify the logic you linked to in test-collection.js similar to what was done run-status.js regarding exclusive?

novemberborn:
> Did not feel comfortable with changing some of the deeper behaviour of how tests where added and removed from collections.
> 
> But if you think its a good idea to change the behaviour to count all the tests even if there is .only tests present. Then i guess we can simplify the logic you linked to in test-collection.js similar to what was done run-status.js regarding exclusive?

Fair enough. But yes, let's change it. We should keep the counting consistent, regardless of whether `.only` was encountered in the test file being collected, or in a later test file.

ThomasBem:
@novemberborn I got some results that confused me a bit when trying to fix some failing tests after implementing what I think would be a solution for test-collection.js.

Then I read this in the description of one of the api.js tests that was failing, https://github.com/avajs/ava/blob/ce42fcb7741039eeb26f0a167b27fd5923936a1b/test/api.js#L22

Is this correct? That it will still run tests that are **NOT** tagged with .only() if they are in a file that has at least one .only() test?

I am still reading up on the code and trying to see if its just me that´s missing something obvious and thats why things are acting seemingly a bit weird. But figured I would at least ask for a clarification regarding this one thing :)

ThomasBem:
Think I figured out what i was doing wrong on my own :) Was just doing things way too complicated... 

But this wording is still somewhat confusing to me. Because it does not run any other tests than the .only() as far as i can tell?

ThomasBem:
@novemberborn PR updated with requested changes. It now properly counts all tests, exclusive or not. Also made some changes to the reporters output to be more in line with #1179 that @jarlehansen is working on.

Now it only reports on .only() / exclusive if there are tests that have not been run.

Any thoughts on the output text? 

> The .only() modifier is used in some tests. 1 test(s) were not run.

A bit clumsy with the "test(s)", but i couldn't think of anything better to cover 1 or multiple tests remaining.

mightyiam:
"1 tests were not run" is fine with me.

novemberborn:
@ThomasBem 

> … I read this in the description of one of the api.js tests that was failing, https://github.com/avajs/ava/blob/ce42fcb7741039eeb26f0a167b27fd5923936a1b/test/api.js#L22
> 
> Is this correct? That it will still run tests that are NOT tagged with .only() if they are in a file that has at least one .only() test?
>
> … this wording is still somewhat confusing to me. Because it does not run any other tests than the .only() as far as i can tell?

That's not how I read it. It's meant to say that if there are two test files, and only one contains exclusive tests, then none of the tests from the other file are run.

lib/reporters/mini.js
```diff
@@ -216,7 +216,11 @@ MiniReporter.prototype.finish = function (runStatus) {
 	}
 
 	if (runStatus.failFastEnabled === true) {
-		status += '\n\n  ' + colors.failFast('`--fail-fast` is on. Any number of tests may have been skipped');
+		status += '\n\n  ' + colors.information('`--fail-fast` is on. Any number of tests may have been skipped');
+	}
+
+	if (runStatus.hasExclusive === true && runStatus.remainingCount > 0) {
+		status += '\n\n  ' + colors.information('The .only() modifier is used in some tests.', runStatus.remainingCount, 'test(s) were not run.');
```
novemberborn: [Use `plur()`](https://github.com/avajs/ava/blob/23ffce31def7d960e9b39d38a95dbd3b5f103f81/lib/reporters/verbose.js#L79:L86), so "1 test was" and "2 tests were".

ThomasBem:
PR updated again :)

novemberborn:
Great work @ThomasBem, keep it up 👍

ThomasBem:
Thanks for the support! 👍

lib/reporters/mini.js
```diff
@@ -216,7 +216,11 @@ MiniReporter.prototype.finish = function (runStatus) {
 	}
 
 	if (runStatus.failFastEnabled === true) {
-		status += '\n\n  ' + colors.failFast('`--fail-fast` is on. Any number of tests may have been skipped');
+		status += '\n\n  ' + colors.information('`--fail-fast` is on. Any number of tests may have been skipped');
+	}
+
+	if (runStatus.hasExclusive === true && runStatus.remainingCount > 0) {
+		status += '\n\n  ' + colors.information('The .only() modifier is used in some tests.', runStatus.remainingCount, plur('test', runStatus.remainingCount), plur('was', 'were', runStatus.remainingCount), 'not run.');
```
novemberborn: @ThomasBem this is very petty, but I just noticed there's a period at the end of the message here (and in the verbose reporter). None of the other messages seem to end in a period. If you've got the time could you do a PR to remove them? 😄

ThomasBem: @novemberborn Sure, I can fix that :)

PR #1179, opened 2017-01-11
Only print fail-fast when there are remaining tests

### LINKED ISSUE #1171: Don't print --fail-fast reminder if no tests were actually skipped
With #1160 AVA now prints a reminder when `--fail-fast` is used and a test fails. Any number of tests could have been skipped.

If all other tests have indeed completed then the reminder is not necessary. A `remainingCount` could be computed and guarded against when printing the reminder, as described in https://github.com/avajs/ava/issues/1134#issuecomment-267961180.

jarlehansen:
This PR tries to fix #1171 by calculating the remaningCount in run-status.js
And then check the value in the reporters, if there are no remaining test it will not print:

> `--fail-fast` is on. Any number of tests may have been skipped

### DIFF
--- a/lib/reporters/mini.js
+++ b/lib/reporters/mini.js
@@ -215,7 +215,7 @@ MiniReporter.prototype.finish = function (runStatus) {
 		});
 	}
 
-	if (runStatus.failFastEnabled === true) {
+	if (runStatus.failFastEnabled === true && runStatus.remainingCount > 0 && runStatus.failCount > 0) {
 		status += '\n\n  ' + colors.information('`--fail-fast` is on. Any number of tests may have been skipped');
 	}
 
--- a/test/reporters/mini.js
+++ b/test/reporters/mini.js
@@ -417,6 +417,8 @@ test('results with unhandled errors', function (t) {
 test('results when fail-fast is enabled', function (t) {
 	var reporter = miniReporter();
 	var runStatus = {
+		remainingCount: 1,
+		failCount: 1,
 		failFastEnabled: true
 	};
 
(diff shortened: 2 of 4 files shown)


lib/run-status.js
```diff
@@ -139,6 +140,7 @@ class RunStatus extends EventEmitter {
 		this.skipCount = sum(this.stats, 'skipCount');
 		this.todoCount = sum(this.stats, 'todoCount');
 		this.failCount = sum(this.stats, 'failCount');
+		this.remainingCount = (this.testCount - (this.passCount + this.failCount));
```
novemberborn: `this.testCount - this.passCount - this.failCount` is clearer.

You need to include other test states in this count, see the highlighted lines in https://github.com/avajs/ava/blob/033d4dcdcbdadbf665c740ff450c2a775a8373dc/lib/run-status.js#L41:L45.

The computation should be tested. That's probably best in `test/run-status.js`, though most of its behavior ends up being tested elsewhere.

test/reporters/mini.js
```diff
@@ -429,6 +430,18 @@ test('results when fail-fast is enabled', function (t) {
 	t.end();
 });
 
+test('results without fail-fast if no skipped tests', function (t) {
+	var reporter = miniReporter();
+	var runStatus = {
+		remainingCount: 0,
```
novemberborn: I think this needs `failCount: 1`, like with the verbose test?

jarlehansen:
Thanks for the comments, I will wait on #1177 to be merged and then update this PR.

jarlehansen:
I have updated this PR with the changes done by @ThomasBem

lib/reporters/mini.js
```diff
@@ -215,7 +215,7 @@ MiniReporter.prototype.finish = function (runStatus) {
 		});
 	}
 
-	if (runStatus.failFastEnabled === true) {
+	if (runStatus.failFastEnabled === true && runStatus.remainingCount > 0) {
```
novemberborn: This should check if there were any failures. Otherwise the warning is also printed when `.only()` is used and `--fail-fast` is on:

```
$ "$(npm bin)"/ava --fail-fast

  1 passed

  `--fail-fast` is on. Any number of tests may have been skipped

  The .only() modifier is used in some tests. 135 tests were not run.
```

sindresorhus:
Thank you @jarlehansen. This is excellent 👌😄

PR #1184, opened 2017-01-13
Resolve relative modules in profile.js just like with CLI/API

### LINKED ISSUE #1183: requires in config not used by profile.js
## Description
#1070 removed the CLI option but also made it not possible to run/debug with ```profile.js``` like

```
inspect node_modules/ava/profile.js src/MyComponent.spec.js
```

Furthermore, while relative paths work when using the CLI, they don't work when using the non-forking test runner ```profile.js```

I'll attach a PR that has a fix.  It duplicates ```resolveModules()``` from ```api.js``` which isn't ideal. I'm definitely open to suggestions if that function should be moved to a common module.

## Configuration
```json
  "ava": {
    "babel": {
      "babelrc": false,
      "presets": [
        "es2015",
        "es2016",
        "react"
      ]
    },
    "concurrency": 6,
    "verbose": true,
    "files": [
      "./src/**/*.spec.js",
      "./src**/*.spec.js"
    ],
    "require": [
      "babel-core/register",
      "./scripts/ava-setup.js"
    ]
  }
```

gconaty:
Fixes #1183

### DIFF
--- a/profile.js
+++ b/profile.js
@@ -11,8 +11,22 @@ const Promise = require('bluebird');
 const pkgConf = require('pkg-conf');
 const findCacheDir = require('find-cache-dir');
 const uniqueTempDir = require('unique-temp-dir');
+const arrify = require('arrify');
+const resolveCwd = require('resolve-cwd');
 const CachingPrecompiler = require('./lib/caching-precompiler');
 const globals = require('./lib/globals');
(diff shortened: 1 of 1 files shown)


novemberborn:
Thanks @gconaty!

PR #1196, opened 2017-01-18
Move the `ava-files` package back into core

### LINKED ISSUE #1192: Move ava-files back into core
In #925 we extracted AVA's file logic into its own package: https://github.com/avajs/ava-files. In hindsight this was premature and we now want to move it back.

This should be reasonably straight-forward, however the tests need to be backported from AVA to `tap`.

See also https://github.com/avajs/ava-files/issues/11.

forresst:
Fixes #1192


package.json
```diff
@@ -44,7 +44,7 @@
   },
   "scripts": {
     "test": "xo && nyc tap --no-cov --timeout=150 test/*.js test/reporters/*.js",
-    "test-win": "tap --no-cov --reporter=classic --timeout=150 test/*.js test/reporters/*.js",
+    "test-win": "tap --no-cov --reporter=classic --timeout=150 test/ava-files.js test/reporters/*.js",
```
sindresorhus: You forgot to change this back ;)

test/ava-files.js
```diff
@@ -0,0 +1,180 @@
+const path = require('path');
```
sindresorhus: Put `'use strict';` above here.

package.json
forresst: Oups ! I'm unmasked, I test under windows!

test/ava-files.js
forresst: Yep Sir !

sindresorhus:
There are some lint issues you need to fix. See Travis.

sindresorhus:
Some tests are failing: https://travis-ci.org/avajs/ava/jobs/193048227

forresst:
@sindresorhus I will look at these tests quickly. Sorry

There is a second point to consider: spread for Node 4 (https://travis-ci.org/avajs/ava/jobs/193048228#L231). 
I know spread does not work without the flag under node 4. It works on the avajs/ava-files repository but not under avajs/ava. Is there a tip or parameter under travis for this to work?
Finally, should we use spread in avajs/ava?

sindresorhus:
No, only use syntax supported on Node.js 4.

novemberborn:
> It works on the avajs/ava-files repository but not under avajs/ava.

I assume it's used in the old tests? We're using AVA to run those tests, and of course AVA takes care of transpiling the spread operator. That's not the case with tap.

forresst:
@sindresorhus ok, i will remove spread

@novemberborn Obviously, how could I miss out on this? I will read the doc in French! 😄

forresst:
All tests passing. Sorry for my mistakes.

mightyiam:
> No, only use syntax supported on Node.js 4.

:1st_place_medal: tough leadership decisions

novemberborn:
This is awesome. Thanks @forresst!

Issue #1195, opened 2017-01-18
Create a Babel preset with AVA-specific transforms


novemberborn:
Per [RFC001](https://github.com/avajs/ava/blob/master/docs/specs/001%20-%20Improving%20language%20support.md) we're placing AVA-specific Babel transforms into its own preset. 

This makes it possible for users to set up a custom Babel pipeline for tests, yet still apply AVA's transforms, without having to know specifics.

It should also be possible to apply these transforms in a build step, e.g. when publishing a package containing reusable tests or helpers.

novemberborn:
Landed in #1197.

Issue #1193, opened 2017-01-17
Update our Babel usage to only apply stage-4 transforms


novemberborn:
Our current Babel usage is a bit confusing, and applies too many transforms (stage-2 even!): https://github.com/avajs/ava/blob/a380fe4f5ffa1d334a2a3786215d3c1ff10a9898/lib/babel-config.js#L43:L48

Per [RFC001](https://github.com/avajs/ava/blob/master/docs/specs/001%20-%20Improving%20language%20support.md) we will only apply stage-4 transforms. We must only apply transforms for features that are not natively supported on Node.js 4 and 6 respectively.  We must not polyfill the environment.

I'm thinking we may have to roll our own preset to do this properly, given that https://www.npmjs.com/package/babel-preset-es2015-node4 is deprecated in favor of https://github.com/babel/babel-preset-env, which pulls in unnecessary dependencies. https://www.npmjs.com/package/babel-preset-node6, confusingly, talks about ES2015 yet we need ES2017.

sindresorhus:
Yeah. We need a custom preset.

mightyiam:
I'm looking forward to this. Debugging should be more clear without unnecessary polyfills.

Did we think about automatically detecting target?

And what's wrong with stage 2? No one has to actually use any proposed future...

novemberborn:
> And what's wrong with stage 2? No one has to actually use any proposed future...

But they might, accidentally.

When AVA started the language features that are now in ES2017 and earlier where not even shipping in Node.js. That's completely changed now, Node.js 8 should have full ES2017 support I think. AVA can still focus on making these features available to older runtimes but it doesn't need to advocate for features in earlier stages of development. This also aligns us with ESLint, which only supports stage-4 features.

mightyiam:
Alright. What about target detection? We *are* on target runtime. So why don't we detect and set the target accordingly?

novemberborn:
> What about target detection? We are on target runtime. So why don't we detect and set the target accordingly?

That's the plan.

Issue #1089, opened 2016-10-21
Document how babel-plugin-transform-runtime changes references to builtins as a pitfall


dwieeb:
I am trying to use reflection and Typescript decorators. My library uses the `reflect-metadata` polyfill. The `babel-plugin-transform-runtime` plugin is clobbering the references to `Reflect` in my test files, resulting in two reflection registries and thus breaking reflection between my library and the test file. I am unable to test decorators from my library. I made a simple repo to illustrate the issue: https://github.com/dwieeb/ava-decorators

I had to use ava from master in order to debug tests in subprocesses and ultimately see that babel is wiping out the global `Reflect` and also substituting my polyfill for some `core-js` thing. 

node 6
npm 3

relevant, unresolved: https://github.com/avajs/ava/issues/947

novemberborn:
> relevant, unresolved: #947

That would be the solution, yes.

Meanwhile you could try something like this: https://github.com/novemberborn/buoyant/blob/4015fd703a4f26a4965f12a60c221fbb372ed3b3/test/lib/Scheduler.js#L11:L12

Let me know how that works for you and I'll see if we can this to the [Common Pitfalls](https://github.com/avajs/ava/blob/master/docs/common-pitfalls.md) documentation.

dwieeb:
Yes, that works. My tests are working as expected now.

lholznagel:
@dwieeb hey, just came across the same issue. What did you do for it to work? Adding `const { Reflect } = global` seems not to work.

Thanks in advance :)

dwieeb:
@lholznagel I did this, but I think `const { Reflect } = global` should work, too(?):

```typescript
const g = <any>global;
const Reflect = g.Reflect;
```

lholznagel:
Ah thanks!
With my code I get this error:
`'Global' has no property 'Reflect' and no string index signature`

Really didn´t think about adding `<any>`. When adding `<any>` it works :+1:
End result looks like this: `const { Reflect } = <any> global;`

My test now works, thanks for your help :+1:

novemberborn:
As of #1197 `master` no longer applies the runtime transform.

PR #1197, opened 2017-01-18
Extract Babel presets

### LINKED ISSUE #947: Option to disable babel-plugin-transform-runtime?
I am porting a Node.js v6 project to use AVA. I can already specify just the Babel plugins I need for my tests to run:

``` json
  "ava": {
    "babel": {
      "plugins": [
        "transform-async-to-generator",
        "transform-es2015-modules-commonjs"
      ]
    },
  }
```

However AVA will always add the `babel-plugin-transform-runtime` plugin. It's not really needed for my use case and actually causes a test to break:

I'm testing whether a module returns a promise, using `t.true(thingThatReturnsAPromise() instanceof Promise)`. The `babel-plugin-transform-runtime` plugin replaces `Promise` within the test with a `core-js` implementation, but the module returns a native promise. Hence `instanceof` fails.

It'd be useful if the `babel-plugin-transform-runtime` could be disabled. Perhaps like this:

``` json
"ava": {
  "babel": {
    "disable-transform-runtime": true
```

The above would leave AVA's Babel config in default mode, but disable the plugin.

To use it with `inherit` mode:

``` json
"ava": {
  "babel": {
    "inherit": true,
    "disable-transform-runtime": true
```

Or to use it with a custom config:

``` json
"ava": {
  "babel": {
    "disable-transform-runtime": true,
    "plugins": [
      "transform-async-to-generator",
      "transform-es2015-modules-commonjs"
    ]
```

Thoughts?

novemberborn:
Removes deprecated assertions (breaking!)

Uses the new Babel presets (#1193, #1195). Will publish those dependencies and update the `package.json` references when CI passes.

Since this removes `babel-runtime` we can close #947 and #1089.


lib/babel-config.js
```diff
@@ -25,68 +24,12 @@ function validate(conf) {
...
-		}
-
-		return value;
+function makeTransformTestFiles(powerAssert) {
+	return babel => {
+		return require('@ava/babel-preset-transform-test-files')(babel, {powerAssert});
 	};
 }
 
-const defaultPresets = lazy(() => {
```
sindresorhus: @novemberborn Do you remember why we have laziness here? There must be a reason. Maybe we should preserve the laziness?

novemberborn: It seemed unnecessary yesterday but I think I figured it out. This config is used even if all transforms have been cached. So we should keep the laziness in place.

novemberborn: Restored the lazy behavior, though in a somewhat simplified form.

tylerjharden:
So this is good to go? And fixes "warning ava > babel-preset-es2015-node4@2.1.1: Use npmjs.org/babel-preset-env instead, see https://github.com/babel/babel-preset-env"?

Very exciting if so.

sindresorhus:
@novemberborn I tried this PR on some of my projects, like Got, Pageres, etc, and seems to be working well.

sindresorhus:
@tylerjharden Yes, it removes the warning. It's good to go when it's done.

novemberborn:
> It's good to go when it's done.

@sindresorhus Could you approve it then? ;-) CI failure is just AppVeyor being confused.

sindresorhus:
@novemberborn Don't you need to publish the extracted modules first and update package.json? Right now they reference Git commits.

novemberborn:
> Don't you need to publish the extracted modules first and update package.json? Right now they reference Git commits.

Hence the `DO NOT MERGE` label 😉

sindresorhus:
It's approved. :shipit: :shipit: :shipit: :shipit:

sindresorhus:
1ab7e29c-117e-11e6-9cbb-f0828bacea8e

mightyiam:
Can't wait for next release.

danez:
This PR removed the stage-2 preset and broke my test when updating as I was using object-spread. Should I change the README and remove the mention of stage-2 or was that removed by mistake?

novemberborn:
@danez yes, this sentence shouldn't mention `stage-2`:

> We use our own bundled Babel with our @ava/stage-4 and stage-2 preset, as well as custom transforms for test and helper files.

We would appreciate a PR!

Issue #1046, opened 2016-09-21
Separate config keys from CLI flags


novemberborn:
Currently the config object in the `package.json` is passed as-is to `meow`, providing default values for the CLI flags. This means that config keys need to match the CLI flags. CLI flags are singular, but sometimes config keys are better when pluralized. E.g. `sources` versus `source`, and the suggested `extension` flag in #631 versus `extensions`.

@avajs/core how about we change some of the config keys to be plurals and map them to `meow`'s default in `cli.js`?

sindresorhus:
Sounds good.

This would also be a good opportunity to document the package.json options separately instead of just referring to the CLI flags and show a JSON config dump.

omerbn:
Hey guys,
I'd like to do it but need further depiction.

novemberborn:
@omerbn AVA loads config from `package.json` here: https://github.com/avajs/ava/blob/262911f02b0857ce3873e353a547bf009bf4e795/lib/cli.js#L24

It is passed to `meow` here: https://github.com/avajs/ava/blob/262911f02b0857ce3873e353a547bf009bf4e795/lib/cli.js#L74

The goal is to pass a different object to `meow` where the keys match the long option names in the `meow` configuration, so that we can change the `package.json` config to use different key names. E.g. pass `source` to `meow` but use `sources` in `package.json`.

yatharthk:
@novemberborn what keys in particular are to be mapped in configuration object? Like the keys we want to be used as `plurals`.

novemberborn:
Hi @yatharthk. There've been a few developments since I started this issue. We won't have an `extensions` flag, the `require` flag has been removed, and we're looking to remove the `source` flag (#937).

I think a good place to start would be to stop passing `default: conf` in the call to `meow`. We could then merge the resulting `cli` object with `conf`. Hopefully that leads to the same behavior. It'll then allow us to decouple the config keys from the CLI flags where needed (though I don't think there are any such cases at the moment).

yatharthk:
@novemberborn Thanks. Shall I pick up the task of pulling out the `default: conf` when calling `meow`?  Is there a relevant issue or we need to create one?

novemberborn:
Hey @yatharthk,

> Is there a relevant issue or we need to create one?

This would be the one.

> Shall I pick up the task of pulling out the default: conf when calling meow?

Thanks for volunteering!

[...]

PR #1198, opened 2017-01-18
Added support for --no-color flag in verbose and mini reporters

ThomasBem:
Started working on fixing #843. So far I have managed to disable color output for both mini and verbose reporters by setting the --no-color flag. 

Now I need to disable color output also if you try console.log with colors inside a test. What I am a little stuck on at the moment is this spawning of child processes. Thats a part of the code base I have yet to understand.

Any pointers on where to start digging? :)

### DIFF
--- a/lib/babel-config.js
+++ b/lib/babel-config.js
@@ -1,9 +1,7 @@
 'use strict';
 const path = require('path');
-const chalk = require('chalk');
 const figures = require('figures');
 const convertSourceMap = require('convert-source-map');
-const colors = require('./colors');
 
 function validate(conf) {
 	if (conf === undefined || conf === null) {
@@ -14,9 +12,9 @@ function validate(conf) {
 	const isValidShortcut = conf === 'default' || conf === 'inherit';
 
 	if (!conf || (typeof conf === 'string' && !isValidShortcut)) {
-		let message = colors.error(figures.cross);
+		let message = figures.cross;
 		message += ' Unexpected Babel configuration for AVA. ';
-		message += 'See ' + chalk.underline('https://github.com/avajs/ava#es2015-support') + ' for allowed values.';
+		message += 'See https://github.com/avajs/ava#es2015-support for allowed values.';
 
 		throw new Error(message);
 	}
--- a/lib/babel-config.js
+++ b/lib/babel-config.js
@@ -1,7 +1,9 @@
 'use strict';
 const path = require('path');
+const chalk = require('chalk');
 const figures = require('figures');
 const convertSourceMap = require('convert-source-map');
+const colors = require('./colors');
 
 function validate(conf) {
 	if (conf === undefined || conf === null) {
@@ -12,9 +14,9 @@ function validate(conf) {
 	const isValidShortcut = conf === 'default' || conf === 'inherit';
 
 	if (!conf || (typeof conf === 'string' && !isValidShortcut)) {
-		let message = figures.cross;
+		let message = colors.error(figures.cross);
 		message += ' Unexpected Babel configuration for AVA. ';
-		message += 'See https://github.com/avajs/ava#es2015-support for allowed values.';
+		message += 'See ' + chalk.underline('https://github.com/avajs/ava#es2015-support') + ' for allowed values.';
 
 		throw new Error(message);
 	}
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -38,6 +38,7 @@ exports.run = () => {
 		  --verbose, -v           Enable verbose output
 		  --no-cache              Disable the transpiler cache
 		  --no-power-assert       Disable Power Assert
+		  --no-color              Disable color output
 		  --match, -m             Only run tests with matching title (Can be repeated)
 		  --watch, -w             Re-run tests when tests and source files change
 		  --timeout, -T           Set global timeout
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -68,9 +68,10 @@ exports.run = () => {
 			'tap',
 			'verbose',
 			'watch',
-			'update-snapshots'
+			'update-snapshots',
+			'color'
 		],
-		default: conf,
+		default: Object.assign({color: true}, conf),
 		alias: {
 			t: 'tap',
 			v: 'verbose',
@@ -118,7 +119,8 @@ exports.run = () => {
 		pkgDir,
 		timeout: cli.flags.timeout,
 		concurrency: cli.flags.concurrency ? parseInt(cli.flags.concurrency, 10) : 0,
-		updateSnapshots: cli.flags.updateSnapshots
+		updateSnapshots: cli.flags.updateSnapshots,
+		color: cli.flags.color
 	});
 
 	let reporter;
@@ -126,9 +128,9 @@ exports.run = () => {
 	if (cli.flags.tap && !cli.flags.watch) {
 		reporter = new TapReporter();
 	} else if (cli.flags.verbose || isCi) {
-		reporter = new VerboseReporter({basePath: pkgDir});
+		reporter = new VerboseReporter({color: cli.flags.color, basePath: pkgDir});
 	} else {
-		reporter = new MiniReporter({watching: cli.flags.watch, basePath: pkgDir});
+		reporter = new MiniReporter({color: cli.flags.color, watching: cli.flags.watch, basePath: pkgDir});
 	}
 
 	reporter.api = api;
--- a/lib/fork.js
+++ b/lib/fork.js
@@ -36,7 +36,9 @@ module.exports = (file, opts, execArgv) => {
 		} : false
 	}, opts);
 
-	const ps = childProcess.fork(path.join(__dirname, 'test-worker.js'), [JSON.stringify(opts)], {
+	const args = [JSON.stringify(opts), opts.color ? '--color' : '--no-color'];
+
+	const ps = childProcess.fork(path.join(__dirname, 'test-worker.js'), args, {
 		cwd: opts.pkgDir,
 		silent: true,
 		env,
--- a/lib/reporters/mini.js
+++ b/lib/reporters/mini.js
@@ -42,11 +42,16 @@ function eraseLines(count) {
 
 class MiniReporter {
 	constructor(options) {
+		this.options = Object.assign({}, options);
+
+		chalk.enabled = this.options.color;
+		for (const key of Object.keys(colors)) {
+			colors[key].enabled = this.options.color;
+		}
+
 		const spinnerDef = spinners[process.platform === 'win32' ? 'line' : 'dots'];
 		this.spinnerFrames = spinnerDef.frames.map(c => chalk.gray.dim(c));
 		this.spinnerInterval = spinnerDef.interval;
-
-		this.options = Object.assign({}, options);
 
 		this.reset();
 		this.stream = process.stderr;
--- a/lib/reporters/mini.js
+++ b/lib/reporters/mini.js
@@ -14,11 +14,6 @@ const extractStack = require('../extract-stack');
 const codeExcerpt = require('../code-excerpt');
 const colors = require('../colors');
 
-chalk.enabled = true;
-Object.keys(colors).forEach(key => {
-	colors[key].enabled = true;
-});
-
 // TODO(@jamestalamge): This should be fixed in log-update and ansi-escapes once we are confident it's a good solution.
 const CSI = '\u001b[';
 const ERASE_LINE = CSI + '2K';
--- a/lib/reporters/verbose.js
+++ b/lib/reporters/verbose.js
@@ -10,13 +10,14 @@ const extractStack = require('../extract-stack');
 const codeExcerpt = require('../code-excerpt');
 const colors = require('../colors');
 
-Object.keys(colors).forEach(key => {
-	colors[key].enabled = true;
-});
-
 class VerboseReporter {
 	constructor(options) {
 		this.options = Object.assign({}, options);
+
+		chalk.enabled = options.color;
+		for (const key of Object.keys(colors)) {
+			colors[key].enabled = options.color;
+		}
 	}
 	start() {
 		return '';
--- a/readme.md
+++ b/readme.md
@@ -163,6 +163,7 @@ $ ava --help
     --verbose, -v           Enable verbose output
     --no-cache              Disable the transpiler cache
     --no-power-assert       Disable Power Assert
+    --no-color              Disable color output
     --match, -m             Only run tests with matching title (Can be repeated)
     --watch, -w             Re-run tests when tests and source files change
     --timeout, -T           Set global timeout
--- a/test/reporters/mini.js
+++ b/test/reporters/mini.js
@@ -27,6 +27,9 @@ process.stdout.columns = 5000;
 const fullWidthLine = chalk.gray.dim('\u2500'.repeat(5000));
 
 function miniReporter(options) {
+	if (options === undefined) {
+		options = {color: true};
+	}
 	const reporter = new MiniReporter(options);
 	reporter.start = () => '';
 	return reporter;
@@ -35,7 +38,7 @@ function miniReporter(options) {
 process.stderr.setMaxListeners(50);
 
 test('start', t => {
-	const reporter = new MiniReporter();
+	const reporter = new MiniReporter({color: true});
 
 	t.is(reporter.start(), ' \n ' + graySpinner + ' ');
 	reporter.clearInterval();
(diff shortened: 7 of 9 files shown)


lib/cli.js
```diff
@@ -124,9 +125,14 @@ exports.run = () => {
 	if (cli.flags.tap && !cli.flags.watch) {
 		reporter = tapReporter();
 	} else if (cli.flags.verbose || isCi) {
-		reporter = verboseReporter();
+		reporter = verboseReporter({
+			disableColors: cli.flags.color !== undefined
```
novemberborn: Without looking into `meow`, I'm surprised this isn't a boolean. Does it become `false` if `--no-color` is set? Comparing against `false` seems better than comparing against `undefined`.

lib/reporters/mini.js
```diff
@@ -28,6 +28,13 @@ function MiniReporter(options) {
 
 	this.options = Object.assign({}, options);
 
+	if (this.options.disableColors) {
+		chalk.enabled = false;
+		Object.keys(colors).forEach(function (key) {
```
novemberborn: I'd personally prefer a `for…of` loop here.

novemberborn:
> Now I need to disable color output also if you try console.log with colors inside a test

That seems unnecessary. If you want to log colors then go ahead. It's just that one should be able to disable AVA's color output.

lib/reporters/mini.js
ThomasBem: I just re-used the existing code that in other places did this but with "=true". But can make changes if you think that`s better.

https://github.com/ThomasBem/ava/blob/7c00c13844da6ab116676027b6482a57cf0f102a/lib/reporters/verbose.js#L9:L11

https://github.com/ThomasBem/ava/blob/7c00c13844da6ab116676027b6482a57cf0f102a/lib/reporters/mini.js#L14:L16

ThomasBem:
@sindresorhus commented this in one of the PR`s that was working on this issue previously.

> It also needs to pass the --no-color flag down to the forked process. If I add console.log(require('chalk').red('colorzzzz')); in a test file, it will still log colors even with $ ava --no-colors. I would expect that to prevent all colors.

https://github.com/avajs/ava/pull/1104

lib/reporters/mini.js
sindresorhus: Yes, go with for-of whenever possible.

lib/cli.js
sindresorhus: @novemberborn By default, if the user doesn't specify either `--color` or `--no-color`, it's not set. 

If we add the following, it will be `true` if not specified, but then we wouldn't know when it's not specified and should be auto-detected:

```js
const cli = meow(``, {
	boolean: ['color'],
	default: {
		color: true
	}
});
```

sindresorhus: I would go with `color` instead of `disableColors` though. Double-negatives are not very good.

ThomasBem: I will double check this. Was stuck on this point for a while trying out different things and reading up on meow and minimist. Got it working on undefined, but maybe I missed something.

novemberborn:
@ThomasBem I guess I disagree with @sindresorhus then 😄 

@sindresorhus why do you think `ava --no-color` should disable colors not coming from AVA itself?

sindresorhus:
> why do you think ava --no-color should disable colors not coming from AVA itself?

Because that's how it would work normally. That's how I designed Chalk. AVA is special since we spawn child processes, but the expected behavior should be preserved, but passing down the flag.

novemberborn:
Ah, hadn't realized `chalk` takes a `--no-color` argument.

We've previously argued against forwarding command line arguments to the workers though.

sindresorhus:
> We've previously argued against forwarding command line arguments to the workers though.

Yes, and we're not gonna start doing it with arbitrary flags, but I would say it's worth making an exception for this one as it matches expected outcome.

lib/cli.js
ThomasBem: @sindresorhus I tried to do what you suggested with default in in meow and that works for my code, 
`
 {
	color: true
}
`
but I don`t understand whats going on with what`s currently in default and how to make it work together with the change you recommended. Tried a few different things, but I keep getting a failing test in cli.js.

https://github.com/avajs/ava/blob/master/lib/cli.js#L75

ThomasBem:
I have updated the PR with the suggested changes regarding the CLI flag and the for...of loops. 

I need some help with keeping the existing "default: conf" in cli.js while also adding the new stuff @sindresorhus mentioned. I could not get it to work together with the new { color: true} so I just removed it for now and its causing a test to fail.

I don`t understand what this conf thing is and what its doing in default in the first place. It makes no sense to me :P

I can understand adding { color: true } saying that color should be true if nothing else is set. But just "conf" makes no sense to me.

I also need some support in where to start with passing the flag down to the child processes. That`s one of the things I have yet to understand with how Ava works :)

novemberborn:
> I don`t understand what this conf thing is and what its doing in default in the first place. It makes no sense to me :P

This may be relevant: https://github.com/avajs/ava/issues/1046#issuecomment-275071873

> I could not get it to work together with the new { color: true} so I just removed it for now and its causing a test to fail.

The test verifies that config in `package.json` overrides the default behavior. Currently the config is passed through the CLI parser (`meow`) using `default`, except that you removed that behavior :wink:

If you do `Object.assign({ color: true }, conf)` instead you should get the behavior you want, and a passing test. I think this is another argument for my suggestion in https://github.com/avajs/ava/issues/1046#issuecomment-275071873.

> I also need some support in where to start with passing the flag down to the child processes. That`s one of the things I have yet to understand with how Ava works :)

See here: https://github.com/avajs/ava/blob/872d2edb938ab1104b304f25aa586833b9875030/lib/fork.js#L39

`opts` should already contain the `colors` value, you'll just have to translate it to the correct flag.

ThomasBem:
Thanks for the help @novemberborn :)
 
I did what you said for the default, and that worked fine. Think I _sort_ of understand whats going on with that now.

I also made something that works for passing the color flag down to the forked process.

I am unsure of how / if I can test that small piece of code though. Any suggestions?

lib/cli.js
```diff
@@ -38,6 +38,7 @@ exports.run = () => {
 		  --verbose, -v           Enable verbose output
 		  --no-cache              Disable the transpiler cache
 		  --no-power-assert       Disable Power Assert
+		  --no-color              Disable color output
```
forresst: This flag should be added to readme.md (like -no-power-assert for example) in [CLI](https://github.com/avajs/ava/blob/master/readme.md#cli) paragraph

ThomasBem: Good catch! Did not even think about that.

ThomasBem: Should be good now :)

readme.md
```diff
@@ -159,6 +159,7 @@ $ ava --help
     --verbose, -v           Enable verbose output
     --no-cache              Disable the transpiler cache
     --no-power-assert       Disable Power Assert
+	--no-color              Disable color output
```
forresst: Please, fix indentation

ThomasBem: Hmm, ill take a look. It does not look that way for me in IDE or when looking at the readme in my branch. I see even Travis failed on some tests in node7 suddenly, when all i did was updated readme...

ThomasBem:
Looks better now? I used IntelliJ instead of VSCode to edit the markdown file.

ThomasBem:
Something going on with Travis? First time i edited the readme.md file, travis failed on node 7. 

Now I fixed a small indentation mistake and travis fails on node 4? Even different tests failing.

ThomasBem:
Seems like things are working much better today. So probably some hiccup with Travis yesterday.

test/reporters/mini.js
```diff
@@ -671,3 +671,26 @@ test('results when hasExclusive is enabled, but there are multiple remaining tes
 	t.end();
 });
 
+test('Result when no-color flag is set', function (t) {
```
sindresorhus: Result => result

test/reporters/verbose.js
```diff
@@ -571,6 +571,31 @@ test('results when hasExclusive is enabled, but there are multiple remaining tes
 	t.end();
 });
 
+test('Result when no-color flag is set', function (t) {
```
sindresorhus: Result => result

lib/reporters/mini.js
```diff
@@ -11,9 +11,9 @@ var repeating = require('repeating');
 var colors = require('../colors');
 
 chalk.enabled = true;
-Object.keys(colors).forEach(function (key) {
+for (let key of Object.keys(colors)) {
 	colors[key].enabled = true;
-});
+}
```
sindresorhus: Why are we setting this both here and in the constructor? Wouldn't it make more sense to only set it in the constructor?

Something like this:

```js
chalk.enabled = this.options.color;
for (const key of Object.keys(colors)) {
	colors[key].enabled = this.options.color;
}
```

lib/reporters/mini.js
```diff
@@ -28,6 +28,13 @@ function MiniReporter(options) {
 
 	this.options = Object.assign({}, options);
 
+	if (this.options.color === false) {
+		chalk.enabled = false;
+		for (let key of Object.keys(colors)) {
```
sindresorhus: `let` => `const`

lib/reporters/mini.js
sindresorhus: @Qix- This is needed because we assign shortcuts to the colors in [`lib/colors.js`](https://github.com/avajs/ava/blob/9616dde582e5333c3528cdd217c37e5b41798c9a/lib/colors.js) (context https://github.com/avajs/ava/pull/1104#issue-187486491) Any way we could make the above loop not nessecary? Maybe if we made `.enabled` property on the prototype?

ThomasBem: You mean adding a MiniReporter.prototype.enabled = function () {} that would enable all the colors?

sindresorhus: @ThomasBem Ignore this ;) Unrelated to this PR, but rather to [`chalk`](https://github.com/chalk/chalk).

ThomasBem: Ah, oki :) I will commit with updates based on your review, but without doing anything regarding the loop.

I did a quick try where I changed how we create mini and verbose reporters in their test setup. Making sure to include {colors: true} as an options. That way we can use your suggestion:
```
chalk.enabled = this.options.color;
for (const key of Object.keys(colors)) {
	colors[key].enabled = this.options.color;
}
```

It works for the mini and verbose tests themselves, but it caused this test to fail because chalk and colors did not seem to work as intended, https://github.com/avajs/ava/blob/master/test/cli.js#L73

sindresorhus: > but it caused this test to fail because chalk and colors did not seem to work as intended, https://github.com/avajs/ava/blob/master/test/cli.js#L73

I think that's as expected as we used to force Chalk in child processes and we no longer do. I guess we can just remove the `colors.error()` call from that test.

sindresorhus:
@ThomasBem When you push new changes, can you push them as additional commits instead of squashing? Makes it easier to see what changed. We squash on merge anyways.

ThomasBem:
Sure, I can do that :)

lib/fork.js
```diff
@@ -36,7 +36,9 @@ module.exports = (file, opts, execArgv) => {
 		} : false
 	}, opts);
 
-	const ps = childProcess.fork(path.join(__dirname, 'test-worker.js'), [JSON.stringify(opts)], {
+	var args = [JSON.stringify(opts), opts.color ? '--color' : '--no-color'];
```
novemberborn: Please use `const`.

(@sindresorhus is this something `xo` should enforce?)

lib/babel-config.js
```diff
@@ -14,9 +12,9 @@ function validate(conf) {
 	const isValidShortcut = conf === 'default' || conf === 'inherit';
 
 	if (!conf || (typeof conf === 'string' && !isValidShortcut)) {
-		let message = colors.error(figures.cross);
+		let message = figures.cross;
 		message += ' Unexpected Babel configuration for AVA. ';
-		message += 'See ' + chalk.underline('https://github.com/avajs/ava#es2015-support') + ' for allowed values.';
+		message += 'See https://github.com/avajs/ava#es2015-support for allowed values.';
```
novemberborn: Why this change? (I must admit I didn't read all of the back and forth earlier though… forgive me if I overlooked the answer)

lib/reporters/verbose.js
```diff
@@ -6,13 +6,16 @@ var plur = require('plur');
...
-});
-
-function VerboseReporter() {
+function VerboseReporter(options) {
 	if (!(this instanceof VerboseReporter)) {
-		return new VerboseReporter();
+		return new VerboseReporter(options);
+	}
+
+	this.options = Object.assign({}, options);
```
novemberborn: Assigning `this.options` seems unnecessary since the constructor didn't receive `options` before. Nothing else is using `this.options`.

lib/babel-config.js
sindresorhus: See: https://github.com/avajs/ava/pull/1198#discussion_r98352347

lib/fork.js
sindresorhus: I'll add the `esnext: true` option when https://github.com/avajs/ava/pull/1154 lands (hopefully in the next few days), as it's the last piece of non-ES2015ified code.

ThomasBem: Sorry, C# habbit :p

lib/reporters/mini.js
Qix-: I'm a little lost, could you explain a bit more?

Qix-: Nvm, Sindre explained it to me. See chalk/chalk#142.

lib/babel-config.js
novemberborn: @sindresorhus you suggested to remove the color comparison from the test, but this removes it from the actual AVA code.

Why do the changes in this PR require colors to be removed elsewhere? Should we then also remove colors from https://github.com/avajs/ava/blob/master/lib/cli.js#L100:L108?

ThomasBem: I removed it from the actual output, since we removed it from test. But if the problem only exits when running tests, not real Ava. Then we can put it back in the implementation.

ThomasBem: I did some testing and it seems like this is only a problem in test? Is this because execCli spawns a new process and the chalk.enabled and colors.enabled stuff is not passed on?

As far as I can tell none of the changes I did changed anything related to that behaviour? So why this did work before?

Since it seems to actually work when running Ava properly, we can always just remove the color from expected in the test? At least we catch that the error is thrown with the correct text. Even though colors and underline will be missing.

Qix-:
The github mobile site doesn't allow inline comments, so apologies for this being out of line.

Foe the child workera, why don't we just utilize support color's new `FORCE_COLOR` environment variable instead of pushing down arbitrary flags?

lib/babel-config.js
sindresorhus: Oh, I didn't notice this removes it from the actual code. In https://github.com/avajs/ava/pull/1198#discussion_r98352347, I was talking about removing the chalk call from the test, not elsewhere.

> Is this because execCli spawns a new process and the chalk.enabled and colors.enabled stuff is not passed on?

Not exactly. Colors used to be forced "on", now it's auto-detected and colors are automatically off when spawned.

sindresorhus:
@Qix- `FORCE_COLOR` is for users that can't pass flags. We control the child process, so we can pass flags. I don't really see the problem with that or any benefit in using `FORCE_COLOR`?

Qix-:
The parent AVA process is passed `--colors`, etc. Then it sets the environment variable `FORCE_COLOR` for child processes to _force_ them to match whatever color level the main (parent) process is running at. No need to do any messy flag juggling or whatever.

Removes the need to determine which flags to pass down. _This is what the environment was made for._

novemberborn:
@Qix- I think your explanation makes sense, but I'll defer to @sindresorhus who I'm sure has got a better understanding of this area than me.

sindresorhus:
> The parent AVA process is passed --colors, etc. Then it sets the environment variable FORCE_COLOR for child processes to force them to match whatever color level the main (parent) process is running at.

I still don't see how that's any easier. It's the same amount of work to pass down `--color`/`--no-color` based on those flags being set for AVA, as it's passing `FORCE_COLOR=0`/`FORCE_COLOR=1` based on AVA receiving `--color`/`--no-color`. The benefit of passing the flags is that we still respect `FORCE_COLOR` being passed to AVA by the user.

> No need to do any messy flag juggling or whatever. No need to determine which flags to pass down. No need to update flags in AVA in the future if supports-color changes how it handles flags. This is the perfect use case.

I don't do what-ifs. We work with how it works now. In the future the environment variable might be renamed, who knows.

novemberborn:
> The benefit of passing the flags is that we still respect FORCE_COLOR being passed to AVA by the user.

👍

ThomasBem:
So I have reverted back to babel-config.js change I made, and instead passed down an additional --color flag in execCli where needed.

Tok me quite a while to get my work in after @vadimdemedes big Magic Assert PR. But I think I managed to get everything up-to-date. But please take a look :)

sindresorhus:
Great work @ThomasBem! Another high-quality pull request :)

PR #1203, opened 2017-01-23
Exit with error if --watch is used in CI

### LINKED ISSUE #1186: Exit with error if --watch is used in CI
(Rather than ignoring `--watch` we've decided to exit with an error instead: https://github.com/avajs/ava/issues/1186#issuecomment-274246060 — @novemberborn)

---

## Description

This is a feature request. Skip '--watch' arg if `isCi`

Since we have known our test running on CI server. Ava should more clever to ignore `--watch` arg.

### Config

```json
{
  "scripts": {
    "test": "ava -w"
  }
}
```

Now I write my test command like this:

```json
{
  "scripts": {
    "test": "[ -z $CI ] && (ava -w); [ $CI ] && (ava)"
  }
}
```

forresst:
Fixes #1186

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -99,6 +99,10 @@ exports.run = () => {
 		throw new Error(colors.error(figures.cross) + ' The TAP reporter is not available when using watch mode.');
 	}
 
+	if ((hasFlag('--watch') || hasFlag('-w')) && isCi) {
+		throw new Error(colors.error(figures.cross) + ' AVA will not run with the --watch (-w) option in CI, because CI processes should terminate, and with the --watch option, AVA will never terminate.');
+	}
+
 	if (hasFlag('--require') || hasFlag('-r')) {
 		throw new Error(colors.error(figures.cross) + ' The --require and -r flags are deprecated. Requirements should be configured in package.json - see documentation.');
 	}
--- a/docs/recipes/watch-mode.md
+++ b/docs/recipes/watch-mode.md
@@ -83,6 +83,10 @@ Dependency tracking works for required modules. Custom extensions and transpiler
 
 The [`.only` modifier] disables watch mode's dependency tracking algorithm. When a change is made, all `.only` tests will be rerun, regardless of whether the test depends on the changed file.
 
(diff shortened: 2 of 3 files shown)


vadimdemedes:
✅ Readme updated
✅ Lint-free code
✅ Test added

Love PRs like this one! Thanks @forresst!

lib/cli.js
```diff
@@ -99,6 +99,10 @@ exports.run = () => {
 		throw new Error(colors.error(figures.cross) + ' The TAP reporter is not available when using watch mode.');
 	}
 
+	if ((hasFlag('--watch') || hasFlag('-w')) && isCi) {
+		throw new Error(colors.error(figures.cross) + ' AVA will not run with the --watch (-w) option in CI, because CI processes should terminate, and with the --watch option, AVA will never terminate.');
```
novemberborn: The explanation seems a bit long for a CLI error. Contrast with the TAP reporter where we just say "sorry, not available".

sindresorhus: I agree it could be more succinct, but I like how it includes the reason for why it's not available. Users don't really like being told something and not knowing "why".

sindresorhus: Feel free to improve it :)

forresst: Proposal for a shorter version:

`' AVA not run if using watch mode in CI, as CI processes should terminate and AVA will never terminate.'`

before:
* 147 characters

after:
* 102 characters

For information, CLI error with Tap reporter : 57 characters

novemberborn: How about `Watch mode is not available in CI, as it prevents AVA from terminating.`

forresst: @novemberborn  why not ?

sindresorhus: > Watch mode is not available in CI, as it prevents AVA from terminating.

👍

forresst: i will make this change

vadimdemedes:
Ah, excuse me for quick merge, I read the refd issue and thought you all agreed on the message.

PR #1212, opened 2017-01-26
Add recipe for precompiling source files with webpack

### LINKED ISSUE #1011: Recipe for using AVA with Webpack aliases
I use Webpack in my app and I defined aliases in config:
webpack.config.dev.js:

``` javascript
....
resolve: {
    // We can now require('file') instead of require('file.jsx')
    extensions: ['', '.js', '.jsx', '.scss'],
    alias: {
      actions: PATHS.app + '/actions',
      components: PATHS.app + '/components',
      store: PATHS.app + '/store',
      reducers: PATHS.app + '/reducers',
      helpers: PATHS.app + '/helpers',
      sagas:  PATHS.app + '/sagas',
      services:  PATHS.app + '/services',
    },
  }
....
```

And I import modules like so:
componts/LoginPage.js

```
import {login} from 'actions/Login'
```

instead of

```
import {login} from '../actions/Login'
```

But AVA doesn't allow to define aliases, so I'm unable to test **componts/LoginPage.js** with AVA since it throws that error:

```
Error: Cannot find module 'actions/Login'
```

Is there a solution?

danny-andrews:
Fixes #1011.

### DIFF
--- a/docs/recipes/precompiling-with-webpack.md
+++ b/docs/recipes/precompiling-with-webpack.md
@@ -1,0 +1,25 @@
+In the [README](https://github.com/avajs/ava#transpiling-imported-modules) it mentions precompiling your imported modules as an alternative to JIT compilation, but it doesn't explain how. This recipe shows how to do this using webpack.
+
+webpack.config.js
+```js
+var nodeExternals = require('webpack-node-externals');
+
+module.exports = {
+  entry: ['./src/tests.js'],
+  target: 'node',
+  output: {
+    path: '_build',
+    filename: 'tests.js'
+  },
+  externals: nodeModules,
+  module: {
+    rules: [
+      { test: /\.(js|jsx)$/, use: 'babel-loader' }
+    ]
+  }
+};
+```
+
+The important bits are `target: 'node'`, which ignores node-specific `require`s (e.g. `fs`, `path`, etc.) and `externals: nodeModules` which prevents webpack from trying to bundle external node librarires which it may choke on.
+
+Now you can simply run `ava _buid/tests.js` to run the tests contained in this output.
--- a/docs/recipes/precompiling-with-webpack.md
+++ b/docs/recipes/precompiling-with-webpack.md
@@ -26,4 +26,4 @@ module.exports = {
 
 The important bits are `target: 'node'`, which ignores Node.js-specific `require`s (e.g. `fs`, `path`, etc.) and `externals: nodeModules` which prevents webpack from trying to bundle external Node.js libraries which it may choke on.
 
-You can now run `$ ava _buid/tests.js` to run the tests contained in this output.
+You can now run `$ ava _build/tests.js` to run the tests contained in this output.
--- a/docs/recipes/precompiling-with-webpack.md
+++ b/docs/recipes/precompiling-with-webpack.md
@@ -1,25 +1,29 @@
-In the [README](https://github.com/avajs/ava#transpiling-imported-modules) it mentions precompiling your imported modules as an alternative to JIT compilation, but it doesn't explain how. This recipe shows how to do this using webpack.
+## Precompiling source files with webpack
 
-webpack.config.js
+The AVA [readme](https://github.com/avajs/ava#transpiling-imported-modules) mentions precompiling your imported modules as an alternative to runtime compilation, but it doesn't explain how. This recipe shows how to do this using webpack.
+
+###### webpack.config.js
+
 ```js
-var nodeExternals = require('webpack-node-externals');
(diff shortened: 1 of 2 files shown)


docs/recipes/precompiling-with-webpack.md
```diff
@@ -0,0 +1,25 @@
...
+var nodeExternals = require('webpack-node-externals');
+
+module.exports = {
+  entry: ['./src/tests.js'],
+  target: 'node',
+  output: {
+    path: '_build',
+    filename: 'tests.js'
+  },
+  externals: nodeModules,
```
forresst: I have not tested the recipe. But we should instead declare `nodeExternals` instead of `nodeModules`? I have a little difficulty understanding the logic. If what you wrote is correct then it might be good to explain why we should do this.

What do you think?

danny-andrews: No, that was just a typo. It's fixed now.

sindresorhus:
Is this Webpack v1 or v2? I think it should be Webpack v2 and be specified as such.

sindresorhus:
Note that I added a commit to this PR: https://github.com/avajs/ava/pull/1212/commits/4fe61974af3c4b074c084053ac199034f6c46197

docs/recipes/precompiling-with-webpack.md
```diff
@@ -0,0 +1,29 @@
+## Precompiling source files with webpack
+
+The AVA [readme](https://github.com/avajs/ava#transpiling-imported-modules) mentions precompiling your imported modules as an alternative to runtime compilation, but it doesn't explain how. This recipe shows how to do this using webpack.
```
sindresorhus: Add a link to this recipe in the AVA readme at https://github.com/avajs/ava#transpiling-imported-modules

sindresorhus:
You also need to add this to the recipes list: https://github.com/avajs/ava#recipes

docs/recipes/precompiling-with-webpack.md
```diff
@@ -0,0 +1,29 @@
...
+				test: /\.(js|jsx)$/,
+				use: 'babel-loader'
+		}]
+	}
+};
+```
+
+The important bits are `target: 'node'`, which ignores Node.js-specific `require`s (e.g. `fs`, `path`, etc.) and `externals: nodeModules` which prevents webpack from trying to bundle external Node.js libraries which it may choke on.
+
+You can now run `$ ava _buid/tests.js` to run the tests contained in this output.
```
novemberborn: `_build`

sindresorhus:
ping @danny-andrews

danny-andrews:
Added the fixes you requested. @sindresorhus

sindresorhus:
> Is this Webpack v1 or v2? I think it should be Webpack v2 and be specified as such.

@danny-andrews ⬆️

sindresorhus:
@TheLarkInn Could you take a look?

docs/recipes/precompiling-with-webpack.md
```diff
@@ -0,0 +1,29 @@
...
+
+```js
+const nodeExternals = require('webpack-node-externals');
+
+module.exports = {
+	entry: ['src/tests.js'],
+	target: 'node',
+	output: {
+		path: '_build',
+		filename: 'tests.js'
```
TheLarkInn: As someone who didn't know about this feature, I'm asking questions from a first time user standpoint that I think are important to address: 

*  I would love to see an example of what this file looks like. Could a code block be added above to show this? Is it just a list of `t.test`? Or better yet is there a repo for this?

TheLarkInn:
Thanks for the ping @sindresorhus 🐱 🐕 🐓 🚀. Was surprised I didn't know this feature existed!! MFW:
68a971d3c92e870fe4beed5a3f8d71a9

docs/recipes/precompiling-with-webpack.md
danny-andrews: It's a file with the contents of all the transpiled test code and all of its dependencies. It's really ugly and wouldn't really help to see.

chaucerbao: The `entry: ['src/tests.js']` can be a list of `t.test` or, better yet, `require` or `import` statements pulling in the tests you want to run, because you probably don't want all your tests in one file, unless your project is tiny. As the name suggests, it's the entry point that Webpack will look at to determine what to compile.

The `output.filename` will contain the compiled code.

danny-andrews:
@sindresorhus Anything else you'd like me to add?

sindresorhus:
@danny-andrews Good stuff. Thanks :)

PR #1213, opened 2017-01-30
Allow tests files to have any extension (i.e. `.jsx`, `.ts`) #631

### LINKED ISSUE #631: Allow tests files to have any extension (i.e. `.jsx`, `.ts`)

mlewando:
I added availability to test other files extensions than js (jsx for egzample). It only allows to treat other file extensions as test files. It doesn't touch trnaspilation (babel) proces.

Fixes #631


mlewando:
and regardless this specific PR: I have a question about tests, they looks broken. I tried to run them locally, but they failed. When I compared my local test results with my changes, an on master branch I had the same errors. Then I tought: maybe on your CI it will work and I have some local environment issues, but it looks like it's not problem only on my environment

novemberborn:
Hi @mlewando, thank you for going through all this effort. We're changing how AVA handles test and source compilation. Supporting other extensions is inextricably linked to how compilation works (e.g. Babel compilation shouldn't apply to TypeScript files). This is documented in [RFC 001].

As it stands we won't be adding a global `extensions`. #631 is blocked by [RFC 001], though since I forgot to append the `blocked` label this wasn't as clear as it could be. So unfortunately I have to reject your PR at this stage.

Please do remain subscribed to #631 since I'll communicate there when it's ready to work on again.

Also in the future it's a good idea to check in with an issue before doing a significant amount of work on it. That way we can make sure somebody else doesn't work on it in parallel, and it can prevent situations like this.

[RFC 001]: https://github.com/avajs/ava/blob/master/docs/specs/001%20-%20Improving%20language%20support.md

mlewando:
Hi @novemberborn, thank you for explanation. I read (looks like not carefully enough) the documentation in [RFC 001](https://github.com/avajs/ava/blob/master/docs/specs/001%20-%20Improving%20language%20support.md) and understand it and #631 as the `extensions` parameter will be one of the first things to do. 

Also I will very appreciate to have some workaround for this extensions problem soon. Thats why I decided to just write one in the meantime as it looks like very simple task. Unfortunately I run into some problems with tests (as you can see in CI statuses) and still have some problems with them. For the sake of training I'm going to make it work on my fork anyway, so there is no lost effort from my side ;)

Of course when #631 will be ready I'll be glad to help with it as much as I can :)

By the way, yours tests runs very slow... after running `npm run test-win` it takes ages to finish and some error logs are so big that it is impossible to know whats wrong. Do you know any simple fix to run test faster and have better error reporting?

novemberborn:
> For the sake of training I'm going to make it work on my fork anyway, so there is no lost effort from my side ;)

Glad to hear!

> By the way, yours tests runs very slow... after running npm run test-win it takes ages to finish and some error logs are so big that it is impossible to know whats wrong. Do you know any simple fix to run test faster and have better error reporting?

We use [`tap`](http://www.node-tap.org/) to test AVA, and it recently gained support for running tests in parallel. So there should be some speed-ups possible.

When developing I tend to run specific test files, e.g. `node_modules/.bin/tap test/ava-files.js`.

PR #1215, opened 2017-01-30
Removed --source, -S CLI flag. Fixes #937

### LINKED ISSUE #937: Remove --sources CLI flag

LasaleFamine:
Removed every reference within `lib/cli.js`.  Replaced the reference for the `Watcher` constructor with `conf.source`.

I don't know if I need to write some tests or there are other references to remove, I tried to check what seemed more relevant.

### DIFF
--- a/docs/recipes/watch-mode.md
+++ b/docs/recipes/watch-mode.md
@@ -67,11 +67,11 @@ In AVA there's a distinction between *source files* and *test files*. As you can
 
 By default AVA watches for changes to the test files, `package.json`, and any other `.js` files. It'll ignore files in [certain directories](https://github.com/novemberborn/ignore-by-default/blob/master/index.js) as provided by the [`ignore-by-default`] package.
 
-You can configure patterns for the source files in the [`ava` section of your `package.json`] file, using the `source` key. This is the recommended way, though you could also use the [`--source` CLI flag].
+You can configure patterns for the source files in the [`ava` section of your `package.json`] file, using the `source` key.
 
 You can specify patterns to match files in the folders that would otherwise be ignored, e.g. use `node_modules/some-dependency/*.js` to specify all `.js` files in `node_modules/some-dependency` as a source, even though normally all files in `node_modules` are ignored. Note that you need to specify an exact directory; `{bower_components,node_modules}/**/*.js` won't work.
 
-If your tests write to disk they may trigger the watcher to rerun your tests. If this occurs you will need to use the `--source` flag.
+If your tests write to disk they may trigger the watcher to rerun your tests. Configure patterns for the source files to avoid this.
 
 ## Dependency tracking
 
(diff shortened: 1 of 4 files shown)


LasaleFamine:
Dunno why `node 4` on Travis timed-out... 😕

sindresorhus:
There are more places that needs updating (Use Ack or The Silver Searcher):

```
~/dev/ava master
❯ ag '\--source'
docs/recipes/watch-mode.md
70:You can configure patterns for the source files in the [`ava` section of your `package.json`] file, using the `source` key. This is the recommended way, though you could also use the [`--source` CLI flag].
74:If your tests write to disk they may trigger the watcher to rerun your tests. If this occurs you will need to use the `--source` flag.
116:[`--source` CLI flag]: https://github.com/avajs/ava#cli

readme.md
164:    --source, -S            Pattern to match source files so tests can be re-run (Can be repeated)

test/cli.js
241:	const child = execCli(['--verbose', '--watch', '--source=source.js', 'test-*.js'], {dirname: 'fixture/watcher/with-dependencies'}, err => {
```

LasaleFamine:
Thanks for the tip. I will be back later today.

docs/recipes/watch-mode.md
```diff
@@ -67,11 +67,11 @@ In AVA there's a distinction between *source files* and *test files*. As you can
 
 By default AVA watches for changes to the test files, `package.json`, and any other `.js` files. It'll ignore files in [certain directories](https://github.com/novemberborn/ignore-by-default/blob/master/index.js) as provided by the [`ignore-by-default`] package.
 
-You can configure patterns for the source files in the [`ava` section of your `package.json`] file, using the `source` key. This is the recommended way, though you could also use the [`--source` CLI flag].
+You can configure patterns for the source files in the [`ava` section of your `package.json`] file, using the `source` key.
 
 You can specify patterns to match files in the folders that would otherwise be ignored, e.g. use `node_modules/some-dependency/*.js` to specify all `.js` files in `node_modules/some-dependency` as a source, even though normally all files in `node_modules` are ignored. Note that you need to specify an exact directory; `{bower_components,node_modules}/**/*.js` won't work.
 
-If your tests write to disk they may trigger the watcher to rerun your tests. If this occurs you will need to use the `--source` flag.
+If your tests write to disk they may trigger the watcher to rerun your tests.
```
novemberborn: Could you add:

> Configure patterns for the source files to avoid this.

readme.md
```diff
@@ -83,13 +83,13 @@ Your `package.json` will then look like this:
...
-  "devDependencies": {
-    "ava": "^0.17.0"
-  }
+	"name": "awesome-package",
+	"scripts": {
+		"test": "ava"
+	},
+	"devDependencies": {
+		"ava": "^0.17.0"
+	}
```
novemberborn: Could you undo the whitespace changes in this file? Your editor may be doing it automatically but the only change that should be here is removing the `--source` flag.

LasaleFamine:
I don't know why all those changes to the README. I think now is fixed.
Let me know if I need to change something more.

sindresorhus:
Nice work @LasaleFamine. Thanks for contributing :)

PR #1219, opened 2017-02-02
Fix flow errors using t.throws and t.notThrows

### LINKED ISSUE #1148: t.throw flow type check is giving a false-positive (Callable signature not found in)
## Description

I am getting an "Callable signature not found in" when using `t.throws`.

The type error originates in the `node_modules/ava/index.js.flow` definition.

This could be simply me misunderstanding what the error is saying.

I'd expect Flow to pickup:

```js
throws(value: Promise<mixed>, error?: ErrorValidator, message?: string): Promise<mixed>;
```

annotation and ignore:

```js
throws(value: () => void, error?: ErrorValidator, message?: string): mixed;
```

I even explicitly cast the result to `Promise<void>`.

If I remove the `Promise<void>` cast, I am getting a different error, https://gist.github.com/gajus/5c42f0262874a2bae7842691eea28a60#file-ava-flow-sh-L2.

### Test Source

```js
test('timesout after 60 seconds', async (t): Promise<void> => {
  const {
    document
  } = createDocument();

  global.document = document;

  const result: Promise<void> = loadScript('http://dailymail.co.uk/');

  clock.tick(6e4 + 1);

  t.throws(result, 'Script loading time has exceeded the maximum configured time limit.');
});
```

```js
$ flow
node_modules/ava/index.js.flow:71
 71: 	throws(value: () => void, error?: ErrorValidator, message?: string): mixed;
     	              ^^^^^^^^^^ function type. Callable signature not found in
 94:   t.throws(result, 'Script loading time has exceeded the maximum configured time limit.');
                ^^^^^^ Promise. See: test/utilities/loadScript.js:94


Found 1 error
```

### Config

```json
{
  "ava": {
    "babel": {
      "plugins": [
        "transform-flow-strip-types",
        "transform-async-to-generator",
        "transform-es2015-modules-commonjs"
      ]
    },
    "require": [
      "babel-register"
    ]
  }
}
```

### Command-Line Arguments

```
ava
```

## Environment

```
Node.js v7.2.1
darwin 16.1.0
0.17.0
3.10.10
```

leebyron:
Instead of double-defining each of these, define them once while using overloading.

Also took some liberties to improve the definition. Specifically, ava doesn't care what values these throwing functions return, but doesn't produce a return value.

Fixes #1148

### DIFF
--- a/index.js.flow
+++ b/index.js.flow
@@ -31,7 +31,7 @@ type ErrorValidator =
 	| ((error: any) => boolean);
 
 /**
- * Asertion Types
+ * Assertion Types
  */
 
 type AssertContext = {
@@ -57,11 +57,15 @@ type AssertContext = {
 	notDeepEqual<U>(value: U, expected: U, message?: string): void;
 	// Assert that function throws an error or promise rejects.
 	// @param error Can be a constructor, regex, error message or validation function.
-	throws(value: PromiseLike<mixed>, error?: ErrorValidator, message?: string): Promise<mixed>;
-	throws(value: () => void, error?: ErrorValidator, message?: string): mixed;
+	throws: {
+		(value: PromiseLike<mixed>, error?: ErrorValidator, message?: string): Promise<Error>;
+		(value: () => mixed, error?: ErrorValidator, message?: string): Error;
+	};
 	// Assert that function doesn't throw an error or promise resolves.
-	notThrows<U>(value: PromiseLike<U>, message?: string): Promise<U>;
-	notThrows(value: () => void, message?: string): void;
+	notThrows: {
+		notThrows<U>(value: PromiseLike<U>, message?: string): Promise<U>;
+		notThrows(value: () => mixed, message?: string): void;
+	};
 	// Assert that contents matches regex.
 	regex(contents: string, regex: RegExp, message?: string): void;
 	// Assert that contents does not match regex.


index.js.flow
```diff
@@ -57,11 +57,15 @@ type AssertContext = {
 	notDeepEqual<U>(value: U, expected: U, message?: string): void;
 	// Assert that function throws an error or promise rejects.
 	// @param error Can be a constructor, regex, error message or validation function.
-	throws(value: PromiseLike<mixed>, error?: ErrorValidator, message?: string): Promise<mixed>;
-	throws(value: () => void, error?: ErrorValidator, message?: string): mixed;
+	throws: {
+		(value: PromiseLike<mixed>, error?: ErrorValidator, message?: string): Promise<void>;
+		(value: () => mixed, error?: ErrorValidator, message?: string): void;
```
novemberborn: > Specifically, ava doesn't care what values these throwing functions return, but doesn't produce a return value.

`t.throws()` returns the error (and throws an assertion error if no error was thrown). Perhaps we could do something like:

```typescript
throws<E>: {
  		(value: PromiseLike<mixed>, error?: ErrorValidator, message?: string): Promise<E>;
  		(value: () => mixed, error?: ErrorValidator, message?: string): E;
}
```

I don't really know how generics work in Flow though. And of course we wouldn't want to require people to specify the error type if they're not going to use it.

index.js.flow
```diff
@@ -57,11 +57,15 @@ type AssertContext = {
...
+	};
 	// Assert that function doesn't throw an error or promise resolves.
-	notThrows<U>(value: PromiseLike<U>, message?: string): Promise<U>;
-	notThrows(value: () => void, message?: string): void;
+	notThrows: {
+		notThrows<U>(value: PromiseLike<U>, message?: string): Promise<U>;
```
novemberborn: Interesting. This isn't actually supposed to return the value, but looking at the code I think it does…

novemberborn:
@gajus could you give this a try?

leebyron:
Updated to have .throws() return Error or Promise<Error>

gajus:
I haven't had time to property isolate this, but using the latest code I am getting a lot of ambiguity errors that didn't exist before, e.g.

```
test/showtime-data-sources-management-api/loaders/DataSourceByDataSourceIdLoader.js:19
 19: test.after.always(() => {
     ^ call of method `always`. Could not decide which case to select
159: type ContextualTestMethod = {
                                 ^ callable object type. See: node_modules/ava/index.js.flow:159
  Case 1 may work:
  160: 	(              implementation: ContextualTest): void;
       	^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ function type. See: node_modules/ava/index.js.flow:160
  But if it doesn't, case 3 looks promising too:
  162: 	(              implementation: Macros<ContextualTestContext>, ...args: Array<any>): void;
       	^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ function type. See: node_modules/ava/index.js.flow:162
  Please provide additional annotation(s) to determine whether case 1 works (or consider merging it with case 3):
   19: test.after.always(() => {
                              ^ return
```

For the following code:

```js
test.after.always(() => {
  mockKnex.unmock(knex);
});
```

gajus:
Adding `:void` return type annotation fixes it, though.

leebyron:
@gajus - that's a good sign, it means the original flow error you encountered was solved so now flow is finding other unrelated issues.

This particular ambiguity error you're encountering I sent a PR for a while ago: #1164 - since it was merged it should be in the latest ava release.

gajus:
> @gajus - that's a good sign, it means the original flow error you encountered was solved so now flow is finding other unrelated issues.

While I agree that it is an expected behaviour, it has broken all of my CI deployments. :-(

I suppose that when it comes to making breaking Flow annotation changes, it is still a grey zone.

gajus:
Okay. What I have said is not true. ava is still `0.x`. Therefore `^0.17` does not automatically pick up `0.18`. Therefore, it is fine to be a breaking change.

leebyron:
I'm not sure that this is a breaking change - you should only be seeing other issues if they were originally masked by this issue - that is this change can bring you from being broken in two ways to being broken in one way, but not from working to being broken.

Since the annotation errors you're seeing are highlighting lines in the flow files from 0.17.0, could you please make sure that you have 0.18.0 installed? Flow support was broken in 0.17.0

gajus:
> I'm not sure that this is a breaking change - you should only be seeing other issues if they were originally masked by this issue - that is this change can bring you from being broken in two ways to being broken in one way, but not from working to being broken.

I use continuous delivery. Every time someone pushes the code, CI performs `npm install & flow`. Suddenly the CI breaks because of a (suppose) path update to ava. Thats a breaking change. Don't you agree?

Imagine if we discover a bug in production and urgently need to push the update. Discovering that Flow suddenly started breaking would be unfortunate.

leebyron:
I agree that a patch update where flow goes from passing to failing for otherwise reasonable code would be suboptimal. I do not think this change would have such an effect.

novemberborn:
> Okay. What I have said is not true. ava is still 0.x. Therefore ^0.17 does not automatically pick up 0.18. Therefore, it can be a breaking change.

Yay us 😄  @gajus I agree that an incompatible change in typing is a breaking issue, though whether fixing a bug is a breaking change is a perennial SemVer debate isn't it.

index.js.flow
novemberborn: Opened https://github.com/avajs/ava/issues/1227

sindresorhus:
🎉

PR #1221, opened 2017-02-03
Show how many tests were skipped by --fail-fast

### LINKED ISSUE #1172: Show how many tests were skipped due to --fail-fast
With #1160 AVA now prints a reminder when `--fail-fast` is used and a test fails. It'd be nice if the reminder could include the number of tests that were skipped.

A `remainingCount` could be computed and included when printing the reminder, as described in https://github.com/avajs/ava/issues/1134#issuecomment-267961180.

As per https://github.com/avajs/ava/issues/1134#issuecomment-267963077 we may not have determined the total number of tests when a failure occurs, so the copy should be something along the lines of "at least 5 tests were skipped".

See also #1171.

sotojuan:
I'm back @avajs/core!

Closes #1172.

Let me know if you want to change the text.

### DIFF
--- a/lib/reporters/mini.js
+++ b/lib/reporters/mini.js
@@ -233,7 +233,8 @@ class MiniReporter {
 		}
 
 		if (runStatus.failFastEnabled === true && runStatus.remainingCount > 0 && runStatus.failCount > 0) {
-			status += '\n\n  ' + colors.information('`--fail-fast` is on. Any number of tests may have been skipped');
+			const remaining = 'At least ' + runStatus.remainingCount + ' ' + plur('test was', 'tests were', runStatus.remainingCount) + ' skipped.';
+			status += '\n\n  ' + colors.information('`--fail-fast` is on. ' + remaining);
 		}
 
 		if (runStatus.hasExclusive === true && runStatus.remainingCount > 0) {
--- a/test/reporters/mini.js
+++ b/test/reporters/mini.js
@@ -643,7 +643,24 @@ test('results when fail-fast is enabled', t => {
 	compareLineOutput(t, output, [
 		'',
 		'',
-		'  ' + colors.information('`--fail-fast` is on. Any number of tests may have been skipped')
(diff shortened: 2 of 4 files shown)


lib/reporters/mini.js
```diff
@@ -233,7 +233,8 @@ class MiniReporter {
 		}
 
 		if (runStatus.failFastEnabled === true && runStatus.remainingCount > 0 && runStatus.failCount > 0) {
-			status += '\n\n  ' + colors.information('`--fail-fast` is on. Any number of tests may have been skipped');
+			const remaining = runStatus.remainingCount + ' skipped ' + plur('test', runStatus.remainingCount) + '.';
```
novemberborn: We don't necessarily know how many tests would have executed when running in `--concurrency` mode. So to be accurate this should be `At least 1 test was skipped.`.

test/reporters/mini.js
```diff
@@ -643,7 +643,7 @@ test('results when fail-fast is enabled', t => {
 	compareLineOutput(t, output, [
 		'',
 		'',
-		'  ' + colors.information('`--fail-fast` is on. Any number of tests may have been skipped')
+		'  ' + colors.information('`--fail-fast` is on. 1 skipped test.')
```
novemberborn: Should have a test for multiple skipped tests.

sotojuan:
I've updated the text to `At least 2 tests skipped.` `plur` doesn't handle `was` vs `were` so I decided to omit  it unless it's fine for me to make a small function in both reporters to handle that. More tests added too.

novemberborn:
> plur doesn't handle was vs were so I decided to omit it unless it's fine for me to make a small function in both reporters to handle that.

I think the current text is fine…? Second-guessing my language skills now.

You can use `plur('test was', 'tests were', runStatus.remainingCount)` by the way, no need for an extra function.

sotojuan:
All right, sorry about that. Text fixed!

Issue #1218, opened 2017-02-02
Snapshot feature will modify existing snapshots without passing `--update-snapshots` flag


wyze:
## Description

The results with snapshots are inconsistent when ran multiple times with no code changes. I've provided a repository with an example below.

Upon cloning the repo and running `npm test` it will pass the tests but leave the snapshots modified (See output of 2nd pass below). If you run tests again, it will throw an error (See 3rd pass output below).

For what it's worth, I was using the `master` branch after the snapshot feature first landed and had no issues. It seems like the bug was introduced when the magic assert feature was introduced. I even tried deleting the snapshot folder and getting a fresh run, but it still produces the same results.

### Error Message & Stack Trace

```sh
# 2nd pass
> ava

  4 passed

~/Projects/ava-snapshot-issue master*

# 3rd pass
> ava

  3 passed
  1 failed

  renders custom width
  test.js:9

   8:
   9:   t.snapshot(tree)
   10: }

  Error: Unexpected token m in JSON at position 28

  Test.snapshot (test.js:9:5)

npm ERR! Test failed.  See above for more details.
```

### Config

Copy the relevant section from `package.json`:

```json
{
  "ava": {
    "babel": "inherit",
    "require": "babel-register"
  }
}
```

### Command-Line Arguments

Copy your npm build scripts or the `ava` command used:

```
npm test
```

## Relevant Links

See reproducable repo here: https://github.com/wyze/ava-snapshot-issue

## Environment

Node.js v7.1.0
darwin 16.1.0
npm 4.0.3
AVA 0.18

vadimdemedes:
Thank you for a detailed bug report, I just reproduced it and looking into it.

wyze:
I think the problem lies with `JSON.stringify`ing the `serializedTree` here: [lib/assert#L171](https://github.com/avajs/ava/blob/master/lib/assert.js#L171)

If I remove the `JSON.stringify` and make it just `const result = toMatchSnapshot.call(context, serializedTree);`, it works every time.

vadimdemedes:
Hm yes, this makes the issue gone indeed. I've been debugging this issue inside jest-snapshot, and the problem lies in saving snapshot file. For some reason, on the second run, it reserializes/reencodes the last item in the snapshot.

wyze:
I tried to see if for some reason the macro was causing the problem, and broke out the tests into long form, and it still produced the same result.

[...]

PR #1223, opened 2017-02-03
[WIP] Replace jest-snapshot

vadimdemedes:
After investigating https://github.com/avajs/ava/issues/1218, I realized jest-snapshot is too much for us and I believe we can bundle a simpler solution. We cut down on unnecessary dependencies and lay more groundwork for universal (React/Preact/Inferno/more) assertions by not making assumptions about the content of your snapshots.

Since our replacement is basically just born, please report any bugs or missing/unexpected behavior.

If this PR lands and if you've used `t.snapshot` already, make sure to update them using `--update-snapshots` flag.

---

Todo:

- [ ] https://github.com/avajs/ava/issues/1218#issuecomment-277389930
- [ ] Needs to handle removing stale snapshots, and do it correctly with test.only/test.skip. (As discussed on Hangouts) *(Added by @sindresorhus)*
- [ ] Update readme regarding the changes we do and new behavior. *(Added by @sindresorhus)*
- [ ] Change directory to `__snapshots__` if inside a `__tests__` directory, otherwise `snapshots` *(added by @novemberborn)*
- [ ] Fix when snapshots are saved https://github.com/avajs/ava/issues/1218#issuecomment-280668480 (*added by @novemberborn)*


lusentis:
Hi,
I was testing this PR on a project, and found an issue with objects containing `undefined` values (and possibly others). I've created a repo to reproduce this: https://github.com/lusentis/ava-test-1223

image

Let me know if I can further help,
thanks!

vadimdemedes:
Hey @lusentis, I don't think it can be considered a bug, since `d` is still contained in that object, we can't skip it because it's `undefined`. If you try the following, it will also error:

```js
t.deepEqual({a: 1, b: undefined}, {a: 1});
```

lusentis:
Sure, I didn't explain it very well...
The issue is that the snapshot file *does not* contain the value which is `undefined`, so t.snapshot() fails.

This code:
```js
        const obj = {
                a: 123,
                b: 'bar',
                c: null,
                d: undefined
        }
        t.snapshot(obj);
```

produces this snapshot when running with `-u`:
```js

{
  "example snapshot of an object with undefined values": {
    "a": 123,
    "b": "bar",
    "c": null
  }
}
```
which is missing the "d" key from the Object. Obiously, `t.snapshot()` fails when run again because it complains that `d: undefined` is not in the snapshot.

vadimdemedes:
Right, it is an unexpected behavior for sure. It is happening, because we use `JSON.stringify` to serialize snapshot data and it skips `undefined` values.

vadimdemedes:
@lukechilds Thanks for trying out this PR and reporting the issue!

vadimdemedes:
A quick fix would be to `JSON.parse(JSON.stringify(actual))`, when comparing to `expected` value stored in the snapshot. Not sure about the drawbacks of this though. Otherwise, we'd have to manually serialize incompatible values like `undefined` and `Function` using a [replacer](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#The_replacer_parameter) parameter.

lusentis:
Found this: https://github.com/kaelzhang/node-code-stringify it might help
Thanks for your work on this! Let me know if I can help testing out.

lib/snapshot.js
```diff
@@ -0,0 +1,58 @@
...
+			throw new TypeError('Test file path is required');
+		}
+
+		this.dirPath = path.join(path.dirname(testPath), '__snapshots__');
+		this.filePath = path.join(this.dirPath, path.basename(testPath) + '.snap');
+		this.tests = {};
+		this.options = options || {};
+
+		if (fs.existsSync(this.filePath)) {
+			this.tests = JSON.parse(fs.readFileSync(this.filePath));
```
novemberborn: I'd just read the file and ignore any errors:

```js
let contents
try {
  contents = fs.readFileSync(this.filePath);
} catch (err) {}

this.tests = contents ? JSON.parse(contents) : {}
```

lib/snapshot.js
```diff
@@ -0,0 +1,58 @@
...
+			pass: false,
+			actual,
+			expected
+		};
+	}
+}
+
+const x = module.exports = Snapshot;
+
+Snapshot.getSnapshot = () => {
```
novemberborn: Why export the `Snapshot` class at all?

lib/snapshot.js
```diff
@@ -0,0 +1,58 @@
...
+	if (!x.snapshot) {
+		x.snapshot = new Snapshot(globals.options.file, {update: globals.options.updateSnapshots});
```
novemberborn: Could you document that this is only loaded in a worker process, so we only need one snapshot instance for the file that worker is testing?

dinoboff:
@novemberborn Assuming it serialise the new value and deserialise it before comparing it to the deserialised old value, it can support toJSON, no?

dinoboff:
ps: when storing a new value, snapshot could warn about any value lost in the process: it could show a diff of the value with the deserialised stored value.

capaj:
@novemberborn For me as a user of ava, I'd expect the same level of robustness as I get from jest snapshots. 
I think work done on https://www.npmjs.com/package/pretty-format is quite awesome and I don't think ava comunity needs to reinvent this piece of snapshoting feature. Maybe you could used that instead of doing `JSON.parse(JSON.stringify(actual))` ? I know it's not taking ava too far from the current state, but do we really need to? Jest snapshots are easily the most awesome feature of jest.

 Also when you've listed the features not supported by raw json snapshost, you failedto  mention Type support: https://www.npmjs.com/package/pretty-format#type-support

novemberborn:
> Maybe you could used that instead of doing JSON.parse(JSON.stringify(actual)) ?

@capaj yea that's a good idea.

@vadimdemedes if we go this route we'd have to do string comparisons on the formatted values. Can we do string diffs in case the new value doesn't match the saved snapshot?

lib/snapshot.js
```diff
@@ -0,0 +1,58 @@
...
+const mkdirp = require('mkdirp');
+const globals = require('./globals');
+
+class Snapshot {
+	constructor(testPath, options) {
+		if (!testPath) {
+			throw new TypeError('Test file path is required');
+		}
+
+		this.dirPath = path.join(path.dirname(testPath), '__snapshots__');
```
novemberborn: Jest convention aside, is there a reason we should stick to this format? Our [naming scheme is `helpers` and `fixtures`](https://github.com/avajs/ava/blob/314ef001ab1c085e2057738e2d2588bde3e792cc/lib/ava-files.js#L76:L80), not `__fixtures__`.

vadimdemedes: I agree, I'd prefer `snapshots` too.

lib/snapshot.js
vadimdemedes: For tests.

vadimdemedes:
We could use pretty-format to serialize values into snapshots, but then we wouldn't get a highlighted diff (when t.snapshot fails). I think it'd look weird, if only t.snapshot didn't support diff highlighting, while other assertions did.

lib/snapshot.js
novemberborn: Fair enough. Perhaps export both `Snapshot` and `getSnapshot` separately (rather than exporting the class as the main?)

Alternatively you could use `proxyquire` and inject the `globals` dependency, though you'd have to reload for each test.

vadimdemedes: > Perhaps export both Snapshot and getSnapshot separately (rather than exporting the class as the main?)

This doesn't seem to change anything. What is the benefit of this?

novemberborn: I think it better separates the "public" API (`getSnapshot()`) from the truly internal API (`Snapshot` constructor) which is needed for tests only.

vadimdemedes: Hm, but the thing is, both `getSnapshot()` and `Snapshot` aren't public. `t.snapshot` is an access point for these, users won't ever need to access the above mentioned stuff.

novemberborn: I know. Just voicing my preference 😄

sindresorhus:
I added a couple of todo items to the PR description.

lib/snapshot.js
novemberborn: Should this be configurable though? Folks who follow a `__tests__` scheme will want to use `__snapshots__`.

sindresorhus: I would really prefer not introducing more options. Can we make it `__snapshots__` if inside a `__tests__` directory, otherwise `snapshots`?

novemberborn: > I would really prefer not introducing more options. Can we make it __snapshots__ if inside a __tests__ directory, otherwise snapshots?

Yea that sounds great to me.

sindresorhus:
I just realized the benefit of doing like Jest is with a JS snapshot file is that they can have template literals with multiple line, so they can have good git diffing when updating snapshots.

novemberborn:
> Needs to handle removing stale snapshots, and do it correctly with test.only/test.skip. (As discussed on Hangouts) (Added by @sindresorhus)

@sindresorhus what are you intending here? I'm under the impression that if all tests are run, the snapshots will be updated correctly. If you use `t.only` then at least you'll see an overly large snapshot diff. Is the only issue one where a test file is removed but we don't clean up the snapshot file? And does Jest handle this?

novemberborn:
I've also added a few TODO items. Let's decide our direction first though: https://github.com/avajs/ava/issues/1275.

sindresorhus:
> Is the only issue one where a test file is removed but we don't clean up the snapshot file?

Not just removing a test file. Imagine the user removes a single test. There will now be a stale entry in the snapshot file. I would expect AVA to remove the stale test entry in the snapshot when I run `--update-snapshots`.

This is made harder by `t.skip()`/`.only()` though. Here's my discussion with Vadim about it. I'm too lazy to summarize.

screen shot 2017-02-20 at 14 41 12
screen shot 2017-02-20 at 14 41 30

> And does Jest handle this?

Yes

novemberborn:
> Not just removing a test file. Imagine the user removes a single test. There will now be a stale entry in the snapshot file. I would expect AVA to remove the stale test entry in the snapshot when I run --update-snapshots.

Yes, I'd imagine it would do that. That's easy to do though. Removing a snapshot file after a test file has been removed / renamed isn't too hard either since we glob all test files and can compare that to the existing snapshot files.

> This is made harder by t.skip()/.only() though

I see. As you said, we can simply not clean up snapshots for skipped tests.

Omniroot:
One more item for the TODO: 
[ ] Allow multiple snapshots per test

Currently, the snapshots are indexed by test name, it seems; however, when I tested in my project, I received false positives when there were multiple snapshots in a test. The current implementation in master (using `jest-snapshot`) appends a number after each one.

For example, if my test is called `"foo is consistent"`, and it has 2 snapshots, We could do something like: `"foo is consistent 1"` and `"foo is consistent 2"`.

What I'm currently doing is a bit nicer to read (diff-wise), which is the snapshot a hash with the different snapshots. That way they each have a name associated with them.

lib/snapshot.js
```diff
@@ -0,0 +1,58 @@
...
+		this.tests = {};
+		this.options = options || {};
+
+		if (fs.existsSync(this.filePath)) {
+			this.tests = JSON.parse(fs.readFileSync(this.filePath));
+		}
+	}
+
+	save() {
+		mkdirp.sync(this.dirPath);
```
Omniroot: We should add something like:

```javascript
let hasTests = false;
for(var key in obj) {
  if(obj.hasOwnProperty(key))
    hasTests = true;
    break;
  }
}
if (!hasTests) return;
```

at the beginning of this function. That way we don't end up making a bunch of empty files/divs.

We also need to ask ourselves what we want to do if someone deletes a test. Should we delete the snapshots too? Or leave them there?

Omniroot:
On my project, the diffs seem to be backwards. For example, if my test does: `t.snapshot({foo: 'foo'})`, and then I change the test to `t.snapshot({bar: 'bar'})`, I think the diff should be:

```diff
Object {
- foo: foo,
+ bar: bar,
}
```

but I see the opposite:

```diff
Object {
- bar: bar,
+ foo: foo,
}
```

Omniroot:
We could use something like this for non-native objects. In particular, it might be useful for `Error` objects.

```javascript
function objify(val) {
  if (! val instanceof Object) return val;
  var obj = {};
  Object.getOwnPropertyNames(val).forEach(function (key) {
    obj[key] = objify(val[key]);
  }, obj);
  return obj;
}
```

lib/snapshot.js
Omniroot: FYI, I forked off and tried this out and it works quite well: https://github.com/Omniroot/ava/pull/2/commits/c95962a4de900e7437965d5f7ecb04b6aaf98b9c

novemberborn:
Humbly, I think this is superseded by https://github.com/avajs/ava/pull/1341#issuecomment-293959914 😄

novemberborn:
Superseded by #1341.

PR #1230, opened 2017-02-05
Add flow type tests

### LINKED ISSUE #1224: Add flow type unit tests
It seems like there have been a few issues recently with flow types being broken, perhaps we should add some unit tests for them. Opening an issue to gauge interest, I'm happy to do the initial setup work if folks think this is worthwhile.

Here's what's involved:

We would add `flow-bin` as a devDependency and add a `.flowconfig` to the project, then add a file (or directory of files) that end in `-test.js.flow`. We would also add a small script that ran `flow check` on these files and pass if no errors were reported.

Flow has a feature where you can annotate portions of your code that you expect a type error, that instead cause a flow check to fail if they *do not* produce a type error. This provides a nice way to write both positive and negative tests.

leebyron:
Closes #1224

This adds flow-bin, adds it to the test running script, and adds some regression tests based on prior issues.

In an immediate proof of value (and proof of my own prior sloppiness) it spotted an issue that #1219 sought to fix and did not do so. Fix included along with test.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,0 +1,7 @@
+[ignore]
+<PROJECT_ROOT>/node_modules/.*
+<PROJECT_ROOT>/.*\.js$
+
+[options]
+emoji=true
+suppress_comment=\\(.\\|\n\\)*\\$ExpectError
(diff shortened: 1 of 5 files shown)


sotojuan:
Thanks Lee!

PR #1234, opened 2017-02-06
Simplify failure output - fixes #1072

### LINKED ISSUE #1072: Simplify failure output
With:

``` js
import test from 'ava';

test('returns true when happy', t => {
    throw new Error('unicorn');
});
```

It currently outputs:

```
❯ ava

  1 failed

  returns true when happy
  Error: unicorn
    Test.fn (test.js:5:8)
```

I think we should simplify it to:

```
❯ ava

  1 failed

  returns true when happy
  Error: unicorn
    test.js:5:8
```

So users can quickly see where the error came from. `Test.fn` doesn't have any value to the user.

This also applies to the verbose reporter (`$ ava --verbose`).
## 

We could also use https://github.com/sindresorhus/ansi-escapes#itermsetcwdpath to make the path clickable on iTerm.

LasaleFamine:
The idea is to check when the `test.error.source.file` is NOT the same of the `test.file` within the `forEach` of the errors.

I saw that when I throw an error from another module the `test.file` refers to the module's file and not the file that is running the test.

In this case for example if I make this check, I'm able to hide the stack trace.
```sh
  1 failed

  lol.js:2

   1: const testing = () => { 
   2:   throw new Error('lol')
   3: }                       

  Error: lol

  testing (lol.js:2:8)
  Test.t [as fn] (test.js:5:2)
```

Waiting for your feedback to move forward!

### DIFF
--- a/lib/reporters/mini.js
+++ b/lib/reporters/mini.js
@@ -203,7 +203,7 @@ class MiniReporter {
 					status += '\n' + indentString(test.error.message, 2) + '\n';
 				}
 
-				if (test.error.stack) {
+				if (test.error.stack && test.error.source.file !== test.file) {
 					status += '\n' + indentString(colors.errorStack(extractStack(test.error.stack)), 2);
 				}
 			});
--- a/lib/reporters/mini.js
+++ b/lib/reporters/mini.js
@@ -203,8 +203,11 @@ class MiniReporter {
 					status += '\n' + indentString(test.error.message, 2) + '\n';
 				}
 
-				if (test.error.stack && test.error.source.file !== test.file) {
-					status += '\n' + indentString(colors.errorStack(extractStack(test.error.stack)), 2);
+				if (test.error.stack) {
+					const extracted = extractStack(test.error.stack);
+					if (extracted.includes('\n')) {
+						status += '\n' + indentString(colors.errorStack(extracted), 2);
+					}
 				}
 			});
 		}
--- a/lib/reporters/verbose.js
+++ b/lib/reporters/verbose.js
@@ -122,7 +122,10 @@ class VerboseReporter {
 				}
 
 				if (test.error.stack) {
-					output += '\n' + indentString(colors.errorStack(extractStack(test.error.stack)), 2);
+					const extracted = extractStack(test.error.stack);
+					if (extracted.includes('\n')) {
+						output += '\n' + indentString(colors.errorStack(extracted), 2);
+					}
 				}
 			});
 		}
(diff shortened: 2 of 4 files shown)


novemberborn:
> I saw that when I throw an error from another module the test.file refers to the module's file and not the file that is running the test.

To clarify, you mean `test.error.source.file` refers to the module's file?

---

Your current approach means no stack traces are ever shown for errors that come out of the test file. I think that's an issue since users may write helper functions in the test file itself.

Instead we should not output the stack when it only contains one line. So this *should* show the stack trace:

```js
const testing = () => {
  throw new Error('lol')
}

test('testing', t => {
  testing()
})
```

But this should *not*:

```js
test('another test', t => {
  throw new Error('lolol')
})
```

That's my interpretation of @sindresorhus' comment https://github.com/avajs/ava/issues/1072#issuecomment-276891510.

LasaleFamine:
> To clarify, you mean test.error.source.file refers to the module's file?

I this case (that will produce mi first post output), `test.error.source.file === 'lol.js'` but `test.file === 'test.js'`:
```javascript 
// test.js
import test from 'ava'
import fn from './lol'

test('asd', t => {
    fn()
    t.fail()
})
```

```javascript 
// lol.js
module.exports.testing = () => {
    throw new Error('lol')
}
```

I understood your point, actually initially I was thinking to check if the ***extracted*** stack trace was only one line, but I didn't know if that single line could contain some other info.  
I can proceed in this way if you agree.

novemberborn:
> I understood your point, actually initially I was thinking to check if the extracted stack trace was only one line, but I didn't know if that single line could contain some other info.
> I can proceed in this way if you agree.

Yes I think that's the way to go.

LasaleFamine:
Can I assume that there aren't `\n` chars within the ***extracted*** `stack`?

Ex: 
```javascript
// lib/reporters/mini.js finish(runStatus)

// Check also if the stack isn't only the `test Fn` (actually check for one-line stack)
if (test.error.stack && extractStack(test.error.stack).search('\n') > -1) {
       status += '\n' + indentString(colors.errorStack(extractStack(test.error.stack)), 2);
}
```

Actually I don't like so much this solution because I'm calling twice the same function and I really need the comment to explain what I'm doing.  
Do you have a better/correct solution for the problem?

novemberborn:
> Actually I don't like so much this solution because I'm calling twice the same function and I really need the comment to explain what I'm doing.

You could assign the stack to a variable:

```js
if (test.error.stack) {
  const extracted = extractStack(test.error.stack);
  if (extracted.includes('\n')) {
    status += '\n' + indentString(colors.errorStack(extracted), 2);
  }
}
```

LasaleFamine:
I have made the change and also added the check to the `verbose` reporter. Also test fixed.  
Let me know if I need to change something.

novemberborn:
screen shot 2017-02-09 at 12 02 56

This is the output for the examples I posted in https://github.com/avajs/ava/pull/1234#issuecomment-277965444. Looks good to me! @sindresorhus?

sindresorhus:
Looks great! Thank you @LasaleFamine :)

PR #1240, opened 2017-02-08
Ensure TAP reporter test does not write to stderr

novemberborn:
Stub process.stderr.write(), rather than just spying on it. Otherwise 'resultresult' gets written to the stderr while the tests run.


sindresorhus:
Hah, I noticed that and thought it was a bug in the latest version of node-tap.

PR #1241, opened 2017-02-08
Speed up flow check

novemberborn:
`flow check` is easily taking 30 seconds on my development machine, presumably it's checking too many files. Move the `.flowconfig` into the flow-types test directory and only include the `index.js.flow` file to improve check speed.

I've verified it still catches type errors by reverting the `notThrows` fix from #1230.

@leebyron does my analysis make sense? Given how the `ignore` expressions work I don't think there's any other way of restricting `flow check` to just the three `.js.flow` files we currently have.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "node": ">=4"
   },
   "scripts": {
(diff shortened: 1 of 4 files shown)


sindresorhus:
Let's merge this now. The slowness is painful. We can improve it further if @leebyron has any additional suggestions.

Issue #1231, opened 2017-02-06
Ignore node_modules and .gitignore when reading code excerpts


unional:
ava: 0.18.1

In 0.17, I can still type "r" to rerun all tests in watch mode.

In 0.18.1 (and 0.18.0) this feature seems to be not working.

unional:
When I Ctrl-C, it reports error like:
```sh
Error: ENOENT: no such file or directory, open 'E:\hwong\panda\module-loader\node_modules\systemjs\src\instantiate.js'
    at Object.fs.openSync (fs.js:558:18)
    at Object.fs.readFileSync (fs.js:468:33)
    at module.exports (E:\hwong\panda\module-loader\node_modules\ava\lib\code-excerpt.js:15:20)
    at runStatus.errors.forEach (E:\hwong\panda\module-loader\node_modules\ava\lib\reporters\mini.js:191:22)
    at Array.forEach (native)
    at MiniReporter.finish (E:\hwong\panda\module-loader\node_modules\ava\lib\reporters\mini.js:178:21)
    at Logger.finish (E:\hwong\panda\module-loader\node_modules\ava\lib\logger.js:38:28)
    at busy.api.run.then.runStatus (E:\hwong\panda\module-loader\node_modules\ava\lib\watcher.js:112:12)
From previous event:
    at Watcher.run.specificFiles [as run] (E:\hwong\panda\module-loader\node_modules\ava\lib\watcher.js:110:68)
    at Watcher.rerunAll (E:\hwong\panda\module-loader\node_modules\ava\lib\watcher.js:270:8)
    at new Watcher (E:\hwong\panda\module-loader\node_modules\ava\lib\watcher.js:130:8)
    at Object.exports.run (E:\hwong\panda\module-loader\node_modules\ava\lib\cli.js:152:20)
    at Object.<anonymous> (E:\hwong\panda\module-loader\node_modules\ava\cli.js:22:24)
    at Module._compile (module.js:571:32)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
    at Module.runMain (module.js:605:10)
    at run (bootstrap_node.js:420:7)
    at startup (bootstrap_node.js:139:9)
    at bootstrap_node.js:535:3
```

Seems like `ava` tries to load file using the sourcemap path?
[...]

PR #1242, opened 2017-02-08
Ensure watcher rethrows logger errors

novemberborn:
The watcher would break if `logger.finish()` threw an exception. Instead rethrow this error (which causes the process to exit).

This surfaced in https://github.com/avajs/ava/issues/1231#issuecomment-278169455.

### DIFF
--- a/lib/watcher.js
+++ b/lib/watcher.js
@@ -107,13 +107,15 @@ class Watcher {
 				}
 			}
 
-			this.busy = api.run(specificFiles || files, {runOnlyExclusive}).then(runStatus => {
-				runStatus.previousFailCount = this.sumPreviousFailures(currentVector);
-				logger.finish(runStatus);
+			this.busy = api.run(specificFiles || files, {runOnlyExclusive})
+				.then(runStatus => {
+					runStatus.previousFailCount = this.sumPreviousFailures(currentVector);
+					logger.finish(runStatus);
 
-				const badCounts = runStatus.failCount + runStatus.rejectionCount + runStatus.exceptionCount;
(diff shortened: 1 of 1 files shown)


unional:
Tested and seems like it still hang the watch mode.

novemberborn:
Strange, it works for me.

* Install `npm i avajs/ava#rethrow-watch-logger-errors`
* Modify `node_modules/ava/lib/code-excerpt.js` to `const source = fs.readFileSync(file + 'foo', 'utf8');`
* Write test file `import test from 'ava'; test(t => { throw new Error("uhoh") })`
* Run AVA: `./node_modules/.bin/ava --watch test.js`

```
 ⠼ test › [anonymous]

  1 failed/private/var/folders/_6/p8qxp_3n62zg9081tvb0lcc80000gn/T/tmp.vx8nC52ks6/node_modules/ava/lib/watcher.js:15
		throw err;
		^

Error: ENOENT: no such file or directory, open '/private/var/folders/_6/p8qxp_3n62zg9081tvb0lcc80000gn/T/tmp.vx8nC52ks6/test.jsfoo'
```

unional:
Async test

novemberborn:
Even with `import test from 'ava'; test(t => Promise.reject(new Error("uhoh")))` the watcher exits for me due to the reporter error.

sindresorhus:
I can confirm it correctly crashes for me too when following the steps in https://github.com/avajs/ava/pull/1242#issuecomment-278918213 and https://github.com/avajs/ava/pull/1242#issuecomment-279230434

PR #1243, opened 2017-02-08
Add TypeScript definition for the snapshot assertion

despairblue:
The snapshot method was missing in the typescript type-definition file.

### DIFF
--- a/types/base.d.ts
+++ b/types/base.d.ts
@@ -72,6 +72,10 @@ export interface AssertContext {
 	 */
 	regex(contents: string, regex: RegExp, message?: string): void;
 	/**
+	 * Assert that contents matches a snapshot.
+	 */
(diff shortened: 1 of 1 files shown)


types/base.d.ts
```diff
@@ -72,6 +72,10 @@ export interface AssertContext {
 	 */
 	regex(contents: string, regex: RegExp, message?: string): void;
 	/**
+	 * Assert that contents matches a snapshot.
+	 */
+	snapshot(contents: any): void;
```
sindresorhus: It accepts an optional message too, like all the other asserts.

despairblue: Thanks, done.

novemberborn:
Thanks @despairblue!

PR #1246, opened 2017-02-08
Symlink fix

forivall:
Fixes an issue that I found when i had errors under symlinks in my `node_modules` folder; `codeExcerpt` would crash because `path.join` doesn't take absolute paths into consideration. `path.resolve` does.

### DIFF
--- a/lib/reporters/mini.js
+++ b/lib/reporters/mini.js
@@ -187,7 +187,7 @@ class MiniReporter {
 				if (test.error.source) {
 					status += '  ' + colors.errorSource(test.error.source.file + ':' + test.error.source.line) + '\n';
 
-					const errorPath = path.join(this.options.basePath, test.error.source.file);
+					const errorPath = path.resolve(this.options.basePath, test.error.source.file);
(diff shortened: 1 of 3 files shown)


novemberborn:
Perhaps we should ensure `error.source.file` is an absolute path. Then we can remove the `path.join()` from the reporters. That's in https://github.com/avajs/ava/blame/2dc47532543a7d5ef8b86b48396b7f53d20963a3/lib/serialize-error.js#L62.

@vadimdemedes, thoughts?

vadimdemedes:
Yep, agree.

But we'll still need relative path to fix #1231 (skip code excerpts and magic-assert for errors coming from `node_modules`).

novemberborn:
> But we'll still need relative path to fix #1231 (skip code excerpts and magic-assert for errors coming from node_modules).

We can always compute the path back to the project directory if necessary.

novemberborn:
> Perhaps we should ensure error.source.file is an absolute path. Then we can remove the path.join() from the reporters.

I've opened https://github.com/avajs/ava/issues/1270 to track this. I'm working on that now.

Thanks for raising this issue @forivall. I'm looking forward to future PRs from you 😄

Issue #1228, opened 2017-02-05
Crashes while looking for helper files in huge project subdirectories


transcranial:
Currently, ava looks for helper files in all project subdirectories excluding node_modules:

https://github.com/avajs/ava/blob/master/lib/ava-files.js#L90-L95

This can overwhelm ava when there exists subdirectories that are huge and/or with complicated fs hierarchies:

```
FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory
 1: node::Abort() [ava]
 2: 0x126389c [ava]
 3: v8::Utils::ReportOOMFailure(char const*, bool) [ava]
 4: v8::internal::V8::FatalProcessOutOfMemory(char const*, bool) [ava]
 5: v8::internal::Factory::NewInternalizedStringImpl(v8::internal::Handle<v8::internal::String>, int, unsigned int) [ava]
 6: v8::internal::StringTable::LookupString(v8::internal::Isolate*, v8::internal::Handle<v8::internal::String>) [ava]
 7: v8::internal::LookupIterator::PropertyOrElement(v8::internal::Isolate*, v8::internal::Handle<v8::internal::Object>, v8::internal::Handle<v8::internal::Object>, bool*, v8::internal::LookupIterator::Configuration) [ava]
 8: v8::internal::Runtime_KeyedGetProperty(int, v8::internal::Object**, v8::internal::Isolate*) [ava]
 9: 0x126964c063a7
Aborted (core dumped)
```

Allowing the specification of exclusion globs for helper files would help prevent this. We reverted back to v0.17.0 for now.

Environment:
node v7.4.0
linux 4.2.0-34-generic
ava v0.18.1

sindresorhus:
Approximately how many files/directories and how deep? Do you specify any arguments to `$ ava` or any AVA config in package.json?

Could you try removing https://github.com/avajs/ava/blob/b886c5b25ae54430f908c49f9932d8f998c41702/lib/ava-files.js#L141-L143 and see if it helps?

Could you try to narrow down to which [patterns](https://github.com/avajs/ava/blob/b886c5b25ae54430f908c49f9932d8f998c41702/lib/ava-files.js#L90-L95) are causing this? 

I find it strange that just globbing for files would cause this. We must be reading in too much data somewhere.

transcranial:
[...]

PR #1250, opened 2017-02-09
Excluding files/folders in .gitignore from ava-files. Fixes #1229 

### LINKED ISSUE #1229: Excluding files/folders in .gitignore from ava-files
See #1228 for an example as to why this would be helpful.

7373Lacym:
Per the discussion [here](https://github.com/avajs/ava/issues/1228).  I think this is what is needed.


sindresorhus:
@7373Lacym The issue is about ignoring files matching glob patterns defined in .gitignore, not ignoring the .gitignore file. Should have been clearer. I've added some more text to the issue now.

7373Lacym:
👍

PR #1251, opened 2017-02-09
Document helper precompilation caveat (in code at least)

novemberborn:
The way helpers are precompiled, if you invoke AVA from inside a directory, with file arguments, then only helper files inside that directory are precompiled.

We'll be able to fix it when we move compilation into the workers, and I'm not even convinced it should be considered a bug at this stage. I did want to document the behavior somewhere, but I'm not sure it has to be in user-facing documentation as it's (hopefully) a rather specific edge-case.

### DIFF
--- a/api.js
+++ b/api.js
@@ -119,6 +119,11 @@ class Api extends EventEmitter {
 	_precompileHelpers() {
 		this._precompiledHelpers = {};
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
I think inline in the code is a good place to document this. We also have https://github.com/avajs/ava/blob/master/maintaining.md for internal notes.

PR #1262, opened 2017-02-14
Resolve Babel options ahead of time 🎉

novemberborn:
[RFC 001](https://github.com/avajs/ava/blob/master/docs/specs/001%20-%20Improving%20language%20support.md) requires us to resolve Babel options in the main process. This PR lands that new behavior, fixing #707 in the process.

Note that with these changes AVA will no longer resolve `.babelrc` files relative to the test file. Instead it only looks at a `.babelrc` file in the project directory.

AVA's `stage-4` and `transform-test-files` presets no longer have the package hashes of their plugins included in the caching hash for Babel transforms. This behavior will shift to `package-hash` itself, see https://github.com/novemberborn/package-hash/issues/22.

The caching hash is now derived from the full Node.js version, rather than just the major. This allows for presets / plugins which vary behavior based on feature differences in minor releases. I'm assuming that people don't update their Node.js versions *all* of the time, so this shouldn't really have an impact. Note that cached files for other versions are preserved, so this doesn't impact users who switch between Node.js versions.

We're now better at detecting whether Babel options have changed, especially for users who `extend` other files. We're also hashing the presets and plugins and will reapply transforms when these versions change.


novemberborn:
We should disable the Babel cache, given that we're now really good at managing it ourselves. Need to set `process.env.BABEL_DISABLE_CACHE = '1'`, perhaps in `lib/cli.js`.

lib/babel-config.js
```diff
@@ -24,85 +27,122 @@ function validate(conf) {
...
-	} else {
-		options = {
-			babelrc: false
-		};
-
-		Object.assign(options, babelConfig);
+function build(projectDir, cacheDir, userOptions, powerAssert) {
+	// Compute a seed based on the Node.js version and the project directory.
+	// Dependency hashes may vary based on the Node.js version, e.g. with the
+	// @ava/stage-4 Babel preset. Sources and dependencies paths are  absolute in
```
sindresorhus: There are two spaces between `are` and  `absolute`

sindresorhus:
Generally looks good. Will give it a more thorough read later today. I've ran it on a bunch of projects of mine and the performance seems to be the same as master, which is good.

sindresorhus:
Sorry about the delay. Did a thorough review now and nothing more to point out. Excellent work as always Mark.

novemberborn:
No worries! I better write docs and release hullabaloo at 1.0 now though 😉

Issue #1263, opened 2017-02-14
test hangs forever when asserting a response from an axios request


RickyMarou:
## Description

test hangs forever when asserting a response from an axios request. 

### Test Source
originally a typescript file : 
```js
import test from 'ava'
import axios from 'axios'

test('axios test', function (t) {
  return axios.get('https://www.google.de/search?q=help')
    .then((response) => {
      console.log(response.data)

      // DOES NOT EXIT
      t.true(response.data === 'something')

      // DOES EXIT
      const data = response.data
      t.true(data === 'something')

      // DOES EXIT
      const assert = response.data === 'something'
      t.true(assert)
    })
})
```

here is the compiled output, nothing suprising really : 
```js
"use strict";
const ava_1 = require("ava");
const axios_1 = require("axios");
ava_1.default('axios test', function (t) {
    return axios_1.default.get('https://www.google.de/search?q=help')
        .then((response) => {
        console.log(response.data.length);
        // DOES NOT EXIT
        t.true(response.data === 'something');
        // DOES EXIT
        const data = response.data;
        t.true(data === 'something');
        // DOES EXIT
        const assert = response.data === 'something';
        t.true(assert);
    });
});
```

### Error Message & Stack Trace

There is no error message, the test just hangs forever. The test only hangs when the response is being asserted, if the response is assigned to a variable and this variable is being asserted, the test ends as expected.

### Config

Copy the relevant section from `package.json`:

```json
{
  "dependencies": {
    "axios": "^0.15.3",
    "bluebird": "^3.4.7",
  },
  "devDependencies": {
    "ava": "^0.18.1",
  }
}
```

### Command-Line Arguments

I have build scripts but the error could be reproduced by running the test (from the transpiled js) with the ava CLI 

```
ava -v tests/build/tests/axios.test.js
```

## Environment
node :
```
Node.js v6.9.4
darwin 15.6.0
```
ava : 
```
0.18.1
```

npm:
```
3.10.10
```

I hope this issue report will be useful :) keep up the good work

novemberborn:
With `t.true(response.data === 'something')`, power-assert will display the entire `response` object. This includes Node.js HTTP objects which are exposed by `axios`. These include symbol keys. [Our version of `pretty-format`](https://github.com/avajs/pretty-format) contains a bug which causes it to crash when it encounters a symbol key. I've opened a PR to fix that: https://github.com/avajs/pretty-format/pull/2

[...]

PR #1265, opened 2017-02-15
Ensure test run failures crash worker

novemberborn:
The first commit is there to catch more errors. The second commit ensures errors like https://github.com/avajs/ava/issues/1263 are surfaced without the test hanging.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -70,7 +70,23 @@ globals.setImmediate(() => {
 	runner.on('test', test);
 
 	process.on('ava-run', options => {
-		runner.run(options).then(exit);
+		runner.run(options)
+			.then(exit)
+			.catch(err => {
+				let exception;
+				try {
+					exception = serializeError(err);
+				} catch (_) {
+					// Avoid using serializeError
+					const err = new Error('Runner failed with an unserializable exception');
+					exception = {
+						name: 'Error',
+						message: err.message,
+						stack: err.stack
+					};
+				}
+				send('uncaughtException', {exception});
+			});
 	});
 
 	process.on('ava-init-exit', () => {
--- a/lib/main.js
+++ b/lib/main.js
@@ -73,19 +73,7 @@ globals.setImmediate(() => {
 		runner.run(options)
 			.then(exit)
 			.catch(err => {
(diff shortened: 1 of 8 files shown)


novemberborn:
Added tests to last commit. Let's hope they pass in CI 😄

lib/main.js
```diff
@@ -70,7 +70,23 @@ globals.setImmediate(() => {
...
+			.then(exit)
+			.catch(err => {
+				let exception;
+				try {
+					exception = serializeError(err);
+				} catch (_) {
+					// Avoid using serializeError
+					const err = new Error('Runner failed with an unserializable exception');
+					exception = {
+						name: 'Error',
```
sindresorhus: Super nitpick, but would prefer `err.name` here just for consistency with the other props and more future proof if we decide to change the error type or something.

lib/main.js
```diff
@@ -70,7 +70,23 @@ globals.setImmediate(() => {
...
-		runner.run(options).then(exit);
+		runner.run(options)
+			.then(exit)
+			.catch(err => {
+				let exception;
+				try {
+					exception = serializeError(err);
+				} catch (_) {
+					// Avoid using serializeError
+					const err = new Error('Runner failed with an unserializable exception');
```
sindresorhus: I wonder, should we put this logic in `serializeError` so it handles reporting about unserializable exceptions everywhere?

test/fixture/trigger-worker-exception/test-fallback.js
```diff
@@ -0,0 +1,5 @@
+import test from '../../../';
+
+test(() => {
+	return Promise.reject(new Error('Hi :)'));
```
sindresorhus: Would look better to use async function here and throw instead. Same with the below fixture.

lib/main.js
novemberborn: It's only called in three places. I'd prefer letting the error propagate correctly, and then treating it as an uncaught exception.

That said I've changed this code to `process.emit('uncaughtException', err)` and moved the fallback stuff to the `uncaughtException` handler in `test-worker.js`.

Issue #1266, opened 2017-02-15
mini reporter fails on files with a sourcemap source URLs


davewasmer:
## Description

For files that are transpiled before they reach ava, if the sourcemap includes a source URL, the mini reporter attempts to load the source code (via [code-excerpt.js](https://github.com/avajs/ava/blob/master/lib/code-excerpt.js#L15)), but it attempts to load it from the source URL specified in the source map. Since the code went through a build step prior to reaching ava, there's no guarantee that a relative source URL is still valid and exists. If the file does not exist, ava will error out.

### Error Message & Stack Trace

In the stack trace below, the `command-acceptance.ts` file does not exist, but `command-acceptance.js` does.

```
/Users/daw/projects/oss/denali/denali/node_modules/ava/lib/cli.js:174
                                        throw err;
                                        ^

Error: ENOENT: no such file or directory, open '/Users/daw/projects/oss/denali/denali/tmp/test/lib/test/command-acceptance.ts'
    at Error (native)
    at Object.fs.openSync (fs.js:640:18)
    at Object.fs.readFileSync (fs.js:508:33)
    at module.exports (/Users/daw/projects/oss/denali/denali/node_modules/ava/lib/code-excerpt.js:15:20)
    at runStatus.errors.forEach (/Users/daw/projects/oss/denali/denali/node_modules/ava/lib/reporters/mini.js:191:22)
    at Array.forEach (native)
    at MiniReporter.finish (/Users/daw/projects/oss/denali/denali/node_modules/ava/lib/reporters/mini.js:178:21)
    at Logger.finish (/Users/daw/projects/oss/denali/denali/node_modules/ava/lib/logger.js:38:28)
    at api.run.then.runStatus (/Users/daw/projects/oss/denali/denali/node_modules/ava/lib/cli.js:167:12)
From previous event:
    at Object.exports.run.err [as run] (/Users/daw/projects/oss/denali/denali/node_modules/ava/lib/cli.js:166:5)
    at Object.<anonymous> (/Users/daw/projects/oss/denali/denali/node_modules/ava/cli.js:22:24)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.runMain (module.js:604:10)
    at run (bootstrap_node.js:394:7)
    at startup (bootstrap_node.js:149:9)
[...]

PR #1267, opened 2017-02-15
Only add basePath to error stacktraces with relative source file paths

davewasmer:
Some sourcemap tools can produce absolute paths for source map urls, which will then be reflected in error stacktraces that feature absolute paths for the stack frame source url.

Previously, the basePath (typically `process.cwd()`) was prefixed to all error stacktrace urls, which sometimes works for relative paths[1], but will fail with absolute paths.

This change checks if the stacktrace url is absolute first - if so, it skips prefixing the url with basePath.

This change is more than cosmetic - the mini reporter uses the stacktrace url to invoke the `codeExcerpt` helper, which tries to load the source file from that url. If it's not correct, `codeExcerpt` will throw.

--------

[1] - See #1266. Intermediate build steps between the sourcemap generation and running ava can break relative urls too. This change does not solve that problem - it only fixes the absolute path scenario.

### DIFF
--- a/lib/reporters/mini.js
+++ b/lib/reporters/mini.js
@@ -187,7 +187,10 @@ class MiniReporter {
 				if (test.error.source) {
 					status += '  ' + colors.errorSource(test.error.source.file + ':' + test.error.source.line) + '\n';
 
-					const errorPath = path.join(this.options.basePath, test.error.source.file);
+					let errorPath = test.error.source.file;
+					if (!path.isAbsolute(test.error.source.file)) {
+						errorPath = path.join(this.options.basePath, test.error.source.file);
+					}
 					const excerpt = codeExcerpt(errorPath, test.error.source.line, {maxWidth: process.stdout.columns});
 					if (excerpt) {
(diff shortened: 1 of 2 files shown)


novemberborn:
Hi @davewasmer. This is similar to #1246. We've decided to ensure `error.source.file` is an absolute path so the reporters don't have to deal with this. I'm working on that at the moment.

Thanks for opening this pull request. Keep them coming 😄

davewasmer:
@novemberborn thanks for the quick response!

Just out of curiosity - how do you plan to get the absolute source url for files that have relative sourcemaps? I didn't see a way around that one, so I'm curious what you've come up with.

novemberborn:
> Just out of curiosity - how do you plan to get the absolute source url for files that have relative sourcemaps? I didn't see a way around that one, so I'm curious what you've come up with.

See https://github.com/avajs/ava/pull/1271. I'm basically assuming that V8 and any source map library make them relative to the working directory. It's best effort, really.

PR #1271, opened 2017-02-16
Code excerpt fixes

### LINKED ISSUE #1268: Restrict code excerpt sources
* [ ] Don't show a code excerpt if the source doesn't exist on disk
* [ ] Don't show a code excerpt if the source exists outside the project directory
* [ ] Don't show a code excerpt if the source is within a `node_modules` directory (within the project directory)

This is a follow-up issue from https://github.com/avajs/ava/issues/1231.

### LINKED ISSUE #1270: Resolve absolute source file path when serializing errors
The source file should be an absolute path. This should be done when the error is serialized, not when the file is accessed.

Related to https://github.com/avajs/ava/pull/1246 and https://github.com/avajs/ava/issues/1267.

novemberborn:
Fixes #1270 
Fixes #1268


vadimdemedes:
This is nice! I like the new `isWithinProject` and `isDependency` props.

Builds are failing though. Otherwise, 👍

novemberborn:
> Builds are failing though

Fixed now. On my local machine `tap` was including `source-map-support`, but that didn't seem to be happening on CI. Perhaps a caching issue, but it's better to include it explicitly anyway.

sindresorhus:
Excellent :)

PR #1281, opened 2017-02-22
Fix reporter bug when rejecting non-errors

jakwuh:
To reproduce try ava with the following test:

```js
import test from 'ava';

test(t => {
    Promise.reject(1);
});
```

### DIFF
--- a/lib/run-status.js
+++ b/lib/run-status.js
@@ -18,10 +18,15 @@ function sum(arr, key) {
 }
 
 function normalizeError(err) {
-	if (!isObj(err)) {
+	if (isObj(err) && 'message' in err && typeof err.message === 'string') {
+		['name', 'stack'].forEach(field => {
+			if (field in err && typeof err[field] !== 'string') {
+				delete err[field];
+			}
+		});
+	} else {
 		err = {
-			message: err,
-			stack: err
+			message: String(err)
 		};
 	}
 
--- a/lib/run-status.js
+++ b/lib/run-status.js
@@ -18,13 +18,7 @@ function sum(arr, key) {
 }
 
 function normalizeError(err) {
-	if (isObj(err) && 'message' in err && typeof err.message === 'string') {
-		['name', 'stack'].forEach(field => {
-			if (field in err && typeof err[field] !== 'string') {
-				delete err[field];
-			}
-		});
-	} else {
+	if (!isObj(err) || typeof err.message !== 'string') {
 		err = {
 			message: String(err)
 		};
--- a/lib/serialize-error.js
+++ b/lib/serialize-error.js
@@ -20,6 +20,11 @@ function filter(propertyName, isRoot, source, target) {
 		return true;
 	}
 
+	if ((propertyName === 'name' || propertyName === 'stack') &&
+		typeof source[propertyName] !== 'string') {
+		return false;
+	}
+
 	if (propertyName === 'stack') {
 		target.stack = beautifyStack(source.stack);
 		return false;
--- a/test/run-status.js
+++ b/test/run-status.js
@@ -87,3 +87,53 @@ test('calculate remaining test count', t => {
 	t.end();
 });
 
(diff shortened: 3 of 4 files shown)


novemberborn:
Interesting find. I think the error comes from this line, really: https://github.com/avajs/ava/blob/98dded553c8c213739cea9743ab4662f40d3125a/lib/run-status.js#L24

Unhandled rejections are normalized: https://github.com/avajs/ava/blob/98dded553c8c213739cea9743ab4662f40d3125a/lib/run-status.js#L72

Perhaps the `message` should be set to `String(err)` and `stack` shouldn't be set at all. Further, error objects that have a non-string `message`, `name` or `stack` should have those properties deleted. That would be in https://github.com/avajs/ava/blob/98dded553c8c213739cea9743ab4662f40d3125a/lib/serialize-error.js#L59.

Long story short, this means errors have the expected shape in the reporters and we don't have to keep checking if certain properties are strings etc. @avajs/core thoughts?

sindresorhus:
> Long story short, this means errors have the expected shape in the reporters and we don't have to keep checking if certain properties are strings etc. @avajs/core thoughts?

👍

jakwuh:
@novemberborn thank you for the response. I think having `message: String(err)` + removing `stack` property is the best solution here. Isn't it?

novemberborn:
@jakwuh yea. Though that's just for rejection reasons that aren't objects. We also need to sanitize objects with unexpected properties.

jakwuh:
@novemberborn here we go, now regardfully listening to any feedback :)

lib/run-status.js
```diff
@@ -18,10 +18,15 @@ function sum(arr, key) {
 }
 
 function normalizeError(err) {
-	if (!isObj(err)) {
+	if (isObj(err) && 'message' in err && typeof err.message === 'string') {
+		['name', 'stack'].forEach(field => {
+			if (field in err && typeof err[field] !== 'string') {
+				delete err[field];
```
novemberborn: I think this really does need to go here: https://github.com/avajs/ava/blob/98dded553c8c213739cea9743ab4662f40d3125a/lib/serialize-error.js#L59

We don't want to modify `err` since it comes from user code. There may be side-effects. But we can't easily copy error objects since `name` and `message` aren't always enumerable. After the call to `cleanYamlObject` in `serialize-error.js` we have a copy we can safely modify.

Also, in this case, I'd prefer a little bit of duplication over your `forEach` trick. IMHO this is a lot easier to follow:

```js
if ('name' in err && typeof err.name !== 'string') {
  delete err.name
}
if ('stack' in err && typeof err.stack !== 'string') {
  delete err.stack
}
```

Even better, `delete` only deletes own-properties, so the `'name' in err` guard is unnecessary.

jakwuh: Yes, this totally makes sense, still I find fairly to say that errors coming to `normalizeError` are already serialized (as I can see from the [test-worker](https://github.com/avajs/ava/blob/master/lib/test-worker.js#L42) code).
Also thank you a lot for your fast and detailed feedback.

jakwuh:
@novemberborn here we go 🙂

lib/serialize-error.js
```diff
@@ -20,6 +20,11 @@ function filter(propertyName, isRoot, source, target) {
 		return true;
 	}
 
+	if ((propertyName === 'name' || propertyName === 'stack') &&
+		typeof source[propertyName] !== 'string') {
```
novemberborn: @sindresorhus any thoughts on this? XO doesn't complain 😉

sindresorhus: Why should it have complained?

sindresorhus:
Woot! Very good work @jakwuh :)

lib/serialize-error.js
novemberborn: Figured it may have opinions on if condition indenting.

jakwuh:
Thanks  😊

Issue #1280, opened 2017-02-22
t.is can cause tests to hang


lukechilds:
`t.is` shows a nice diff if strings don't match which is pretty cool. However it appears that it can choke on large strings and cause the tests to hang which is not so cool.

I was comparing two b64 string representations  of an image, and it caused the Travis VM to hang for over 10mins until it got shut down. 

For example, this will cause AVA to hang:

```js
t.is(b64, expectedB64);
```

So I'm currently using this instead, which is fine:

```js
t.true(b64 === expectedB64);
```

It really threw me off and took me waaay longer to figure out what was going on because of AVA hanging. I could ctrl-c to get my terminal back but there where AVA processes still running that I had to manually kill each time. I'm not sure if this is a bug in the diff implementation or maybe it's expected that a large random string will cause it to hang. Even then, it would be better if this could be handled more elegantly, possibly a max time limit that the diff is allowed to run for before it should be killed?

For reference the two b64 strings I was comparing were:

```
[...]

PR #1285, opened 2017-02-24
Use `diff-match-patch` for text diffs

lukechilds:
Resolves https://github.com/avajs/ava/issues/1280

`diff` is really slow for large strings. This just swaps out the `diff` npm package for `diff-match-patch` which is waaaay quicker. Down from ~20mins to ~1.5s in the linked issue.

### DIFF
--- a/lib/format-assert-error.js
+++ b/lib/format-assert-error.js
@@ -2,6 +2,7 @@
 const indentString = require('indent-string');
 const chalk = require('chalk');
 const diff = require('diff');
+const DiffMatchPatch = require('diff-match-patch');
 
 const cleanUp = line => {
 	if (line[0] === '+') {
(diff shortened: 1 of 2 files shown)


vadimdemedes:
Thanks for submitting such an easy-win PR, love it! Obviously, I'm all for perf optimization, especially with such a little change. However, this module seems to be unmaintained. The last commit is from 2015, despite having open issues and PRs. 

Perhaps you are aware of some alternatives?

Also, can we swap the usage of 'diff' to the other library, so that we don't have 2 dependencies that basically do the same thing?

lukechilds:
Ah, I chose that module as it was recommended in the issue: https://github.com/avajs/ava/issues/1280#issuecomment-282258975

Didn't realise it was unmaintained, I'll have a look for some others.

> Also, can we swap the usage of 'diff' to the other library, so that we don't have 2 dependencies that basically do the same thing?

Only if it supports generating a patch. The `diff` lib was being used to generate a diff for strings and a patch for arrays/objects.

vadimdemedes:
Well, if there are no viable and just-as-fast diff libraries out there, we can go with this one at least for string diffs, I think. It's better to use this library, than to have 20m of waiting time.

lukechilds:
The only other I've found is: https://github.com/liddiard/text-diff

It's also pretty outdated and is just a wrapper around the same Google lib that the other project is a clone of: https://code.google.com/p/google-diff-match-patch/

vadimdemedes:
This one doesn't have tests, so it's not a good candidate. Ok, so I see there aren't that many diff libs out there, I think we may just go with the one you used in the PR. I will test how it works and what output it provides, thank you!

lukechilds:
No probs 👍

lukechilds:
Is it not already fixed in https://github.com/avajs/ava/pull/1285/commits/d402277ed8670eec5a4276ff9236d12589046959?

lukechilds:
@novemberborn Is it not already fixed in https://github.com/avajs/ava/pull/1285/commits/d402277ed8670eec5a4276ff9236d12589046959?

novemberborn:
Oh interesting. My merge button was set to *Rebase and merge*, which can't work, but I *can* squash and merge.

novemberborn:
Thanks @lukechilds!

lukechilds:
I like to call that squirging. No problem 👍

leoswing:
> I like to call that squirging. No problem 👍

Does the google diff generate a patch? which it could be used for diff2html and produce a nice diff like git diff

Issue #1286, opened 2017-02-25
String diff output starts red and ends white


lukechilds:
I think this is a bug, although maybe it's intentional and I'm missing something.

The text output for the string diff always starts red for neutral text (as in text that has been neither added or removed in the diff). Then elsewhere the neutral text is white not red. You can see what I mean in the screenshot from the readme:

![](https://github.com/avajs/ava/raw/a45cdb9eb7e04097952b3a88b504e12d8885fbd3/media/magic-assert-strings.png)

Notice how `"this is so` is red, but the last `"` is white.

I think all the text should be white apart from `cool|bad`.

vadimdemedes:
I agree, we should change the string color from red to white. Thanks for reporting this!

lukechilds:
No probs, happy to submit a PR but I don't seem to be able to figure out where it's coming from.

This is the code for the diff:

```js
if (err.actualType === 'string' && err.expectedType === 'string') {
  const patch = diff.diffChars(err.actual, err.expected);
  const msg = patch
    .map(part => {
      if (part.added) {
        return chalk.bgGreen.black(part.value);
      }

      if (part.removed) {
        return chalk.bgRed.black(part.value);
      }

      return part.value;
    })
    .join('');

  return `Difference:\n\n${msg}\n`;
}
```

As you can see the section for the neutral text applies no colouring: `return part.value;`. So it's weird that it's red the first time it's returned but not the other times. That makes me think that it's inheriting it from a previous unclosed ANSI colour code, however the string is interpolated into the same string that starts with the word `Difference:`:

```js
return `Difference:\n\n${msg}\n`;
```

But `Difference:` isn't red. Any ideas?

vadimdemedes:
The red color for strings is a default value set in @ava/pretty-format module (at the end of index.js in default props for 'theme' prop, if I'm not mistaken).

Now that I think of it, it'd be inconsistent if strings in other types of output (comparing objects, for example) were red, but white in string diff?

lukechilds:
Ok, so that flips the problem round, it's the white text that is incorrect. Instead of the neutral colour being white it should be red.

lukechilds:
Hmmn, I'm struggling to think of a clean way to fix this. Obviously a really simple fix would be to just colour the neutral text red, but then if the string pretty-format colour changes then it'll be wrong.

So we could check the default pretty-format colour and use that, but then if pretty-format was ran with custom colour options we won't know and again the colour will be wrong.

[...]

PR #1287, opened 2017-02-25
Fix string diff colours

lukechilds:
Make sure all string diff output is red by default.

Resolves https://github.com/avajs/ava/issues/1286

### DIFF
--- a/lib/format-assert-error.js
+++ b/lib/format-assert-error.js
@@ -1,5 +1,6 @@
 'use strict';
 const indentString = require('indent-string');
+const stripAnsi = require('strip-ansi');
 const chalk = require('chalk');
 const diff = require('diff');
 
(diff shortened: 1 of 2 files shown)


vadimdemedes:
Tests are failing, but other than that - works great, just tested it!

lib/format-assert-error.js
```diff
@@ -2,6 +2,7 @@
 const indentString = require('indent-string');
 const chalk = require('chalk');
 const diff = require('diff');
+const stripAnsi = require('strip-ansi');
```
vadimdemedes: Could you move it above `chalk`?

lukechilds: 👌

lukechilds:
Fixed test and updated `require` position 👍

vadimdemedes:
Thank you, good fix!

PR #1290, opened 2017-02-28
Improve Babel recipe

### LINKED ISSUE #727: Document how to use babel-polyfill to polyfill the test environment
AVA does not load polyfills when running tests. This means features like `['foo', 'bar'].includes('foo')` are not available. We should update the [ES2015](https://github.com/sindresorhus/ava#es2015-support) docs to point this out, and add a recipe explaining how to use `babel-polyfill` with AVA.

There should be a caveat to the latter that users shouldn't load `babel-polyfill` unless they _also_ use it in their applications, since the polyfill will apply to both test _and_ source files.

See #263 for background.

florianb:
- Fixes #727 by describing usage of `babel-polyfill`

### DIFF
--- a/docs/recipes/babelrc.md
+++ b/docs/recipes/babelrc.md
@@ -30,6 +30,21 @@ You can override the default Babel configuration AVA uses for test transpilation
 }
 ```
 
+## Provide ES2015 Polyfills to your tests
+
+You might run into issues when relying on ES2015-features your current environment do not support. Since Ava isn't providing any polyfills per default you might want to enable them by adding `babel-polyfill` manually to the `require` section:
+
+```json
+{
+	"ava": {
+    "require": [
+			"babel-register",
+			"babel-polyfill"
+		]
+  }
+}
+```
+
 ## Transpiling Sources
 
 To transpile your sources, you will need to define a [`babel config` ](http://babeljs.io/docs/usage/babelrc/) in `package.json` or a `.babelrc` file. Also, you will need to tell AVA to load [`babel-register`](http://babeljs.io/docs/usage/require/) in every forked process, by adding it to the `require` section of your AVA config:
--- a/docs/recipes/babelrc.md
+++ b/docs/recipes/babelrc.md
@@ -30,15 +30,18 @@ You can override the default Babel configuration AVA uses for test transpilation
 }
 ```
 
-## Provide ES2015 Polyfills to your tests
+## Use Babel Polyfills
 
-You might run into issues when relying on ES2015-features your current environment do not support. Since Ava isn't providing any polyfills per default you might want to enable them by adding `babel-polyfill` manually to the `require` section:
+AVA lets you write your tests using new JavaScript syntax, even on Node.js versions that otherwise wouldn't support it. However, it doesn't add or modify built-ins of your current environment. Using AVA would, for example, not provide modern features such as `Array.prototype.includes()` to an underlying NodeJ.js 4 environment.
+
+By loading [Babel's Polyfill module](https://babeljs.io/docs/usage/polyfill/) you're able to opt these features in. Note that this will modify the environment, which may influence how your program itself behaves.
+
+You can enable `babel-polyfill` by adding it to AVA's require options:
 
 ```json
 {
   "ava": {
     "require": [
-      "babel-register",
       "babel-polyfill"
     ]
   }
--- a/docs/recipes/babelrc.md
+++ b/docs/recipes/babelrc.md
@@ -32,11 +32,11 @@ You can override the default Babel configuration AVA uses for test transpilation
 
 ## Use Babel Polyfills
 
-AVA lets you write your tests using new JavaScript syntax, even on Node.js versions that otherwise wouldn't support it. However, it doesn't add or modify built-ins of your current environment. Using AVA would, for example, not provide modern features such as `Array.prototype.includes()` to an underlying NodeJ.js 4 environment.
+AVA lets you write your tests using new JavaScript syntax, even on Node.js versions that otherwise wouldn't support it. However, it doesn't add or modify built-ins of your current environment. Using AVA would, for example, not provide modern features such as `Array.prototype.includes()` to an underlying Node.js 4 environment.
 
-By loading [Babel's Polyfill module](https://babeljs.io/docs/usage/polyfill/) you're able to opt these features in. Note that this will modify the environment, which may influence how your program itself behaves.
+By loading [Babel's Polyfill module](https://babeljs.io/docs/usage/polyfill/) you're able to opt these features in. Note that this will modify the environment, which may influence how your program behaves.
 
-You can enable `babel-polyfill` by adding it to AVA's require options:
+You can enable `babel-polyfill` by adding it to AVA's `require` option:
 
 ```json
 {
--- a/docs/recipes/babelrc.md
+++ b/docs/recipes/babelrc.md
@@ -13,9 +13,7 @@ There are multiple options for configuring how AVA transpiles your tests using B
 
 ## AVA's default transpiler behavior
 
-AVA by default provides some nifty JavaScript features, like [`async`](https://github.com/avajs/ava#async-function-support), to your test-environment. To make this possible AVA transpiles the tests and helper files using the [`@ava/stage-4`](https://github.com/avajs/babel-preset-stage-4) Babel preset.
-
-This is a great feature for projects where you desire to not transpile your source before deploying but enjoy using ECMAScript for writing tests.
+AVA lets you use some nifty JavaScript features, like [async functions](https://github.com/avajs/ava#async-function-support). To make this work on older Node.js versions AVA transpiles the tests and helper files using the [`@ava/stage-4`](https://github.com/avajs/babel-preset-stage-4) Babel preset. This is great for projects where you do not use Babel for your source, but do want to use the newest JavaScript features for your tests.
 
 ## Customizing how AVA transpiles your tests
 
@@ -36,7 +34,7 @@ You can override the default Babel configuration AVA uses for test transpilation
 
 AVA lets you write your tests using new JavaScript syntax, even on Node.js versions that otherwise wouldn't support it. However, it doesn't add or modify built-ins of your current environment. Using AVA would, for example, not provide modern features such as `Array.prototype.includes()` to an underlying Node.js 4 environment.
 
-By loading [Babel's Polyfill module](https://babeljs.io/docs/usage/polyfill/) you can opt in to these features in. Note that this will modify the environment, which may influence how your program behaves.
+By loading [Babel's Polyfill module](https://babeljs.io/docs/usage/polyfill/) you can opt in to these features. Note that this will modify the environment, which may influence how your program behaves.
 
 You can enable `babel-polyfill` by adding it to AVA's `require` option:
 
--- a/docs/recipes/babelrc.md
+++ b/docs/recipes/babelrc.md
@@ -13,7 +13,9 @@ There are multiple options for configuring how AVA transpiles your tests using B
 
 ## AVA's default transpiler behavior
 
-By default, AVA transpiles your tests and helper files using the [`@ava/stage-4`](https://github.com/avajs/babel-preset-stage-4) Babel preset. This is a great option for small modules where you do not desire a build step to transpile your source before deploying to `npm`.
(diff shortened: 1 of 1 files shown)


docs/recipes/babelrc.md
```diff
@@ -30,6 +30,21 @@ You can override the default Babel configuration AVA uses for test transpilation
...
 
+## Provide ES2015 Polyfills to your tests
+
+You might run into issues when relying on ES2015-features your current environment do not support. Since Ava isn't providing any polyfills per default you might want to enable them by adding `babel-polyfill` manually to the `require` section:
+
+```json
+{
+  "ava": {
+    "require": [
+      "babel-register",
```
novemberborn: This isn't necessary for `babel-polyfill`, so let's leave it out.

docs/recipes/babelrc.md
```diff
@@ -30,6 +30,21 @@ You can override the default Babel configuration AVA uses for test transpilation
 }
 ```
 
+## Provide ES2015 Polyfills to your tests
```
novemberborn: Perhaps `Polyfill tests`? I'm trying to avoid `ES2015` mentions since the standard is continuously evolving. AVA tries to support stage-4 syntax which means currently we're supporting more than ES2016, but not ES2017 since it isn't official yet.

docs/recipes/babelrc.md
```diff
@@ -30,6 +30,21 @@ You can override the default Babel configuration AVA uses for test transpilation
...
+
+You might run into issues when relying on ES2015-features your current environment do not support. Since Ava isn't providing any polyfills per default you might want to enable them by adding `babel-polyfill` manually to the `require` section:
```
novemberborn: This is a good start. I think it could be clearer on what AVA does out of the box, and what kind of features end up being missing. There's also a big downside to using `babel-polyfill`, which is that it impacts how your program behaves (not just the tests!)

What do you think about this:

> AVA lets you write tests using new JavaScript syntax, even on Node.js version that do not support it otherwise. However it doesn't add or modify built-ins. For example `Array.prototype.includes()` is not available in Node.js 4 even when you use AVA.
>
> You can opt in to these features by loading [Babel's Polyfill](https://babeljs.io/docs/usage/polyfill/) module. Note that this modifies the environment, which impacts how your program itself behaves.
>
> You can load `babel-polyfill` by adding it to AVA's `require` options:

docs/recipes/babelrc.md
florianb: Thank you @novemberborn - that's a absolute valid point. I will change that.
- [ ] Remove ES2015-reference

docs/recipes/babelrc.md
florianb: Thank you very much @novemberborn - i somehow assumed it could be wanted to write not too much. 🤔 I guess that assumption was just a bit stupid. 😄

- [ ] Elaborate the description

docs/recipes/babelrc.md
florianb: Absolutely, thanks!

- [ ] Remove the polyfill-snippet

docs/recipes/babelrc.md
florianb: @novemberborn I rewrite the description. I think it reads better now (even though my mother-tongue is German, so don't expect too much) but i have still a note on it:

- The documentation should probably elaborate on the distinction between the execution environment of tests and the execution environment of the tested code. The current wording may leave that as a question to the beginner.

docs/recipes/babelrc.md
```diff
@@ -30,15 +30,18 @@ You can override the default Babel configuration AVA uses for test transpilation
 }
 ```
 
-## Provide ES2015 Polyfills to your tests
+## Use Babel Polyfills
 
-You might run into issues when relying on ES2015-features your current environment do not support. Since Ava isn't providing any polyfills per default you might want to enable them by adding `babel-polyfill` manually to the `require` section:
+AVA lets you write your tests using new JavaScript syntax, even on Node.js versions that otherwise wouldn't support it. However, it doesn't add or modify built-ins of your current environment. Using AVA would, for example, not provide modern features such as `Array.prototype.includes()` to an underlying NodeJ.js 4 environment.
```
novemberborn: There's a typo in `NodeJ.js`.

docs/recipes/babelrc.md
```diff
@@ -30,15 +30,18 @@ You can override the default Babel configuration AVA uses for test transpilation
...
+
+By loading [Babel's Polyfill module](https://babeljs.io/docs/usage/polyfill/) you're able to opt these features in. Note that this will modify the environment, which may influence how your program itself behaves.
```
novemberborn: > able to opt in to these features.

Let's get rid of `itself` in `how your program itself behaves`.

docs/recipes/babelrc.md
```diff
@@ -30,15 +30,18 @@ You can override the default Babel configuration AVA uses for test transpilation
...
+
+You can enable `babel-polyfill` by adding it to AVA's require options:
```
novemberborn: Perhaps:

> … `require` option:

(Marking `require` as code, and singular "option".)

docs/recipes/babelrc.md
florianb: Sorry for that..

- [ ] Fix typo

docs/recipes/babelrc.md
florianb: - [ ] Remove `itself`

docs/recipes/babelrc.md
florianb: - [ ] make `require`-option more explicit

florianb:
Thanks for your understanding. 😄

So what we would need to say is that the user must ensure he provides the same "polyfilled" environment for AVA as he does to the execution environment of his program (in the case he uses more features than defined in AVA's stage-4)?
In my opinion this is a more general topic and would better be pointed out at the beginning of the whole file, probably in addition to the default's transpiler behavior section?

novemberborn:
> So what we would need to say is that the user must ensure he provides the same "polyfilled" environment for AVA as he does to the execution environment of his program (in the case he uses more features than defined in AVA's stage-4)

Yes.

> In my opinion this is a more general topic and would better be pointed out at the beginning of the whole file, probably in addition to the default's transpiler behavior section?

AVA's test transpilation doesn't impact the user's program. I agree it's a general topic which is why I was trying to get away with just hinting at it.

florianb:
Okay, i guess i'd like to try to clarify this a bit more on the default's transpiler behavior section. If you don't mind, of course. 😄

novemberborn:
@florianb by all means.

florianb:
Okay @novemberborn, i hope i could make the description of the default transpilation behavior a little bit more clear. I tried to become a little bit more explicit, even though i am not sure i got the  meaning of

> This is a great option for small modules where you do not desire a build step to transpile your source before deploying to `npm`

I think that former sentence implies that AVA allows you using ES-features even in projects you otherwise don't do any transpilation.

docs/recipes/babelrc.md
```diff
@@ -30,6 +32,24 @@ You can override the default Babel configuration AVA uses for test transpilation
 }
 ```
 
+## Use Babel Polyfills
+
+AVA lets you write your tests using new JavaScript syntax, even on Node.js versions that otherwise wouldn't support it. However, it doesn't add or modify built-ins of your current environment. Using AVA would, for example, not provide modern features such as `Array.prototype.includes()` to an underlying Node.js 4 environment.
+
+By loading [Babel's Polyfill module](https://babeljs.io/docs/usage/polyfill/) you're able to opt these features in. Note that this will modify the environment, which may influence how your program behaves.
```
novemberborn: > By loading Babel's Polyfill module you can opt in to these features. …

docs/recipes/babelrc.md
```diff
@@ -13,7 +13,9 @@ There are multiple options for configuring how AVA transpiles your tests using B
 
 ## AVA's default transpiler behavior
 
-By default, AVA transpiles your tests and helper files using the [`@ava/stage-4`](https://github.com/avajs/babel-preset-stage-4) Babel preset. This is a great option for small modules where you do not desire a build step to transpile your source before deploying to `npm`.
+AVA by default provides some nifty ECMAScript features, like [`async`](https://github.com/avajs/ava#async-function-support), to your test-environment. To make this possible AVA transpiles the tests and helper files using the [`@ava/stage-4`](https://github.com/avajs/babel-preset-stage-4) Babel preset.
```
novemberborn: Let's just call it `JavaScript`.

How about:

> AVA lets you use some nifty JavaScript features in your tests, like [async functions](https://github.com/avajs/ava#async-function-support). To make this work on older Node.js versions AVA transpiles the tests and helper files using the [`@ava/stage-4`](https://github.com/avajs/babel-preset-stage-4) Babel preset. This is great for projects where you don't otherwise use Babel for your source, but do want to use the newest JavaScript features in your tests.

florianb: I am somehow torn between using JS over ECMAScript for these reasons:

- JavaScript sounds just slight and handy
- This paragraph is about the differences between good old vanilla JS and the new (not backward compatible) language features introduced by ES.

I think this topic is read by people who already know the term "ECMAScript" and i think in this case it supports understanding (and avoids confusion for less experienced programmers ["Isn't that feature an ES-feature?"]).

But i am really not insisting on that - so if you'd prefer using JS, i am fine with that.

docs/recipes/babelrc.md
florianb: Thanks! 😄 

- [ ] fix style

docs/recipes/babelrc.md
novemberborn: Our main page talks about JavaScript, so I don't see why we should start using a different term here.

florianb: Okay - i just thought about the use of ["ES2017"](https://github.com/avajs/ava#es2017-support) to explicitly spot on ES-features.

But as i said, i'm not insisting on it.

- [ ] Replace ES by JS

docs/recipes/babelrc.md
```diff
@@ -30,6 +32,24 @@ You can override the default Babel configuration AVA uses for test transpilation
 }
 ```
 
+## Use Babel Polyfills
+
+AVA lets you write your tests using new JavaScript syntax, even on Node.js versions that otherwise wouldn't support it. However, it doesn't add or modify built-ins of your current environment. Using AVA would, for example, not provide modern features such as `Array.prototype.includes()` to an underlying Node.js 4 environment.
+
+By loading [Babel's Polyfill module](https://babeljs.io/docs/usage/polyfill/) you can opt in to these features in. Note that this will modify the environment, which may influence how your program behaves.
```
novemberborn: Remove the "in" after "features".

docs/recipes/babelrc.md
```diff
@@ -13,7 +13,9 @@ There are multiple options for configuring how AVA transpiles your tests using B
 
 ## AVA's default transpiler behavior
 
-By default, AVA transpiles your tests and helper files using the [`@ava/stage-4`](https://github.com/avajs/babel-preset-stage-4) Babel preset. This is a great option for small modules where you do not desire a build step to transpile your source before deploying to `npm`.
+AVA by default provides some nifty JavaScript features, like [`async`](https://github.com/avajs/ava#async-function-support), to your test-environment. To make this possible AVA transpiles the tests and helper files using the [`@ava/stage-4`](https://github.com/avajs/babel-preset-stage-4) Babel preset.
+
+This is a great feature for projects where you desire to not transpile your source before deploying but enjoy using ECMAScript for writing tests.
```
novemberborn: Thanks for changing this to `JavaScript` (though it says `ECMAScript` in the second paragraph). I still think this reads better though:

> AVA lets you use some nifty JavaScript features in your tests, like [async functions](https://github.com/avajs/ava#async-function-support). To make this work on older Node.js versions AVA transpiles the tests and helper files using the [`@ava/stage-4`](https://github.com/avajs/babel-preset-stage-4) Babel preset. This is great for projects where you don't otherwise use Babel for your source, but do want to use the newest JavaScript features in your tests.

The latest Node.js version support most of ES2017, so it helps to explain why AVA still needs to do transpilation. I feel the sentences read a bit better as well.

novemberborn:
Thanks @florianb!

florianb:
Your are welcome @novemberborn - thanks for your help!

PR #1295, opened 2017-03-04
[Bug report] Add tests for observable-support: rxjs, xstream, most

shofel:
Ava's [observable support](https://github.com/avajs/ava#observable-support) works well for `xen-observable` and `rxjs`, but doesn't work for `xstream` and `most.js`.

There are two reasons for this.
Firstly, 'xstream' and 'most' observables don't have `.forEach()` method. This could be solved by sindresorhus/observable-to-promise#4.
Secondly, 'xstream' and 'most.js' observables are not detected by `isObservable()` function at [`lib/test.js::onAssertionEvent()`](https://github.com/avajs/ava/blob/157ef25aff0bae7145a6133928d2bdd0da853a78/lib/test.js#L75) and at [`lib/test.js::run()`](https://github.com/avajs/ava/blob/157ef25aff0bae7145a6133928d2bdd0da853a78/lib/test.js#L231).

At the same time `isObservable` detects them successfully in observable-to-promise project (see the tests in [PR#4](https://github.com/sindresorhus/observable-to-promise/pull/4))

### DIFF
--- a/package.json
+++ b/package.json
@@ -177,9 +177,11 @@
     "inquirer": "^3.0.5",
     "is-array-sorted": "^1.0.0",
     "lolex": "^1.4.0",
+    "most": "^1.2.2",
     "nyc": "^10.0.0",
     "proxyquire": "^1.7.4",
     "rimraf": "^2.5.0",
+    "rxjs": "^5.2.0",
     "signal-exit": "^3.0.0",
     "sinon": "^2.0.0",
     "source-map-fixtures": "^2.1.0",
@@ -187,6 +189,7 @@
     "temp-write": "^3.1.0",
     "touch": "^1.0.0",
     "xo": "^0.18.0",
+    "xstream": "^10.3.0",
     "zen-observable": "^0.5.1"
   },
   "xo": {
--- a/test/observable.js
+++ b/test/observable.js
@@ -136,12 +136,13 @@ Observable.of('zen-observable').subscribe(logger);
  * @param ctor {(x) => Observable}
  */
 function detects(name, ctor) {
-	test(`detects ${name} observables`, t => {
+	test(`ava detects ${name} observable as a return value`, t => {
 		ava(a => {
 			a.plan(1);
-
 			const observable = ctor(1);
-			observable.subscribe(x => a.is(x, 1));
+			observable.subscribe({
+				next: x => a.is(x, 1)
+			});
 			return observable;
 		}).run().then(result => {
 			t.is(result.passed, true);
--- a/test/observable.js
+++ b/test/observable.js
@@ -1,7 +1,10 @@
 'use strict';
 const test = require('tap').test;
-const Test = require('../lib/test');
-const Observable = require('zen-observable'); // eslint-disable-line import/order
+const Test = require('../lib/test'); // eslint-disable-line import/order
+const Observable = require('zen-observable');
+const xs = require('xstream').Stream;
+const Rx = require('rxjs');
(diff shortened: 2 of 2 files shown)


sindresorhus:
@shofel Can you fix the merge conflict? Do you know anyone in the Xstream/Most community we could mention here? I don't really have that much experience with Observables.

shofel:
@sindresorhus I've resolved the conflict.

After some time writing tests for streams I have some thoughts in general)))

Let's look at the example from ava documentation.
``` js
test(t => {
	t.plan(3);
	return Observable.of(1, 2, 3, 4, 5, 6)
		.filter(n => n % 2 === 0)
		.map(() => t.pass());
});
```
Using `.map()` for side effects is a hack in FRP and in FP in general. All the side effects should be isolated in subscribers. Thus above beautiful snippet turns into
``` js
test(t => {
	t.plan(3);
	return Observable.of(1, 2, 3, 4, 5, 6)
		.filter(n => n % 2 === 0)
		.subscribe({
			next: () => t.pass()
		});
});
```

But now we return not a stream, but a subscription object. Thus we need a var to save the stream:
``` js
test(t => {
	t.plan(3);
	let stream = Observable.of(1, 2, 3, 4, 5, 6)
		.filter(n => n % 2 === 0);
	stream.subscribe({
		next: () => t.pass()
	});
	return stream;
});
```

This is a lot less fruitful. Unexpectedly, it looks like the callback syntax is the most natural and semantically correct in this case:
``` js
test.cb(t => {
	t.plan(3);
	Observable.of(1, 2, 3, 4, 5, 6)
		.filter(n => n % 2 === 0)
		.subscribe({
			next: () => t.pass(),
			complete: () => t.end()
		});
});
```

Given all above, I don't really know if returning an observable from a test has got any advantage.

shofel:
I've been also thinking on `await` syntax for asserts
``` js
test(async t => {
	const oneTwoThree = xstream.from([1, 2, 3])
	t.deepEqual(await oneTwoThree, [1, 2, 3])
});
```

And it turned out the RxJS team have had [a discussion](https://github.com/ReactiveX/rxjs/issues/556) about this, and the take out of that was the decision to not have this because there are more caveats and ambiguity than profit.

shofel:
As for now, I write tests like this:
```js
import {toPromise} from './helpers/observable'

test('a stream of 1', async t => {
	let one = xs.of(1)
	t.deepEqual(await toPromise(one), [1])
})
```

Where `toPromise` is almost the same as [observable-to-promise](https://github.com/sindresorhus/observable-to-promise)
``` js
export function toPromise (observable) {
  let values = []
  let remember = x => { values = values.concat(x) }

  return new Promise((resolve, reject) => {
    observable.subscribe({
      error: reject,
      complete: () => resolve(values),
      next: remember
    })
  })
}
```

And as long as we check only values, but not timing, it works well.

novemberborn:
> Firstly, 'xstream' and 'most' observables don't have .forEach() method. This could be solved by sindresorhus/observable-to-promise#4.

This PR has landed, and AVA depends on the latest `observable-to-promise` version.

> Secondly, 'xstream' and 'most.js' observables are not detected by isObservable() function at lib/test.js::onAssertionEvent() and at lib/test.js::run().

This seems to be resolved:

```
> require("is-observable")(require('xstream').default.fromArray([1,2]))
true
> require("is-observable")(require('most').from([1,2]))
true
```

In conclusion I believe these issues have been resolved. Thanks for your report @shofel!

PR #1297, opened 2017-03-05
Separate config keys from CLI flags

### LINKED ISSUE #1046: Separate config keys from CLI flags

yatharthk:
Fixes #1046 

This separates out package config keys from CLI flags by picking up defaults from `conf` and passing them to `meow` as defaults. Furthermore, the resultant `cli.flags` is copied back `conf` so that it can be used elsewhere (instead of `cli.flags`).

@novemberborn The changes have been made as per our discussion about the same. Let me know if something doesn't seems like fitting in right!

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -21,7 +21,7 @@ const babelConfigHelper = require('./babel-config');
 Promise.longStackTraces();
 
 exports.run = () => {
-	const conf = pkgConf.sync('ava');
+	let conf = pkgConf.sync('ava');
 
 	const filepath = pkgConf.filepath(conf);
 	const projectDir = filepath === null ? process.cwd() : path.dirname(filepath);
@@ -72,7 +72,21 @@ exports.run = () => {
 			'update-snapshots',
 			'color'
 		],
-		default: Object.assign({color: true}, conf),
+		default: {
+			init: conf.init,
+			failFast: conf.failFast,
+			serial: conf.serial,
+			cache: conf.cache,
+			powerAssert: conf.powerAssert,
+			tap: conf.tap,
+			verbose: conf.verbose,
+			watch: conf.watch,
+			match: conf.match,
+			timeout: conf.timeout,
+			concurrency: conf.concurrency,
+			updateSnapshots: conf.updateSnapshots,
+			color: true || conf.color
+		},
 		alias: {
 			t: 'tap',
 			v: 'verbose',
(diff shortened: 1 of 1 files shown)


lib/cli.js
```diff
@@ -72,7 +72,21 @@ exports.run = () => {
...
+			cache: conf.cache,
+			powerAssert: conf.powerAssert,
+			tap: conf.tap,
+			verbose: conf.verbose,
+			watch: conf.watch,
+			match: conf.match,
+			timeout: conf.timeout,
+			concurrency: conf.concurrency,
+			updateSnapshots: conf.updateSnapshots,
+			color: true || conf.color
```
novemberborn: I think this should be `'color' in conf ? color.conf : true`. The idea is that it defaults to `true` if not configured at all, but otherwise the config value should prevail.

It's strange that this doesn't fail any tests though.

lib/cli.js
```diff
@@ -72,7 +72,21 @@ exports.run = () => {
...
+		},
```
novemberborn: Would you mind sorting the keys alphabetically?

lib/cli.js
```diff
@@ -107,31 +121,34 @@ exports.run = () => {
 		throw new Error(colors.error(figures.cross) + ' The --require and -r flags are deprecated. Requirements should be configured in package.json - see documentation.');
 	}
 
+	// copy resultant cli.flags into conf for use with Api and elsewhere
+	conf = Object.assign(conf, cli.flags);
```
novemberborn: `Object.assign(conf, …)` modifies and returns `conf`. There's no need to reassign it (since you're reassigning it to itself).

lib/cli.js
yatharthk: Oh, I don't know how I missed out on this. Was stupid! 😛 So what you're suggesting here is something like `color: conf.color ? conf.color : true`, right?

lib/cli.js
yatharthk: Sure. 👍

lib/cli.js
novemberborn: No, that'll still force the value to be true if `conf.color` is `false`.

If `color` is in `conf`, assume it's a usable value, else use `true`.

yatharthk: I get it. Then I guess we can use `(conf.color !== undefined) ? conf.color : true`. Any thoughts?

novemberborn: Why not `'color' in conf ? color.conf : true`?

yatharthk: Cool. No problem. I wasn't so sure about that. You just enhanced my knowledge. Thank you. :)

sindresorhus:
Nice work @yatharthk 👌

yatharthk:
@sindresorhus Happy to contribute 😄. Thanks to @novemberborn for all the help!

PR #1298, opened 2017-03-05
Make test context optionally type aware for TypeScript

mmkal:
This sets out to solve the problem highlighted in https://github.com/avajs/ava/issues/1291

Usage example: see comment further down.


.gitignore
```diff
@@ -3,3 +3,6 @@ node_modules
 coverage
 bench/.results
 types/generated.d.ts
+
+# VS Code configuration
+.vscode
```
novemberborn: You should add this to your machine's `.gitignore`, rather than adding it in this repo.

types/base.d.ts
```diff
@@ -99,20 +96,19 @@ export interface CallbackTestContext extends TestContext {
...
 	(t: T, ...args: any[]): void;
 	title? (providedTitle: string, ...args: any[]): string;
 }
 export type Macros<T> = Macro<T> | Macro<T>[];
 
-export function test(name: string, run: ContextualTest): void;
-export function test(run: ContextualTest): void;
-export function test(name: string, run: Macros<ContextualTestContext>, ...args: any[]): void;
-export function test(run: Macros<ContextualTestContext>, ...args: any[]): void;
+export interface ITest<T> {
```
novemberborn: Let's just use `Test<T>`.

types/base.d.ts
```diff
@@ -9,11 +7,10 @@ export type ErrorValidator
 export interface Observable {
 	subscribe(observer: (value: {}) => void): void;
 }
-
 export type Test = (t: TestContext) => PromiseLike<void> | Iterator<any> | Observable | void;
-export type ContextualTest = (t: ContextualTestContext) => PromiseLike<void> | Iterator<any> | Observable | void;
+export type GenericTest<T> = (t: GenericTestContext<T>) => PromiseLike<void> | Iterator<any> | Observable | void;
```
ivogabe: Add `export type ContextualTest = GenericTest<{ context: any }>;`, otherwise this might break some users. Same for `ContextualCallbackTest`

ivogabe:
Overall this looks good. Could you add some comments to `make.js`? It might be a bit hard to read now.

types/base.d.ts
mmkal: Can do - I think my C# is showing 😳

mmkal:
@novemberborn re whether we could do:
```TypeScript
function contextualize<T>(getContext: () => T): ava.ITest<T> { ... }
```
Maybe what I could do is export two types - `Test<T>` (renamed from `ITest`) and `GenericContextualTest<T>` which could be defined by
```TypeScript
export type GenericContextualTest<T> = Test<{ context: T }>;
```
The name `GenericContextualTest` isn't great, but I'll see what I can do about avoiding a clash with `ContextualTest` when implementing @ivogabe's suggestion.

Anyway, this way people wouldn't have to constantly repeat the `context: ...` part. I'll add something along those lines soon and update.

.gitignore
mmkal: Sure. I didn't want to do that initially because I have other repos that I want `.vscode` files checked in for, but maybe I can use `.git/info/exclude` instead.

mmkal:
@novemberborn @ivogabe updated with suggested changes.
Although I couldn't rename `ITest` to `Test` as there's already a `Test` type so I went with `DefineTest`.

mmkal:
@novemberborn @ivogabe updated with suggested changes, and improved a couple of other things:
I renamed `ITest` as suggested, but I am now using it as the interface for all `test`-like functions - and they are all now optionally generic. Plus, `types/generated.d.ts` is now about 30% smaller because each `test`-like function now implements this single interface rather than having its declaration spread over four lines.

Usage example:

```TypeScript
import * as ava from 'ava'

function contextualize<T>(getContext: () => T): ava.ContextualTestFunction<T> {
    ava.test.beforeEach(t => {
        Object.assign(t.context, getContext());
    });

    return ava.test;
}

const test = contextualize(() => ({ foo: 'bar' }));

test.beforeEach(t => {
    t.context.foo = 123; // error:  Type '123' is not assignable to type 'string'
});

test.after.always.failing.cb.serial('very long chains are properly typed', t => {
    t.context.fooo = 'a value'; // error: Property 'fooo' does not exist on type '{ foo: string }'
});

test('an actual test', t => {
    t.deepEqual(t.context.foo.map(c => c), ['b', 'a', 'r']); // error: Property 'map' does not exist on type 'string'
});
```

In the above example, `contextualize` is a generic method, and infers the type of the object passed to it to be `{ foo: string }`. By using the test method returned instead of the vanilla ava one, we can take advantage of type annotations. 

This would allow people to extend ava with things like their favourite assertion/mocking libraries, and have strongly typed, convenient test setups:

```TypeScript
import MyClass from '../MyClass';
import calc = require('npm-calculator');
import chai = require('chai');
import sinon = require('sinon');
import contextualize from './ava-contextualize'; // same contextualize method as above as a module

const test = contextualize(() => {
    const myClass = new MyClass();
    const sandbox = sinon.sandbox.create();
    const expect = chai.expect;
    return { myClass, sandbox, expect };
});

test('times uses npm-calculator multiply', t => {
    t.context.sandbox.stub(calc.multiply).returns(42);
    const sixTimesNine = t.context.myClass.times(6, 9);
    t.context.expect(calc.multiply).to.be.calledOnce;
    t.context.expect(sixTimesNine).to.equal(42);
});
```

mmkal:
Here's a [gist of the generated types file](https://gist.github.com/mmkal/6d99027728d8088174fde481e0b84a36), since it can't be seen on the PR.
@novemberborn @despairblue @ivogabe I'd love to get this merged as I'm planning to use it in my current project. Sorry about so many small commits - I'd be happy for it to go in as it is now, if it looks OK to you.

types/make.js
```diff
@@ -24,9 +24,25 @@ const base = fs.readFileSync(path.join(__dirname, 'base.d.ts'), 'utf8');
...
+	testFunctionDeclarations(),
+	generatePrefixed([])
+].join('\n');
+
 fs.writeFileSync(path.join(__dirname, 'generated.d.ts'), output);
 
+// Returns the declarations for the 'test' function and core types
+function testFunctionDeclarations() {
+	return (
+`export default test;
```
ivogabe: Move this to `base.d.ts`

types/base.d.ts
```diff
@@ -9,11 +7,16 @@ export type ErrorValidator
...
 export type CallbackTest = (t: CallbackTestContext) => void;
-export type ContextualCallbackTest = (t: ContextualCallbackTestContext) => void;
+export type GenericCallbackTest<T> = (t: GenericCallbackTestContext<T>) => void;
+
+interface Context<T> { context: T }
```
ivogabe: Add `export`

types/base.d.ts
```diff
@@ -99,20 +102,19 @@ export interface CallbackTestContext extends TestContext {
...
 	(t: T, ...args: any[]): void;
 	title? (providedTitle: string, ...args: any[]): string;
 }
 export type Macros<T> = Macro<T> | Macro<T>[];
 
-export function test(name: string, run: ContextualTest): void;
-export function test(run: ContextualTest): void;
-export function test(name: string, run: Macros<ContextualTestContext>, ...args: any[]): void;
-export function test(run: Macros<ContextualTestContext>, ...args: any[]): void;
+interface TestFunctionCore<T> {
```
ivogabe: Can we think of a better name here? Technically this is not the function that does the tests, it only registers a test. So maybe something like `RegisterBase<T>`, and rename `TestFunction` `Register<T>`?

And add `export` here

types/make.js
mmkal: If I do, `base.d.ts` will no longer compile because `ContextualTestFunction` and `TestFunction` don't exist in base.d.ts. 

Or do you think that doesn't matter so much because `base.d.ts` isn't used for anything except seeding `generated.d.ts`?

ivogabe: That's not an issue, `base.d.ts` is indeed only used by the generation script to write `generated.d.ts`.

mmkal:
@ivogabe updated with the `export`s and renames you suggested. [Gist](https://gist.github.com/mmkal/6d99027728d8088174fde481e0b84a36) also updated.

sindresorhus:
This should be documented somewhere, probably here: https://github.com/avajs/ava/blob/master/docs/recipes/typescript.md

mmkal:
@sindresorhus added the above usage example.

ivogabe:
Looks good to me, I've tested it on one of my projects and it worked just fine.

This does however not work with TypeScript 2.0, 2.1 is required (because of the use of `T['skip']`). 2.1 was released december last year. But that's probably not an issue since AVA is not 1.0 yet.

readme.md
```diff
@@ -578,6 +578,8 @@ Keep in mind that the `beforeEach` and `afterEach` hooks run just before and aft
 
 Remember that AVA runs each test file in its own process. You may not have to clean up global state in a `after`-hook since that's only called right before the process exits.
 
+### Test context
```
sindresorhus: `###` => `####`

sindresorhus:
Thank you @mmkal. This is a great improvement! :)

mmkal:
Thanks @sindresorhus @novemberborn @ivogabe ! Does it get auto-published to NPM? Or is it done ad-hoc?

sindresorhus:
@mmkal No, releases are manually published. Was planning to do a new one next week (It's a lot of work to manually test and write release notes).

PR #1301, opened 2017-03-08
Improve YAML blocks in TAP reporter

novemberborn:
* Properly print multiline strings (assuming \n is the line separator)
* Print additional name & message properties
* Determine 'at' value similar to serialize-error.js. Recompute since here we want the full line
* Print unhandled errors using the same logic
* Strip ANSI from actual and expected values

### DIFF
--- a/lib/reporters/tap.js
+++ b/lib/reporters/tap.js
@@ -1,14 +1,37 @@
 'use strict';
 const format = require('util').format;
+const indentString = require('indent-string');
 const stripAnsi = require('strip-ansi');
+const yaml = require('js-yaml');
+const extractStack = require('../extract-stack');
 
 // Parses stack trace and extracts original function name, file name and line
-function getSourceFromStack(stack, index) {
-	return stack
-		.split('\n')
-		.slice(index, index + 1)
-		.join('')
-		.replace(/^\s+ /, '');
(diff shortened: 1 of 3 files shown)


lib/reporters/tap.js
```diff
@@ -1,14 +1,44 @@
...
+}
+
+function yamlBlock(error, includeMessage) {
+	const output = [];
+
+	if (error.name) {
+		output.push(`    name: ${yamlBlockPropertyValue(error.name)}`);
+	}
+	if (includeMessage && error.message) {
+		output.push(`    message: ${yamlBlockPropertyValue(error.message)}`);
```
sindresorhus: I think we need to handle quotes.

For example:

```
message: 'hello' this is not valid YAML
```

Wouldn't it be better to just use a real YAML stringifier? Like [`js-yaml`](https://github.com/nodeca/js-yaml)

novemberborn: Yes, not sure why I thought I could get away with the manual approach 😉

novemberborn:
Updated.

sindresorhus:
Looks good :)

I wish TAP had some kind of testsuite we could run our implementation at to check for conformance.

PR #1302, opened 2017-03-09
Refactor assertions

### LINKED ISSUE #1150: t.notThrows() gives no info on thrown error in case an error is thrown
## Description

I have yet to make a test which uses `t.notThrows()` pass. Is it by design to not get any info on the thrown error when `t.notThrows()` actually errors out? All I get is:

`Got unwanted exception..`

I would like to see the stack trace.

Currently, I'm expecting the code to fail with a ` SequelizeUniqueConstraintError: Validation error` error. But since I'm not seeing anything other than `Got unwanted exception..`, I cannot verify this. Later on, after fixing this, if the test fails again, I would like to see the stack trace to see if it's still because of the constraint violation.

### Test Source

```js
  let [transaction, onSuccess] = await adapter.setupTransaction();
  // fails with "Got unwanted exception..". 
  await t.notThrows(Promise.all(G.deltas.map(delta => {
    return adapter.addContentObjectDelta(G.co1.getContentObjectId(), delta, {}, { transaction });
  })));
  await onSuccess();
```

```
AssertionError: Got unwanted exception..
```

### Config

```json
{
  "ava": {
    "require": [
      "babel-register",
      "babel-polyfill"
    ],
    "babel": "inherit",
    "files": [
      "test/**/*.test.js"
    ],
    "failFast": true
  },
}
```

### Command-Line Arguments

```
ava --serial
```

`--verbose` does not help.

## Environment

```
Node.js v6.3.1
darwin 15.4.0
ava: 0.17.0
npm: 3.10.3
```

### LINKED ISSUE #1227: t.notThrows(promise) should return an "empty" promise
This is as expected:

```js
const value = Symbol()
t.notThrows(() => value) === undefined
```

However:

```js
const value = Symbol()
(await t.notThrows(Promise.resolve(value))) === value
```

`t.notThrows(promise)` should return a promise, but for an `undefined` value.

The offending line is at https://github.com/avajs/ava/blob/dcdfbee689e1538420f5793bbaca09c7e78ddedd/lib/assert.js#L124.

The Flow and TypeScript definitions should be updated too:

* https://github.com/avajs/ava/blob/dcdfbee689e1538420f5793bbaca09c7e78ddedd/index.js.flow#L63
* https://github.com/avajs/ava/blob/dcdfbee689e1538420f5793bbaca09c7e78ddedd/types/base.d.ts#L68

novemberborn:
See commits. Behavior should be the same, aside from these changes:

* Fixes #1227 
* Fixes #1150 
* Assertions now return `undefined` rather than `null`
* Removes the undocumented `t.assertCount` getter
* All test failures (exception thrown, rejected promise returned, `t.end(error)`) are now wrapped in an assertion error, but we'll still print the actual error
* Do the same for when `t.throws()` / `t.notThrows()` are called with an improper argument. We can now even print the actual argument passed!
* In the TAP reporter, the error YAML block now has varying properties (e.g. do not assume it has `operator` or even `assertion`)
* Test serialization no longer overwrites non-standard error properties

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -1,158 +1,327 @@
 'use strict';
-const assert = require('core-assert');
+const coreAssert = require('core-assert');
 const deepEqual = require('lodash.isequal');
 const observableToPromise = require('observable-to-promise');
 const indentString = require('indent-string');
 const isObservable = require('is-observable');
 const isPromise = require('is-promise');
 const jestSnapshot = require('jest-snapshot');
+const enhanceAssert = require('./enhance-assert');
 const snapshotState = require('./snapshot-state');
 
-const x = module.exports;
-const noop = () => {};
+class AssertionError extends Error {
+	constructor(opts) {
+		super(opts.message || '');
+		this.name = 'AssertionError';
 
-Object.defineProperty(x, 'AssertionError', {value: assert.AssertionError});
+		this.actual = opts.actual;
+		this.assertion = opts.assertion;
+		this.expected = opts.expected;
+		this.hasActual = 'actual' in opts;
+		this.hasExpected = 'expected' in opts;
+		this.operator = opts.operator;
 
-function create(val, expected, operator, msg, fn) {
-	return {
-		actual: val,
-		expected,
-		message: msg || ' ',
-		operator,
-		stackStartFunction: fn
-	};
-}
+		// Reserved for power-assert statements
+		this.statements = null;
 
-function test(ok, opts) {
-	if (!ok) {
-		const err = new assert.AssertionError(opts);
-		err.showOutput = ['fail', 'throws', 'notThrows'].indexOf(err.operator) === -1;
-		throw err;
-	}
-}
-
-x.pass = msg => {
-	test(true, create(true, true, 'pass', msg, x.pass));
-};
-
-x.fail = msg => {
-	msg = msg || 'Test failed via t.fail()';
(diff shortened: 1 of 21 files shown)


novemberborn:
Tests are red across the board, but it works on my machine…

I'll look into that tomorrow. Please hit me with feedback regardless 😄

novemberborn:
> Tests are red across the board, but it works on my machine…
> 
> I'll look into that tomorrow.

Clearing CI caches helped. Pushed more changes too.

lib/assert.js
```diff
@@ -122,7 +122,11 @@ function wrapAssertions(callbacks) {
 			} else if (isObservable(fn)) {
 				promise = observableToPromise(fn);
 			} else if (typeof fn !== 'function') {
-				throw new TypeError('t.throws must be called with a function, Promise, or Observable');
+				fail(this, new AssertionError({
+					actual: fn,
+					message: '`t.throws()` must be called with a function, Promise, or Observable'
+				}));
```
novemberborn: `--fail-fast` implies all tests stop. This still immediately fails the test. The behavior should be similar to before this refactor, except that the error can no longer be caught by calling code.

sindresorhus: I just tried and you're right. It is the existing behavior. I still think user mistakes like this should fail the test right away no matter what flags the user used though, but that's a separate discussion.

sindresorhus:
🎉🎉🎉

PR #1310, opened 2017-03-15
Fix `nodeExternals` in the precompiling-with-webpack recipe

screendriver:
Fixed `nodeExternals` in `webpack.config.js` because the provided example did not work. `externals` have to be an array and `nodeExternals` is a function that must be called and not referenced.


sindresorhus:
Don't change the indentation.

screendriver:
Ok. May I ask why? The original indentation was tabs and not spaces. Is it preferred to use tabs?

screendriver:
By the way: fixed.

sindresorhus:
> Is it preferred to use tabs?

Yes, that's what we use.

screendriver:
Oki doki. I fixed that ☺️

Issue #1226, opened 2017-02-04
Include only plain objects/arrays in t.true/t.false output of statements


vadimdemedes:
AVA's magic-assert output causes a crash when this kind of assertion fails:

```js
t.true(some.spy.calledOnce);
```

or similar, when functions/instances are involved. This is happening, because `@ava/pretty-print` tries to print `some.spy` and `some` nicely. When parsing statements in `lib/enhance-assert`, we should ignore anything but plain objects and arrays.

novemberborn:
Could you elaborate on what specifically causes the crash? 

> When parsing statements in lib/enhance-assert, we should ignore anything but plain objects and arrays.

How does this affect the output for `t.true(some.spy.calledOnce)`?

vadimdemedes:
Sure. You can easily cause a crash if one of the statements contain (or is) a reference to a global scope (`global`). There's a ton of stuff in there and it doesn't make sense to print that.

Same goes with all the libraries. Unless it's a plain object or array, we can't pretty print it. Even if we could, why would we? The output needs to be meaningful and compact. So in the above example, `t.true(some.spy.calledOnce)` it's enough to print that `some.spy.calledOnce` is `false`.

novemberborn:
> Same goes with all the libraries. Unless it's a plain object or array, we can't pretty print it. Even if we could, why would we? The output needs to be meaningful and compact. So in the above example, t.true(some.spy.calledOnce) it's enough to print that some.spy.calledOnce is false.

Fair enough.

I'd love to be able to print `Map` and `Set` though. And for those values we can't print, indicate their type.

vadimdemedes:
Sure, `Map` and `Set` are ok. Talking about instances of classes/libraries. For example, you wouldn't want to pretty print an instance of Mongoose model, right?

novemberborn:
> For example, you wouldn't want to pretty print an instance of Mongoose model, right?

It can be useful sometimes, but probably no 😄  Perhaps those kinds of instances could be restricted to showing the first three properties, with a little note saying "59 more properties not shown".

vadimdemedes:
I think implementing this might be overly complicated and could take a while. This bug is quite annoying actually, I've seen it at least 6 times during last 2 days (yes, I've set up a counter 😄). I'd just go for a set of accepted types to be pretty-formatted in the "statements" output and push a fix. We can always make it more complex later ;)

novemberborn:
[...]

Issue #1205, opened 2017-01-23
Show relevant parts of the object/array for t.true/t.false assertion output


vadimdemedes:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%2460%20Funded-%2300A156.svg" />](https://issuehunt.io/r/avajs/ava/issues/1205)

**Note, this issue is related only to #1154.**

Given this assertion:

```js
t.true(arr[56]);
```

if the size of `arr` is (for example) `100`, AVA will output the whole array without pointing to the specific item that caused an assertion error. We should show only the relevant parts of the array/object to identify the cause of an issue quicker. Plus, we really don't want to trash user's terminal.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Sponsors (Total: $60.00)

- [<img src='https://avatars3.githubusercontent.com/u/44827199?v=4' alt='issuehunt' width=24 height=24> issuehunt](https://issuehunt.io/u/issuehunt) ($60.00)

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/avajs/ava/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

yatharthk:
@vadimdemedes thanks for pointing it out. I actually faced this thing last night and was thinking to open an issue.

novemberborn:
My suggestion in #1204 is to show the type of `arr` but nothing else.

leandrooriente:
Guys, do you have any update regarding this request?

I'm facing the same issue.
If I use `t.is(resetBehaviour.calledOnce, false);` it outputs this:
```bash
51:     wrapper.unmount();
   52:     t.is(resetBehaviour.calledOnce, false);
   53: });

  Difference:

  - true
  + false

  Test.is (src/common/components/App/App.test.js:52:7)
  processEmit [as emit] (node_modules/nyc/node_modules/signal-exit/index.js:155:32)
  processEmit [as emit] (node_modules/nyc/node_modules/signal-exit/index.js:155:32)
```

But if i use `t.false(resetBehaviour.calledOnce)` it returns this huge output: 

```bash
51:     wrapper.unmount();
   52:     t.false(resetBehaviour.calledOnce);
   53: });

  Value is not `false`:

  true

  resetBehaviour.calledOnce
  => true

  resetBehaviour
  => Function proxy {
    args: [
      [],
    ],
[...]

Issue #1204, opened 2017-01-23
Limit the number of statements in the output for t.true/t.false


vadimdemedes:
**Note, this issue is related only to #1154.**

Given this assertion:

```js
const obj = {
  a: {
    b: {
      c: [false]
    }
  }
};

t.true(obj.a.b.c[0]);
```

AVA will output the value of each statement (`obj.a.b.c[0]`, `obj.a.b.c`, `obj.a.b`, `obj.a`, `obj`), which of course is not optimal, since the helpful values are probably only the last two. We should limit the number of statements to prevent a long list of unhelpful values.

novemberborn:
https://github.com/avajs/ava/issues/1222#issuecomment-280671306 is another example of this. I can reproduce that using:

```
import test from 'ava'
import {jsdom} from 'jsdom'

test(t => {
  const doc = jsdom('<html></html>')
  t.true(doc.documentElement.classList.contains('foo'))
})
```

With AVA 0.18.1 we dump out each [`property` in that `MemberExpression`](http://astexplorer.net/#/gist/846391c23915214505ae12eb9eff7240/464b4b4eeda117c73259ff9f866914d9b0fb92b4). With AVA 0.17.0 we only show:

```
  t.true(doc.documentElement.classList.contains('foo'))
                                       |
                                       false
```

Why did we remove the [old formatter code](https://github.com/avajs/ava/pull/1154/files#diff-b19474e8e1de352b59bce4b9e0266d47L49)? To get consistency in how we were displaying objects etc?

Would it be an appropriate short-term fix to just show the [last `property`](http://astexplorer.net/#/gist/2e381ccd5b753aa4618226b41d931e90/8677539983fecfa4bd53e6479b77937f3bcd0063), and hope it's not too massive?

Longer term it might still be useful to show each `property` in the `MemberExpression`, but perhaps just their type? I don't think we can do that with `pretty-format` though.

vadimdemedes:
> Why did we remove the old formatter code? To get consistency in how we were displaying objects etc?

Yeah, for consistency. Otherwise it'd be pretty weird to see completely different looking output for different assertions.

> Would it be an appropriate short-term fix to just show the last property, and hope it's not too massive?

Totally agree. Perhaps we could even remove "statements" output for t.true/t.false until it's more production ready and just show actual/expected. What do you think?

> Longer term it might still be useful to show each property in the MemberExpression, but perhaps just their type? I don't think we can do that with pretty-format though.

[...]

PR #1312, opened 2017-03-15
More assertion changes

novemberborn:
* Cleanup and consistency changes
* Rather than printing `Actual`/`Expected`, vary those labels by the assertion, depending on whether the assertion was improperly used or failed. This creates more helpful output
* For power-assert enhanced assertions, print the assertion failure first, followed by the statement context. Limit the depth of the formatted context to just 1. Addresses #1226, #1205 and #1204. (One last improvement would be to limit the number of properties / array items shown, but we can't do that with `pretty-format`.)
* Detect improper usage of `t.regex()` / `t.notRegex()`

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -300,6 +300,21 @@ function wrapAssertions(callbacks) {
 		},
 
 		regex(string, regex, message) {
+			if (typeof string !== 'string') {
+				throw new AssertionError({
+					assertion: 'regex',
+					message: '`t.regex()` must be called with a string',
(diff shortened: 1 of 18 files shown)


vadimdemedes:
Damn it, Mark is killing it here!

> Rather than printing Actual/Expected, vary those labels by the assertion, depending on whether the assertion was improperly used or failed. This creates more helpful output

Big fan of this change, removes the need to guess what output means.

sotojuan:
LGTM as well. Great!

sindresorhus:
🙌 Great as always.

PR #1313, opened 2017-03-19
Include anonymous functions (whose name node cannot infer) in stack traces

neoeno:
### Issue

Functions whose name node cannot infer do not get included in stack traces.

I came across this issue when working with curried functions.

### Reproduction 

This error in a code using an anonymous function:

```js
const fn = () => () => doesnotexist
fn()()
```

Produces a stacktrace line without a function name:

```
ReferenceError: doesnotexist is not defined
    at /Users/caden/Code/libs/ava/repro.js:1:86
    at Object.<anonymous> (/Users/caden/Code/libs/ava/repro.js:2:5)
    at Module._compile (module.js:556:32)
    at Object.Module._extensions..js (module.js:565:10)
    at Module.load (module.js:473:32)
    at tryModuleLoad (module.js:432:12)
    at Function.Module._load (module.js:424:3)
    at Module.runMain (module.js:590:10)
```

The regex in [extract-stack.js:2](https://github.com/avajs/ava/blob/master/lib/extract-stack.js#L2) expects the function name & brackets to be there:

```js
const stackLineRegex = /^.+ \(.+:[0-9]+:[0-9]+\)$/;
```

So that line gets swallowed. Which means we lose context for our error message. This is a shame.

### Fix

This commit fixes the issue by adding another regex to `extract-stack.js` which will match anonymous functions. 

The regex may be a little broad — I couldn't think of any obvious instances where it would fail but undoubtedly some contrived examples exist.

I have included a unit test and an integration test. A few different areas in the codebase amend stacktraces, so I thought it was best to do so.

### Finally

Thanks for all your great work on Ava!

### DIFF
--- a/lib/extract-stack.js
+++ b/lib/extract-stack.js
@@ -1,5 +1,5 @@
 'use strict';
-const stackLineRegex = /^.+ \(.+:[0-9]+:[0-9]+\)$/;
+const stackLineRegex = /^.+( \(.+:[0-9]+:[0-9]+\)|:[0-9]+:[0-9]+)$/;
 
 module.exports = stack => {
 	return stack
--- a/test/cli.js
+++ b/test/cli.js
@@ -105,6 +105,14 @@ test('throwing a named function will report the to the console', t => {
 	});
 });
 
+test('include anonymous functions in error reports', t => {
+	execCli('fixture/error-in-anonymous-function.js', (err, stdout, stderr) => {
+		t.ok(err);
+		t.match(stderr, /test\/fixture\/error-in-anonymous-function\.js:4:8/);
+		t.end();
+	});
+});
+
 test('improper use of t.throws will be reported to the console', t => {
 	execCli('fixture/improper-t-throws.js', (err, stdout, stderr) => {
 		t.ok(err);
--- a/test/extract-stack.js
+++ b/test/extract-stack.js
@@ -34,3 +34,13 @@ test('strip beginning whitespace from stack', t => {
 	t.is(extractStack(stack), 'Test.t (test.js:1:1)');
 	t.end();
 });
+
+test('includes anonymous function lines', t => {
+	const stack = [
+		'error message',
+		'path/to/test.js:1:1'
+	].join('\n');
+
+	t.is(extractStack(stack), 'path/to/test.js:1:1');
+	t.end();
+});
--- a/test/fixture/error-in-anonymous-function.js
+++ b/test/fixture/error-in-anonymous-function.js
@@ -1,0 +1,9 @@
+import test from '../../';
+
+const getAnonymousFn = () => () => {
+	throw Error();
+};
+
+test(t => {
+	getAnonymousFn()();
+});


novemberborn:
@neoeno could you share an example of how this changes AVA's output, and what is missing in the current output? The stack trace you shared comes from Node.js itself but I can't quite see how it impacts AVA's output.

neoeno:
Of course @novemberborn — sorry I didn't do this already!

Here's a repo with reproduction, output both before and after this PR: https://github.com/neoeno/ava-pr-1313-example

For posterity, here's the code:

```js

import test from 'ava';

const fn = () => () => doesnotexist;

test("some test", t => {
  fn()();
});
```

Here's the current output (as of 3da8603):

```
$ ava test.js

  1 failed

  some test
  /Users/kay/Code/libs/tester/test.js:6

   5: test("some test", t => {
   6:   fn()();
   7: });

  Error:

    [ReferenceError: doesnotexist is not defined]

  Error thrown in test
```

And here's the output after this PR:

```
$ ava test.js

  1 failed

  some test
  /Users/kay/Code/libs/tester/test.js:3

   2:
   3: const fn = () => () => doesnotexist;
   4:

  Error:

    [ReferenceError: doesnotexist is not defined]

  Error thrown in test

  test.js:3:24
  Test.t [as fn] (test.js:6:3)
```

neoeno:
I'll fix that formatting error too 😨

novemberborn:
@vadimdemedes how did you arrive at the regular expression? Looks like `extract-stack` was introduced in #1154.

lib/extract-stack.js
```diff
@@ -1,10 +1,11 @@
 'use strict';
-const stackLineRegex = /^.+ \(.+:[0-9]+:[0-9]+\)$/;
+const namedFunctionStackLineRegex = /^.+ \(.+:[0-9]+:[0-9]+\)$/;
+const anonymousFunctionStackLineRegex = /^.+:[0-9]+:[0-9]+$/;
 
 module.exports = stack => {
 	return stack
 		.split('\n')
-		.filter(line => stackLineRegex.test(line))
+		.filter(line => namedFunctionStackLineRegex.test(line) || anonymousFunctionStackLineRegex.test(line))
```
novemberborn: I think you should combine this into a singular regular expression.

neoeno: As you prefer!

vadimdemedes:
@novemberborn I think I tried to make a regex that'd be as simple as possible and match "standard" stack lines. The point of that was to get only stack from err.stack.

sindresorhus:
I think the regex could still be looser. I decided to extract it into a separate module as it can be useful for other projects too: https://github.com/sindresorhus/extract-stack

@neoeno Can you update to use that module and remove `lib/extract-stack.js` and its tests?

neoeno:
I'll give it a go — though afaict it won't be as simple as replacing `extract-stack.js` with the new module (which is great by the way!).

This is because at the moment `extractStack` accepts a pre-processed stack-trace, of of this form:

```
error message
  Test.t (test.js:1:1)
```

No `at`s — which from reading the code I think the `extract-stack` module relies upon?

Here's where my knowledge gets a bit hazy, but it looks like it passes through `clean-stack` elsewhere — removing the `at`s as well as some other stuff — via paths like this: [main.js:37](https://github.com/avajs/ava/blob/master/lib/main.js#L37) -> [serialize-error.js:52](https://github.com/avajs/ava/blob/master/lib/serialize-error.js#L52) -> [beautify-stack:27](https://github.com/avajs/ava/blob/master/lib/beautify-stack.js#L27) -> clean-stack. There may be one or two more...

So the solution would seem to be to do all this stacky work in one place and delegate to `extract-stack` its due share — but I've no idea how much work that will be. Might be simple anyway, so I'll give it a shot and get back to you.

novemberborn:
My hunch is that we should extract the stack when serializing (filter out unnecessary bits early), and beautify it in the mini and verbose reporters. Though I'm not even sure what `beautify-stack` does that `extract-stack` doesn't.

bchapman:
Hit the same issue.  I'm using an async wrapper function inside of a test, and any code inside loses it's context in the traceback.  This would be super helpful to get the error context back.

patrickrand:
+1 for this

novemberborn:
@neoeno will you have a chance to get back to this? If not, perhaps @bchapman, @zabawaba99 or @patrickrand want to give it a go? 😄

neoeno:
Though this is still on my radar I suspect I won't get to it imminently. As such, if anyone is keen to take it on I should certainly give way!

zabawaba99:
I don't think I'll be able to get to this immediately either but I will post if I'm able to start looking into it!

Issue #1094, opened 2016-10-26
Custom assertions


yamsellem:
Is there a way to add some custom assertions to `t` existing ones?

And maybe, adding some basic needs to the existing ones:
— `t.instanceof(object, clazz)`
— `t.length(array, length)`
— `t.members(array, item, item)`
— `t.includes(string, substring)`

For custom ones, I guess `t.context` can be used, but feels cumbersome.

sindresorhus:
Not at the moment, but your needs can be solved with the existing methods:

```js
t.true(object instanceof clazz);
t.is(array.length, length);
t.true(array.every(x => [item, item].includes(x)));
t.true(string.includes(substring));
```

You can already use any external assertion library. The only caveat is that `t.plan()` doesn't work.

alathon:
I think a very strong point of AVA (and tape) is precisely that there are few assertions by default - this reduces the amount of 'magic' that occurs, and reduces bugs in testing code that occurs because you aren't 100% sure how a particular assertion works.

yamsellem:
@alathon Being small does not necessarily means being ascetic.

```javascript
t.is(array.length, length); // is better with t.truthy(array) before
t.true(array.every(x => [item, item].includes(x))); // seems cumbersome
t.true(string.includes(substring)); // could be simpler
```

Some basic assertion on strings and arrays seems fair to me — and don't break the 'no magic' of the thing, IMO.

sotojuan:
I guess the problem is that it's a slippery slope. These kind of extra assertions are better in userland as separate modules, functions, or snippets. I personally don't see anything wrong with the code block in the last comment but you could make it look "prettier" by abstracting those functions away. That way you can use your own custom logic as well.

The smaller core AVA is, the easier it is to use.

vdemedes:
@sotojuan Good point, totally agree.

I think we should make `t` extendable with custom methods (assertions). Was thinking to create a separate module for `t.jsxEqual()`, would be nice to make this possible.

odigity:
I'm in the same boat.  I would like to write some custom assertions for convenience and add them to `t` in my project.

The [docs](https://github.com/avajs/ava#custom-assertions) say "you can use custom assertions", but provide no information on *how*.

vadimdemedes:
[...]

Issue #1047, opened 2016-09-23
Simplify t.throws()


novemberborn:
This is my proposal for resolving #661. In short I want to stop deferring to https://nodejs.org/api/assert.html#assert_assert_throws_block_error_message and remove various parameter overloads that are currently supported by the assertion.

Here's what I'd like to support:

``` js
t.throws(fn) // Throws an Error (or subclass thereof)
t.throws(fn, SyntaxError) // Throws a SyntaxError (or subclass thereof)
t.throws(fn, 'string') // Throws an Error (or subclass thereof), whose .message === 'string'
t.throws(fn, /regexp/) // Throws an Error (or subclass thereof), whose /regexp/.test(err.message) === true
```

If you need to test the errors class and message you should use the `t.throws` return value:

``` js
const err = t.throws(fn, SyntaxError)
t.true(err.message === 'expecting integer')
```

Passing anything other than `undefined`, functions, strings or regular expressions as the second argument results in an `TypeError` from `t.throws()` itself.

The first argument is allowed to be a promise or observable, as is any return value from the `fn` call. Of course this makes `t.throws` asynchronous, so users may need to do:

``` js
const err = await t.throws(fn)
```

---

Questions:
1. Does `t.throws(fn, Constructor)` require the thrown value to be an `Error`? For example:
   
   ``` js
   t.throws(() => { throw 'foo' }, String) // Is this allowed?
   ```
2. Should `t.throws(fn, 'string')` and `t.throws(fn, /regexp/)` be supported at all, or would it be preferable to dereference the error and then use another assertion?
   
   ``` js
   const err = t.throws(fn)
   t.true(err.message === 'expecting integer')
   ```
3. Is there a need for `t.throws(fn, SyntaxError, 'string')` and `t.throws(fn, SyntaxError, /regexp/)`
4. Does anybody have experience with / examples of asserting `Error` instances across contexts?

vdemedes:
I absolutely support your proposal!

Answers to your questions:

1. I don't think it should be allowed. ESLint's docs on `no-throw-literal` rule explain it best:

> It is considered good practice to only throw the Error object itself or an object using the Error object as base objects for user-defined exceptions. The fundamental benefit of Error objects is that they automatically keep track of where they were built and originated.

Services like BugSnag, Sentry, Rollbar, etc will provide zero information, if code throws strings, instead of Error objects.

2. I think they should be supported, with behavior exactly like you outlined in your first code example:

```js
[...]

Issue #261, opened 2015-11-23
verbose mode


jamestalmage:
With #259, we have moved from only printing the last assertion failure, to only printing the first.

Once #243 (chainable methods) happens, I think we should allow printing of every failed assertion to give the most information possible.

```js
test.verbose('title', t => {
  t.fail('a');
  t.fail('b');
});
```

Perhaps `only` mode should also just imply `verbose` mode (you should have way less noise with a single test).

sindresorhus:
Not sure we should have a `verbose` modifier. I don't see why would want to augment just some tests. I'd rather have a `--verbose` flag for all tests.

> Perhaps only mode should also just imply verbose mode (you should have way less noise with a single test).

I like the idea of `only` triggering verbose mode for those tests.

Qix-:
`ava -v` would be wonderful.

vdemedes:
As for me, I can't find the use-case for a verbose mode in my mind. If you could help me with that, that'd be great :) If I have 10 assertions with 1st assertion causing the rest to fail, why do we need to see the other 9? Are there any examples in some projects or other test runners, where this functionality is needed/implemented?

jamestalmage:
`tap`/`tape` show you every assertion failure.

Usually, yes - you only need to see the first assertion. Though sometimes additional information can be helpful:

```js
var a = sum(2, 2); // => 4 (correct)
var b = sum(4, 4); // => 7 (wrong - for some weird reason)
var c = sum(a, b); // => 11 (correct for the wrong input).

t.is(c, 12); // this will fail (telling us there is something wrong - but not what the root of the problem is)
t.is(b, 8);  // this will also fail and in this case is the more helpful information
```

Granted, the solution above is pretty obvious - reorder the assertions. But that is not so easy to see in every problem.

jamestalmage:
Here is an example where I wish I could have seen every failure: 
  https://github.com/jamestalmage/babel-plugin-detective/blob/master/test/babel-6-test.js#L165

I could not use `deepEquals`, and using `assert` which throws at the first error was annoying because you can not know ahead of time which failure will highlight your mistake.

vdemedes:
> tap/tape shows you every assertion failure.

This shouldn't really bother us, since AVA is not trying to replicate tap/tape, but trying to stand out.

> Though sometimes additional information can be helpful

In that case, verbose mode should definitely be optional and dead-clear to enable. I'm not a fan of enabling verbose mode automatically for `test.only()`.

[...]

PR #1314, opened 2017-03-20
Overhaul test implementation 👷🏗

novemberborn:
Another massive PR… the first bunch of commits are cleanups and refactorings that lay groundwork for the subsequent work.

Highlights:

* `t.throws()` and `t.notThrows()`, when passed a promise or observable, no longer reject the returned promise with the `AsssertionError`. Instead the promise is fulfilled with `undefined`. This behavior is consistent with when a function is passed, and with other assertions which also do not throw when they fail.

* `Concurrent`, `Sequence` and `Test` have been refactored and are (hopefully) easier to understand.

* Similarly `test-worker` and `process-adapter` are slightly easier to understand.

* It's now a failure to add an assertion after a test finishes. That said, the failure is only shown if you used `t.throws()` or `t.notThrows()` assertions with a promise or observable. I'll open an issue when this PR lands discussing how to make the failure count in other scenarios.

* It's now a failure for a test to finish without executing *any* assertions. A `failWithoutAssertions` option has been added to the `package.json` configuration. It defaults to `true`, users should set this to `false` to disable the behavior. This will break deployments with where custom assertions are used, or tests that don't use AVA's assertion library for other reasons. Use `t.pass()` to ensure at least 1 assertion is executed.

* We can now detect when `t.end()` is never called in a `test.cb()` test, or if a returned promise never resolves, or a returned observable never completes. This means tests will no longer hang. This does not work if user code keeps the event loop busy by starting or connecting to a server or running long timers.

This also prepares for integrating https://github.com/avajs/babel-plugin-ava-throws-helper/pull/8.


novemberborn:
> t.throws() and t.notThrows(), when passed a promise or observable, no longer reject the returned promise with the AsssertionError. Instead the promise is fulfilled with undefined. This behavior is consistent with when a function is passed, and with other assertions which also do not throw when they fail.

Need to update the type definitions to take this into account.

novemberborn:
>> t.throws() and t.notThrows(), when passed a promise or observable, no longer reject the returned promise with the AsssertionError. Instead the promise is fulfilled with undefined. This behavior is consistent with when a function is passed, and with other assertions which also do not throw when they fail.
>
> Need to update the type definitions to take this into account.

That said, it would make this example quite annoying:

```ts
const err = await t.throws(promise)
t.is(err.message, 'foo')
```

Because if `err` can be undefined, this would have to be written as:

```ts
const err = await t.throws(promise)
if (err) {
  t.is(err.message, 'foo')
}
```

And yes the code will crash without that guard, but that error is ignored given the original assertion failure in `t.throws()`. How accurate do we want to be?

sindresorhus:
> t.throws() and t.notThrows(), when passed a promise or observable, no longer reject the returned promise with the AsssertionError. Instead the promise is fulfilled with undefined. This behavior is consistent with when a function is passed, and with other assertions which also do not throw when they fail.

I don't get it.

```js
test(async t => {
	const err = await t.throws(Promise.reject(new Error('foo')));
	console.log(err.message);
});
```

The `console.log` here will never be reached as `t.throws` throws:

```js
test(async t => {
	const err = await t.throws(Promise.resolve(new Error('foo')));
	console.log(err.message);
});
```

What will change?

novemberborn:
> What will change?

It'll no longer throw, just like `t.throws(() => {})` doesn't throw, and `t.true(false)` doesn't throw.

sindresorhus:
I see. I use the pattern of assigning `t.throws` to `err` a lot, so having to guard `err` would be inconvenient. Maybe we should hold on off this until we've done https://github.com/avajs/ava/issues/1047, so users doesn't have to change their code twice?

lib/test-collection.js
```diff
@@ -124,14 +125,14 @@ class TestCollection extends EventEmitter {
 			context = null;
 		}
 
-		return new Test(hook.metadata, title, hook.fn, context, this._emitTestResult);
+		return new Test(hook.metadata, title, hook.fn, false, context, this._emitTestResult);
```
sindresorhus: Time to use an object for the arguments?

readme.md
```diff
@@ -673,6 +676,8 @@ You can use any assertion library instead of or in addition to the built-in one,
 
 This won't give you as nice an experience as you'd get with the [built-in assertions](#assertions) though, and you won't be able to use the [assertion planning](#assertion-planning) ([see #25](https://github.com/avajs/ava/issues/25)).
 
+You'll have to configure AVA to not fail tests if no assertions are executed, because AVA can't tell if custom assertions pass. Set the `failWithoutAssertions` option to `false` in AVA's [`package.json` configuration](#configuration).
```
sindresorhus: This feature, while useful, is going to hit a lot of people. Any chance we could add support for assertions throwing `AssertionError`?

I know we have https://github.com/avajs/ava/issues/1094, but that's more about integrating with `t` and `t.plan()`. I sometimes use assertions helper, and would like to use them in AVA without having to do anything extra.

lib/test.js
```diff
@@ -262,23 +259,36 @@ class Test {
...
-						values: [formatAssertError.formatWithLabel('Rejection reason:', err)]
-					}));
-					return this.finish();
-				}
-			);
+			return new Promise(resolve => {
+				this.finishDueToInactivity = () => {
+					const err = returnedObservable ?
+						new Error('Observable returned by test never completed') :
+						new Error('Promise returned by test never resolved');
```
sindresorhus: ❤️ This commit is an amazing improvement! I remember trying to achieve something like this in AVA 0.0.2, but I could not get it working. Probably because `beforeExit` didn't exist then.

sindresorhus:
> Concurrent, Sequence and Test have been refactored and are (hopefully) easier to understand.

Significantly more readable, indeed.

sindresorhus:
Mark, you're on 🔥!

test/fixture/stalled-tests/callback.js
```diff
@@ -0,0 +1,6 @@
+'use strict';
+
+const test = require('../../..');
```
novemberborn: Yea I can clean that up. I probably just copied these from elsewhere.

lib/sequence.js
```diff
@@ -1,86 +1,59 @@
...
-			// If last item failed and we should bail, return results and stop
-			if (this.bail && !this.passed) {
-				return this._results();
-			}
+		let allPassed = true;
+		const runNext = () => {
+			let promise;
 
-			const result = this.tests[i].run();
+			for (let next = iterator.next(); !next.done; next = iterator.next()) {
```
novemberborn: Because of promises.  The alternative is maintaining an `index` variable to resume the iteration. That's what iterators do all by themselves though.

lib/test-collection.js
novemberborn: Yea I was having my doubts about this. On the one hand it's an internal API so hey, but on the other it's a lot of arguments…

readme.md
novemberborn: > This feature, while useful, is going to hit a lot of people. Any chance we could add support for assertions throwing AssertionError?

That's not the problem, actually. Any custom assertion that throws an error causes the test to fail, just because tests shouldn't throw errors.

The issue is when all custom assertions pass and no errors are thrown. The intent is for the test to pass, but we don't know that custom assertions were executed. So it'll fail. That's why I added this option.

IMHO the question is what the default behavior should be.

novemberborn:
> I see. I use the pattern of assigning t.throws to err a lot, so having to guard err would be inconvenient

OK so with `0.18`:

```js
import test from 'ava'

test('sync', t => {
  const err = t.throws(() => {})
  console.error('After sync', err)
})
```

This results in:

```
❯ "$(npm bin)"/ava --verbose test.js

After sync null
  ✖ sync Missing expected exception..

  1 test failed [14:07:57]

  sync
  /private/var/folders/_6/p8qxp_3n62zg9081tvb0lcc80000gn/T/tmp.HrBeKHcUVs/test.js:4

   3: test('sync', t => {
   4:   const err = t.throws(() => {})
   5:   console.error('After sync', err)

  Missing expected exception..
```

Now async:

```js
test('async', async t => {
  const err = await t.throws(Promise.resolve())
  console.error('After async', err)
})
```

```
❯ "$(npm bin)"/ava --verbose test.js

  ✖ async Missing expected exception..

  1 test failed [14:08:26]

  async

  Missing expected exception..
```

What happens is that the promise returned by `t.throws()` is rejected, so the `AssertionError` is thrown by `await` and the subsequent code never runs.

With this branch:

```
❯ "$(npm bin)"/ava --verbose test.js

After sync undefined
  ✖ sync
After async undefined
  ✖ async

  2 tests failed [14:10:04]

  sync
  /private/var/folders/_6/p8qxp_3n62zg9081tvb0lcc80000gn/T/tmp.HrBeKHcUVs/test.js:4

   3: test('sync', t => {
   4:   const err = t.throws(() => {})
   5:   console.error('After sync', err)




  async
```

(OK just now noticing that the output for `async` seems to be truncated, that seems like a bug.)

Now, both `sync` and `async` behave the same. Will subsequent code crash? If so, that's the same as the current behavior for `async`. Will an assertion fail? If so, that's the same as the current behavior for `sync`. Can the user tell? No, since we only show the first assertion.

novemberborn:
> (OK just now noticing that the output for async seems to be truncated, that seems like a bug.)

Pushed a commit to fix this.

readme.md
sindresorhus: Could we maybe have it on by default, but disable it if the user imports any of the popular assertion libs? That way we can have a good default, but also support people using external assertion libs without them having to configure something.

sindresorhus:
Got it. My problem with this change is that it breaks a popular pattern that we also document:

```js
test('async', async t => {
  const err = await t.throws(Promise.resolve());
  t.is(err.message, 'foo');
})
```

Will result in:

```
[TypeError: Cannot read property 'message' of undefined]
```

Which is not very helpful for the user.

---

Do you remember why and when we stopped throwing on assertion failures? Seems like that would have resolved all of this, as no code would run after a failed assertion. What is even the benefit of assertions not throwing?

Another workaround would be to just return an empty object until we implement https://github.com/avajs/ava/issues/1047. I just don't want to force the user to change their code twice.

readme.md
novemberborn: I'm not sure how we can make that detection reliable and consistent across test files.

novemberborn:
> Got it. My problem with this change is that it breaks a popular pattern that we also document:
> 
> ```js
> test('async', async t => {
>   const err = await t.throws(Promise.resolve());
>   t.is(err.message, 'foo');
> })
> ```
> 
> Will result in:
> 
> ```
> [TypeError: Cannot read property 'message' of undefined]
> ```
> 
> Which is not very helpful for the user.

Interesting! That's happening because we now only check the pending assertions for failures after we've checked the return value. I think that's a bug, even if `t.throws()` returned a rejected promise we'd be *wrapping* it with a new `AssertionError`, since the `Test` code assumes that no `AssertionError`s leak out of the test.

> Do you remember why and when we stopped throwing on assertion failures? Seems like that would have resolved all of this, as no code would run after a failed assertion.

I don't know. It's hard to tell, currently `empower` doesn't rethrow but I don't know if it did before or if we disabled that.

> What is even the benefit of assertions not throwing?

If you have a callback test (or even when you wrap a legacy API in a promise), code may run in a separate call stack. If the assertion throws that might become an uncaught exception, completely crashing your test run.

> Another workaround would be to just return an empty object until we implement #1047. I just don't want to force the user to change their code twice.

That's an option, though with the bug I described above it would mean the `t.is(err.message, 'foo')` assertion gets reported, not the `t.throws()` assertion failure. And once I fix that bug I think not throwing becomes harmless, since subsequent errors are not shown to the user.

(Unless of course they cause an uncaught exception in some odd test, but even returning an empty object could cause that.)

novemberborn:
> That's happening because we now only check the pending assertions for failures after we've checked the return value. I think that's a bug

I've pushed a fix:

```
❯ "$(npm bin)"/ava test.js

  1 failed

  async
  /private/var/folders/_6/p8qxp_3n62zg9081tvb0lcc80000gn/T/tmp.KKt3WiWUVJ/test.js:4

   3: test('async', async t => {
   4:   const err = await t.throws(Promise.resolve());
   5:   t.is(err.message, 'foo');

  Test.<anonymous> (test.js:4:23)
  step (test.js:9:191)
  Test.<anonymous> (test.js:9:99)
  Test.__dirname [as fn] (test.js:3:1)
```

readme.md
sindresorhus: Me neither. Maybe not worth it. We can discuss in a new issue after this is merged. Let's go with on by default for now.

sindresorhus:
@novemberborn Amazing. I'm good with it now. :shipit:  :shipit:  :shipit:

jamestalmage:
No comments on this PR yet, just some background:

As far as I know, we have never allowed assertions failures to throw. This is a relic from `tap`/`tape` which exhibit the same behavior. The big difference for those libraries is that they show the status of every assertion in their output. Not throwing allows them to check the status of future tests. We only show our first assertion failure. This is a result of our parallel execution of tests, which would make it difficult to inline assertions.

In `tap` / `tape`, this works as a good test for your `add` function:

```js
test('add', t => {
  t.is(add(2, 2), 4);
  t.is(add(2, 3), 5);
  t.is(add(3, 3), 6);
  // ....
});
```

In AVA, you would stop at the first failure, so it's better to use macros:

```js
function macro(t, lhs, rhs, expected) {
  t.is(add(lhs, rhs), expected);
}

test(macro, 2, 2, 4);
test(macro, 2, 3, 5);
test(macro, 3, 3, 6);
```

This moves each assertion into separate tests, so a failure of one does not mask failures down the line.

`tap` offers some convenience here for very simple tests. You don't need to separate it out like you do in AVA to see every failure.

Right now, what we are doing does not make a lot of sense. We only display the first failed assertion, but we continue on after that assertion collecting data we will never display. We have discussed a flag to display every failed assertion in #261.

lib/test-worker.js
```diff
@@ -1,38 +1,56 @@
...
 
-const opts = process.opts;
-const testPath = opts.file;
+// Check if the test is being run without AVA cli
+{
+	/* eslint-disable import/order */
+	const path = require('path');
+	const chalk = require('chalk');
+
+	const isForked = typeof process.send === 'function';
```
jamestalmage: I am pretty sure moving this in here will break `karma-ava`. Anything that relies on the Node environment (vs the browser) should remain in `process-adapter`.

`process-adapter` is swapped out by `karma-ava` when compiling for the browser with an implementation that polyfills any missing functionality.

test/fixture/long-running.js
```diff
@@ -2,6 +2,7 @@ import test from '../../';
 
 test.cb('slow', t => {
 	setTimeout(t.end, 5000);
+	t.pass();
```
jamestalmage: We shouldn't force users to use `t.pass` in this scenario. You can pass `t.end` as an error first callback, so passing `t.end` really should count as performing an assertion.

lib/test-worker.js
novemberborn: `karma-ava` is pretty experimental though. We don't have a stable API so IMHO it's fine if this breaks `karma-ava`.

If necessary this could live in its own file. It's odd in `process-adapter`. Heck it's a bit odd to have it in `test-worker` versus just `main`.

test/fixture/long-running.js
novemberborn: Good point. Will push a fix imminently.

lib/test-worker.js
jamestalmage: I am not so concerned about breaking `karma-ava` in the short term. But we need a strategy for where we will locate Node stuff that needs to be polyfilled for the browser. This is going to necessarily create some "oddness", as you will encounter the need to abstract out stuff you might prefer to inline throughout the codebase.

lib/sequence.js
```diff
@@ -1,86 +1,70 @@
...
-		const length = this.tests.length;
+		const iterator = this.runnables[Symbol.iterator]();
 
-		for (let i = 0; i < length; i++) {
-			// If last item failed and we should bail, return results and stop
-			if (this.bail && !this.passed) {
-				return this._results();
+		let activeRunnable;
+		const onBeforeExit = () => {
+			if (activeRunnable.finishDueToInactivity) {
```
jamestalmage: Where is the `finishDueToInactivity` method defined?

lib/test-worker.js
novemberborn: Sure, let's cross that bridge when we get there.

lib/sequence.js
novemberborn: In `Test`. See https://github.com/avajs/ava/pull/1314/commits/3e3b2131ff08838e86e34e99940a0a8ec1208e30.

Issue #1275, opened 2017-02-17
Snapshot recap


novemberborn:
A summary of snapshot related issues. See #1218, #1220, #1254 and #1223.

---

Fundamentally, `jest-snapshot` is string based. The snapshots contain human-readable representations of expected values, which are then compared against another representation of the actual value when tests are run.

Snapshot testing is useful when code generates complex tree structures that are tedious to verify by writing assertions, and that may change significantly as part of day-to-day development. Humans manually verify that the snapshot is as expected the first time it's generated. To do this, the snapshot needs to be readable. Then, when the actual value changes and the snapshot is regenerated, source control diffing makes it easy to see what's actually changed and if that is as expected.

The original snapshot PR (#1113) used `jest-snapshot` as-is (thanks @lithin!). Here, `jest-snapshot` was also responsible for the error message:

```
expect(value).toMatchSnapshot()

Received value does not match stored snapshot 1.

- Snapshot
+ Received

Object {
-   "foo": "bar",
+   "foo": "bar!",
  }
```

Soon after this landed @vadimdemedes started working on improving our error output, specifically showing diffs for `t.deepEqual()` and `t.is()` assertions (#1154). This work also included `t.snapshot()`. However, to generate these diffs we need actual objects, not the error message that `jest-snapshot` returns. (Plus, that `expect(value).toMatchSnapshot()` line is a bit jarring.)

The solution was to [send a JSON string into `jest-snapshot`](https://github.com/avajs/ava/pull/1154/files#diff-f76cd6d4ba3dad8b6b9a393af1fce15bR168), with a workaround for React trees. This means though that the snapshot contains a fairly unreadable JSON string (#1254). Also, whereas `jest-snapshot` normalizes key order in `Object` objects, AVA can no longer rely on that because we're now just comparing JSON strings. Then when we generate a diff it comes up empty since our diffing logic does not care about key order (#1220).

@vadimdemedes started a PR (#1223) to replace `jest-snapshot` with our own implementation, using `t.deepEqual()` for the comparisons. The advantage of this approach is that diffs look the same for all our (diffing) assertions. Unfortunately, because it uses JSON serialization it limits the kinds of object structures we can safely snapshot:

* `undefined`, `Symbol()` and function property values are dropped
* `undefined`, `Symbol()` and function array items are replaced by `null`
* regular expression property values and array items are replaced by `{}`
* no support for built-ins such as `Map` and `Set`
* no support for custom classes: instances are reduced to plain objects
[...]

PR #1317, opened 2017-03-21
Make snapshots usable again

### LINKED ISSUE #1218: Snapshot feature will modify existing snapshots without passing `--update-snapshots` flag

### LINKED ISSUE #1220: Snapshot feature shows empty diff when changing order of keys on object
## Description

When using the snapshot feature, if the new value is an object that is identical to the snapshot **but with a different key order**, the diff shown is **empty**.

### Test Source

Before:

```js
	...
	t.snapshot({
		coolNumber: 5 + 8,
		otherNumber: 9 + 8,
	});
	...
```

After altering the test so that the keys are in reversed order:
```js
	...
	t.snapshot({
		otherNumber: 9 + 8,
		coolNumber: 5 + 8,
	});
	...
```

```
  1 failed

  banana
  test/test.js:6

   5: test('banana', async t => {
   6:   t.snapshot({             
   7:     otherNumber: 9 + 8,    

  Difference:

 
  Please check your code or --update-snapshots

  Test.<anonymous> (test/test.js:6:4)
  Test.fn (test/test.js:5:1)
```

### Config

package.json:
```json
{
  "ava": {
    "require": [
      "babel-register"
    ],
    "babel": "inherit"
  }
}
```

.babelrc:
```json
{
    "plugins": [
        "transform-runtime"
    ],
    "presets": [
        "stage-0",
        "es2015-node4"
    ]
}
```

### Command-Line Arguments

```
ava
```

## Environment

```
Node.js v4.6.2
linux 4.8.0-34-generic

AVA 0.18.0

npm 3.10.9
```

### LINKED ISSUE #1254: Snapshot is doubly-escaped / hard to see diffs
The example from the readme gives:

```js
exports[`HelloWorld component 1`] = `"\"\\n<h1>\\n    Hello World...!\\n</h1>\\n\""`;
```

(looks like it's json-stringified, with extra `"` quotes)

I'd love to use the snapshot feature, but non-diffable snapshots are a no-go. Could the exported value not just be a json object instead of a stringified json object?

novemberborn:
Had to start with another refactor, this time of `Runner`.

Anyway, the cool part is making snapshots work again:

* Make the Runner manage the snapshot state. Thread an accessor to the
`t.snapshot()` assertion.

* Save snapshots when runner has finished. Fixes #1218.

* Use jest-snapshot directly, without serializing values. Use jest-diff
to generate the diff if the snapshot doesn't match. This does mean the
output is not colored and has other subtle differences with how other
assertions format values, but that's OK for now. Fixes #1220, fixes #1254.

* Pin jest-snapshot and jest-diff versions. This isn't ideal but we're
using private APIs and making other assumptions, so I'm not comfortable
with using a loose SemVer range.

It's still not ideal (see #1275), especially since the diff output is different from other assertions, but at least it makes snapshots usable.


test/fixture/assert.snap
```diff
@@ -0,0 +1,19 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
sindresorhus: We need to change this to `AVA Snapshot v1`. We don't want to mislead people with being fully compatible with Jest and it's also confusing for AVA users that it doesn't say AVA.

We might need to open an issue on Jest about getting an option for this, but for now I guess we can just read the file in, modify it, and write it out again.

sindresorhus:
This looks very good. I almost couldn't tell the difference with our diff. Usable now is better than perfect at some point.

test/fixture/assert.snap
novemberborn: I think `jest-snapshot` just treats these files as CommonJS modules so rewriting the header should be harmless. Will do.

lib/runner.js
```diff
@@ -95,100 +135,55 @@ class Runner extends EventEmitter {
...
 			return Promise.resolve(null);
 		}
 
-		this.tests.on('test', this._addTestResult);
-
 		this.hasStarted = true;
-
-		return Promise.resolve(this.tests.build().run()).then(this._buildStats);
+		this.tests.on('test', result => this.addTestResult(result));
+		return new Promise(resolve => {
```
novemberborn: Ah `Promise` here is indeed from `bluebird`. I guess I'll rename it to `Bluebird.try()`.

lib/runner.js
```diff
@@ -95,100 +135,55 @@ class Runner extends EventEmitter {
...
 		if (options.runOnlyExclusive && !this.tests.hasExclusive) {
 			return Promise.resolve(null);
 		}
 
-		this.tests.on('test', this._addTestResult);
-
 		this.hasStarted = true;
-
-		return Promise.resolve(this.tests.build().run()).then(this._buildStats);
+		this.tests.on('test', result => this.addTestResult(result));
```
novemberborn: Sure.

test/fixture/assert.snap
novemberborn: Actually no, `jest-snapshot` requires snapshot files to start with this header: https://github.com/facebook/jest/blob/38ca942f6f2979ed25f54510173dbd2761dd7e04/packages/jest-snapshot/src/utils.js#L37

We could add a comment after but that becomes annoyingly complex.

sindresorhus: @novemberborn Can you open an issue on Jest to make it configurable? They seem open to making it more generic when needed.

I'm happy to land this as is for now, but long-term we'll want our own naming and versioning.

novemberborn: > Can you open an issue on Jest to make it configurable? They seem open to making it more generic when needed.

I don't know, if you look at the [error messages it very much assumes running inside Jest](https://github.com/facebook/jest/blob/38ca942f6f2979ed25f54510173dbd2761dd7e04/packages/jest-snapshot/src/utils.js#L40:L76).

> I'm happy to land this as is for now, but long-term we'll want our own naming and versioning.

I think we want our own alternative, really.

sindresorhus: Too bad, would have been nice to reuse that code.

// @cpojer

cpojer:
They are still Jest snapshots though, aren't they? If we can, we should keep the header because it looks to me like they are compatible. We can update error messages and the documentation link that it links to, though. Feel free to send PRs to Jest to make the public API more usable as we'll undoubtedly break private APIs in the future.

Issue #1288, opened 2017-02-25
Ava slows when there are many non-run files


taylor1791:
## Description

I started a project a while ago and I noticed that `ava` slowed down over time. At the start it used to clock in at under a second. As I added more code and my configuration changed it is now around 15 seconds. This is an extremely long feedback cycle. Using `--watch` does not appear to produce a significant improvement.

I managed to reproduce the problem in a [test repo](https://github.com/taylor1791/ava-slow-test). Further testing show it has to do with the _number of files in the directory_. It does not matter what type of files they are. In the example repo, I have a 1 simple test, a copy of node, and a benchmark script. When I remove the copy of node it is significantly faster. Here is a sample run of the script.

```
user@computer:~/src/ava-slow-test% ./benchmark.sh
Running tests...

  1 passed


real	0m7.377s
user	0m6.863s
sys	0m1.379s
Removing lib and re-running tests...

  1 passed


real	0m0.737s
user	0m0.589s
sys	0m0.100s
Resetting lib...
```

## Environment

```
user@computer:~/src/ava-slow-test% node -e "var os=require('os');console.log('Node.js ' + process.version + '\n' + os.platform() + ' ' + os.release())"
Node.js v7.4.0
darwin 16.4.0
user@computer:~/src/ava-slow-test% ./node_modules/.bin/ava --version
0.18.2
user@computer:~/src/ava-slow-test% npm --version
4.0.5
```

## Links

[Test Repository](https://github.com/taylor1791/ava-slow-test)

taylor1791:
I just noticed that this may be related to #1228.

novemberborn:
AVA needs to find your tests. It uses globbing patterns, which by default are quite wide (`test.js test-*.js test/**/*.js **/__tests__/**/*.js **/*.test.js`). If all your tests are in one particular folder you could specify the `files` option in the AVA config and point it at that directory.

@avajs/core I think we can consider this a [common pitfall](https://github.com/avajs/ava/blob/master/docs/common-pitfalls.md), so we should at least document it. Perhaps we should log a warning if globbing takes X amount of time?

sotojuan:
I definitely think it should at least be a common pitfall. Thanks @taylor1791 for the detailed issue!

taylor1791:
[...]

Issue #909, opened 2016-06-09
Only exclude `helpers` directory when inside a test directory


rymohr:
Any chance the exclude rule for `helpers` can be relaxed so it only applies to files within `test/helpers`? I just ran into a `couldn't find any files to test` error for a test file that definitely existed. Ended up being because it was in a `helpers` dir nested deep within the package I was working on.

I wasn't even aware ava had a concept of helpers until I ran into this. Where are they documented?

Here's the related PR https://github.com/avajs/ava/pull/357 and issues https://github.com/avajs/ava/issues/355 https://github.com/avajs/ava/issues/750

_Edit: added issue #750_

jamestalmage:
I think the current pattern is here to stay. Instead I think we should try to better document the patterns.

I think a `structuring-your-project` recipe is in order.


Also, I think our `no tests found` should link to it.

rymohr:
Sounds like you've made up your mind. But for what it's worth, it's not so much better docs on how to structure your project. For me it was the idea of bringing ava in to unit test small chunks of a very large existing project with an established file structure. I like my tests to live side-by-side with the files they are testing and the current `helper` exclude policy forced me to restructure my project in order to use ava.

Also, by renaming the issue you've made it hard for anyone in the same boat to find this issue in the future. Renaming issues for clarity is fine, but please don't use renaming to change the intent of the issue.

jamestalmage:
> I wasn't even aware ava had a concept of helpers until I ran into this. Where are they documented?

That's what made me assume it was a documentation issue.

--

I guess it would be possible to provide yet another glob pattern for users to define what constitutes a helper, but we're pretty against adding config options unless there is a strong need for it.

Can you fully define your directory structure?

rymohr:
If it doesn't make sense to relax the exclude rule from `helpers` to `test/helpers` then it would definitely be nice to be able to override the ignore.

Here's an example of where we ran into this problem:

```
client/
  modules/
    surveys/
      helpers/
        prepareScreen.js
        prepareScreen.test.js
```

In our case we just renamed `helpers` to `utils` to get around the issue for now but I still think it's strange that ava chokes on this by default. The `node_modules` exclude is a smart default but I feel like the `helpers` exclude is a special case that should be opt-in.

jamestalmage:
I'm wondering if we couldn't relax it just a bit, to something like this:

```
src/
  helpers/        <- not excluded
  foo/
   helpers/       <- not excluded
[...]

PR #1320, opened 2017-03-23
Fix default helper files glob to cover helper files inside subdirectories

### LINKED ISSUE #1319: Ava does not transpile helpers!
## Description

Ava should transpile helper files without requiring `babel-register` or any such a thing.

If the test files are inside a `test` folder, this works.

**However:** if the files are inside another folder (e.g: `my-tests` instead of `test`) **or** are inside a subfolder (e.g: `test/unit`) this does not work.

**TLDR: If tests are inside a subfolder of `test` directory, ava DOES NOT TRANSPILE helper files**

### Test Source

```js
/* FILE: test/unit/helpers/index.js */
import Promise from 'bluebird'

export const delay = (n) => Promise.delay(n)

/* ============================== */

/* FILE: test/unit/index.js */
import test from 'ava'
import { delay } from './helpers'

test('should return true', async (t) => {
  console.log('starting ===>', Date.now())
  await delay(2000)
  console.log('finishing ==>', Date.now())

  return true
})
```

```
~/H/l/ava (master) [n] ⋊> ./node_modules/.bin/ava test
/Users/grvcoelho/Hyperspace/labs/ava/test/unit/helpers/index.js:1
(function (exports, require, module, __filename, __dirname) { import Promise from 'bluebird'
                                                              ^^^^^^
SyntaxError: Unexpected token import
    at Object.exports.runInThisContext (vm.js:78:16)
    at Module._compile (module.js:543:28)
    at Module._extensions..js (module.js:580:10)
    at extensions.(anonymous function) (/Users/grvcoelho/Hyperspace/labs/ava/node_modules/require-precompiled/index.js:16:3)
    at Object.require.extensions.(anonymous function) [as .js] (/Users/grvcoelho/Hyperspace/labs/ava/node_modules/ava/lib/process-adapter.js:105:4)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
    at Module.require (module.js:498:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/Users/grvcoelho/Hyperspace/labs/ava/test/unit/request.js:3:1)
    at Module._compile (module.js:571:32)
    at extensions.(anonymous function) (/Users/grvcoelho/Hyperspace/labs/ava/node_modules/require-precompiled/index.js:13:11)
    at Object.require.extensions.(anonymous function) [as .js] (/Users/grvcoelho/Hyperspace/labs/ava/node_modules/ava/lib/process-adapter.js:105:4)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
    at Module.require (module.js:498:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/Users/grvcoelho/Hyperspace/labs/ava/node_modules/ava/lib/test-worker.js:33:1)
    at Module._compile (module.js:571:32)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
    at Module.runMain (module.js:605:10)
    at run (bootstrap_node.js:420:7)
    at startup (bootstrap_node.js:139:9)
    at bootstrap_node.js:535:3

  1 exception

  ✖ test/unit/request.js exited with a non-zero exit code: 1

```

### Config

**No configuration, which means no babel-register or such things**

```json

{
  "ava": {
  ...
  }
}
```

### Command-Line Arguments

```
ava
```

## Relevant Links

**Repo with detailed instructions to reproduce the behavior**: https://github.com/wasting-light/avajs-bug-report-x89

## Environment

```
Node.js: v7.4.0
OS: darwin 16.4.0
Ava: 0.18.2
npm: 4.0.5
```

grvcoelho:
## Description

When we have a `helpers` folder inside a subdirectory (e.g: `test/unit/helpers`), the helper files are not transpiled by Ava.

This PR fixes the glob that finds the defaultHelperFiles to cover subdirectories as well.

## Related

This fixes #1319

### DIFF
--- a/lib/ava-files.js
+++ b/lib/ava-files.js
@@ -88,9 +88,9 @@ const defaultIncludePatterns = () => [
 ];
 
 const defaultHelperPatterns = () => [
-	'**/__tests__/helpers/**/*.js',
+	'**/__tests__/**/helpers/**/*.js',
 	'**/__tests__/**/_*.js',
-	'**/test/helpers/**/*.js',
+	'**/test/**/helpers/**/*.js',
 	'**/test/**/_*.js'
 ];
 


novemberborn:
I'm not sure about this… we already have issues with helper globbing performance (#1288). Helper exclusion isn't great either (#909). Perhaps we should solve those problems first.

Just spitballing here, but what if:

* `files` patterns that match files cause those files to be treated as tests, not helpers
  * We support negation patterns to refine selections if necessary
  * We exclude `{project-dir}/node_modules` by default, though ideally you can override this by providing a more specific path
* `files` patterns that match directories mean that containing files are treated as tests, unless they're inside `fixtures` and `helpers` directories (that are children of the matched directories), or start with `_` (regardless of how deeply they are nested)

This wouldn't really help your use case, but I think it makes test file matching more understandable. We could then add support for helper patterns, so you can match those and we transpile them. A utility command that prints out what files are matched and how they're treated would also be useful.

grvcoelho:
> I'm not sure about this… we already have issues with helper globbing performance (#1288). Helper exclusion isn't great either (#909). Perhaps we should solve those problems first.

@novemberborn: I've read the discussions and I'm not sure if I understood the problem correctly, but the issue with #1288 is because of number of files and not globbing itself, right?

Because, ATM in my case, if I rename everything inside my `helpers` folder to have a `_` prefix, it will be considered a helper and be transpiled. So overall, it is the same amount of files, I just don't need to put the `_` prefix.

ATM, `_` prefixed helpers are transpiled inside subdirectories, but helpers that are inside `helpers` folder are not. So the behaviors are different if you decide do `_`-prefix your helpers or to put them inside `helpers` folder.

> We could then add support for helper patterns, so you can match those and we transpile them.

That would be great in the future 🙂

novemberborn:
> I've read the discussions and I'm not sure if I understood the problem correctly, but the issue with #1288 is because of number of files and not globbing itself, right?

I think it's because of the `**` which matches any directory, leading to a *lot* of matching.

I'm not saying this PR isn't an improvement, it's just that the whole helper story is a bit of a mess and I'm not sure whether we should add more twists to it.

sindresorhus:
I'm also concerned about the performance implications of this.

> files patterns that match files cause those files to be treated as tests, not helpers

@novemberborn How is this different from how it already works? Or do you mean the ability to do `$ ava _foo.js` and have it run as a test instead of being a helper?

novemberborn:
>> files patterns that match files cause those files to be treated as tests, not helpers
>
> @novemberborn How is this different from how it already works? Or do you mean the ability to do $ ava _foo.js and have it run as a test instead of being a helper?

@sindresorhus uhm, I hadn't even thought of it like that. I guess if you don't use a pattern it could be used as-is…? Probably easier to never treat any `_`-prefixed file as a test.

wmertens:
This approach seems to slow down my AVA runs a lot, I think it descends into my `build/` dir but I don't really know, there is no debugging output for the globbing.

How about not scanning for helper files, but instead somehow intercepting the `require` mechanism that Babel uses to transpile, only allowing it to transpile imports that match helper globs?

wmertens:
I just went into the module, prepended the helper globs with my source dirs (e.g. `'{src,lib}/**/__tests__/helpers/**/*.js'`), and `ava` run time decreased by 20s!

novemberborn:
> This approach seems to slow down my AVA runs a lot, I think it descends into my build/ dir but I don't really know, there is no debugging output for the globbing.

@wmertens is this a comment on this PR, or AVA's current behavior?

> How about not scanning for helper files, but instead somehow intercepting the require mechanism that Babel uses to transpile, only allowing it to transpile imports that match helper globs?

This is tricky to do performantly at the moment.

Let's move the performance discussion to https://github.com/avajs/ava/issues/1228#issuecomment-320452723.

novemberborn:
I still don't think adding more patterns is the right answer here. Improving the `files` matching, and supporting `helpers` patterns should be sufficient.

@grvcoelho #1319 is valid though. I'm sorry this is taking so long.

Issue #1330, opened 2017-04-01
Fail test even after it has passed


novemberborn:
09b23e0d5fd00d48f6b5b1f9c43641faa00f1d1d made it an error if an assertion is added after a test has finished. This can occur if the test finished earlier than the user intended. Once a test has passed though we cannot update its state in the reporters.

The feedback in 09b23e0d5fd00d48f6b5b1f9c43641faa00f1d1d could only be provided to users if the test was waiting for a pending assertion. That behavior itself is confusing, which means that when #1327 is fixed we won't be able to provide any feedback at all.

If a test has passed, but then later an assertion is added before other tests have finished, we should change the test state to a failure and update the reporters. The mini reporter should work well, since it hides previously passing tests. The verbose reporter would have a duplicate entry for the test but perhaps that's OK. I'm not sure whether it's acceptable to add a duplicate entry in the TAP reporter. The alternative is buffering the output for each test file which is not ideal.

sindresorhus:
> The verbose reporter would have a duplicate entry for the test but perhaps that's OK.

That's ok.

novemberborn:
This also applies to callback tests ending more than once.

novemberborn:
Note that this will still be best-effort. The test worker will exit once all tests have completed, even if there is a `setTimeout()` scheduled which double-ends a callback test, or makes an extraneous assertion.

novemberborn:
With https://github.com/avajs/ava/pull/1776, we don't write test results until the test file has finished executing. So we'll be able to fail a test even if it seemed to pass at first.

PR #1331, opened 2017-04-01
Allow new assertions while waiting for pending ones

### LINKED ISSUE #1327: Too aggressive in declaring the test "finished"
The following works in `0.18.2`:

```js
test(t => {
	t.notThrows(new Promise(resolve => {
		setTimeout(() => {
			t.pass();
			resolve();
		}, 100);
	}));
});
```

On master, it fails with:

```
Assertion passed, but test has already ended
```

We should wait until all pending assertions have resolved before declaring a test "finished" and throwing when new assertions show up.

Adding a call to `t.plan()` causes it to behave oddly going all the way back (we even added a test solidifying this odd behavior in #360):

```js
test(t => {
   t.notThrows(new Promise(resolve => {
       t.plan(2);
       setTimeout(() => {
         t.pass();
         resolve();
       }, 100);
   }));
});
```

The above fails with:

```
Planned for 2 assertions, but got 1.
```

Similar results for as far back as I tested.

I think we should allow for added assertions, as long as we still have pending assertions. When `t.plan` is used, I don't think we should check for too few assertions until we've drained all pending assertions.

novemberborn:
Whilst not a great way to write tests, users may end up adding more
assertions from inside a promise chain that's passed to the `t.throws()`
or `t.notThrows()` assertions. These should be counted for the plan, and
not fail the test as being extraneous.

Fixes #1327.

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -109,7 +109,7 @@ class Test {
 		this.endCallbackFinisher = null;
 		this.finishDueToAttributedError = null;
 		this.finishDueToInactivity = null;
-		this.finishing = false;
+		this.finished = false;
 		this.pendingAssertions = [];
 		this.pendingThrowsAssertion = null;
 		this.planCount = null;
@@ -152,7 +152,7 @@ class Test {
 	}
 
 	countPassedAssertion() {
-		if (this.finishing) {
+		if (this.finished) {
 			this.saveFirstError(new Error('Assertion passed, but test has already ended'));
 		}
 
@@ -160,7 +160,7 @@ class Test {
 	}
 
 	addPendingAssertion(promise) {
-		if (this.finishing) {
+		if (this.finished) {
 			this.saveFirstError(new Error('Assertion passed, but test has already ended'));
 		}
 
@@ -169,7 +169,7 @@ class Test {
 	}
 
 	addFailedAssertion(error) {
-		if (this.finishing) {
+		if (this.finished) {
 			this.saveFirstError(new Error('Assertion failed, but test has already ended'));
 		}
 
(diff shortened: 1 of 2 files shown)


lib/test.js
```diff
@@ -152,15 +152,15 @@ class Test {
...
 			this.saveFirstError(new Error('Assertion passed, but test has already ended'));
 		}
 
 		this.assertCount++;
 	}
 
 	addPendingAssertion(promise) {
-		if (this.finishing) {
+		if (this.finished) {
 			this.saveFirstError(new Error('Assertion passed, but test has already ended'));
```
jamestalmage: I realize this is not from this PR, but a pending assertion technically hasn't passed yet. This message should be different.

lib/test.js
```diff
@@ -365,21 +365,19 @@ class Test {
...
 		if (this.assertError || this.pendingAssertions.length === 0) {
 			return this.completeFinish();
 		}
 
-		// Finish after potential errors from pending assertions have been consumed.
-		return Promise.all(this.pendingAssertions).then(() => this.completeFinish());
+		// Retry finishing after potential errors from pending assertions have been consumed.
+		const consumedErrors = Promise.all(this.pendingAssertions);
+		// Note that more assertions may be added in the meantime.
+		this.pendingAssertions = [];
```
jamestalmage: Maybe add a test to validate the works it happens twice (pending resolve, but more pending have been added, and yet another sync will be added while you wait for those to resolve).

lib/test.js
novemberborn: I considered that when I first wrote it, but the nuance seemed superfluous. And since #1330 it's not even shown anymore.

lib/test.js
novemberborn: Done.

test/test.js
```diff
@@ -646,6 +641,54 @@ test('assertions return promises', t => {
 	});
 });
 
+// https://github.com/avajs/ava/issues/1330
+test('no crash when adding assertions after the test has ended', t => {
+	t.plan(3);
+
+	ava(a => {
+		a.pass();
+		setTimeout(() => {
```
sindresorhus: I really don't think this should be ok. Many people do the mistake of using `setTimeout` inside a sync test and it silently passing. If users need to do this they should either use an async function or `test.cb()`.

novemberborn: @sindresorhus It's not OK, that's what #1330 is about. This merely verifies that the internal handling doesn't cause a crash, since the error cannot currently be communicated.

sindresorhus: Ah, ok.

PR #1334, opened 2017-04-02
Don't highlight formatted values with --no-color

### LINKED ISSUE #843: Switch to disable color output to console

novemberborn:
Fixes #843 (https://github.com/avajs/ava/issues/843#issuecomment-290935567). Values shouldn't be highlighted during formatting if colors are disabled.

Note that this only works when colors are disabled explicitly. @sindresorhus perhaps we should change the default value of the `color` option if we can detect colors are not supported?


sindresorhus:
> perhaps we should change the default value of the color option if we can detect colors are not supported?

👍

novemberborn:
@sindresorhus see new commits 😃

PR #1353, opened 2017-04-10
Make `t.is()` use `Object.is()`

### LINKED ISSUE #1339: Feature request: Use `Object.is()` instead of `===` in `t.is()`
I was using AVA to test some math stuff the other day and noticed that I needed to use `t.deepEqual()` to test for `NaN`. If the `===` test was swapped out for `Object.is()` in [`assert.js`](https://github.com/avajs/ava/blob/master/lib/assert.js#L62) it would be able to correctly handle `NaN` being 'equal' to `NaN` without having to resort to `deepEqual`.

The loss we get with switching to `Object.is` as far as [MDN says](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness#A_model_for_understanding_equality_comparisons) is that `+0` and `-0` are not longer treated as equal. That's actually probably okay, as they are 'more different' than two `NaN`s are, and certainly the `NaN` problem is far more common than "why doesn't it think my two zeroes are the same?" it could be argued that, during a test, you may well wish to know if you are getting two different zeroes.

Checking the `package.json` [it says](https://github.com/avajs/ava/blob/master/package.json#L43) that the 'engine' must be `node >= 4`, and Node.js 4 does indeed support `Object.is()` according to [the ES compatibility table](https://kangax.github.io/compat-table/es6/#node4).

---

All that said, the same MDN article linked above, [further down](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness#When_to_use_Object.is_versus_triple_equals), says that maybe using `Object.is()` could introduce more confusion (though it'd be interesting to know if power assert can correctly give meaningful output in a diff of `+0` and `-0`, as that would probably be what this hinges on) so may best be avoided. If it is decided that this is the case, I thinking adding the specific `isNaN` checks to `is` in `assert.js` would still be really helpful as, arguably, `NaN` *is* (in the sense of `t.is`) `NaN`, it's just not "strictly 100% guaranteed to be actually the exact same *type* of `NaN` value as the `NaN` you passed into the `t.is` function", which is not what people want when they say` t.is`.

alexrussell:
Refactor `is` assertion to use `Object.is()` rather than `===`.
Also add a few more test cases in for `is` to make it clearer what it is expected to do.

As per discussion in #1339, this will make it more sensible to compare `NaN` against `NaN`.

The good:

- `NaN` now compares to `NaN` in a way that people expect (truthily)

The odd:

- `+0`/`0` and `-0` are now **not** equal. As we discussed in #1339, this actually feels fine - this being a testing framework, it's good to be strict on this math stuff, while sensible on the NaN stuff.

Fixes #1339 

/cc @novemberborn, @sotojuan, @sindresorhus

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -59,31 +59,29 @@ function wrapAssertions(callbacks) {
 		},
 
 		is(actual, expected, message) {
-			if (actual === expected) {
+			if (Object.is(actual, expected)) {
 				pass(this);
 			} else {
 				const diff = formatAssertError.formatDiff(actual, expected);
 				const values = diff ? [diff] : [
 					formatAssertError.formatWithLabel('Actual:', actual),
-					formatAssertError.formatWithLabel('Must be strictly equal to:', expected)
+					formatAssertError.formatWithLabel('Must be strictly equal (using Object.is) to:', expected)
 				];
 
 				fail(this, new AssertionError({
 					assertion: 'is',
 					message,
-					operator: '===',
 					values
 				}));
 			}
 		},
 
 		not(actual, expected, message) {
-			if (actual === expected) {
+			if (Object.is(actual, expected)) {
 				fail(this, new AssertionError({
 					assertion: 'not',
 					message,
-					operator: '!==',
-					values: [formatAssertError.formatWithLabel('Value is strictly equal:', actual)]
+					values: [formatAssertError.formatWithLabel('Value is strictly equal (using Object.is):', actual)]
 				}));
 			} else {
 				pass(this);
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -65,7 +65,7 @@ function wrapAssertions(callbacks) {
 				const diff = formatAssertError.formatDiff(actual, expected);
 				const values = diff ? [diff] : [
 					formatAssertError.formatWithLabel('Actual:', actual),
-					formatAssertError.formatWithLabel('Must be strictly equal (using Object.is) to:', expected)
+					formatAssertError.formatWithLabel('Must be the same as:', expected)
 				];
 
 				fail(this, new AssertionError({
@@ -81,7 +81,7 @@ function wrapAssertions(callbacks) {
 				fail(this, new AssertionError({
 					assertion: 'not',
 					message,
-					values: [formatAssertError.formatWithLabel('Value is strictly equal (using Object.is):', actual)]
+					values: [formatAssertError.formatWithLabel('Value is the same as:', actual)]
 				}));
 			} else {
 				pass(this);
--- a/readme.md
+++ b/readme.md
@@ -632,8 +632,8 @@ function macro(t, input, expected) {
 	t.is(eval(input), expected);
 }
 
-test('2 + 2 === 4', macro, '2 + 2', 4);
-test('2 * 3 === 6', macro, '2 * 3', 6);
+test('Object.is(2 + 2, 4)', macro, '2 + 2', 4);
+test('Object.is(2 * 3, 6)', macro, '2 * 3', 6);
 ```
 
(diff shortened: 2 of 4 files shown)


test/assert.js
```diff
@@ -130,10 +235,31 @@ test('.is()', t => {
...
+	});
+
+	failsWith(t, () => {
+		assertions.is(0, -0, 'my message');
+	}, {
+		assertion: 'is',
+		message: 'my message',
+		values: [
+			{label: 'Actual:', formatted: /0/},
+			{label: 'Must be strictly equal (using Object.is) to:', formatted: /0/}
```
novemberborn: I suspect this outputs `0`, rather than `-0`? https://github.com/avajs/ava/pull/1341 would fix, but I think we should work around this in this branch too.

lib/assert.js
```diff
@@ -59,31 +59,29 @@ function wrapAssertions(callbacks) {
...
 		is(actual, expected, message) {
-			if (actual === expected) {
+			if (Object.is(actual, expected)) {
 				pass(this);
 			} else {
 				const diff = formatAssertError.formatDiff(actual, expected);
 				const values = diff ? [diff] : [
 					formatAssertError.formatWithLabel('Actual:', actual),
-					formatAssertError.formatWithLabel('Must be strictly equal to:', expected)
+					formatAssertError.formatWithLabel('Must be strictly equal (using Object.is) to:', expected)
```
novemberborn: Perhaps we can take a leaf from the [`Object.is()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) description and say `Must be the same value as:`.

readme.md
```diff
@@ -914,23 +914,23 @@ Assert that `value` is `false`.
 
 ### `.is(value, expected, [message])`
 
-Assert that `value` is equal to `expected`.
+Assert that `value` is equal to `expected` (using `Object.is()` for equality testing).
```
novemberborn: We could say:

> Assert that `value` is the same as `expected`. This is based on [`Object.is()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is).

alexrussell:
@novemberborn the latest commit addresses all three of these. I have gone for "the same as" rather than "the same value as" (you used both in your two comments on this) so let me know if you'd prefer "the same value as".

Also, the `0`/`-0` thing was my bad - Just adding the `-` into the test expectation worked, so it appears that the value formatter does indeed include the negative sign. Phew!

I think when I originally did this work on the tests I was really unclear as to how the AVA test helpers worked with all the formatted values and regexes, etc., so I went for a conservative `0` and simply forgot to go back and see whether it worked as I'd expect.

readme.md
```diff
@@ -914,11 +914,11 @@ Assert that `value` is `false`.
 
 ### `.is(value, expected, [message])`
 
-Assert that `value` is equal to `expected` (using `Object.is()` for equality testing).
+Assert that `value` is the same as `expected`. This is based on using [`Object.is()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is).
```
novemberborn: The "using" seems superfluous. Either `This is based on` or `This relies on` I think. Also for `.not()`.

alexrussell: My bad. Updated.

sindresorhus:
I'm a little bit sceptical of differentiating between `-0` and `+0`, but I don't think it happens that often in real code anyways. Let's try it out and we can always revert if it's a big annoyance to users.

sindresorhus:
Thank you @alexrussell :)

alexrussell:
Woohoo thanks for merging 🎉

Issue #1092, opened 2016-10-23
Please do not recommend browser-env for browser testing


domenic:
That package uses [an explicitly unsupported jsdom anti-pattern](https://github.com/tmpvar/jsdom/wiki/Don't-stuff-jsdom-globals-onto-the-Node-global) and is very bad practice to recommend to consumers of ava.

sindresorhus:
// @lukechilds

lukechilds:
Regarding the two main points in that link, the docs do recommend only requiring the properties needed to prevent unnecessary globals and I don't think attaching them to `global` is as big of an issue as it normally would be due to the way AVA runs each test file in a separate process.

That said, I'm aware this isn't the cleanest solution in the world. To match the [example](https://github.com/tmpvar/jsdom/wiki/Don't-stuff-jsdom-globals-onto-the-Node-global#an-example-in-a-test-runner) I think we would have create a new `jsdom` instance inside every test because they all run async in AVA. It would also require that the module has been built for the browser so it can be injected in via a script tag.

I feel like the current solution gets you 99% of the way there in one function call and makes sense until AVA is running natively in the browser.

domenic:
> It would also require that the module is first built for the browser so it can be injected in via a script tag.

Yes. That is exactly the point! You need to test the same scenario as you will be deploying to your users.

lukechilds:
I understand your point. I'm not quite sure how you would get the react example working in the current recipe. It loads extra modules for testing which would also need to be built for browsers.

I don't disagree with what you're saying, I still think this method is ok to easily test simple modules, but maybe it would be better to not recommend `browser-env` because more complex modules may end up passing tests by using Node.js functionality that wouldn't be available in a sandboxed jsdom environment.

Akkuma:
To defend browser-env. I feel it does exactly what it should/could do. I'm using it for testing a browser web extension, which is inherently even more difficult to test. I already have browser automation tests to guarantee if X is done Y happens. However, I can see the need for it to go away once AVA supports running in a real browser.

domenic:
I understand that some people want to abuse jsdom in ways it's not meant to be used and that are not supported. I especially understand if the author of a module geared around doing so wants to do that. But I don't think a widely-used test-runner such as ava should be recommending such practices.

lukechilds:
I can see your concern. Maybe we should add a disclaimer in the browser recipe that makes it clear that this is not recommended and is a temporary workaround until AVA is supported in the browser.

[...]

PR #1355, opened 2017-04-11
Add warning to browser recipe

lukechilds:
Ok so since writing [`browser-env`](https://github.com/lukechilds/browser-env) I have come up with a cleaner solution, [`window`](https://github.com/lukechilds/window).It's been working well for me so far and resolves some of the issues raised in #1092.

However, it requires a [small code change](https://github.com/lukechilds/window#universal-testing-pattern) and is only viable for smaller modules without lots of dependencies that also require a browser environment. For other modules you'll still need `browser-env`. For that reason I think it would be best to continue to recommend `browser-env`, but highlight the issues raised in #1092 and offer `window` as a possible alternative.

Happy to change the wording, just pretty much copied it over from what I've got in the `browser-env` readme.

### DIFF
--- a/docs/recipes/browser-testing.md
+++ b/docs/recipes/browser-testing.md
@@ -9,6 +9,10 @@ This recipe works for any library that needs a mocked browser environment.
 
 ## Install browser-env
 
+> **❗️ Important note**
+>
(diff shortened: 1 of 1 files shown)


novemberborn:
`window` looks neat!

I wonder if the recipe could start with that, and then lay out why (though ideal) it's not always a good fit. With the caveat that could lead up to `browser-env`. This way perhaps people realize `window` is indeed sufficient.

What do you think?

lukechilds:
Thanks :)

I considered that but decided it was best to leave as is because, while `window` is definitely the ideal solution, it's only viable if they aren't using 3rd party browser-only modules and if they're happy to make a slight modification to their module. I think that instantly removes a large chunk of the people that will be reading the recipe.

Also I think it could cause some confusion to less advanced users. They may try and use `window` not understanding why they should be using `browser-env` and get cryptic `a is not defined` errors from minified browser libs. I think `browser-env` should be the default because it'll always work and then the users who are able to use `window` can opt in.

novemberborn:
Fair enough.

PR #1373, opened 2017-04-30
Fail hard when `--concurrency` is used without a value

### LINKED ISSUE #1367: --concurrency without value should fail
See https://github.com/avajs/ava/issues/1365#issuecomment-296146387. `--concurrency` requires a value but you can run AVA without specifying it.

yatharthk:
Fixes #1367

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -118,6 +118,10 @@ exports.run = () => {
 		throw new Error(colors.error(figures.cross) + ' Watch mode is not available in CI, as it prevents AVA from terminating.');
 	}
 
+	if (hasFlag('--concurrency') || hasFlag('-c')) {
+		throw new Error(colors.error(figures.cross) + ' The --concurrency and -c flags must be provided the maximum number of test files to run with.');
+	}
+
 	if (hasFlag('--require') || hasFlag('-r')) {
 		throw new Error(colors.error(figures.cross) + ' The --require and -r flags are deprecated. Requirements should be configured in package.json - see documentation.');
 	}
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -118,8 +118,8 @@ exports.run = () => {
 		throw new Error(colors.error(figures.cross) + ' Watch mode is not available in CI, as it prevents AVA from terminating.');
 	}
 
-	if (hasFlag('--concurrency') || hasFlag('-c')) {
-		throw new Error(colors.error(figures.cross) + ' The --concurrency and -c flags must be provided the maximum number of test files to run with.');
+	if (cli.flags.concurrency === '') {
+		throw new Error(colors.error(figures.cross) + ' The --concurrency and -c flags must be provided the maximum number of test files to run at once.');
 	}
 
 	if (hasFlag('--require') || hasFlag('-r')) {
--- a/test/cli.js
+++ b/test/cli.js
@@ -364,11 +364,22 @@ test('bails when config contains `"tap": true` and `"watch": true`', t => {
 	});
 });
 
-test('bails when config does not contain `--concurrency` and `-c` values', t => {
-	execCli(['test.js'], {dirname: 'fixture/concurrency'}, (err, stdout, stderr) => {
-		t.is(err.code, 1);
-		t.match(stderr, 'The --concurrency and -c flags must be provided the maximum number of test files to run with.');
-		t.end();
+['--concurrency', '-c'].forEach(concurrencyFlag => {
+	test(`bails when ${concurrencyFlag} provided without value`, t => {
+		execCli([concurrencyFlag, 'test.js'], {dirname: 'fixture/concurrency'}, (err, stdout, stderr) => {
+			t.is(err.code, 1);
+			t.match(stderr, 'The --concurrency and -c flags must be provided the maximum number of test files to run at once.');
+			t.end();
+		});
+	});
+});
+
+['--concurrency', '-c'].forEach(concurrencyFlag => {
+	test(`works when ${concurrencyFlag} provided with value`, t => {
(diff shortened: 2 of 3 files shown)


test/cli.js
```diff
@@ -364,6 +364,14 @@ test('bails when config contains `"tap": true` and `"watch": true`', t => {
 	});
 });
 
+test('bails when config does not contain `--concurrency` and `-c` values', t => {
```
yatharthk: @novemberborn This test fails with error below:

```
bails when config does not contain `--concurrency` and `-c` values
  not ok spawn /Users/yatharthk/.node-spawn-wrap-62339-36f9a5f0c8ca/node ENOENT
    at:
      line: 1026
      column: 11
      file: util.js
      function: exports._errnoException
    code: ENOENT
    errno: ENOENT
    syscall: spawn /Users/yatharthk/.node-spawn-wrap-62339-36f9a5f0c8ca/node
    path: /Users/yatharthk/.node-spawn-wrap-62339-36f9a5f0c8ca/node
    spawnargs:
      - ../../../cli.js
      - test.js
    test: bails when config does not contain `--concurrency` and `-c` values
```

I'm not really sure what's expected here. Can you also let me know if `{dirname: 'fixture/concurrency'}` is correct? I didn't understand how this works!

lib/cli.js
```diff
@@ -118,6 +118,10 @@ exports.run = () => {
 		throw new Error(colors.error(figures.cross) + ' Watch mode is not available in CI, as it prevents AVA from terminating.');
 	}
 
+	if (hasFlag('--concurrency') || hasFlag('-c')) {
```
sindresorhus: This is not the correct way to check. AVA can be run with either `ava --concurrency=1` or `ava --concurrency 1`.

lib/cli.js
```diff
@@ -118,6 +118,10 @@ exports.run = () => {
...
+		throw new Error(colors.error(figures.cross) + ' The --concurrency and -c flags must be provided the maximum number of test files to run with.');
```
sindresorhus: `the maximum number of test files to run with.` => `the maximum number of test files to run at once.`

lib/cli.js
yatharthk: @sindresorhus I see. In that case, `((hasFlag('--concurrency') || hasFlag('-c')) && !cli.flags.concurrency)` should work good, right?

sindresorhus: No, that wouldn't work if the user had `concurrency` defined in package.json.

The correct check would be:

```js
if (cli.flags.concurrency === '') {
```

As it would be an empty string if the value is not defined, as we already define `concurrency` to be a string in the `meow` config.

yatharthk: Okay, okay. Thanks for explaining.

test/cli.js
yatharthk: @sindresorhus I read the test code in detail to understand how tests are written for `cli`! Do we need to write it in a similar way as we do for `watcher` et al? Any specific cases to cover?

cc/ @novemberborn

novemberborn: These tests can be hard to get right. I'd start with running the CLI directly (`node cli.js test/fixture/concurrency/test.js` etc) to make sure it works correctly. Then try invoking it from the test.

yatharthk: @novemberborn This works! Can you tell me what all cases to cover in `test/fixture/concurrency/test.js` ?

novemberborn: CLI invocation with `--concurrency` and `-c` without a value. Presumably there are tests where a value *is* provided, but if not those should be added.

test/cli.js
```diff
@@ -364,6 +364,25 @@ test('bails when config contains `"tap": true` and `"watch": true`', t => {
 	});
 });
 
+['--concurrency', '-c'].forEach(concurrencyFlag => {
+	test(`bails when ${concurrencyFlag} provided without value`, t => {
+		execCli([concurrencyFlag, 'test.js'], {dirname: 'fixture/concurrency'}, (err, stdout, stderr) => {
+			t.is(err.code, 1);
+			t.match(stderr, 'The --concurrency and -c flags must be provided the maximum number of test files to run at once.');
```
yatharthk: @novemberborn This test doesn't seem to pass. Can you see any  fault in the `match()`?

test/cli.js
```diff
@@ -364,6 +364,25 @@ test('bails when config contains `"tap": true` and `"watch": true`', t => {
...
+			t.end();
+		});
+	});
+});
+
+['--concurrency', '-c'].forEach(concurrencyFlag => {
```
yatharthk: @novemberborn This tests for success if the value is provided for the flags. Looks good?

test/cli.js
novemberborn: What does `stderr` actually contain?

I had some trouble running this directly from the command line, it seemed to start running TAP tests instead. Try `execCli(['test.js', concurrencyFlag], …)` instead.

test/cli.js
```diff
@@ -364,6 +364,25 @@ test('bails when config contains `"tap": true` and `"watch": true`', t => {
 	});
 });
 
+['--concurrency', '-c'].forEach(concurrencyFlag => {
+	test(`bails when ${concurrencyFlag} provided without value`, t => {
+		execCli(['test.js', concurrencyFlag], {dirname: 'fixture/concurrency'}, (err, stdout, stderr) => {
```
yatharthk: @novemberborn Thanks. Tests pass now. :)

cc/ @sindresorhus

yatharthk:
@novemberborn Any progress on this?  cc/ @sindresorhus

sindresorhus:
Looks good. Thank you @yatharthk :)

yatharthk:
@sindresorhus @novemberborn Happy to contribute :)

Issue #1043, opened 2016-09-21
Revisit programmatic API


novemberborn:
AVA exposes an API through `require('ava/api')`. It's what other Node.js programs should use (e.g. Gulp & Grunt, #345). Currently it's undocumented (#284).

Let's provide a reliable, documented, SemVer-compliant & public API.

Feature wise it should be possible to run individual tests, test files, and start watch mode, as well as interrupt test runs (#710). Integrations should be able to collect test results, both as they happen and as individual tests finish (see also #141).

Integrations should be able to use AVA configuration in `package.json` files without duplicating code from the CLI (#700).

This is blocked until AVA's internals settle further, so let's start by figuring out what we want out of this API. This won't be there for a 1.0 release.

novemberborn:
While yes, we do want to support this, I don't think we need to keep this issue open. We'll get back to this eventually.

novemberborn:
@MKRhere no. I think it's more likely we'll require AVA to be run in a child process, but allow you to load your own extensions into that process.

MKRhere:
Thanks a lot for responding. I understand the reasons behind the decision, I think. I strongly needed an API-first framework, so I've built one in the meanwhile. It doesn't do parallelisation yet, of course :)

Thanks a lot! Overall a great job.

PR #1374, opened 2017-05-02
add options argument to cli.run

ahdinosaur:
hey :cat:

use case: when using `ava` as part of a [framework](https://dogstack.js.org) cli, i want to be able to pass in some configuration to the `ava` cli, so `ava` runs as the framework expects.

what do y'all think of a change like this?

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -20,8 +20,13 @@ const babelConfigHelper = require('./babel-config');
 // Bluebird specific
 Promise.longStackTraces();
 
-exports.run = () => {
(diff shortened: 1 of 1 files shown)


novemberborn:
We have an open issue for implementing a programmatic API: #1043. We're not ready to commit to that though. Perhaps you could share your use case on that issue?

ahdinosaur:
okay sweet, saw that issue, but don't have any specific requirements not already listed, basically just want to minimally wrap the existing cli and pass in a default configuration. thanks for you help @novemberborn.  :smiley_cat:

PR #1376, opened 2017-05-04
Use the `import-local` module

sindresorhus:
https://github.com/sindresorhus/import-local

I've used this method in a few other projects, so I thought I might as well make a module out of it.

I could use some feedback on the approach too: https://github.com/sindresorhus/import-local/blob/87a0c91a9a0ecea64471a95479b0366af296b989/index.js#L7-L10

I'm aware the existing test for this is failing. I was not sure how to properly mock this as I don't think you can use `proxyrequire` on sub-dependencies. @novemberborn Any suggestions?


novemberborn:
> I could use some feedback on the approach too

Approach seems good. I like how you take the relative path from the potential global directory, and then map it to a local import.

> I'm aware the existing test for this is failing. I was not sure how to properly mock this as I don't think you can use proxyrequire on sub-dependencies. @novemberborn Any suggestions?

We could stub `import-local` and see if the CLI code handles the return value correctly. Alternatively if you want more of an integration test I think you'll have to use fixtures to set up a "global" AVA and then see if it picks up the local version.

novemberborn:
I think if you create a fixture directory that contains a `node_modules/ava/cli.js` file then that file will be treated as the local file, assuming you `execCli()` with `dirname` set to that fixture directory. From inside the `cli.js` file you could just do `require('../../../../cli')` and it'll execute normally (I didn't count the `../` so this may not be accurate).

At this point running the normal `cli.js` from `execCli` gives you *global*, and with the correct `dirname` it should switch to *local*.

This should replace https://github.com/avajs/ava/blob/2ed94853ddd2ca4614a82b2a79cc251c09212331/test/cli.js#L408-L436.

sindresorhus:
@novemberborn Thanks for the tips. Should be good now.

PR #1382, opened 2017-05-12
Point to correct JavaScript file on Windows

mrchief:
`node_modules/.bin/ava.cmd` doesn't work on Windows (seen this mistake repeated in too many repos). The correct way is to reference the cli file under the module itself.


sindresorhus:
> seen this mistake repeated in too many repos

It's not a mistake. With your change, we're now relying on where the `cli.js` file is placed in the package. That could change. Although not likely since we control it.

Windows supports executing without extension as long as the extension is in `%PATHEXT%`, which `cmd` is, so this sounds more like a WebStorm issue.

sindresorhus:
See: https://youtrack.jetbrains.com/issue/WI-18109

// @develar

PR #1383, opened 2017-05-14
Add test setup recipe

### LINKED ISSUE #481: Recipe for using a plain function instead of `beforeEach` hook

JasonRitchie:
Fixes #481. I also took all the contributions from the issue and the previous PR #512 and tried to make a unified recipe. Thank you in advance for any feedback.


novemberborn:
Hi @JasonRitchie, thanks for opening this PR. Please bear with us until we can find the time to go through this and suggest any improvements.

novemberborn:
Hey @JasonRitchie, looks like I dropped the ball on this one. Thanks for writing this. I've gone through and made a few edits. Let me know if you have any thoughts on them. Otherwise I'll land this in a few days.

Thanks for making this PR, and I'm sorry it took so long to land.

JasonRitchie:
No worries, @novemberborn. Looks good to me!

PR #1401, opened 2017-06-04
Only forward `color` flag to worker if supplied by user

### LINKED ISSUE #1393: Remove custom CLI flags in child worker before requiring test file
We shouldn't expose custom CLI flags to the test file. It might change the behavior of user's code.

Currently we expose `--color` in the test file, even though the user just ran `$ ava` with no flag.

I had to implement a workaround for now: https://github.com/sindresorhus/xo-init/blob/9755f41134d0c8199674c1888d88cca3375358a3/test.js#L8

@novemberborn Thoughts?

kevva:
Fixes #1393.

Didn't bother to add tests since we already have ones in place for this [here](https://github.com/avajs/ava/blob/master/test/cli.js#L549-L563). Also removed a test that is moot now.

### DIFF
--- a/lib/fork.js
+++ b/lib/fork.js
@@ -36,7 +36,7 @@ module.exports = (file, opts, execArgv) => {
 		} : false
 	}, opts);
 
-	const args = [JSON.stringify(opts), opts.color ? '--color' : '--no-color'];
+	const args = [JSON.stringify(opts)].concat(process.argv.slice(2));
 
 	const ps = childProcess.fork(path.join(__dirname, 'test-worker.js'), args, {
 		cwd: opts.projectDir,
--- a/lib/fork.js
+++ b/lib/fork.js
@@ -4,6 +4,7 @@ const path = require('path');
 const fs = require('fs');
 const Promise = require('bluebird');
 const debug = require('debug')('ava');
+const hasFlag = require('has-flag');
 const AvaError = require('./ava-error');
 
 if (fs.realpathSync(__filename) !== __filename) {
@@ -36,7 +37,11 @@ module.exports = (file, opts, execArgv) => {
 		} : false
 	}, opts);
 
-	const args = [JSON.stringify(opts)].concat(process.argv.slice(2));
+	const args = [JSON.stringify(opts)];
+
+	if (hasFlag('--color') || hasFlag('--no-color')) {
+		args.push(`--color=${opts.color}`);
+	}
 
 	const ps = childProcess.fork(path.join(__dirname, 'test-worker.js'), args, {
 		cwd: opts.projectDir,
--- a/lib/fork.js
+++ b/lib/fork.js
@@ -44,7 +44,7 @@ module.exports = (file, opts, execArgv) => {
 	}
 
 	if (hasFlag('--no-color')) {
-		args.push(`--no-color`);
+		args.push('--no-color');
 	}
 
 	const ps = childProcess.fork(path.join(__dirname, 'test-worker.js'), args, {
--- a/lib/fork.js
+++ b/lib/fork.js
@@ -39,8 +39,12 @@ module.exports = (file, opts, execArgv) => {
 
 	const args = [JSON.stringify(opts)];
 
-	if (hasFlag('--color') || hasFlag('--no-color')) {
+	if (hasFlag('--color')) {
 		args.push(`--color=${opts.color}`);
+	}
+
+	if (hasFlag('--no-color')) {
+		args.push(`--no-color`);
 	}
 
 	const ps = childProcess.fork(path.join(__dirname, 'test-worker.js'), args, {
--- a/lib/test-worker.js
+++ b/lib/test-worker.js
@@ -1,9 +1,9 @@
 'use strict';
+const chalk = require('chalk');
 
 // Check if the test is being run without AVA cli
 {
 	const path = require('path');
-	const chalk = require('chalk');
 
 	const isForked = typeof process.send === 'function';
 	if (!isForked) {
@@ -26,6 +26,9 @@ const opts = adapter.opts;
 globals.options = opts;
 
 const serializeError = require('./serialize-error');
+
+// Initialize color support
+chalk.enabled = typeof opts.color === 'boolean' && opts.color;
 
 (opts.require || []).forEach(require);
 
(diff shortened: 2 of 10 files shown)


novemberborn:
`process.argv` is still a little different from what users might expect. Should we [splice out the options](https://github.com/avajs/ava/blob/c4c5e7bb3ec43726c03d9834f55e157c5feacf8f/lib/process-adapter.js#L29)?

I took https://github.com/avajs/ava/issues/1393#issuecomment-303965511 to read us having consensus on *only* forwarding `--color` and `--no-color`. Previously we've argued that users should use environment variables to pass runtime values to their tests. These flags don't sit well with that though.

Does it make sense to pass AVA's command line arguments? Should we instead look for a `--` argument and forward everything after that? But then does that imply users need to do `ava --color -- --color` if they want colors from AVA *and* whatever code runs in their tests? OTOH that would make it possible to do `ava --color -- --no-color`.

@sindresorhus?

sindresorhus:
> Should we splice out the options?

👍 

> I took #1393 (comment) to read us having consensus on only forwarding --color and --no-color. Previously we've argued that users should use environment variables to pass runtime values to their tests. These flags don't sit well with that though.

Yes, we should forward them if they're specified by the user. It's an exception since the flags apply to both us and other modules. `ava --color -- --color` would be super weird. I do not want that.

> OTOH that would make it possible to do ava --color -- --no-color.

I don't see why you would want to do that. Colors is a binary thing. Either you want them or you don't. I don't think we should complicate it more.

sindresorhus:
@kevva You need to handle the color option in the worker, by setting the Chalk singleton `.enabled` property to the same value as the `.enabled` property of Chalk in the CLI process. So the automatic color support is brought into the child worker, which defaults to no colors always because it's not a TTY.

duartealexf:
Is this going to be merged anytime soon? Please, I really need to pass some arguments to my tests :)

novemberborn:
@duartealexf the PR is not ready yet. If you could take @kevva's commit and iterate on it in a new PR then we could get it to a state where we can land it. I'm sure he won't mind.

kevva:
@novemberborn, I'm going to address your comments soon :).

novemberborn:
Sorry @duartealexf, @kevva is still on the case :)

kevva:
Ok, so I've changed the PR to only forward the `color` flag if it's set, and I'm also setting `chalk.enabled` in the worker.

Should we include support for different levels as in https://github.com/avajs/ava/pull/1455 too?

kevva:
Some tests are failing now because they relied upon checking the `ava` options forwarded as arguments before.

lib/fork.js
```diff
@@ -36,7 +37,11 @@ module.exports = (file, opts, execArgv) => {
 		} : false
 	}, opts);
 
-	const args = [JSON.stringify(opts), opts.color ? '--color' : '--no-color'];
+	const args = [JSON.stringify(opts)];
+
+	if (hasFlag('--color') || hasFlag('--no-color')) {
+		args.push(`--color=${opts.color}`);
```
novemberborn: Shouldn't this forward `--no-color` if specified? And forward `--color` exactly as it was received?

lib/test-worker.js
```diff
@@ -27,6 +27,9 @@ globals.options = opts;
 
 const serializeError = require('./serialize-error');
 
+// Initialize color support
+chalk.enabled = typeof opts.color === 'boolean' && opts.color;
```
novemberborn: `opts.color` already is a boolean, courtesy of https://github.com/avajs/ava/blob/1df502df8cba18f92afcbaed730d8c724687ca45/lib/cli.js#L78.

kevva: Yeah, I did this just to make [this](https://github.com/avajs/ava/blob/master/test/fork.js#L144) test pass since it doesn't run the CLI.

novemberborn: That's an odd test. Just remove it 😄

lib/fork.js
kevva: To forward them *exactly* like they were received requires some sort of parsing. We can forward `--no-color` instead of `--color=false` if it was specified though.

lib/fork.js
```diff
@@ -36,7 +37,15 @@ module.exports = (file, opts, execArgv) => {
...
 
-	const args = [JSON.stringify(opts), opts.color ? '--color' : '--no-color'];
+	const args = [JSON.stringify(opts)];
+
+	if (hasFlag('--color')) {
+		args.push(`--color=${opts.color}`);
+	}
+
+	if (hasFlag('--no-color')) {
+		args.push(`--no-color`);
```
novemberborn: Although XO doesn't complain, I'd have expected a regular string here, not a templated one.

kevva: Oh, escaped my eyes.

kevva:
> My take, if --color is set, it should be forwarded in exactly the same way. If it's not set, it shouldn't be forwarded.

Reason I didn't implement it that way is because it would've required some sort of command line arguments parsing, and it wouldn't look too good using `meow` in there. I extracted the command line parsing out of `meow` though, and created an option that I think would fix this use case https://github.com/kevva/argvments#any.

novemberborn:
> Reason I didn't implement it that way is because it would've required some sort of command line arguments parsing, and it wouldn't look too good using meow in there. I extracted the command line parsing out of meow though, and created an option that I think would fix this use case kevva/argvments#any.

Nice! So we could do the following:

||No flags|`--color`|`--color=mode`|`--no-color`|
|---|---|---|---|---|
|Set flag on worker process|No|Yes, as is|Yes, as is|Yes, as is|
|AVA configures `chalk` directly|Yes, with supported mode from main process|Yes, with supported mode from main process|No|No|

(I'm not sure now whether AVA configures `chalk` or another package.)

@kevva does this strike the right balance between forwarding color mode by default and not behaving against the user's intention when `--color` or `--no-color` are provided?

And do we need to handle `FORCE_COLOR` environment variables?

novemberborn:
@kevva any thoughts on my last comment?

@nowells does my proposal satisfy your concerns raised in #1455?

novemberborn:
#1700 clears up what arguments are exposed to the test worker. Let's discuss the specifics around configuring color in #1701.

Issue #1396, opened 2017-06-01
Allow passing in stacktrace to assertion functions


Alxandr:
## Description

I've created a fable wrapper for ava over at https://github.com/YoloDev/Fable.Ava, allowing people to use ava when testing their compiled-to-JS-F# code. This includes a wrapper on top of the ava api to make it more F# friendly, so that users can write code such as this (this is part of the code I use to test the bindings library itself):

```fsharp
prop {
  do! SyncSpec.plan 9

  let obj = createObj [ "test" ==> true; "kind" ==> "snapshot"; "with" ==> "ava"; "num" ==> 1 ]

  do! messageTwiceFails (Assert.snapshot obj)
  do! Assert.isDeepEqual (Assert.snapshot obj) (Assert.Snapshot (obj, None))
  do! Assert.isDeepEqual (Assert.snapshot obj <?> "foo") (Assert.Snapshot (obj, Some "foo"))
  do! mockCall (Assert.snapshot obj) (AsserterMock.snapshot (obj, None))
  do! mockCall (Assert.snapshot obj <?> "foo") (AsserterMock.snapshot (obj, Some "foo"))
  
  do! Assert.snapshot obj
  do! Assert.snapshot obj <?> "snapshot with name"
} |> addTest "snapshot"
```

This works really well, but there is one problem, when a test fail the error printed to the console is completely useless. For instance, the following code

```fsharp
prop {
  do! SyncSpec.plan 1

  do! Assert.isTrue false <?> "true is false"
} |> addTest "failing"
```

yields the following console output:

![image](https://cloud.githubusercontent.com/assets/112334/26704388/0020970c-472f-11e7-9caf-f9b4994d0c62.png)

Now the console output is gorgeous in AVA, using the sourcemaps to look up the F#, but it's showing the wrong line in the wrong file, which makes it completely unhelpful. Therefore I suggest adding a new parameter to the assertion methods which is the stacktrace at the point where the assertion method was called. This is more or less what is done for async tests already (except that the stack traces are collected by ava itself for instance [here](https://github.com/avajs/ava/blob/master/lib/assert.js#L176)), but would allow consumers of the library to take advantage of this. I would be willing to look at writing a PR during the summer if my suggestion is accepted.

## Environment

```
Node.js v7.10.0
darwin 16.6.0
ava 0.24.5
npm 4.2.0
```

novemberborn:
[...]

PR #1402, opened 2017-06-04
Add support for explicit callsites in asserts

Alxandr:
See #1396. I've not been able to test that it works with snapshots, because I don't really understand how the snapshoting works, otherwize there is updated tests for all asserts. It's not been decided whether or not this is a feature ava want's to support, but I figured it's better to have an actual implementation to talk about. I will likely be using this fork in my Fable bindings.

Currently there are two tests failing locally for me, but they were failing in master as well locally.

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -33,18 +33,18 @@ class AssertionError extends Error {
 exports.AssertionError = AssertionError;
 
 class CallSite {
-	constructor() {
+	constructor(caller) {
 		// This is not lazy in node. However, using pure `Error` instances might not be
 		// backwards compatible (nor will it allow us to remove unwanted frames easily).
-		Error.captureStackTrace(this, CallSite);
+		Error.captureStackTrace(this, caller || CallSite);
 	}
 
 	toString() {
 		return '[[CallSite]]';
 	}
 
-	getStack(target = '') {
-		return this.stack.replace('[[CallSite]]', target);
+	getStack(target) {
+		return this.stack.replace('[[CallSite]]', target || '');
 	}
 }
 exports.CallSite = CallSite;
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -22,18 +22,32 @@ class AssertionError extends Error {
 		// Reserved for power-assert statements
 		this.statements = [];
 
+		// To keep compatibility with old API, don't know if needed
 		if (opts.stack) {
 			this.stack = opts.stack;
+		} else if (opts.callSite) {
+			this.stack = opts.callSite.getStack(this.toString());
 		}
 	}
 }
 exports.AssertionError = AssertionError;
 
-function getStack() {
-	const obj = {};
-	Error.captureStackTrace(obj, getStack);
-	return obj.stack;
+class CallSite {
+	constructor() {
+		// This is not lazy in node. However, using pure `Error` instances might not be
+		// backwards compatible (nor will it allow us to remove unwanted frames easily).
+		Error.captureStackTrace(this, CallSite);
+	}
+
+	toString() {
+		return '[[CallSite]]';
+	}
+
+	getStack(target = '') {
+		return this.stack.replace('[[CallSite]]', target);
+	}
 }
+exports.CallSite = CallSite;
 
 function wrapAssertions(callbacks) {
 	const pass = callbacks.pass;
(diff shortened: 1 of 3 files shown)


Alxandr:
Btw; a bonus of the abstraction of the `CallSite` class, is that the internals can be changed once standard stacktraces lands in ecmascript/node (see https://github.com/tc39/proposal-error-stacks). It can also be made lazy etc., without the callers having to know.

lib/assert.js
```diff
@@ -22,18 +22,32 @@ class AssertionError extends Error {
 		// Reserved for power-assert statements
 		this.statements = [];
 
+		// To keep compatibility with old API, don't know if needed
```
novemberborn: This constructor is private, so we should be able to clean this up.

lib/assert.js
```diff
@@ -51,14 +65,15 @@ function wrapAssertions(callbacks) {
 			pass(this);
 		},
 
-		fail(message) {
+		fail(message, {callSite} = {}) {
```
novemberborn: We're targeting Node.js 4 so this syntax is a bit too fancy I'm afraid 😉

lib/main.js
```diff
@@ -96,6 +97,7 @@ globals.setImmediate(() => {
 });
 
 module.exports = runner.chain;
+module.exports.CallSite = CallSite;
```
novemberborn: I think we should export this as `require("ava/CallSite")`.

lib/assert.js
```diff
@@ -22,18 +22,32 @@ class AssertionError extends Error {
...
+		// backwards compatible (nor will it allow us to remove unwanted frames easily).
+		Error.captureStackTrace(this, caller || CallSite);
+	}
+
+	toString() {
+		return '[[CallSite]]';
+	}
+
+	getStack(target) {
+		return this.stack.replace('[[CallSite]]', target || '');
```
novemberborn: Is the assumption that `[[CallSite]]` comes from the `toString()` method above? Could there be more than one of those in the stack? Why is it OK to replace it with an empty string?

Alxandr: You're correct. When you call `Error.captureStackTrace(obj, fn)`, it will produce `obj.toString()\n<stack trace>`. I simply inserted a token that should not happen during any stacktrace so that it could be replaced at a later time.

lib/assert.js
Alxandr: Right. Well, that should be easy enough to fix.

lib/main.js
Alxandr: That would require putting it as it's own file in the directory root though, next to `index` and `cli`, wouldn't it?

Alxandr:
The `CallSite` abstraction is just a simple abstraction for just that, call sites (meaning "where a function was executed"). There were two points for this, first was being open to the new stacktrace API without having to change the public API, the second was not parsing the stack myself. When you do `Error.captureStackTrace(obj)`, it will insert `obj.toString()` first into the stacktrace, on its own line, therefore I just made `toString()` return a replace token. The API is then `callSite.getStack(errorMessage)`, so to get the exact same result you would get from a custom `Error` object, you do `callSite.getStack(myError.toString())`.

The tap library you linked seems to deal with stack manipulation (I've never seen it), but just because both talk about callsites, doesn't mean they have anything to do with eachother. Callsite is a common name for a data type that refers to where a function was invoked.

Typically, I would generate code like this in `Fable.Ava`:

```js
const assertion = Assert.isDeepEqual(actual, expected, new CallSite());
asserter.assert(assertion);
```

However, since this is F#, I automatically insert the last parameter, so the user doesn't know anything about the callsite unless he calls the overload that explicitly passes one. The F# is still kind of at the experimental stage though (not public anywhere yet).

If I was creating this library to use in JS, I would change it to the following:

```js
const isDeepEqual = (actual, expected) => ({
  type: 'deepEqual',
  actual,
  expected,
  callSite: new CallSite(isDeepEqual)
});
```

But this has proven slightly problematic in F#, but I'm still working at it :).

lib/main.js
novemberborn: There could be a `CallSite.js` which reexports `CallSite` from `lib/assert.js`. Manipulating `runner.chain` like you do here doesn't seem ideal.

novemberborn:
> The CallSite abstraction is just a simple abstraction for just that, call sites (meaning "where a function was executed").

Makes sense. Let's see what the rest of @avajs/core has to say on this 😄

lib/main.js
Alxandr: Fair enough. I'm in the middle of moving though, so it might be a week or so till I have the time to look at this again. By then maybe there's more feedback :)

novemberborn: No worries. Good luck with the move!

novemberborn:
Closing this, but note that #1396 is still open 😄

PR #1407, opened 2017-06-10
Version warning when local version is behind

### LINKED ISSUE #1336: Version warning when only local version is behind.
If the global version is up to date, but the local version is behind, you still get an error message telling you to update, but it recommends using the `-g` flag.

tdeschryver:
Fixes #1336 

Added variable `process.env.AVA_LOCAL_CLI` to show the correct message.

### DIFF
--- a/package.json
+++ b/package.json
@@ -173,7 +173,7 @@
 		"time-require": "^0.1.2",
 		"trim-off-newlines": "^1.0.1",
 		"unique-temp-dir": "^1.0.0",
-		"update-notifier": "^2.1.0"
+		"update-notifier": "^2.2.0"
 	},
 	"devDependencies": {
 		"cli-table2": "^0.2.0",
--- a/package.json
+++ b/package.json
@@ -173,7 +173,7 @@
 		"time-require": "^0.1.2",
 		"trim-off-newlines": "^1.0.1",
 		"unique-temp-dir": "^1.0.0",
-		"update-notifier": "^2.2.0"
+		"update-notifier": "^2.3.0"
 	},
 	"devDependencies": {
 		"cli-table2": "^0.2.0",


cli.js
```diff
@@ -11,7 +11,8 @@ const localCLI = resolveCwd('ava/cli');
 // Use `path.relative()` to detect local AVA installation,
 // because __filename's case is inconsistent on Windows
 // see https://github.com/nodejs/node/issues/6624
-if (localCLI && path.relative(localCLI, __filename) !== '') {
+process.env.AVA_LOCAL_CLI = process.env.AVA_LOCAL_CLI || (localCLI && path.relative(localCLI, __filename) !== '');
```
novemberborn: This bit `(localCLI && path.relative(localCLI, __filename) !== '')` needs to be cast to a string: `String(localCLI && path.relative(localCLI, __filename) !== '')`. Otherwise the comparison on the next line won't be true.

lib/cli.js
```diff
@@ -100,7 +100,7 @@ exports.run = () => {
 		}
 	});
 
-	updateNotifier({pkg: cli.pkg}).notify();
+	updateNotifier({pkg: cli.pkg}).notify({isGlobal: process.env.AVA_LOCAL_CLI === 'true'});
```
novemberborn: Shouldn't this be `!== 'true'`?

tdeschryver: Yep, you're right. My bad!

cli.js
tdeschryver: Oh I didn't know that. When I was debugging it, it was a automatically converted to a string.
Its changed now.

tdeschryver:
@novemberborn I was trying to figure out how I could test the `isGlobal` value and to be honest I'm kinda stuck. The only way to test it would be to call `lib/cli` directly  and stub almost everything (`update-notifier`, `pkg-conf`, `api`,  ...), and put a test inside the `.notify()` function?

novemberborn:
@tdeschryver in your current test, try this:

```js
const notifySpy = sinon.spy()

proxyquire('../cli', {
  debug: debugStub,
  'resolve-cwd': resolveCwdStub,
  './lib/cli': proxyquire('../lib/cli', {
    'update-notifier' () {
      return {notify: notifySpy}
    }
  })
});
```

You should then be able to assert on the `notifySpy` arguments.

tdeschryver:
@novemberborn I don't think that will work (or I am missing something), the problem is we are stubbing `resolveCwd` with `/fixture/empty`, thus the `lib/cli` is never being called. I tried creating a new test where `lib/cli` is being called, the problem here is that it will test some files and to ignore this we'll have to stub a lot of things I think.

novemberborn:
@tdeschryver could you push your attempt? I'll try on my end.

tdeschryver:
@novemberborn I discarded my changes, would it help you if I rewrite it?

novemberborn:
@tdeschryver yea I'll try and give this a go. #1376 landing also threw a wrench into the works.

novemberborn:
@sindresorhus `import-local` breaks the approach taken in this PR. We need to set `AVA_LOCAL_CLI` *before* the local CLI module is required. `import-local` requires the module immediately. Perhaps it could export an `existsSync` method?

sindresorhus:
I think you're trying to solve this problem in the wrong place. Better to fix it in `update-notifier` so it's fixed for everyone. See: https://github.com/yeoman/update-notifier/issues/112

tdeschryver:
@sindresorhus is it OK if I'm going to take a look at it?

sindresorhus:
@tdeschryver Go ahead 😃

tdeschryver:
Bumped `update-notifier` since [yeoman/update-notifier#114](https://github.com/yeoman/update-notifier/pull/114) is merged.

sindresorhus:
Thanks @tdeschryver :)

PR #1408, opened 2017-06-10
Output message when `t.throws()` fails

### LINKED ISSUE #1342: Output from `t.throws()` failing is missing a message
With latest master (ecdadfa9c2b5d5c40e4edd14df0319f222007c60).

```
import test from 'ava';

const foo = () => Promise.resolve('foo');

test(async t => {
	await t.throws(foo());
});
```

Outputs:

```
  1 failed

  [anonymous]
  /Users/sindresorhus/dev/private/ava-playground/test.js:6

   5: test(async t => {
   6:   await t.throws(foo());
   7: });

  Test.<anonymous> (test.js:6:10)
  step (test.js:9:191)
  Test.<anonymous> (test.js:9:99)
  Test.fn (test.js:5:1)
```

tdeschryver:
Fixes #1342

Maybe the test can be written better?

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -41,7 +41,6 @@ function wrapAssertions(callbacks) {
 	const fail = callbacks.fail;
 
 	const noop = () => {};
-	const makeNoop = () => noop;
 	const makeRethrow = reason => () => {
 		throw reason;
 	};
@@ -119,6 +118,7 @@ function wrapAssertions(callbacks) {
 		},
 
 		throws(fn, err, message) {
+			const assertion = 'throws';
 			let promise;
 			if (isPromise(fn)) {
 				promise = fn;
@@ -126,7 +126,7 @@ function wrapAssertions(callbacks) {
 				promise = observableToPromise(fn);
 			} else if (typeof fn !== 'function') {
 				fail(this, new AssertionError({
-					assertion: 'throws',
+					assertion,
 					improperUsage: true,
 					message: '`t.throws()` must be called with a function, Promise, or Observable',
 					values: [formatAssertError.formatWithLabel('Called with:', fn)]
@@ -174,10 +174,16 @@ function wrapAssertions(callbacks) {
 			if (promise) {
 				// Record stack before it gets lost in the promise chain.
 				const stack = getStack();
-				const intermediate = promise.then(makeNoop, makeRethrow).then(fn => test(fn, stack));
+				const intermediate = promise.then(value => {
+					throw new AssertionError({
+						assertion,
+						message: 'Expected promise to be rejected, but it was resolved instead',
+						values: [formatAssertError.formatWithLabel('Resolved with:', value)]
+					});
+				}, reason => test(makeRethrow(reason), stack));
+
 				pending(this, intermediate);
-				// Don't reject the returned promise, even if the assertion fails.
-				return intermediate.catch(noop);
+				return intermediate;
 			}
 
 			try {
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -118,7 +118,6 @@ function wrapAssertions(callbacks) {
 		},
 
 		throws(fn, err, message) {
-			const assertion = 'throws';
 			let promise;
 			if (isPromise(fn)) {
 				promise = fn;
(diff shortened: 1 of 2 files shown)


lib/assert.js
```diff
@@ -174,10 +174,16 @@ function wrapAssertions(callbacks) {
...
+						assertion,
+						message: 'Expected promise to be rejected, but it was resolved instead',
+						values: [formatAssertError.formatWithLabel('Resolved with:', value)]
+					});
+				}, reason => test(makeRethrow(reason), stack));
+
 				pending(this, intermediate);
-				// Don't reject the returned promise, even if the assertion fails.
-				return intermediate.catch(noop);
+				return intermediate;
```
tdeschryver: I had to remove the `catch` because otherwise I couldn't access the `AssertionError` in the test.

lib/assert.js
```diff
@@ -119,14 +118,15 @@ function wrapAssertions(callbacks) {
...
+			const assertion = 'throws';
 			let promise;
 			if (isPromise(fn)) {
 				promise = fn;
 			} else if (isObservable(fn)) {
 				promise = observableToPromise(fn);
 			} else if (typeof fn !== 'function') {
 				fail(this, new AssertionError({
-					assertion: 'throws',
+					assertion,
```
novemberborn: I think it's clearer to just repeat this property.

lib/assert.js
novemberborn: The `catch` is necessary. We don't expose the `AssertionError` instance to the test author. Instead add a `.catch()` callback in the `pending()` implementation here: <https://github.com/tdeschryver/ava/blob/4a19a96da89a1533025295a7cd78b74c12528f04/test/assert.js#L15>.

test/assert.js
```diff
@@ -648,6 +652,17 @@ test('.throws() fails if passed a bad value', t => {
 	t.end();
 });
 
+test('promise .throws() fails when promise is resolved', t => {
+	return assertions.throws(Promise.resolve('foo'))
+	.catch(err => {
+		testFailResults(t, err, {
```
novemberborn: Rather than extracting the comparison logic into `testFailResults`, if you capture `lastFailure` through the `pending()` function you could write an `eventuallyFailsWith()`:

```js
function eventuallyFailsWith(t, promise, subset) {
  return promise.then(() => failsWith(t, () => {}, subset))
}
```

Given that `assertion.throws()` should return a promise that always resolves, and the error will be captured separately.

tdeschryver: Thanks for the tip and reviews!
I got it to work now :), but I had to pass the `lastFailure` to the `failsWith` function otherwise it would be assigned to `null`. Is this OK?

novemberborn: My bad, overlooked that `failsWith()` resets `lastFailure`.

Instead extract the failure testing code into its own function, e.g.:

```js
function assertFailure() {
  if (!lastFailure) {
  		t.fail('Expected assertion to fail');
 		return;
 	}
 
 	t.is(lastFailure.assertion, subset.assertion);
 	t.is(lastFailure.message, subset.message);
 	t.is(lastFailure.name, 'AssertionError');
 	t.is(lastFailure.operator, subset.operator);
 	if (subset.statements) {
 		t.is(lastFailure.statements.length, subset.statements.length);
 		lastFailure.statements.forEach((s, i) => {
 			t.is(s[0], subset.statements[i][0]);
 			t.match(s[1], subset.statements[i][1]);
 		});
 	} else {
 		t.same(lastFailure.statements, []);
 	}
 	if (subset.values) {
 		t.is(lastFailure.values.length, subset.values.length);
 		lastFailure.values.forEach((s, i) => {
 			t.is(s.label, subset.values[i].label);
 			t.match(s.formatted, subset.values[i].formatted);
 		});
 	} else {
 		t.same(lastFailure.values, []);
  	}
  }
}
```

And call that from `failsWith()` and `eventuallyFailsWith()` (so the latter doesn't call `failsWith()` anymore). `eventuallyFailsWith()` should also reset `lastFailure`.

tdeschryver: That's indeed a cleaner solution.
Changes are commited!

sindresorhus:
Thank you @tdeschryver :)

PR #1410, opened 2017-06-10
Set absolute output path in Webpack config used for precompiling

bapjiws:
Using `output.path: '_build'` (tested in Webpack v2.5.1) will lead to:
```
Invalid configuration object. Webpack has been initialised using a configuration object that does not match the API schema.
 - configuration.output.path: The provided value "_build" is not an absolute path!
```


novemberborn:
Hey @bapjiws sorry for the delay in getting back to you. Thanks for this PR. As it turns out this was previously spotted in #1385 which I just landed. There's many other improvements to the webpack recipe in that PR, too. I hope you'll enjoy it.

PR #1413, opened 2017-06-12
Add command for updating snapshots in watch mode

### LINKED ISSUE #1387: Press 'u' in watch mode to update snapshots
Ava borrows Jest's snapshot feature, and from #1275 and other issues, this feature is evolving in Ava. That's really awesome.

Jest, in addition, has an interactive watcher that:

- Runs only tests related to files that have changed
- Replaces previous console output of passing tests
- Provides commands to interact with the running watcher

One of the commands is `u`, which updates snapshots of tests that just ran.

I became familiar with Ava thru @novemberborn when working on `babel-plugin-import-glob`. I really like Ava, especially with the TAP-like syntax.

Ava just needs an interactive watcher to compliment the snapshot feature.

efegurkan:
This is my humble implementation for pressing 'u' in watch mode to update snapshots.

fixes #1387

### DIFF
--- a/docs/recipes/watch-mode.md
+++ b/docs/recipes/watch-mode.md
@@ -91,6 +91,10 @@ If you run AVA in your CI with watch mode, the execution will exit with an error
 
 You can quickly rerun all tests by typing <kbd>r</kbd> on the console, followed by <kbd>Enter</kbd>.
 
+## Updating snapshots
+
+You can update failing snapshots by typing <kbd>u</kbd> on the console, followed by <kbd>Enter</kbd>.
+
 ## Debugging
 
 Sometimes watch mode does something surprising like rerunning all tests when you thought only a single test would be run. To see its reasoning you can enable a debug mode. This will work best with the verbose reporter:
--- a/lib/watcher.js
+++ b/lib/watcher.js
@@ -84,7 +84,8 @@ class Watcher {
 
 		this.clearLogOnNextRun = true;
 		this.runVector = 0;
-		this.run = specificFiles => {
+		this.run = (specificFiles, options) => {
+			const opts = options || {};
 			if (this.runVector > 0) {
 				const cleared = this.clearLogOnNextRun && logger.clear();
 				if (!cleared) {
@@ -117,7 +118,7 @@ class Watcher {
 			}
 
 			this.touchedFiles.clear();
-			this.busy = api.run(specificFiles || files, {runOnlyExclusive})
+			this.busy = api.run(specificFiles || files, {runOnlyExclusive, updateSnapshots: opts.updateSnapshots || false})
 				.then(runStatus => {
 					runStatus.previousFailCount = this.sumPreviousFailures(currentVector);
 					logger.finish(runStatus);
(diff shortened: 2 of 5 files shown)


lib/watcher.js
```diff
@@ -79,7 +79,7 @@ class Watcher {
 
 		this.clearLogOnNextRun = true;
 		this.runVector = 0;
-		this.run = specificFiles => {
+		this.run = (specificFiles, options = {}) => {
```
sindresorhus: You can't use default arguments. We still support Node.js 4.

test/watcher.js
```diff
@@ -212,7 +212,7 @@ group('chokidar', (beforeEach, test, group) => {
 		t.ok(logger.reset.notCalled);
 		t.ok(logger.start.notCalled);
 		t.ok(api.run.calledOnce);
-		t.strictDeepEqual(api.run.firstCall.args, [files, {runOnlyExclusive: false}]);
+		t.strictDeepEqual(api.run.firstCall.args, [files, {runOnlyExclusive: false, updateSnapshots: false}]);
```
sindresorhus: I would extract `{runOnlyExclusive: false, updateSnapshots: false}` into a variable and use it in all these asserts. Would make it easier to add additional options in the future.

sindresorhus:
The feature makes sense to me :) @avajs/core Thoughts?

It also needs to be documented.

lib/watcher.js
efegurkan: that explains the failing tests indeed 👍

docs/recipes/watch-mode.md
```diff
@@ -91,6 +91,10 @@ If you run AVA in your CI with watch mode, the execution will exit with an error
 
 You can quickly rerun all tests by typing <kbd>r</kbd> on the console, followed by <kbd>Enter</kbd>.
 
+## Updating snapshots
+
+You can update failing snapshots by typing<kbd>u</kbd> on the console, followed by <kbd>Enter</kbd>.
```
novemberborn: Needs a space between `typing` and `<kbd>`.

lib/watcher.js
```diff
@@ -267,14 +268,22 @@ class Watcher {
...
 			});
 		});
 	}
 	rerunAll() {
 		this.dirtyStates = {};
 		this.run();
 	}
+	rerunWithUpdate() {
+		this.dirtyStates = {};
+		this.run(undefined, {updateSnapshots: true});
```
novemberborn: Should this update snapshots in *all* tests, or just those tests that just ran? I'm leaning towards the latter. Do you know what Jest settled on?

@lukescott?

lukescott: I believe it only updates snapshots of tests that were just ran.

efegurkan: for jest it updates only for running tests.

docs/recipes/watch-mode.md
efegurkan: oops 🤷‍

lib/watcher.js
novemberborn: Let's do that here as well then. Storing the [`specificFiles || files` argument here](https://github.com/efegurkan/ava/blob/08675e83375781fcf9468ce30d95e1e84f87f7d4/lib/watcher.js#L115) should be sufficient.

PR #1414, opened 2017-06-14
Fix debugging with Webstorm on Node.js 8

dohomi:
## Description
On Node 8 Webstorm debug is not working any longer. The reason is, that the `--inspect-brk` is not handled correctly inside of the cli.js file

```
Starting inspector on 127.0.0.1:53733 failed: address already in use
```
### Fix
Adding --inspect-brk to api.js

## Environment
Node v8.1, latest AVAJS

### DIFF
--- a/api.js
+++ b/api.js
@@ -173,7 +173,7 @@ class Api extends EventEmitter {
 
 		// --inspect-brk is used in addition to --inspect to break on first line and wait
 		execArgv.some((arg, index) => {
-			const isDebugArg = arg === '--inspect' || arg === '--inspect-brk' || arg.indexOf('--inspect=') === 0 || arg.indexOf('--inspect-brk=') === 0;
+			const isDebugArg = /^--inspect(-brk)?($|=)/.test(arg);
 			if (isDebugArg) {
 				debugArgIndex = index;
 			}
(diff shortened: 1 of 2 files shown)


novemberborn:
Hey @dohomi, this looks good!

I pushed a commit which replaces these comparisons with a regular expression instead. That led me to trying to understand the existing API tests which seem to be limited and a bit confusing (notably `--inspect-brk` and `--debug-brk` aren't tested). So there's some commits that add / fix comments for that too.

Could you check whether `--inspect-brk` still works for you?

@sindresorhus happy with the regexes?

sindresorhus:
> happy with the regexes?

👍

dohomi:
@novemberborn yes I think this is better approach 👍

api.js
```diff
@@ -171,9 +171,9 @@ class Api extends EventEmitter {
 		const execArgv = this.options.testOnlyExecArgv || process.execArgv;
 		let debugArgIndex = -1;
 
-		// --debug-brk is used in addition to --inspect to break on first line and wait
+		// --inspect-brk is used in addition to --inspect to break on first line and wait
 		execArgv.some((arg, index) => {
-			const isDebugArg = arg === '--inspect' || arg.indexOf('--inspect=') === 0;
+			const isDebugArg = /^--inspect(-brk)?($|=)/.test(arg);
```
willmendesneto: There's no `--inspect-brk`. Could you update t=your regex removing that validation, please?

novemberborn: It's there in Node.js 8: https://nodejs.org/api/debugger.html#debugger_v8_inspector_integration_for_node_js

novemberborn:
Thanks @dohomi!

dohomi:
Thanks for merging

Issue #849, opened 2016-05-21
Improve recipe for testing endpoints


sindresorhus:
We already have the [Endpoint testing](https://github.com/avajs/ava/blob/master/docs/recipes/endpoint-testing.md) recipe, but it's not very comprehensive. Testing endpoints is such a common tasks, I think we should expand on it.

Like using [`nock`](https://github.com/node-nock/nock)/[`chakram`](https://github.com/dareid/chakram), testing against a real end-point (and how to handle concurrency), etc.

silvenon:
I could expand it with nock testing, I'm just not sure whether [cleanAll](https://github.com/node-nock/nock#cleanall) is needed when testing with AVA. I guess I'm still unclear about [process isolation](https://github.com/avajs/ava#process-isolation) and if it's applicable to this situation.

jamestalmage:
You wouldn't need `cleanAll` in an `after` hook. The process is going to be killed, so there's no good reason to clean up mocks at the end. `after` cleanup is only necessary for persistent state (i.e. disk state). 

You might want `cleanAll` in an `afterEach` if all your tests were `serial`.

-- 

When it comes to testing real endpoints and handling concurrency, you have a few options. 

Let's assume a simple social app. You have a "friends list", you can add and remove people from the list, and you can get recommendations for "friends of friends".

1. Avoid concurrency all together, and just use `serial` tests, and a clean application state each time. This avoids any interference and ensures no test relies on state left over from a previous one.

2. Allow concurrency, but make sure you use disjoint operations that do not interfere with each other. In the social app described above, this would mean creating all new user accounts for every test (using randomized strings for account names or something). You could allow all these tests to run against the same server at once because there should be no data overlap. None of your friends in the current test are friends with any accounts created in other tests, so you shouldn't be getting any inter-test interference in the "recommendations". There is some risk here of creating inter test dependencies if you aren't careful, but if done correctly it also stresses your production code in a way that `1` does not.

3. Launch a new app server for every test, using `get-port` to find open ports. This is probably only advisable for lightweight apps that can be spun up and down easily.

jamestalmage:
[`download`](https://github.com/kevva/download/blob/master/test.js#L14-L24) seems to have a decent test suite. We could use that as a template for our `nock` recipe. (note that doesn't really test an endpoint as much as mock one).

[...]

PR #1420, opened 2017-06-23
Add docs for endpoint testing with mongoose 

zellwk:
Docs for endpoint tests. #849. 

Let me know if I'm opening the PR correctly?

### DIFF
--- a/docs/recipes/endpoint-testing-with-mongoose.md
+++ b/docs/recipes/endpoint-testing-with-mongoose.md
@@ -53,7 +53,7 @@ Next start the in-memory MongoDB instance and connect to Mongoose:
 const mongod = new MongodbMemoryServer()
 
 // Create connection to Mongoose before tests are run
-test.before(async t => {
+test.before(async () => {
 	const uri = await mongod.getConnectionString();
 	await mongoose.connect(uri, {useMongoClient: true});
 });
@@ -64,7 +64,7 @@ When you run your first test, MongoDB downloads the latest MongoDB binaries. The
 You'll want to populate your database with dummy data. Here's an example:
 
 ```js
-test.beforeEach(async t => {
+test.beforeEach(async () => {
 	const user = new User({
 		email: 'one@example.com',
 		name: 'One'
@@ -76,7 +76,7 @@ test.beforeEach(async t => {
 Dummy data should be cleared after each test:
 
 ```js
-test.afterEach.always(async t => await User.remove());
+test.afterEach.always(async () => await User.remove());
 ```
 
 Now you can use SuperTest to send off a request for your app endpoint. Use AVA for your assertions:
@@ -114,7 +114,7 @@ test.serial('litmus create user', async t => {
 Finally disconnect from and stop MongoDB when all tests are done:
 
 ```js
-test.after.always(async t => {
+test.after.always(async () => {
 	mongoose.disconnect()
 	mongod.stop()
 })
--- a/docs/recipes/endpoint-testing-with-mongoose.md
+++ b/docs/recipes/endpoint-testing-with-mongoose.md
@@ -6,9 +6,9 @@ This recipe shows you how to test your endpoints with AVA and Mongoose, assuming
 
 This recipe uses the following libraries:
 
-1. [MongoDB memory server](https://github.com/nodkz/mongodb-memory-server) (A MongoDB in-memory Server)
-2. [Supertest](https://github.com/visionmedia/supertest) (An endpoint testing library)
-3. [Mongoose](http://mongoosejs.com)
+1. [`mongod-memory-server`](https://github.com/nodkz/mongodb-memory-server) (A MongoDB in-memory Server)
+2. [SuperTest](https://github.com/visionmedia/supertest) (An endpoint testing library)
+3. [Mongoose ODM](http://mongoosejs.com)
 
 Install the first two libraries by running the following code:
 
@@ -24,9 +24,15 @@ You should have Mongoose installed already. If not, run the following code to in
 $ npm install mongoose
 ```
 
+## Prerequisites
+
+You'll need a server file and a Mongoose model. See the [`server.js`](https://github.com/zellwk/ava-mdb-test/blob/master/server.js) and [`models/User.js`](https://github.com/zellwk/ava-mdb-test/blob/master/models/User.js) examples.
+
+Note that `server.js` does not start the app. Instead this must be done by SuperTest, so that the app endpoints can be tested. If you're using Express for your application, make sure you have a startup file that imports `app` and calls `app.listen()`.
+
 ## Your test file
 
-First, include the libraries you need.
+First, include the libraries you need:
 
 ```js
 // Libraries required for testing
@@ -40,94 +46,77 @@ import app from '../server'
 import User from '../models/User'
 ```
 
-### Your server file
-
-When you test endpoints, you don't want to start your app. Supertest does this for us automatically when you pass your app into it.
-
-So, if you're using express for your application, make sure you have a startup file that imports `app` and calls `app.listen`
-
-Here's an example of the [server file](https://github.com/zellwk/ava-mdb-test/blob/master/server.js)
-
-And here's an example of your Mongoose [model](https://github.com/zellwk/ava-mdb-test/blob/master/models/User.js)
-
-### Back to your test file
-
-**First, start your MongoDB instance and connect to Mongoose:**
+Next start the in-memory MongoDB instance and connect to Mongoose:
 
 ```js
-// Start MongoDB Instance
+// Start MongoDB instance
 const mongod = new MongodbMemoryServer()
 
-// Create connection to mongoose before all tests
-test.before(async t => mongoose.connect(await mongod.getConnectionString(), { useMongoClient: true }))
+// Create connection to Mongoose before tests are run
+test.before(async t => {
+	const uri = await mongod.getConnectionString();
+	await mongoose.connect(uri, {useMongoClient: true});
+});
 ```
 
-When you run your first test, MongoDB downloads the latest MongoDB Binaries. It may take a minute. (The download is ~70mb).
-
-**Add fixtures for each test**
+When you run your first test, MongoDB downloads the latest MongoDB binaries. The download is ~70MB so this may take a minute.
 
 You'll want to populate your database with dummy data. Here's an example:
 
 ```js
 test.beforeEach(async t => {
-  const user = new User({
-  	email: 'one@example.com',
-  	name: 'One'
-  })
-  await user.save()
-})
+	const user = new User({
+		email: 'one@example.com',
+		name: 'One'
+	});
+	await user.save();
+});
 ```
 
-**Clear your dummy data after each test**:
+Dummy data should be cleared after each test:
 
 ```js
-// Cleans up database after every test
-test.afterEach.always(async t => await User.remove())
+test.afterEach.always(async t => await User.remove());
 ```
 
-**Write your tests**
-
-Use Supertest to fire a request for your endpoint. Then, do the rest with AVA normally.
+Now you can use SuperTest to send off a request for your app endpoint. Use AVA for your assertions:
 
 ```js
-// First test
-// Note: tests must be serial tests.
-// It is NOT RECOMMENDED to run parallel tests within an AVA test file when using Mongoose (see why below)
+// Note that the tests are run serially. See below as to why.
+
 test.serial('litmus get user', async t => {
-  const { app } = t.context
-  const res = await request(app)
-    .get('/litmus')
-    .send({email: 'one@example.com'})
-  t.is(res.status, 200)
-  t.is(res.body.name, 'One')
-})
+	const {app} = t.context;
+	const res = await request(app)
+		.get('/litmus')
+		.send({email: 'one@example.com'});
+	t.is(res.status, 200);
+	t.is(res.body.name, 'One');
+});
 
-// Second test
 test.serial('litmus create user', async t => {
-  const { app } = t.context
-  const res = await request(app)
-    .post('/litmus')
-    .send({
-      email: 'new@example.com',
-      name: 'New name'
-    })
+	const {app} = t.context;
+	const res = await request(app)
+		.post('/litmus')
+		.send({
+			email: 'new@example.com',
+			name: 'New name'
+		});
 
-  t.is(res.status, 200)
-  t.is(res.body.name, 'New name')
+	t.is(res.status, 200);
+	t.is(res.body.name, 'New name');
 
-  // Verifies that user is created in DB
-  const newUser = await User.findOne({email: 'new@example.com'})
-  t.is(newUser.name, 'New name')
-})
+	// Verify that user is created in DB
+	const newUser = await User.findOne({email: 'new@example.com'});
+	t.is(newUser.name, 'New name');
+});
 ```
 
-**Shutdown your server and connection when done**:
+Finally disconnect from and stop MongoDB when all tests are done:
 
 ```js
-// Disconnect MongoDB and Mongoose after all tests are done
 test.after.always(async t => {
-  mongoose.disconnect()
-  mongod.stop()
+	mongoose.disconnect()
+	mongod.stop()
 })
 
 ```
@@ -136,12 +125,10 @@ And you're done!
 
 ## Reusing the configuration across files
 
-You may choose to abstract code for `test.before`, `test.beforeEach`, `test.afterEach.always` and `test.after.always` into a separate file.
+You may choose to extract the code for the `test.before`, `test.beforeEach`, `test.afterEach.always` and `test.after.always` hooks into a separate file. Have a look at https://github.com/zellwk/ava-mdb-test for an example.
 
-To see a demo of this configuration file, look at https://github.com/zellwk/ava-mdb-test
+## Using `test.serial` instead of `test`
 
-## Why `test.serial` instead of `test`
+Your tests likely change the database. Using `test()` means they run concurrently, which may cause one test to affect another. Instead if you use `test.serial()` then the tests will run one at a time. You can then clean up your database between test runs, making the tests more predictable.
 
-You need Mongoose to use the same connection in both your app and AVA to test endpoints properly. The easiest way is to make sure your tests run serially, clearing test fixtures after every test with `test.after.always` before the next one begins.
-
-There's a harder way (which allows you to run parallel tests), where you create separate Mongoose connections for each test. You'll also have to modify your Schemas to support this behavior. More information can be found [here](https://github.com/nodkz/mongodb-memory-server#several-mongoose-connections-simultaneously)
+You could run tests concurrently if you create separate Mongoose connections for each test. This is harder to set up, though. More information can be found [here](https://github.com/nodkz/mongodb-memory-server#several-mongoose-connections-simultaneously).
--- a/docs/recipes/endpoint-testing-with-mongoose.md
+++ b/docs/recipes/endpoint-testing-with-mongoose.md
@@ -66,7 +66,6 @@ const app = express()
 // # Middlewares
 // ======================================
 app.use(bodyParser.json())
-app.use(expressValidator())
 
 // ======================================
 // # Routes
--- a/docs/recipes/endpoint-testing-with-mongoose.md
+++ b/docs/recipes/endpoint-testing-with-mongoose.md
@@ -1,60 +1,73 @@
-When you test endpoints, you'd want to test the same code you use for production as much as possible. This recipe helps you set up endpoint testing with a different mongoose database per test file.
+# Endpoint testing with Mongoose
 
-Note that AVA doesn't have a built-in method for testing endpoints, but you can use libraries like [`supertest`](https://github.com/visionmedia/supertest) to help you with it.
+This recipe shows you how to test your endpoints with AVA and Mongoose.
 
-## Setting up
+## Setup
 
 This recipe uses the following libraries:
 
-1. Mongomem (A MongoDB in-memory Server)
-2. Supertest (An endpoint testing library)
-3. Mongoose
+1. [MongoDB memory server](https://github.com/nodkz/mongodb-memory-server) (A MongoDB in-memory Server)
+2. [Babel Polyfill](https://babeljs.io/docs/usage/polyfill/) (required for MongoDB memory server)
+3. [Supertest](https://github.com/visionmedia/supertest) (An endpoint testing library)
+4. Mongoose
 
-Install the libraries by running the following code:
+Install the first three libraries by running the following code:
 
 ```console
-$ npm install --save-dev mongomem supertest
+$ npm install --save-dev mongodb-memory-server babel-polyfill supertest
+```
+
+You should have Mongoose installed already. If not, run the following code to install it:
+
+(Note: You need at least mongoose v4.11.3)
+
+```console
+$ npm install mongoose
+```
+
+## Setting up AVA
+
+Since MongoDB Memory server requires Babel polyfill to work, the easiest way to set up AVA is through the `ava` key in your `package.json` file.
+
+```json
+"ava": {
+    "files": ["test/**/*"],
+    "require": [
+      "babel-register",
+      "babel-polyfill"
+    ]
+  },
+```
+
+If you want AVA to run your tests whenever changes are made, add the `--watch` option.
+
+```json
+"scripts": {
+  "watch": "ava --watch"
+},
 ```
 
 ## Your test file
 
-**First, include all libraries you're going to use:**
+First, include the libraries you need.
 
 ```js
+// Libraries required for testing
 import test from 'ava'
 import request from 'supertest'
-import { MongoDBServer } from 'mongomem'
-// Optionally, import your models here too
+import MongodbMemoryServer from 'mongodb-memory-server'
+import mongoose from 'mongoose'
+
+// Your server and models
+import app from '../server'
+import User from '../models/User'
 ```
 
-**Next, start your MongoDB instance with Mongomem**
+### Your server file
 
-```js
-test.before('start server', async t => MongoDBServer.start())
-```
+In your `server` file, you cannot use `app.listen` to start your app. Export it to another file and call `app.listen` there.
 
-**For each test you run, you need to configure a new Express instance** that connects to a different MongoDB URL. Make sure you add your server only after Mongoose is started.
-
-```js
-const setupMongoose = async _ => {
-  await mongoose.connect(await MongoDBServer.getConnectionString())
-  return mongoose
-}
-
-test.beforeEach(async t => {
-  const db = await setupMongoose()
-  const app = require('./server')
-
-  // Setup any fixtures you need here. This is a placeholder code
-  await setupFixtures()
-
-  // Pass app and mongoose into your tests
-  t.context.app = app
-  t.context.db = db
-})
-```
-
-Your server file may resemble the following. In this file, you cannot use `app.listen` to start your app. (Export it to another file and call `app.listen` there).
+This `server` file should resemble the following:
 
 ```js
 const express = require('express')
@@ -70,58 +83,200 @@ app.use(bodyParser.json())
 // ======================================
 // # Routes
 // ======================================
+app.get('/litmus', async (req, res) => {
+  const { email } = req.body
+  res.json(await User.findOne({email}))
+})
 
-app.use('/', routes)
+app.post('/litmus', async (req, res) => {
+  const { email, name } = req.body
+  const user = new User({email, name})
+  res.json(await user.save())
+})
 
 module.exports = app
 ```
 
-**After each test, you need to clear your fixtures and disconnect the Mongoose server**:
+### Back to your test file
+
+**First, start your MongoDB instance and connect to Mongoose:**
 
 ```js
-// use .always to help with teardown
-test.afterEach.always(async t => {
-  const { db } = t.context
-  // Note: removeFixtures is a placeholder. Write your own
-  await removeFixtures()
-  await db.connection.close()
+// Start MongoDB Instance
+const mongod = new MongodbMemoryServer()
+
+// Create connection to mongoose before all tests
+test.before(async t => mongoose.connect(await mongod.getConnectionString(), { useMongoClient: true }))
+```
+
+When you run your first test, MongoDB downloads the latest MongoDB Binaries. It may take a minute. (The download is ~70mb).
+
+Note: Since we're using async/await, you need Node v7.6 and above.
+
+**Add fixtures for each test**
+
+You'll want to populate your database with dummy data. Here's an example:
+
+```js
+test.beforeEach(async t => {
+  const user = new User({
+  	email: 'one@example.com',
+  	name: 'One'
+  })
+  await user.save()
 })
 ```
 
-**Next, you write your tests**.
-
-Since you're opening and closing mongose connections per test, make sure your tests run serially in this file. Here, you get get your express app from `t.context.app` (we did this step in `beforeEach`). Then, simply use `app` in supertest as follows:
+**Clear your dummy data after each test**:
 
 ```js
-// Note the serial tests
-test.serial('litmus get test', async t => {
+// Cleans up database after every test
+test.afterEach.always(async t => await User.remove())
+```
+
+**Write your tests**
+
+Use Supertest to fire a request for your endpoint. Then, do the rest with AVA normally.
+
+Note: Make sure your tests run serially with `test.serial`.
+
+```js
+// First test
+test.serial('litmus get user', async t => {
   const { app } = t.context
   const res = await request(app)
-    .get('/test1')
-    .send({
-      email: 'example@example.com'
-    })
+    .get('/litmus')
+    .send({email: 'one@example.com'})
   t.is(res.status, 200)
-  t.is(res.body.name, `somename`)
+  t.is(res.body.name, 'One')
 })
 
-test.serial('litmus post test', async t => {
+// Second test
+// Note: subsequent tests must be serial tests.
+// It is NOT RECOMMENDED to run parallel tests within an AVA test file when using Mongoose
+test.serial('litmus create user', async t => {
   const { app } = t.context
   const res = await request(app)
-    .post('/test2')
+    .post('/litmus')
     .send({
-      email: 'example@example.com'
-      name: `A random name`
+      email: 'new@example.com',
+      name: 'New name'
     })
+
   t.is(res.status, 200)
-  t.is(res.body.name, 'A random name')
+  t.is(res.body.name, 'New name')
+
+  // Verifies that user is created in DB
+  const newUser = await User.findOne({email: 'new@example.com'})
+  t.is(newUser.name, 'New name')
 })
 ```
 
-**After you're done with all your tests, close your MongoDB Server**:
+**Shutdown your server and connection when done**:
 
 ```js
-test.after.always('cleanup', t => MongoDBServer.tearDown())
+// Disconnect MongoDB and mongoose after all tests are done
+test.after.always(async t => {
+  mongoose.disconnect()
+  mongod.stop()
+})
+
 ```
 
-And you're done! :)
+And you're done!
+
+## Reusing the configuration across files
+
+You may choose to abstract code for `test.before`, `test.beforeEach`, `test.afterEach.always` and `test.after.always` into a separate file. It should look similar to this:
+
+```
+// utils.js
+// File for abstracting generic before, beforeEach, afterEach and after code
+
+const MongodbMemoryServer = require('mongodb-memory-server').default
+const mongoose = require('mongoose')
+
+// Your models and server
+const app = require('../server')
+const User = require('../models/User')
+
+const mongod = new MongodbMemoryServer()
+
+// Create connection to mongoose before all tests
+exports.before = async t =>
+  mongoose.connect(await mongod.getConnectionString(), { useMongoClient: true })
+
+// Create fixtures before each test
+exports.beforeEach = async t => {
+  const user = new User({email: 'one@example.com', name: 'One'})
+  const user2 = new User({email: 'two@example.com', name: 'Two'})
+  const user3 = new User({email: 'three@example.com', name: 'Three'})
+
+  await user.save()
+  await user2.save()
+  await user3.save()
+
+  // Saves app to t.context so tests can access app
+  t.context.app = app
+}
+
+// Clean up database after every test
+exports.afterEach = async t => await User.remove()
+
+// Disconnect MongoDB and mongoose after all tests are done
+exports.after = async t => {
+  mongoose.disconnect()
+  mongod.stop()
+}
+
+```
+
+Your test file then becomes much simpler:
+
+```
+import test from 'ava'
+import request from 'supertest'
+import User from '../models/User'
+
+import {before, beforeEach, afterEach, after} from './utils'
+
+test.before(before)
+test.beforeEach(beforeEach)
+test.afterEach.always(afterEach)
+
+// First test
+test.serial('litmus get user', async t => {
+  const { app } = t.context
+  const res = await request(app)
+    .get('/litmus')
+    .send({email: 'one@example.com'})
+  t.is(res.status, 200)
+  t.is(res.body.name, 'One')
+})
+
+// Second test
+// Note: subsequent tests must be serial tests.
+// It is NOT RECOMMENDED to run parallel tests within an AVA test file when using Mongoose
+test.serial('litmus create user', async t => {
+  const { app } = t.context
+  const res = await request(app)
+    .post('/litmus')
+    .send({
+      email: 'new@example.com',
+      name: 'New name'
+    })
+
+  t.is(res.status, 200)
+  t.is(res.body.name, 'New name')
+
+  // Verifies that user is created in DB
+  const newUser = await User.findOne({email: 'new@example.com'})
+  t.is(newUser.name, 'New name')
+})
+
+test.after.always(after)
+```
+
+## A demo repo
+
+Here's a [demo repo](https://github.com/zellwk/ava-mdb-test) if you like to see the recipe in action.
--- a/docs/recipes/endpoint-testing-with-mongoose.md
+++ b/docs/recipes/endpoint-testing-with-mongoose.md
@@ -82,6 +82,7 @@ module.exports = app
 // use .always to help with teardown
 test.afterEach.always(async t => {
   const { db } = t.context
+  // Note: removeFixtures is a placeholder. Write your own
   await removeFixtures()
   await db.connection.close()
 })
--- a/docs/recipes/endpoint-testing-with-mongoose.md
+++ b/docs/recipes/endpoint-testing-with-mongoose.md
@@ -1,6 +1,6 @@
 # Endpoint testing with Mongoose
 
-This recipe shows you how to test your endpoints with AVA and Mongoose.
+This recipe shows you how to test your endpoints with AVA and Mongoose, assuming you use Express as your framework.
 
 ## Setup
 
@@ -9,7 +9,7 @@ This recipe uses the following libraries:
 1. [MongoDB memory server](https://github.com/nodkz/mongodb-memory-server) (A MongoDB in-memory Server)
 2. [Babel Polyfill](https://babeljs.io/docs/usage/polyfill/) (required for MongoDB memory server)
 3. [Supertest](https://github.com/visionmedia/supertest) (An endpoint testing library)
-4. Mongoose
+4. [Mongoose](http://mongoosejs.com)
 
 Install the first three libraries by running the following code:
 
@@ -19,7 +19,7 @@ $ npm install --save-dev mongodb-memory-server babel-polyfill supertest
 
 You should have Mongoose installed already. If not, run the following code to install it:
 
-(Note: You need at least mongoose v4.11.3)
+(Note: You need at least Mongoose v4.11.3)
 
 ```console
 $ npm install mongoose
@@ -32,18 +32,9 @@ Since MongoDB Memory server requires Babel polyfill to work, the easiest way to 
 ```json
 "ava": {
     "require": [
-      "babel-register",
       "babel-polyfill"
     ]
   },
-```
-
-If you want AVA to run your tests whenever changes are made, add the `--watch` option.
-
-```json
-"scripts": {
-  "watch": "ava --watch"
-},
 ```
 
 ## Your test file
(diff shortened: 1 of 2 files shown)


docs/recipes/endpoint-testing-with-mongoose.md
```diff
@@ -0,0 +1,127 @@
...
+import test from 'ava'
+import request from 'supertest'
+import { MongoDBServer } from 'mongomem'
+// Optionally, import your models here too
+```
+
+**Next, start your MongoDB instance with Mongomem**
+
+```js
+test.before('start server', async t => await MongoDBServer.start())
```
OmgImAlexis: The await is redundant here since it's in a return statement.

docs/recipes/endpoint-testing-with-mongoose.md
```diff
@@ -0,0 +1,127 @@
...
+Since you're opening and closing mongose connections per test, make sure your tests run serially in this file. Here, you get get your express app from `t.context.app` (we did this step in `beforeEach`). Then, simply use `app` in supertest as follows:
+
+```js
+// Note the serial tests
+test.serial('litmus get test', async t => {
+  const { app } = t.context
+  const res = await request(app)
+    .get('/test1')
+    .send({
+      email: 'someemail@gmail.com'
```
OmgImAlexis: Might want to switch this to `example@example.com` just to be safe that you're not using a real email in an example.

OmgImAlexis:
Hopefully this is just a copying error but while testing this in my own project all I manage to get is an error thrown and my tests all fail. 

```
➜  api.wvvw.me git:(master) ✗ yarn ava -- test/user.spec.js  --verbose
yarn ava v0.24.6
$ "/Users/xo/code/api.wvvw.me/node_modules/.bin/ava" test/user.spec.js --verbose

  ✖ start server Rejected promise returned by test

  1 test failed [12:36:40]

  start server

  Rejected promise returned by test

  Rejection reason:

    "Mongod shutting down"


error Command failed with exit code 1.
```

novemberborn:
@zellwk we already have a recipe that uses `mongomem` and `mongoose`, written by the creator of `mongomem`. Rather than adding a new recipe, how do we improve the current one?

docs/recipes/endpoint-testing-with-mongoose.md
zellwk: 👍

docs/recipes/endpoint-testing-with-mongoose.md
zellwk: 👍

zellwk:
@OmgImAlexis Hmm. Would you mind trying again? Recipe looks right to me. Whats in your `user.spec.js`

@novemberborn I can edit the file directly, would that work better? Regarding that recipe, I tested it and it doesn't work directly with my app because the Mongoose instances are different, which is why I decided to make this recipe.

OmgImAlexis:
Here's my `./user.spec.js`. My `./main.js` exports app as default.
```js
import test from 'ava';
import request from 'supertest';
import {MongoDBServer} from 'mongomem';

import {setupFixtures, removeFixtures, setupMongoose} from './utils';

test.before('start server', async () => {
    await MongoDBServer.start();
});

test.beforeEach(async t => {
    const db = await setupMongoose();
    const app = require('../main');

    // Setup any fixtures you need here. This is a placeholder code
    await setupFixtures();

    // Pass app and mongoose into your tests
    t.context.app = app;
    t.context.db = db;
});

test.afterEach.always(async t => {
    const {db} = t.context;
    // Note: removeFixtures is a placeholder. Write your own
    await removeFixtures();
    await db.connection.close();
});

// Note the serial tests
test('create', async t => {
    const {app} = t.context;
    const res = await request(app).post('/user').send({
        username: 'ava',
        password: 'avarocks'
    });
    t.is(res.status, 201);
    t.is(res.body.username, 'ava');
});

test.after.always('cleanup', () => MongoDBServer.tearDown());
```

`./utils/index.js`
```js
import mongoose from 'mongoose';
import {MongoDBServer} from 'mongomem';

const setupFixtures = () => {};
const removeFixtures = () => {};
const setupMongoose = async () => {
    await mongoose.connect(await MongoDBServer.getConnectionString());
    return mongoose;
};

export {
    setupFixtures,
    removeFixtures,
    setupMongoose
};
```

zellwk:
Looks right. Can you tell me how you handled `/user`?

OmgImAlexis:
In `./main.js` I import all my routes and then use `app.use('/user', user);`

`./routes/user.js`

```js
import {Router} from 'express';
import HTTPError from 'http-errors';

import log from '../log';
import config from '../config';
import {User} from '../models';
import {isValidObjectId} from '../middleware';

const router = new Router();

router.get(['/', '/:id'], isValidObjectId, async (req, res, next) => {
    if (req.params.id) {
        const user = await User.find({
            _id: req.params.id
        }).exec().catch(next);

        return res.send(user);
    }

    const users = await User.find({}).sort({
        date: -1
    }).exec().catch(next);

    res.send(users);
});

router.post('/', (req, res, next) => {
    if (!config.get('signups.enabled')) {
        // @TODO: This may not be the best status code for this.
        return next(new HTTPError.MethodNotAllowed('Signups are currently disabled.'));
    }

    const user = new User({
        username: req.body.username || '',
        email: req.body.email || '',
        password: req.body.password
    });
    log.debug(user);
    user.save().then(() => {
        log.debug('User saved.');
        return res.send({
            user
        });
    }).catch(err => {
        log.debug('User error.');
        // Duplicate field
        if (err.code === 11000) {
            return next(new HTTPError.Conflict('User already exists.'));
        }
        return next(err);
    });
});

export default router;
```

zellwk:
Do you mind trying out a handler like this?

```
app.get('/user', (req, res) => {
  const { email } = req.body
  User.findOne({email})
   .then(r => res.json(r))
  .catch(e => /* handle whatever errors */)
})
```

OmgImAlexis:
Okay this is an issue with mongomem, looks like it's trying to bind to `0.0.0.0:27017` even though it's using `get-port`.

Ref: https://github.com/CImrie/mongomem/issues/2

zellwk:
Ah yes. I had the same problem too. Just to confirm. This recipe works fine for you. Yeah?

OmgImAlexis:
Yep after releasing the port from my docker container it's now fine.

OmgImAlexis:
As per https://github.com/CImrie/mongomem/issues/2#issuecomment-310947227 maybe this should use `mongodb-memory-server` instead since ava needs to run all tests in parallel?

novemberborn:
> @novemberborn I can edit the file directly, would that work better? Regarding that recipe, I tested it and it doesn't work directly with my app because the Mongoose instances are different, which is why I decided to make this recipe.

Both recipes are very similar to each other. Ideally we have one. If there's some additional use case or nuance we can discuss it in the recipe.

zellwk:
@OmgImAlexis I have tried using mondob-memory-server with the following code, but was unable to get tests to run in parallel still. 

```
test.beforeEach(async t => {
  const mongod = new MongodbMemoryServer()
  const uri = await mongod.getConnectionString()
  await mongoose.connect(uri)
  await setupFixtures()
  const app = require('../../server')

  t.context.mongod = mongod
  t.context.app = app
  t.context.db = mongoose
})

test.afterEach.always(async t => {
  const { db, mongod } = t.context
  await removeFixtures()
  db.connection.close()
  mongod.stop()
})
```

The error I get when attempting to run tests in parallel is this: 

```
litmus › beforeEach for litmus post test
  /Users/zellwk/Projects/private-repos/course-server/node_modules/mongoose/lib/connection.js:211

  Rejected promise returned by test

  Rejection reason:

    [Error: Trying to open unclosed connection.]

  NativeConnection.Connection._handleOpenArgs (node_modules/mongoose/lib/connection.js:211:11)
  NativeConnection.Connection.open (node_modules/mongoose/lib/connection.js:306:37)
  Mongoose.connect (node_modules/mongoose/lib/index.js:259:47)
  Test.<anonymous> (test/routes/litmus.spec.js:17:18)
  step (test/routes/litmus.spec.js:25:191)
```

Here, it seems like, without using `mongooseInstance = new mongoose.Mongoose()` as [this recipe suggests](https://github.com/avajs/ava/blob/master/docs/recipes/isolated-mongodb-integration-tests.md), we can't test databases in parallel. But if we do use `new mongoose.Mongoose()`, we have to change how `mongoose` connects within the app. (And I have no idea how to do it). If you have any suggestions, I'd be happy to try!

zellwk:
> Both recipes are very similar to each other. Ideally we have one. If there's some additional use case or nuance we can discuss it in the recipe.

@novemberborn Sure. We can merge it into one recipe eventually. Can we iron out the issues mentioned above before doing so? That way, we can see if there's really an additional use case or nuance.

OmgImAlexis:
@nodkz since you're the dev for `mongodb-memory-server` would you be able to help here?

docs/recipes/endpoint-testing-with-mongoose.md
```diff
@@ -0,0 +1,127 @@
...
+**For each test you run, you need to configure a new Express instance** that connects to a different MongoDB URL. Make sure you add your server only after Mongoose is started.
+
+```js
+const setupMongoose = async _ => {
+  await mongoose.connect(await MongoDBServer.getConnectionString())
+  return mongoose
+}
+
+test.beforeEach(async t => {
+  const db = await setupMongoose()
```
novemberborn: Would be better to inline the `setupMongoose` function here.

docs/recipes/endpoint-testing-with-mongoose.md
```diff
@@ -0,0 +1,127 @@
...
+  const app = require('./server')
```
novemberborn: This isn't reloaded for every test. Better to require it outside the test hook.

docs/recipes/endpoint-testing-with-mongoose.md
```diff
@@ -0,0 +1,127 @@
...
+
+  // Setup any fixtures you need here. This is a placeholder code
+  await setupFixtures()
```
novemberborn: This seems pretty critical actually. How does `app` use the `db` instance?

docs/recipes/endpoint-testing-with-mongoose.md
```diff
@@ -0,0 +1,127 @@
...
+
+**After each test, you need to clear your fixtures and disconnect the Mongoose server**:
+
+```js
+// use .always to help with teardown
+test.afterEach.always(async t => {
+  const { db } = t.context
+  // Note: removeFixtures is a placeholder. Write your own
+  await removeFixtures()
+  await db.connection.close()
```
novemberborn: https://github.com/avajs/ava/blob/master/docs/recipes/isolated-mongodb-integration-tests.md doesn't do this. I wonder if it should.

docs/recipes/endpoint-testing-with-mongoose.md
zellwk: The 'require' statement configures Mongoose according to the current test (I believe). May not work properly if Mongoose is not set up per test.

docs/recipes/endpoint-testing-with-mongoose.md
zellwk: Another way is to disconnect mongoose. Works the same, as far as I can tell. Awaiting @nodkz's input on `mongodb-memory-server` right now. Otherwise, tests can't run in parallel. Should we wait? Or should I edit according to your comments first?

docs/recipes/endpoint-testing-with-mongoose.md
zellwk: Sure

docs/recipes/endpoint-testing-with-mongoose.md
novemberborn: Modules are cached after the first time they're loaded, so requiring `./server` for each test shouldn't make a difference.

nodkz:
@OmgImAlexis @zellwk 
Sorry guys for the late response. 

You may test mongoose in several ways:
- As described in README for [JEST example](https://github.com/nodkz/mongodb-memory-server#simple-jest-test-example)
- Or something complex via mocking mongoose module. Like it done in graphql-compose-mongoose:
  - [mock mongoose](https://github.com/nodkz/graphql-compose-mongoose/blob/master/src/__mocks__/mongooseCommon.js) just mock `connect` method and adds callback on `disconnect` for shutdowning mongo server.
  - [user model](https://github.com/nodkz/graphql-compose-mongoose/blob/master/src/__mocks__/userModel.js) used mocked mongoose module to define schema/model
  - [test1](https://github.com/nodkz/graphql-compose-mongoose/blob/master/src/resolvers/__tests__/findById-test.js) - `beforeAll(() => UserModel.base.connect());` calls only once mocked connect method per test file, and `afterAll(() => UserModel.base.disconnect());` for disconnecting mongoose and shutdowning server and freeing event loop for correct test exit
  - [test2](https://github.com/nodkz/graphql-compose-mongoose/blob/master/src/resolvers/__tests__/removeOne-test.js) - same operations.

This tests works in parallel. Each **test file** has own mongodb server.

---

Also it should work for your case, where you use a dedicated database **for every test case** in one test file. If not, please provide test repo i try to figure out what is the problem.

---

I see following disadvantages in your example:
- **low performance** it's not good to use for **every test case** dedicated server. It's ok to use one server **per one test file**. If you need for **every test case** different db:
  - spin up one server before all tests `const mongoServer = new MongodbMemoryServer();`
  - and for **every test case**  use different dbs via [`await mongoServer.getConnectionString('someDbName')`](https://github.com/nodkz/mongodb-memory-server#several-mongoose-connections-simultaneously).
- also it **may be some internal `mongoose` problems** if AVA runs test in event loop, not in separate processes. Cause `mongoose` package use globals inside and it may produce some errors when different test cases set different connection strings and tries to connect.
- `mongomem` in parallel tests will download 70MB files first time for every test (it's bad for CI envs). `mongodb-memory-server` has internal lock file for such cases and should download binaries only once. Also you may provide required version of mongo server via [`new MongodbMemoryServer({ binary: { version: '3.4.4' }})`](https://github.com/nodkz/mongodb-memory-server#available-options); `mongomem` does not allow you to choose mongod version.

zellwk:
@nodkz Thanks for chipping in! I've tried your suggestions, but I still couldn't get ava tests to run in parallel. By parallel, I mean tests within a test file. So far, my test file looks like this: 

```
import test from 'ava'
import app from '../server'
import request from 'supertest'
import User from '../models/User'
import mongoose from '../handlers/mongoose'
import MongodbMemoryServer from 'mongodb-memory-server'

const mongod = new MongodbMemoryServer()

test.before(async t => {
  mongoose.connect(await mongod.getConnectionString())
})

test.beforeEach(async t => {
  const user = new User({email: 'one@example.com', name: 'One'})
  const user2 = new User({email: 'two@example.com', name: 'Two'})
  const user3 = new User({email: 'three@example.com', name: 'Three'})
  await user.save()
  await user2.save()
  await user3.save()
  t.context.app = app
})

test.afterEach.always(async t => {
  await User.remove()
})

test('litmus get test', async t => {
  const { app } = t.context
  const res = await request(app)
    .get('/litmus')
    .send({email: 'one@example.com'})
  t.is(res.status, 200)
  t.is(res.body.name, 'One')
})

// Works if we use test.serial instead of test
test('litmus post test', async t => {
  const { app } = t.context
  const res = await request(app)
    .post('/litmus')
    .send({
      email: 'new@example.com',
      name: 'New'
    })
  t.is(res.status, 200)
  t.is(res.body.name, 'New')
  t.is(200, 200)
})

test.after.always(async t => {
  mongoose.disconnect()
  mongod.stop()
})
```

Is this the best we can do? If it helps with debugging, you can find a demo repo at https://github.com/zellwk/ava-mdb-test.

docs/recipes/endpoint-testing-with-mongoose.md
zellwk: You're right about this. Let me make the requested changes after confirming if there's a better implementation with @nodkz, as commented below.

docs/recipes/endpoint-testing-with-mongoose.md
zellwk: Will expand on this!

zellwk:
@nodkz, @novemberborn, @OmgImAlexis any suggestions on the demo repo I've sent? If not, I'll go ahead and edit this docs again.

nodkz:
@zellwk miss your prev message. Now I'm looking your repo. Write about results today later.

nodkz:
Tests do not work on node 7.2.1
----
Got following error:
```
$ ava --watch
/Users/nod/www/_sandbox/ava-mdb-test/server.js:16
app.get('/litmus', async (req, res) => {
                         ^
SyntaxError: Unexpected token (
    at Object.exports.runInThisContext (vm.js:78:16)
    at Module._compile (module.js:543:28)
    at loader (/Users/nod/www/_sandbox/ava-mdb-test/node_modules/babel-register/lib/node.js:144:5)
    at require.extensions.(anonymous function) (/Users/nod/www/_sandbox/ava-mdb-test/node_modules/babel-register/lib/node.js:154:7)
    at extensions.(anonymous function) (/Users/nod/www/_sandbox/ava-mdb-test/node_modules/require-precompiled/index.js:16:3)
    at Object.require.extensions.(anonymous function) [as .js] (/Users/nod/www/_sandbox/ava-mdb-test/node_modules/ava/lib/process-adapter.js:100:4)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)

    at Function.Module._load (module.js:439:3)
    at Module.require (module.js:498:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/Users/nod/www/_sandbox/ava-mdb-test/test/litmus.spec.js:2:1)
    at Module._compile (module.js:571:32)
    at extensions.(anonymous function) (/Users/nod/www/_sandbox/ava-mdb-test/node_modules/require-precompiled/index.js:13:11)
    at Object.require.extensions.(anonymous function) [as .js] (/Users/nod/www/_sandbox/ava-mdb-test/node_modules/ava/lib/process-adapter.js:100:4)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
    at Module.require (module.js:498:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/Users/nod/www/_sandbox/ava-mdb-test/node_modules/ava/lib/test-worker.js:61:1)
    at Module._compile (module.js:571:32)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
    at Module.runMain (module.js:605:10)
    at run (bootstrap_node.js:420:7)
    at startup (bootstrap_node.js:139:9)
    at bootstrap_node.js:535:3
```

On 8.2.0 works perfectly. I think that NOTE is required in docs about minimal node version.

Fix mongoose deprecationWarning: `open()`
---

Following console warning:
```
(node:5017) DeprecationWarning: `open()` is deprecated in mongoose >= 4.11.0, use `openUri()` instead, or set the `useMongoClient` option if using `connect()`
 or `createConnection()`. See http://mongoosejs.com/docs/connections.html#use-mongo-client
```
Was fixed in the beggining of this week (needs to install the last mongoose version 4.11.3). Also it's required small config changes:
```diff
// handlers/mongoose.js
- mongoose.connect = async (uri) => {
+ mongoose.connect = async (uri, opts) => {
-   originalConnect.bind(mongoose)(uri)
+   originalConnect.bind(mongoose)(uri, opts)

// test/litmus.spec.js
- mongoose.connect(await mongod.getConnectionString())
+ mongoose.connect(await mongod.getConnectionString(), { useMongoClient: true })
```

For more info you may read [docs](http://mongoosejs.com/docs/connections.html#use-mongo-client) and may see this issues [#5423](https://github.com/Automattic/mongoose/issues/5423), [huge #5399](https://github.com/Automattic/mongoose/issues/5399)


Parallel tests inside one file do not work with one mongoose model. And never will not work.
----

Tests cases inside one file when runs in parallel uses the same proccess (checked it with `console.log(process.pid)`). So it means that all test cases work via EventLoop and on top of context used THE SAME mongoose module (THE SAME mongoose model USER). So in current setup it is impossible to test mongoose in parallel with different connections. Connection will be one and the same. So you got a dubplicate key error 11000.

How to solve it:
1. **DO NOT USE parallel tests inside one file [RECOMMENDED]. You always may split your tests in several files for which ava will use different process for every file (tested with your repo and it works perfectly). So you will not meet with mongoose globals restrictions.**

2. Create one more mock for User model. If you want to ran in parallel one mongoose schema with different connections you must create two mongoose models:
```js
import mongoose from 'mongoose';
import MongodbMemoryServer from 'mongodb-memory-server';

mongoose.Promise = Promise;
const mongoServer = new MongodbMemoryServer();

const connections = {
  conn1: mongoose.createConnection(),
  conn2: mongoose.createConnection(),
  ...
  ...
};

const mongooseOpts = {
  server: {
    promiseLibrary = Promise;
    auto_reconnect: true,
    reconnectTries: Number.MAX_VALUE,
    reconnectInterval: 1000,
  },
};

mongoServer1.getConnectionString('server1_db1').then((mongoUri) => {
  connections.conn1.open(mongoUri, mongooseOpts);
  connection.once('open', () => {
    console.log(`MongoDB successfully connected to ${mongoUri}`);
  });
});

mongoServer1.getConnectionString('server1_db2').then((mongoUri) => {
  connections.conn2.open(mongoUri, mongooseOpts);
  connection.once('open', () => {
    console.log(`MongoDB successfully connected to ${mongoUri}`);
  });
});

...
...

export default connections;


// somewhere in other file
import { Schema } from 'mongoose';
import { conn1, conn2 } from './file_above';

const userSchema = new Schema({
  name: String,
});

export default {
  UserForConnection1: conn1.model('user', userSchema),
  UserForConnection2: conn2.model('user', userSchema),
  ...
}
```
It can be simplifed. But stays still complex for test. For second case you should somehow mock `User` that for first test it got `UserForConnection1`, and for second test `UserForConnection2` for third test `...` and so one. I'm crying.

**So my decision: DO NOT USE PARALLEL TESTS INSIDE ONE FILE FOR ONE MONGOOSE MODEL**. You may split huge test into several files.

zellwk:
@nodkz 👍. Thanks so much for all the help! I'm going to update this PR with your latest findings and ping the others.

zellwk:
@OmgImAlexis, @novemberborn and @nodkz, The recipe is updated. Please take a look and let me know if it's good to go. 

I think this recipe should overwrite the [original endpoint testing recipe](https://github.com/avajs/ava/blob/master/docs/recipes/endpoint-testing.md). They're pretty similar. What do you think?

nodkz:
@zellwk good write up 👍

```js
import {before, beforeEach, afterEach, after} from './utils'

test.before(before)
test.beforeEach(beforeEach)
test.afterEach.always(afterEach)
```

Try to reduce this code a little bit more in such manner:

```js
import { prepareMongoose } from './utils';

prepareMongoose(test);

// ------utils.js--------
prepareMongoose(test) {
  test.before(before)
  test.beforeEach(beforeEach)
  test.afterEach.always(afterEach)
}
```

I do not sure will it work, but if so it makes tests nicer.

docs/recipes/endpoint-testing-with-mongoose.md
```diff
@@ -1,60 +1,73 @@
...
-**Next, start your MongoDB instance with Mongomem**
+## Setting up AVA
 
-```js
-test.before('start server', async t => MongoDBServer.start())
+Since MongoDB Memory server requires Babel polyfill to work, the easiest way to set up AVA is through the `ava` key in your `package.json` file.
+
+```json
+"ava": {
+    "files": ["test/**/*"],
```
OmgImAlexis: This shouldn't be needed since it's part of the default set.

docs/recipes/endpoint-testing-with-mongoose.md
```diff
@@ -70,58 +83,200 @@ app.use(bodyParser.json())
...
 ```
 
-And you're done! :)
+And you're done!
+
+## Reusing the configuration across files
+
+You may choose to abstract code for `test.before`, `test.beforeEach`, `test.afterEach.always` and `test.after.always` into a separate file. It should look similar to this:
+
+```
```
OmgImAlexis: Add js to keep highlighting working.

docs/recipes/endpoint-testing-with-mongoose.md
```diff
@@ -70,58 +83,200 @@ app.use(bodyParser.json())
...
+exports.after = async t => {
+  mongoose.disconnect()
+  mongod.stop()
+}
+
+```
+
+Your test file then becomes much simpler:
+
+```
```
OmgImAlexis: Add js to keep highlighting working.

OmgImAlexis:
For the example I'd suggest linking to something controlled by the avajs org so it can be updated if the recipe changes without a third party needing to do anything.

zellwk:
@OmgImAlexis Changed as requested. For the example, should I cede ownership of the repo to AVA? If not, let me know what's the best way. 

@nodkz Thanks Pavel. I'm pretty sure what you've suggested works. Not sure if @OmgImAlexis and @novemberborn would like to be more explicit about `test.before`, `test.beforeEach` and `test.afterEach.always`.

docs/recipes/endpoint-testing-with-mongoose.md
```diff
@@ -0,0 +1,281 @@
...
+This recipe shows you how to test your endpoints with AVA and Mongoose.
+
+## Setup
+
+This recipe uses the following libraries:
+
+1. [MongoDB memory server](https://github.com/nodkz/mongodb-memory-server) (A MongoDB in-memory Server)
+2. [Babel Polyfill](https://babeljs.io/docs/usage/polyfill/) (required for MongoDB memory server)
+3. [Supertest](https://github.com/visionmedia/supertest) (An endpoint testing library)
+4. Mongoose
```
novemberborn: Would be nice to link to http://mongoosejs.com/ here.

docs/recipes/endpoint-testing-with-mongoose.md
```diff
@@ -0,0 +1,281 @@
...
+
+Install the first three libraries by running the following code:
+
+```console
+$ npm install --save-dev mongodb-memory-server babel-polyfill supertest
+```
+
+You should have Mongoose installed already. If not, run the following code to install it:
+
+(Note: You need at least mongoose v4.11.3)
```
novemberborn: The capitalization of Mongoose should be consistent throughout the recipe.

docs/recipes/endpoint-testing-with-mongoose.md
```diff
@@ -0,0 +1,281 @@
...
+
+```console
+$ npm install mongoose
+```
+
+## Setting up AVA
+
+Since MongoDB Memory server requires Babel polyfill to work, the easiest way to set up AVA is through the `ava` key in your `package.json` file.
```
novemberborn: I find this misleading. It just needs `regenerator-runtime`, right? Going by https://www.npmjs.com/package/regenerator-runtime (which is depended on by `babel-polyfill`) you could install that as a dependency and then require `regenerator-runtime/runtime`. That should only be done if you're testing on Node.js releases older than 8.

`babel-polyfill` impacts your environment in many other ways. This recipe should focus on what is absolutely necessary.

docs/recipes/endpoint-testing-with-mongoose.md
```diff
@@ -0,0 +1,281 @@
...
+
+```json
+"ava": {
+    "require": [
+      "babel-register",
```
novemberborn: Requiring `babel-register` can have unintended consequences. I don't see how it's required for this recipe?

docs/recipes/endpoint-testing-with-mongoose.md
```diff
@@ -0,0 +1,281 @@
...
+      "babel-polyfill"
+    ]
+  },
+```
+
+If you want AVA to run your tests whenever changes are made, add the `--watch` option.
```
novemberborn: While the watch mode is nice (it's what got me involved with AVA) I don't see how this is relevant to this recipe.

docs/recipes/endpoint-testing-with-mongoose.md
```diff
@@ -0,0 +1,281 @@
+# Endpoint testing with Mongoose
+
+This recipe shows you how to test your endpoints with AVA and Mongoose.
+
+## Setup
+
+This recipe uses the following libraries:
```
novemberborn: I think you should mention here that the recipe assumes you have an Express server

docs/recipes/endpoint-testing-with-mongoose.md
```diff
@@ -0,0 +1,281 @@
...
+import mongoose from 'mongoose'
+
+// Your server and models
+import app from '../server'
+import User from '../models/User'
+```
+
+### Your server file
+
+In your `server` file, you cannot use `app.listen` to start your app. Export it to another file and call `app.listen` there.
```
novemberborn: Here you should repeat that this example is based on Express. It's not a requirement, so stating it up front may make it easier for others to see how they can apply it to their own setup.

Perhaps place the `app.listen` bit after the code sample. You could phrase it like "Note that this doesn't call `app.listen()`. We'll need to do that ourselves in our tests. For your application make sure you have a startup file that imports `app` and calls `listen()`."

docs/recipes/endpoint-testing-with-mongoose.md
```diff
@@ -0,0 +1,281 @@
...
+// Start MongoDB Instance
+const mongod = new MongodbMemoryServer()
+
+// Create connection to mongoose before all tests
+test.before(async t => mongoose.connect(await mongod.getConnectionString(), { useMongoClient: true }))
+```
+
+When you run your first test, MongoDB downloads the latest MongoDB Binaries. It may take a minute. (The download is ~70mb).
+
+Note: Since we're using async/await, you need Node v7.6 and above.
```
novemberborn: AVA takes care of this for you. I'd assume (especially when `regenerator-runtime` is loaded this recipe works all the way back on Node.js 4).

docs/recipes/endpoint-testing-with-mongoose.md
```diff
@@ -0,0 +1,281 @@
...
+// Your server and models
+import app from '../server'
+import User from '../models/User'
+```
+
+### Your server file
+
+In your `server` file, you cannot use `app.listen` to start your app. Export it to another file and call `app.listen` there.
+
+This `server` file should resemble the following:
```
novemberborn: I'm thinking this should just link to https://github.com/zellwk/ava-mdb-test/blob/master/server.js. Not sure it it needs to be copied into the test. Similarly you may want to link to https://github.com/zellwk/ava-mdb-test/blob/master/models/User.js which, not knowing anything about Mongoose, I find quite relevant.

docs/recipes/endpoint-testing-with-mongoose.md
```diff
@@ -0,0 +1,281 @@
...
+```js
+// Cleans up database after every test
+test.afterEach.always(async t => await User.remove())
+```
+
+**Write your tests**
+
+Use Supertest to fire a request for your endpoint. Then, do the rest with AVA normally.
+
+Note: Make sure your tests run serially with `test.serial`.
```
novemberborn: You should explain why these tests must use `.serial`. What happens if I just use `test()`?

docs/recipes/endpoint-testing-with-mongoose.md
```diff
@@ -0,0 +1,281 @@
...
+  const res = await request(app)
+    .get('/litmus')
+    .send({email: 'one@example.com'})
+  t.is(res.status, 200)
+  t.is(res.body.name, 'One')
+})
+
+// Second test
+// Note: subsequent tests must be serial tests.
+// It is NOT RECOMMENDED to run parallel tests within an AVA test file when using Mongoose
```
novemberborn: All this could be covered in the earlier paragraph.

docs/recipes/endpoint-testing-with-mongoose.md
```diff
@@ -0,0 +1,281 @@
...
+  mongod.stop()
+})
+
+```
+
+And you're done!
+
+## Reusing the configuration across files
+
+You may choose to abstract code for `test.before`, `test.beforeEach`, `test.afterEach.always` and `test.after.always` into a separate file. It should look similar to this:
```
novemberborn: While a useful tip, I don't think you need to show a full example.

docs/recipes/endpoint-testing-with-mongoose.md
zellwk: I tried adding regenerator-runtime, but it didn't work. Looks like regenerator-runtime is required in mongodb-memory-server? 

Here's the stack trace: 

```
/Users/zellwk/Projects/test-repos/ava-mdb-test/node_modules/mongodb-memory-server/lib/MongoMemoryServer.js:10
  var _ref = _asyncToGenerator(regeneratorRuntime.mark(function _callee(port, dbName) {
                               ^

ReferenceError: regeneratorRuntime is not defined
    at /Users/zellwk/Projects/test-repos/ava-mdb-test/node_modules/mongodb-memory-server/lib/MongoMemoryServer.js:10:32
    at Object.<anonymous> (/Users/zellwk/Projects/test-repos/ava-mdb-test/node_modules/mongodb-memory-server/lib/MongoMemoryServer.js:44:2)
```

docs/recipes/endpoint-testing-with-mongoose.md
zellwk: You're right, babel-register is not required. My bad. Thanks!

docs/recipes/endpoint-testing-with-mongoose.md
zellwk: 👍

docs/recipes/endpoint-testing-with-mongoose.md
zellwk: 👍

docs/recipes/endpoint-testing-with-mongoose.md
zellwk: 👍

docs/recipes/endpoint-testing-with-mongoose.md
zellwk: True. Let me omit it.

docs/recipes/endpoint-testing-with-mongoose.md
zellwk: 👍

docs/recipes/endpoint-testing-with-mongoose.md
zellwk: Sure, I'll link out.

docs/recipes/endpoint-testing-with-mongoose.md
zellwk: Should I link to this PR? Since @nodkz explained in detail, but there's probably no need for others to go into the same depth.

docs/recipes/endpoint-testing-with-mongoose.md
zellwk: 👍

docs/recipes/endpoint-testing-with-mongoose.md
zellwk: Sure, linked.

zellwk:
@novemberborn I made the changes you requested. One thing still concerns me regarding 'regenerator-runtime'. I can't get the tests to work without `babel-polyfill`. I've included the stack trace in the comments, but here it is again: 

```
/Users/zellwk/Projects/test-repos/ava-mdb-test/node_modules/mongodb-memory-server/lib/MongoMemoryServer.js:10
  var _ref = _asyncToGenerator(regeneratorRuntime.mark(function _callee(port, dbName) {
                               ^

ReferenceError: regeneratorRuntime is not defined
    at /Users/zellwk/Projects/test-repos/ava-mdb-test/node_modules/mongodb-memory-server/lib/MongoMemoryServer.js:10:32
    at Object.<anonymous> (/Users/zellwk/Projects/test-repos/ava-mdb-test/node_modules/mongodb-memory-server/lib/MongoMemoryServer.js:44:2)
```

@nodkz Would you mind checking the above? Let me know if `regenerator-runtime` is required directly in the tests, or within mongodb-memory-server? 

> I'm not sure what to do with https://github.com/avajs/ava/blob/master/docs/recipes/isolated-mongodb-integration-tests.md though. Again it's more focused on just MongoDB testing, with a bit of Mongoose added on. It uses a different memory server though. Perhaps it'd be sufficient for these recipes to reference each other.

I've tried the recipe before creating this one. The mongoose version doesn't work. Had a tough time understanding everything, until I hacked stuff together with @nodkz and @OmgImAlexis help in this PR. 

So, I recommend removing the Mongoose parts, but keep the MongoDB parts for that recipe.

docs/recipes/endpoint-testing-with-mongoose.md
OmgImAlexis: @zellwk open an issue with them, it should be added when it's transpiled to ES5 before they publish it.

nodkz:
@zellwk try new mongodb-memory-server@1.4.0 it imports `regenerator-runtime` explicitly in its code, so I suppose now tests should work without `babel-polyfill`.

docs/recipes/endpoint-testing-with-mongoose.md
```diff
@@ -141,6 +108,8 @@ Note: Make sure your tests run serially with `test.serial`.
 
 ```js
 // First test
+// Note: tests must be serial tests.
+// It is NOT RECOMMENDED to run parallel tests within an AVA test file when using Mongoose
```
OmgImAlexis: This really needs a reason why.

nodkz:
In mongodb-memory-server@1.4.1 ship some changes with babel-polyfill. 

Now your tests may work without babel-polyfill and regenerator.

https://github.com/nodkz/mongodb-memory-server/issues/9#issuecomment-323044701

docs/recipes/endpoint-testing-with-mongoose.md
novemberborn: It should be summarized in the recipe. Otherwise people may either think that they should always use `test.serial()`, or they'll go "that can't be right" and remove it.

novemberborn:
>> I'm not sure what to do with /docs/recipes/isolated-mongodb-integration-tests.md@master though. Again it's more focused on just MongoDB testing, with a bit of Mongoose added on. It uses a different memory server though. Perhaps it'd be sufficient for these recipes to reference each other.
>
> I've tried the recipe before creating this one. The mongoose version doesn't work. Had a tough time understanding everything, until I hacked stuff together with @nodkz and @OmgImAlexis help in this PR.
> 
> So, I recommend removing the Mongoose parts, but keep the MongoDB parts for that recipe.

That's a good idea. Could you add that to this PR? I can fix it up afterwards to if you'd like.

novemberborn:
> In mongodb-memory-server@1.4.1 ship some changes with babel-polyfill.
> 
> Now your tests may work without babel-polyfill and regenerator.
> 
> nodkz/mongodb-memory-server#9 (comment)

Thanks @nodkz! @zellwk could you check that's all good with this recipe, so we don't need the polyfill and stuff?

zellwk:
> Thanks @nodkz! @zellwk could you check that's all good with this recipe, so we don't need the polyfill and stuff?

Done! `babel-polyfill` is no longer required. Woot! 

> > So, I recommend removing the Mongoose parts, but keep the MongoDB parts for that recipe.

> That's a good idea. Could you add that to this PR? I can fix it up afterwards to if you'd like.

Done too! Removed the Mongoose parts and pointed them to the mongoose docs instead. I'm not sure whether we should change Mongmem to Mongodb-memory server on that recipe. Thoughts, @nodkz? 

@OmgImAlexis and @novemberborn, I also added an explanation on why the user should use `test.serial` instead of `test` with Mongoose. LMK if it looks good.

docs/recipes/endpoint-testing-with-mongoose.md
```diff
@@ -156,3 +139,13 @@ And you're done!
 You may choose to abstract code for `test.before`, `test.beforeEach`, `test.afterEach.always` and `test.after.always` into a separate file.
 
 To see a demo of this configuration file, look at https://github.com/zellwk/ava-mdb-test
+
+## Why `test.serial` instead of `test`
```
OmgImAlexis: Why not just use a new database for each test? That's how we run tests with [agenda](https://github.com/agenda/agenda/pull/506) and we don't have any issues.

nodkz:
Firstly I'm using `mongomem` too, but when met with its downsides was written`mongodb-memory-server`.
`mongomem` downsides:
- does not have `lock` for downloading binary (so when you start in parallel several tests from scratch, bins also will be downloaded several times).
- does not allow to choose mongodb version and bunch of other options.

Full list of options for `mongodb-memory-server`:
```js
const mongod = new MongodbMemoryServer({
  instance?: {
    port?: ?number, // by default choose any free port
    dbPath?: string, // by default create in temp directory
    storageEngine?: string, // by default `ephemeralForTest`
    debug?: boolean, // by default false
  },
  binary?: {
    version?: string, // by default '3.4.4'
    downloadDir?: string, // by default %HOME/.mongodb-prebuilt
    platform?: string, // by default os.platform()
    arch?: string, // by default os.arch()
    http?: any, // see mongodb-download package
    debug?: boolean, // by default false
  },
  debug?: boolean, // by default false
  autoStart?: boolean, // by default true
});
```

docs/recipes/endpoint-testing-with-mongoose.md
zellwk: When you use mongoose, you interact with the database through Mongoose. Since there can only be one Mongoose instance (and that Mongoose instance must always point to the same database), we cant access multiple databases even if we create them. 

Do I make sense?

zellwk:
In that case, shall we continue with Mongomem with the mongodb recipe?

docs/recipes/endpoint-testing-with-mongoose.md
OmgImAlexis: Yes you can. http://mongoosejs.com/docs/connections.html#multiple_connections

zellwk: Ah! I understood @nodkz's explanation (above) wrongly. Although you can create multiple connections with Mongoose, you need to create multiple models. Let me switch the explanation as soon as I find some time to.

nodkz: See `User` and `UserOnServer2` models constructed on one schema with different connections in the following example:
https://github.com/nodkz/mongodb-memory-server#several-mongoose-connections-simultaneously

zellwk: @OmgImAlexis updated with a slightly better explanation. Do I make sense in this new one?

OmgImAlexis: That does make sense but since ava is meant to be run in parallel it'd be better to explain how to set it up by default instead of expecting tests to run in serial.

zellwk: I fully understand that AVA runs tests in parallel, and your desire to set it up by default. I would love to do that too. 

However, based on @nodkz's explanation, and my personal research over the past month, I strongly believe endpoint testing with Mongoose and AVA is an exception where serial tests would be preferred. 

The setup for parallel tests changes how a person would create their schemas, and thus their app, too much for the additional benefit of running tests in parallel within each file. Besides, if you create multiple AVA test files, each test file still runs in parallel to other files, which shouldn't slow down testing too much. 

I'll be glad to change the docs if you could show me a good way to run endpoint testing in parallel with Mongoose and Supertest.

novemberborn:
Thanks @zellwk et al. I think this is pretty much there! I did push some edits, so please let me know if I've butchered anything 😇 

I've also changed the section on using `test.serial`: I'm pretty sure Mongoose already uses the same connection in the tests and the app. Notably, the connection is made in a `test.before()` hook, not a `test.beforeEach()` hook. Thus the real concern is tests interleaving and the database being in an unknown or conflicting state.

Please let me know if I got this wrong.

> In that case, shall we continue with Mongomem with the mongodb recipe?

We can modify that recipe to use `mongodb-memory-server` in a follow-up PR.

docs/recipes/endpoint-testing-with-mongoose.md
```diff
@@ -40,94 +46,77 @@ import app from '../server'
...
+	t.is(newUser.name, 'New name');
+});
 ```
 
-**Shutdown your server and connection when done**:
+Finally disconnect from and stop MongoDB when all tests are done:
 
 ```js
-// Disconnect MongoDB and Mongoose after all tests are done
 test.after.always(async t => {
```
OmgImAlexis: This can just be `() => {` since t isn't used and there's nothing using async.

docs/recipes/endpoint-testing-with-mongoose.md
```diff
@@ -40,94 +46,77 @@ import app from '../server'
...
+});
 ```
 
-**Clear your dummy data after each test**:
+Dummy data should be cleared after each test:
 
 ```js
-// Cleans up database after every test
-test.afterEach.always(async t => await User.remove())
+test.afterEach.always(async t => await User.remove());
```
OmgImAlexis: No need for the `t` argument here.

I'd change it to `test.afterEach.always(() => User.remove());`

docs/recipes/endpoint-testing-with-mongoose.md
```diff
@@ -40,94 +46,77 @@ import app from '../server'
...
 
-When you run your first test, MongoDB downloads the latest MongoDB Binaries. It may take a minute. (The download is ~70mb).
-
-**Add fixtures for each test**
+When you run your first test, MongoDB downloads the latest MongoDB binaries. The download is ~70MB so this may take a minute.
 
 You'll want to populate your database with dummy data. Here's an example:
 
 ```js
 test.beforeEach(async t => {
```
OmgImAlexis: No need for the `t` argument.

docs/recipes/endpoint-testing-with-mongoose.md
```diff
@@ -40,94 +46,77 @@ import app from '../server'
...
 
 ```js
-// Start MongoDB Instance
+// Start MongoDB instance
 const mongod = new MongodbMemoryServer()
 
-// Create connection to mongoose before all tests
-test.before(async t => mongoose.connect(await mongod.getConnectionString(), { useMongoClient: true }))
+// Create connection to Mongoose before tests are run
+test.before(async t => {
```
OmgImAlexis: No need for the `t` argument.

docs/recipes/endpoint-testing-with-mongoose.md
```diff
@@ -6,9 +6,9 @@ This recipe shows you how to test your endpoints with AVA and Mongoose, assuming
 
 This recipe uses the following libraries:
 
-1. [MongoDB memory server](https://github.com/nodkz/mongodb-memory-server) (A MongoDB in-memory Server)
-2. [Supertest](https://github.com/visionmedia/supertest) (An endpoint testing library)
-3. [Mongoose](http://mongoosejs.com)
+1. [`mongod-memory-server`](https://github.com/nodkz/mongodb-memory-server) (A MongoDB in-memory Server)
+2. [SuperTest](https://github.com/visionmedia/supertest) (An endpoint testing library)
+3. [Mongoose ODM](http://mongoosejs.com)
```
OmgImAlexis: Why was this changed when "Mongoose" is know as "Mongoose" not "Mongoose ODM"?

novemberborn: Ah I took that from the document title, but they don't refer to it that way elsewhere. Will change back, thanks!

novemberborn:
@OmgImAlexis updated.

docs/recipes/endpoint-testing-with-mongoose.md
```diff
@@ -76,7 +76,7 @@ test.beforeEach(async t => {
 Dummy data should be cleared after each test:
 
 ```js
-test.afterEach.always(async t => await User.remove());
+test.afterEach.always(async () => await User.remove());
```
OmgImAlexis: `async` and `await` are redundant here since you're returning a `Promise` from `User.remove()`.

novemberborn: The explicitness doesn't hurt I think.

OmgImAlexis: There's no need for it and anyone using `xo` will need to remove it if they copy+paste. Better to just remove it and set an example by only shipping code that's linted even if it's only in the docs.

zellwk:
> I've also changed the section on using test.serial: I'm pretty sure Mongoose already uses the same connection in the tests and the app. Notably, the connection is made in a test.before() hook, not a test.beforeEach() hook. Thus the real concern is tests interleaving and the database being in an unknown or conflicting state.

@novemberborn LGTM 👍

novemberborn:
Thank you so much for your efforts @zellwk, @OmgImAlexis and @nodkz!

cyberwombat:
Hi all. I've written a module to handle parallel testing with Mongo. It's in early test but please check it out. https://github.com/cyberwombat/mongoprime

PR #1432, opened 2017-06-30
Include raw actual and expected objects in AssertionError

ArtemGovorov:
We support AVA in wallaby.js and have a few custom renderers for displaying assertions diffs in various editors, like:

VS Code:

screen shot 2017-06-30 at 12 09 25 pm

WebStorm:

screen shot 2017-06-30 at 12 11 50 pm

Wallaby App:

screen shot 2017-06-30 at 12 12 30 pm

Other testing frameworks that we support (also various assertion libs, such as `chai`) pass raw `actual` and `expected` objects with their assertion errors, so we (and other custom reporters) are able to display custom diffs.

AVA's new coloured ANSI diff display is great, but in order to display diffs outside of the ANSI supporting viewer (such as in the cases listed above), it would be great if the `AssertionError` contained raw `actual` and `expected` objects, so that a customer reporter integrated with AVA could use them for doing custom diffs/reporting (the screenshots above show how it'd look like once the PR is landed and published in AVA).

The pull request implements passing the raw `actual` and `expected` objects to the `AssertionError` constructor.

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -29,7 +29,7 @@ function formatWithLabel(label, value) {
 }
 
 class AssertionError extends Error {
-	constructor(opts, raw) {
+	constructor(opts) {
 		super(opts.message || '');
 		this.name = 'AssertionError';
 
@@ -42,7 +42,7 @@ class AssertionError extends Error {
 		// Raw expected and actual objects are stored for custom reporters
 		// (such as wallaby.js), that manage worker processes directly and
 		// use the values for custom diff views
-		this.raw = raw;
+		this.raw = opts.raw;
 
 		// Reserved for power-assert statements
 		this.statements = [];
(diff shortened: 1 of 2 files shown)


novemberborn:
> The pull request implements passing the raw actual and expected objects to the AssertionError constructor.

How does Wallaby access these properties? `AssertionError` is serialized so anything that isn't JSON will get lost as the test results make their way to the main AVA process.

ArtemGovorov:
@novemberborn Wallaby doesn't use AVA process. We create an instance of the `ava/lib/runner` class and use it to run tests. To get the test result we patch `addTestResult` and get the `test.assertError`.

ArtemGovorov:
@novemberborn Great, thanks. Made the requested changes and left the comment (new tests should also help not to remove the property accidentally).

lib/assert.js
```diff
@@ -29,7 +29,7 @@ function formatWithLabel(label, value) {
 }
 
 class AssertionError extends Error {
-	constructor(opts) {
+	constructor(opts, raw) {
```
novemberborn: Apologies, I meant passing `raw` as a property in `opts`.

ArtemGovorov:
@novemberborn Fair enough. Yep, we are aware that we are using AVA's internal APIs and things get broken sometimes, but thankfully our AVA users report them fairly quickly. Made the requested change.

novemberborn:
Thanks @ArtemGovorov!

PR #1447, opened 2017-07-12
Add more Vue.js recipes

suchmaske:
I have struggled the last days to properly test vue components using ava. To make life easier for others, I added some additional recipes for testing vuejs using ava. 

For example:

* testing the DOM of the component after a change in the data model
* testing data model changes when using async requests and promises
* testing DOM updates when promises are involved

### DIFF
--- a/docs/recipes/vue.md
+++ b/docs/recipes/vue.md
@@ -102,14 +102,13 @@ import test from 'ava';
 import Vue from 'vue';
 import Component from 'component.vue';
 
-test('it sees the updated message', t => {
+test('see the updated message', t => {
 
 	const vm = new Vue(Component).$mount();
 	t.is(vm.$el.textContent, 'my message'); 
 
 	vm.message = 'my new message';
 	// this fails here: t.is('my new message', vm.$el.textContent)
-
 
 	Vue.nextTick(() => {
 		t.is(vm.$el.textContent, 'my new message');
@@ -128,7 +127,7 @@ import Vue from 'vue';
 import Component from 'component.vue';
 
 // Don't forget to add the async prefix
-test('it updates the model using promises', async t => {
+test('update the model using promises', async t => {
 
 	// assume the message data property is a pending promise defined in the Component.
 	const vm = new Vue(Component).$mount();
--- a/docs/recipes/vue.md
+++ b/docs/recipes/vue.md
@@ -5,10 +5,12 @@ Translations: [Français](https://github.com/avajs/ava-docs/blob/master/fr_FR/do
 ## Dependencies
 
 - [Require extension hooks](https://github.com/jackmellis/require-extension-hooks):
-	- `npm i --save-dev require-extension-hooks require-extension-hooks-vue require-extension-hooks-babel`
+
+  `npm install --save-dev require-extension-hooks require-extension-hooks-vue require-extension-hooks-babel`
 
 - [browser-env](browser-testing.md)
-	- `npm i --save-dev browser-env`
+
+	`npm install --save-dev browser-env`
 
 ## Setup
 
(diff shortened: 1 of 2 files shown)


docs/recipes/vue.md
```diff
@@ -60,6 +80,71 @@ test('renders', t => {
...
+import Component from 'component.vue';
+
+test('see the updated message', t => {
+	const vm = new Vue(Component).$mount();
+	t.is(vm.$el.textContent, 'my message');
+
+	vm.message = 'my new message';
+	// this fails here: t.is('my new message', vm.$el.textContent)
+
+	Vue.nextTick(() => {
```
novemberborn: I assume this is asynchronous? That'll mean the assertion is run when the test is ended. We don't currently warn about that, but it means the assertion is ignored. You'd have to get a promise out of this or use `test.cb()` with `t.end()`.

sindresorhus: From the Vue docs:

> New in 2.1.0+: returns a Promise if no callback is provided and Promise is supported in the execution environment. - https://vuejs.org/v2/api/#Vue-nextTick

So you can just use an async function for the test and `await` it here.

docs/recipes/vue.md
```diff
@@ -60,6 +80,71 @@ test('renders', t => {
...
+
+// Don't forget to add the async prefix
+test('update the model using promises', async t => {
+	// assume the message data property is a pending promise defined in the Component.
+	const vm = new Vue(Component).$mount();
+
+	// just testing the data model can be achieved by using await
+	t.is(await vm.message, 'my message');
+
+	await vm.message.then((data) => {
```
sindresorhus: The `.then()` here is moot. You're already `await`'ing it. Just get value back the async/await way.

docs/recipes/vue.md
```diff
@@ -36,9 +38,27 @@ const Vue = require('vue');
 // Setup Vue.js to remove production tip
 Vue.config.productionTip = false;
 
+// Setup Vue files to be processed by `require-extension-hooks-vue`
+hooks('vue').plugin('vue').push();
+// Setup Vue and JS files to be processed by `require-extension-hooks-babel`
+hooks(['vue', 'js']).plugin('babel').push();
+```
+
+When you are using ES6 modules, use the dist-file of Vue. Not doing that may cause a not completely accessible (and thus not testable) Vue model.
```
sindresorhus: ES6 => ES2015

docs/recipes/vue.md
```diff
@@ -36,9 +38,27 @@ const Vue = require('vue');
...
+
+```js
+import browserEnv from 'browser-env';
+import hooks from 'require-extension-hooks';
+import Vue from 'vue/dist/vue';
```
OmgImAlexis: Does this type of import not require webpack? I didn't think this worked out of the box with babel, etc. yet.

sindresorhus:
@suchmaske @blake-newman @OmgImAlexis @knpwrs ping :)

docs/recipes/vue.md
blake-newman: This is not true. Using the hooks compiles the es6 to es2015. You can use the source components.

docs/recipes/vue.md
blake-newman: Not sure of the point of this example, same as above but es imports. 

@sindresorhus can you confirm @OmgImAlexis is also correct.

docs/recipes/vue.md
blake-newman: Agree with using async and await syntax.

docs/recipes/vue.md
```diff
@@ -60,6 +80,71 @@ test('renders', t => {
...
+import Component from 'component.vue';
+
+// Don't forget to add the async prefix
+test('update the model using promises', async t => {
+	// assume the message data property is a pending promise defined in the Component.
+	const vm = new Vue(Component).$mount();
+
+	// just testing the data model can be achieved by using await
+	t.is(await vm.message, 'my message');
+
```
blake-newman: Why would vm.message be a promise? No need to `await` this

knpwrs:
I think it's worth pointing out that [Vue has started work on official testing tools](https://github.com/vuejs/vue-test-utils). They have an [example with AVA](https://github.com/eddyerburgh/vue-test-utils-ava-example) available.

blake-newman:
@knpwrs Indeed. Will create a PR to link to that example repository

novemberborn:
@blake-newman @knpwrs is this PR still relevant?

@suchmaske will you have time to revive this PR?

suchmaske:
@novemberborn I recently have changed my workflow to [vue-test-utils](https://github.com/vuejs/vue-test-utils). I will take the feedback and try to rework the PR in the next days.

knpwrs:
I don't personally use Vue enough anymore to offer any official recommendations.

novemberborn:
That's great, thank you @suchmaske.

blake-newman:
@novemberborn i don't think the PR is relevant. Also there is many examples, via `vue-test-utils` that enhance what is on offering here. 

Sorry for the late reply

sindresorhus:
Closing this for lack of activity. Happy to reopen if anyone wants to finish it.

PR #1448, opened 2017-07-12
Resolve missing Buffer.alloc function on older versions of Node.js

### LINKED ISSUE #1446: `Buffer.alloc` is not a function
## Description
Seems like the `Buffer.alloc` api was introduced in node v5.10+. Since I'm using node v5.9.1, and ava's `package.json` specifies valid engines as `">=4"`, I'll have to upgrade my node install to play with ava. Running a silly test throws a NAF error as shown below.

I know I can upgrade my node to 5.10.0 and keep going. Just reporting this in case there are others facing this problem.

Moving forward, should the `package.json`'s `engines` key to at least `5.10.0` of node?
Or keep the `engines` as it is, and change the `Buffer.alloc` call to something suitable for older node versions?

### Test code
```
const test = require('ava')
const users = require('../resources/users.methods')

test('should return all users in db', async function (t) {
  const fetchedUsers = await users.getAll()
  t.truthy(fetchedUsers.length)
})
```

```
$ ava
D:\Work\server-express\node_modules\ava\lib\snapshot-manager.js:21
const VERSION_HEADER = Buffer.alloc(2);
                              ^

TypeError: Buffer.alloc is not a function
    at Object.<anonymous> (D:\Work\server-express\node_modules\ava\lib\snapshot-manager.js:21:31)
    at Module._compile (module.js:413:34)
    at Object.Module._extensions..js (module.js:422:10)
    at Module.load (module.js:357:32)
    at Function.Module._load (module.js:314:12)
    at Module.require (module.js:367:17)
    at require (internal/module.js:16:19)
    at Object.<anonymous> (D:\Work\server-express\node_modules\ava\lib\assert.js:10:25)
    at Module._compile (module.js:413:34)
    at Object.Module._extensions..js (module.js:422:10)
    at Module.load (module.js:357:32)
    at Function.Module._load (module.js:314:12)
    at Module.require (module.js:367:17)
    at require (internal/module.js:16:19)
    at Object.<anonymous> (D:\Work\server-express\node_modules\ava\lib\serialize-error.js:5:16)
    at Module._compile (module.js:413:34)

  1 exception

  × routes\users\users.methods.test.js exited with a non-zero exit code: 1
```

### Config

Defaults. No custom config in `package.json`

### Command-Line Arguments

```
ava
```

## Environment

```
Node.js v5.9.1
win32 6.1.7601
ava 0.20.0
```

ajtorres9:
Fixes #1446 

**NOTE:** I incorporated the `safe-buffer` package as specified in the issue, but when the tests are run in `node@4.3` the CLI test "watcher does not rerun test files when they write snapshot files" timeouts. To test if it was due to slow processing, I increased the the test timeout to 8 minutes and it still failed. I am not entirely sure how to address this issue and would appreciate any guidance.

### DIFF
--- a/media/screenshot-fixtures/magic-assert-buffers.js
+++ b/media/screenshot-fixtures/magic-assert-buffers.js
@@ -1,5 +1,4 @@
 import test from 'ava';
-import {Buffer} from 'safe-buffer';
 
 test('buffers', t => {
 	const actual = Buffer.from('decafbadcab00d1e'.repeat(4), 'hex')
--- a/media/screenshot-fixtures/magic-assert-buffers.js
+++ b/media/screenshot-fixtures/magic-assert-buffers.js
@@ -1,4 +1,5 @@
 import test from 'ava';
+import {Buffer} from 'safe-buffer';
 
 test('buffers', t => {
 	const actual = Buffer.from('decafbadcab00d1e'.repeat(4), 'hex')
--- a/package-lock.json
+++ b/package-lock.json
@@ -2,7 +2,6 @@
   "name": "ava",
   "version": "0.20.0",
   "lockfileVersion": 1,
-  "requires": true,
   "dependencies": {
     "@ava/babel-plugin-throws-helper": {
       "version": "2.0.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -2,6 +2,7 @@
   "name": "ava",
   "version": "0.20.0",
   "lockfileVersion": 1,
+  "requires": true,
   "dependencies": {
     "@ava/babel-plugin-throws-helper": {
       "version": "2.0.0",
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ node_js:
   - 8
   - 6
   - 4
-  - 4.3
 env:
   - FRESH_DEPS=false
   - FRESH_DEPS=true
@@ -12,8 +11,6 @@ matrix:
     - node_js: 6
       env: FRESH_DEPS=true
     - node_js: 4
-      env: FRESH_DEPS=true
-    - node_js: 4.3
       env: FRESH_DEPS=true
 cache:
   directories:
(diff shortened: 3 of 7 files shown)


package-lock.json
```diff
@@ -2,7 +2,6 @@
   "name": "ava",
   "version": "0.20.0",
   "lockfileVersion": 1,
-  "requires": true,
```
novemberborn: Could you fix it so this stays in the `package-lock.json`?

media/screenshot-fixtures/magic-assert-buffers.js
```diff
@@ -1,4 +1,5 @@
 import test from 'ava';
+import {Buffer} from 'safe-buffer';
```
novemberborn: Let's not worry about this file, it's used by maintainers so we won't run it on Node.js 4 😉

novemberborn:
> NOTE: I incorporated the safe-buffer package as specified in the issue, but when the tests are run in node@4.3 the CLI test "watcher does not rerun test files when they write snapshot files" timeouts. To test if it was due to slow processing, I increased the the test timeout to 8 minutes and it still failed. I am not entirely sure how to address this issue and would appreciate any guidance.

I can't reproduce that locally, so it's probably something specific to the CI environment. Let's not run these tests in 4.3 then, it's not worth the hassle.

novemberborn:
> I can't reproduce that locally

Ah actually I can. Some confusion with my version manager.

Still, I can't say I'm interested in debugging those issues. We're just doing this to help Lambda users, otherwise people should really be on the latest Node.js 4 release.

novemberborn:
@ajtorres9 I'm hoping to get an AVA release out today, so I took the liberty of acting on my suggestions 👼

novemberborn:
Thanks @ajtorres9!

ajtorres9:
Glad I could help @novemberborn 😄

Issue #420, opened 2016-01-13
Group log output with the relevant test.


jamestalmage:
Continuation from [here](https://github.com/sindresorhus/ava/issues/392#issuecomment-168344251).

> 1. Use domains to determine which test `console.log` is used from.
> 2. Provide a `t.log` method.

`2` is a less controversial choice, but won't help for logging from production code.

One possibility for such grouping would be to only display the logs when the test failed. I think this is an especially a good idea for `t.log`, as it lets you leave them in your tests without consequence if you think they will help troubleshoot future failures.

vdemedes:
> only display the logs when the test failed

I'm strongly against this one as it's very irritating when obvious & simple things (like `console.log()`) don't work as expected.

Domains looks like a good solution for our issues. We could also group exceptions & rejections, not just console.log()s. It was mentioned before in some issue too.

The problem with domains is this:

<img width="728" alt="screen shot 2016-01-13 at 9 35 50 pm" src="https://cloud.githubusercontent.com/assets/697676/12307115/a4b15692-ba3d-11e5-80d7-105afc383fc5.png">

But if a **replacement** will be provided, that shouldn't be an issue to just migrate to a new API.

vdemedes:
Am I crazy or we could use `cluster` module for that? It allows to listen to `error` events when worker (test) fails. Also we could attach to `message` events and:

```js
console.log = function () {
  process.send(...);
};
```

sindresorhus:
> I'm strongly against this one as it's very irritating when obvious & simple things (like console.log()) don't work as expected.

:+1: I also want them visible by default.

> Provide a t.log method.

Do we really need this if we manage to route console calls correctly?

> The problem with domains is this:

Domains are deprecated, but so much depend on it that they're not going to remove it anytime soon (source: Node.js core people). They don't even have an alternative yet. And the alternative is going to be powerful enough to enable building a user-land implementation of domains. More here: https://github.com/sindresorhus/ava/issues/214#issuecomment-157402230.

jamestalmage:
> only display the logs when the test failed

>> I'm strongly against this one as it's very irritating when obvious & simple things (like console.log()) don't work as expected.

You are probably right. I was more thinking this for `t.log`. My only irritation is that if I put `console.log` in a production file that is called by a lot of tests, I probably only want to see the output of the failing ones. Still - probably best not to.

vdemedes:
In that case, let's move on with domains on board.

sindresorhus:
[...]

PR #1452, opened 2017-07-13
Add support for test logging

nowells:
There are times when you wish to include comments in your tests that will be shown in context of the test result and not streamed to `stdout` like `console.log` statements are, where they will likely be out of order of the tests they are associated with.

I will happily add tests, and change the API, etc. I just want to get general 👍 or 👎 to the concept of adding comments before I flesh out the tests and other cleanup.

One of the main use cases I have for this, is I am using ava as a runner for my SauceLabs/BrowserSrack/Selenium tests, and I would like to have the link to the job results from SauceLabs/BrowserStack be output in context with the test run. There are certainly many other usecases for this, however, just to give a relevant example of when this is useful.

### `--verbose`
1__tmux_and_zoom_meeting_id__571-235-5176

### `--tap`
```
TAP version 13                                                                                                                                                              [30/1823]
# file-one › passing test
ok 1 - file-one › passing test
  a comment for passing test
# file-two › passing test
ok 2 - file-two › passing test
  a comment for passing test
# file-one › failing test
not ok 3 - file-one › failing test
  ---
    name: AssertionError
    assertion: is
    values:
      'Difference:': |-
        - true
        + false
    at: 'Test.t [as fn] (file-one.test.js:11:7)'
  ...
  a comment for failing test
# file-two › failing test
not ok 4 - file-two › failing test
  ---
    name: AssertionError
    assertion: is
    values:
      'Difference:': |-
        - true
        + false
    at: 'Test.t [as fn] (file-two.test.js:11:7)'
  ...
  a comment for failing test

1..4
# tests 4
# pass 2
# fail 2
```

### DIFF
--- a/lib/reporters/mini.js
+++ b/lib/reporters/mini.js
@@ -5,6 +5,7 @@ const lastLineTracker = require('last-line-stream/tracker');
 const plur = require('plur');
 const spinners = require('cli-spinners');
 const chalk = require('chalk');
+const figures = require('figures');
 const cliTruncate = require('cli-truncate');
 const cross = require('figures').cross;
 const indentString = require('indent-string');
@@ -163,6 +164,15 @@ class MiniReporter {
 				}
 
 				status += '  ' + colors.title(test.title) + '\n';
+
+				if (test.logs) {
+					test.logs.forEach(log => {
+						status += '  ' + colors.information(figures.info) + ' ' + colors.log(log) + '\n';
+					});
+
+					status += '\n';
+				}
+
 				if (test.error.source) {
 					status += '  ' + colors.errorSource(test.error.source.file + ':' + test.error.source.line) + '\n';
 
--- a/lib/reporters/mini.js
+++ b/lib/reporters/mini.js
@@ -167,7 +167,13 @@ class MiniReporter {
 
 				if (test.logs) {
 					test.logs.forEach(log => {
-						status += '  ' + colors.information(figures.info) + ' ' + colors.log(log) + '\n';
+						const logLines = indentString(colors.log(log), 6);
+						const logLinesWithLeadingFigure = logLines.replace(
+							/^ {6}/,
+							`    ${colors.information(figures.info)} `
+						);
+
+						status += logLinesWithLeadingFigure + '\n';
 					});
 
 					status += '\n';
--- a/lib/reporters/tap.js
+++ b/lib/reporters/tap.js
@@ -49,7 +49,7 @@ class TapReporter {
 		return 'TAP version 13';
 	}
 	test(test) {
-		let output;
+		const output = [];
 
 		let directive = '';
 		const passed = test.todo ? 'not ok' : 'ok';
@@ -62,23 +62,23 @@ class TapReporter {
 
 		const title = stripAnsi(test.title);
 
-		if (test.error) {
-			output = [
-				'# ' + title,
-				format('not ok %d - %s', ++this.i, title),
-				dumpError(test.error, true)
-			];
-		} else {
-			output = [
-				`# ${title}`,
-				format('%s %d - %s %s', passed, ++this.i, title, directive).trim()
-			];
-		}
+		const appendLogs = () => {
+			if (test.logs) {
+				test.logs.forEach(log => {
+					output.push(`  ${log}`);
+				});
+			}
+		};
 
-		if (test.comments) {
-			test.comments.forEach(comment => {
-				output.push(`  ${comment}`);
-			});
+		output.push(`# ${title}`);
+
+		if (test.error) {
+			output.push(format('not ok %d - %s', ++this.i, title));
+			appendLogs();
+			output.push(dumpError(test.error, true));
+		} else {
+			output.push(format('%s %d - %s %s', passed, ++this.i, title, directive).trim());
+			appendLogs();
 		}
 
 		return output.join('\n');
--- a/lib/reporters/tap.js
+++ b/lib/reporters/tap.js
@@ -65,7 +65,13 @@ class TapReporter {
 		const appendLogs = () => {
 			if (test.logs) {
 				test.logs.forEach(log => {
-					output.push(`  ${log}`);
+					const logLines = indentString(log, 4);
+					const logLinesWithLeadingFigure = logLines.replace(
+						/^ {4}/,
+						'  * '
+					);
+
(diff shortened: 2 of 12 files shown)


lib/reporters/verbose.js
```diff
@@ -24,29 +24,32 @@ class VerboseReporter {
 		return '';
 	}
 	test(test, runStatus) {
+		const lines = [];
 		if (test.error) {
-			return '  ' + colors.error(figures.cross) + ' ' + test.title + ' ' + colors.error(test.error.message);
```
nowells: This whole stream of changes looks more noisy than it is. This is simply switching from an immediate `return` approach to appending to an output array so that we can concat comments after the test summary.

novemberborn:
Looks like you've inadvertently implemented #420! 👍 

This looks great. I think it should work with the (default) mini reporter too, since that still prints failing tests.

Bike shedding consensus from #420 was to use `t.log()` rather than `t.comment()`.

nowells:
@novemberborn thanks for the review! I will:

* [ ] Switch from `commen`t to `log`
* [ ] Add log output to mini reporter as well on failure
* [ ] Add tests for `t.log`

I should hopefully have this up to you for review by the end of the day. Cheers!

nowells:
@novemberborn let me know if my rework looks good, and if you think there are other types of tests that I should add. Thanks!

sindresorhus:
@nowells Can you document it in the readme?

sindresorhus:
> I think it should work with the (default) mini reporter too, since that still prints failing tests.

How would that work? The comments would need context.

nowells:
> Can you document it in the readme?

Absolutely! Done.

> How would that work? The comments would need context.

The way I implemented it, in `mini` reporter the log messages show up in failure cases (arguably the only place you really care to see them in that context) but in success cases I do not display them.

lib/reporters/tap.js
```diff
@@ -62,17 +62,23 @@ class TapReporter {
 
 		const title = stripAnsi(test.title);
 
+		const appendLogs = () => {
+			if (test.logs) {
+				test.logs.forEach(log => {
+					output.push(`  ${log}`);
```
novemberborn: If `log` contains linebreaks, should the indentation apply on those new lines? If so, this should use https://www.npmjs.com/package/indent-string (like elsewhere in the reporters). May not be a bad idea to use that anyway.

lib/reporters/verbose.js
```diff
@@ -104,6 +107,15 @@ class VerboseReporter {
 				}
 
 				output += '  ' + colors.title(test.title) + '\n';
+
+				if (test.logs) {
+					test.logs.forEach(log => {
+						output += '    ' + colors.information(figures.info) + ' ' + colors.log(log) + '\n';
```
novemberborn: Same question here around linebreaks in the `log` value.

readme.md
```diff
@@ -875,6 +875,10 @@ Plan how many assertion there are in the test. The test will fail if the actual
 
 End the test. Only works with `test.cb()`.
 
+###### `t.log(message)`
+
+Print a log message contextually alongside the test result instead of immediately streaming the message to stdout like `console.log`.
```
novemberborn: Along the lines of https://nodejs.org/api/console.html#console_console_log_data_args, perhaps:

> Print a log message contextually alongside the test result instead of immediately printing it to `stdout` like `console.log`.

lib/reporters/tap.js
sindresorhus: > should the indentation apply on those new lines?

Yes

nowells: Good catch! I'm going to make it so that it will result in something like this, where I add the `i` figure to the first line.

```
  ✔ file-one › passing test
    ℹ a comment for passing test
      with a newline
    ℹ a comment for passing test without a newline
```

readme.md
nowells: Done

nowells:
The tests that are failing appear to be failing on master as well?

lib/reporters/mini.js
```diff
@@ -163,6 +164,21 @@ class MiniReporter {
...
 
 				status += '  ' + colors.title(test.title) + '\n';
+
+				if (test.logs) {
+					test.logs.forEach(log => {
+						const logLines = indentString(colors.log(log), 6);
+						const logLinesWithLeadingFigure = logLines.replace(
+							/^ {6}/,
+							`    ${colors.information(figures.info)} `
+						);
```
novemberborn: This might be better as:

```js
const logLines = indentString(colors.log(log), 1, {
  indent: `    ${colors.information(figures.info)} `,
  includeEmptyLines: true
});
```

(Though that `indent` value should probably be a constant, but feel free to redeclare it in the `verbose` logger.)

See [`indent-string`'s API](https://www.npmjs.com/package/indent-string#api). Including empty lines would be good given the figure prefix.

You'll have to update the `indent-string` dependency to `3.2.0` since the `includeEmptyLines` feature is pretty new.

novemberborn:
> The tests that are failing appear to be failing on master as well?

I'm hoping https://github.com/avajs/ava/pull/1477 fixes that.

lib/reporters/mini.js
nowells: @novemberborn the reason I didn't do that was I only wanted to `i` symbol to show up once per log line (to indicate where a log message starts)

```
    * this is a message that will have multiple lines, but there only the first line
      will have the `*` to indicate a log message, and the indent of the other lines
      is at the indent level of the first log message (after the `*`)
    * Some other log message
```

Does that make sense? Or would you rather it be

```
    * this is a message that will have multiple lines, but there only the first line
    * will have the `*` to indicate a log message, and the indent of the other lines
    * is at the indent level of the first log message (after the `*`)
    * Some other log message, but how do I know that.
```

novemberborn: Oh! No that's fine, I misread the code.

I suppose the replacement string could be a constant, but on the other hand it's clearer if it's near the indentation values. So hey :)

nowells:
thanks for the great feedback and guiding me through this @novemberborn!

sindresorhus:
Thanks for working on this @nowells 🙌

nowells:
@sindresorhus thanks for making such a wonderful testing system!!! I look forward to helping out on other features as well. :heart:

PR #1455, opened 2017-07-15
Do not clobber forked process color level

nowells:
Chalk uses `supports-color` to determine what color level should be
rendered. However, currently our parent `ava` process uses that to
colorize its output, however, when we fork our test processes we clobber
the color configuration that `supports-color` uses to determine the
level (see https://github.com/chalk/supports-color/blob/master/index.js#L37-L42)

This PR carries the current color level through to the forked processes
so if they log output that is colorized it will remain so with the same
fidelity of colors as the parent ava process.

### DIFF
--- a/lib/fork.js
+++ b/lib/fork.js
@@ -4,7 +4,16 @@ const path = require('path');
 const fs = require('fs');
 const Promise = require('bluebird');
 const debug = require('debug')('ava');
+const supportsColor = require('supports-color');
 const AvaError = require('./ava-error');
+
+const colorLevels = {
+	0: 'false',
+	1: 'true',
+	2: '256',
+	3: '16m'
+};
+const currentColorLevel = colorLevels[supportsColor.level];
 
 if (fs.realpathSync(__filename) !== __filename) {
 	console.warn('WARNING: `npm link ava` and the `--preserve-symlink` flag are incompatible. We have detected that AVA is linked via `npm link`, and that you are using either an early version of Node 6, or the `--preserve-symlink` flag. This breaks AVA. You should upgrade to Node 6.2.0+, avoid the `--preserve-symlink` flag, or avoid using `npm link ava`.');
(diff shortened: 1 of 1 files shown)


novemberborn:
Interesting. #1401 aims to forward `--color` only if the user has explicitly provided it. See also the discussion in #1393.

Of course the forked processes still can't detect the color support, and we can't necessarily expect the user to know what to specify either. Does this mean we should provide a value for `--color` when the user provides it without a value in the `ava` invocation?

@sindresorhus, @kevva?

sindresorhus:
> Does this mean we should provide a value for --color when the user provides it without a value in the ava invocation?

Yes, the forked process should get the exact same color level as AVA.

novemberborn:
>> Does this mean we should provide a value for --color when the user provides it without a value in the ava invocation?
>
> Yes, the forked process should get the exact same color level as AVA.

Re-reading my original comment I'm not exactly sure what I meant! I think the conclusion is:

* Only pass `--color` to the worker process if it was provided to the main AVA process
* If it was provided with a value, pass it exactly as used
* If it was provided without a value, have AVA detect color support, and pass it with the appropriate value so that the worker process is configured with the same level of support

@sindresorhus did I get this right?

sindresorhus:
> If it was provided without a value, have AVA detect color support, and pass it with the appropriate value so that the worker process is configured with the same level of support

This should also happen if no `--color` flag was provided.

---

And it should also pass the `--no-color` flag.

novemberborn:
>> If it was provided without a value, have AVA detect color support, and pass it with the appropriate value so that the worker process is configured with the same level of support
>
> This should also happen if no --color flag was provided.

@sindresorhus that's not my understanding based on our discussion in https://github.com/avajs/ava/pull/1401#issuecomment-307383459.

sindresorhus:
@novemberborn It should just work as if the child process was run in the same process as AVA.

novemberborn:
> It should just work as if the child process was run in the same process as AVA.

Yes, but per https://github.com/avajs/ava/issues/1393 we should only pass `--color` if it was provided to AVA. We need a different way of enabling colors if the flag was not provided. And I'm still not sure if we should rewrite the `--color` flag if it was provided to AVA without a value.

novemberborn:
#1700 clears up what arguments are exposed to the test worker. Let's discuss the specifics around configuring color in #1701.

Issue #1460, opened 2017-07-18
Feature Request: Export current module from AVA


lukechilds:
It gets kind of annoying having to import the module you want to test with:

```js
import module from '../';
```

Especially if you move a test into a sub folder and then have to change it to:

```js
import module from '../../';
```

I think it would be really neat if AVA could expose the current module itself, so:

```js
import test from 'ava';
import module from '../../';
```

Would become:

```js
import { test, module } from 'ava';
```

Happy to put together a PR if this is something you're interested in.

120687:
Id line me :rindupoma
Send me contact account protect bot .please..

Pada tanggal 18 Jul 2017 18:55, "Luke Childs" <notifications@github.com>
menulis:

ORESoftware:
Interesting

lukechilds:
Just put this together to demo the functionality: https://github.com/lukechilds/get-root-module

No docs or anything yet but pretty obvious what it's doing from the source.

I'll submit a PR to demo the functionality in AVA.

ORESoftware:
@lukechilds I am looking at the source for get root module:

https://github.com/lukechilds/get-root-module/blob/master/src/index.js

not really sure what it's doing 

```js
const findRoot = require('find-root');

try {
	const modulePath = findRoot(process.cwd());
	module.exports = require(modulePath);
} catch (err) {
	module.exports = undefined;
}
```

I looked at the find-root project, and it's just looking for the project root directory, so you are looking to find the "index" file of a project?

lukechilds:
@ORESoftware I explained in more detail in the PR #1461

> Traverses up the directory tree from process.cwd() (whatever dir ava was executed from on the cli). When a dir with a package.json is found, that dir is required, if no package.json is found module will be undefined.

ORESoftware:
@lukechilds ok bro so that's cool

one alternative this, is linking your project to itself with npm link:

```bash
cd <your-project-root> && npm link
npm link your-project
```

that way you can require('your-project') from anywhere, including your test scripts

ORESoftware:
@lukechilds there is actually a small problem with the PR

in NPM land, requiring a directory will first look at

`{package.json}.main`

if that entry is not there, it will look for `index.js`

if that's not there, no module can be returned/required, so you'd have to handle the case where nothing can be loaded

novemberborn:
This would indeed be useful, though users could use `get-root-module` directly, right?

[...]

PR #1461, opened 2017-07-19
Export root module

lukechilds:
Resolves https://github.com/avajs/ava/issues/1460

Needs tests and I need to add docs to `get-root-module` but this is just to demo my proposal in https://github.com/avajs/ava/issues/1460.

Allows you to do:

```js
import { test, module } from 'ava';
```

Rather than having to do:

```js
import test from 'ava';
import module from '../';

// and then in sub folders
import module from '../../';
```

Traverses up the directory tree from `process.cwd()` (whatever dir `ava` was executed from on the cli). When a dir with a `package.json` is found, that dir is `require`d, if no package.json is found `module` will be undefined.

I'll wait to see if you're interested in merging this before going any further.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -1,4 +1,5 @@
 'use strict';
+const rootModule = require('get-root-module');
 const worker = require('./test-worker');
 const adapter = require('./process-adapter');
 const serializeError = require('./serialize-error');
@@ -102,3 +103,5 @@ module.exports = runner.chain;
 // an ES2015 default import (`import test from 'ava'`)
 // See: https://github.com/Microsoft/TypeScript/issues/2242#issuecomment-83694181
 module.exports.default = runner.chain;
+
+module.exports.module = rootModule;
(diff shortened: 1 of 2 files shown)


sindresorhus:
AVA already knows the root, so we could just expose that instead. (See `projectDir` [here](https://github.com/avajs/ava/blob/31b1380194cec664b8816bb31e333cba64ff27d9/lib/cli.js#L143)) But let's discuss whether this is a good idea or not in https://github.com/avajs/ava/issues/1460 first.

lukechilds:
```js
const conf = pkgConf.sync('ava');

const filepath = pkgConf.filepath(conf);
const projectDir = filepath === null ? process.cwd() : path.dirname(filepath);
```

Wouldn't that default to `process.cwd()` if there's no AVA conf? Which would mean if I was in `/project-root/src/helpers` and then ran `npm test` that dir would be `require`d rather than traversing up to `/project-root`?

sindresorhus:
No, it defaults to the first package.json

sindresorhus:
Closing this in favor of a recipe for now.

PR #1467, opened 2017-07-24
Limit concurrency to the number of CPU cores

### LINKED ISSUE #966: Default concurrency
In https://github.com/avajs/ava/pull/791, we added a flag to control the concurrency. It turned out operating systems are really bad at doing their job properly, and they couldn't handle the amount of processes we tried to start up when there's a lot of test files. We also have the overhead starting up `node` and requiring dependencies in each process. This was a big problem in React projects where you do one test file per component, and you usually have a lot of components.

We should consider doing a default concurrency limit. Users should still be able to set it to unlimited, but I think it would be in our best interest to adapt to reality. I've previously had good success with `(os.cpus().length || 1) * 2`, but we should get some real numbers on it from various systems.

~~**Everyone, please run your test suite with the `--concurrency` flag and try out different numbers for it and post back the results. Also include the number of test files in the suite and the output of:**~~

Edit: Let's wait on this until https://github.com/avajs/ava/pull/945 is merged and released, per https://github.com/avajs/ava/issues/966#issuecomment-233035952.

```
node -p "require('os').cpus().length"
```
## 

@avajs/core Thoughts?

sindresorhus:
Fixes #966

This considerably improves the performance for the common case. Most users have many test files and their CPUs got overloaded. And few knew about the concurrency option.

This has the downside of making `test.only()` not be exclusive with multiple test files, but there are just too many limitations caused by it. I think this change is worth it.

I've chosen to go with a default concurrency of the number of CPU cores. We can tweak it based on real usage data later on.

### DIFF
--- a/api.js
+++ b/api.js
@@ -2,6 +2,7 @@
 const EventEmitter = require('events');
 const path = require('path');
 const fs = require('fs');
+const os = require('os');
 const commonPathPrefix = require('common-path-prefix');
 const uniqueTempDir = require('unique-temp-dir');
 const findCacheDir = require('find-cache-dir');
@@ -160,16 +161,17 @@ class Api extends EventEmitter {
 					this._setupTimeout(runStatus);
 				}
 
-				let overwatch;
+				let concurrency = Math.max(2, os.cpus().length || 1);
+
 				if (this.options.concurrency > 0) {
-					const concurrency = this.options.serial ? 1 : this.options.concurrency;
-					overwatch = this._runWithPool(files, runStatus, concurrency);
-				} else {
-					// _runWithoutPool exists to preserve legacy behavior, specifically around `.only`
-					overwatch = this._runWithoutPool(files, runStatus);
+					concurrency = this.options.concurrency;
 				}
 
-				return overwatch;
+				if (this.options.serial) {
+					concurrency = 1;
+				}
+
+				return this._runWithPool(files, runStatus, concurrency);
 			});
 	}
 	_computeForkExecArgs(files) {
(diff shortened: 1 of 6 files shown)


vadimdemedes:
I'm so looking forward to this change!

novemberborn:
> This has the downside of making test.only() not be exclusive with multiple test files, but there are just too many limitations caused by it. I think this change is worth it.

We should follow `tap` and implement a `--only` flag, which then causes only `test.only()` tests to run. See http://www.node-tap.org/cli/.

api.js
```diff
@@ -160,16 +161,17 @@ class Api extends EventEmitter {
 					this._setupTimeout(runStatus);
 				}
 
-				let overwatch;
+				let concurrency = Math.max(2, os.cpus().length || 1);
```
novemberborn: Could you add a comment explaining the logic here? Like, why use `2` on single-core machines?

When would `os.cpus()` return an empty array?

docs/common-pitfalls.md
```diff
@@ -16,6 +16,8 @@ AVA uses [is-ci](https://github.com/watson/is-ci) to decide if it's in a CI envi
 
 You may be using a service that only allows a limited number of concurrent connections. For example, many database-as-a-service businesses offer a free plan with a limit on how many clients can be using it at the same time. AVA can hit those limits as it runs multiple processes, but well-written services should emit an error or throttle in those cases. If the one you're using doesn't, the tests will hang.
 
+By default, AVA will use as many processes as there are CPU cores in your machine.
```
novemberborn: More actually.

readme.md
```diff
@@ -400,7 +400,7 @@ test.only('will be run', t => {
 });
 ```
 
-`.only` applies across all test files, so if you use it in one file, no tests from the other file will run.
+*Note:* The `.only` modifier applies to the test file it's defined in, so if you run multiple test files, tests in other files will stil run. If you want to only run the `test.only` test, specify just that test file to AVA.
```
novemberborn: Typo in `stil run`.

Rather than `specify just that test file to AVA`, say `provide just that test file to AVA`?

api.js
sindresorhus: > Could you add a comment explaining the logic here? Like, why use 2 on single-core machines?

Will do. It's to make it still concurrent on VMs that report only one CPU: https://github.com/sindresorhus/grunt-concurrent/commit/ef076acedd2bfcac84f9ce038d84af884b12d841 Does that make sense?

> When would os.cpus() return an empty array?

Some years ago someone reported an issue about it returning an empty array for a VM or something. I can't find the issue now, so let's just remove it and see.

readme.md
sindresorhus: 👍 Fixed

api.js
sindresorhus: Actually, I'm just going to remove it. Too edge-casey and magic.

sindresorhus:
> We should follow tap and implement a --only flag, which then causes only test.only() tests to run. See http://www.node-tap.org/cli.

Agreed → https://github.com/avajs/ava/issues/1472

sindresorhus:
@novemberborn Anything else? Travis is failing on `FRESH_DEPS=true`, but [so is master](https://travis-ci.org/avajs/ava/jobs/253231510), so I don't think that's relevant.

novemberborn:
> Travis is failing on FRESH_DEPS=true, but so is master, so I don't think that's relevant.

I'm hoping https://github.com/avajs/ava/pull/1477 fixes that.

PR #1477, opened 2017-07-30
Use --verbose when testing CLI output

novemberborn:
It cannot be guaranteed that all lines output by the mini reporter are
captured in the stderr of the CLI invocation during testing. Use
--verbose when matching output that the mini reporter may otherwise
rewrite.

Somewhere on at least Node.js 8.2.1, when the package-lock.json is
discarded prior to installation, something in our dependency tree is
different which results in the test title no longer being in the stderr
of the AVA CLI.

Compare:

```
⠋ ESC[2KESC[1AESC[2KESC[G
⠙ ESC[2KESC[1AESC[2KESC[G
⠹ ESC[2KESC[1AESC[2KESC[G
⠸ ESC[2KESC[1AESC[2KESC[G
⠼ ESC[2KESC[1AESC[2KESC[G
⠴ ESC[2KESC[1AESC[2KESC[G
⠦ ESC[2KESC[1AESC[2KESC[G
⠦ test › works…

 1 passedESC[2KESC[1AESC[2KESC[1AESC[2KESC[1AESC[2KESC[G
 1 passed [18:32:54]
```

Versus:

```
⠋ ESC[2KESC[1AESC[2KESC[G
⠙ ESC[2KESC[1AESC[2KESC[G
⠹ ESC[2KESC[1AESC[2KESC[G
⠸ ESC[2KESC[1AESC[2KESC[G
⠼ ESC[2KESC[1AESC[2KESC[G
⠴ ESC[2KESC[1AESC[2KESC[G
⠴ …

 1 passedESC[2KESC[1AESC[2KESC[1AESC[2KESC[1AESC[2KESC[G
⠦ …

 1 passedESC[2KESC[1AESC[2KESC[1AESC[2KESC[1AESC[2KESC[G
 1 passed [18:35:49]
```

### DIFF
--- a/test/cli.js
+++ b/test/cli.js
@@ -365,7 +365,7 @@ test('watcher reruns test files when snapshot dependencies change', t => {
 test('`"tap": true` config is ignored when --watch is given', t => {
 	let killed = false;
 
-	const child = execCli(['--watch', 'test.js'], {dirname: 'fixture/watcher/tap-in-conf'}, () => {
+	const child = execCli(['--watch', '--verbose', 'test.js'], {dirname: 'fixture/watcher/tap-in-conf'}, () => {
 		t.ok(killed);
 		t.end();
 	});


novemberborn:
See https://travis-ci.org/avajs/ava/jobs/258140018 for a failing test.

sindresorhus:
I guess it's caused by https://github.com/sindresorhus/cli-truncate/commit/f379a0ea694135dd58e437f8131102af8dd20c0f

novemberborn:
Maybe. I suppose something removes spurious output that would get rewritten immediately? I don't think it's a problem if the mini reporter doesn't write that to stdout, and it only broke one test.

Not sure if we should explicitly bump any dependencies?

Regardless will merge this to unblock other PRs.

sindresorhus:
> Not sure if we should explicitly bump any dependencies?

Nah

PR #1478, opened 2017-07-31
Fail hard when `--concurrency` is set to invalid values

### LINKED ISSUE #1476: Fail when `--concurrency` is set to `0`
Inspired by #1463. `--concurrency=0` currently means AVA spins up as many workers as there are test files, and when #1467 lands it means AVA spins up as many workers as there are CPU cores. In either case, we should refuse to run tests when the value is set to `0` since AVA's behavior won't match user expectations.

See a similar test in `lib/cli.js`: https://github.com/avajs/ava/blob/a0d5b373a19e9eb0022ad2936b46a1608f11b53f/lib/cli.js#L121

tdeschryver:
Fixes #1476.

**Changes:**
- Added `0` check.
- Changed the message if  `concurrency` was set to an empty string, because the error message seemed a bit weird.
- Introduced an extra check to validate if `concurrency` is an integer.
- Changed default `concurrency` from `0` to `1` if no `concurrency` is given.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -119,7 +119,15 @@ exports.run = () => {
 	}
 
 	if (cli.flags.concurrency === '') {
-		throw new Error(colors.error(figures.cross) + ' The --concurrency and -c flags must be provided the maximum number of test files to run at once.');
+		throw new Error(colors.error(figures.cross) + ' The --concurrency and -c flags must be provided.');
+	}
+
+	if (cli.flags.concurrency && isNaN(parseInt(cli.flags.concurrency, 10))) {
+		throw new Error(colors.error(figures.cross) + ' The --concurrency and -c flags must be an integer.');
+	}
+
+	if (parseInt(cli.flags.concurrency, 10) === 0) {
+		throw new Error(colors.error(figures.cross) + ' The --concurrency and -c flags must be greater than 0.');
 	}
 
 	if (hasFlag('--require') || hasFlag('-r')) {
@@ -142,7 +150,7 @@ exports.run = () => {
 		resolveTestsFrom: cli.input.length === 0 ? projectDir : process.cwd(),
 		projectDir,
 		timeout: conf.timeout,
-		concurrency: conf.concurrency ? parseInt(conf.concurrency, 10) : 0,
+		concurrency: cli.flags.concurrency ? parseInt(cli.flags.concurrency, 10) : 1,
 		updateSnapshots: conf.updateSnapshots,
 		color: conf.color
 	});
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -150,7 +150,7 @@ exports.run = () => {
 		resolveTestsFrom: cli.input.length === 0 ? projectDir : process.cwd(),
 		projectDir,
 		timeout: conf.timeout,
-		concurrency: cli.flags.concurrency ? parseInt(cli.flags.concurrency, 10) : 1,
+		concurrency: conf.concurrency ? parseInt(conf.concurrency, 10) : 1,
 		updateSnapshots: conf.updateSnapshots,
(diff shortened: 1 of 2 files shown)


lib/cli.js
```diff
@@ -119,7 +119,15 @@ exports.run = () => {
 	}
 
 	if (cli.flags.concurrency === '') {
-		throw new Error(colors.error(figures.cross) + ' The --concurrency and -c flags must be provided the maximum number of test files to run at once.');
+		throw new Error(colors.error(figures.cross) + ' The --concurrency and -c flags must be provided.');
+	}
+
+	if (cli.flags.concurrency && isNaN(parseInt(cli.flags.concurrency, 10))) {
```
novemberborn: I think this should be `Number.isInteger(Number.parseFloat('cli.flags.concurrency'))`. That way even values like `9.5` are rejected, which parse to `9` with `parseInt()`.

lib/cli.js
```diff
@@ -142,7 +150,7 @@ exports.run = () => {
 		resolveTestsFrom: cli.input.length === 0 ? projectDir : process.cwd(),
 		projectDir,
 		timeout: conf.timeout,
-		concurrency: conf.concurrency ? parseInt(conf.concurrency, 10) : 0,
+		concurrency: conf.concurrency ? parseInt(conf.concurrency, 10) : 1,
```
novemberborn: The default should remain `0`. This allows AVA to pick an optimal value for your machine, see https://github.com/avajs/ava/blob/465fcecc9ae0d3274d4d41d3baaca241d6a40130/api.js#L164:L167.

tdeschryver: Oh sorry, I assumed the default had to change, because you wouldn't be able to set it to `0`.

lib/cli.js
tdeschryver: Yep! Didn't thought about this case.

tdeschryver:
@novemberborn seems fair to me. I'll remove the `0` check.

tdeschryver:
@novemberborn would a check on a negative input be useful here? :thinking:

lib/cli.js
```diff
@@ -119,7 +119,11 @@ exports.run = () => {
 	}
 
 	if (cli.flags.concurrency === '') {
-		throw new Error(colors.error(figures.cross) + ' The --concurrency and -c flags must be provided the maximum number of test files to run at once.');
+		throw new Error(colors.error(figures.cross) + ' The --concurrency and -c flags must be provided.');
+	}
+
+	if (cli.flags.concurrency && !Number.isInteger(Number.parseFloat(cli.flags.concurrency, 10))) {
```
novemberborn: `parseFloat` doesn't take a radix.

lib/cli.js
```diff
@@ -119,7 +119,12 @@ exports.run = () => {
 	}
 
 	if (cli.flags.concurrency === '') {
-		throw new Error(colors.error(figures.cross) + ' The --concurrency and -c flags must be provided the maximum number of test files to run at once.');
+		throw new Error(colors.error(figures.cross) + ' The --concurrency and -c flags must be provided.');
+	}
+
+	if (cli.flags.concurrency &&
+		(!Number.isInteger(Number.parseFloat(cli.flags.concurrency)) || parseInt(cli.flags.concurrency, 10) < 0)) {
+		throw new Error(colors.error(figures.cross) + ' The --concurrency and -c flags must be a positive integer.');
```
novemberborn: This should be `nonnegative integer`, since `0` is allowed. I feel a bit bad about this nitpick ☹️

tdeschryver: No problem! I should be the one saying sorry...

novemberborn:
Thank you @tdeschryver!

PR #1479, opened 2017-08-01
Fixed makeApp() in endpoint testing recipe

ydaniv:
I've wasted blood, sweat, and tears over trying to get tests working, just to realize that I simply had to add `bodyParser` to express.

I hope this will help others.

Thanks for all the fish! (:


novemberborn:
Nice catch @ydaniv!

PR #1483, opened 2017-08-06
Update debugging with webstorm recipe

melisoner2006:
Update debugging with Webstorm recipe to include Node options flag --inspect-brk.

### DIFF
--- a/docs/recipes/debugging-with-webstorm.md
+++ b/docs/recipes/debugging-with-webstorm.md
@@ -13,6 +13,8 @@ In the `JavaScript file` field specify the path to AVA in the project's `node_mo
 
 In the `Application parameters` pass the CLI flags you're using and the test files you would like to debug, for example `--verbose test.js`.
 
+In the `Node parameters` pass the `--inspect-brk` flag to enable the Node inspector or `--inspect-brk=port` if you want to listen it on a specific port.
+
 Save the configuration.
 
 ## Setup using npm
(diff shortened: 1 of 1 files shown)


docs/recipes/debugging-with-webstorm.md
```diff
@@ -13,6 +13,8 @@ In the `JavaScript file` field specify the path to AVA in the project's `node_mo
 
 In the `Application parameters` pass the CLI flags you're using and the test files you would like to debug, for example `--verbose test.js`.
 
+In the `Node parameters` pass the `--inspect-brk` flag to enable the Node inspector or `--inspect-brk=port` if you want to listen it on a specific port.
```
novemberborn: Need to move "it" around:

> if you want it to listen on a specific port.

develar: @melisoner2006 Please add note that `--inspect-brk` works only in the node 7+.

Not all users uses latest 8 :)

develar: Probably will be better if text `if you want to listen` will be in the start of sentence, because it is advanced and optional instruction (english is not my native language, so, I don't known how much it is relevant).

melisoner2006: @novemberborn `--inspect` does work for me, but there's risk of using it. `--inspect-brk` stops before the user's script starts. `--inspect` flag doesn't guarantee that: if there are breakpoints early in the entry point they might be skipped until the debugger is attached. At least, this is my understanding.

melisoner2006: @develar Thanks for pointing out the node version. I updated my statement to include a solution for both 7+ and previous versions.

I also removed the statement about `--inspect-brk=port`. There are other flags that can also be used and I thought it is not the place to mention all of them. `--inspect-brk` is enough to make it work and if someone wants to try different things than they should look at Node.js documentation.

prigara: Looks good to me. That's for updating the recipe.

novemberborn:
Thank you @melisoner2006!

Also thank you @prigara and @develar for your feedback.

PR #1486, opened 2017-08-09
Fix typo in `t.notThrows` example

clayzermk1:
Trivial documentation fix proposal. I think there is a typo in the example for `t.notThrows`. The test name is "rejects" but shouldn't a test with a `notThrows` assertion be checking that the promise under test resolves instead of rejects? It's confusing when compared with the example for `throws` directly above that has a test named "rejects" as well.

Cheers and thanks for Ava!


sindresorhus:
You're right. Must have been a copy-paste mistake. Thanks :)

Issue #1459, opened 2017-07-18
Snapshot fixture location with transpiled tests


impaler:
When writing tests with something like the Typescript, tests can be transpiled to an output folder and run there. This is problematic when you need to keep the generated snapshots  `*.snap` and `*.md` fixtures in version control and not the build folder.

A lot of build processes will clean the build directory all together with a `rmraf` or similar along with the fixtures. Currently there does not seem to be an api to have a custom location for the [generated snapshot fixtures](https://github.com/avajs/ava/commit/ebd572a02debe8e17802f795e6cd618073c8d787). 

Got any suggestions would anyone consider a user defined location for the snapshot fixtures? @novemberborn

Cheers, ava is great :+1:

novemberborn:
Oh that's an interesting problem to have!

One option may be to resolve the original file location through source maps. That should probably deal with your use case?

Of course source maps aren't always reliable, so I do see the need for a configured location. But then it should be one directory that is used for all test files. There is a risk of multiple test files having the same filename, but existing in different directories. I'm not sure how to handle that problem. We could reproduce the directory structure (relative to the `package.json`), or concatenate the relative file path (but I'm wary of running into file name length restrictions). What do you think @impaler?

impaler:
@novemberborn yes and this feature is the only issue I have when using Ava and Typescript. Ideally using sourcemaps to have the snapshot fixtures next to the source *.ts files would be the best solution, developer experience wise.

I can't speak to the reliability issue with source maps you mention. Currently I am using the a basic tsc command to compile all *.ts files in my project along with the ava tests in ts, to one build folder. I can then the node debugger in Intellij Idea to configured to the ava "profile.js" to successfully use breakpoints etc on the original typescript source without issue, its fantastic. Its working with the sourcemaps generated from the typescript compiler just fine.

On the custom location, I agree one snapshot base directory would make sense for keeping things manageable. Regarding name collisions, maybe a subdirectory of this base path could be derived from the transpiled javascript test __filepath itself?

novemberborn:
I think we should:

* Check if the test file has a source map associated with it, and attempt to resolve the source file path
* Have that source file path take precedence over the test file path when it comes to loading and saving the snapshot
* Add a setting in `package.json#ava` to force snapshots to be stored in the specified global directory. This takes precedence over the source file detection
* When snapshots are stored in a global directory, mirror the file hierarchy of the test file (with the source file path still taking precedence) inside the global directory
[...]

Issue #1427, opened 2017-06-28
Flaky AppVeyor test for watcher not rerunning tests when they write snapshot files


novemberborn:
The `watcher does not rerun test files when they write snapshot files` [test](https://github.com/avajs/ava/blob/0e82f8f/test/cli.js#L310:L334) is flaky on AppVeyor. It relies on the watcher picking up the file system events. Presumably the [event debouncer](https://github.com/avajs/ava/blob/652705b735b818c9a206b35cc06f3760e6dcea13/lib/watcher.js#L28) is too fast for AppVeyor. This may also be a problem for users in general.

Perhaps we need to slow the debounce internal on AppVeyor, or even make it configurable? @avajs/core?

sindresorhus:
Just make it slower for AppVeyor. AppVeyor is notoriously slow.

sindresorhus:
@novemberborn Can we close this as we're not using AppVeyor anymore?

PR #1489, opened 2017-08-13
Snapshot location

impaler:
@novemberborn As discussed in #1459 the proposal to support custom snapshot locations and sourcemaps is:

- Check if the test file has a source map associated with it, and attempt to resolve the source file path
- Have that source file path take precedence over the test file path when it comes to loading and saving the snapshot
- Add a setting in package.json#ava to force snapshots to be stored in the specified global directory. This takes precedence over the source file detection
- When snapshots are stored in a global directory, mirror the file hierarchy of the test file (with the source file path still taking precedence) inside the global directory
- We should also add an integration test where the test file comes from a TypeScript compiler

Although there are some the test fixtures added for integration tests, I created an example project to more easily test the implementation.
https://github.com/impaler/ava-snapshot-location


lib/snapshot-manager.js
```diff
@@ -365,8 +386,31 @@ function determineSnapshotDir(projectDir, testDir) {
...
+function determineSourceMappedDir(options) {
+	const sourceMapFilePath = path.join(options.testDir, `${options.name}.map`);
+	const sourceMapContent = tryReadSourceMap(sourceMapFilePath);
+
+	if (sourceMapContent) {
+		const targetFilename = path.basename(sourceMapContent.file, 'js');
+		const testFileSource = sourceMapContent.sources.find(file => {
+			const fileName = file.replace(path.extname(targetFilename), '');
+			return fileName.indexOf(targetFilename) > -1;
+		});
```
novemberborn: This logic isn't quite right. According to the [source map "spec"](https://docs.google.com/document/d/1U1RGAehQwRypUTovF1KRlpiOFze0b-_2gc6fAH0KY0k/view) `file` is optional. There might be a `sourceRoot` which should be prepended to the `sources`.

I couldn't quite find a library which exports the correct source resolution. I think this will be sufficient for our use case (though I haven't tested it):

```js
// Assume there is but a single source. Multiple sources imply multiple possible
// snapshot locations, which won't work.
const firstSource = sourceMapContent.sources[0];
const rootedSource = sourceMapContent.sourceRoot ?
  sourceMapContent.sourceRoot + firstSource :
  firstSource;
// `convert-source-map` doesn't provide the path to the source map file, if the
// source map came from such a file. Assuming that if it did, it is stored right
// next to the test file. Or, alternatively, if it is stored elsewhere,
// resolving `rootedSource` relative to the test file results in a path that
// is the same as if it had been resolved relative to the source map file.
// Note that `options.relFile` is relative to `options.projectDir`.
const sourceFile = path.resolve(path.join(options.projectDir, options.relFile), rootedSource);
return path.dirname(sourceFile);
```

I'm not sure how much we need to guard against malformed source maps (e.g. no or empty `sources`), or source maps that unexpectedly resolve to an HTTP source.

readme.md
```diff
@@ -1041,6 +1041,20 @@ You can then check your code. If the change was intentional you can use the `--u
 $ ava --update-snapshots
 ```
 
+You can also set a custom location for the generated snapshot fixtures. This can be configured in the `package.json#ava` configuration for every test run.
```
novemberborn: I'd prefer:

> You can specify a fixed location for storing the snapshot files from each test under the `"ava"` key in your `package.json`:

readme.md
```diff
@@ -1041,6 +1041,20 @@ You can then check your code. If the change was intentional you can use the `--u
...
+
+```json
+{
+  "ava": {
+    "snapshotLocation": "custom-directory"
+  }
+}
+```
+
+Snapshots fixtures will be then saved in a directory structure that reflects the test sourcecode.
```
novemberborn: > The snapshot files will be saved in a directory structure that mirrors that of your test files.

readme.md
```diff
@@ -1041,6 +1041,20 @@ You can then check your code. If the change was intentional you can use the `--u
...
+
+If you are transpiling the tests with sourcemaps, ava will derive the test location from the `*.js.map` files, see more in the [TypeScript](docs/recipes/typescript.md) recipe.
```
novemberborn: > If you are running AVA against precompiled test files, AVA will try and use source maps to determine the location of the original files. Snapshots will be stored next to these files, following the same rules as if AVA had executed the original files directly. This is great if you're writing your tests in TypeScript (see our [TypeScript recipe](docs/recipes/typescript.md)).

test/fixture/snapshots/test-sourcemaps/build/feature/__tests__/test.js
```diff
@@ -0,0 +1,9 @@
+import test from '../../../../../../..';
+test('feature test title', t => {
+    t.snapshot({ foo: 'bar' });
+    t.snapshot({ answer: 40 });
+});
+test('another feature test', t => {
+    t.snapshot(new Map());
+});
+//# sourceMappingURL=test.js.map
```
novemberborn: Could you make sure these files have a newline at the end of them?

test/cli.js
```diff
@@ -680,6 +680,49 @@ test('legacy snapshot files are reported to the console', t => {
...
 
+test('snapshots infer their location from sourcemaps', t => {
+	t.plan(8);
+	const dirname = path.join('fixture/snapshots/test-sourcemaps');
+	const testLocationTypes = [
+		'src',
+		'src/test/snapshots',
+		'src/feature/__tests__/__snapshots__'
+	];
+	const removeExists = relFilePath => {
```
novemberborn: This doesn't remove anything though?

You should remove these files at least prior to running the test, otherwise on local setups you won't be able to tell if the code has broken if they're persisted from previous runs.

test/fixture/snapshots/test-sourcemaps/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "scripts": {
+	"build": "tsc"
```
novemberborn: Does this assume `tsc` is installed globally? I know we have a dev dependency on it but running `build` inside this directory won't pick it up. This should either refer to the top-level `node_modules/.bin/tsc` or use `npx tsc`.

(Also, going by the GitHub output here, the indentation seems off in this file and `tsconfig.json`.)

docs/recipes/typescript.md
```diff
@@ -83,3 +83,7 @@ test('an actual test', t => {
 ```
 $ npm test
 ```
+
+## Snapshots
+
+Ava derives the base location for the snapshot fixtures relative to the typescript as if it were standard javascript tests. It determines this from the associated `*.js.map` sourcemap files. Enabling sourcemaps with the typescript compiler can be done in either the [tsconfig](http://www.typescriptlang.org/docs/handbook/tsconfig-json.html) `tsconfig.json#compilerOptions#sourceMap=true` or in the `tsc` cli option `--sourceMap`.
```
novemberborn: Rather than this paragraph, update the sample `tsconfig.json` to enable source maps. Honestly it should always be enabled, since it helps AVA show correct stack traces and code snippets when assertions fail.

test/fixture/snapshots/test-sourcemaps/build/feature/__tests__/test.js
impaler: Is that just for constency? in this case everything under `...build/* ` is directly from the typescript compiler, I usually hesitate to edit compiled files for style only.

test/cli.js
impaler: Yes this is wrong, I have updated this be a little more dry https://github.com/avajs/ava/pull/1489/commits/12f1ff8d73e0ae47c109eed9737470c9d421fa8e#diff-583d616f4b01bf1fa153e71c0cef7de7R704. If there are more integration tests on projects like this I can think of some better ways to do it, like maybe the project should be first duplicated to a /tmp location and so on.

test/fixture/snapshots/test-sourcemaps/package.json
impaler: Yes it does, I don't think there will be any need to run build here for the tests anyway. So I fixed the indentation and removed "scripts".

lib/cli.js
```diff
@@ -144,6 +144,7 @@ exports.run = () => {
 		timeout: conf.timeout,
 		concurrency: conf.concurrency ? parseInt(conf.concurrency, 10) : 0,
 		updateSnapshots: conf.updateSnapshots,
+		snapshotLocation: conf.snapshotLocation,
```
impaler: Sure no preference here from me projectDir, I think I saw Dir used somewhere else :)

lib/snapshot-manager.js
```diff
@@ -355,7 +367,16 @@ class Manager {
 	}
 }
 
-function determineSnapshotDir(projectDir, testDir) {
+function determineSnapshotDir(options) {
+	const projectDir = options.projectDir;
+	const testDir = determineSourceMappedDir(options);
+	if (options.location) {
+		// "snapshotLocation" in ava package.json config, snapshots derive a location
+		// relative from the source testDir not the preset "__snapshots__" folder names
```
impaler: Good point, I guess the comment highlighted some of the ambiguitiy, please checkout the latest revision.

lib/snapshot-manager.js
```diff
@@ -355,7 +367,16 @@ class Manager {
...
+		const resolvedLocation = path.resolve(options.projectDir, options.location);
+		const relativeTestLocation = options.testDir.replace(options.projectDir, '');
```
impaler: Yes, please checkout the latest revision.

lib/snapshot-manager.js
```diff
@@ -365,8 +386,31 @@ function determineSnapshotDir(projectDir, testDir) {
 	return testDir;
 }
 
+function determineSourceMappedDir(options) {
+	const sourceMapFilePath = path.join(options.testDir, `${options.name}.map`);
+	const sourceMapContent = tryReadSourceMap(sourceMapFilePath);
```
impaler: Indeed thanks a bunch I am new to that library very cool. Seems to work well with the inline version in my testing https://github.com/impaler/ava-snapshot-location/tree/master/inline-sourcemaps

lib/snapshot-manager.js
impaler: Thanks I learned a few things here. I guess we wont be able to answer how much more to guard without trying more than just typescript. With the way things are layed out it shouldn't be too hard to expand on this.

readme.md
impaler: :+1:

readme.md
impaler: :+1:

readme.md
impaler: :+1:

impaler:
Thanks for the great review @novemberborn :)

I won't claim that we have complete integration test coverage on this snapshot feature, but it has the main aspects. Otherwise as I said in a comment I think we would need some more utilities to duplicate project folders and run all scenarios like I am testing in https://github.com/impaler/ava-snapshot-location.

Based on what I can see on appveyor the ci issues seem unrelated.

lib/snapshot-manager.js
```diff
@@ -368,16 +357,12 @@ class Manager {
...
-	const projectDir = options.projectDir;
 	const testDir = determineSourceMappedDir(options);
-	if (options.location) {
-		// "snapshotLocation" in ava package.json config, snapshots derive a location
-		// relative from the source testDir not the preset "__snapshots__" folder names
-		const resolvedLocation = path.resolve(options.projectDir, options.location);
-		const relativeTestLocation = options.testDir.replace(options.projectDir, '');
-		return path.join(resolvedLocation, relativeTestLocation);
+	if (options.fixedLocation) {
+		const relativeTestLocation = testDir.replace(options.projectDir, '');
```
novemberborn: Use `path.relative(options.projectDir, testDir)`.

lib/snapshot-manager.js
```diff
@@ -388,23 +373,20 @@ function determineSnapshotDir(options) {
...
-		if (testFileSource) {
-			if (options.location) {
-				const basePath = path.join(options.projectDir, options.location);
-				const sourceTestFilePath = path.dirname(path.resolve(options.testDir, testFileSource));
-				const relativePath = sourceTestFilePath.replace(options.projectDir, '');
-				return path.join(basePath, relativePath);
-			}
-			return path.dirname(path.resolve(options.testDir, testFileSource));
-		}
+	const sourceMapContent = tryRead(options.relFile).toString();
```
novemberborn: `tryRead()` returns `null` if the file doesn't exist. You'll have to read it into a variable and bail out of if it's `null`.

Perhaps this function should return `null` if no source-mapped directory could be determined. Then `determineSnapshotDir()` can do `testDir = determineSourceMappedDir(options) || options.testDir`.

test/cli.js
```diff
@@ -701,23 +701,34 @@ test('snapshots infer their location from sourcemaps', t => {
...
+		.map(snapshotRelativeDir => {
+			const snapshotPath = path.join(__dirname, relativeFixtureDir, snapshotDir, snapshotRelativeDir);
+			return [
+				path.join(snapshotPath, 'test.js.md'),
+				path.join(snapshotPath, 'test.js.snap')
+			];
+		})
+		.reduce((a, b) => a.concat(b), []);
+	const removeExistingSnapshotFixtureFiles = filePath => {
+		return fs.existsSync(filePath) && fs.unlinkSync(filePath);
```
novemberborn: There is still room for the file to be deleted right in between `existsSync` and `unlinkSync`. Would be better to do https://github.com/impaler/ava/blob/12f1ff8d73e0ae47c109eed9737470c9d421fa8e/test/cli.js#L568:L574, even if it's more verbose.

test/cli.js
novemberborn: I'm not too concerned about how DRY this is. The other test looks good, just need to take that same approach here.

test/fixture/snapshots/test-snapshot-location/package.json
```diff
@@ -0,0 +1,8 @@
+{
+	"ava": {
+		"files": [
+			"src/**/*test.js"
+		],
+	"snapshotDir": "snapshot-fixtures"
+	}
```
novemberborn: Nitpick alert: the indentation in this file is inconsistent. It's just a fixture, but once seen I can't not comment 😉

test/fixture/snapshots/test-sourcemaps/build/feature/__tests__/test.js
novemberborn: Fair enough.

test/fixture/snapshots/test-snapshot-location/package.json
impaler: :thumbsup:

test/cli.js
impaler: Sure that will also throw any fs issues, I updated the pull request :)

impaler:
The windoze build still seems intermittent and unrelated, since the same error is happening in other branches it seems out of scope.

novemberborn:
@impaler this is great! I've pushed some minor fixes, and some larger fixes to `determineSourceMappedDir()`. See the commits.

Could you give those a spin with your project? Will merge this if it's all good.

impaler:
Thanks @novemberborn, yes everything is working as I expected with your changes :)

And looks like appveyor is now working, know if it was it related to changes here?

novemberborn:
> And looks like appveyor is now working, know if it was it related to changes here?

There are some known issues, e.g. https://github.com/avajs/ava/issues/1427. But I've been having loads of problems in https://github.com/avajs/ava/pull/1499 too. So 🤷‍♂️.

novemberborn:
Thanks @impaler!

PR #1491, opened 2017-08-20
Document config options

### LINKED ISSUE #1473: `files` vs. `source` vs. `match`?
It’s not clear to me what the difference is between the following three `package.json` properties. Would you mind explaining in the readme (AFAICT, it isn’t documented elsewhere, either)?

```js
		"files": [
			"my-test-folder/*.js",
			"!**/not-this-file.js"
		],
		"source": [
			"**/*.{js,jsx}",
			"!dist/**/*"
		],
		"match": [
			"*oo",
			"!foo"
		],
```

Questions:

* Are `files` the test files? Shouldn’t `test` be in the name of this property?
* Is `source` all files (tests *and* implementations)? Is this option needed for the `--watch` mode?
    * Would it make sense to just specify the implementations here and to watch (the union of) the test files and the implementation files?
* `match` looks like an additional filter mechanism. Why is such a mechanism needed?

kristianmandrup:
Fixes #1473 with additional docs from comments/guidance mentioned in that issue.
As I'm not an expert on the inner workings of `ava`, please correct any mistakes or make further clarifications thanks.


novemberborn:
Thanks! I've pushed a few edits, please let me know what you think.

PR #1499, opened 2017-08-28
Assume npm@5.4, work around permissions issue

novemberborn:
There's a fair bit of churn between `package-lock.json` files generated with different npm versions (even minor releases). Contributors may be using an older npm version or even Yarn. Thus the `package-lock.json` may regress or not even be updated.

~This PR proposes we pin npm to an exact version. Then in Travis we can check whether the `package-lock.json` is modified and fail the build if this occurs. This should catch regressions or missing dependencies.~

Note that CI is currently failing due to a bug in npm@5.4.0: <https://github.com/npm/npm/issues/18324>. Consequently we shouldn't land it until we can pin to a patched version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ cache:
   directories:
     - $HOME/.npm
 before_install:
-  - npm install --global npm@^5.2.0
+  - npm install --global npm@5.4.0
   - npm --version
 install:
   - if [[ ${FRESH_DEPS} == "true" ]]; then npm install --no-shrinkwrap --prefer-online; else npm install --prefer-offline; fi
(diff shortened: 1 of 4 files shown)


novemberborn:
> Note that CI is currently failing due to a bug in npm@5.4.0: npm/npm#18324. Consequently we shouldn't land it until we can pin to a patched version.

Unfortunately 5.4.0 was promoted to `latest` despite this bug. I've pushed a commit which tries to work around this by marking the binary as executable when tests are run. We can remove this workaround once npm installs the binary correctly again.

novemberborn:
Turns out that due to https://github.com/npm/npm/issues/18135 and our `chokidar` dependency we'll always have churn in Travis. I'll be landing this without the churn detection, so that we can commit the churn caused by npm@5.4 (as compared to 5.3).

novemberborn:
We're now running into https://github.com/npm/npm/issues/17671. I'm disabling tests with locked dependencies on AppVeyor for the time being.

PR #1500, opened 2017-08-29
Update snapshot signature for flow types

philippotto:
I updated the flow types for the snapshot function to allow an options object as a parameter.


sindresorhus:
Great. Can you add a test to https://github.com/avajs/ava/tree/master/test/flow-types ? Should just showing usage of the different ways of using `t.snapshot` so we can make sure it works and doesn't regress.

philippotto:
Sure thing, done.

novemberborn:
Lovely work, thanks @philippotto!

PR #1502, opened 2017-09-01
Match anonymous tests when all of the match expressions are negative

rhendric:
<strike>(Matching on `'*'`, for example, still excludes tests with no titles, as it did prior to this commit.)</strike>

The motivation here is if I have a test suite with some anonymous tests and some titled tests, and some of the titled tests are tagged with something like `[long-running]`, and I want to run `ava -m '!*[long-running]*'`, I don't want to exclude the anonymous tests. But with current Ava they are excluded.

<strike>For what it's worth, anyone who misses the old behavior could get it back with `-m '*' -m <negative match expression(s)>`.</strike>

**Edit:** Changed so that `-m '*'` matches all tests, including anonymous. For purposes of matching, anonymous tests have empty string titles.

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -112,7 +112,7 @@ class Runner extends EventEmitter {
 		}
 
 		if (metadata.type === 'test' && this.match.length > 0) {
-			metadata.exclusive = title !== null && matcher([title], this.match).length === 1;
+			metadata.exclusive = matcher([title || ''], this.match).length === 1;
 		}
 
 		const validationError = validateTest(title, fn, metadata);
--- a/test/runner.js
+++ b/test/runner.js
@@ -711,6 +711,32 @@ test('options.match overrides .only', t => {
 	});
 });
 
+test('options.match includes anonymous tests in negative matches', t => {
+	t.plan(4);
+
+	const runner = new Runner({
+		match: ['!*oo']
+	});
+
+	runner.chain.test('foo', a => {
+		t.fail();
+		a.pass();
+	});
+
+	runner.chain.test(a => {
+		t.pass();
+		a.pass();
+	});
+
+	runner.run({}).then(() => {
+		const stats = runner.buildStats();
+		t.is(stats.skipCount, 0);
+		t.is(stats.passCount, 1);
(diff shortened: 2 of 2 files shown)


lib/runner.js
```diff
@@ -112,7 +113,7 @@ class Runner extends EventEmitter {
 		}
 
 		if (metadata.type === 'test' && this.match.length > 0) {
-			metadata.exclusive = title !== null && matcher([title], this.match).length === 1;
+			metadata.exclusive = title === null ? this.matchTitleless : matcher([title], this.match).length === 1;
```
novemberborn: I think it may be sufficient to remove the `title !== null` check. `matcher([null],["!*oo"])` gives `[null]` which passes the `length === 1` check. Perhaps use `matcher([title || ''], this.match)`, since the [matcher](https://www.npmjs.com/package/matcher) expects strings.

rhendric: I'm okay with that simplification, but you should know that it does change behavior—that would essentially treat a titleless test the same as a test with an empty string title, so `-m '*'` would truly include all tests, even the titleless ones, and there would no longer be an escape hatch to go back to the old behavior. I was trying to lean towards making this PR as conservative as possible, but I can't think of a real reason to want to distinguish between anonymous tests and empty title tests, so if you're fine with that then I'll make the change.

novemberborn:
> I can't think of a real reason to want to distinguish between anonymous tests and empty title tests, so if you're fine with that then I'll make the change.

Yea me neither. I've scanned the discussion in https://github.com/avajs/ava/issues/476 and I don't think we ever considered this. Generally we'd recommend against using anonymous tests, so if you do use them and need to somehow exclude them I guess you might just have to name them 😉

rhendric:
So shall it be!

ORESoftware:
why not default to the file path for the name of the test if no test title is given

novemberborn:
@ORESoftware we already display the title as "anonymous". This PR postulates that `!*foo` should match anonymous tests, since, by virtue of having no title, their title does not end in `foo`.

rhendric:
@novemberborn, ready to merge?

PR #1508, opened 2017-09-05
Include anonymous functions in stacktraces

jugglinmike:
This patch corrects Ava's reporting of stack traces which contain frames of anonymous functions. See pull request gh-1313 for the original problem description and steps to reproduce.

@neoeno originally submitted that patch in March of this year. The thread has been inactive for the 5 months since the initial review phase. This pull request is an attempt to implement that feedback and fix the bug.

Most notably, this involved implementing @novemberborn's recommendation:

> My hunch is that we should extract the stack when serializing (filter out
> unnecessary bits early), and beautify it in the mini and verbose reporters.
> Though I'm not even sure what `beautify-stack` does that `extract-stack`
> doesn't.

To fully accommodate this change, I've updated the tests for Ava's reporters to use the internal `serialize-error` module. This makes them somewhat more like integration tests, but this was already true to a certain extent--previously, some tests relied on the internal `beautify-stack` module.

We could update the tests to explicitly construct the errors as we expect them to result from the serialization process. Although that would tend to isolate the tests, it might also make them harder to maintain (as any change to the serialization format would have to be reflected in the tests themselves). For this initial draft, I opted for this factoring because the output of `serialize-error` seems like a pretty strong contract to depend on.

As part of the discussion of gh-1313, @sindresorhus created [`extract-stack`](https://www.npmjs.com/package/extract-stack), a new npm module. For reasons shared in the discussion that followed, that module is not suitable in its current form. Because of the similar-sounding functionality between modules (`beatify-stack`, `extract-stack`, `clean-stack`, and `stack-utils`), I have instead opted to rename `extractStack` to the slightly more precise `extractFrames`. Due to the re-factoring described above, I was also able to move the function definition into the `beautify-stack` module itself, which I believe further limits the ambiguity between module responsibilities. To shore things up a little more, I've included some documentation for `beautify-stack` to describe its intended behavior along with an example.


lib/serialize-error.js
```diff
@@ -90,5 +89,12 @@ module.exports = error => {
 		}
 	}
 
+	if (typeof error.stack === 'string') {
+		retval.summary = error.stack.split('\n')[0];
+	} else {
+		retval.summary = typeof error === 'function' ?
```
novemberborn: What's the origin of this `typeof error === 'function'` check?

test/helper/error-from-worker.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+const serializeError = require('../../lib/serialize-error');
+
+module.exports = function (err, {type, file, stack} = {}) {
```
novemberborn: We're targeting Node.js 4 still, so this destructuring won't work unfortunately.

Perhaps just use `Object.assign()` without checking for values?

lib/serialize-error.js
jugglinmike: Yeah, this seemed a little hinky to me, too. I implemented that to satisfy [an existing test](https://github.com/avajs/ava/blob/1df502df8cba18f92afcbaed730d8c724687ca45/test/cli.js#L197-L205) which looks pretty intentional. It was introduced here: https://github.com/avajs/ava/pull/206 . Maybe @jamestalmage can give us some perspective.

test/helper/error-from-worker.js
jugglinmike: Sure thing

jugglinmike:
> Could you have a look at why tests are failing though?

Sure thing. This was a tricky issue, but I think I have it fixed over at https://github.com/avajs/ava/pull/1524

lib/serialize-error.js
novemberborn: The code comment seems informative: https://github.com/avajs/ava/blob/15bd79402c55468a4b13aedceba503ee26e964a1/lib/serialize-value.js#L11:L12

This is no longer applicable I think. We catch and wrap all errors from tests, so it's only uncaught exceptions and unhandled rejections that are passed directly to `serialize-error`. And right now we don't even guard against `throw null`. Let's remove this edge-case and we can open a follow-up for handling non-Error exceptions / rejections.

jugglinmike: The test demonstrates a case where a function value is thrown and not caught. Removing the test and this condition would prevent Ava from displaying the source of the function in such cases.

Granted, it seems extremely rare for a function value to be thrown in the first place. I just want to make sure we're on the same page before proceeding. Can you verify?

novemberborn: Yes let's get rid of this check.

The code needs to be more defensive here but we'll handle that in a follow-up issue.

jugglinmike:
Okay, I've pushed up another commit to remove that check. Since it reflects an intentional but subjective change, I'd prefer if we could include it in a standalone commit. That will help future contributors understand that the change in behavior was not a regression.

There have been some recent commits to `master` that cause merge conflicts in this changeset. I've taken the liberty to rebase and force-push the result. In the interest of transparency, I've made the original version of this branch available here:

https://github.com/avajs/ava/compare/master...bocoup:include-anonymous-functions-in-stacktraces-orig-2

novemberborn:
Hi @jugglinmike, I haven't forgotten about you 😄  Glad to see the tests are passing, I'm hoping to merge this later this week.

jugglinmike:
No worries; I know how things get with maintaining big FOSS projects like this. Take your time, and thanks for keeping me in the loop!

novemberborn:
Thank you @jugglinmike! There should be a new release out within the next few days.

jugglinmike:
Great! And thank you!

PR #1496, opened 2017-08-26
Remove Bluebird.longStackTraces()

anshulwadhawan:
Fixes #1492.

PR #1514, opened 2017-09-11
#1496 integration tests

### LINKED ISSUE #1492: Remove Bluebird.longStackTraces()
We enable long stack traces: http://bluebirdjs.com/docs/api/promise.longstacktraces.html

However these don't matter for users. By default tests will run with the native `Promise` implementation, even with async-await on Node.js 4 and 6. AVA's code, which at times does use Bluebird, is always removed from the stack traces, so it doesn't matter what Bluebird captures.

As long as we process the `require` modules early enough, users can still enable long stack traces if they use Bluebird in their own code. It would be good to have an integration test for this though.

See https://github.com/avajs/ava/blob/dd9e8b2effe541f9f232ee622452343dac5895dd/lib/test-worker.js#L33

mliou8:
Integration test for making sure that long stack trace works. 

Fixes #1492.


mliou8:
@novemberborn Hey Mark - this is what I've got so far for the integration tests. I'm not sure why there's so many commits? Hopefully this is what you meant in the issue discussion.

package-lock.json
```diff
@@ -2,7 +2,6 @@
   "name": "ava",
   "version": "0.22.0",
   "lockfileVersion": 1,
-  "requires": true,
```
novemberborn: Could you remove these changes from your PR?

test/fixture/long-stack-trace/package.json
```diff
@@ -0,0 +1,7 @@
+{
+	"ava": {
+		"name": "application-name",
+		"version": "0.0.1",
```
novemberborn: You probably meant for these two fields to be at the top-level?

Node.js doesn't care about them though. Feel free to remove them.

test/fixture/long-stack-trace/test.js
```diff
@@ -0,0 +1,14 @@
+import test from '../../../';
+const Promise = require('./enable-trace.js');
```
novemberborn: `import Promise from './enable-trace';`

test/fixture/long-stack-trace/test.js
```diff
@@ -0,0 +1,14 @@
...
+
+test(async t => {
+	const resolve = await Promise.resolve().then(function outer() {
```
novemberborn: Would be good to add a comment linking to http://bluebirdjs.com/docs/api/promise.longstacktraces.html, since that's where the code comes from.

test/fixture/long-stack-trace/test.js
```diff
@@ -0,0 +1,14 @@
...
+    return Promise.resolve().then(function inner() {
```
novemberborn: Indentation seems a bit off here, though looks like we're no linting this file so hey 🤷‍♂️

mliou8:
@novemberborn  Hey mark thanks for the comments! I've pushed up some changes, please let me know if there's anything else I should change.

novemberborn:
Thanks @mliou8!

PR #1515, opened 2017-09-11
When removing .snap files, only recalculate impacted test

### LINKED ISSUE #1511: Deleting a snap file runs all tests
## Description

I think this is a nice workflow:
* write complex data processing code
* feed it example data in tests with `t.snapshot()`
* inspect result `.md` file in editor tab
* if snapshot is no longer good, remove .snap file so it gets regenerated

👉  this runs all the tests again, instead of just the test using the snap file 😢 

ava v0.22

wmertens:
Fixes #1511

This also introduces a slight change in semantics: `package.json` and `**/*.snap` are always considered source files, since they are part of how AVA works and not of the application source files. If that is undesirable, they can be put in the ignore list.

### DIFF
--- a/docs/recipes/watch-mode.md
+++ b/docs/recipes/watch-mode.md
@@ -65,7 +65,7 @@ AVA uses [`chokidar`] as the file watcher. Note that even if you see warnings ab
 
 In AVA there's a distinction between *source files* and *test files*. As you can imagine the *test files* contain your tests. *Source files* are all other files that are needed for the tests to run, be it your source code or test fixtures.
 
-By default AVA watches for changes to the test files, snapshot files, `package.json`, and any other `.js` files. It'll ignore files in [certain directories](https://github.com/novemberborn/ignore-by-default/blob/master/index.js) as provided by the [`ignore-by-default`] package.
+By default AVA watches for changes to the test files and any `.js` files. It'll ignore files in [certain directories](https://github.com/novemberborn/ignore-by-default/blob/master/index.js) as provided by the [`ignore-by-default`] package. It also always watches `package.json` an snapshot files (even when you specify other source patterns), unless they are specifically ignored.
 
 You can configure patterns for the source files in the [`ava` section of your `package.json`] file, using the `source` key.
 
--- a/lib/ava-files.js
+++ b/lib/ava-files.js
@@ -145,7 +145,7 @@ class AvaFiles {
 		});
 	}
 	isSource(filePath) {
-		let mixedPatterns = [];
+		const mixedPatterns = [];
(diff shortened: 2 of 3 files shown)


lib/ava-files.js
```diff
@@ -262,13 +266,22 @@ class AvaFiles {
...
+		}
+		if (paths.indexOf('package.json') === -1) {
+			paths.unshift('package.json');
+		}
+		if (paths.indexOf('**/*.snap') === -1) {
+			paths.unshift('**/*.snap');
 		}
 
-		paths = paths.concat(this.files);
+		paths.push(...this.files);
```
novemberborn: We're targeting Node.js 4, which means spread isn't available 😞

wmertens: ok fixed

wmertens:
Addressed your comments, but I'm not sure how to test the ignore pattern…

docs/recipes/watch-mode.md
```diff
@@ -65,7 +65,7 @@ AVA uses [`chokidar`] as the file watcher. Note that even if you see warnings ab
 
 In AVA there's a distinction between *source files* and *test files*. As you can imagine the *test files* contain your tests. *Source files* are all other files that are needed for the tests to run, be it your source code or test fixtures.
 
-By default AVA watches for changes to the test files, snapshot files, `package.json`, and any other `.js` files. It'll ignore files in [certain directories](https://github.com/novemberborn/ignore-by-default/blob/master/index.js) as provided by the [`ignore-by-default`] package.
+By default AVA watches for changes to the test files and any `.js` files. It'll ignore files in [certain directories](https://github.com/novemberborn/ignore-by-default/blob/master/index.js) as provided by the [`ignore-by-default`] package. It also always watches `package.json` an snapshot files (even when you specify other source patterns), unless they are specifically ignored.
```
novemberborn: > unless they are specifically ignored

Swap `they` for `these`, I think.

This should have an example of how to ignore them.

novemberborn:
Hey @wmertens, do you have the time to finish this up, or would you rather I give it a go?

wmertens:
Hmm, no time at all right now, maybe next week; so feel free to assist, thanks! :)

novemberborn:
No worries @wmertens. I'll probably focus on other issues first.

novemberborn:
Hey @wmertens I'm going to close this, if only for housekeeping reasons. Feel free to reopen / make a new PR if you have the time to get back to this.

I have some thoughts on how AVA should select test, helper and source files that will impact watch mode. I've added the appropriate label to #1511 so maybe I'll be able to solve it all in one go when I get to that stage.

PR #1130, opened 2016-11-27
Recommend Yarn for installation

anoff:
Fixes #1129 
Also bumped the expected version in the `package.json`

PR #1518, opened 2017-09-15
Prioritize npm over Yarn install method in the readme

ntwb:
The statement prior to this PR: _"Yarn  currently provides significant speed improvements over npm during the installation process. Consider Yarn if the installation is too slow for your needs."_

I would suggest this is no longer true with npm 5, this PR removes that statement

Secondly, I switched to npm to be the "lead" method, if you will "preffered" method to install Ava. I made this change due to my assumption that because this package now contains a `package-lock.json` file and not a `yarn.lock` file that npm 5 is now the preferred method 😏 

Edit: For reference, Yarn was added in #1130

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -69,18 +69,18 @@ test(t => {
 
 ### Add AVA to your project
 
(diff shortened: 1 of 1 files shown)


ntwb:
_hehehe, I try to use US spelling when not in/on an Australian site, sometimes I forget 😂_

novemberborn:
> I made this change due to my assumption that because this package now contains a package-lock.json file and not a yarn.lock file that npm 5 is now the preferred method 😏

Oops, my biases are showing 😉 

Thanks @ntwb!

PR #1523, opened 2017-09-16
Set NODE_ENV to to 'test' if not already set

### LINKED ISSUE #1470: Feature request: set NODE_ENV=test
## Description

To set `process.env.NODE_ENV='test'` if it is not already set.

P-Seebauer:
Fixes #1470

Still needs tests and docs.

I haven't found a good place in the docs to put the "warnings about the dangers".

Also stops `process.env.AVA_PATH` from getting overwritten in `fork.js`. I did not find a reason why the `Object.assign` is inside the if statement. (I have had a look at #559 and the likes and it seems to me that there is a similar issue with the AVA_PATH waiting to happen). Since there are no tests, I believe that the overwrite was not intentional (and could move to another place, if it was).


novemberborn:
Thanks @P-Seebauer!

> I haven't found a good place in the docs to put the "warnings about the dangers".

Perhaps https://github.com/avajs/ava/#process-isolation comes closest, as it talks about the test process. Add another paragraph explaining what value of `NODE_ENV` to expect, and warning users that this may influence their code?

> Also stops process.env.AVA_PATH from getting overwritten in fork.js. I did not find a reason why the Object.assign is inside the if statement. (I have had a look at #559 and the likes and it seems to me that there is a similar issue with the AVA_PATH waiting to happen). Since there are no tests, I believe that the overwrite was not intentional (and could move to another place, if it was).

I think you're right. I must've assumed it was a copy already. The value isn't used in the main process though so it was harmless.

P-Seebauer:
I think, I'm done.

Sorry about the commit/build spam, I wasn't aware that github links the branch this way.

I'm not too happy about the Documentation part. Do you have any suggestions what could be done better?

readme.md
```diff
@@ -1127,6 +1127,8 @@ t.true(a.test(b) || b === c)
 
 Each test file is run in a separate Node.js process. This allows you to change the global state or overriding a built-in in one test file, without affecting another. It's also great for performance on modern multi-core processors, allowing multiple test files to execute in parallel.
 
+For each process, `NODE_ENV` gets set to `'test'` if it was not set. This could change some ouput and/or the behaviour of some modules (e.g. express' final error handler does [*not* log errors when `NODE_ENV` is `'test'`](https://github.com/expressjs/express/blob/c087a45b9cc3eb69c777e260ee880758b6e03a40/lib/application.js#L630)). Is very convenient to set some defaults (change database, disable logging, etc…), but you shouldn't rely on it too much (if you check your [coverage](#code-coverage), you'll see it go down). As `NODE_ENV` is always set, `'NODE_ENV' in process.env` always is `false` when called from your tests.
```
novemberborn: How about this?

> AVA will set `process.env.NODE_ENV` to `test`, unless the `NODE_ENV` environment variable has been set. This is useful if the code you're testing has test defaults (for example when picking what database to connect to, or environment-specific Babel options). It may cause your code or its dependencies to behave differently though. Note that `'NODE_ENV' in process.env` will always be `true`.

The Express example seems overly specific, as do the references to code coverage. And the result of `'NODE_ENV' in process.env` will be `true`, not `false`.

P-Seebauer: Yeah looks much better.

novemberborn:
Thanks @P-Seebauer!

PR #1524, opened 2017-09-16
Stabilize colors

jugglinmike:
Stabilize colors

> Prior to this commit, the library and its tests relied on the same code
> to add color to terminal output. This prevented the tests from
> recognizing changes to that code.
> 
> Such a change was introduced in a recent release of the `chalk` module
> [1]: it was updated to disable text coloring when run without a TTY.
> This is the case in the test environment environment provided by the
> `tap` module (which Ava uses for its tests).
> 
> This fix disabled color output in the tests, but it caused no failures
> because both the expected values and the actual values were effected.
> However, this situation amounted to a decrease in test coverage because
> the expected color of the terminal output could no longer be verified.
> 
> Update the tests environment to simulate a TTY so that Ava continues to
> produce colored text despite being run in a sub-process provided by
> `tap`. Update the test logic to reduce dependency on Ava internals in
> order to avoid similar silent regressions in the future.
>
> [1] https://github.com/chalk/chalk/commit/0827d3b6327dfd5e132589b1308703c9b4e127ef

This issue was a surprise for me after first submitting gh-1508 because I had been using npm version 3. That version does not recognize the `package_lock.json` file, so I had been vetting my changes with newer versions of the module dependencies than are available in the continuous integration environment (this is why the tests pass for [the build that sets `FRESH_DEPS` to `true`](https://github.com/avajs/ava/blob/1df502df8cba18f92afcbaed730d8c724687ca45/.travis.yml#L22)).

### DIFF
--- a/test/helper/colors.js
+++ b/test/helper/colors.js
@@ -14,14 +14,25 @@ function make(name) {
 }
 const bold = make('bold');
 const white = make('white');
-const dim = make('dim');
 const gray = make('gray');
 
+// The following color definitions are contextual so that they produce expected
+// values which mimic the behavior of the Chalk library.
+const isSimpleWindowsTerm = process.platform === 'win32' && !(process.env.TERM || '').toLowerCase().startsWith('xterm');
+const openDim = isSimpleWindowsTerm ? '' : ansiStyles.dim.open;
+const openBlue = isSimpleWindowsTerm ? '\u001B[94m' : ansiStyles.blue.open;
+// "Use `bold` by default on Windows"
+// https://github.com/chalk/chalk/issues/36
+const blue = string => openBlue + string + ansiStyles.blue.close;
+// "(Windows) chalk.gray.dim not visible"
+// https://github.com/chalk/chalk/issues/58
+const dimGray = string => gray(openDim + string + ansiStyles.dim.close);
+
 module.exports = {
-	blue: make('blue'),
+	blue,
 	boldWhite: string => bold(white(string)),
-	dimGray: string => gray(dim(string)),
-	gray: make('gray'),
+	dimGray,
+	gray,
 	green: make('green'),
 	magenta: make('magenta'),
 	red: make('red'),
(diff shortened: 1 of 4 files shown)


novemberborn:
> This issue was a surprise for me after first submitting gh-1508 because I had been using npm version 3. That version does not recognize the package_lock.json file, so I had been vetting my changes with newer versions of the module dependencies than are available in the continuous integration environment (this is why the tests pass for the build that sets FRESH_DEPS to true).

This is odd. The `package-lock.json` can help ensure everybody has the same dependencies when developing, and stop us from accidentally using a dependency's API that came with a newer version of that dependency than is in our SemVer range. Sounds like there is an unexpected breaking change in perhaps `chalk`? If we update our version of `chalk` would that help? How does https://github.com/avajs/ava/pull/1401 impact this problem?

I'm perpetually confused by color management in AVA, and this PR seems appropriate despite my questions in the previous paragraph. Still, I'd like to understand it all a bit better before hitting the merge button.

(And, as it stands, there are some linting errors.)

jugglinmike:
> Sounds like there is an unexpected breaking change in perhaps chalk? If we
> update our version of chalk would that help? How does #1401 impact this
> problem?

I think I'd call it a bug fix because as I understand it, it disables colors
when a TTY is not present, which generally seems expected. I don't believe that
the referenced issue about the `--color` flag is directly related because the
test process is being created by `tap`, not Ava itself.

With the linting problem out of the way, I can see some Windows-specific test
failures. It looks like there are some complexities in Chalk that need to be
re-created in the new test helper. I submitted another "fixup" commit to
implement that. In some ways, it's unfortunate for the test logic to grow in
complexity like this. In other ways, this is functionality that consumers
depend on, so it's probably for the best that it is explicitly guaranteed by
this project's tests.

novemberborn:
Makes sense, thanks @jugglinmike!

jugglinmike:
My pleasure!

PR #1528, opened 2017-09-20
Remove React section from browser testing recipe

### LINKED ISSUE #1522: Remove React from Browser Testing Recipe
Can we remove the React example from the browser testing recipe?

https://github.com/avajs/ava/blob/master/docs/recipes/browser-testing.md

We now have a dedicated recipe for React. I didn't write the React section of that recipe, it was left in from the previous revision before I added `browser-env`/`window`. I always thought it seemed a bit convoluted and it appears it's actually invalid (references an undefined `div` variable).

Can this be removed in favour of the React recipe?

Also, if you wanna keep the React version in the browser recipe, I have since written [ReactJSDOM](https://github.com/lukechilds/react-jsdom) which may be a neater solution.

lukechilds:
Fixes #1522


lukechilds:
Sorry, commits are a bit messy because I used this branch for a PR previously. Squash + merge will do the job.

kevva:
Thanks 💯

Issue #1530, opened 2017-09-23
Watch mode filter by filename and by test name


unional:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%2480%20Funded-%2300A156.svg" />](https://oss.issuehunt.io/r/avajs/ava/issues/1530)

ava@0.22

borrow the idea from `jest`:

```sh
Watch Usage
 › Press p to filter by a filename regex pattern.
 › Press t to filter by a test name regex pattern.
 › Press q to quit watch mode.
 › Press Enter to trigger a test run.
```

These two features are very valuable.
They allow me to focus on a subset of tests when I am working on a specific part of the system.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $80.00)

- [<img src='https://avatars3.githubusercontent.com/u/44827199?v=4' alt='issuehunt' width=24 height=24> issuehunt](https://oss.issuehunt.io/u/issuehunt) ($80.00)

### Submitted pull Requests
- [#3372 Implemented Watch mode filter by filename and by test name #1530](https://oss.issuehunt.io/r/avajs/ava/pull/3372)
---

### Tips

- Checkout the [Issuehunt explorer](https://oss.issuehunt.io/r/avajs/ava/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://oss.issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

novemberborn:
Sure.

Does `Enter` rerun all tests, like `r` does with AVA? That's quite neat actually.

unional:
`Enter` reruns the filtered list of tests. They use `a` to rerun all tests.

novemberborn:
> Enter reruns the filtered list of tests. They use a to rerun all tests.

Sounds like we should borrow that too 😉

IssueHuntBot:
@issuehunt has funded $80.00 to this issue.

---
- Submit pull request via [IssueHunt](https://issuehunt.io/repos/26820798/issues/1530) to receive this reward.
- Want to contribute? Chip in to this issue via [IssueHunt](https://issuehunt.io/repos/26820798/issues/1530).
- Checkout the [IssueHunt Issue Explorer](https://issuehunt.io/issues) to see more funded issues.
- Need help from developers? [Add your repository](https://issuehunt.io/r/new) on IssueHunt to raise funds.

radasam:
Hi, Is this still worth working on?

[...]

PR #1533, opened 2017-09-24
Add support of tty raw mode

### LINKED ISSUE #1531: Change watch listening mode to raw
ava@0.22

It would be nice to change watch listening mode to raw so that users don't have to type `r + ENTER` or `u + ENTER`. Just `r` or `u` is sufficient.

I can work on a PR for that.

However, I notice the code also check for `rs`: https://github.com/avajs/ava/blob/master/lib/watcher.js#L278
which does the same thing as `r`

Any reason for that? if not, I can remove it (which simplifies the code).

unional:
Fixes #1531 

This PR mostly works except killing the process requires hitting [Ctrl+C] twice, which guess is related to the test is running in a different process.

Let me know what is the right way to fix it and I'll adjust the PR. 🌷

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -182,16 +182,11 @@ exports.run = () => {
 	if (conf.watch) {
 		try {
 			const watcher = new Watcher(logger, api, files, arrify(conf.source));
-			watcher.observeStdin(process.stdin);
+			watcher.observeStdin(process.stdin).then(() => {
+				logger.exit(0);
+			}).catch(handleWatchError);
 		} catch (err) {
-			if (err.name === 'AvaError') {
-				// An AvaError may be thrown if `chokidar` is not installed. Log it nicely.
-				console.error(`  ${colors.error(figures.cross)} ${err.message}`);
-				logger.exit(1);
-			} else {
-				// Rethrow so it becomes an uncaught exception
-				throw err;
-			}
+			handleWatchError(err);
 		}
 	} else {
 		api.run(files)
@@ -206,5 +201,16 @@ exports.run = () => {
 					throw err;
 				});
 			});
+	}
+
+	function handleWatchError(err) {
+		if (err.name === 'AvaError') {
+			// An AvaError may be thrown if `chokidar` is not installed. Log it nicely.
+			console.error(`  ${colors.error(figures.cross)} ${err.message}`);
+			logger.exit(1);
+		} else {
+			// Rethrow so it becomes an uncaught exception
+			throw err;
+		}
 	}
 };
--- a/docs/recipes/watch-mode.md
+++ b/docs/recipes/watch-mode.md
@@ -89,11 +89,11 @@ If you run AVA in your CI with watch mode, the execution will exit with an error
 
 ## Manually rerunning all tests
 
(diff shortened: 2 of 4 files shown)


novemberborn:
> This PR mostly works except killing the process requires hitting [Ctrl+C] twice, which guess is related to the test is running in a different process.

I think this is what made us revert https://github.com/avajs/ava/pull/658. How does Jest handle this?

unional:
Jest just `process.exit(0)`
https://github.com/facebook/jest/blob/9c429cd15d639ab42cfe0a9fa7032095af3b675f/packages/jest-cli/src/watch.js#L143
My guess is that they put the watcher in the main process?

I add this PR as a prelude of #1530, so that user can press 'p' and then type in the filter string and 'Enter' to execute.

unional:
Once we figure out how to solve the double [Ctrl-C], I can change this PR to use `keypress` similar to #658 to get the code coverage up.

lib/cli.js
```diff
@@ -182,16 +182,11 @@ exports.run = () => {
 	if (conf.watch) {
 		try {
 			const watcher = new Watcher(logger, api, files, arrify(conf.source));
-			watcher.observeStdin(process.stdin);
+			watcher.observeStdin(process.stdin).then(() => {
+				logger.exit(0);
+			}).catch(handleWatchError);
```
novemberborn: I don't quite see how the new `observeStdin` would throw this error where the old one doesn't. Why does it need to be handled here?

lib/cli.js
```diff
@@ -182,16 +182,11 @@ exports.run = () => {
 	if (conf.watch) {
 		try {
 			const watcher = new Watcher(logger, api, files, arrify(conf.source));
-			watcher.observeStdin(process.stdin);
+			watcher.observeStdin(process.stdin).then(() => {
+				logger.exit(0);
```
novemberborn: Instead of exiting directly, can we stop all workers and let the process exit on its own?

unional:
> To clarify, we're blocked until that's resolved, yes?

Yes. Need to figure out how to handle that. In `jest`, it just `process.exit(0)`. Is there some special handling in `ava`?

lib/cli.js
unional: Sure. How to stop those workers?
By the way, I saw in `test-worker.js`:
```js
process.on('ava-exit', () => { process.exit(0); })
```
So the watcher should fire that even in all sub-processes?

lib/cli.js
unional: Yes. The code still needs some love. `observeStdin` is returning a promise that is not rejecting.
Will fix that.

lib/cli.js
novemberborn: Not quite. The child processes should be [sent the `SIGINT` signal](https://nodejs.org/api/child_process.html#child_process_subprocess_kill_signal). See https://github.com/avajs/ava/blob/bb91862eaac1a3e89f651afb15ec9691745ee222/lib/fork.js.

novemberborn:
Hi @unional, I'm sorry to see that this is closed. Did you come to the conclusion that this isn't possible the way we currently want it, or did you close it because your otherwise busy, or perhaps for another reason?

unional:
I think this can be better addressed when the code is refactored. I'll give it a try on that route first before getting back to this. 🌷

PR #1538, opened 2017-09-30
Add t.log() type for Flowtype and TypeScript

Couto:
When using Flowtype, the type validation will fail since there are is no type
declared for the newly implemented `t.log()`.

This commit adds Flowtype and TypeScript declarations for `t.log()`.


novemberborn:
Lovely, thanks @Couto!

PR #1539, opened 2017-10-01
update verbose.js to show total duration

fruch:
the way verbose was showing a timestamp, was very confusing.
I don't care much when the tests run, I want to know how much time it took.

now we have both
  21 tests passed [took: 1.6s at 14:45:18]

I think this is clearer

### DIFF
--- a/lib/reporters/verbose.js
+++ b/lib/reporters/verbose.js
@@ -95,7 +95,11 @@ class VerboseReporter {
 		].filter(Boolean);
 
 		if (lines.length > 0) {
-			lines[0] += ' ' + chalk.gray.dim('[' + new Date().toLocaleTimeString('en-US', {hour12: false}) + ']');
+			var duration = 0;
+            		runStatus.tests.forEach( test => {
+                		duration += test.duration;
+            		});
+		        lines[0] += ' ' + chalk.gray.dim('[took: ' + prettyMs(duration) + ' at ' + new Date().toLocaleTimeString('en-US', {hour12: false}) + ']');
 			output += lines.join('\n') + '\n';
 		}
 


novemberborn:
Interesting. The timestamp was introduced with https://github.com/avajs/ava/pull/737 but admittedly it's not terribly useful outside of watch mode.

We're not looking to add test duration though. There's too much variability in how tests are run, and how, when they're run concurrently, they may affect each other's performance. Printing a number doesn't make it any easier to figure out why tests are slow.

Would you be interested in changing this PR to only log the timestamps in watch mode?

fruch:
most my tests are serial anyhow, and I do care about their timing.

If you think taking timestamp from start of the run till the report, would be better, I'll add that.

I do care to see how long my tests take, total, and each one. (and if we are it, if there were a way to configure to duration threshold per test would be nice)

sindresorhus:
@fruch Please read the previous discussion about this: https://github.com/avajs/ava/pull/1038

fruch:
Now that I've read other thread, I still can't figured why we can report it... 

I care less about precision.

my use case is that I want to be able to see the run time of my test suite easily on multi platforms.

we put those tests on a precommit hook, which mean we run it quite a lot.

we want to be able to track it this number easily.

and we are a bit spoiled by previous tests runners (for example py.test) that does have this option available.

yes time( ava ) can do something similar, but some of our developer happen to be working on windows.

I get is you are trying to keep it small, and not introduce too much features 
but its shame that I'll have to keep a fork for such small thing.

novemberborn:
I'd like to see an API exposing details around test runs so this kind of reporting can be provided through a third-party module. That's not an immediate priority though.

Meanwhile, given how much the number depends on how your tests are set up, we don't want to provide it without making sure people have the context to avoid needlessly optimizing their test runs just to make that number go lower.

Thank you @fruch for opening this PR. Hopefully there are other parts of AVA you're interested in contributing to 👍

PR #1543, opened 2017-10-05
Recipe instructions for making jQuery available in browser

ptim:
I didn't readily find instructions for using jquery in the browser - hopefully this helps!
(not creating a topic branch as it's a docs PR stright from GH.. is that okay?)

### DIFF
--- a/docs/recipes/browser-testing.md
+++ b/docs/recipes/browser-testing.md
@@ -38,6 +38,15 @@ import browserEnv from 'browser-env';
 browserEnv(['window', 'document', 'navigator']);
 ```
 
+If your components use jQuery, you can make it available like so:
+```js
+import browserEnv from 'browser-env';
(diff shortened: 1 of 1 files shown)


docs/recipes/browser-testing.md
```diff
@@ -38,6 +38,15 @@ import browserEnv from 'browser-env';
 browserEnv(['window', 'document', 'navigator']);
 ```
 
+If your components use jQuery, you can make it available like so:
```
novemberborn: I wonder if you could rephrase this a little. What's interesting is that you're adding additional global variables. jQuery could be an example of that:

> You can expose more global variables by assigning them to the `global` object. For instance jQuery is typically available through the `$` variable:

ptim:
Thanks for the feedback @novemberborn! Your phrasing gives much more context - thanks!

PR #1551, opened 2017-10-16
Limit concurrency to 2 in a CI environment and 8 otherwise

lukechilds:
The new default of using CPU cores for concurrency works well most of the time, however when running tests in Docker (Travis) it will use the CPU count of the host machine, not what's been allotted to the container.

Discovered when the Got tests where timing out for no apparent reason. [I ran `os.cpus().length` on the container](https://travis-ci.org/lukechilds/got/jobs/288682178) and it returned 32. Capping at 4 resolved the issue.

This PR changes AVA to use whichever is lower out of `os.cpus().length` or `4` for concurrency by default. This seems like a sensible soft limit. You can still manually specify a concurrency that is larger than 4.

### DIFF
--- a/api.js
+++ b/api.js
@@ -161,7 +161,13 @@ class Api extends EventEmitter {
 					this._setupTimeout(runStatus);
 				}
 
-				let concurrency = Math.min(os.cpus().length, 4);
+				let concurrency = os.cpus().length;
+
+				if (process.env.CI) {
+					Math.min(concurrency, 2);
+				} else {
+					Math.min(concurrency, 8);
+				}
 
 				if (this.options.concurrency > 0) {
 					concurrency = this.options.concurrency;
--- a/api.js
+++ b/api.js
@@ -6,6 +6,7 @@ const os = require('os');
 const commonPathPrefix = require('common-path-prefix');
 const uniqueTempDir = require('unique-temp-dir');
 const findCacheDir = require('find-cache-dir');
+const isCi = require('is-ci');
 const resolveCwd = require('resolve-cwd');
 const debounce = require('lodash.debounce');
 const autoBind = require('auto-bind');
@@ -161,7 +162,7 @@ class Api extends EventEmitter {
 					this._setupTimeout(runStatus);
 				}
 
-				let concurrency = Math.min(os.cpus().length, process.env.CI ? 2 : 8);
+				let concurrency = Math.min(os.cpus().length, isCi ? 2 : 8);
 
 				if (this.options.concurrency > 0) {
 					concurrency = this.options.concurrency;
--- a/api.js
+++ b/api.js
@@ -162,7 +162,7 @@ class Api extends EventEmitter {
 					this._setupTimeout(runStatus);
 				}
 
-				let concurrency = Math.min(os.cpus().length, isCi ? 2 : 8);
+				let concurrency = Math.min(os.cpus().length, isCi ? 2 : Infinity);
 
 				if (this.options.concurrency > 0) {
 					concurrency = this.options.concurrency;
--- a/api.js
+++ b/api.js
@@ -161,7 +161,7 @@ class Api extends EventEmitter {
 					this._setupTimeout(runStatus);
 				}
 
-				let concurrency = os.cpus().length;
+				let concurrency = Math.min(os.cpus().length, 4);
 
 				if (this.options.concurrency > 0) {
 					concurrency = this.options.concurrency;
--- a/api.js
+++ b/api.js
@@ -161,13 +161,7 @@ class Api extends EventEmitter {
 					this._setupTimeout(runStatus);
 				}
 
-				let concurrency = os.cpus().length;
-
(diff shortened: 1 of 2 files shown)


sindresorhus:
Relevant Travis issue: https://github.com/travis-ci/travis-ci/issues/4696

I think we should only cap it to `4` when `process.env.CI` exists. (Maybe even lower? CIs can be pretty slow) When not on a CI, we could cap it at maybe 6?

lukechilds:
So maybe 2 on CI? That way we don't hold everything up for one slow test file. Even one CPU should be fine swapping between two threads.

And when not on a CI I'd say go even further. Maybe cap at 8. Majority of the time if this is run locally it'll be on bare metal so `os.cpus().length` will be safe to use and limit to machine thread count anyway.

Although we'll encounter the same issue on local machines if you're running your tests in docker. But then users are less likely to have a large number of threads (most likely 4 - 8) so that's unlikely to cause to many issues.

Alternatively, as this seems mostly related to Docker, we could set the limits with an [`is-docker`](https://github.com/sindresorhus/is-docker) check. Otherwise, trust os.cpus().length`.

sindresorhus:
Alright, so let's cap it at 2 for CI and 8 for normal.

@novemberborn Thoughts?

lukechilds:
Done, can be rewritten as

```js
let concurrency = Math.min(os.cpus().length, process.env.CI ? 2 : 8);
```

but thought maybe that's too much for one line?

api.js
```diff
@@ -163,6 +163,12 @@ class Api extends EventEmitter {
 
 				let concurrency = os.cpus().length;
 
+				if (process.env.CI) {
+					Math.min(concurrency, 2);
+				} else {
+					Math.min(concurrency, 8);
+				}
```
sindresorhus:
```js
Math.min(concurrency, process.env.CI ? 2 : 8);
```

?

sindresorhus: You're not assigning this to anything.

sindresorhus:
> but thought maybe that's too much for one line?

It looks fine. That's how I would have written it.

sindresorhus:
Can you also mention the capping behavior in the readme?

api.js
lukechilds: Woops, sorry running on 2 hours sleep 😬

lukechilds:
Yeah, done.

Also in the docs:

> By default, AVA will use as many processes as there are CPU cores in your machine.

Technically this should be threads not cores. Do you want me to change that?

sindresorhus:
> Technically this should be threads not cores. Do you want me to change that?

No, it's CPU cores: 

> The os.cpus() method returns an array of objects containing information about each CPU/core installed. - https://nodejs.org/api/os.html#os_os_cpus

lukechilds:
Node.js docs are misleading, it's definitely threads:

```
$ sysctl -n machdep.cpu.brand_string
Intel(R) Core(TM) i5-5257U CPU @ 2.70GHz

$ sysctl -a | grep cpu.core_count
machdep.cpu.core_count: 2

$ sysctl -a | grep cpu.thread_count
machdep.cpu.thread_count: 4

$ node -e "console.log(require('os').cpus().length)"
4
```

sindresorhus:
I see. Alright, update it to threads.

sindresorhus:
Node.js issue: https://github.com/nodejs/node/issues/16279

sindresorhus:
Let's go with `logical cores` per Node.js issue discussion and make `logical cores` link to https://superuser.com/questions/1105654/logical-vs-physical-cpu-performance

lukechilds:
Good idea, done 👍

sindresorhus:
> but I don't understand why we'd need to cap this at all for non-CI?

For cases where Docker is used locally since it reports the incorrect count. And, from experience, anything more than 4-5 doesn't improve the performance. But yes, it's pretty arbitrary, so not sure whether it makes sense.

lukechilds:
@novemberborn Just to clarify, it only caps the default from `os.cpus().length`. You can still manually specify whatever concurrency you want for CI or local.

But yeah, I like I mentioned in an above comment, it's unlikely to cause issues locally unless they have a really high thread count.

> Majority of the time if this is run locally it'll be on bare metal so os.cpus().length will be safe to use and limit to machine thread count anyway.
>
> Although we'll encounter the same issue on local machines if you're running your tests in docker. But then users are less likely to have a large number of threads (most likely 4 - 8) so that's unlikely to cause issues.

novemberborn:
@lukechilds @sindresorhus OK, I'm still leaning towards *not* capping this outside of a CI environment. Just seems like unnecessary logic.

sindresorhus:
Ok, let's drop it. We can add it back if it's an actual proven problem.

lukechilds:
@novemberborn was not aware of `is-ci`. Looks like a more robust solution 👍

novemberborn:
Thanks @lukechilds!

PR #1555, opened 2017-10-17
Print watch mode shortcut commands

### LINKED ISSUE #1525: Print watch mode shortcut commands
[Issuehunt badges](https://issuehunt.io/r/avajs/ava/issues/1525)

Print the shortcut commands for rerunning all tests (`r`) and updating snapshots used in the previous test run (`u`) after watch mode output.

Via @wmertens in https://github.com/avajs/ava/issues/1509#issuecomment-328374098.

> There is a $40.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/avajs/ava/issues/1525).

rilut:
Fixes #1525

### DIFF
--- a/lib/watcher.js
+++ b/lib/watcher.js
@@ -1,4 +1,5 @@
 'use strict';
+const chalk = require('chalk');
 const nodePath = require('path');
 const debug = require('debug')('ava:watcher');
 const diff = require('lodash.difference');
@@ -18,6 +19,8 @@ function rethrowAsync(err) {
 
 const MIN_DEBOUNCE_DELAY = 10;
 const INITIAL_DEBOUNCE_DELAY = 100;
+const rerunMessage = chalk.gray.dim("To rerun all tests, type 'r', followed by Enter\nTo update snapshots used in the previous tests, type 'u', followed by Enter\n");
+
 
 class Debouncer {
 	constructor(watcher) {
@@ -126,7 +129,7 @@ class Watcher {
 				.then(runStatus => {
 					runStatus.previousFailCount = this.sumPreviousFailures(currentVector);
 					logger.finish(runStatus);
-					logger.write('[ava] To rerun tests, enter `r`\n[ava] To update snapshots used in the previous test run, enter `u`');
+					logger.write(rerunMessage);
 
 					const badCounts = runStatus.failCount + runStatus.rejectionCount + runStatus.exceptionCount;
 					this.clearLogOnNextRun = this.clearLogOnNextRun && badCounts === 0;
--- a/lib/watcher.js
+++ b/lib/watcher.js
@@ -1,6 +1,6 @@
 'use strict';
-const chalk = require('chalk');
 const nodePath = require('path');
+const chalk = require('chalk');
 const debug = require('debug')('ava:watcher');
 const diff = require('lodash.difference');
 const chokidar = require('chokidar');
@@ -19,8 +19,7 @@ function rethrowAsync(err) {
 
 const MIN_DEBOUNCE_DELAY = 10;
 const INITIAL_DEBOUNCE_DELAY = 100;
-const rerunMessage = chalk.gray.dim("To rerun all tests, type 'r', followed by Enter\nTo update snapshots used in the previous tests, type 'u', followed by Enter\n");
-
+const rerunMessage = chalk.gray.dim(`To rerun all tests, type 'r', followed by Enter\nTo update snapshots used in the previous tests, type 'u', followed by Enter\n`);
 
 class Debouncer {
 	constructor(watcher) {
--- a/lib/watcher.js
+++ b/lib/watcher.js
@@ -126,6 +126,7 @@ class Watcher {
 				.then(runStatus => {
 					runStatus.previousFailCount = this.sumPreviousFailures(currentVector);
 					logger.finish(runStatus);
+					logger.write('[ava] To rerun tests, enter `r`\n[ava] To update snapshots used in the previous test run, enter `u`');
 
 					const badCounts = runStatus.failCount + runStatus.rejectionCount + runStatus.exceptionCount;
 					this.clearLogOnNextRun = this.clearLogOnNextRun && badCounts === 0;
--- a/test/cli.js
+++ b/test/cli.js
@@ -325,7 +325,9 @@ test('watcher does not rerun test files when they write snapshot files', t => {
 				killed = true;
 			}, 500);
 		} else if (passedFirst && !killed) {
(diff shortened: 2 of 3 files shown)


lib/watcher.js
```diff
@@ -123,6 +123,7 @@ class Watcher {
 				.then(runStatus => {
 					runStatus.previousFailCount = this.sumPreviousFailures(currentVector);
 					logger.finish(runStatus);
+					logger.write('[ava] To rerun tests, enter `r`\n[ava] To update snapshots used in the previous test run, enter `u`');
```
novemberborn: I don't think this needs to be prefixed with `[ava]`. Currently the key needs to be followed by `Enter` in order to be recognized. How about:

> To rerun all tests, type 'r', followed by Enter
> To update snapshots used in the previous tests, type 'u', followed by Enter

Perhaps using `chalk.gray.dim` as the color.

novemberborn:
Hi @rilut, just wondering if there's more commits forthcoming? Would be good to have some assertions on the `logger.write()` spy, like is done for the other logger methods. Thanks!

rilut:
Oh sure, will do

novemberborn:
@rilut thanks for following up. I've just restarted CI, so hopefully all the tests will pass. Still need to give this a whirl, have been a bit busy sorry.

novemberborn:
Ah no, the CLI integration tests are failing because there's new output. Could you look into that @rilut? 😄

rilut:
@novemberborn I'll look into it

test/cli.js
```diff
@@ -325,7 +325,9 @@ test('watcher does not rerun test files when they write snapshot files', t => {
 				killed = true;
 			}, 500);
 		} else if (passedFirst && !killed) {
-			t.is(buffer.replace(/\s/g, ''), '');
+			const rerunMessage = `To rerun all tests, type 'r', followed by Enter\nTo update snapshots used in the previous tests, type 'u', followed by Enter\n`;
+			const rerunMessageWithoutWhitespace = buffer.replace(/\s/g, '')
+			t.is(buffer.replace(/\s/g, ''), rerunMessageWithoutWhitespace);
```
rilut: @novemberborn not sure if asserting this is the right way to do this

test/cli.js
```diff
@@ -325,7 +325,9 @@ test('watcher does not rerun test files when they write snapshot files', t => {
 				killed = true;
 			}, 500);
 		} else if (passedFirst && !killed) {
-			t.is(buffer.replace(/\s/g, ''), '');
+			const rerunMessage = `To rerun all tests, type 'r', followed by Enter\nTo update snapshots used in the previous tests, type 'u', followed by Enter\n`;
+			const rerunMessageWithoutWhitespace = rerunMessage.replace(/\s/g, '');
+			t.is(buffer.replace(/\s/g, ''), rerunMessageWithoutWhitespace);
```
rilut: @novemberborn I'm not sure if what I did (asserting this) is the right way to do this

rilut:
not sure what happened with the node:6 FreshDeps=false tests

rilut:
```test/cli.js ....................................... 138/140 5m
  snapshots work (tests)
  not ok timeout!
    signal: SIGTERM
    handles:
      - type: ChildProcess
        events:
          - close
      - type: Socket
        events:
          - end
          - finish
          - _socketEnd
      - type: Socket
        events:
          - end
          - finish
          - _socketEnd
          - close
          - error
          - data
      - type: Socket
        events:
          - end
          - finish
          - _socketEnd
          - close
          - error
          - data
    expired: TAP
    stack: |
      emit (node_modules/nyc/node_modules/signal-exit/index.js:77:11)
      process.listener (node_modules/nyc/node_modules/signal-exit/index.js:91:7)
      processEmit (node_modules/nyc/node_modules/signal-exit/index.js:155:32)
      process.emit (node_modules/source-map-support/source-map-support.js:439:21)
      processEmit (node_modules/signal-exit/index.js:155:32)
    test: snapshots work (tests)
  
  not ok test count !== plan
    +++ found
    --- wanted
    -1
    +61
    results:
      ok: false
      count: 61
      pass: 60
      fail: 2
      bailout: false
      todo: 0
      skip: 0
      plan:
        start: null
        end: null
        skipAll: false
        skipReason: ''
        comment: ''
      failures:
        - ok: false
          id: 61
          time: 5460.13
          name: snapshots work (tests)
        - tapError: no plan
```

lib/watcher.js
```diff
@@ -18,6 +19,7 @@ function rethrowAsync(err) {
 
 const MIN_DEBOUNCE_DELAY = 10;
 const INITIAL_DEBOUNCE_DELAY = 100;
+const rerunMessage = chalk.gray.dim(`To rerun all tests, type 'r', followed by Enter\nTo update snapshots used in the previous tests, type 'u', followed by Enter\n`);
```
novemberborn: Looking at this in my terminal I think this should just be `chalk.gray`. But we then should also update the timestamp output in the mini & verbose loggers to not be dimmed.

test/cli.js
novemberborn: Most tests use [`t.match`](http://www.node-tap.org/asserts/#tmatchfound-pattern-message-extra) for this. You could do an assertion for each line.

lib/watcher.js
```diff
@@ -126,6 +128,7 @@ class Watcher {
 				.then(runStatus => {
 					runStatus.previousFailCount = this.sumPreviousFailures(currentVector);
 					logger.finish(runStatus);
+					logger.write(rerunMessage);
```
novemberborn: Unfortunately this doesn't work with the mini reporter. I think `finish()` should take an optional second parameter, call it `footer` perhaps, and both the mini and verbose loggers can then include this in their regular output.

sindresorhus:
screen shot 2017-12-13 at 13 40 23

It shouldn't reprint the message when I press `r`+`Enter`.

rilut:
@sindresorhus Maybe we just write the rerun message on the first start?

rilut:
@sindresorhus How do I write the rerunMessage that can be deleted after the `status` in MiniLogger.finish? Using `ansiEscapes.eraseLines(ct);`?

novemberborn:
@rilut I think what @sindresorhus is seeing is due to my earlier comment:

> Unfortunately this doesn't work with the mini reporter: the test output itself is overwritten.

My suggestion for resolving this is https://github.com/avajs/ava/pull/1555/files#r153833345.

novemberborn:
Closing this since it has stalled. If anybody is looking to revive this PR please check in first, I'm considering refactoring the reporter pipeline which impacts this PR.

PR #1561, opened 2017-10-22
TypeScript recipe: add "working with macros"

mightyiam:
Hey, here is a little something that works for me. Please review carefully, because I don't really know whether I'm doing it correctly. It just seems to work.


novemberborn:
@mightyiam what happens without this type definition? I'd have expected this to work out of the box, that's why the `Macro` interface exists.

mightyiam:
@novemberborn 

My `macro`, like any other macro—I figure—is a function that is created by the user.

What do you mean by "out of the box", please?

novemberborn:
I'm confused as to why you're explicitly typing `const macro: Macro<AssertContext>`. Why is that necessary?

mightyiam:
I am explicitly typing `macro` as `Macro` in order to be able to assign to `macro.title`.

mightyiam:
@novemberborn better?

docs/recipes/typescript.md
```diff
@@ -50,6 +50,23 @@ test(async (t) => {
...
 ```
 
+## Working with [macros](https://github.com/avajs/ava#test-macros)
+
+In order to be able to assign the `title` property to a macro:
+
+```ts
+import test, { AssertContext, Macro } from 'ava';
+
+const macro: Macro<AssertContext> = (t: AssertContext, input, expected) => {
```
novemberborn: Looking at this in VS Code the type of `t` is inferred. Given how often you'd end up typing that parameter perhaps the example shouldn't explicitly type it.

mightyiam: You're right, because:
https://github.com/avajs/ava/blob/64b7755d97cba07ed62d797c6ce2f85ffa11a01e/types/base.d.ts#L118

docs/recipes/typescript.md
```diff
@@ -50,6 +50,23 @@ test(async (t) => {
...
 ```
 
+## Working with [macros](https://github.com/avajs/ava#test-macros)
+
+In order to be able to assign the `title` property to a macro:
+
+```ts
+import test, { AssertContext, Macro } from 'ava';
+
+const macro: Macro<AssertContext> = (t, input, expected) => {
```
mightyiam: `t` is implicit, @novemberborn.

novemberborn:
Thanks @mightyiam!

Issue #1554, opened 2017-10-17
Include babel-plugin-syntax-object-rest-spread in default Babel config


novemberborn:
As of 8.3.0, Node.js supports object rest/spread without any flags. Out of the box Babel does not recognize this syntax though. We should include the `babel-plugin-syntax-object-rest-spread` plugin in our default Babel config.

Note that we only support stage-4 features, and object rest/spread is still stage-3. Thus the question is for which Node.js versions do we include this plugin? I'm leaning towards 8.3.0 and above. This would mean that if you're targeting just those versions you can use object rest/spread with AVA without having to make any customizations. If you're targeting older versions you'll need to configure the full `babel-plugin-transform-object-rest-spread`, which comes with the syntax plugin.

PR #1564, opened 2017-10-23
Include babel object rest spread plugin for node versions >= 8.3.0

codeslikejaggars:
First thoughts on fixing #1554 by checking the node version.

I'd like to also add tests, but could use some advice on how to augment the existing test suite to handle integration tests for different node versions.


novemberborn:
Very nice @codeslikejaggars!

Regarding testing, I'd add some tests to `test/babel-config.js` which compare the `plugins` to see if `babel-plugin-syntax-object-rest-spread` is included. Override `process.versions.node` so that there's a test with a version older than 8.6.0, then 8.6.0 exactly, and then say a 9.0.0 version.

We could do an integration test but that's tricky since we need to test the Node.js version to see whether the test should pass or fail… let's instead just see if the Babel config is generated correctly.

codeslikejaggars:
Thanks @novemberborn, I added a few tests that update the node version and validate the output.

test/babel-config.js
```diff
@@ -9,6 +9,14 @@ const configManager = require('hullabaloo-config-manager');
 const babelConfigHelper = require('../lib/babel-config');
 
 const fixture = name => path.join(__dirname, 'fixture', name);
+const setNodeVersion = value => Object.defineProperty(process.versions, 'node', {value});
+const resetNodeVersion = setNodeVersion.bind(null, process.versions.node);
+
+// Execute fn with a given stubbed node version, then reset to the real version
+const withNodeVersion = (version, fn) => {
+	setNodeVersion(version);
+	return Promise.resolve(fn).then(() => resetNodeVersion());
```
novemberborn: You never actually execute `fn` though 😉 

I'd return the result of `fn` which makes avoids some odd indentation in the tests themselves. For example:

```js
function withNodeVersion (version, run) {
  setNodeVersion(version);
  const promise = new Promise(resolve => {
    resolve(run());
  });
  promise.then(resetNodeVersion, resetNodeVersion);
  return promise;
}
```

codeslikejaggars:
😬 yep.. should probably actually execute the tests

novemberborn:
Hey @codeslikejaggars I've pushed some minor changes, and added some documentation.

Turns out Node.js has supported this since 8.3.0 so I've adjusted the version numbers accordingly.

Please let me know if this all looks OK to you! 😄

codeslikejaggars:
👍 Looks good to me, thanks @novemberborn

novemberborn:
@codeslikejaggars had to force push to fix a merge conflict, sorry.

novemberborn:
Thank you @codeslikejaggars 🥇

PR #1568, opened 2017-10-25
Fixes #1553. t.is() prompts to user when objects are different references

### LINKED ISSUE #1553: Improve t.is output when objects are different references
## Description

The output for a failed `t.is()` assertion should be clearer when the `actual` and `expected` arguments are identical objects but not the same reference.

Object diffing works well when their properties are different, but not when the references are different. Specifically, the issue is that there is nothing useful in the diff and no additional message explaining the actual error.

Since equality in JavaScript is a frequent source of confusion for beginners, I think fixing this will be particularly helpful for those users.

In my case, I changed an API from modifying its input to returning a copy. So I know the solution is `t.deepEqual()`, but I only know that because I understand what is happening under the covers.

### Test Source

```js
test(() => {
    t.is({ hi : 'bye' }, { hi : 'bye' });
});
```

### Output

```
  Difference:

    {
      hi: 'bye',
    }
```

### Config

Everything default.

## Environment

Node 8.6.0
npm 5.4.2
ava 0.22.0
darwin 17.0.0

Lifeuser:
t.is() performs deep equality check on objects and prompts to user if objects are equal.
Test for this assertion is changed.

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -89,7 +89,7 @@ function wrapAssertions(callbacks) {
 			} else {
 				const result = concordance.compare(actual, expected, concordanceOptions);
 				// Primitive values should be the same, so if actual or expected is primitive
-  				// then the values will never compare. No result.actual and result.expected values returned
+				// then the values will never compare. No result.actual and result.expected values returned
 				const notPrimitives = result.actual && result.expected;
 				if (result.pass && notPrimitives) {
 					return fail(this, new AssertionError({
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -92,20 +92,20 @@ function wrapAssertions(callbacks) {
 				const expectedDescriptor = result.expected || concordance.describe(expected, concordanceOptions);
 
 				if (result.pass) {
-					return fail(this, new AssertionError({
+					fail(this, new AssertionError({
 						assertion: 'is',
 						message,
 						raw: {actual, expected},
 						values: [formatDescriptorWithLabel('Values are deeply equal to each other, but they are not the same:', actualDescriptor)]
 					}));
+				} else {
+					fail(this, new AssertionError({
+						assertion: 'is',
+						message,
+						raw: {actual, expected},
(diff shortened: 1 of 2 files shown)


lib/assert.js
```diff
@@ -87,8 +87,18 @@ function wrapAssertions(callbacks) {
...
 			} else {
-				const actualDescriptor = concordance.describe(actual, concordanceOptions);
-				const expectedDescriptor = concordance.describe(expected, concordanceOptions);
+				const result = concordance.compare(actual, expected, concordanceOptions);
+				if (result.pass) {
+					return fail(this, new AssertionError({
+						assertion: 'is',
+						message,
+						raw: {actual, expected},
+						values: [formatWithLabel(`Objects are identical, but it's not the same object:`, actual)]
```
novemberborn: This should use a regular string, not a template string.

As for the label itself, how about:

> Values are deeply equal to each other, but they're not the same:

This references both `t.deepEqual` and the ["Assert `value` is the same as `expected`" explanation](https://github.com/avajs/ava/tree/3b81e2ca13db91e558234f55badc312209677d82#isvalue-expected-message) of `t.is()`.

Lifeuser: Label sounds good for me!

I used it because of "it's". Linting style doesn't allow doublequotes.
So I changed your label a bit for this reason (they're ->> they are) and used regular string.

I also moved assignments up, so we don't run .describe second time in deeply equal case.)

novemberborn: > I also moved assignments up, so we don't run .describe second time in deeply equal case.)

Nice!

> I used it because of "it's". Linting style doesn't allow doublequotes.
> So I changed your label a bit for this reason (they're ->> they are) and used regular string.

I don't know if we're consistent in this regard, but `’` is a good apostrophe to use. Or else the single quote can be escaped. Anyway, the copy is fine now.

novemberborn:
@Lifeuser I pushed a tiny fixup, but this is good to go as soon as CI passes. Thank you!

PR #1569, opened 2017-10-25
Use babel-generator to regenerate enhanced assertion statements (#1513)

### LINKED ISSUE #1513: Missing space between var1 instanceof var2 in output
## Description

When a test fails while using `test.true(var1 instanceof var2)` the output removes the spaces.

### Test Source

```js
test('returns itself', t => {                
    t.true(t.context.script instanceof Script);
});
```

```
  1 failed
  1 rejection

  returns itself

  /Users/xo/code/@lad/mongoose/test/test.js:10

    9: test('returns itself', t => {                
   10:   t.true(t.context.script instanceof Script);
   11: });                                          

  Value is not `true`:

  false

  t.context.scriptinstanceofScript
  => false
```

### Config

```json
{
  "ava": {
  ...
  }
}
```

### Command-Line Arguments

```
ava
```

## Environment

```
Node.js v8.4.0
darwin 16.7.0
ava 0.22.0
npm 5.3.0
```

codeslikejaggars:
Fixes #1513 by using babel-generator to regenerate statements from the AST instead of concatenating tokens without whitespace. 

Note that this doesn't totally rematerialize the original syntax since the tokens included with the power assert context don't retain quote or whitespace information. This implementation outputs strings with single quotes to match the output of concordance.

### DIFF
--- a/lib/enhance-assert.js
+++ b/lib/enhance-assert.js
@@ -16,12 +16,7 @@ const PATTERNS = [
 	't.notRegex(contents, regex, [message])'
 ];
 
-const computeStatement = node => {
-	return generate(node, {
-		quotes: 'single'
-	}).code;
-};
-
+const computeStatement = node => generate(node, {quotes: 'single'}).code;
 const getNode = (ast, path) => dotProp.get(ast, path.replace(/\//g, '.'));
 
 const formatter = context => {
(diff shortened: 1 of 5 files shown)


novemberborn:
> Note that this doesn't totally rematerialize the original syntax since the tokens included with the power assert context don't retain quote or whitespace information.

Yea that's fine.



> This implementation outputs strings with single quotes to match the output of concordance.

Nice touch 😄

novemberborn:
🎉

PR #1570, opened 2017-10-25
Dont run before and after hooks when all tests are skipped (#1283)

### LINKED ISSUE #1283: Should before and after run with skip?
ava: 0.18.2

```js
import test from 'ava'

test.before(() => {
  throw new Error('should not run?')
})
test.after(() => {
  throw new Error('should not run?')
})

test.skip('some skipped test', t => { })
```

Currently the `before()` and `after()` is run even if the file contain only skipped tests.
Is that the expected behavior?

codeslikejaggars:
Attempting to fix #1283. This change removes the `before` and `after` hooks from the run sequence when all tests in the collection are skipped. It intentionally keeps the `always` hook in the run sequence, since it would probably be surprising to a user if `after.always` didn't run.

### DIFF
--- a/lib/test-collection.js
+++ b/lib/test-collection.js
@@ -178,19 +178,32 @@ class TestCollection extends EventEmitter {
 	_buildTests(tests) {
 		return tests.map(test => this._buildTestWithHooks(test));
 	}
+	_hasUnskippedTests() {
+		const getUnskippedCount = tests => tests.filter(test => {
+			return !(test.metadata && test.metadata.skipped === true);
+		}).length;
+		return (getUnskippedCount(this.tests.serial) + getUnskippedCount(this.tests.concurrent)) > 0;
+	}
 	build() {
-		const beforeHooks = new Sequence(this._buildHooks(this.hooks.before));
-		const afterHooks = new Sequence(this._buildHooks(this.hooks.after));
-
 		const serialTests = new Sequence(this._buildTests(this.tests.serial), this.bail);
 		const concurrentTests = new Concurrent(this._buildTests(this.tests.concurrent), this.bail);
 		const allTests = new Sequence([serialTests, concurrentTests]);
 
-		let finalTests = new Sequence([beforeHooks, allTests, afterHooks], true);
+		let finalTests;
+		// Only run before and after hooks when there are unskipped tests
+		if (this._hasUnskippedTests()) {
+			const beforeHooks = new Sequence(this._buildHooks(this.hooks.before));
+			const afterHooks = new Sequence(this._buildHooks(this.hooks.after));
+			finalTests = new Sequence([beforeHooks, allTests, afterHooks], true);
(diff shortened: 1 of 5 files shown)


lib/test-collection.js
```diff
@@ -178,19 +178,32 @@ class TestCollection extends EventEmitter {
 	_buildTests(tests) {
 		return tests.map(test => this._buildTestWithHooks(test));
 	}
+	_hasUnskippedTests() {
+		const getUnskippedCount = tests => tests.filter(test => {
+			return !(test.metadata && test.metadata.skipped === true);
+		}).length;
+		return (getUnskippedCount(this.tests.serial) + getUnskippedCount(this.tests.concurrent)) > 0;
```
novemberborn: This would be clearer with [`Array#some()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some).

lib/test-collection.js
```diff
@@ -178,19 +178,32 @@ class TestCollection extends EventEmitter {
...
 		const concurrentTests = new Concurrent(this._buildTests(this.tests.concurrent), this.bail);
 		const allTests = new Sequence([serialTests, concurrentTests]);
 
-		let finalTests = new Sequence([beforeHooks, allTests, afterHooks], true);
+		let finalTests;
+		// Only run before and after hooks when there are unskipped tests
+		if (this._hasUnskippedTests()) {
+			const beforeHooks = new Sequence(this._buildHooks(this.hooks.before));
+			const afterHooks = new Sequence(this._buildHooks(this.hooks.after));
+			finalTests = new Sequence([beforeHooks, allTests, afterHooks], true);
```
novemberborn: You could create the `allTests` sequence in this branch.

lib/test-collection.js
```diff
@@ -178,19 +178,32 @@ class TestCollection extends EventEmitter {
...
+		} else {
+			finalTests = new Sequence([allTests], true);
```
novemberborn: This could be an empty sequence, since presumably all tests are skipped if it gets to this point. That would make the logic easier to grasp IMHO.

codeslikejaggars: I don't think that would work. If it's an empty sequence, the skipped tests won't be reported in the runner output.

lib/test-collection.js
codeslikejaggars: Not sure I can, see response to the comment below.

codeslikejaggars:
Updated the readme and added a test for `after.always`.

lib/test-collection.js
novemberborn: Ah that makes sense 👍

novemberborn:
Awesome work @codeslikejaggars 🚀

sindresorhus:
Great work @codeslikejaggars 👌✨🎉

PR #1574, opened 2017-10-27
Ignore nested ignored directories (#1501)

codeslikejaggars:
Update file matching code to ignore the [ignore-by-default](https://github.com/novemberborn/ignore-by-default/blob/master/index.js) directories at any depth by prepending `**/`.

It's definitely worth considering whether it makes sense to ignore all those directories at any depth. For example, it may make sense to ignore files in `node_modules`, `.git`, and `.nyc_output` at any depth, but not `coverage` which would be a reasonable module name in some projects.

### DIFF
--- a/lib/ava-files.js
+++ b/lib/ava-files.js
@@ -94,7 +94,8 @@ const defaultHelperPatterns = () => [
 	'**/test/**/_*.js'
 ];
 
-const getDefaultIgnorePatterns = () => defaultIgnore.map(dir => `${dir}/**/*`);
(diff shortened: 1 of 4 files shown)


novemberborn:
Thanks for your hard work on this @codeslikejaggars. We have a lot of issues to do with globbing and performance and I have a feeling that those issues are all connected, and require a holistic approach. I need to do more thinking and research on this, and I'd like to avoid making any changes until then.

codeslikejaggars:
Makes sense @novemberborn

PR #1586, opened 2017-11-14
Update readme to remove unreleased feature

ahorrocks2:
My code (using Node 8.8.1) was throwing syntax errors on the spread operator, which was confusing since I had read the documentation and was sure that ava would pull in the necessary babel plugin. I saw that it had not actually pulled the plugin in and almost put an issue in until I noticed that the fix was submitted 17 days ago and the last release was 20 days ago. 

I think that this information should be removed from the documentation until the issue fix is released. Thanks!


novemberborn:
I'm sorry about the confusion. Without a dedicated documentation website this is bound to happen. We're quite happy to receive documentation updates along with pull requests when new features are introduced. It wouldn't be feasible to land these later when we do a release.

For future reference (and for others coming across this PR), you can browse the recipes by version tag though, see the `v0.23.0` bit in the URL here: https://github.com/avajs/ava/tree/v0.23.0/docs/recipes.

Thank you for your PR and interest in AVA.

PR #1593, opened 2017-11-17
Add ES Modules recipe

### LINKED ISSUE #1590: HOW-TO: ES Modules with @std/esm
If you're using [@std/esm](https://github.com/standard-things/esm) for your projects, `ava` will not be able to read your `import test from 'ava';` if you're not transpiling ES modules.

The following configuration in `package.json` will help you out:

```json
{
...
"@std/esm": {
  "esm": "all",
  "cjs": true,
  "gz": true,
  "await" true
},
"ava": {
  "require": [
    "@std/esm"
  ]
}
...
}
```

Hope this helps! 🎉

motss:
@sindresorhus 

To fix #1590 

Recipe for ava + @std/esm + typescript


docs/recipes/es-modules-with-@std-esm.md
```diff
@@ -0,0 +1,146 @@
...
+    "require": [
+      "@std/esm"
+    ]
+  },
+  "@std/esm": {
+    "esm": "all",
+    "cjs": true,
+    "await": true,
+    "gz": true
+  }
```
jdalton: I don't believe all those options are needed. Can you try snipping it down to something like:

```
"@std/esm": "cjs"
```

motss: @jdalton Sure.

jdalton: Sweet 🍬 !

sindresorhus:
Thanks for starting this @motss. I've tweaked the text and simplified parts of it a little bit.

Can you also add it to the AVA readme: https://github.com/avajs/ava#recipes

docs/recipes/es-modules-with-@std-esm.md
```diff
@@ -0,0 +1,109 @@
+# ES Modules
+
+As of Node.js 8.5.0, [ES modules](http://2ality.com/2017/09/native-esm-node.html) are natively supported, but behind a command line option `--experimental-modules`.
+
+However, it works a bit different on Node.js than in web browsers that also natively supports ES modules. In Node.js, you need to use the `.mjs` extension instead of the commonly known `.js` extension. The `.mjs` extension in web browsers too, but they have to be served with the correct media type (`application/javascript`). There is a [ongoing effort](https://tools.ietf.org/html/draft-bfarias-javascript-mjs-00) on standardizing `.mjs`.
+
+The good news is that there's a module called [`@std/esm`](https://github.com/standard-things/esm) that enable you to write and run ES modules in Node.js 4 and above.
```
jdalton: that enable -> that enables

docs/recipes/es-modules-with-@std-esm.md
```diff
@@ -0,0 +1,109 @@
...
+		"require": [
+			"@std/esm"
+		]
+	},
+	"@std/esm": "cjs"
+}
+```
+
+You can now write tests with AVA, `@std/esm`, and TypeScript:
+
```
jdalton: 👆 So rad!

docs/recipes/es-modules-with-@std-esm.md
```diff
@@ -0,0 +1,109 @@
...
+```
+$ npm install --save-dev typescript
+```
+
+Then modify your `tsconfig.json` accordingly:
+
+```json
+{
+	…
+	"moduleResolution": "node",
```
jdalton: 👆 is the `...` in the json common in the recipes? It makes a weird error highlight when using ```json

sindresorhus: It's just a way to signify that there should be more content than just this in the file. Any better suggestion?

jdalton:
Let me know if a specific usage pops out for AVA. I wouldn't mind making `@std/esm` detect AVA and applying AVA specific defaults or something.

docs/recipes/es-modules-with-@std-esm.md
```diff
@@ -0,0 +1,109 @@
+# ES Modules
+
+As of Node.js 8.5.0, [ES modules](http://2ality.com/2017/09/native-esm-node.html) are natively supported, but behind a command line option `--experimental-modules`.
+
+However, it works a bit different on Node.js than in web browsers that also natively supports ES modules. In Node.js, you need to use the `.mjs` extension instead of the commonly known `.js` extension. The `.mjs` extension in web browsers too, but they have to be served with the correct media type (`application/javascript`). There is a [ongoing effort](https://tools.ietf.org/html/draft-bfarias-javascript-mjs-00) on standardizing `.mjs`.
```
novemberborn: I think this should say:

> The `.mjs` extension can work in web browsers too, but such files have to be served with the correct media type

And:

> There is an [ongoing effort](https://tools.ietf.org/html/draft-bfarias-javascript-mjs-00) on standardizing `.mjs`.

(`an` rather than `a`)

docs/recipes/es-modules-with-@std-esm.md
```diff
@@ -0,0 +1,109 @@
...
+	"@std/esm": "cjs"
+}
+```
+
+You can now use native ES modules with AVA:
+
+```js
+import test from 'ava';
+
+test('2 + 2 = 4', t => {
```
novemberborn: It'd be nice if this would have an actual example of a `.mjs` file. Notably the test files are still transpiled using Babel, which means the above really reads` const test = require('ava')`, and there are no `.mjs` files being used. Same goes for the TypeScript example below.

jdalton:
I'm fine not dwelling on mjs. The recommended config is "@std/esm":"cjs", which enables cjs compat mode.

docs/recipes/es-modules-with-@std-esm.md
```diff
@@ -0,0 +1,109 @@
...
+Modify your package.json accordingly:
+
+```json
+{
+	…
+	"scripts": {
+		"test": "ava"
+	},
+	"dependencies": {
+		"@std/esm": "^0.16.0"
```
jdalton: Latest is `0.17.0` now. Is there a standard way to express a dependency without using a version for the recipe so folks don't lazy copy pasta an out of date version?

novemberborn: I don't know… perhaps the recipe should say to use `npm install` and not show any `package.json`? That would make copying really hard.

motss:
@sindresorhus @jdalton Does the PR look good? Ready to merge?

motss:
See avajs/ava#1618

sindresorhus:
What do we do about this PR now that #1618 is merged?

jdalton:
The general advice is still solid. Maybe bump the version documented and run with it.

feross:
You might want to update this PR to point out that `.mjs` files won't work with ava until https://github.com/avajs/ava/issues/631#issuecomment-357733734 is fixed.

Running `ava` or `ava test/**/*.mjs` or `ava test/my-test.mjs` results in:

```
  1 exception

  ✖ Couldn't find any files to test
```

This was surprising to me since the documentation in this PR mentions the `.mjs` explicitly implying that it works!

jdalton:
The `"@std/esm": "cjs"` could technically be `"@std/esm": "js"` since it would no longer need the cjs babel interop to be enabled.

novemberborn:
I've pushed a rewrite that focuses this recipe on using `@std/esm`, with AVA 1.0 in mind.

@jdalton is the configuration OK, and the example accurate?

jdalton:
Looks great! Let's roll with that configuration and loosen it if needed later.

docs/recipes/es-modules.md
```diff
@@ -0,0 +1,47 @@
...
+	"ava": {
+		"require": [
+			"@std/esm"
+		]
+	},
+	"@std/esm": "js"
+}
+```
+
+By default AVA converts ES module syntax to CommonJS. [You can disable this](./babel.md#preserve-es-module-syntax).
```
sindresorhus: You don't need to prefix the file path with `./`. It works without too.

novemberborn:
Thanks all!

motss:
🎉 Thanks for merging the PR. Well done all! 👏

Issue #1158, opened 2016-12-19
--fail-fast keeps running other test files


novemberborn:
The test process that encounters the error simply [exits if `--fail-fast` is enabled](https://github.com/avajs/ava/blob/033d4dcdcbdadbf665c740ff450c2a775a8373dc/lib/main.js#L44:L46). Since each file runs in its own process, no other tests from that file are reported after the first failure is encountered.

However, other test processes run to completion. I'm not sure whether to:

* terminate them outright, which prevents any cleanup, even if there are no failures in those processes
* silence their output, which at least means the test output isn't cluttered with passing tests, but makes it harder to see why `ava` won't exit
* shrug and say that this is expected

When running with limited concurrency we'll also run into this scenario, so we should implement the same behavior. Additionally we mustn't launch new test processes.

Any thoughts on what to do with processes that are already running?

novemberborn:
You can reproduce this issue by creating `one.test.js`:

```js
import test from 'ava'

test(t => t.fail())
```

and `two.test.js`:

```js
import test from 'ava'

test(t => t.pass())
```

Then `ava --verbose --fail-fast`.

ThomasBem:
I also noticed this while working on issue #1134 today. Here is a screenshot of me trying to run two different files both with a failing test somewhere in the middle.

![screen shot 2016-12-21 at 19 58 17](https://cloud.githubusercontent.com/assets/3599597/21402236/f458f636-c7b7-11e6-9ea0-c7f7f8f6091c.png)

When things are as small as these examples it does not seem like much of a problem. But if you are running a huge test suite for a complex application. Then allowing Ava to keep running tests kind of defeats the purpose of --fail-fast in my opinion. 

So to keep the --fail-fast option working as intended, i would say shutdown any running processes. But i don´t know what problems that might cause since you talk about preventing cleanup:
> terminate them outright, which prevents any cleanup, even if there are no failures in those processes

mightyiam:
My thoughts:

1. do not run any more test files
1. do not run any more tests
1. allow tests to complete, including their `afterEach`
1. allow test files to complete, including their `after`

Only then, exit.

novemberborn:
> So to keep the --fail-fast option working as intended, i would say shutdown any running processes. But i don´t know what problems that might cause since you talk about preventing cleanup

[...]

PR #1596, opened 2017-11-22
Draft: list 'pending' tests

### LINKED ISSUE #583: Print pending tests when interrupting AVA, or after a timeout

ace-n:
Fixes #583.

**N.B:** I haven't added any _new_ tests yet - but I plan to once the implementation details are nailed down.

Example _test.js_
```
const test = require(`ava`);

test('instant', (t) => {
	t.pass()
})

test.cb('2 seconds', (t) => {
	setTimeout(() => {
		t.pass()
		t.end()
	}, 2000);
})

test.cb('6 seconds', (t) => {
	setTimeout(() => {
		t.pass()
		t.end()
	}, 6000);
})

test.cb('12 seconds', (t) => {
	setTimeout(() => {
		t.pass()
		t.end()
	}, 12000);
})
```

```
$ ava test.js -T 3s
  2 passed
  2 pending

  These tests did not finish:

   6 seconds

   12 seconds

  1 exception

  ✖ Exited because no new tests completed within the last 3000ms of inactivity
```

### DIFF
--- a/api.js
+++ b/api.js
@@ -37,6 +37,7 @@ function getBlankResults() {
 	return {
 		stats: {
 			knownFailureCount: 0,
+			pendingCount: 0,
 			testCount: 0,
 			passCount: 0,
 			skipCount: 0,
--- a/api.js
+++ b/api.js
@@ -248,16 +248,28 @@ class Api extends EventEmitter {
 			});
 		});
 
+		let interrupted = false;
+
 		return this._computeForkExecArgs(files)
 			.then(argvList => {
 				execArgvList = argvList;
 			})
 			.return(files)
 			.map((file, index) => {
-				return new Promise(resolve => {
+				return new Promise((resolve, reject) => {
 					const forkArgs = execArgvList[index];
+					if (interrupted) {
+						const relFile = path.relative('.', file);
+						return reject(new AvaError(`File ${relFile} was not run.`));
+					}
+
 					const test = this._runFile(file, runStatus, forkArgs);
 					tests.push(test);
+
+					process.on('SIGINT', () => {
+						interrupted = true;
+						test.exit();
+					});
 
 					// If we're looking for matches, run every single test process in exclusive-only mode
 					const options = {
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -207,4 +207,7 @@ exports.run = () => {
 				});
 			});
 	}
+
+	// Trap SIGINT to prevent AVA quitting immediately
+	process.on('SIGINT', () => {});
 };
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -207,7 +207,4 @@ exports.run = () => {
 				});
 			});
 	}
-
-	// Trap SIGINT to prevent AVA quitting immediately
-	process.on('SIGINT', () => {});
 };
--- a/lib/main.js
+++ b/lib/main.js
@@ -27,6 +27,9 @@ Error.stackTraceLimit = Infinity;
 
 function test(props) {
 	if (isFailed) {
+		if (props.pending) {
+			adapter.send('test', props);
+		}
 		return;
 	}
 
@@ -93,6 +96,10 @@ globals.setImmediate(() => {
 	});
 
 	process.on('ava-init-exit', () => {
+		exit();
+	});
+
+	process.on('SIGINT', () => {
 		exit();
 	});
 });
--- a/lib/main.js
+++ b/lib/main.js
@@ -99,9 +99,8 @@ globals.setImmediate(() => {
 		exit();
 	});
 
-	process.on('SIGINT', () => {
-		exit();
-	});
+	// Catch SIGINTs to prevent the worker exiting ungracefully
+	process.on('SIGINT', () => {});
 });
 
 module.exports = runner.chain;
--- a/lib/reporters/mini.js
+++ b/lib/reporters/mini.js
@@ -153,8 +153,6 @@ class MiniReporter {
 			for (const test of runStatus.pendingTests) {
 				const title = test.title;
 				status += '\n   ' + colors.title(title) + '\n';
-				// TODO: Output description with link
-				// status += colors.stack(description);
 			}
 			status += '\n';
 		}
--- a/lib/reporters/mini.js
+++ b/lib/reporters/mini.js
@@ -46,6 +46,7 @@ class MiniReporter {
 	reset() {
 		this.clearInterval();
 		this.passCount = 0;
+		this.pendingCount = 0;
 		this.knownFailureCount = 0;
 		this.failCount = 0;
 		this.skipCount = 0;
@@ -75,6 +76,8 @@ class MiniReporter {
 			this.skipCount++;
 		} else if (test.error) {
 			this.failCount++;
+		} else if (test.pending) {
+			this.pendingCount++;
 		} else {
 			this.passCount++;
 			if (test.failing) {
@@ -121,6 +124,7 @@ class MiniReporter {
 	reportCounts(time) {
 		const lines = [
 			this.passCount > 0 ? '\n  ' + colors.pass(this.passCount, 'passed') : '',
+			this.pendingCount > 0 ? '\n  ' + colors.pending(this.pendingCount, 'pending') : '',
 			this.knownFailureCount > 0 ? '\n  ' + colors.error(this.knownFailureCount, plur('known failure', this.knownFailureCount)) : '',
 			this.failCount > 0 ? '\n  ' + colors.error(this.failCount, 'failed') : '',
 			this.skipCount > 0 ? '\n  ' + colors.skip(this.skipCount, 'skipped') : '',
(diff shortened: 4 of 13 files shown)


lib/runner.js
```diff
@@ -146,17 +147,22 @@ class Runner extends EventEmitter {
 	}
 
 	buildStats() {
+		this.tests.finalize();
```
novemberborn: How does this work given that `TestCollection#finalize()` does the finalizing via `setImmediate()`?

I think it'd be better to get an array of pending tests from `TestCollection` and then iterate through it, calling `addTestResult()`. I'd then modify `addTestResult()` to take a second `pending` argument and provide the correct value here and in `run()` below. That way `Test` doesn't have to be aware of `pending` at all.

`TestCollection#finalize()` can then be removed. For clarity I think `buildStats()` should be renamed to `exit()`.

lib/run-status.js
```diff
@@ -126,10 +132,11 @@ class RunStatus extends EventEmitter {
 		this.tests = flatten(this.tests);
 		this.passCount = sum(this.stats, 'passCount');
 		this.knownFailureCount = sum(this.stats, 'knownFailureCount');
+		this.pendingCount = sum(this.stats, 'pendingCount');
 		this.skipCount = sum(this.stats, 'skipCount');
 		this.todoCount = sum(this.stats, 'todoCount');
 		this.failCount = sum(this.stats, 'failCount');
-		this.remainingCount = this.testCount - this.passCount - this.failCount - this.skipCount - this.todoCount - this.knownFailureCount;
+		this.remainingCount = this.testCount - this.passCount - this.failCount - this.skipCount - this.todoCount - this.knownFailureCount - this.pendingCount;
```
novemberborn: At first blush you'd think this is now always `0`, but that wouldn't be the case with `--fail-fast` not emitting the pending tests. Perhaps a comment would help clarify that possibility.

lib/reporters/mini.js
```diff
@@ -144,6 +148,17 @@ class MiniReporter {
 
 		let status = this.reportCounts(time) + '\n';
 
+		if (this.pendingCount > 0) {
+			status += '\n  ' + colors.pending('These tests did not finish:') + '\n';
+			for (const test of runStatus.pendingTests) {
+				const title = test.title;
+				status += '\n   ' + colors.title(title) + '\n';
+				// TODO: Output description with link
+				// status += colors.stack(description);
```
novemberborn: This doesn't apply to pending tests.

lib/cli.js
```diff
@@ -207,4 +207,7 @@ exports.run = () => {
 				});
 			});
 	}
+
+	// Trap SIGINT to prevent AVA quitting immediately
+	process.on('SIGINT', () => {});
```
novemberborn: This relies on the signal still being propagated to the worker processes, right? I think we need to take a more explicit approach. For one the API needs to know not to start new workers when concurrency is being limited. Since the API is managing the worker pool it could shut down the workers, and the workers could then emit their pending tests. See https://github.com/avajs/ava/blob/e401bd151b1a81f79ccb106f034b2e18715ee125/api.js#L245:L247

(Though looking at that code I have a suspicion timeouts don't prevent new workers from being started!)

We'd need to make sure the signal isn't propagated, or otherwise is ignored in the workers.

lib/reporters/verbose.js
```diff
@@ -91,6 +91,7 @@ class VerboseReporter {
 				'  ' + colors.error(runStatus.failCount, plur('test', runStatus.failCount), 'failed') :
 				'  ' + colors.pass(runStatus.passCount, plur('test', runStatus.passCount), 'passed'),
 			runStatus.knownFailureCount > 0 ? '  ' + colors.error(runStatus.knownFailureCount, plur('known failure', runStatus.knownFailureCount)) : '',
+			runStatus.pendingCount > 0 ? '  ' + colors.todo(runStatus.pendingCount, plur('test', runStatus.pendingCount), 'pending') : '',
```
novemberborn: I'm sure it's on your list, but this still needs to print the titles of these tests.

lib/main.js
```diff
@@ -27,6 +27,9 @@ Error.stackTraceLimit = Infinity;
 
 function test(props) {
 	if (isFailed) {
+		if (props.pending) {
+			adapter.send('test', props);
+		}
```
novemberborn: This only applies to when `--fail-fast` is used. I don't think we don't need to print pending tests in that circumstance.

novemberborn:
Nice work @ace-n! Test state jumps through a lot of hoops on its way to the main process so well done following it all.

I'd like to suggest a slightly different approach though, one that is a bit more explicit and requires fewer low-level changes. See the inline comments below. Let me know what you think!

lib/test-collection.js
```diff
@@ -225,6 +226,18 @@ class TestCollection extends EventEmitter {
 		}
 		return false;
 	}
+
+	finalize() {
+		globals.setImmediate(() => {
+			Array.from(this.pendingTestInstances).forEach(test => {
```
sindresorhus: Use a `for-of` loop instead of `forEach`.

lib/runner.js
ace-n: Agreed - though this does burrow a bit into the `TestCollection`, I feel like adding such a loop at the top of `buildStats()` (or `exit()`, if we were to rename it) is a cleaner way to do this.

lib/test-collection.js
ace-n: Done.

lib/main.js
ace-n: I imagine people want to be able to identify _any_ non-passing tests with `--fail-fast`. That's doable without this PR if all tests are serial (--> any tests after the failed one are pending), but harder to reason about if test order is nondeterministic or involves multiple files executing concurrently.

Thoughts?

lib/reporters/mini.js
ace-n: Done.

lib/reporters/verbose.js
ace-n: Done. I missed this originally - good catch!

lib/run-status.js
ace-n: See follow-up [above][https://github.com/avajs/ava/pull/1596#discussion_r154272744].

lib/cli.js
ace-n: The teardown process is initiated by the `SIGINT` handler added to `lib/main.js` - which then calls the `exit()` function in that same file.

If you disable the `SIGINT` handler in `lib/main.js` while keeping this one, AVA (from what I can see through the CLI) basically ignores `SIGINT`s. If you disable this `SIGINT` handler however, `lib/main.js` doesn't have enough time to gracefully exit.

This was largely based on experimenting with the code though - so it's definitely possible that I'm initiating the wrong teardown call/event.

lib/main.js
novemberborn: > I imagine people want to be able to identify any non-passing tests with --fail-fast. 

No, all tests should stop running as soon as a single failure occurs. See #1158.

lib/runner.js
```diff
@@ -146,17 +147,28 @@ class Runner extends EventEmitter {
 	}
 
 	buildStats() {
+		for (const test of this.tests.pendingTestInstances) {
+			this.addTestResult({
+				pending: true,
+				result: test
+			});
+		}
+		this.tests.pendingTestInstances.clear(); // Pending tests must be cleared
```
novemberborn: I'm not sure why?

lib/cli.js
novemberborn: I'd like to make sure we try intercepting `SIGINT` in the main process, and then using the `Api` instance to explicitly exit all workers. We may then need to trap `SIGINT` in the workers to ensure they don't exit before they're told to do so.

lib/runner.js
novemberborn: > I feel like adding such a loop at the top of buildStats() (or exit(), if we were to rename it) is a cleaner way to do this.

Which is what you've done, right?

ace-n:
Correct.

api.js
```diff
@@ -247,17 +248,29 @@ class Api extends EventEmitter {
...
 				execArgvList = argvList;
 			})
 			.return(files)
 			.map((file, index) => {
-				return new Promise(resolve => {
+				return new Promise((resolve, reject) => {
 					const forkArgs = execArgvList[index];
+					if (interrupted) {
+						const relFile = path.relative('.', file);
+						return reject(new AvaError(`File ${relFile} was not run.`));
```
novemberborn: AVA prints this as an uncaught exception, which isn't ideal. Do you think this even needs to be printed, or can these test files be skipped silently?

api.js
```diff
@@ -247,17 +248,29 @@ class Api extends EventEmitter {
...
+					}
+
 					const test = this._runFile(file, runStatus, forkArgs);
 					tests.push(test);
 
+					process.on('SIGINT', () => {
```
novemberborn: This will trigger "memory leak" warnings when there are more than 10 test files. Each `test` is pushed onto the `tests` array though so you could use that to set up one `SIGINT` listener and exit each active test.

lib/main.js
novemberborn: > iterative development/debugging. Personally, I'd find knowing which tests are pending useful in this case - especially if I'm trying to troubleshoot why said tests are pending.

It would show tests that haven't yet completed when another test fails. I don't think that's what we're trying to solve with this PR.

novemberborn:
> Another side-question: do we want to report pending after() calls? e.g. if all your tests pass, but AVA doesn't terminate automatically - should we tell users which after() call is pending?
> 
> Perhaps this should be part of another issue altogether? I've personally ran into this in the past, but I don't know whether this feature would be useful for others.

That would be useful, yes, but it could be done in a separate issue.

novemberborn:
There's been a lot of changes recently that impact this PR. I'm also looking to refactor how test results are collected and logged. That'll make this feature easier to implement.

Closing this for now. @ace-n if you want to pick this up again, please check in first to see if I've actually managed to do that refactoring 👼 Meanwhile thank you for putting in so much effort!

Issue #1556, opened 2017-10-17
No transpile option


cdaringe:
## Description

- add support for a flag/mode that skips transpilation of all files.

ava doesn't support a "no transpilation" option.  a no transpile option would improve performance and, more importantly, yield a stable, easy debugging experience for those of us who don't need/want transpilation.  i use ava for many of it's great features, transpilation not being one of them.  transpilation+source_maps+editors frequently don't play nicely.

observe:

![vscode-ava](https://user-images.githubusercontent.com/1003261/31685899-79fb7a6c-b339-11e7-8012-b6d62919ead9.gif)
^^ _shows setting breakpoints, running ava, transpiliation happens behind the scenes, and my editor losing it's cool as it lost refs to the actual lines breakpoints were originally on, pretest_

this is already a known issue without a solution--[it even says so in the docs](https://github.com/avajs/ava/blob/master/docs/recipes/debugging-with-vscode.md#debug).  these issues "just go away" when i can use raw node, and the runtime runs just my code, untouched!

thanks for the consideration!  great work.

sindresorhus:
Yeah, I'd like to see that too. Now that Node.js 8 has native async/await, there's no longer a need for me personally to transpile. The main reason we chose to do transpilation by default initially was to get async/await.

sindresorhus:
@avajs/core Thoughts?

novemberborn:
Yes. See also https://github.com/avajs/ava/issues/709 (don't worry I'll clean up the issues).

Say we have a `"babel': false` option. Should that also imply `"powerAssert": false`? I'm leaning towards yes, since `power-assert` requires transpilation, whether via Babel or https://github.com/power-assert-js/espower-loader. But it might be an unexpected way to lose that feature. Same goes for our [throws helper](https://github.com/avajs/babel-plugin-throws-helper). Should there be an easy way to opt back in to it?

sindresorhus:
> Say we have a "babel': false option. Should that also imply "powerAssert": false?

Yes, unless we can figure out a way to run it efficiently at [runtime](https://github.com/power-assert-js/espower-loader) on a failure.

> Should there be an easy way to opt back in to it?

I don't see how you could opt into the throws helper again without having Babel, and then using `babel: false` is moot.

novemberborn:
[...]

PR #1605, opened 2017-12-05
Allow tests to be ran with dummy Babel compilation

bguerout:
Hi there,

Based on the discussions in #709 and #1556, here is a pull-request to by pass the babel compilation. This is mostly inspired by https://github.com/andywer/ava-ts.

It adds a `compileJavascriptFiles` option to switch between `CachingPrecompiler` and `NativeCachingPrecompiler`. Instead of calling `babel.transform`, this new class just returns native code by calling `toString()` function.
We tested this pull-request against our tests and we were able to debug our code without source mapping issue. Note that we don't use Babel in our project.

This pull request is not ready to be merged (no tests,...). This is mostly to share with you a way to by pass Babel compilation. Is this the right way to do it or do you have another solutions in mind?

### DIFF
--- a/api.js
+++ b/api.js
@@ -16,6 +16,7 @@ const arrify = require('arrify');
 const ms = require('ms');
 const babelConfigHelper = require('./lib/babel-config');
 const CachingPrecompiler = require('./lib/caching-precompiler');
+const NativeCachingPrecompiler = require('./lib/native-caching-precompiler');
 const RunStatus = require('./lib/run-status');
 const AvaError = require('./lib/ava-error');
 const AvaFiles = require('./lib/ava-files');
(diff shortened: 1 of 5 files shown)


novemberborn:
Hey @bguerout, thanks for the PR!

I'm leaning towards the solution proposed in https://github.com/avajs/ava/issues/1556#issuecomment-340003308 — as you've probably noticed I haven't cleaned up the existing issues to clarify the intended approach. Does this solution seem good to you?

bguerout:
Thanks for your feedback.
I'm not sure to understand how `transpileEnhancements` and `babel` options will be used.
Nevertheless feel free to close this pull request.
We can still discuss about theses options into #1556

novemberborn:
> I'm not sure to understand how transpileEnhancements and babel options will be used.

In essence I'm proposing to break up `compileJavascriptFiles` into `transpileEnhancements || babel`. So you can opt out of Babel transforms with regards to syntax support, but it's surprising if that disables other advertised AVA features. Features we may want to apply if down the line we have built-in TypeScript support, for instance. You'd have to explicitly disable those features, and with both flags set to `false` AVA wouldn't transpile test files at all.

bguerout:
Ok sounds good to me.

novemberborn:
@bguerout cool! Would you like to take that on, in lieu of this PR?

PR #1610, opened 2017-12-09
Use supertap to generate TAP output

vadimdemedes:
Uses https://github.com/vadimdemedes/supertap to generate TAP output, without changing a single line of tests and a lot of the code was extracted from AVA itself, so it's pretty much backwards compatible.


lib/reporters/tap.js
```diff
@@ -41,79 +39,42 @@ class TapReporter {
...
-		}
-
-		return output.join('\n');
+		return supertap.test(test.title, {
+			passed: !test.error,
+			index: ++this.i,
+			todo: test.todo,
+			skip: test.skip,
+			comment: test.logs,
+			error: test.error ? dumpError(test.error, true) : null
```
novemberborn: `dumpError` doesn't return an `Error` instance, which `supertap`s API says is expected. Perhaps the API documentation should be loosened, or clearer on what properties are expected on non-`Error` objects?

vadimdemedes:
> Heh, "pretty much" implies there might be some behavioral changes?

I don't think so, since I haven't modified AVA's tests and they pass 😉

lib/reporters/tap.js
vadimdemedes: Agree, will clarify that.

vadimdemedes: Done.

PR #1618, opened 2017-12-23
support @std/esm

jamestalmage:
This fixes support for `@std/esm`.

//cc: @jdalton


lib/test-worker.js
```diff
@@ -27,7 +27,13 @@ globals.options = opts;
 
 const serializeError = require('./serialize-error');
 
-(opts.require || []).forEach(x => require(x));
+(opts.require || []).forEach(x => {
+	if (/.*std[/\\]esm[/\\]index\.js$/.test(x)) {
+		require = require(x)(module); // eslint-disable-line no-global-assign
```
jamestalmage: Basically, we are making sure that we use `@std/esm`'s require function instead of the standard one when we load the test file [later in the file](https://github.com/avajs/ava/blob/8bb386a89fcd139d3da828290f90a0a8c535405a/lib/test-worker.js#L62).


I am a bit surprised that this is necessary, since I figured `@std/esm` would install a require extension hook (via `pirates` or similar). @jdalton, am I mistaken here? Do you not implement a require hook via `require.extensions['.js']`?

jdalton: Since `@std/esm` is meant to be a production dependency, as well as a dev dependency, we can't just overload the global `require.extensions` *(node folks would flip out, esp. since Lodash v5 will be using `@std/esm`)*.

If `@std/esm` is invoked through a CLI then it will hook into `require.extensions` otherwise it produces its own loader instance for folks to use. Whenever ava requires `@std/esm` through the `--require` option `@std/esm` is in CLI mode so will hook `require.extensions`. In this case the thing preventing its use is the `precompile` check which dead-ends the load chain and prevents it from getting to `@std/esm`.

test/fixture/std-esm/.esmrc
```diff
@@ -0,0 +1,4 @@
+{
+	"cjs": true,
+	"esm": "js"
+}
```
jdalton: 👆 You can shorthand to just `"cjs"` in the .esmrc file. In the next release you can use `{ "esm":"cjs" }` as a shorthand for the above too.

jamestalmage: Like `"cjs"\n` is the full contents of the file?

lib/test-worker.js
jamestalmage: `precompile` would only be short circuiting imports for tests, but those will have already used Babel to convert the `import` statements to `require` statements.

test/fixture/std-esm/.esmrc
jdalton: Ya, kinda gross but it works. You may want to wait for the other shorthand form.

lib/test-worker.js
jamestalmage: I poked around in the `@std/esm` codebase, and didn't see where you were actually hooking `require.extensions` (even in "CLI mode").

jdalton: > `precompile` would only be short circuiting imports for tests, but those will have already used Babel to convert the `import` statements to `require` statements.

That's the only issue affecting `@std/esm` *(the test files)*.

> I poked around in the `@std/esm` codebase, and didn't see where you were actually hooking `require.extensions` (even in "CLI mode").

I think `@std/esm` might be OK with just having a`babel: false` or `precompile: false` option.

jdalton: That way it avoids any package specific sniffs.

jamestalmage: OK, so this is actually only a problem when Babel is configured **not** to transpile ESM modules. Setting `babel:false` does not actually disable Babel - we still apply a few transforms to the tests for `power-assert` support, and providing better errors for common misuses of the API. 

Since we're always going to load Babel, `precompile: false` becomes a huge performance hit, as Babel just takes too long to load.

jdalton: > OK, so this is actually only a problem when Babel is configured not to transpile ESM modules.

Correct. The user wanted to just use `@std/esm` which is where I pointed to #709.

jdalton: >  we still apply a few transforms to the tests for power-assert support

`power-assert` ships with [a UMD build](https://unpkg.com/power-assert@1.4.4/build/power-assert.js) *(doesn't require Babel)*

jdalton: The [babel-plugin-throws-helper](https://github.com/avajs/babel-plugin-throws-helper) could always be optional or could be written for [acorn](https://github.com/ternjs/acorn) + [magic-string](https://github.com/Rich-Harris/magic-string) or other lighter solutions.

jamestalmage: > power-assert ships with a UMD build (no Babel required). The babel-plugin-throws-helper could always be optional or could be written for acorn+magic-string or other lighter solutions.

Interesting ideas, indeed the whole stack (esm / power-assert / throws-helper) could be pretty easily handled in acorn.

jdalton: `@std/esm` uses acorn+magic-string :yum:

lib/test-worker.js
```diff
@@ -27,7 +27,13 @@ globals.options = opts;
 
 const serializeError = require('./serialize-error');
 
-(opts.require || []).forEach(x => require(x));
+(opts.require || []).forEach(x => {
+	if (/.*std[/\\]esm[/\\]index\.js$/.test(x)) {
```
novemberborn: This should start with `[/\\]@std` I think, to better match just `@std/esm`.

jamestalmage:
@novemberborn - The only other issue would be providing a way to bypass the main thread precompilation. In that case, the package sniffing I am doing wouldn't be needed.

jamestalmage:
It also replaces "require" for when we load the tests.

billyjanitsch:
(Apologies for the notification spam.) As the original reporter of standard-things/esm#197, I just want to express appreciation for everyone working on this. Always awesome to see OSS projects come together to support interoperability for users. Thanks y'all ❤️

jdalton:
>  I just want to express appreciation for everyone working on this. Always awesome to see OSS projects come together to support interoperability for users. Thanks y'all ❤️

I was pretty floored by everyone jumping in to help out. It's such a pleasant surprise to have to rein in adjustments. Double thanks to y'all ❤️

jdalton:
Okay, `@std/esm` [v0.19.0](https://github.com/standard-things/esm/releases/tag/0.19.0) is released *(which supports that shorthand mentioned above)*.

Issue #520, opened 2016-02-07
Adding support for a dedicated configuration file


kkemple:
I'm not sure how viable this is currently due to the way pkg-conf is walking up until it finds a `package.json` file, but it would be nice to be able to set ava config in an `.avarc` file in the root of the project.

I'm not a huge fan of keeping config in my package.json or of overly verbose cli commands in `scripts` config. This is also pretty common practice with other tooling frameworks so I don't think it's too crazy of an idea.

I think an initial approach could be to check the current working directory for a `.avarc` file and if not found fall back to `pkgConf.sync`.

I would be more than willing to put together a PR if this seems like something you find appealing.

sindresorhus:
We're intentionally only using package.json for simplicity and consistency.

> I'm not a huge fan of keeping config in my package.json

Why? *We* are not big fans of having an endless amount of meta files in our repos. It's IMHO nice to have one place for all config. This is becoming more and more common, with both AVA, ESLint, XO, nyc, standard, etc, tools doing it.

> This is also pretty common practice with other tooling frameworks

Common practice doesn't necessarily mean it's a good one.

kkemple:
okay, thanks for the response

mattecapu:
> Why? We are not big fans of having an endless amount of meta files in our repos

**I am**. I prefer separation over concentration, and it's handy to just ctrl-c ctrl-v a dotfile. That said, I think no one is really wrong or right and **is just a matter of taste**.

From what I can see, to add support for a dotfile there are just a handful of lines to edit in [cli.js](https://github.com/sindresorhus/ava/blob/cac933e11b8b2d503eef5b8d614b18cd4f5666ed/cli.js#L38), [profile.js](https://github.com/sindresorhus/ava/blob/a03f8264edbd99d9c6a86827106ad40dfdc42b10/profile.js#L23) and [test/cli.js](https://github.com/sindresorhus/ava/blob/c928cb59b6d2c3ed9aa3df05bc922b885ca5e985/test/cli.js#L100).

My naive implementation would just grab the config object from `.avarc` instead of `package.json`. A one-liner. I'd be really happy to make a PR for this (also because I love this project and I'd like to contribute). The only thing to decide is precedence, and I'll leave `package.json` first for legacy reasons.

Are there any complications I'm overlooking?

sindresorhus:
@mattecapu Sorry, we're not interested in having multiple ways to define config.

calebmer:
[...]

PR #1627, opened 2018-01-04
Adds `.avarc` meta file support.

jtag05:
Simple implementation resolving #520.

### DIFF
--- a/profile.js
+++ b/profile.js
@@ -9,6 +9,7 @@ const EventEmitter = require('events');
 const meow = require('meow');
 const Promise = require('bluebird');
 const pkgConf = require('pkg-conf');
+const fs = require('pkg-conf');
 const findCacheDir = require('find-cache-dir');
 const uniqueTempDir = require('unique-temp-dir');
 const arrify = require('arrify');
@@ -35,11 +36,10 @@ globals.clearTimeout = clearTimeout.bind(null);
 
 Promise.longStackTraces();
 
-const conf = pkgConf.sync('ava', {
-	defaults: {
-		babel: 'default'
-	}
-});
+const conf = Object.assign(
+	{ babel: 'default' },
+	(fs.exists('.avarc') ? JSON.parse(fs.readFileSync('.avarc')) : pkgConf.sync('ava'))
+);
 
 // Define a minimal set of options from the main CLI
 const cli = meow(`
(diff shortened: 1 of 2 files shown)


novemberborn:
Hey @jtag05, thanks for the PR. #520 has already been closed as wontfix. Clearly this is a contentious issue but I'd like to see an argument in favor of placing configuration outside of `package.json` that goes beyond "it's what I'm used to from other tools", which I think is where we left #520.

Anyway, #520 is the place for that discussion, though of course the "it's what I'm used to" point has already been made.

pho3nixf1re:
@novemberborn there is a perfectly valid argument in #520 that was recently posted. The package.json file has become a dumping ground for configuration. In large projects it makes it huge and unwieldy. Digging through a giant JSON file is not a pleasant experience and is reminiscent of the Java world's XML configs. My team prefers the dotfiles as it neatly separates each tool's configuration that is easily explorable. It is not just because "we are used to it". Every other tool we use manages to support this trivial feature that we value but AVA. It seems absurd that even in the face of community outcry and the presence of a perfectly fine PR that the maintainers here ignore the need.

jtag05:
I'd like to see an argument against implementing 7 lines of code that has no performance issues. I understand that preferences are exactly that, but rather than taking the stance of "why should we?", as a utility should you not be approaching this from the perspective of "Why not?" as a means of maximizing adoption?

I think it's fair to argue that whichever route a developer chooses (meta file or package.json) all configuration should be stored the same way. As I have previously stated in the #520 thread, I was a supporter of `package.json` based configutation, but I've since run into issues on projects where a `package.json` file get's seriously bloated. Take for instance if you use a custom `eslintConfig` and store it in `package.json`, that can easily run 300 lines.

profile.js
```diff
@@ -9,6 +9,7 @@ const EventEmitter = require('events');
 const meow = require('meow');
 const Promise = require('bluebird');
 const pkgConf = require('pkg-conf');
+const fs = require('pkg-conf');
```
tacooper: I think you meant `require('fs')` here.

psirenny:
To add to what @pho3nixf1re said, I prefer dotfiles because they can easily be added to a .npmignore file. Consumers of my packages don't care about various config choices and don't need to download bloated package.json file.

novemberborn:
@psirenny the latest beta release (tagged `next` on npm) has support for `ava.config.js` files. Not a dotfile but you can keep config out of `package.json` and away from npm.

PR #1629, opened 2018-01-07
Fix: add types for t.snapshot(content, options)

ajafff:
While writing my tests in TypeScript I noticed the missing overload signature for `t.snapshot(content, options [, message])`.
This add the missing type declaration.


novemberborn:
Lovely, thanks @ajafff!

PR #1634, opened 2018-01-11
Debug serially in the "Debugging tests with VS Code" recipe

troysandal:
Debug serially to avoid confusing execution paths in async code, e.g. swapping from one test's execution context to the next.


novemberborn:
Thanks for your PR @troysandal! Those execution paths may be the cause of test failures, though. Rather than suggesting this as the default invocation, perhaps you could repeat the `configurations` section with the `--serial` flag included, and a brief explanation of why this might be useful?

troysandal:
@novemberborn great point Mark, I added a new section under Debug that addresses this, thank you.

novemberborn:
Thanks @troysandal. I've made a few more changes, please let me know if you're happy with them.

troysandal:
Changes look great @novemberborn and @sindresorhus thank you.

PR #1638, opened 2018-01-15
Use absolute map paths for absolute cache paths

### LINKED ISSUE #817: Cache source-map urls should use absolute paths.
I run WIndows tests in Parallels, where my OSX homedir is mapped to the `Z:` drive. Many of our tests use the `--no-cache` option which causes our normal caching behavior to be redirected to a temp directory so there are no collisions between runs. Unfortunately, there's no way to represent a relative path between the `Z:` and `C:` drive.

Once https://github.com/thlorenz/convert-source-map/pull/41 is merged, we should be able to just attach the sourceMapUrl comments with absolute paths.

kugtong33:
Fixes #817 

As far as `CachingPrecompiler` is used, `options.path` would always be assigned with absolute paths as I observed on `api.js`, `profile.js`, `fork.js`, `hooks.js`. If ever the cache directory path is not absolute, the relative map path would be added instead.

Also updated the `convert-source-map` version to `^1.5.1` as suggested.

### DIFF
--- a/lib/caching-precompiler.js
+++ b/lib/caching-precompiler.js
@@ -76,12 +76,7 @@ class CachingPrecompiler {
 
 		// Append source map comment to transformed code
 		// So that other libraries (like nyc) can find the source map
-		let relativeMapPath;
-		if (!path.isAbsolute(mapPath)) {
-			const dirPath = path.dirname(filePath);
-			relativeMapPath = path.relative(dirPath, mapPath);
-		}
-		const comment = convertSourceMap.generateMapFileComment(relativeMapPath || mapPath);
+		const comment = convertSourceMap.generateMapFileComment(mapPath);
 
 		return `${result.code}\n${comment}`;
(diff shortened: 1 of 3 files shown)


kugtong33:
I'll wait for further feedback, before I'll write tests for relative cache directory paths :)

novemberborn:
Thanks for the doing the research @kugtong33!

> If ever the cache directory path is not absolute, the relative map path would be added instead.

Best I can tell the cache dir is also absolute. We should thus be able to use `mapPath` as is, without making it relative.

kugtong33:
@novemberborn , not sure about the failed appveyor node v4 build

novemberborn:
Thanks @kugtong33!

kugtong33:
@novemberborn , yes I deleted it and then hit `npm install` again, hmm should have just updated the version and let npm do the update itself to the lock file

PR #1643, opened 2018-01-20
Update code-excerpt to ^2.1.1

### LINKED ISSUE #1636: Windows Line Feed Code Excerpt Display Bug
## Description

There seems to be a small display bug when printing the code-excerpt for an assertion failure into the console for CRLF encoded files due to windows **\r\n** line feed.
The code-excerpt node module returns lines with **\r** at the end for CRLF encoded files.
ava/lib/code-excerpt.js line 29: 
```js
const excerpt = codeExcerpt(contents, line, {around: 1});
//[ { line: 14, value: '  1 + 1 + 1 + 1;\r' },
//  { line: 15, value: '  t.is(1,2)\r' },
//  { line: 16, value: '  1;\r' } ]
```
These are then joined with **\n** and sent to the equal-length module which re-splits the lines on **\n** and adds trailing spaces to equalize the length of each line:
```js
[ '  1 + 1 + 1 + 1;\r',
  '  t.is(1,2)\r     ',
  '  1;\r            ' ]
```
Due to the **\r**, the trailing spaces override the preceding string, resulting in blanked out characters:
image
When it should show:
image

### Ava Source
ava/lib/code-excerpt.js lines 29 to 56:
```js
const excerpt = codeExcerpt(contents, line, {around: 1});
if (!excerpt) {
	return null;
}

const lines = excerpt.map(item => ({
	line: item.line,
	value: truncate(item.value, maxWidth - String(line).length - 5)
}));

const joinedLines = lines.map(line => line.value).join('\n');
const extendedLines = equalLength(joinedLines).split('\n');

return lines
	.map((item, index) => ({
		line: item.line,
		value: extendedLines[index]
	}))
	.map(item => {
		const isErrorSource = item.line === line;

		const lineNumber = formatLineNumber(item.line, line) + ':';
		const coloredLineNumber = isErrorSource ? lineNumber : chalk.grey(lineNumber);
		const result = ` ${coloredLineNumber} ${item.value}`;

		return isErrorSource ? chalk.bgRed(result) : result;
	})
	.join('\n');
```

### Equal Length Source
equal-length/index.js line 8:
```js
const lines = input.split('\n');
```

### Code Excerpt Source
code-excerpt/index.js line 27:
```js
source = tabsToSpaces(source).split('\n');
```
### 1 Avoid Equalizing Length
Remove lines 43 to 46 of ava/lib/code-excerpt.js and remove call to equal-length library:

```js
const excerpt = codeExcerpt(contents, line, {around: 1});
if (!excerpt) {
	return null;
}

const lines = excerpt.map(item => ({
	line: item.line,
	value: truncate(item.value, maxWidth - String(line).length - 5)
}));

return lines
	.map(item => {
		const isErrorSource = item.line === line;

		const lineNumber = formatLineNumber(item.line, line) + ':';
		const coloredLineNumber = isErrorSource ? lineNumber : chalk.grey(lineNumber);
		const result = ` ${coloredLineNumber} ${item.value}`;

		return isErrorSource ? chalk.bgRed(result) : result;
	})
	.join('\n');
```

This fixes the display bug, as after the final join each **\r** is properly matched with a **\n** to create a valid newline. I'm not too sure why the lines need to be equalized personally, so I think this makes things a little simpler.

### 2 Tweak Code-Excerpt Library - (my preference if both can't be implemented)
Replacing code-excerpt/index.js line 27:
```js
source = tabsToSpaces(source).split(/\r?\n/);
```
fixes the problem as the data being passed into the line equalizer is free of any **\r**.
https://stackoverflow.com/questions/21895233/how-in-node-to-split-string-by-newline-n

## Finally
Thank you in advance for any work you do on this as it took me ages to puzzle it out, hope it is helpful :).

kugtong33:
Fixes #1636 

Updated code-excerpt dependency to 2.1.1 , this version contains the fix for the abnormal display on CRLF line breaks

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1282,9 +1282,9 @@
       }
     },
     "code-excerpt": {
-      "version": "2.1.0",
-      "resolved": "https://registry.npmjs.org/code-excerpt/-/code-excerpt-2.1.0.tgz",
-      "integrity": "sha1-XcwIHoj0p+O1VOnjXX7yMtR/gUc=",
+      "version": "2.1.1",
+      "resolved": "https://registry.npmjs.org/code-excerpt/-/code-excerpt-2.1.1.tgz",
+      "integrity": "sha512-tJLhH3EpFm/1x7heIW0hemXJTUU5EWl2V0EIX558jp05Mt1U6DVryCgkp3l37cxqs+DNbNgxG43SkwJXpQ14Jw==",
       "requires": {
         "convert-to-spaces": "1.0.2"
       }
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
 		"cli-spinners": "^1.0.0",
 		"cli-truncate": "^1.0.0",
 		"co-with-promise": "^4.6.0",
-		"code-excerpt": "^2.1.0",
+		"code-excerpt": "^2.1.1",
 		"common-path-prefix": "^1.0.0",
 		"concordance": "^3.0.0",
 		"convert-source-map": "^1.5.1",


sindresorhus:
Thanks, but this is not needed. Semver handles this for us. Read more here: https://nodesource.com/blog/semver-tilde-and-caret You just need to reinstall AVA. The package-lock.json doesn't apply when you install AVA from npm.

novemberborn:
@sindresorhus this is useful, actually. I don't think we can rely on npm updating transitive dependencies if the currently installed version is still within the acceptable range. It might also pull an existing version from a cache. Explicitly changing the version range ensures fixes are included in the next release.

sindresorhus:
@novemberborn What's the point of semver then, if we have to manually bump for each patch release of dependencies?

sindresorhus:
And what about all the other dependencies with semver ranges that are not forcing the latest patch release? We're probably missing out on a lot of fixes in those, by this logic.

I'm not again this PR, I'm just genuinely curious about how we should handle things :)

novemberborn:
I think the trick is that npm might dedupe dependencies according to the allowed SemVer ranges. This means that it's not necessarily guaranteed you'll receive the *latest* version. You'll receive the most widely compatible instead.

I've also seen this in application projects with package locks, where the best way to upgrade was to remove and then reinstall a package. Perhaps it's better with newer npm versions.

For us maintainers then, if there's a dependency upgrade that comes about due to a bug report, we should upgrade the minimum version so there's no circumstance under which we are linked to the buggy version.

Similarly our use of package locks helps ensure we're not accidentally coupled to a newer dependency feature whilst still allowing a previous version to be used.

Issue #1332, opened 2017-04-02
Ability to run test files in the same process


sindresorhus:
Split from https://github.com/avajs/ava/issues/1322.

> … add an option to not spawn AVA test files as child-processes. Pretty much what `ava/profile.js` does now. I also need this for testing Electron apps, as their tester needs to inject some globals. Having such option would also fix our React/etc performance issues.

It could be a global option that when turned on would run the test files in the same process in a new script context using the `vm` module instead of a child process.

jeffijoe:
This would still run each test in parallel, right?

novemberborn:
> This would still run each test in parallel, right?

Yes.

ORESoftware:
@novemberborn yes parallel, but not in separate processes, because `require('vm')` doesn't spawn a separate process (...or does it?)

sindresorhus:
> because `require('vm')` doesn't spawn a separate process (...or does it?)

It does not. That's the whole point of this issue, to not spawn.

tniessen:
> This would still run each test in parallel, right?

Just FYI, the `vm` module has nothing to do with concurrency, and it is not comparable to executing tests in separate processes. There is neither strict isolation nor will any code be executed in parallel, apart from the usual event loop.

novemberborn:
Prompted by #1631 it might be interesting to still start workers, but reuse them for different test files. We'd use `vm` for sandboxing. Would be good to check Jest's sandboxing approach as well.

lo1tuma:
I’ve experimented a little bit with implementing sandboxing in ava via the `vm` module. This seems to be quite tricky because per default you don’t have a node-environment in the context of the vm, which means you don’t have e.g. `require`.
I also tried [`NodeVM` from `vm2` ](https://github.com/patriksimek/vm2#nodevm) which unfortunately has [a bug](https://github.com/patriksimek/vm2/issues/62) which leads to throwing 
 a `TypeError` for every assertion (e.g. `t.is()`). AFAIK `vm2` injects all the commonjs globals and the function wrapper manually to the sandbox, so I would assume that it doesn’t works nicely with ES modules.

I’ve also struggled a little bit with the way how the runner instance is loaded in each test file (i.e. `import test from 'ava'`). The [current approach](https://github.com/avajs/ava/blob/f4b2d196e813cf3549e19703e5a79ebc988fb8ca/lib/worker/main.js#L20) only works because ava uses a separate process for every test file.

link89:
[...]

Issue #1635, opened 2018-01-11
Make t.log() behave like console.log()


edbrannin:
## Description

When using `--tap` output, trying to log an object results in an error.  Worse, the stack trace points to ava internals, not to my test code.

If `t.log` is given an object:

* Minimum expected behavior: Throw immediately
* Preferred expected behavior: Pass through `JSON.stringify(arg, null, 2)`

I do a lot of `t.log(JSON.stringify(resp.body, null, 2));` and it would be great to just say `t.log(resp.body);` instead.

### Test Source

```js
import { serial as test } from 'ava';

test('Log an object', (t) => {
  const myObject = { a: 1 };
  t.log(myObject);
  t.is(true, true);
});
```

### Error Message & Stack Trace

```
^^/d/o/b/ava >>> ava --tap test.js 

/Users/redacted/dev/oss/bug-reports/ava/node_modules/indent-string/index.js:9
		throw new TypeError(`Expected \`input\` to be a \`string\`, got \`${typeof str}\``);
		^

TypeError: Expected `input` to be a `string`, got `object`
    at module.exports (/Users/redacted/dev/oss/bug-reports/ava/node_modules/indent-string/index.js:9:9)
    at test.logs.forEach.log (/Users/redacted/dev/oss/bug-reports/ava/node_modules/ava/lib/reporters/tap.js:64:23)
    at Array.forEach (<anonymous>)
    at appendLogs (/Users/redacted/dev/oss/bug-reports/ava/node_modules/ava/lib/reporters/tap.js:63:15)
    at TapReporter.test (/Users/redacted/dev/oss/bug-reports/ava/node_modules/ava/lib/reporters/tap.js:83:4)
    at Logger.test (/Users/redacted/dev/oss/bug-reports/ava/node_modules/ava/lib/logger.js:27:28)
    at emitTwo (events.js:126:13)
    at RunStatus.emit (events.js:214:7)
    at RunStatus.handleTest (/Users/redacted/dev/oss/bug-reports/ava/node_modules/ava/lib/run-status.js:105:8)
    at emitOne (events.js:121:20)
    at ChildProcess.emit (events.js:211:7)
    at ChildProcess.ps.on.event (/Users/redacted/dev/oss/bug-reports/ava/node_modules/ava/lib/fork.js:79:7)
    at emitTwo (events.js:126:13)
    at ChildProcess.emit (events.js:214:7)
    at emit (internal/child_process.js:772:12)
[...]

PR #1645, opened 2018-01-21
[WIP] Run tests in a single process

vadimdemedes:
Fixes https://github.com/avajs/ava/issues/1332.

Note, that there's no sandboxing like in Jest. This PR runs tests in the same AVA process, similar to Mocha. Feel free to try it out in your own projects, propose better solutions/code and generally tell us what you think.

**TODO:**

- [ ] Fix existing tests.
- [ ] Add new tests.
- [ ] Discuss how to ship it (name of the flag / make it default / experimental flag / etc).
- [ ] Capture `console.log()` and send its args to `lib/fork` for proper output, like in fork mode. I'd rather support https://github.com/avajs/ava/issues/1635 instead.
- [ ] Detect which dependencies belong to other test files in dependency tracking (`lib/process-adapter`).
- [ ] Fix `@std/esm` support (doesn't work, because `lib/main` doesn't have access to `module.parent.parent`, which is `undefined`).


novemberborn:
What do you think about still running the tests in a separate worker process, but *reusing* that process?

vadimdemedes:
Interesting idea, I think it would help with correct console.log() handling.

novemberborn:
It also protects the main process from any shenanigans the tests may be up to. We'd still get process concurrency.

sindresorhus:
> What do you think about still running the tests in a separate worker process, but reusing that process?

I think reusing could be useful for multi-process too, but this PR should focus on only the same process. The use-cases are things that need to be run in the same process as the AVA CLI, for example, debugging or Electron. So I'd rather have multi-process mode reuse processes and let this one be without any process forking. If users want reuse of a single process, they can just set `concurrency` to 1 in multi-process mode.

novemberborn:
> The use-cases are things that need to be run in the same process as the AVA CLI, for example, debugging or Electron.

Interesting, we can start with top-level process then.

I think rather than mimicking the rather low-level `fork()`, we should abstract the "pool" and how it communicates with the API and `run-status`.

novemberborn:
> I think rather than mimicking the rather low-level fork(), we should abstract the "pool" and how it communicates with the API and run-status.

I've been doing a fair bit of refactoring in this area, and have some ideas on how to make `run-status` better too. That should make this feature easier to implement.

vadimdemedes:
> I think rather than mimicking the rather low-level fork(), we should abstract the "pool" and how it communicates with the API and run-status.

That doesn't sound like an easier way to implement this right now to me :) This PR isn't that complex, is it? What improvements specifically do you have in mind?

novemberborn:
> What improvements specifically do you have in mind?

Getting rid of the various events that are emitted and forwarded and rewritten and reemitted, as well as the stats counting and test tracking we do all over the place.

vadimdemedes:
But this PR doesn't change events, it just switches `process.on` and `process.send` with adapters. Refactoring that you have in mind could be done afterwards with the same outcome imo.

novemberborn:
I'd really like this to be a first class implementation, rather than a mimicking of how `fork` and `test-worker` interact. Cause I've been looking a lot at that code lately and I still can't load it all into my head.

I have been doing and will continue to do refactoring to clear up the existing code, and hopefully that makes it relatively straight-forward to do that first class implementation. It's definitely in the back of my mind.

vadimdemedes:
> I'd really like this to be a first class implementation, rather than a mimicking of how fork and test-worker interact. Cause I've been looking a lot at that code lately and I still can't load it all into my head.

Definitely agree with this. Ok then, we can put this PR on hold for now, I'm super busy lately anyway.

novemberborn:
Hey @vadimdemedes, this is quite outdated by now. Going to close if you don't mind 😄

PR #1647, opened 2018-01-21
Allow custom Node executable via environment variable for children processes

CodeOtter:
# Story

As a developer, I would like to provide my own path to a Node executable so the test children can be ran using that custom Node executable.

# Description

I hate Babel (the line mangling and source mapping, the requirement to compile code in advance for NPM, the implied requirements to compile test code, the unreadable stack traces in production, its malware-like capacity to destroy CPUs on older machines, the language security disaster that awaits us all for using it, the rewarding of terrible mentalities centered around turning JavaScript into a crippled, inferior mutant of C# and Java when, really, just learn Go and call it a day) and my project intentionally goes out of its way to not use it.

I do use Flow because, unlike most use cases of Babel, it's using transpilers to solve an actual problem and not minor annoyances.  I like to keep my developer exposure to Babel as small as possible, which means I need Ava's children processes to run via `flow-node` and not native `node`.

This PR lets you specify an environment variable (`NODE_EXEC`) to contain the path to a custom Node executable for children processes.  This is probably not the best approach, and even if the PR gets rejected, the spirit of the PR should, at the very least, be considered and solved by another form.

### DIFF
--- a/bench/run.js
+++ b/bench/run.js
@@ -13,7 +13,7 @@ function runTests(_args) {
 		const args = [cliPath].concat(arrify(_args));
 		const start = Date.now();
 
-		childProcess.execFile(process.execPath, args, {
+		childProcess.execFile(process.env.NODE_EXEC || process.execPath, args, {
 			cwd: __dirname,
 			maxBuffer: 100000 * 200
 		}, (err, stdout, stderr) => {
--- a/test/cli.js
+++ b/test/cli.js
@@ -32,7 +32,7 @@ function execCli(args, opts, cb) {
 	let stderr;
 
 	const processPromise = new Promise(resolve => {
-		child = childProcess.spawn(process.execPath, [cliPath].concat(args), {
+		child = childProcess.spawn(process.env.NODE_EXEC || process.execPath, [cliPath].concat(args), {
 			cwd: dirname,
 			env,
 			stdio: [null, 'pipe', 'pipe']


novemberborn:
Would [`flow-node`'s `require-hook`](https://www.npmjs.com/package/flow-node#use-the-require-hook) work for your use case?

CodeOtter:
Yes, I am currently solving this problem with a dedicated `test/index.js` file that Ava requires, which contains that require hook:

`test/index.js`
```
require('flow-remove-types/register')({
  all: true,
  pretty: true,
  sourcemaps: true
})
```

`package.json`
```
{
  "ava": {
    "files": [
      "test/**/*.{js,jsx}",
      "!test/index.js"
    ],
    "source": [
      "src/**/*.{js,jsx}"
    ],
    "require": [
      "./test"
    ]
  }
}
```

Given JavaScript's established and, apparently, never-ending proclivity to invent new DSLs/dialects/parsers/AST opinions-per-project, this PR is designed to enhance Ava's future proofing, as far as that interest is relevant. :P

novemberborn:
I much prefer the `register` approach, actually. One concern I have with your suggested approach is it may break AVA's assumption that the same Node.js version is used in the main and worker processes. 

Given that `flow-node` really doesn't do much other than loading the `register` module I don't think it adds enough value. Thank you for taking the effort to make the PR though!

Issue #483, opened 2016-01-29
Don't set global stack trace limit to be infinite


jamestalmage:
Infinity depth stack traces hardly seem necessary, but [that is what we are currently doing](https://github.com/sindresorhus/ava/blob/761c2e349b306769e9c8f090a2cef6cb0c933664/index.js#L32).
#482 fixes that for `t.plan()`, which I think is important, since `t.plan()` generates a stack trace whether there is a test failure or not.

My remaining question is whether or not we should shorten them globally. I am not overly concerned about the performance impact of failing assertions, more so other libraries (like bluebird longStackTrace support) that might be creating overly long stacks and slowing things down. 

Some investigation is needed here. If an `Infinity` value does not hurt the speed of a passing test suite, then let's leave it there. Otherwise we should bikeshed on how long it should be.

novemberborn:
> My remaining question is whether or not we should shorten them globally. I am not overly concerned about the performance impact of failing assertions, more so other libraries (like bluebird longStackTrace support) that might be creating overly long stacks and slowing things down.

Changing the stack trace limit may also impact code-under-test. Instead we should set it to `Infinity` when recording the assertion failure, and then reset it afterwards.

novemberborn:
We set the limit here: https://github.com/avajs/ava/blob/dd9e8b2effe541f9f232ee622452343dac5895dd/lib/main.js#L25

Instead we should modify the `AssertionError` constructor to capture an infinite stack trace: https://github.com/avajs/ava/blob/dd9e8b2effe541f9f232ee622452343dac5895dd/lib/assert.js#L31:L54

We need to do the same in the `getStack()` method: https://github.com/avajs/ava/blob/dd9e8b2effe541f9f232ee622452343dac5895dd/lib/assert.js#L57:L61

We have some existing code for this already: https://github.com/avajs/ava/blob/dd9e8b2effe541f9f232ee622452343dac5895dd/lib/test.js#L24:L31

kugtong33:
@novemberborn , I think I'll stick with easier issues for now, can I get this? :)

PR #1648, opened 2018-01-22
Dont' set `Error.stackTraceLimit` to Infinity globally

okyantoro:
Refs: https://github.com/avajs/ava/issues/483

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -49,14 +49,22 @@ class AssertionError extends Error {
 
 		if (opts.stack) {
 			this.stack = opts.stack;
+		} else {
+			const limitBefore = Error.stackTraceLimit;
+			Error.stackTraceLimit = Infinity;
+			Error.captureStackTrace(this);
+			Error.stackTraceLimit = limitBefore;
 		}
 	}
 }
 exports.AssertionError = AssertionError;
 
 function getStack() {
+	const limitBefore = Error.stackTraceLimit;
+	Error.stackTraceLimit = Infinity;
 	const obj = {};
 	Error.captureStackTrace(obj, getStack);
+	Error.stackTraceLimit = limitBefore;
 	return obj.stack;
 }
 
--- a/test/cli.js
+++ b/test/cli.js
@@ -83,6 +83,16 @@ test('enabling long stack traces will provide detailed debug information', t => 
 	});
 });
 
+test('`AssertionError` should capture infinity stack trace', t => {
+	execCli('fixture/infinity-stack-trace.js', (err, stdout, stderr) => {
+		t.ok(err);
+		t.match(stderr, /c \(.+?infinity-stack-trace\.js:6:20\)/);
+		t.match(stderr, /b \(.+?infinity-stack-trace\.js:7:18\)/);
+		t.match(stderr, /a \(.+?infinity-stack-trace\.js:8:18\)/);
+		t.end();
+	});
+});
+
 test('timeout', t => {
 	execCli(['fixture/long-running.js', '-T', '1s'], (err, stdout, stderr) => {
 		t.ok(err);
--- a/test/fixture/infinity-stack-trace.js
+++ b/test/fixture/infinity-stack-trace.js
@@ -1,0 +1,11 @@
+import test from '../../';
+
+Error.stackTraceLimit = 1;
+
+test(t => {
+	const c = () => t.fail();
+	const b = () => c();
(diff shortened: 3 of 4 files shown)


lib/assert.js
```diff
@@ -38,6 +38,7 @@ class AssertionError extends Error {
 		this.improperUsage = opts.improperUsage || false;
 		this.operator = opts.operator;
 		this.values = opts.values || [];
+		this.stackTraceLimit = Infinity;
```
novemberborn: Is this sufficient? I don't see a reference to `stackTraceLimit` as an *instance* property in https://github.com/v8/v8/wiki/Stack-Trace-API.

Instead I suspect we need to manually call `Error.captureStackTrace(this)` below, if `opts.stack` is falsy.

okyantoro: You are right. Thank you for pointing this out.
I made change to the code.
Thanks

novemberborn:
Thanks @okyantoro. Would you like to try writing an integration test for this?

> It'd be good to have an integration test where the default stack limit is set to say 1, and then cause an assertion failure that should have a deeper stack available.

okyantoro:
sure @novemberborn 
I'll try to write the test but I need guide where to start to add the test
Thank you

novemberborn:
@okyantoro These kinds of tests tend to end up in `test/cli.js`. See this example: https://github.com/avajs/ava/blob/bcb77fc46dbb77ddb14ac8eda8f2cc7e856c2416/test/cli.js#L78:L84

The idea is to write a fixture that is an an AVA test file. In your case it should contain an assertion that fails. Then in `cli.js` you can match `stderr` for various lines from the expected stack trace.

okyantoro:
Thank you @novemberborn 
I have added a test. Could you review this and give some some suggestions about what I do wrong and what should it be with this.
Thank you

test/fixture/infinity-stack-trace.js
```diff
@@ -0,0 +1,14 @@
+import test from '../../';
+import {AssertionError} from '../../lib/assert';
+
+test(t => {
+	const limitBefore = Error.stackTraceLimit;
+	Error.stackTraceLimit = 1;
```
novemberborn: Since this test is run in its own process you could set the limit at the top level, before the `test()` call.

test/fixture/infinity-stack-trace.js
```diff
@@ -0,0 +1,14 @@
...
+	try {
+		throw new AssertionError({});
+	} catch (err) {
+		console.error(err.stack);
```
novemberborn: Rather than instantiating the error, call say `t.fail()` from inside some nested functions:

```js
const a = () => b();
const b = () => c();
const c = () => t.fail();
```

AVA should print the full stack trace which you can then assert against in `test/cli.js`.

okyantoro: updated

test/cli.js
```diff
@@ -83,6 +83,15 @@ test('enabling long stack traces will provide detailed debug information', t =>
 	});
 });
 
+test('`AssertionError` should capture infinity stack trace', t => {
+	execCli('fixture/infinity-stack-trace.js', (err, stdout, stderr) => {
+		t.ok(err);
+		t.match(stderr, /const c/);
+		t.match(stderr, /const b/);
```
novemberborn: If I run the fixture locally (`node ./cli.js test/fixture/infinity-stack-trace.js`) I get:

```
  1 failed

  [anonymous]

  /Users/mark/GitHub/avajs/ava/test/fixture/infinity-stack-trace.js:6

   5: test(t => {
   6:   const c = () => t.fail();
   7:   const b = () => c();

  Test failed via `t.fail()`

  c (test/fixture/infinity-stack-trace.js:6:20)
  b (test/fixture/infinity-stack-trace.js:7:18)
  a (test/fixture/infinity-stack-trace.js:8:18)
  Test.t [as fn] (test/fixture/infinity-stack-trace.js:10:2)
```

Right now you're matching the code snippet but not the stack trace. Try `t.match(stderr, /c \(.+?infinity-stack-trace\.js:6/)` and so forth for `b` and `a`. (Note that I haven't actually tried this regular expression, apologies in advance if it's wrong! 😄 )

okyantoro:
Thank you @novemberborn 
I learn a lot from here.

novemberborn:
Thanks @okyantoro 🎉

Issue #1371, opened 2017-04-27
Unhandled rejection error on "t.throws()" test of trivial async function


sdd:
## Description

Unhandled rejection error on test of async function that throws

### Test Source

```js
import test from 'ava';

const f = async () => {
    throw new Error();
};

test(async t => {

    const error = await t.throws(async () => {
        return await f();
    });

    t.is(error.message, '');
});
```

### Error Message & Stack Trace

```
1 failed [08:28:15]
  1 rejection

  test › [anonymous]
  /Users/scotty/workspace/fetch-nice/test/test.js:9

   8:
   9:     const error = await t.throws(async () => {
   10:         return await f();

  Test.<anonymous> (test/test.js:9:27)
  step (test/test.js:9:191)
  Test.<anonymous> (test/test.js:9:99)
  Test.__dirname [as fn] (test/test.js:7:1)

  Unhandled Rejection
  Error
    test/test.js:4:11
    Generator.next (<anonymous>)
    step (test/test.js:9:191)
    test/test.js:9:437
    test/test.js:9:99
    f (test/test.js:3:7)
    test/test.js:10:22
    Generator.next (<anonymous>)
    step (test/test.js:9:191)
    test/test.js:9:437
    test/test.js:9:99
    _tryBlock (node_modules/core-assert/index.js:311:5)
    _throws (node_modules/core-assert/index.js:330:12)
    Function.assert.throws (node_modules/core-assert/index.js:360:3)
    Test.<anonymous> (test/test.js:9:27)
    Generator.next (<anonymous>)
    step (test/test.js:9:191)
    test/test.js:9:437
    Test.<anonymous> (test/test.js:9:99)
    Test.__dirname [as fn] (test/test.js:7:1)
```

### Config

```json
{
  "ava": {
  "require": [
      "babel-register",
      "babel-polyfill"
    ]
  }
}
```

### Command-Line Arguments

```
ava -w
```

## Environment

Node.js v7.6.0
darwin 15.6.0
ava 0.19.1
npm 4.1.2

sindresorhus:
Simpler test case:

```js
import test from 'ava';

test(async t => {
	await t.throws(async () => {
		throw new Error();
	});
});
```

novemberborn:
This is correct behavior: the function doesn't throw, so the assertion fails. AVA ignores the return value, so the promise rejection ends up being unhandled and is reported separately.

[...]

PR #1650, opened 2018-01-23
Make `t.throws()` accept async function as parameter

okyantoro:
When an async callback is given to ```t.throws()```, it will execute the function and put the returned ```Promise``` to ```promise``` variable. The rest is same as processing ```Promise```.
This commit introduce new private function ```isAsync()``` inside ```asserts.wrapAssertions()```

Refs: https://github.com/avajs/ava/issues/1371

### DIFF
--- a/index.js.flow
+++ b/index.js.flow
@@ -55,13 +55,13 @@ type AssertContext = {
 	deepEqual<U>(value: U, expected: U, message?: string): void;
 	// Assert that value is not deep equal to expected.
 	notDeepEqual<U>(value: U, expected: U, message?: string): void;
-	// Assert that function throws an error or promise rejects.
+	// Assert that the promise rejects, or the function throws or returns a rejected promise.
 	// @param error Can be a constructor, regex, error message or validation function.
 	throws: {
 		(value: PromiseLike<mixed>, error?: ErrorValidator, message?: string): Promise<any>;
 		(value: () => mixed, error?: ErrorValidator, message?: string): any;
 	};
-	// Assert that function doesn't throw an error or promise resolves.
+	// Assert that the promise resolves, or the function doesn't throws or returns a resolved promise.
 	notThrows: {
 		(value: PromiseLike<mixed>, message?: string): Promise<void>;
 		(value: () => mixed, message?: string): void;
--- a/index.js.flow
+++ b/index.js.flow
@@ -61,7 +61,7 @@ type AssertContext = {
 		(value: PromiseLike<mixed>, error?: ErrorValidator, message?: string): Promise<any>;
 		(value: () => mixed, error?: ErrorValidator, message?: string): any;
 	};
-	// Assert that the promise resolves, or the function doesn't throws or returns a resolved promise.
+	// Assert that the promise resolves, or the function doesn't throw or return a resolved promise.
 	notThrows: {
 		(value: PromiseLike<mixed>, message?: string): Promise<void>;
 		(value: () => mixed, message?: string): void;
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -198,13 +198,14 @@ function wrapAssertions(callbacks) {
 				coreAssertThrowsErrorArg = err;
 			}
 
+			let maybePromise;
 			const test = (fn, stack) => {
 				let actual;
 				let threw = false;
 				try {
 					coreAssert.throws(() => {
 						try {
-							fn();
+							maybePromise = fn();
 						} catch (err) {
 							actual = err;
 							threw = true;
@@ -224,7 +225,7 @@ function wrapAssertions(callbacks) {
 				}
 			};
 
-			if (promise) {
+			const handlePromise = promise => {
 				// Record stack before it gets lost in the promise chain.
 				const stack = getStack();
 				const intermediate = promise.then(value => {
@@ -238,6 +239,10 @@ function wrapAssertions(callbacks) {
 				pending(this, intermediate);
 				// Don't reject the returned promise, even if the assertion fails.
 				return intermediate.catch(noop);
+			};
+
+			if (promise) {
+				return handlePromise(promise);
 			}
 
 			try {
@@ -245,6 +250,14 @@ function wrapAssertions(callbacks) {
 				pass(this);
 				return retval;
 			} catch (err) {
+				if (maybePromise) {
+					if (isPromise(maybePromise)) {
+						return handlePromise(maybePromise);
+					}
+					if (isObservable(maybePromise)) {
+						return handlePromise(observableToPromise(maybePromise));
+					}
+				}
 				fail(this, err);
 			}
 		},
@@ -265,9 +278,12 @@ function wrapAssertions(callbacks) {
 				return;
 			}
 
+			let maybePromise;
 			const test = (fn, stack) => {
 				try {
-					coreAssert.doesNotThrow(fn);
+					coreAssert.doesNotThrow(() => {
+						maybePromise = fn();
+					});
 				} catch (err) {
 					throw new AssertionError({
 						assertion: 'notThrows',
@@ -278,17 +294,29 @@ function wrapAssertions(callbacks) {
 				}
 			};
 
-			if (promise) {
+			const handlePromise = promise => {
 				// Record stack before it gets lost in the promise chain.
 				const stack = getStack();
 				const intermediate = promise.then(noop, reason => test(makeRethrow(reason), stack));
 				pending(this, intermediate);
 				// Don't reject the returned promise, even if the assertion fails.
 				return intermediate.catch(noop);
+			};
+
+			if (promise) {
+				return handlePromise(promise);
 			}
 
 			try {
 				test(fn);
+				if (maybePromise) {
+					if (isPromise(maybePromise)) {
+						return handlePromise(maybePromise);
+					}
+					if (isObservable(maybePromise)) {
+						return handlePromise(observableToPromise(maybePromise));
+					}
+				}
 				pass(this);
 			} catch (err) {
 				fail(this, err);
--- a/readme.md
+++ b/readme.md
@@ -952,7 +952,7 @@ test('rejects', async t => {
 });
 ```
 
-When testing an `async function` you must also wait for the assertion to complete:
+When testing an asynchronous function you must also wait for the assertion to complete:
 
 ```js
 test('throws', async t => {
--- a/readme.md
+++ b/readme.md
@@ -913,7 +913,7 @@ Assert that `value` is not deeply equal to `expected`. The inverse of `.deepEqua
 
 ### `.throws(function|promise, [error, [message]])`
 
-Assert that `function` throws an error, or `promise` rejects with an error.
+Assert that `function` throws an error, `promise` rejects with an error, or `function` returns a rejected `promise`.
 
 `error` can be an error constructor, error message, regex matched against the error message, or validation function.
 
@@ -952,9 +952,21 @@ test('rejects', async t => {
 });
 ```
 
+When testing an `async function` you must also wait for the assertion to complete:
+
+```js
+test('throws', async t => {
+	const error = await t.throws(async () => {
+		throw new TypeError('🦄');
+	}, TypeError);
+
+	t.is(error.message, '🦄');
+});
+```
+
 ### `.notThrows(function|promise, [message])`
 
-Assert that `function` does not throw an error or that `promise` does not reject with an error.
+Assert that `function` does not throw an error, `promise` does not reject with an error, or `function` returns a promise that does not reject with an error.
 
 Like the `.throws()` assertion, when testing a promise you must wait for the assertion to complete:
 
--- a/test/assert.js
+++ b/test/assert.js
@@ -656,11 +656,7 @@ test('.throws()', t => {
 		});
 	});
 
-	return eventuallyFailsWith(t, assertions.throws(() => Promise.resolve('foo')), {
-		assertion: 'throws',
-		message: 'Expected promise to be rejected, but it was resolved instead',
-		values: [{label: 'Resolved with:', formatted: /'foo'/}]
-	});
+	t.end();
 });
 
 test('.throws() returns the thrown error', t => {
@@ -750,13 +746,7 @@ test('.notThrows()', t => {
 		values: [{label: 'Threw:', formatted: /foo/}]
 	});
 
-	return eventuallyFailsWith(t, assertions.notThrows(() => {
-		return Promise.reject(new Error('foo'));
-	}), {
-		assertion: 'notThrows',
-		message: '',
-		values: [{label: 'Threw:', formatted: /foo/}]
-	});
+	t.end();
 });
 
 test('.notThrows() returns undefined for a fulfilled promise', t => {
--- a/test/assert.js
+++ b/test/assert.js
@@ -6,6 +6,7 @@ const stripAnsi = require('strip-ansi');
 const React = require('react');
 const renderer = require('react-test-renderer');
 const test = require('tap').test;
+const Observable = require('zen-observable');
 const assert = require('../lib/assert');
 const snapshotManager = require('../lib/snapshot-manager');
 const Test = require('../lib/test');
@@ -655,7 +656,11 @@ test('.throws()', t => {
 		});
 	});
 
-	t.end();
+	return eventuallyFailsWith(t, assertions.throws(() => Promise.resolve('foo')), {
+		assertion: 'throws',
+		message: 'Expected promise to be rejected, but it was resolved instead',
+		values: [{label: 'Resolved with:', formatted: /'foo'/}]
+	});
 });
 
 test('.throws() returns the thrown error', t => {
@@ -673,6 +678,28 @@ test('.throws() returns the rejection reason of promise', t => {
 	const expected = new Error();
 
 	return assertions.throws(Promise.reject(expected)).then(actual => {
+		t.is(actual, expected);
+		t.end();
+	});
+});
+
+test('.throws() returns the rejection reason of a promise returned by the function', t => {
+	const expected = new Error();
+
+	return assertions.throws(() => {
+		return Promise.reject(expected);
+	}).then(actual => {
+		t.is(actual, expected);
+		t.end();
+	});
+});
+
+test('.throws() returns the error of an observable returned by the function', t => {
+	const expected = new Error();
+
+	return assertions.throws(() => {
+		return new Observable(observer => observer.error(expected));
+	}).then(actual => {
 		t.is(actual, expected);
 		t.end();
 	});
@@ -723,11 +750,33 @@ test('.notThrows()', t => {
 		values: [{label: 'Threw:', formatted: /foo/}]
 	});
 
-	t.end();
+	return eventuallyFailsWith(t, assertions.notThrows(() => {
+		return Promise.reject(new Error('foo'));
+	}), {
+		assertion: 'notThrows',
+		message: '',
+		values: [{label: 'Threw:', formatted: /foo/}]
+	});
 });
 
 test('.notThrows() returns undefined for a fulfilled promise', t => {
 	return assertions.notThrows(Promise.resolve(Symbol(''))).then(actual => {
+		t.is(actual, undefined);
+	});
+});
+
+test('.notThrows() returns undefined for a fulfilled promise returned by the function', t => {
+	return assertions.notThrows(() => {
+		return Promise.resolve(Symbol(''));
+	}).then(actual => {
+		t.is(actual, undefined);
+	});
+});
+
+test('.notThrows() returns undefined for an observable returned by the function', t => {
+	return assertions.notThrows(() => {
+		return Observable.of(Symbol(''));
+	}).then(actual => {
 		t.is(actual, undefined);
 	});
 });
--- a/test/observable.js
+++ b/test/observable.js
@@ -48,7 +48,7 @@ test('returning an observable from a legacy async fn is an error', t => {
 	t.end();
 });
 
-test('handle throws with thrown observable', t => {
+test('handle throws with erroring observable', t => {
 	let result;
 	ava(a => {
 		a.plan(1);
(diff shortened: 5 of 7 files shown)


okyantoro:
Hello, I am new here :smile: 
Need suggestions about this pull request 
Thank you

okyantoro:
Broken on node 4 and 6

novemberborn:
> Broken on node 4 and 6

`async` is only supported in Node.js 8.

I don't think it matters whether the function is truly an `AsyncFunction`. Instead we should check the return value of `fn` and if it's a promise or an observable wait for it to resolve, like we already do when you pass a promise or observable directly.

okyantoro:
> Instead we should check the return value of fn and if it's a promise or an observable wait for it to resolve, like we already do when you pass a promise or observable directly.

@novemberborn So we need to try to execute the `fn` inside a ```try {} catch(e) {}```, right?

novemberborn:
> So we need to try to execute the fn inside a try {} catch(e) {}, right?

@okyantoro close. The way you're doing it now you're calling `fn()` twice if it doesn't return a promise/observable. It should only be called once.

I think the trick is to [capture `fn()`'s return value](https://github.com/avajs/ava/blob/bcb77fc46dbb77ddb14ac8eda8f2cc7e856c2416/lib/assert.js#L191), and then when [`coreAssert.throws()` itself, well, throws, see if a promise/observable was returned](https://github.com/avajs/ava/blob/bcb77fc46dbb77ddb14ac8eda8f2cc7e856c2416/lib/assert.js#L200) and run the [promise handling routine](https://github.com/avajs/ava/blob/bcb77fc46dbb77ddb14ac8eda8f2cc7e856c2416/lib/assert.js#L212:L224) (which should be extracted into a function, and you'd have to convert the observable to a promise as well).

This isn't great because of how we wrap `coreAssert.throws()`. #1047 contains the proposal that removes this dependency, but it's not a deal breaker for this feature.

okyantoro:
@novemberborn I realize that calling `fn()` twice is dangerous, especially when the `fn()` it self modify outer states.
I try different approach like what you suggested and call the `fn()` only once.

lib/assert.js
```diff
@@ -222,13 +223,25 @@ function wrapAssertions(callbacks) {
 				pending(this, intermediate);
 				// Don't reject the returned promise, even if the assertion fails.
 				return intermediate.catch(noop);
+			};
+
+			if (promise) {
+				return handlePromise.call(this, promise);
```
novemberborn: `handlePromise` is an arrow function, so `.call(this` isn't necessary. Just do `return handlePromise(promise)`.

lib/assert.js
```diff
@@ -222,13 +223,25 @@ function wrapAssertions(callbacks) {
...
 				const retval = test(fn);
 				pass(this);
 				return retval;
 			} catch (err) {
+				if (maybePromise) {
+					if (isPromise(maybePromise)) {
+						return handlePromise.call(this, maybePromise);
+					}
+					if (isObservable(maybePromise)) {
+						return handlePromise.call(this, observableToPromise(maybePromise));
```
novemberborn: Could you add a test for this code path?

okyantoro:
hi @novemberborn I'll continue this on saturday or sunday. Thank you for the suggestions and help

lib/assert.js
okyantoro: sure, we can use already provided library `zen-observable` in package.json

okyantoro:
hello @novemberborn 
In `notThrows()`, the `fn()` is executed inside [`coreAssert.doesNotThrow()`](https://github.com/avajs/ava/blob/4f896c2832a186bf0c7f6b0f21b0e36b8e8fdcf2/lib/assert.js#L270) directly. Is it okay to wrap the `fn()` by using arrow function so we can check the returned value of `fn()`?

novemberborn:
> In notThrows(), the fn() is executed inside coreAssert.doesNotThrow() directly. Is it okay to wrap the fn() by using arrow function so we can check the returned value of fn()?

@okyantoro yup.

okyantoro:
@novemberborn when I try to add function signature for `function` that return `promise`, I get the following error:

```
Error: test/flow-types/regression-1148.js.flow:6
  6: 	t.throws(() => { throw new Error(); });
     	^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ call of method `throws`. Method cannot be called on any member of intersection type
  6: 	t.throws(() => { throw new Error(); });
     	^ intersection
  Member 1:
   91: type ContextualTestContext         = TestContext & { context: any; };
                                            ^^^^^^^^^^^ TestContext. See: index.js.flow:91
  Error:
    6: 	t.throws(() => { throw new Error(); });
       	^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ call of method `throws`. Method cannot be called on any member of intersection type
   91: type ContextualTestContext         = TestContext & { context: any; };
                                            ^^^^^^^^^^^ intersection. See: index.js.flow:91
    Member 1:
     85: type TestContext = AssertContext & {
                            ^^^^^^^^^^^^^ AssertContext. See: index.js.flow:85
    Error:
      6: 	t.throws(() => { throw new Error(); });
         	^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ call of method `throws`. Could not decide which case to select
      6: 	t.throws(() => { throw new Error(); });
         	^^^^^^^^ callable object type
      Case 2 may work:
       62: 		(value: () => PromiseLike<mixed>, error?: ErrorValidator, message?: string): Promise<any>;
           		^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ function type. See: index.js.flow:62
      But if it doesn't, case 3 looks promising too:
       63: 		(value: () => mixed, error?: ErrorValidator, message?: string): any;
           		^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ function type. See: index.js.flow:63
      Please provide additional annotation(s) to determine whether case 2 works (or consider merging it with case 3):
        6: 	t.throws(() => { throw new Error(); });
           	              ^ return
    Member 2:
                                            v
     85: type TestContext = AssertContext & {
     86: 	title: string;
     87: 	plan(count: number): void;
    ...:
     90: };
         ^ object type. See: index.js.flow:85
    Error:
      6: 	t.throws(() => { throw new Error(); });
         	  ^^^^^^ property `throws`. Property not found in
                                            v
     85: type TestContext = AssertContext & {
     86: 	title: string;
     87: 	plan(count: number): void;
    ...:
     90: };
         ^ object type. See: index.js.flow:85
  Member 2:
   91: type ContextualTestContext         = TestContext & { context: any; };
                                                          ^^^^^^^^^^^^^^^^^ object type. See: index.js.flow:91
  Error:
    6: 	t.throws(() => { throw new Error(); });
       	  ^^^^^^ property `throws`. Property not found in
   91: type ContextualTestContext         = TestContext & { context: any; };
                                                          ^^^^^^^^^^^^^^^^^ object type. See: index.js.flow:91

Error: test/flow-types/regression-1148.js.flow:9
  9: 	t.notThrows(() => { return; });
     	^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ call of method `notThrows`. Method cannot be called on any member of intersection type
  9: 	t.notThrows(() => { return; });
     	^ intersection
  Member 1:
   91: type ContextualTestContext         = TestContext & { context: any; };
                                            ^^^^^^^^^^^ TestContext. See: index.js.flow:91
  Error:
    9: 	t.notThrows(() => { return; });
       	^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ call of method `notThrows`. Method cannot be called on any member of intersection type
   91: type ContextualTestContext         = TestContext & { context: any; };
                                            ^^^^^^^^^^^ intersection. See: index.js.flow:91
    Member 1:
     85: type TestContext = AssertContext & {
                            ^^^^^^^^^^^^^ AssertContext. See: index.js.flow:85
    Error:
      9: 	t.notThrows(() => { return; });
         	^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ call of method `notThrows`. Could not decide which case to select
      9: 	t.notThrows(() => { return; });
         	^^^^^^^^^^^ callable object type
      Case 2 may work:
       68: 		(value: () => PromiseLike<mixed>, message?: string): Promise<void>;
           		^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ function type. See: index.js.flow:68
      But if it doesn't, case 3 looks promising too:
       69: 		(value: () => mixed, message?: string): void;
           		^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ function type. See: index.js.flow:69
      Please provide additional annotation(s) to determine whether case 2 works (or consider merging it with case 3):
        9: 	t.notThrows(() => { return; });
           	                 ^ return
    Member 2:
                                            v
     85: type TestContext = AssertContext & {
     86: 	title: string;
     87: 	plan(count: number): void;
    ...:
     90: };
         ^ object type. See: index.js.flow:85
    Error:
      9: 	t.notThrows(() => { return; });
         	  ^^^^^^^^^ property `notThrows`. Property not found in
                                            v
     85: type TestContext = AssertContext & {
     86: 	title: string;
     87: 	plan(count: number): void;
    ...:
     90: };
         ^ object type. See: index.js.flow:85
  Member 2:
   91: type ContextualTestContext         = TestContext & { context: any; };
                                                          ^^^^^^^^^^^^^^^^^ object type. See: index.js.flow:91
  Error:
    9: 	t.notThrows(() => { return; });
       	  ^^^^^^^^^ property `notThrows`. Property not found in
   91: type ContextualTestContext         = TestContext & { context: any; };
                                                          ^^^^^^^^^^^^^^^^^ object type. See: index.js.flow:91

Error: test/flow-types/regression-1148.js.flow:12
 12: 	const error = t.throws(() => { throw new Error(); });
     	              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ call of method `throws`. Method cannot be called on any member of intersection type
 12: 	const error = t.throws(() => { throw new Error(); });
     	              ^ intersection
  Member 1:
   91: type ContextualTestContext         = TestContext & { context: any; };
                                            ^^^^^^^^^^^ TestContext. See: index.js.flow:91
  Error:
   12: 	const error = t.throws(() => { throw new Error(); });
       	              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ call of method `throws`. Method cannot be called on any member of intersection type
   91: type ContextualTestContext         = TestContext & { context: any; };
                                            ^^^^^^^^^^^ intersection. See: index.js.flow:91
    Member 1:
     85: type TestContext = AssertContext & {
                            ^^^^^^^^^^^^^ AssertContext. See: index.js.flow:85
    Error:
     12: 	const error = t.throws(() => { throw new Error(); });
         	              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ call of method `throws`. Could not decide which case to select
     12: 	const error = t.throws(() => { throw new Error(); });
         	              ^^^^^^^^ callable object type
      Case 2 may work:
       62: 		(value: () => PromiseLike<mixed>, error?: ErrorValidator, message?: string): Promise<any>;
           		^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ function type. See: index.js.flow:62
      But if it doesn't, case 3 looks promising too:
       63: 		(value: () => mixed, error?: ErrorValidator, message?: string): any;
           		^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ function type. See: index.js.flow:63
      Please provide additional annotation(s) to determine whether case 2 works (or consider merging it with case 3):
       12: 	const error = t.throws(() => { throw new Error(); });
           	                            ^ return
    Member 2:
                                            v
     85: type TestContext = AssertContext & {
     86: 	title: string;
     87: 	plan(count: number): void;
    ...:
     90: };
         ^ object type. See: index.js.flow:85
    Error:
     12: 	const error = t.throws(() => { throw new Error(); });
         	                ^^^^^^ property `throws`. Property not found in
                                            v
     85: type TestContext = AssertContext & {
     86: 	title: string;
     87: 	plan(count: number): void;
    ...:
     90: };
         ^ object type. See: index.js.flow:85
  Member 2:
   91: type ContextualTestContext         = TestContext & { context: any; };
                                                          ^^^^^^^^^^^^^^^^^ object type. See: index.js.flow:91
  Error:
   12: 	const error = t.throws(() => { throw new Error(); });
       	                ^^^^^^ property `throws`. Property not found in
   91: type ContextualTestContext         = TestContext & { context: any; };
                                                          ^^^^^^^^^^^^^^^^^ object type. See: index.js.flow:91
```

or maybe I did something wrong when added the function signature.

index.js.flow
```diff
@@ -56,12 +56,14 @@ type AssertContext = {
 	// Assert that value is not deep equal to expected.
 	notDeepEqual<U>(value: U, expected: U, message?: string): void;
 	// Assert that function throws an error or promise rejects.
+	// If the given function returns Promise, then the Promise will be evaluated instead.
```
sindresorhus: `Promise` => `a Promise`

index.js.flow
```diff
@@ -56,12 +56,14 @@ type AssertContext = {
...
 	// @param error Can be a constructor, regex, error message or validation function.
 	throws: {
 		(value: PromiseLike<mixed>, error?: ErrorValidator, message?: string): Promise<any>;
 		(value: () => mixed, error?: ErrorValidator, message?: string): any;
 	};
 	// Assert that function doesn't throw an error or promise resolves.
+	// If the given function returns Promise, then the Promise will be evaluated instead.
```
sindresorhus: `Promise` => `a Promise`

readme.md
```diff
@@ -948,7 +948,7 @@ Assert that `value` is not deeply equal to `expected`. The inverse of `.deepEqua
 
 ### `.throws(function|promise, [error, [message]])`
 
-Assert that `function` throws an error, or `promise` rejects with an error.
+Assert that `function` throws an error, `promise` rejects with an error, or `function` that return rejected `promise`.
```
```
or `function` that return rejected `promise`.
```

⬇️ 

```
or `function` returns a rejected `promise`.
```

readme.md
```diff
@@ -987,9 +997,19 @@ test('rejects', async t => {
 });
 ```
 
+Because a `function` that return a `promise` is allowed to be a parameter, then you can pass `async function` that throws exception as parameter:
```
sindresorhus: This sentence needs to be better worded.

readme.md
```diff
@@ -987,9 +997,19 @@ test('rejects', async t => {
...
+	await t.throws(async () => {
+		throw new TypeError('🦄');
+	});
+});
+```
+
 ### `.notThrows(function|promise, [message])`
 
-Assert that `function` does not throw an error or that `promise` does not reject with an error.
+Assert that `function` does not throw an error or that `promise` does not reject with an error. When you pass a `function` that return a `promise` as parameter, the returned promise will be evaluated instead.
```
sindresorhus: `return` => `returns`

readme.md
okyantoro: maybe we can make it simple like:

> When testing an `async function` you must also wait for the assertion to complete:

@sindresorhus is it better?

index.js.flow
```diff
@@ -56,12 +56,14 @@ type AssertContext = {
 	// Assert that value is not deep equal to expected.
 	notDeepEqual<U>(value: U, expected: U, message?: string): void;
 	// Assert that function throws an error or promise rejects.
+	// If the given function returns a Promise, then the Promise will be evaluated instead.
```
novemberborn: Let's go with:

```
// Assert that the promise rejects, or the function throws or returns a rejected promise.
```

index.js.flow
```diff
@@ -56,12 +56,14 @@ type AssertContext = {
...
 	// @param error Can be a constructor, regex, error message or validation function.
 	throws: {
 		(value: PromiseLike<mixed>, error?: ErrorValidator, message?: string): Promise<any>;
 		(value: () => mixed, error?: ErrorValidator, message?: string): any;
 	};
 	// Assert that function doesn't throw an error or promise resolves.
+	// If the given function returns a Promise, then the Promise will be evaluated instead.
```
novemberborn: Let's go with:

```
// Assert that the promise resolves, or the function doesn't throws or returns a resolved promise.
```

readme.md
```diff
@@ -987,9 +997,19 @@ test('rejects', async t => {
...
 ```
 
+When testing an `async function` you must also wait for the assertion to complete:
+
+```js
+test('rejects', async t => {
+	await t.throws(async () => {
+		throw new TypeError('🦄');
+	});
+});
```
novemberborn: Perhaps replace this example with the async function one you added above, and remove it there.

readme.md
```diff
@@ -987,9 +997,19 @@ test('rejects', async t => {
...
+```
+
 ### `.notThrows(function|promise, [message])`
 
-Assert that `function` does not throw an error or that `promise` does not reject with an error.
+Assert that `function` does not throw an error or that `promise` does not reject with an error. When you pass a `function` that returns a `promise` as parameter, the returned promise will be evaluated instead.
```
novemberborn: > Assert that `function` does not throw an error, `promise` does not reject with an error, or `function` returns a promise that does not reject with an error.

test/assert.js
```diff
@@ -678,6 +679,28 @@ test('.throws() returns the rejection reason of promise', t => {
 	});
 });
 
+test('.throws() returns the rejection reason of function that return rejected `Promise`', t => {
```
novemberborn: > .throws() returns the rejection reason of a promise returned by the function

test/assert.js
```diff
@@ -678,6 +679,28 @@ test('.throws() returns the rejection reason of promise', t => {
...
+
+	return assertions.throws(() => {
+		return Promise.reject(expected);
+	}).then(actual => {
+		t.is(actual, expected);
+		t.end();
+	});
+});
+
+test('.throws() throws exception from `Observable`', t => {
```
novemberborn: > .throws() returns the error of an observable returned by the function

test/assert.js
```diff
@@ -732,6 +755,22 @@ test('.notThrows() returns undefined for a fulfilled promise', t => {
 	});
 });
 
+test('.notThrows() returns undefined for a function that return fulfilled promise', t => {
```
novemberborn: > .notThrows() returns undefined for a fulfilled promise returned by the function

test/assert.js
```diff
@@ -732,6 +755,22 @@ test('.notThrows() returns undefined for a fulfilled promise', t => {
...
+	return assertions.notThrows(() => {
+		return Promise.resolve(Symbol(''));
+	}).then(actual => {
+		t.is(actual, undefined);
+	});
+});
+
+test('.notThrows() returns undefined for a function that return `Observable`', t => {
```
novemberborn: > .notThrows() returns undefined for an observable returned by the function

okyantoro:
hi @novemberborn 
Thank you for reviewing
I think **flow** will not allow if we have two function signatures with subset type like:
```js
(value: () => PromiseLike<mixed>, error?: ErrorValidator, message?: string): Promise<any>;
(value: () => mixed, error?: ErrorValidator, message?: string): any;
```
correct me if I am wrong.

> Don't forget to add pass and failure cases to https://github.com/avajs/ava/blob/a051d3e18dba92c893fddb08490a8627f586c231/test/assert.js#L624:L659 and https://github.com/avajs/ava/blob/master/test/assert.js#L701:L727.

isn't it already covered by these lines:
```js
test('.throws() returns the rejection reason of promise', t => {
	const expected = new Error();

	return assertions.throws(Promise.reject(expected)).then(actual => {
		t.is(actual, expected);
		t.end();
	});
});

test('.throws() returns the rejection reason of a promise returned by the function', t => {
	const expected = new Error();

	return assertions.throws(() => {
		return Promise.reject(expected);
	}).then(actual => {
		t.is(actual, expected);
		t.end();
	});
});
```
and
```js
test('.notThrows() returns undefined for a fulfilled promise', t => {
	return assertions.notThrows(Promise.resolve(Symbol(''))).then(actual => {
		t.is(actual, undefined);
	});
});

test('.notThrows() returns undefined for a fulfilled promise returned by the function', t => {
	return assertions.notThrows(() => {
		return Promise.resolve(Symbol(''));
	}).then(actual => {
		t.is(actual, undefined);
	});
});
```
?

Thank you

novemberborn:
> I think flow will not allow if we have two function signatures with subset type like:
> 
> (value: () => PromiseLike<mixed>, error?: ErrorValidator, message?: string): Promise<any>;
> (value: () => mixed, error?: ErrorValidator, message?: string): any;
> correct me if I am wrong.

I don't know @okyantoro. Don't worry about it, we can always open an issue for somebody else to refine the signature.

>> Don't forget to add pass and failure cases to /test/assert.js@a051d3e#L624:L659 and /test/assert.js@master#L701:L727.
> 
> isn't it already covered by these lines:

Those tests are specifically about the expected pass/fail behavior, given all the ways the assertion can be used. The other tests are about their return value. They exercise the same code paths but they're a little different in intent. It's useful to see all possible behaviors in one place.

okyantoro:
@novemberborn Oh I see. Updated. Is it correct?

novemberborn:
@okyantoro this is great! I pushed some minor tweaks but will merge when CI passes.

Regarding my previous comment:

> Those tests are specifically about the expected pass/fail behavior, given all the ways the assertion can be used. The other tests are about their return value. They exercise the same code paths but they're a little different in intent. It's useful to see all possible behaviors in one place.

Turns out I was wrong here, `test/assert.js` doesn't contain explicit assertions for promises and observables. These are in `test/promise.js` and `test/observable.js`. There's some legacy test organization in the way there 😄 I've made the necessary changes in the commit I just pushed.

okyantoro:
Thank you @novemberborn

PR #1653, opened 2018-01-24
Make `t.log()` behave like `console.log()`

### LINKED ISSUE #1635: Make t.log() behave like console.log()

kugtong33:
Fixes #1635 

Use concordance to format all arguments except string arguments as the existing test will fail and I find it awkward this way. Update test to cater `t.log()` and `t.log(n1, n2, n3, ...)`.

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -122,7 +122,8 @@ function wrapAssertions(callbacks) {
 			}
 		},
 
-		log(...args) {
+		log() {
+			const args = Array.prototype.slice.call(arguments);
 			args.forEach((value, index) => {
 				args[index] = typeof value === 'string' ?
 					value : concordance.format(value, concordanceOptions);
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -122,8 +122,15 @@ function wrapAssertions(callbacks) {
 			}
 		},
 
-		log(text) {
-			log(this, text);
+		log(...args) {
+			args.forEach((value, index) => {
+				args[index] = typeof value === 'string' ?
+					value : concordance.format(value, concordanceOptions);
+			});
+
+			if (args.length > 0) {
+				log(this, args.join(' '));
+			}
 		},
 
 		deepEqual(actual, expected, message) {
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -123,11 +123,10 @@ function wrapAssertions(callbacks) {
 		},
 
 		log() {
-			const args = Array.prototype.slice.call(arguments);
-			args.forEach((value, index) => {
-				args[index] = typeof value === 'string' ?
-					value : concordance.format(value, concordanceOptions);
-			});
+			const args = Array.from(arguments, value =>
+				typeof value === 'string' ?
+					value : concordance.format(value, concordanceOptions)
+			);
 
 			if (args.length > 0) {
 				log(this, args.join(' '));
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -123,10 +123,11 @@ function wrapAssertions(callbacks) {
 		},
 
 		log() {
-			const args = Array.from(arguments, value =>
-				typeof value === 'string' ?
-					value : concordance.format(value, concordanceOptions)
-			);
+			const args = Array.from(arguments, value => {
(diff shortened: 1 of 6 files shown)


lib/assert.js
```diff
@@ -122,8 +122,15 @@ function wrapAssertions(callbacks) {
 			}
 		},
 
-		log(text) {
-			log(this, text);
+		log(...args) {
```
novemberborn: For the next few months we're still targeting Node.js 4, which doesn't support rest parameters. You'll have to use plain old `arguments` instead.

lib/assert.js
```diff
@@ -122,8 +122,16 @@ function wrapAssertions(callbacks) {
...
 		},
 
-		log(text) {
-			log(this, text);
+		log() {
+			const args = Array.prototype.slice.call(arguments);
+			args.forEach((value, index) => {
+				args[index] = typeof value === 'string' ?
+					value : concordance.format(value, concordanceOptions);
+			});
```
novemberborn: Since you're really mapping `args` anyway this would be better as:

```js
const args = Array.from(arguments, value => {
  // ...
});
```

kugtong33: that makes sense! I was trying to find a better way :)

novemberborn:
Thanks @kugtong33!

PR #1657, opened 2018-01-26
Add context support on before hook

### LINKED ISSUE #1579: Allow context in before hook
It would be useful to have context available in `before` hooks, so that messy workarounds like in https://github.com/avajs/ava/issues/1542 aren't required. It's useful in cases where you're doing integration testing and are downloading several URLs whose content will be used across multiple tests.

Without being able to share context in before hooks you would need to resort to the workaround, or download the page for each test each time, even though the content doesn't change between tests.

kugtong33:
Fixes #1579 

Create the `context` object before passing it to the `before` hook and then clone it, `Object.create(context)`, before passing it the `beforeEach` hook so that each `beforeEach - test - afterEach` sequence has its own copy of the context.


lib/test-collection.js
```diff
@@ -195,18 +195,22 @@ class TestCollection extends EventEmitter {
...
-		const allTests = new Sequence([serialTests, concurrentTests]);
+		const sequenceTests = context => {
+			const serialTests = new Sequence(this._buildTests(this.tests.serial, context), this.bail);
+			const concurrentTests = new Concurrent(this._buildTests(this.tests.concurrent, context), this.bail);
+			return new Sequence([serialTests, concurrentTests]);
+		};
 
 		let finalTests;
 		// Only run before and after hooks when there are unskipped tests
 		if (this._hasUnskippedTests()) {
```
novemberborn: `_hasUnskippedTests()` depends on `allTests` having been created. Wrapping the test building in `sequenceTests()` won't work. Looks like this behavior isn't covered by an integration test though.

I think it's fine to still build `beforeHooks` and `afterHooks`. The trick is to not *run* them if all tests are skipped. We still "run" the skipped tests since we want to show them in the test results, but they don't actually execute.

lib/test-collection.js
```diff
@@ -165,26 +165,26 @@ class TestCollection extends EventEmitter {
...
 	}
 
-	_buildTestWithHooks(test) {
+	_buildTestWithHooks(test, context) {
 		if (test.metadata.skipped || test.metadata.todo) {
 			return new Sequence([this._skippedTest(this._buildTest(test))], true);
 		}
 
-		const context = {context: {}};
+		const contextRef = context ? Object.create(context) : {context: {}};
```
novemberborn: It's actually `context.context` that needs to be (shallowly) cloned. Let's use <https://www.npmjs.com/package/lodash.clone>.

You'll have to create a context ref for the `before` and `after` hooks, and *before* those hooks have run you need to pass a cloned ref to the `beforeEach` and `afterEach` hooks as well as the tests. Something like this should work if you [modify the `t.context` getter and setter](https://github.com/avajs/ava/blob/a051d3e18dba92c893fddb08490a8627f586c231/lib/test.js#L55:L69).

```js
class ContextRef {
  constructor () {
    this.value = {}
  }

  get () {
    return this.value
  }

  set (newValue) {
    this.value = newValue
  }

  copy () {
    return new LateBinding(this)
  }
}

class LateBinding extends ContextRef {
  constructor (ref) {
    super()
    this.ref = ref
    this.bound = false
  }

  get () {
    if (!this.bound) this.set(clone(this.ref.get()))
    return super.get()
  }

  set (newValue) {
    this.bound = true
    super.set(newValue)
  }
}
```

The same copied ref should be passed to the `beforeEach` and `afterEach` hooks as well as the tests. The first time the context is accessed it'll be "bound" to the value set in the `before` hooks. You can pass the original ref to the `after` hooks.

novemberborn:
> if it was that easy then it should have been a good-for-beginner issue :D, thanks for the detailed review, will work on it

I probably spend half an hour thinking about your code and the expected behavior so I wouldn't say it's *that* easy! 😉

lib/test-collection.js
kugtong33: @novemberborn , this is quite circular, and gives me an error on XO, I now understand that the `contextRef` is injected into the `Test` class then to the `ExecutionContext` (that is why you pointed me that way), is it necessary that we should create the `LateBinding` class? Is it possible that we can just return the `copy()` method a clone of itself?

lib/test-collection.js
kugtong33: If I created the context above and injected it directly, I think it should have work as intended, but why did the `after` hook still got a mutated(mutations from `afterEach`, `beforeEach`, `tests`) context, when I ran the tests?

```javascript
const context = {context: {}};

const serialTests = new Sequence(this._buildTests(this.tests.serial, context), this.bail);
const concurrentTests = new Concurrent(this._buildTests(this.tests.concurrent, context), this.bail);
const allTests = new Sequence([serialTests, concurrentTests]);

...

if (this._hasUnskippedTests()) {
    const beforeHooks = new Sequence(this._buildHooks(this.hooks.before, null, context));
    const afterHooks = new Sequence(this._buildHooks(this.hooks.after, null, context));
    finalTests = new Sequence([beforeHooks, allTests, afterHooks], true);
} else {
    finalTests = new Sequence([allTests], true);
}
```

novemberborn:
@kugtong33 if you could push your best try I'll have a look tomorrow. Don't worry if it doesn't actually work.

kugtong33:
@novemberborn , just checking in, I will still be working with this one, had a terrible PC problem for the last 2 days

kugtong33:
@novemberborn , sorry for the late update, I tried the suggestion above that is to bound `context.context` value, I did it in a simple way, create a `Context` class with only one method `clone` then pass the *bound context* to the `beforeEach-test-afterEach` sequence, also updated the shared context test to allow `t.context` on `after` hook.

I used `lodash.clonedeep` since the test used an array as an example.

PS: if this gets accepted, can I be the one to update the docs as well? :D

lib/test-collection.js
```diff
@@ -1,10 +1,42 @@
...
+class ContextBind {
+	constructor(ref) {
+		this.ref = ref;
+		this.bound = false;
+		this.boundContext = {};
+	}
+
+	get context() {
+		if (!this.bound) {
+			this.context = cloneDeep(this.ref.context);
```
novemberborn: I find relying on the setter's behavior to be quite surprising. It's not immediately clear that `this.bound` and `this.boundContext` are actually updated.

novemberborn:
@kugtong33 I've just merged #1670 which changes the TypeScript and Flow type definitions. Currently they're hardcoded to set `t.context` to `null` in `before` and `after` hooks. Could you update them to pass `Context` instead? You'll have to merge the `master` branch into this one, and change `AfterInterface<null>` and `BeforeInterface<null>` to `AfterInterface<Context>` and `BeforeInterface<Context>` in the `index.d.ts` and `index.js.flow` files. Thanks!

kugtong33:
:+1:  for the clear intent, there was a lot going on when I checked on master and didn't want this discussion to get messed up by the new commits so I left it that way, and thanks for the guidance :)

kugtong33:
@novemberborn , it's fine, will work with the documentation on a separate merge request :)

PR #1661, opened 2018-01-28
Share glob caches

novemberborn:
Share glob caches when finding test and helper files, and when rerunning all files in watch mode. This should provide a performance improvement, though I suppose certain file changes in watch mode may no longer be detected. I can't quite figure out what those would be though.


sindresorhus:
> I can't quite figure out what those would be though.

Me neither. We'll just have to see.

PR #1670, opened 2018-01-29
Restrict modifier chaining

### LINKED ISSUE #1182: Restrict test modifier chaining
Currently test modifiers allow silly and confusing combinations. We have [linting rules](https://github.com/avajs/eslint-plugin-ava/issues/166#issuecomment-272231304) to warn about them, and [disallow some of them in the type definitions](https://github.com/avajs/ava/issues/1005). We should go further and restrict the modifier chains so only useful combinations are allowed, but provide useful error feedback if disallowed combinations are used.

Examples of silly chains:

* `test.always.afterEach.cb()`
* `test.after.test()`
* `test.always.always.serial.always.afterEach(() => {})`

Examples of confusing chains:

* `test.serial.afterEach()` (https://github.com/avajs/ava/issues/1178)

See also the discussion in https://github.com/avajs/ava/issues/1005.

novemberborn:
Explicitly specify allowable chains, with some ground rules in mind:

Test chaining rules:

* `serial` must come at the start
* `only` and `skip` must come at the end
* `failing` must come at the end, but can be followed by `only` and `skip`
* `only` and `skip` cannot be chained together
* no repeating

Hook chaining rules:

* `always` comes immediately after "after hooks"
* `skip` must come at the end
* no `only`
* no repeating

Additionally:

* `todo` cannot be chained, except after `serial`
* hooks are not available on `serial`

This commit also removes now unnecessary assertions from TestCollection.

Fixes #1182.

---

This allows the TypeScript definition to be simplified. I've done an initial refactor. Note that I've not included any JSDoc comments. I think we can do a lot better than we what we had before but I didn't want to sort that out just yet. Will open a follow-up ticket. I have fixed `throws` / `notThrows` taking observables. Would be nice to get a better type definition for `Observable` though. Will ask more explicitly in a follow-up ticket.

* [ ] Create issue for `Observable` typing in TS and Flow
* [ ] Create issue for embedding documentation in the TS and Flow definitions
* [ ] Create issue in `eslint-plugin-ava` to check chains when linting


novemberborn:
Finished the TS refactor and did a similar one for Flow too. Also wrote some documentation for Flow, though still need to add the best way of stripping the type annotations. Will file a follow-up ticket for that too.

novemberborn:
I've documented how to strip flow type annotations and improved typings for `t.throws()` and `t.notThrows()`. TS should be fine without explicit type hints, though Flow needs some help. I've updated the documentation to reflect this.

lib/main.js
```diff
@@ -95,9 +95,21 @@ globals.setImmediate(() => {
...
-// See: https://github.com/Microsoft/TypeScript/issues/2242#issuecomment-83694181
-module.exports.default = runner.chain;
+const makeCjsExport = () => {
+	function test() {
+		return runner.chain.apply(null, arguments);
+	}
+	return Object.assign(test, runner.chain);
+};
+
+// Support ComonJS modules by exporting a test function that can be fully
```
sindresorhus: ComonJS => CommonJS

lib/main.js
```diff
@@ -95,9 +95,21 @@ globals.setImmediate(() => {
...
+// chained. Also support ES module loaders by exporting __esModule and a
+// default. Support `import {* as ava} from 'ava'` use cases by exporting a
```
sindresorhus: `import {* as ava} from 'ava'` => `import * as ava from 'ava'`

sindresorhus:
Good to see this being cleaned up before the 1.0 release.

sindresorhus:
Can you open an issue on `eslint-plugin-ava` about making a rule for this?

novemberborn:
@sindresorhus Thanks. Will make the fixes and open follow-up issues. Will sneak in TypeScript 2.7 as well.

I'm considering revisiting the `import {serial as test}` / `test.before()` use-case in this PR. I think having accurate imports is more important than trying to "help". We should make `serial` with hooks work as expected too, and not support it before that's done.

Issue #1682, opened 2018-02-04
Allow tests to take stdin from the master test process


jviotti:
## Description

We have a use case where we need some end to end tests to take stdin input (key strokes) from the user before doing some assertions (the test waits for some human input intervention). In particular, we have test suite that performs hardware tests on a board. Some tests are hard to automate (ie LED blinks), so before investing in more advanced technology, we decided to prompt the user to check certain things. In the case of LEDs, the test case will start blinking the LED for like 5 seconds, and ask the user if he can confirm it worked.

We noticed that if we use something like Inquirer.js, the parent Ava process that receives the key strokes will not forward them to the child test processes, as the `silent: true` option from `child_process.fork()` is set.

I understand that this might not make sense given that Ava spawns test processes in parallel, and therefore key strokes sent to the master process might be piped to more than one test at the same time, so what about piping stdin to the child's stdin if `serial: true`? 

We have a small patch that pipes stdio and makes it all work for us:

```diff
diff --git a/lib/fork.js b/lib/fork.js
index 0ca0f45..655ed8f 100644
--- a/lib/fork.js
+++ b/lib/fork.js
@@ -43,6 +43,10 @@ module.exports = (file, opts, execArgv) => {
                execArgv: execArgv || process.execArgv
        });

+       if (opts.serial) {
+         process.stdin.pipe(ps.stdin);
+       }
+
        const relFile = path.relative('.', file);

        let exiting = false;
```

What do you think?

### Test Source

```js
const ava = require('ava')
const inquirer = require('inquirer')

ava.test('should pass', (test) => {
  return inquirer.prompt([
    {
      type: 'confirm',
      name: 'result',
      message: 'Did the test pass?',
      default: false
    }
  ]).then((answer) => {
    test.true(answer.result)
  })
})
```

### Config

Copy the relevant section from `package.json`:

```json
{
  "ava": {
    "verbose": true,
    "serial": true
  }
}
```

## Environment

Tell us which operating system you are using, as well as which versions of Node.js, npm, and AVA. Run the following to get it quickly:

- Node.js v6.1.0
- darwin 17.4.0

novemberborn:
@jviotti this is a very interesting use case. Typically our response is that you should wrap the invocation of AVA in some other script.

You're right that stdin forwarding would work in `--serial` mode, but it wouldn't work in watch mode. There's just too many ways in which users may expect to use AVA that your proposal is not compatible with.

[...]

PR #1683, opened 2018-02-04
Pipe stdin to child test processes if serial is set to true

jviotti:
This allows serial test suites to consume key strokes from standard
input during test executions.

Fixes: https://github.com/avajs/ava/issues/1682
Signed-off-by: Juan Cruz Viotti <jv@jviotti.com>


novemberborn:
@jviotti thanks for digging into the code base to figure out how this could be implemented. I'm closing this  given my issue comment https://github.com/avajs/ava/issues/1682#issuecomment-363746586.

Issue #1684, opened 2018-02-04
Default hooks to run concurrently, with explicit serial hooks possible


novemberborn:
Currently, hook execution subtly differs from test execution. Hooks always execute sequentially, even if they're not marked as being serial.

Prompted by #1178, hooks are removed from `test.serial` in #1670. I think we should keep such hooks, but they should work properly:

* Hooks always execute in the order they're defined (by category, e.g. `before`, `afterEach.always`, etc)
* `beforeEach`, `afterEach` and `afterEach.always` hooks execute for each test, when the test executes
  * Serial tests are executed before concurrent ones, if any
  * Concurrent tests are executed after serial ones, if any, without regard for the state of other concurrent tests
  * For each test, `beforeEach` hooks are executed after `before` hooks, if any
  * For each test, `afterEach` hooks are executed after the test has finished
  * For each test, `afterEach.always` hooks are executed after `afterEach` hooks, if any
  * For serial tests, the `beforeEach` hooks are executed after the previous serial test's `afterEach.always` hook (or `afterEach` hook, or test itself, if any)
* `after` hooks are executed after all tests have finished
* `after.always` hooks are executed after `after` hooks
* Unlike tests, serial hooks are not reordered to run first
* Serial hooks are executed after preceding hooks in their category have finished
* Concurrent hooks execute concurrently, without regard for the state of other concurrent hooks (in their category)

Here, "serial" means tests and hooks explicitly marked as serial. "Concurrent" means tests and hooks not marked as serial, that execute asynchronously either by using an asynchronous method, returning a promise or observable, or using the `cb` interface. Synchronous tests not explicitly marked as serial may execute concurrently with previous concurrent test, even though they finish immediately.

This proposal lets us bring back hooks to `test.serial`, but now their execution will be properly understood. It might break some tests with multiple concurrent hooks in each category, but I expect that to be rare.

Issue #1416, opened 2017-06-20
--match incorrectly flags test as `.only`, making them unskippable


muyu66:
I meet a question that i can't resolve using AVA

I build a.js file from a.ts using "tsc && ava".

It is source code:

```
import test from 'ava';

test.before(t => {
..........
});

test.skip('test1111', t => {
..........
});

test('test22222', t => {
...........
});
```

But, the code throw a exception :

```
TAP version 13
/home/zhouyu/Web/framework/node_modules/.0.19.1@ava/lib/runner.js:119
            throw new TypeError(validationError);
            ^

TypeError: `only` tests cannot be skipped
    at Runner.addTest (/home/zhouyu/Web/framework/node_modules/.0.19.1@ava/lib/runner.js:119:10)
    at Function.Runner.chain.optionChain (/home/zhouyu/Web/framework/node_modules/.0.19.1@ava/lib/runner.js:99:10)
I am confused that my 'test1111' test isn't a ''only' test.
```

This is my package.json :
```
  "ava": {
    "files": [
      "build/test/*.js"
    ],
    "source": [
      "**/*.{js}"
    ],
    "match": [
      "*"
    ],
    "failFast": false,
    "tap": true
  }
```

My ava version is : **"ava": "^0.19.1"**

This is my usage problem, or, AVA bug?

Help me, thanks.

ORESoftware:
interesting

novemberborn:
The problem is the `match` option. Note that with `*` it doesn't need to be set, since it'll match anything.

The way `match` is implemented is we mark a test as `exclusive` when it matches: https://github.com/avajs/ava/blob/212b5de5759fdb267d837620c3eec7e5695a23e1/lib/runner.js#L114

Unfortunately we do the same when `test.only()` is used! We should set an additional property that tracks whether the test is exclusive because of `test.only()` or `--match`, and make `.skip()` only throw on `test.only()`.

PR #1693, opened 2018-02-06
Refactor how test files are told to run their tests, and more

### LINKED ISSUE #1674: Collecting and running tests needs some cleanup
As raised in #1672 our requirement that tests must be declared synchronously is no longer properly enforced. I think this broke when we switched to limited concurrency (#1467).

Originally all workers sent their tests to the main process, and once all workers were ready did we start running the tests. Any test declared after this point would [throw an error at the point of declaration](https://github.com/avajs/ava/blob/d42865be9f9484a6a3ab7e90a91283ec50c0d405/lib/runner.js#L71:L730). We still have the code logic for this, and a unit test, but the unit test is at the `runner` level rather than an integration test. It doesn't seem to have an effect in #1672.

Further, despite us saying that tests should be declared synchronously, the ["start running your tests" instruction](https://github.com/avajs/ava/blob/d42865be9f9484a6a3ab7e90a91283ec50c0d405/api.js#L274) is sent over IPC which is inherently asynchronous.

We also don't detect later `test()` calls [when we're looking to only run exclusive tests](https://github.com/avajs/ava/blob/d42865be9f9484a6a3ab7e90a91283ec50c0d405/lib/runner.js#L216).

We need to update our test running logic:

* When `test()` is first called, schedule running them using `process.nextTick()`. This should happen in `lib/runner.js` itself to avoid any other asynchronicity from being introduced.
* Any `test()` call in a subsequent tick should cause that test to be marked as failing with an appropriate reason. The test should never run.
* Workers should run tests in the tick after the first one has been registered. There should be no need to wait for IPC with the main process.
* There is no requirement that `test()` is called immediately. This allows for asynchronous module loading and top-level `await`.

It's important that no new tests are registered once tests start running. Allowing it makes it hard to know when all tests are finished, or to tell when a test might be stuck. I think the above approach should work since presumably the worker process doesn't exit whilst there is something going on that will end up registering a new test, and once that completes we'll immediately fail that test without ever running it.

novemberborn:
Apologies, this is a big one! See the commits for details, but the highlights are:

* Remove cruft remaining from when we switched to *always* using the limited concurrency mode.
* The cache directory is now always relative to the project directory. Previously another unnecessary path lookup was done, which I think may have resulted in different cache paths if there were `package.json` files at different levels of the project.
* Compile all test and helpers files before running any test. This doesn't improve time-to-first-test but it does mean that test files can depend on each other. AVA will correctly load the precompiled file. No tests are run if any file fails to compile.
* The real file path is passed to the precompiler.
* `execArgv` is computed when running a test file, not all at once for all files.
* `--fail-fast` now works with limited concurrency mode, preventing new test files from being run. It also fails fast after a timeout or if a test file crashed. This is part of the work needed for #1158.
* The mini and verbose reporters print the number of untested files after `--fail-fast`.
* Improved output for unhandled rejections / uncaught exceptions in the mini and verbose reporters.
* Test runners will now run tests as soon as (in the next tick) they're declared. The main process no longer instructs the workers to startrunning tests. Fixes #1674.


novemberborn:
Additionally:

Fixes #1684. Fixes #1416. Refs #1158.

Properly support serial hooks. Hooks are divided into the following categories:

* before
* beforeEach
* afterEach
* afterEach.always
* after
* after.always

For each category all hooks are run in the order they're declared in. This is different from tests, where serial tests are run before concurrent ones.

By default hooks run concurrently. However a serial hook is not run before all preceding concurrent hooks have completed. Serial hooks are never run concurrently.

Always hooks are now always run, even if --fail-fast is enabled.

Internally, TestCollection, Sequence and Concurrent have been removed. This has led to a major refactor of Runner, and some smaller breaking changes and bug fixes:

* Unnecessary validations have been removed
* Macros can be used with hooks
* A macro is recognized even if no additional arguments are given, so it can modify the test (or hook) title
* --match now also matches todo tests
* Skipped and todo tests are shown first in the output
* --fail-fast prevents subsequent tests from running as long as the failure occurs in a serial test

novemberborn:
And finally, if a failure occurs in one worker, attempt to interrupt other workers. This only works as long as the other worker has not yet started running concurrent tests. Fixes #1158.

sindresorhus:
I've looked through each commit and everything looks good 👌

Issue #1662, opened 2018-01-28
Help update recipes for Babel 7


novemberborn:
We're upgrading AVA to use Babel 7. As part of this the way you configure AVA's Babel pipeline has changed.

There are various recipes which either use Babel 6 dependencies or AVA's previous configuration. For some recipes we just need to update the dependencies and configuration, for others we may need to test them a bit to make sure everything still works with Babel 7 in the mix.

Babel 7 uses the `@babel` scope on npm. Typically dependencies are renamed from `babel-plugin-transform-something` to `@babel/plugin-transform-something`.

AVA's configuration places the Babel options inside a `testOptions` key:

```json
{
  "ava": {
    "babel": {
      "testOptions": {}
    }
  }
}
```

`inherits` has been removed, since it has become the default behavior. `default` has also been removed. See the [Babel recipe](https://github.com/avajs/ava/blob/master/docs/recipes/babel.md) for all the details.

Here's the recipes that need updating:

* [x] [Code coverage](https://github.com/avajs/ava/blob/master/docs/recipes/code-coverage.md)
* [x] [JSPM and SystemJS for ES2015](https://github.com/avajs/ava/blob/master/docs/recipes/jspm-systemjs.md)
* [x] [Precompiling source files with webpack](https://github.com/avajs/ava/blob/master/docs/recipes/precompiling-with-webpack.md)
* [x] [Testing React components](https://github.com/avajs/ava/blob/master/docs/recipes/react.md)
* [x] [Testing Vue.js components](https://github.com/avajs/ava/blob/master/docs/recipes/vue.md)

Let's make this happen! 😃

BusbyActual:
Claimed : ^).

novemberborn:
@BusbyActual woohoo!

There's no tests, no. These are descriptions of how you would use AVA for certain use cases. I suppose we could host a repository where you can put "playgrounds", as it were. Not sure if it makes sense for all recipes though.

BusbyActual:
For 

- [ ] [Testing Vue.js components](https://github.com/avajs/ava/blob/master/docs/recipes/vue.md) I don't see any changes needed. Require extension hooks appears to be a bit different than @babel/register.

novemberborn:
@BusbyActual the configuration needs to be cleaned up: `"babel": "inherit"` is now the default. I don't know if that extension hook works with Babel 7.  There's also an outdated link to the Babel recipe.

NickHurst:
[...]

PR #1694, opened 2018-02-08
Documentation updates

BusbyActual:
Progress #1662

- [x]  [code-coverage ](https://github.com/avajs/ava/blob/master/docs/recipes/code-coverage.md)


docs/recipes/code-coverage.md
```diff
@@ -93,12 +99,16 @@ Note that the build script really has very little to do with AVA, and is just a
...
+
+```
+npm install --save-dev babel-register
+```
 
 ```json
 {
 	"ava": {
-		"require": ["babel-core/register"]
+		"require": ["babel-register"]
```
novemberborn: This should be `@babel/register` I think.

docs/recipes/code-coverage.md
```diff
@@ -54,7 +54,13 @@ First, we need a Babel configuration. The following is just an example. You will
 ```json
 {
 	"babel": {
-		"presets": ["es2015"],
+		"presets": [
+	    ["env", {
```
novemberborn: Babel seems to be [documenting these](https://www.npmjs.com/package/@babel/preset-env) as `@babel/preset-env` rather than just `env`.

docs/recipes/code-coverage.md
```diff
@@ -114,7 +124,13 @@ Combining the above steps, your complete `package.json` should look something li
...
+	      "targets": {
+	        "browsers": ["last 2 versions", "safari >= 7"]
+	      }
+	    }]
+	  ],
 		"plugins": ["transform-runtime"],
```
novemberborn: I suppose `transform-runtime` is not needed if `@babel/preset-env` is used?

BusbyActual:
@novemberborn I like the idea of separate PR's for smaller changes to be reviewed. Hmm there's conflicts in docs

1.  registration documentation is different on [npm's](https://www.npmjs.com/package/@babel/register) vs [babel's](https://babeljs.io/docs/usage/babel-register/) site? Do we want to honor npm?
2. preset documentation is different from [npm's](https://www.npmjs.com/package/@babel/preset-env)  vs [babel's](https://babeljs.io/docs/plugins/preset-env/) site? Do we want to honor npm?

novemberborn:
@BusbyActual I think they have a beta website for Babel 7 somewhere. You'll want to be looking at that one, the current site still covers Babel 6.

BusbyActual:
Updated pr with doc changes and pulled latest

novemberborn:
Thanks for the update @BusbyActual. I was looking at the new version of this recipe and it just seemed too complicated to me. Lots has changed since it was first written. I've done a PR to simplify it instead: https://github.com/avajs/ava/pull/1710.

What do you think?

BusbyActual:
That looks quite a bit simpler. Do we want to integrate the babel changes into the new docs instead?

novemberborn:
> Do we want to integrate the babel changes into the new docs instead?

I'm not sure it's needed. I mean ideally the nyc documentation gets updated but we already have documentation on how to configure Babel, use `@babel/register`, etc.

I'm going to close this in favor of #1710. Hopefully you'll do more PRs for the other recipes! And if the recipes seem overly complicated please feel free to raise that, rather than converting them.

PR #1696, opened 2018-02-11
Bound execution context and changed skip

### LINKED ISSUE #1425: Skipping a snapshot assertion should not cause the next snapshot assertion to fail
Given:

```js
test(t => {
  t.snapshot({foo: 'bar'})
  t.snapshot({baz: 'thud'})
})
```

If the first snapshot is skipped, the second snapshot will fail:

```js
test(t => {
  t.skip.snapshot({foo: 'bar'})
  t.snapshot({baz: 'thud'})
})
```

This is because the test maintains an internal [snapshot invocation counter](https://github.com/avajs/ava/blob/9d3b1bad681e1aad18ad8cd53407a7ccb1fc5782/lib/test.js#L116), and skipping an assertion [does not increment this counter](https://github.com/avajs/ava/blob/9d3b1bad681e1aad18ad8cd53407a7ccb1fc5782/lib/test.js#L96).

`SkipApi.prototype.snapshot` should be implemented such that it increments the counter as if `t.snapshot()` was used.

There's another complication when skipped snapshot assertions are present while the snapshots are updated. The snapshot manager [doesn't allow missing entries](https://github.com/avajs/ava/blob/9d3b1bad681e1aad18ad8cd53407a7ccb1fc5782/lib/snapshot-manager.js#L300). I'm not sure whether that's solvable. It's probably easier to fail the skipped snapshot assertion when updating snapshots, or if the skipped snapshot doesn't exist.

### LINKED ISSUE #1592: Pre-bind assertion methods
Or simply all methods on the test context `t`.

I didn't really buy the use-case in https://github.com/avajs/ava/issues/1255, but after that I've encountered some situations where it could have been useful.

Sometimes I need to skip an assert based on some condition:

```js
test(t => {
	const isTrue = condition ? t.true : t.skip.true;
	isTrue('a', 'b');
});
```

The above results in an error:

```
TypeError {
    message: 'Cannot read property \'_test\' of undefined',
  }
```

Or use a different assertion. I realize it's not a very good practice, but sometimes it's just the best way to do it.

Right now I have to do:

```js
test(t => {
	const isTrue = (condition ? t.true : t.skip.true).bind(t);
	isTrue('a', 'b');
});
```

Which is not very nice.

novemberborn:
* Assertions are now bound. Fixes #1592.
* Assertions are now skipped by calling `.skip()` at the end, e.g. `t.fail.skip()`. This is consistent with the test interface, where `skip()` can only be used at the end of the chain.
* The private `t._test` value has been removed. Instead a WeakMap is used internally to look up the corresponding Test instance given an ExecutionContext.
* Fix `t.snapshot.skip()`. Ensure subsequent `t.snapshot()` calls compare the correct snapshot. Forbid snapshots from being skipped while updating them. Fixes #1425.

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -72,7 +72,6 @@ function wrapAssertions(callbacks) {
 	const pass = callbacks.pass;
 	const pending = callbacks.pending;
 	const fail = callbacks.fail;
-	const log = callbacks.log;
 
 	const noop = () => {};
 	const makeRethrow = reason => () => {
@@ -127,18 +126,6 @@ function wrapAssertions(callbacks) {
 				}));
 			} else {
 				pass(this);
-			}
-		},
-
-		log() {
-			const args = Array.from(arguments, value => {
-				return typeof value === 'string' ?
-					value :
-					concordance.format(value, concordanceOptions);
-			});
-
-			if (args.length > 0) {
-				log(this, args.join(' '));
 			}
 		},
 
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -334,7 +334,7 @@ function wrapAssertions(callbacks) {
 
 			let result;
 			try {
-				result = this._test.compareWithSnapshot(options);
+				result = this.compareWithSnapshot(options);
 			} catch (err) {
 				if (!(err instanceof snapshotManager.SnapshotError)) {
 					throw err;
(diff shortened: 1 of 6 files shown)


novemberborn:
> Fix t.snapshot.skip(). Ensure subsequent t.snapshot() calls compare the correct snapshot. Forbid snapshots from being skipped while updating them. Fixes #1425.

See https://github.com/avajs/ava/issues/1425#issuecomment-364766611 though.

sindresorhus:
> See #1425 (comment) though.

I agree. Way too much of an edge-case.

Issue #1309, opened 2017-03-15
Probably not compiling with babel: `Unexpected import token`


renatorib:
## Description

Testings are failing in `Unexpected import token` (probably due to non babel compiling issue)

### Test Source

initials.js
```js
const initials = (string: string) =>
  string.split(' ').map(word => word[0]).join('').slice(0, 2);

export default initials;
```

initials.test.js 
```js
import test from 'ava';
import initials from './initials';

test(t => t.expect(initials('Renato Ribeiro')).toBe('RR'));
```

### Error Message & Stack Trace

Here's a print (Sorry, I have a problem with my term that I can't copy text :cry:)

![image](https://cloud.githubusercontent.com/assets/3277185/23932815/f85670e0-0918-11e7-8f19-27ef8ba966e1.png)

### Config

package.json
```json
{
  "...": "...",
  "ava": {
    "babel": "inherit"
  }
}
```

.babelrc
```json
{
  "presets": [
    "next/babel"
  ],
  "plugins": [
    "transform-export-extensions"
  ]
}
``` 

Note: also tested with `"require": ["babel-register"]` in ava configuration

### Command-Line Arguments

```
ava
```

## Relevant Links

Repo: https://github.com/renatorib/next-jsonplaceholder
Initials test commit: https://github.com/renatorib/next-jsonplaceholder/commit/f5602a1f77163ca9a0fdac4f9bd76ccf10955fcd

## Environment

```
$ ava --version
0.18.2

$ node --version
v7.5.0 
```

novemberborn:
> (probably due to non babel compiling issue)

Yes.

You've disabled AVA's default transpilation of test files. It's now up to you to make sure your code runs in Node.js. You'll have to make sure to transpile the module syntax in your own Babel config.

renatorib:
> You've disabled AVA's default transpilation of test files.
Where?

My custom babel config support `import/export` and latest es versions.

---

When I remove this config:

``` 
"ava": {
  "babel": "inherit"
}
```

This error persists.

novemberborn:
> And when I remove this config:
> 
> "ava": {
>   "babel": "inherit"
> }
> This error persists.

I actually get this error, which seems to be some Flow syntax that isn't removed:

```
❯ npm t

> Placeholder@ test /private/var/folders/_6/p8qxp_3n62zg9081tvb0lcc80000gn/T/tmp.zWXKadVBb1/next-jsonplaceholder
> ava

[...]

PR #1697, opened 2018-02-11
Make source compilation clearer

mdvorscak:
Following discussion of #1309, this PR attempts to clarify the documentation to avoid confusion when first setting up AVA with ESNext source files.


novemberborn:
Thanks @mdvorscak!

PR #1699, opened 2018-02-12
Update links to French docs translations

forresst:
Add link to es-module in french


novemberborn:
Fantastic as always @forresst.

PR #1700, opened 2018-02-12
Allow worker CLI flags to be specified

### LINKED ISSUE #1393: Remove custom CLI flags in child worker before requiring test file

novemberborn:
This removes AVA's internal flags, set by `fork.js` and consumed by `test-worker.js`. Fixes #1393.

AVA now forwards arguments, provided after an `--` argument terminator, to the worker process. Arguments are available from `process.argv[2]` onwards.

---

I've also removed some `pkg-conf` tests. These were accessing the options passed to the worker, which isn't great. They didn't seem terribly useful either, though I haven't confirmed whether they're replicating other tests. That said they don't cover explicit logic either.


sindresorhus:
Yay! I've wanted this many times.

PR #1711, opened 2018-02-14
Document context support in before & after hooks

### LINKED ISSUE #1579: Allow context in before hook

novemberborn:
Fixes #1579.

Heads up @kugtong33.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -602,7 +602,7 @@ Remember that AVA runs each test file in its own process. You may not have to cl
 
 #### Test context
 
-The `.beforeEach()` & `.afterEach()` hooks can share context with the test:
+Hooks can share context with the test:
 
 ```js
 test.beforeEach(t => {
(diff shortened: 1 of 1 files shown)


kugtong33:
:+1: , I was heading on creating a PR for this, should I reference it still? :D

novemberborn:
@kugtong33 looks like I beat you to it, sorry. Wanted to get this in so I can do a new release. Let me know if there's anything you'd like to see changed in my version though.

kugtong33:
@novemberborn , I think the important part is to determine what values are shared/referenced and what values are independent in each test, and better examples would be fine, as I have added on my side we can give examples for referenced values and primitive values and how they can use contexts correctly.

We can also create a recipe for it, migrating declared variables outside of hooks into using contexts. On my side, I am heavily using `supertest` and I would love to try to use contexts on passing `supertest` responses from hooks to tests.

novemberborn:
@kugtong33 a recipe would be good. It's too much to capture in the README itself.

PR #1713, opened 2018-02-14
Do not clobber forked process color level.

### LINKED ISSUE #1701: Ensure workers inherit color level from main process
Workers don't inherit the correct color level from the main process. @nowells raised this in #1455. This got entangled with a solution for #1393, by @kevva in #1401. I'm now fixing that with #1700.

As of #1700 workers remove AVA's internal flags from `process.argv`. We could apply the solution from #1455 so workers correctly inherit the color level.

I'm concerned that `chalk` / `supports-color` will behave inconsistently in the test processes. If the user code shares its `chalk` / `supports-color` dependencies with AVA, they'll inherit the color level from the main process. Otherwise the color level is detected afresh. I don't think AVA should foist the `--color` / `--no-color` flags onto the worker process. We could solve this inconsistency by bundling `chalk`. That way we can be sure to only configure *our* copy and leave user code unaffected. Users can still manipulate color levels in their test processes through environment variables or additional command line arguments.

@nowells, @kevva, @sindresorhus, what do you think?

jamiebuilds:
Cherry picked changes from #1455 to resolve #1701

---

Chalk uses `supports-color` to determine what color level should be
rendered. However, currently our parent `ava` process uses that to
colorize its output, however, when we fork our test processes we clobber
the color configuration that `supports-color` uses to determine the
level (see https://github.com/chalk/supports-color/blob/master/index.js#L37-L42)

This PR carries the current color level through to the forked processes
so if they log output that is colorized it will remain so with the same
fidelity of colors as the parent ava process.

### DIFF
--- a/lib/fork.js
+++ b/lib/fork.js
@@ -4,7 +4,16 @@ const path = require('path');
 const fs = require('fs');
 const Promise = require('bluebird');
 const debug = require('debug')('ava');
+const supportsColor = require('supports-color');
 const AvaError = require('./ava-error');
+
+const colorLevels = {
+	0: 'false',
+	1: 'true',
+	2: '256',
+	3: '16m'
+};
+const currentColorLevel = colorLevels[supportsColor.level];
 
 if (fs.realpathSync(__filename) !== __filename) {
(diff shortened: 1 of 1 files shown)


novemberborn:
Thanks @jamiebuilds. What do you think about bundling `chalk` and `supports-color` with AVA? That's the contention in #1701 that still needs to be resolved. (Please respond there instead.)

novemberborn:
We should make some more changes, see https://github.com/avajs/ava/issues/1701#issuecomment-377975554. I still need to land #1722 though.

novemberborn:
Closing due to inactivity. There's been changes in how AVA uses `chalk`, so we need a slightly different approach. See https://github.com/avajs/ava/issues/1701#issuecomment-377975554.

PR #1716, opened 2018-02-17
Cast test context to correct type.

### LINKED ISSUE #1723: Typescript typing context in strict generates error
### Test Source

```js
import anyTest, {TestInterface} from 'ava';
const test: TestInterface<{foo: 'bar'}> = anyTest;
```

```
$ $(npm bin)/tsc --strict testava.ts 
node_modules/ava/index.d.ts(62,73): error TS2304: Cannot find name 'Iterator'.
testava.ts(2,7): error TS2322: Type 'TestInterface<{}>' is not assignable to type 'TestInterface<{ foo: "bar"; }>'.
  Type '{}' is not assignable to type '{ foo: "bar"; }'.
    Property 'foo' is missing in type '{}'.```
```
### Config
typescript@2.7.2
ava@1.0.0-beta.3

samuelli:
When using `ava@1.0.0-beta.3` with `typescript@2.7.1`, the provided documentation fails to compile properly with an error that the type is not assignable. eg.
```
error TS2322: Type 'TestInterface<{}>' is not assignable to type 'TestInterface<{ foo: 'bar'; }>'.
  Type '{}' is not assignable to type '{ foo: 'bar'; }'.
    Property 'server' is missing in type '{}'.
```

This change casts the test object from Ava to the correctly paramaterized `TestInterface`.

Closes #1723


novemberborn:
I don't think there's a difference between assigning to an explicit type and casting with `as`. The documented code samples compile for me.

Could you share more of your actual code? Perhaps there's something else awry.

samuelli:
@novemberborn I filed #1723 with more details and a minimal repro. It appears to only be an issues when using `strict` mode in typescript.

novemberborn:
@samuelli right you are. I've added a regression test, and we're now running TypeScript in strict mode. See my commits. Does it look good to you?

samuelli:
That looks good @novemberborn! Thanks!

PR #1719, opened 2018-02-18
Add comments for assert and test interface definitions

### LINKED ISSUE #1686: Restore embedded documentation in TS and Flow definitions
I completely refactored the type definitions in #1670 without retaining the embedded documentation. Sorry about that 😞 

It'd be great to restore and even improve the embedded documentation. The `index.d.ts` and `index.js.flow` files are quite similar to each other, so work should start on one and then by copied to the other. Screen grabs of how the documentation is displayed in code editors would be useful so we can review how much documentation should be included.

kugtong33:
Fixes #1686 

Re-add deleted comments on the assert definition methods and added comments on the test interface definitions.

We can discuss the remaining definitions as I am not sure how to proceed from here.


index.d.ts
```diff
@@ -17,33 +17,61 @@ export interface SnapshotOptions {
 	id?: string;
 }
 
+// When an assert fail, the message string will be shown in the error
```
novemberborn: Would this ever be shown to the user? I agree that we shouldn't document it for each assertion though.

index.d.ts
```diff
@@ -17,33 +17,61 @@ export interface SnapshotOptions {
...
 export interface Assertions {
+	// Assert that actual is deep equal to expected
 	deepEqual<ValueType = any>(actual: ValueType, expected: ValueType, message?: string): void;
+	// Failing assertion
```
novemberborn: Perhaps "Cause the test to fail"?

index.d.ts
novemberborn: Would this ever be shown to the user? I agree that we shouldn't document it for each assertion though.

index.d.ts
```diff
@@ -17,33 +17,61 @@ export interface SnapshotOptions {
...
 	fail(message?: string): void;
+	// Assert that actual is equal to false
 	false(actual: any, message?: string): void;
+	// Assert that actual is falsy (false, 0, '', "", null, undefined, NaN)
```
novemberborn: No need to include `""` in this list.

index.d.ts
```diff
@@ -17,33 +17,61 @@ export interface SnapshotOptions {
...
 	falsy(actual: any, message?: string): void;
+	// Assert that error is falsy
 	ifError(error: any, message?: string): void;
+	// Assert that actual is equal to expected (use deepEqual for objects/arrays)
```
novemberborn: Perhaps "strictly equal". Using this with objects or arrays can be a valid use case so the reference to `deepEqual` seems distracting. There's also edge cases with negative zero and NaN but that's perhaps too detailed for inline documentation 😉

index.d.ts
```diff
@@ -17,33 +17,61 @@ export interface SnapshotOptions {
...
 	is<ValueType = any>(actual: ValueType, expected: ValueType, message?: string): void;
+	// Assert that actual is not equal to expected (use notDeepEqual for objects/arrays)
 	not<ValueType = any>(actual: ValueType, expected: ValueType, message?: string): void;
+	// Assert that actual is not deep equal to expected
 	notDeepEqual<ValueType = any>(actual: ValueType, expected: ValueType, message?: string): void;
+	// Assert that string does not follow the regex pattern
```
novemberborn: "match" rather than "follow"

index.d.ts
```diff
@@ -17,33 +17,61 @@ export interface SnapshotOptions {
...
 	notRegex(string: string, regex: RegExp, message?: string): void;
+	// Assert that a function that never returns does not throw an error
```
novemberborn: The "never returning" part is a type detail that's not relevant to the documentation. It's so that you can write `t.notThrows(() => { throw new Error() })` without type errors.

index.d.ts
```diff
@@ -17,33 +17,61 @@ export interface SnapshotOptions {
...
 	notThrows(value: () => never, message?: string): void;
+	// Assert that a function that returns an observable does not throw an error
 	notThrows(value: () => ObservableLike, message?: string): Promise<void>;
+	// Assert that a function that returns a promise does not throw an error
 	notThrows(value: () => PromiseLike<any>, message?: string): Promise<void>;
+	// Assert that a function that returns any value does not throw an error
```
novemberborn: Similarly here, returning "any value" is besides the point. Just "Assert that a function does not throw." is sufficient.

index.d.ts
```diff
@@ -78,15 +106,25 @@ export interface TestInterface<Context = {}> {
 	(title: string, macro: Macro<Context> | Macro<Context>[], ...args: Array<any>): void;
 	(macro: Macro<Context> | Macro<Context>[], ...args: Array<any>): void;
 
+	// Test hook that runs after all tests are done
```
novemberborn: Try and describe what the modifier *does*, rather than what it *is*. E.g. here "Specify hooks that run once all tests have completed". Or below, "Skip this test."

index.d.ts
kugtong33: Using vscode __1.20.1__, this is not shown, but I think it is for the better to document the message parameter on each method definition

kugtong33:
jsdoc-embedded

@novemberborn , apologies if I didn't got it right away (that it needs to be shown in the editor somehow), I think it is better if we use jsdoc

novemberborn:
I don't know… it's a bother to write and not all that readable in the tooltips. Given that the method signature is shown, referring to the variables should be sufficient IMHO. Personally I'd love to see a succinct summary of what the assertion does. Specifics can be found in the online documentation.

With the `deepEqual` example, `actual` doesn't have to be an object. It can be any value. It seems pointless to spell that out, and in any case the method signature shows it. Moreover, it makes it clear that `actual` and `expected` should have the *same* type. The `message` parameter is really hard to succinctly summarize, and I'd argue it's not even necessary to do so (except for `t.throws()` where it can be misleading).

kugtong33:
updated-typedef-doc
updated-typedef-test-doc

@novemberborn , my second take on it, missing documentation on `t.log`, `t.skip`, `t.context` and among other things, you can now add your edits :+1:

novemberborn:
Thanks @kugtong33. I've pushed an updated TS definition. I may have gone overboard a little…

> * Refactor the `Assertion` interface so the `.skip()` modifier can be added to each assertion.
> * Refactor the `ExecutionContext` interface to remove top-level `.skip` and allow `.log()` and `.plan()` to be skipped.
> * Add error value to `.end()` in `CbExecutionContext`.
> * Allow plain implementations to be passed to hooks without a title
> * Document all assertions, execution context methods and hook & test declarations and modifiers.
> * Explode the `t.throws()` definitions to give better information depending on how the assertion is used.
> * Link to documentation where appropriate.

What do you think? I like how, when writing tests, the documentation tells you what the functions do. I suppose it reads a bit different from other API docs but I think it's more useful for AVA.

How did you do your screen recordings?

I haven't updated the Flow definition yet, and the changes I made mean it's a bit trickier to copy things across. I feel I can barely ask if you'd be up for doing that… but are you? 😄

kugtong33:
@novemberborn , absolutely yes, I will look into the new updates and implement it on the flow definitions

I am using [peek](https://github.com/phw/peek) for the screen recordings

kugtong33:
@novemberborn 

Can I restructure the flow definitions, that it would closely resemble the typescript definitions?

novemberborn:
@kugtong33 that would be ideal. It makes it easier to update both files when we add features, since you can diff them against each other and see where you need to make changes.

kugtong33:
@novemberborn 

Restructured flow definition with what I understood, certain definitions that I directly copied from the ts file doesn't work with flow, but I tried it to be as close as possible.

Will add screen captures tomorrow

novemberborn:
Wow great work @kugtong33! I've previously had trouble getting Flow to work well for me. If you can see the right descriptions then I'm happy. I did push one minor tweak.

> I see that tests for flow and typescript definitions are only done when there is a need to regress, what if we add dedicated tests, is it possible?

That'd be great to have. It's just a lot of use cases to write 😄

novemberborn:
> When you select from the existing methods
> 
> …
> 
> When you write the method name with parenthesis

That difference is fine I think. I tried `test.serial.after.always(() => {})` in VS Code and it showed the right information when hovering each part of that chain. I'm going to land this, thanks @kugtong33!

PR #1721, opened 2018-02-19
Remove the `t.ifError` assert method

### LINKED ISSUE #1720: Remove ifError assertion?
Should we remove the `t.ifError()` assertion? It's functionally equivalent to `t.falsy()`. I suppose it's meant to be a "fail test if this variable contains an error" kind of assertion:

```js
test.cb('look no error', t => {
  obj.asyncWithCallback((err, result) => {
    t.ifError(err); // Guard against there being an error
    t.is(result.value, 'bob'); // Now access `result`
    t.end(); // Aaaand done!
  });
});
```

This is odd though. `t.end()` can be passed an error, and besides we don't stop running the test implementation when an assertion fails so it doesn't do anything useful.

@avajs/core?

kugtong33:
Fixes #1720 

Remove `ifError()` method , remove `t.ifError` documentation on readme and update affected tests

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -22,7 +22,6 @@ export interface Assertions {
 	fail(message?: string): void;
 	false(actual: any, message?: string): void;
 	falsy(actual: any, message?: string): void;
-	ifError(error: any, message?: string): void;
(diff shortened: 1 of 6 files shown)


novemberborn:
Fantastic, thanks @kugtong33!

PR #1729, opened 2018-02-22
Update readme.md to include running init with npx

martypdx:
Added instructions for running `ava --init` with `npx` when installing ava locally and at least `npm@0.5.2`

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -100,7 +100,7 @@ Your `package.json` will then look like this:
 }
 ```
 
-NOTE: Running `npx` requires [`npm` version `5.2.0`](https://github.com/npm/npm/releases/tag/v5.2.0) or greater. Otherwise, you'll have to manually configure the `test` script in your `package.json` to use `ava` (see above).
+NOTE: Running `npx` requires [`npm@5.2.0`](https://github.com/npm/npm/releases/tag/v5.2.0) or greater. Otherwise, you'll have to manually configure the `test` script in your `package.json` to use `ava` (see above).
 
 ### Create your test file
 
--- a/readme.md
+++ b/readme.md
@@ -79,7 +79,7 @@ $ npm install ava@next --save-dev
 $ npx ava --init
 ```
 
-If you prefer using `yarn`:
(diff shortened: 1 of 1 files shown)


novemberborn:
Thanks for the PR @martypdx. Do you think we Should generally advocate for `npx ava`? Both for `--init` and just plain running tests? And remove the global install advise? @sindresorhus?

martypdx:
My .02:

Yes, for `--init`, no for running tests. 

For required project dev dependency I would advocate purely installing locally. I only `npm -g` for utilities and tools that are _optionally_ used in a project.

Since `--init` is only run once, it makes sense to use `npx ava --init`.

For ongoing scripts (test running) I would advocate creating `script` `test` because it's needed by CI and as cli flags getting added beyond `ava` it provides a documented, standard way you know how to test the project.

novemberborn:
Right. I think if we promote `npx ava --init` then it wouldn't make sense to install globally.

I agree that normally you should run tests through `npm t`, but often when developing I reach for `npx ava`. 

I think that means we no longer have to advocate installing AVA globally.

martypdx:
@novemberborn updated the PR to just provide local install instructions. I don't use `yarn`, so I think from research that is right way to run `--init`, but haven't tried it myself

readme.md
```diff
@@ -101,23 +100,8 @@ Your `package.json` will then look like this:
...
-```
-
-Alternatively using Yarn:
-
-```console
-$ yarn add --dev ava@next
-```
-
-You'll have to configure the `test` script in your `package.json` to use `ava` (see above).
+NOTE: Running `npx` requires `npm` version `0.5.2` or greater. Otherwise, you'll have to
```
novemberborn: It's `5.2.0`, not `0.5.2` 😉 Perhaps add a link to https://github.com/npm/npm/releases/tag/v5.2.0?

sindresorhus:
👍

martypdx:
Updated to correct npm version with link to that update that outlines `npx`. Also made references to `package.json` and commands consistently code formatted

novemberborn:
Thanks @martypdx!

PR #1734, opened 2018-03-02
Updating recipes #2

BusbyActual:
Progress #1662

- [x] [Precompiling source files with webpack](https://github.com/avajs/ava/blob/master/docs/recipes/precompiling-with-webpack.md)

The files weren't too different so this could be a non-issue. I did see an outdated link and polished that up.

### DIFF
--- a/docs/recipes/precompiling-with-webpack.md
+++ b/docs/recipes/precompiling-with-webpack.md
@@ -95,6 +95,7 @@ module.exports = {
 				use: {
 					loader: 'babel-loader',
 					options: {
+            presets: ['env'],
 						cacheDirectory: true
 					}
 				}
@@ -181,6 +182,7 @@ module.exports = {
 				use: {
 					loader: 'babel-loader',
 					options: {
+            presets: ['env'],
 						cacheDirectory: true
 					}
 				}
--- a/docs/recipes/precompiling-with-webpack.md
+++ b/docs/recipes/precompiling-with-webpack.md
@@ -31,7 +31,6 @@ module.exports = {
 				test: /\.(js|jsx)$/,
 				use: 'babel-loader',
 				options: {
-					presets: ['@babel/preset-env'],
 					cacheDirectory: true
 				}
 			}
@@ -94,7 +93,6 @@ module.exports = {
 				use: {
 					loader: 'babel-loader',
 					options: {
-						presets: ['@babel/preset-env'],
 						cacheDirectory: true
 					}
 				}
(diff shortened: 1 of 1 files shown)


docs/recipes/precompiling-with-webpack.md
```diff
@@ -49,7 +48,7 @@ You can now run `$ ava _build/test.js` to run the tests contained in this output
 
 ### Multiple test files
 
-Things are a little more complicated with multiple test files. We recommend [using babel-register](babelrc.md) until the performance penalty becomes too great.
+Things are a little more complicated with multiple test files. We recommend [using babel-register](babel.md) until the performance penalty becomes too great.
```
novemberborn: This should be `@babel/register`. It'd be neat to link to the [right anchor](https://github.com/avajs/ava/blob/master/docs/recipes/babel.md#compile-sources).

docs/recipes/precompiling-with-webpack.md
```diff
@@ -180,9 +181,11 @@ module.exports = {
 		rules: [
 			{
 				test: /\.(js|jsx)$/,
+        exclude: /(node_modules|bower_components)/,
 				use: {
 					loader: 'babel-loader',
```
novemberborn: Does this need changing?

docs/recipes/precompiling-with-webpack.md
```diff
@@ -1,6 +1,3 @@
-> **Please note, this recipe has not yet been updated for Babel 7 support in AVA 1.0.**
-
----
```
novemberborn: This line ought to be removed as well.

docs/recipes/precompiling-with-webpack.md
BusbyActual: It's not 'necessary' it was in the examples in babels docs. I can remove it for brevity.

BusbyActual:
Made requested changes. The commit spacing looks strange however in my local it's fine?

image

novemberborn:
> The commit spacing looks strange however in my local it's fine?

Your lines use spaces, but the rest of them use tabs. If you configure your editor to show invisible characters you'll be able to see the difference.

I've pushed some tweaks. Why did you add the `env` preset example? Initially I upgraded it to the recommended preset syntax from Babel 7, but then I realized that might depend on the user installing the beta version of `babel-loader`. So instead I think it's better to not mention any of that.

BusbyActual:
I think I didn't see babel-loader was in betal at that version.

PR #1738, opened 2018-03-07
Replace --init flag with npx recommendation

### LINKED ISSUE #1737: Remove ava-init and recommend use of npx
## Description

Now that `npx` presents a great solution for running one-off chore commands, I want to propose going a step farther from d3bade8cd7fac004c1c6d2eff9b3f498257baa60 and removing the `--init` command (& associated dependency) entirely, and recommend running `npx ava-init` for new projects.

This will make AVA installs a little smaller, and I think it's a better separation of concerns, because `--init` isn't useful past the initial setup. For bonus points, the command is two characters shorter than `npx ava --init` :)

You could even have `ava-init` install `ava` if it isn't already present.

Happy to PR this if you like the idea.

billyjanitsch:
Fixes #1737.

This removes the `--init` flag and associated dependency on `ava-init` in favor of recommending `npx ava-init`. See the linked issue for reasoning.

Happy to update this to `@ava/init` if the package is republished there. 😊

(Does yarn have an npx equivalent? Maybe it doesn't matter, since Yarn users presumably also have npx installed.)

### DIFF
--- a/docs/recipes/debugging-with-webstorm.md
+++ b/docs/recipes/debugging-with-webstorm.md
@@ -19,7 +19,7 @@ Save the configuration.
 
 ## Setup using npm
 
-Execute `ava --init` in your project directory to add AVA to your `package.json`.
+Execute `npx ava-init` in your project directory to add AVA to your `package.json`.
 
 Your `package.json` will look something like this:
 
@@ -46,7 +46,7 @@ Your IDE will then execute `npm run test` and thus call `node_modules/.bin/ava` 
 
 In the `Node parameters`, for Node.js 7+ pass `--inspect-brk` or `--debug-brk` for earlier versions.
 
-Don't forget to select a Node.js interpreter. 
+Don't forget to select a Node.js interpreter.
 
 Save the configuration.
 
--- a/docs/recipes/debugging-with-webstorm.md
+++ b/docs/recipes/debugging-with-webstorm.md
@@ -19,7 +19,7 @@ Save the configuration.
 
 ## Setup using npm
 
-Execute `npx ava-init` in your project directory to add AVA to your `package.json`.
+Execute `npx @ava/init` in your project directory to add AVA to your `package.json`.
 
 Your `package.json` will look something like this:
 
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -30,7 +30,6 @@ exports.run = () => {
 		  ava [<file|directory|glob> ...]
 
 		Options
-		  --init                  Add AVA to your project
 		  --watch, -w             Re-run tests when tests and source files change
 		  --match, -m             Only run tests with matching title (Can be repeated)
 		  --update-snapshots, -u  Update snapshots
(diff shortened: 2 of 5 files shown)


novemberborn:
> Happy to update this to @ava/init if the package is republished there. 😊

If you could do the PR to https://github.com/avajs/ava-init to change the name and anything else that might need changing, I can publish it 👍 

This is advanced but I wonder if `npx @ava/init` should offer to install either `latest` or `next`? I think as your PR is currently written we make it really hard to install the beta releases.

> (Does yarn have an npx equivalent? Maybe it doesn't matter, since Yarn users presumably also have npx installed.)

I'm happy with that assumption. [`npx` is open source](https://www.npmjs.com/package/libnpx) so Yarn can always adopt it 😉

billyjanitsch:
>If you could do the PR

https://github.com/avajs/ava-init/pull/8 :)

>This is advanced but I wonder if npx @ava/init should offer to install either latest or next? I think as your PR is currently written we make it really hard to install the beta releases.

Personally I don't think this is necessary. Upgrading to `next` is easy enough: `npx @ava/init && npm i -D ava@next`.

If you do want the feature, I think `latest` should remain the default, but I could add a `--next` flag to the `@ava/init` CLI.

novemberborn:
> If you do want the feature, I think latest should remain the default, but I could add a --next flag to the @ava/init CLI.

That sounds good.

billyjanitsch:
Ok, can I do it in another PR? The existing one is already pretty big.

novemberborn:
> Personally I don't think this is necessary. Upgrading to next is easy enough: npx @ava/init && npm i -D ava@next.
> 
> If you do want the feature, I think latest should remain the default, but I could add a --next flag to the @ava/init CLI.

To give you a longer response, it's a little awkward but I don't know how long we'll have to be in beta (it really depends on Babel). I'd like to avoid too many hoops so people get the beta releases.

> Ok, can I do it in another PR? The existing one is already pretty big.

Just landed the first one so go ahead :) Let me know if you need a beta release published for testing.

billyjanitsch:
>To give you a longer response

No worries, I understand! PR's up: https://github.com/avajs/ava-init/pull/9

billyjanitsch:
Updated this PR to reflect the new name. How do you feel about the instructions defaulting to `--next`? Seems more appropriate since the rest of the readme reflects the beta release.

novemberborn:
> How do you feel about the instructions defaulting to --next? Seems more appropriate since the rest of the readme reflects the beta release.

Yea that's perfect.

What are your thoughts on advocating purely `npx @ava/init`? I suppose folks may figure out they can also do `npm i -D ava@next` (and the Yarn equivalent). I wonder if we should have an explicit "Manual installation" section? Similarly perhaps we should point out that `npx @ava/init` works with Yarn?

novemberborn:
By the way I've published `@ava/init` and deprecated all `ava-init` releases. Thank you for pushing this @billyjanitsch!

billyjanitsch:
>What are your thoughts on advocating purely `npx @ava/init`? I wonder if we should have an explicit "Manual installation" section? Similarly perhaps we should point out that `npx @ava/init` works with Yarn?

Good points -- I especially think a "manual installation" note is important for the sake of being more transparent about what `@ava/init` is doing. Added to the PR.

novemberborn:
> Not sure what's going on with CI. I can try rebasing?

We've got some tests that flake out. I've restarted the specific jobs in Travis.

billyjanitsch:
Thanks! I think this is good to go, then, unless you want any other changes.

PR #1741, opened 2018-03-10
Upgrade std/esm

jdalton:
👋 Hi All,

I've upgraded `@std/esm` to `esm`. While the package is published and ready I'm presenting this PR so changes can land before or along with the release blog next week.

With the upgrade to `esm` CJS interoperability is enabled by default (zero configuration needed).


lib/test-worker.js
```diff
@@ -234,7 +234,7 @@ adapter.installPrecompilerHook();
 
 try {
 	(opts.require || []).forEach(x => {
-		if (/[/\\]@std[/\\]esm[/\\]index\.js$/.test(x)) {
+		if (/[/\\](@std[/\\])?esm[/\\]index\.js$/.test(x)) {
```
novemberborn: Given that `x` here is an absolute path I'm a little nervous about just looking for `/esm/`. That seems like it could yield false positives more often than `@std/esm` did.

It's probably fine if we'd check for `node_modules` as well. Alternatively perhaps `esm` could tag its exported function with a registered symbol that we can check for?

```js
const required = require(x)
try {
  if (required[Symbol.for('esm')]) {
    require = required(module)
  }
} catch (_) {}
```

Do you think we should drop support for `@std` here?

.gitignore
```diff
@@ -1,4 +1,5 @@
 node_modules
+.DS_Store
```
novemberborn: Could you remove this addition? I don't think we've seen this file in PRs at all so it doesn't seem necessary.

jdalton: I just had one slip into this PR (which is why I rebased and added it). It's a meta data on MacOS and it's generated automatically when you view a directory. I was surprised you didn't have it.

lib/test-worker.js
jdalton: Ah, yes I can make that more detectable!

jdalton: We have a symbol namespace we use currently (though not exposed on the exports).
It goes something like `"esm" + ZWJ + ":thing"` so for the exports I could expose a symbol named for `"esm"+ZWJ`, so without the `:thing` bit.

jdalton: I'll noodle around on it and report back.

.gitignore
novemberborn: Personally I have it in `~/.gitignore`. I'd just rather not balloon `.gitignore` with OS level stuff.

jdalton: Coo cool, no prob. I'll remove it.

sindresorhus: Only things related to the project should be in the project gitignore. The rest should be in your own global gitignore. 

https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

jdalton:
Ok updated.

jdalton:
Anything else I need to do?

novemberborn:
Thanks for adding the symbol @jdalton!

jdalton:
No problem. Let me know if you ever need anything else.

git-jiby-me:
@novemberborn any idea when is there gonna be a next release with this changes ?

novemberborn:
@git-jiby-me I'm hoping to get a release out this weekend.

Issue #1730, opened 2018-02-24
Recipe detailing context in per-file and per-test hooks


novemberborn:
Via https://github.com/avajs/ava/pull/1711#issuecomment-365646530 — it'd be great to document how `t.context` works in `.before()` / `.after()` and `.beforeEach()` / `.afterEach()` hooks. Examples should illustrate when context is copied and how to share references and values.

Seiyial:
Hi, I've recently used `.before()` / `.after()` and `.beforeEach()` / `.afterEach()` hooks in my project and would love to help with this. I'm rather new to open source contributing so any feedback/advice would be appreciated. Thanks!

novemberborn:
@Seiyial that's great! Have a look at some of [our other recipes](https://github.com/avajs/ava/tree/master/docs/recipes) to see what we're looking for. Some of the more recently updates ones are:

* https://github.com/avajs/ava/blob/master/docs/recipes/babel.md
* https://github.com/avajs/ava/blob/master/docs/recipes/flow.md
* https://github.com/avajs/ava/blob/master/docs/recipes/test-setup.md
* https://github.com/avajs/ava/blob/master/docs/recipes/typescript.md

Fork the repository, create a new Markdown file in the `docs/recipes` and go from your experiences. Then do a PR and we can help with any improvements.

Seiyial:
@novemberborn Sorry for the delay. Please let me know how you think about it, thanks!

mh81:
I have some free time this weekend, I'd be happy to build on @Seiyial's work in his PR and update it with @novemberborn's PR feedback.  Any objection?

novemberborn:
Go for it @mh81. If you could retain @Seiyial's commits that'd be great.

PR #1744, opened 2018-03-12
Add recipe docs for t.context (#1730)

Seiyial:
Recipe detailing context in per-file and per-test hooks #1730


novemberborn:
Hi @Seiyial, thanks for your PR. I've been busy lately so it may take a bit longer before I can properly look at this.

Seiyial:
Hi @novemberborn , sorry I missed the notification.

By assertions do you mean `t.deepEqual()` etc? I personally thought of using `console.log()` because the readers may not yet be familiarised with ava's syntax. But come to think of it, the assert syntax looks quite foolproof so I guess using assert statements might be better.

Oh I see, that sounds important. I'll guess I should highlight that in too.

novemberborn:
> By assertions do you mean t.deepEqual() etc? I personally thought of using console.log() because the readers may not yet be familiarised with ava's syntax. But come to think of it, the assert syntax looks quite foolproof so I guess using assert statements might be better.

Yes. People will have to learn those assertions at some point, so I'd rather the examples are close to actual use cases rather than simplified.

novemberborn:
Hey @Seiyial are you still interested in working on this?

Seiyial:
Hi @novemberborn , sorry for the late reply again. I have been busy and will be busier in the coming weeks, so it may be considerable if someone else is free to do it faster. That being said, I've updated the recipe in `docs/recipes/context.md` based on your comments in this thread, including changing `console.log` to assertions and adding `test.afterEach.always()` to the diagram. The shallow cloning is also highlighted with examples.

Thanks for looking through it!

Just in case the diagram still needs editing and someone else is taking it, the raw version is here: [https://bit.ly/2IWsSbL](https://bit.ly/2IWsSbL)

sindresorhus:
@Seiyial Still interested in finishing this? :)

Seiyial:
Sure! Am freer now. Please let me know if it needs any more changes, thanks!

novemberborn:
Heya @Seiyial. I just reread the [Test context](https://github.com/avajs/ava#test-context) section in the README. I think it describes the behavior quite succinctly. Perhaps the recipe could start by repeating that section?

Then for the examples, we should have one example that modifies of `t.context` properties, and a second example that modifies a nested object (which isn't cloned). Both examples should use all hooks.

What do you think?

sindresorhus:
Ping @Seiyial :)

Seiyial:
@sindresorhus so sorry I got busy halfway! But as discussed at #1730 @mh81 has picked this up and is working on it now :)

novemberborn:
Thanks @Seiyial. I'm closing this issue for housekeeping purposes.

PR #1745, opened 2018-03-18
Update dependencies & Fix missing imports

clemtrek:

### DIFF
--- a/bench/run.js
+++ b/bench/run.js
@@ -2,6 +2,7 @@
 const childProcess = require('child_process');
 const path = require('path');
 const fs = require('fs');
+const Promise = require('bluebird');
 const arrify = require('arrify');
 const makeDir = require('make-dir');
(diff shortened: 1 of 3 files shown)


bench/run.js
```diff
@@ -2,6 +2,7 @@
 const childProcess = require('child_process');
 const path = require('path');
 const fs = require('fs');
+const Promise = require('bluebird');
```
sindresorhus: Don't do unrelated changes.

sindresorhus:
The PR needs a more descriptive title.

bench/run.js
clemtrek: Shall I remove them and add them in another PR?

novemberborn:
`@ava/init` will land in https://github.com/avajs/ava/pull/1738. I've removed the deprecation message from npm since it's not actually useful for anybody installing AVA.

Before you make a new PR for the promise change, could you elaborate on why it's necessary?

PR #1746, opened 2018-03-19
Make test file extensions configurable

### LINKED ISSUE #631: Allow tests files to have any extension (i.e. `.jsx`, `.ts`)

Jaden-Giordano:
Allows for the use of extensions like `jsx`, `ts`, etc..

Fixes https://github.com/avajs/ava/issues/631


Jaden-Giordano:
Referencing #631

Jaden-Giordano:
Gotcha, I'll make the updates and update this PR. Thanks for the filler!

Jaden-Giordano:
@novemberborn Looking at the code its a bit above my current knowledge with ava.

novemberborn:
@Jaden-Giordano no worries. I'm happy to give pointers if you want?

Jaden-Giordano:
There are currently errors I'm attempting to resolve. I keep getting `ResolveError: D:\Projects\OpenSource\ava-test\.babelrc: Couldn't find preset "@babel/core" relative to directory` when testing ava in a project.

Jaden-Giordano:
@novemberborn You think you can help me with that?

novemberborn:
Hi @Jaden-Giordano, thanks for continuing to work on this. I've been a bit busy lately so it may take a few more days before I can get back to you on this, sorry.

Jaden-Giordano:
No worries, I assumed so, thanks for being so helpful so far.

Jaden-Giordano:
### Update:
So I managed to move past those errors, but now the cli and api tests are not passing, Im assuming this is happening since I added two properties to the config (`extensions` and `babel.extensions`), I'll continue to look into what the root cause of these issues are, but if anyone is willing to help who is familiar with the api and cli files help/feedback would be greatly appreciated. 

### Next steps after fixes:
After this fix, we should be ready to move on to notifying the watch.

lib/babel-config.js
```diff
@@ -22,7 +22,7 @@ function validate(conf) {
 		return {testOptions: {}};
 	}
 
-	if (!conf || typeof conf !== 'object' || !conf.testOptions || typeof conf.testOptions !== 'object' || Array.isArray(conf.testOptions) || Object.keys(conf).length > 1) {
+	if (!conf || typeof conf !== 'object' || !conf.testOptions || typeof conf.testOptions !== 'object' || Array.isArray(conf.testOptions)) {
```
novemberborn: The `Object.keys()` bit was there to ensure only known keys were used. So we still need that, but it's gotten a bit trickier since more than 1 key is allowed.

It should be possible to provide `extensions` and not `testOptions`.

lib/caching-precompiler.js
```diff
@@ -30,7 +30,11 @@ class CachingPrecompiler {
 		this.transform = this._createTransform();
 	}
 
-	precompileFile(filePath) {
+	precompileFile(filePath, excludeExtensions) {
+		if (excludeExtensions.includes(path.extname(filePath))) {
```
novemberborn: I think this should be guarded against in the calling code.

lib/ava-files.js
```diff
@@ -115,9 +116,12 @@ class AvaFiles {
 		});
 
 		if (files.length === 0) {
-			files = defaultIncludePatterns();
+			files = defaultIncludePatterns(options.extensions || ['js']);
```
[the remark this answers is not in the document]
Navomi-Jaden: Thanks for this catch!

api.js
```diff
@@ -43,6 +43,12 @@ class Api extends EventEmitter {
 		const apiOptions = this.options;
 		runtimeOptions = runtimeOptions || {};
 
+		const doNotCompileExtensions = apiOptions.extensions || [];
+		const babelExtensions = apiOptions.babelConfig.extensions || [];
+
+		// Combine all extensions possible for testing.
+		const allExts = [...doNotCompileExtensions, ...babelExtensions];
```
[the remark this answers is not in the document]
Navomi-Jaden: Gotcha

lib/caching-precompiler.js
Navomi-Jaden: Good idea moving this to be checked here https://github.com/avajs/ava/blob/master/api.js#L134

Navomi-Jaden:
@novemberborn Thank you for the insight it helped immensely! That fixed most of the issues. It is working if I install from my git repo with this current branch and run it with `babel.extensions = ['jsx']` too. Maybe you have some idea what is going on in the CI? With that being said I'm going to go ahead and move on to informing the watcher.

Jaden-Giordano:
@novemberborn Thank you for the insight it helped immensely! That fixed most of the issues. It is working if I install from my git repo with this current branch and run it with babel.extensions = ['jsx'] too. Maybe you have some idea what is going on in the CI? With that being said I'm going to go ahead and move on to informing the watcher.

Jaden-Giordano:
Sounds good, take you're time.

novemberborn:
@Jaden-Giordano if you could merge in (or rebase on) master, that should fix some of the CI issues.

Jaden-Giordano:
I think I did the rebase correctly. Will be a couple days until I can finish up your recommendations and write those tests.

api.js
```diff
@@ -43,6 +43,12 @@ class Api extends EventEmitter {
 		const apiOptions = this.options;
 		runtimeOptions = runtimeOptions || {};
 
+		const doNotCompileExtensions = apiOptions.extensions || [];
+		const babelExtensions = apiOptions.babelConfig ? apiOptions.babelConfig.extensions || ['js'] : ['js'];
```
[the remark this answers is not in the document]
Jaden-Giordano: Oh alright I see.

api.js
```diff
@@ -43,6 +43,12 @@ class Api extends EventEmitter {
...
+
+		// Combine all extensions possible for testing.
+		const allExts = [...doNotCompileExtensions, ...babelExtensions];
```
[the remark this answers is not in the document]
Jaden-Giordano: For this im just using `concat` then `filter` to filter out any duplicate entries:
```js
const allExtensions = doNotCompileExtensions
    .concat(babelExtensions)
    .filter((ext, i, self) => self.indexOf(ext) === i);
```

lib/babel-config.js
```diff
@@ -13,19 +13,33 @@ const syntaxAsyncGeneratorsPath = require.resolve('@babel/plugin-syntax-async-ge
...
 const transformTestFilesPath = require.resolve('@ava/babel-preset-transform-test-files');
 
+function validateKeys(conf) {
+	const keys = Object.keys(conf);
+	for (let i = 0; i < keys.length; i++) {
+		if (keys[i] !== 'testOptions' && keys[i] !== 'extensions') {
+			return false;
+		}
+	}
+	return true;
```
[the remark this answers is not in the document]
Jaden-Giordano: So elegant!

lib/babel-config.js
```diff
@@ -13,19 +13,33 @@ const syntaxAsyncGeneratorsPath = require.resolve('@babel/plugin-syntax-async-ge
...
+		return {extensions: ['js'], testOptions: {}};
 	}
 
-	if (!conf || typeof conf !== 'object' || !conf.testOptions || typeof conf.testOptions !== 'object' || Array.isArray(conf.testOptions) || Object.keys(conf).length > 1) {
+	if (!conf || typeof conf !== 'object' || !conf.testOptions || typeof conf.testOptions !== 'object' || Array.isArray(conf.testOptions) || !validateKeys(conf)) {
 		throw new Error(`${colors.error(figures.cross)} Unexpected Babel configuration for AVA. See ${chalk.underline('https://github.com/avajs/ava/blob/master/docs/recipes/babel.md')} for allowed values.`);
 	}
 
+	if (conf.extensions === undefined) {
+		conf.extensions = ['js'];
```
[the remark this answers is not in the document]
Jaden-Giordano: I'm just going to strip away this code, this shouldn't even be needed with the recommendations you gave for `api.js`.

lib/babel-config.js
```diff
@@ -13,19 +13,33 @@ const syntaxAsyncGeneratorsPath = require.resolve('@babel/plugin-syntax-async-ge
...
 		return null;
 	}
 
 	if (conf === undefined) {
-		return {testOptions: {}};
+		return {extensions: ['js'], testOptions: {}};
 	}
 
-	if (!conf || typeof conf !== 'object' || !conf.testOptions || typeof conf.testOptions !== 'object' || Array.isArray(conf.testOptions) || Object.keys(conf).length > 1) {
+	if (!conf || typeof conf !== 'object' || !conf.testOptions || typeof conf.testOptions !== 'object' || Array.isArray(conf.testOptions) || !validateKeys(conf)) {
```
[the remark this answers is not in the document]
Jaden-Giordano: Instead of multiple `if` conditions we could use `invalid` instead then split them up by doing:
```js
let invalid = false;

invalid = !conf || typeof conf !== 'object' || ...;
invalid = ...

if (invalid) {
    throw ...
}
```
This should be a bit shorter than a bunch of if statements.

Jaden-Giordano:
I was having trouble figuring out what tests to write without writing a test that did the same thing as other tests, so I only wrote one test that tests if the tests successfully precompiles before being ran then passing if tests in the fixture ran successfully. The test I wanted to write to check if the tests run when neither extensions nor babelExtensions are defined (defaults to then not compile the test files and looks for js files) was basically the same as this: https://github.com/avajs/ava/blob/master/test/api.js#L1020.

api.js
```diff
@@ -43,6 +43,22 @@ class Api extends EventEmitter {
...
+			doNotCompileExtensions.push('js');
+		}
+
+		if (apiOptions.babelConfig && babelExtensions.length === 0) {
+			babelExtensions.push('js');
+		}
+
+		// Combine all extensions possible for testing. Removing duplicate extensions.
+		const allExtensions = doNotCompileExtensions
+			.concat(babelExtensions)
```
novemberborn: If you wouldn't mind, let's be fancy here and use `[...doNotCompileExtensions, ...babelExtensions]` 💯 ❣️

api.js
```diff
@@ -43,6 +43,22 @@ class Api extends EventEmitter {
...
+			.filter((ext, i, self) => self.indexOf(ext) === i);
```
novemberborn: Would this be too much? `[...new Set([...doNotCompileExtensions, ...babelExtensions])]` 😊 

The `Set` constructor will deduplicate, and we can then turn it back into an array using the spread syntax.

lib/babel-config.js
```diff
@@ -22,7 +26,12 @@ function validate(conf) {
 		return {testOptions: {}};
 	}
 
-	if (!conf || typeof conf !== 'object' || !conf.testOptions || typeof conf.testOptions !== 'object' || Array.isArray(conf.testOptions) || Object.keys(conf).length > 1) {
+	let invalid = false;
+
+	invalid = !conf || typeof conf !== 'object' || !conf.testOptions || typeof conf.testOptions !== 'object' || Array.isArray(conf.testOptions);
+	invalid = !validateKeys(conf) || (conf.extensions && !Array.isArray(conf.extensions));
```
novemberborn: This second line needs to start with `invalid = invalid || …`.

It'll be clearer to put the condition in an `if` clause and then explicitly assign `true`.

test/caching-precompiler.js
```diff
@@ -1,7 +1,7 @@
 'use strict';
 const fs = require('fs');
 const path = require('path');
-const test = require('tap').test;
+const tap = require('tap');
```
novemberborn: Could you revert this change?

Jaden-Giordano: oh I thought i reverted this, sorry about that, yeah.

lib/babel-config.js
Jaden-Giordano: I placed these into an if statement like you recommended

api.js
Jaden-Giordano: Oh i didn't know it would remove duplicates, that's pretty awesome actually (gotta change stuff in my personal projects).

Jaden-Giordano:
So the test would just end up precompiling and checking if the test passed but thats what the test I wrote already does so I just left that the same but added a test to ava files to look for files with a specified extension.

api.js
```diff
@@ -43,6 +43,20 @@ class Api extends EventEmitter {
...
+		if (!apiOptions.babelConfig && doNotCompileExtensions.length === 0) {
+			doNotCompileExtensions.push('js');
+		}
+
+		if (apiOptions.babelConfig && babelExtensions.length === 0) {
+			babelExtensions.push('js');
+		}
+
+		// Combine all extensions possible for testing. Removing duplicate extensions.
+		const allExtensions = [...new Set([...doNotCompileExtensions, ...babelExtensions])];
```
novemberborn: This just needs to ensure there is no overlap between both arrays. You could compare the length of `allExtensions` to the combined length of `doNotCompileExtensions` and `babelExtensions`: it should be equal.

We should probably include the ambiguous extensions in the error message though. Some applications of [`lodash.difference`](https://www.npmjs.com/package/lodash.difference) should do the trick.

Given that we're now throwing an error, could you move this logic into the constructor? Feel free to save `allExtensions` on the instance.

test/caching-precompiler.js
```diff
@@ -77,8 +77,8 @@ test('should reuse existing source maps', t => {
 	const tempDir = uniqueTempDir();
 	const precompiler = new CachingPrecompiler({path: tempDir, getBabelOptions, babelCacheKeys});
 
-	precompiler.precompileFile(fixture('es2015-source-maps.js'));
-	const options = transformSpy.lastCall.args[1];
+	precompiler.precompileFile(fixture('es2015-source-maps.js'), []);
+	const options = babel.transform.lastCall.args[1];
```
novemberborn: I suspect you may have had a merge conflict here? Use `transformSpy` rather than `babel.transform` in this line.

api.js
```diff
@@ -125,21 +139,25 @@ class Api extends EventEmitter {
...
-						});
+						return new AvaFiles({cwd: this.options.resolveTestsFrom, extensions: allExtensions})
+							.findTestHelpers().then(helpers => {
+								return {
+									cacheDir: precompilation.cacheDir,
+									map: files.concat(helpers).reduce((acc, file) => {
+										try {
+											const realpath = fs.realpathSync(file);
+											let hash = realpath;
+											if (!doNotCompileExtensions.includes(path.extname(realpath))) {
```
novemberborn: OK so the tricky thing is we still want to compile enhancements, unless that's *also* disabled. This currently does use Babel, but it won't apply AVA's regular presets. I think this means instantiating a second `precompilation` object, and doing a second build of Babel configuration.

I feel bad for not spotting this earlier, and it's a bit tricky to line all this up correctly. Let me know if you want me to push a commit for this instead.

Jaden-Giordano: @novemberborn If you could on this one I can handle the rest, work is a bit busy so I don't have alot of time to work on this. Thanks for all the help!

test/caching-precompiler.js
Jaden-Giordano: Oh alright gotcha

api.js
```diff
@@ -37,6 +38,27 @@ class Api extends Emittery {
...
+			babelExtensions.push('js');
+		}
+
+		this.options.extensions = doNotCompileExtensions;
+
+		// Combine all extensions possible for testing. Removing duplicate extensions.
+		this.allExtensions = [...new Set([...doNotCompileExtensions, ...babelExtensions])];
+
+		if (this.allExtensions.length !== doNotCompileExtensions.length + babelExtensions.length) {
+			const duplicates = difference([...doNotCompileExtensions, ...babelExtensions]);
```
novemberborn: You need to compare two arrays for `difference` to return anything. But then it actually gives the extensions that are not repeated, so:

```js
const notRepeated = difference(doNotCompileExtensions, babelExtensions);
const duplicates = this.allExtensions.filter(ext => !notRepeated.includes(ext));
```

I suppose this still needs a test?

readme.md
```diff
@@ -262,6 +262,7 @@ To ignore a file or directory, prefix the pattern with an `!` (exclamation mark)
 			"@babel/register"
 		],
 		"babel": {
+			"extensions": ["jsx"]
```
novemberborn: Missing comma.

readme.md
```diff
@@ -282,8 +283,10 @@ Arguments passed to the CLI will always take precedence over the configuration i
 - `tap`: if `true`, enables the [TAP reporter](#tap-reporter)
 - `snapshotDir`: specifies a fixed location for storing snapshot files. Use this if your snapshots are ending up in the wrong location
 - `compileEnhancements`: if `false`, disables [power-assert](https://github.com/power-assert-js/power-assert) — which otherwise helps provide more descriptive error messages — and detection of improper use of the `t.throws()` assertion
+- `extensions`: Extensions to opt of of precompilation.
```
novemberborn: This needs some elaboration. How about:

> Extensions of test files that are not precompiled using AVA's Babel presets. Note that files are still compiled to enable power-assert and other features, so you may also need to set `compileEnhancements` to `false` if your files are not valid JavaScript.

readme.md
```diff
@@ -282,8 +283,10 @@ Arguments passed to the CLI will always take precedence over the configuration i
...
 - `require`: extra modules to require before tests are run. Modules are required in the [worker processes](#process-isolation)
 - `babel`: test file specific Babel options. See our [Babel recipe] for more details
+- `babel.extensions`: Extensions to include in precompilation.
```
novemberborn: How about:

> Extensions of test files that will be precompiled using AVA's Babel presets.

novemberborn:
@Jaden-Giordano I haven't had a chance yet to look into this. I just realized we need some changes with Babel (https://github.com/avajs/ava/issues/1789) so I might try and land that before doing my work here.

Anyway this is super close so once that unblocks we should be good to go.

novemberborn:
@Jaden-Giordano just wanted to say I haven't forgotten about this 😄  #1789 is pretty much done so I'm hoping to land that next weekend. I'll then try and get this in as well before cutting a release.

novemberborn:
Phew! This is ready now, @Jaden-Giordano please have a look at the commits I just pushed.

I've updated the Babel and ES module recipes. I've fixed the top-level extensions / enhancements-only code path, and we can now handle extensions that have periods in them.

novemberborn:
I don't think `profile.js` script supports custom extensions, but I'll file a follow-up issue for that.

PR #1749, opened 2018-03-23
[WIP] Use babel-register's source-map-support

### LINKED ISSUE #974: ava overrides `source-map-support` of `babel-register`
If I transpile imported modules with `babel-register`, stack traces of the imported modules are not correctly reported because`ava` overrides `source-map-support` of `babel-register`.

This is due to the dependency mismatch between `ava` and `babel-register`. `babel-register` depends on `source-map-support` 0.2.10 and `ava` depends on the version 0.4.0, making the both use the different `source-map-support`s so one overriding `Error.prepareStackTrace`.

When I bump `source-map-support` in `babel-register` (babel/babel#3571), the issue goes away.
### Config

``` json
{
  "ava": {
    "require": [
      "babel-register"
    ],
    "babel": "inherit"
  ...
  }
}
```

rhendric:
If babel-register is used in the test process, AVA needs to use its
version of source-map-support. Multiple source-map-supports in use in
the same process don't work together, and can cause stack trace line
numbers to be incorrect.

Fixes #974.

---

Note: no automated tests for this yet. It might take me a while to figure out how to write a good one, but I figured I'd submit this as a work in progress to get early feedback.

### DIFF
--- a/lib/process-adapter.js
+++ b/lib/process-adapter.js
@@ -2,7 +2,6 @@
 const fs = require('fs');
 const path = require('path');
 const debug = require('debug')('ava');
-const sourceMapSupport = require('source-map-support');
 const installPrecompiler = require('require-precompiled');
 
 // Parse and re-emit AVA messages
@@ -74,6 +73,17 @@ const sourceMapCache = new Map();
 const cacheDir = opts.cacheDir;
 
 exports.installSourceMapSupport = () => {
+	// Workaround for https://github.com/avajs/ava/issues/974
+	const sourceMapSupport = (() => {
+		try {
+			/* eslint-disable import/no-extraneous-dependencies, import/no-unresolved */
+			if (require.resolve('babel-register') in require.cache) {
+				return require('babel-register/node_modules/source-map-support');
+			}
+			/* eslint-enable import/no-extraneous-dependencies, import/no-unresolved */
+		} catch (err) {}
+		return require('source-map-support');
+	})();
 	sourceMapSupport.install({
 		environment: 'node',
 		handleUncaughtExceptions: false,


novemberborn:
Thanks for the PR @rhendric!

AVA loads `source-map-support` before loading the `require` options, so we'll never be able to detect `babel-register` this way. There's also no guarantee of how `source-map-support` is laid out on disk. Ideally this is solved in `source-map-support`, but I admit I wouldn't know how. What do you think?

rhendric:
Huh. Looking at `lib/test-worker.js`, I do see that `installSourceMapSupport` is called before `opts.require` is handled. And yet, this fix works for my project—with vanilla master AVA, I get the wrong line numbers in stack traces and error line highlighting, but with this patch, I get the right ones. I'll need some time to look into what exactly is going on here.

`source-map-support` is already in the business of tinkering with Node.js internals, so I'm fine with pushing a fix through there. But then that PR plus more PRs in `babel-require` (and also `@babel/require`, I guess—huh, I should add support for that in this patch, shouldn't I?) and here to bump `source-map-support` version all have to land before this is fixed in AVA, and there's no guarantee that `source-map-support` maintainer(s) will see this as their responsibility. Can I pursue that in parallel (I agree it's the better long-term solution), and submit a patch to upgrade `source-map-support` and revert this patch if and when it lands?

novemberborn:
I've never quite figured out why the various `source-map-support` usages are incompatible. The solution may be as simple as ensuring all packages require the same version. Currently `@babel/register` pulls in `^0.4.2` and AVA pulls in `^0.5.3`.

Updating the register packages seems doable. Ideally `source-map-support` starts to follow SemVer (release a `1.0`) so that as long as packages specify the same major npm can dedupe and everybody uses the same instance.

novemberborn:
Per https://github.com/avajs/ava/issues/974#issuecomment-390478448 we need to revisit AVA's approach.

PR #1755, opened 2018-03-31
Fix Missing Import In bench/run.js

clemtrek:
`bench/run.js` did not import `bluebird` but used `Promise.each()` which resulted in an error.

Image


novemberborn:
Thanks @clemtrek! Looks like this broke in https://github.com/avajs/ava/commit/3da8603b87ae43a77f581738233ea98c3de5a10f — which landed over a year ago! Perhaps we should just remove the bench tests 😊

PR #1759, opened 2018-04-01
t.fail: return never instead of void

ajafff:
Changes the TypeScript declaration of `t.fail` to return `never` instead of `void`. It'is more correct because the call will never exit with a return value.

This allows to use `t.fail` in shorthand arrow callbacks:

```ts
declare function doStuff(param: string, cb: () => number): void;

test('doStuff', (t) => {
  // this is now possible
  doStuff('foo', () => t.fail('should not be called'));
  
  // before the change you would need to actually return something
  doStuff('foo', () => t.fail('should not be called') || 0);
  // or throw
  doStuff('foo', () => { throw t.fail('should never be called'); });
});
```

Since `never` is assignable to everything, this will not break existing code.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -53,7 +53,7 @@ export interface DeepEqualAssertion {
 
 export interface FailAssertion {
 	/** Fail the test. */
-	(message?: string): void;
+	(message?: string): never;
 
 	/** Skip this assertion. */
 	skip(message?: string): void;


novemberborn:
@ajafff very cool!

This should be applied to the other methods as well, right? Other than the `throws` assertion.

Would you know if there's a similar construct available for Flowtype?

ajafff:
AFAICT this only applies to `t.fail`. Everything else, e.g. `t.is`, returns `void` in the success case.

I don't know any flow, so I can't tell if there's a similar concept like typescript's `never`.

novemberborn:
> AFAICT this only applies to t.fail. Everything else, e.g. t.is, returns void in the success case.

Other than `t.throws()`, no assertion returns a value. I don't see how `t.fail()` is special here. Shouldn't we type everything to have `never` return value? Perhaps I'm missing something with when `never` should be used.



> I don't know any flow, so I can't tell if there's a similar concept like typescript's never

No worries 😄

ajafff:
> Shouldn't we type everything to have never return value? Perhaps I'm missing something with when never should be used.

The return type `never` means the function has no reachable endpoint. Either because of an endless loop or by throwing an exception. That's only the case with `t.fail()` as it will always throw.

A function that has a normal completion in at least one case (e.g. `t.is()`) but doesn't return anything of interest has the return type `void`.


Tooling can for example detect function calls that return `never` and mark everything following this statement as dead/unreachable code. That's true for `t.fail()` as it makes all following statements unreachable. That's not the case with `t.is()` and friends.

novemberborn:
Ah. But `t.fail()` doesn't throw. Other than `t.throws()`, the outcome of an assertion is not observable to the test.

ajafff:
I see. I thought it would behave similar to `assert.fail` in other testing libraries.

I'm closing this as the changes in this PR are not valid. Thank you for your time.

novemberborn:
Ah sorry for the misunderstanding. Thanks for putting in the PR though!

PR #1761, opened 2018-04-02
Adds support for ava.config.js files

### LINKED ISSUE #520: Adding support for a dedicated configuration file

good-idea:
Fixes #520 

 - [x] When invoked, AVA walks the file hierarchy looking for a package.json file
 - [x] AVA checks whether a sibling ava.config.js file exists
 - [x] If the package.json file contains an "ava" configuration, and ava.config.js exists, AVA will exit with a warning about ambiguous configuration
 - [x] Otherwise the package.json configuration takes precedence over ava.config.js
 - [x] ava.config.js is loaded through esm so can be either a CJS or ESM file. The loader should support ESM transpilation by looking for a thruthy __esModule property: if set then the value of the default property is used
 - [x] An actual configuration object may be exported, or a factory function
 - [x] If a factory function is exported it'll be called with the project directory (for future compatibility). It must return a configuration object
 - [x] The configuration object must not have a then property
 - (These are the same rules as for .babelrc.js files in Babel 7)
 - [x] ~Smart JS/JSON filetype in babel-config.js~
 - [x] Update docs
 - [x] Explore [`esm` options](https://www.npmjs.com/package/esm#options) so we don't have to support the module.exports workarounds etc


pvdlg:
Have you checked [cosmiconfig](https://github.com/davidtheclark/cosmiconfig)? It would fulfill a lot of the requirements, and the one missing might be implement over there.

novemberborn:
We'd end up using very little of `cosmiconfig`. It's relatively straight-forward to hook this up directly in AVA.

@good-idea thanks for your PR. I've been quite busy lately so I make need another few days before I can get back to you.

good-idea:
@novemberborn No hurry, I don't think I'll be able to chip away at this until the end of the week.

This is my first contribution to OSS 🤓 so any feedback is welcome!

lib/load-config.js
```diff
@@ -0,0 +1,90 @@
...
+};
+
+const findPackageDotJson = findFileUp('package.json');
+
+const findConfigFile = findFileUp(/ava\.config(\.\w+)?\.js/);
+
+const findConfigFiles = () => {
+	const packageFile = findPackageDotJson();
+	const configFile = findConfigFile(
+		packageFile ? path.dirname(packageFile) : process.cwd(),
```
novemberborn: Let's simplify this. We can keep the current `pkg-conf` logic to find the `package.json`, if any. Afterwards, the `ava.config.js` file *must* be in the `projectDir`. We won't walk the file system to find it. (This means the `findFileUp` logic above is not necessary.)

test/cli.js
```diff
@@ -40,7 +40,9 @@ function execCli(args, opts, cb) {
 
 		child.on('close', (code, signal) => {
 			if (code) {
-				const err = new Error(`test-worker exited with a non-zero exit code: ${code}`);
+				const err = new Error(
+					`test-worker exited with a non-zero exit code: ${code}`,
```
novemberborn: Looks like you're using Prettier. Could you exclude unrelated changes from the PR?

novemberborn:
@good-idea if you could merge in (or rebase on) master, that should fix some of the CI issues.

good-idea:
@novemberborn thanks, I'll working on this later today!

test/cli.js
good-idea: Got it!

good-idea:
@novemberborn I think this just about does it, let me know if you see anything else that should change!

lib/load-config.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
+const path = require('path');
+const fs = require('fs');
+// ESM seems to break `path` and `fs`, require them first before setting it up..
```
novemberborn: `esm` shouldn't impact `require()` calls though. Happy to see the Node.js builtins loaded first, but your comment hints at an `esm` bug. Is this still happening for you?

lib/load-config.js
```diff
@@ -0,0 +1,39 @@
...
+const requireESM = require('esm')(module);
+const pkgConf = require('pkg-conf');
+
+const loadConfigFile = projectDir => {
+	const loaded = requireESM(path.resolve(projectDir, 'ava.config.js'));
+	const config = (loaded.__esModule) ? loaded.default : loaded;
```
novemberborn: Does `esm` take care of this?

lib/load-config.js
```diff
@@ -0,0 +1,39 @@
...
+	if (config.then) {
+		throw new Error('Config file must not return a promise');
+	}
+	return (typeof config === 'function') ? config(projectDir) : config;
```
novemberborn: Let's pass `{projectDir}`, so we can extend this with other options later.

lib/load-config.js
```diff
@@ -0,0 +1,39 @@
...
+		throw new Error('Warning: You have configuration settings in both package.json and ava.config.js. Please use only one or the other.');
+	}
+
+	const config = configFileExists ? loadConfigFile(projectDir) : packageConf;
+
+	const _meta = {
+		projectDir
+	};
+
+	return Object.assign(config, defaults, {_meta});
```
novemberborn: This copies the `defaults` *into* `config`. The `_meta` bit seems unnecessary to me.

Better would be `Object.assign({}, defaults, config, {projectDir})`.

profile.js
```diff
@@ -38,9 +38,9 @@ const conf = pkgConf.sync('ava', {
 	}
 });
 
-const filepath = pkgConf.filepath(conf);
-const projectDir = filepath === null ? process.cwd() : path.dirname(filepath);
+console.log(conf);
```
novemberborn: Don't forget to remove 😉

novemberborn:
Oh, one more thing. The `fileType` values in https://github.com/avajs/ava/blob/718349ece833c30052138862785a4e7ee7bc02e6/lib/babel-config.js#L136 need to change to `'JS'` when the config originates from a `.js` file. I'm happy to fix and test that myself though.

lib/load-config.js
good-idea: I haven't looked into why, but if I set up ESM first like so:

lib/load-config.js
good-idea: no, it doesn't seem to. There's a test for this that fails if you don't do the check for `__esModule`.

lib/load-config.js
good-idea: ✅

lib/load-config.js
good-idea: ✅

profile.js
good-idea: ✅

good-idea:
@novemberborn This last update should address the notes in your review -- I didn't look into the fileType issue in babel-conifg yet, though. I updated the initial PR description to include a todo for this.

lib/load-config.js
novemberborn: This works for me, Node.js 8.11.0, and `esm@3.0.22`.

lib/load-config.js
novemberborn: Ah, I meant doing `config({projectDir})`, not `loadConfigFile({projectDir})`.

lib/load-config.js
good-idea: `esm@3.0.15` was installed, I updated to `3.0.22` and it now works as expected.

lib/load-config.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
+require = require('esm')(module); // eslint-disable-line no-global-assign
```
novemberborn: Let's do `const requireEsm` though. We don't need to change any other imports.

package.json
```diff
@@ -92,6 +92,7 @@
 		"emittery": "^0.3.0",
 		"empower-core": "^0.6.1",
 		"equal-length": "^1.0.0",
+		"esm": "^3.0.22",
```
novemberborn: The `package-lock.json` diff seems surprisingly large given this one change. Might be good to `git checkout master package-lock.json` and `npm i` again. (At least, I think that's the right `git` incantation.)

lib/load-config.js
```diff
@@ -0,0 +1,34 @@
...
+const path = require('path');
+const fs = require('fs');
+// ESM seems to break `path` and `fs`, require them first before setting it up..
```
novemberborn: This comment can be removed now.

lib/load-config.js
```diff
@@ -0,0 +1,34 @@
...
+const pkgConf = require('pkg-conf');
+
+const loadConfigFile = projectDir => {
+	const loaded = require(path.resolve(projectDir, 'ava.config.js'));
+	const config = (loaded.__esModule) ? loaded.default : loaded;
```
novemberborn: Would be good to have a fixture with an `__esModule` export.

Let's ignore the edge case of `export default {__esModule: true, default: {}}` 😉

lib/load-config.js
good-idea: no problem, ✅

package.json
good-idea: Ok, better now: from 6,951 to 266

good-idea:
@novemberborn Just added a new commit with info in the docs. Let me know what you think!
I've also updated the checklist in the original comment.
I don't think I'll have time to dive into babel-config.js until next weekend. I'm happy to, though -- but if so could you give me a little direction on how you think it would best be implemented?

lib/load-config.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+const requireEsm = require('esm')(module); // eslint-disable-line no-global-assign
```
sindresorhus: I think think `// eslint-disable-line no-global-assign` is needed.

readme.md
```diff
@@ -287,6 +287,61 @@ Arguments passed to the CLI will always take precedence over the configuration i
...
+
+To use an `ava.config.js` file:
+
+ 1. It must be in the same directory as your `package.json`.
+ 2. Your `package.json` must not contain an `ava` property.
+
+The config file can use either ESM or CommonJS syntax. Both of these are valid:
+
+```js
+module.exports = {
```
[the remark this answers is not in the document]
novemberborn: @sindresorhus I can't quite recall. I can see users outputting this file from a build step though, so it'd be annoying if that wouldn't work.

But we could start there. I think `esm` can be configured to not allow CJS.

lib/load-config.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+const requireEsm = require('esm')(module);
+const path = require('path');
+const fs = require('fs');
+const pkgConf = require('pkg-conf');
+
+const loadConfigFile = projectDir => {
+	const loaded = requireEsm(path.resolve(projectDir, 'ava.config.js'));
+	const config = (loaded.__esModule) ? loaded.default : loaded;
```
novemberborn: I want to play around with the [`esm` options](https://www.npmjs.com/package/esm#options). I *think* we can make it always treat these files as ESM, so we don't have to support the `module.exports` workarounds etc. Would be neat to start with that. What do you think @sindresorhus?

lib/load-config.js
```diff
@@ -0,0 +1,33 @@
...
+	if (config.then) {
+		throw new Error('Config file must not return a promise');
+	}
+	return (typeof config === 'function') ? config({projectDir}) : config;
```
novemberborn: The `config({projectDir})` return value *also* must not be a promise. Happy to fix this up myself.

readme.md
```diff
@@ -287,6 +287,63 @@ Arguments passed to the CLI will always take precedence over the configuration i
...
+};
+```
+
+```js
+export default {
+	// …
+};
+```
+
+The config file may also export a factory function, which will be called with an object containing a `projectDir` property. This function must return a config object, and cannot be a promise.
```
novemberborn: Perhaps:

> This function must return a config object, it mustn't return a promise.

readme.md
```diff
@@ -287,6 +287,63 @@ Arguments passed to the CLI will always take precedence over the configuration i
...
+
+```js
+const config = ({projectDir}) => {
+	if (projectDir === '...') {
```
novemberborn: `projectDir` will be an absolute path, so it'd be good if the example didn't show dots which imply it's somehow relative.

Providing the directory is useful if you have share AVA config amongst projects, e.g. in a work setting or a monorepo. You could then differentiate between projects when necessary. Perhaps this is something we should cover in a recipe, outside of this PR.

readme.md
```diff
@@ -287,6 +287,63 @@ Arguments passed to the CLI will always take precedence over the configuration i
...
+
+	return {
+		// Config B
+	};
+}
+
+export default config;
+```
+
+Or, return different configuration based on the environment: `NODE_ENV=development ava`
```
novemberborn: I'm not sure about advocating AVA to be run with different `NODE_ENV` settings. It'd be more interesting to detect CI environments — although at that point we should forward the detection to the function. Perhaps remove for now and track in a follow-up issue?

lib/load-config.js
sindresorhus: 👍

good-idea:
@novemberborn I made some changes to address the above -- there another big diff with `package-lock.json`, so I did the same as before. Let me know if everything is looking OK.

novemberborn:
@good-idea I haven't had a chance yet to look into this. I just realized we need some changes with Babel (https://github.com/avajs/ava/issues/1789) so I might try and land that before doing my work here.

Anyway this is super close so once that unblocks we should be good to go.

novemberborn:
@good-idea just wanted to say I haven't forgotten about this 😄  #1789 is pretty much done so I'm hoping to land that next weekend. I'll then try and get this in as well before cutting a release.

good-idea:
@novemberborn No hurry on my end! I've been using my own fork for now :)
Looking forward to the next release

novemberborn:
@good-idea I've merged latest `master` and pushed documentation improvements and some cleanup of the config loading. We're also now using just ESM.

There's a few scenarios that are lacking tests. Once I've done that I'll land this 🎉

PR #1763, opened 2018-04-04
Add inline docs to Assertions interface in type definitions

malimichael:
I believe it is valuable to have inline docs here. If I type `t.` in my editor I like to be able to scroll through all the assertion options and see what they do.

### DIFF
--- a/index.js.flow
+++ b/index.js.flow
@@ -46,15 +46,15 @@ export interface Assertions {
 	/**
 	 * Assert that `actual` is [the same
 	 * value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) as `expected`.
-	*/
+	 */
 	is: IsAssertion;
 
 	/**
 	 * Assert that `actual` is not [the same
 	 * value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) as `expected`.
-	*/
+	 */
 	not: NotAssertion;
-	
+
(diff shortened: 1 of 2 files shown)


novemberborn:
Thanks for the PR @malimichael. Can I ask what editor you're using? I do see the behavior you're describing in Atom.

Could you make the same changes in `index.js.flow`? We're trying to keep these files as similar as possible.

malimichael:
@novemberborn I am using VS Code v 1.22.1

malimichael:
This is what Im getting before changes for all assertions before changes

screenshot 2018-04-07 19 40 02


And this afterwards 

screenshot 2018-04-07 19 38 41

index.js.flow
```diff
@@ -31,20 +31,61 @@ export type SnapshotOptions = {
...
+	/**
+	 * Assert that `actual` is [the same
+	 * value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) as `expected`.
+	 */
 	is: IsAssertion;
+
+	/**
+	 * Assert that `actual` is not [the same
+	 * value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) as `expected`.
+	*/
```
sindresorhus: This last `*/` is not aligned.

malimichael: 😅

novemberborn:
I caught a few more whitespace fixes, but it's all good now.

Thank you @malimichael!

PR #1766, opened 2018-04-10
Escape backticks in pronuncation of AVA

harrysarson:
In the README, github interprets the backticks in the pronucaion of AVA as code delimiters.

I added `\` before the backticks which prevents this formatting.

I don't understand the phonetic alphabet so I am not 100% sure that it is now correct but there definately should not be code formatting here.

Before: `/ˈeɪvə/` ay-və
Now: \`/ˈeɪvə/\` ay-və

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1131,7 +1131,7 @@ In contrast AVA is highly opinionated and runs tests concurrently, with a separa
 
 ### How is the name written and pronounced?
 
-AVA, not Ava or ava. Pronounced [`/ˈeɪvə/` ay-və](media/pronunciation.m4a?raw=true).
+AVA, not Ava or ava. Pronounced [\`/ˈeɪvə/\` ay-və](media/pronunciation.m4a?raw=true).
 
 ### What is the header background?
 


novemberborn:
Thanks for your PR @harrysarson!

Looking at Wikipedia I don't think the backticks have any meaning in IPA. For instance, from the [Leiden](https://en.wikipedia.org/wiki/Leiden) page:

> Leiden (/ˈlaɪdən/; Dutch pronunciation: [ˈlɛi̯də(n)] (About this sound listen); in English and archaic Dutch also Leyden) is a city and municipality in the province of South Holland, Netherlands. 

See also <https://en.wikipedia.org/wiki/Help:IPA/English>.

My guess is that the backticks are used for visual effect rather than semantics.

harrysarson:
ah I see, so the whole pronouciation should have been wihin the code block?

Like:  `/ˈeɪvə/ ay-və`

Or should the second `/` be at the end too?

novemberborn:
@harrysarson I have no idea, sorry. @sindresorhus I assume you put this in originally?

sindresorhus:
> My guess is that the backticks are used for visual effect rather than semantics.

Yup

PR #1776, opened 2018-04-22
WIP: Consolidate mini and verbose reporters

### LINKED ISSUE #653: mini reporter should surface failures faster.
[Issuehunt badges](https://issuehunt.io/r/avajs/ava/issues/653)

From https://github.com/sindresorhus/ava/pull/618#issuecomment-193960009

For most test suites, the mini reporter just spits out test titles faster than the human eye can process. This gives you warm fuzzies knowing that things are actually happening, but it's not super useful. Which test failed is generally the most valuable feedback your test suite gives you, so it seems silly to make users wait until all the tests have run to deliver valuable information we already have.

My first idea was that, in addition to the current test, the reporter should persist the title of the most recent failure. This would work well if there was a single failure early on, but not so good if there were lots of failures, rapid fire, throughout the test run.

*_UPDATE: *_ We are going with @vdemedes solution here: https://github.com/sindresorhus/ava/issues/653#issuecomment-198668213

~~I now think we could do this even better by processing keys from stdin, I am thinking something like the following:~~
- ~~The tests titles fly by, you see the error count incrementing, and the title of the most recent failure is "sticky" (shown until a new failure comes along). This is basically the same idea I had before.~~
- ~~Pressing :arrow_up: puts you into error review mode, and the title of the most recent failure (at the time the :arrow_up: was pressed) is now _really_ sticky - incoming failures do not cause the "sticky" failure title to be updated.~~
- ~~Pressing :arrow_up: again takes you one failure earlier in the "failed test" list.~~
- ~~Pressing :arrow_down: takes you later in the list.~~
- ~~Pressing `enter/return` exits error review mode.~~

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $60.00)

### Submitted pull Requests
- [#2488 Consolidate mini and verbose reporter](https://issuehunt.io/r/avajs/ava/pull/2488)
---

### Tips

</details>

### LINKED ISSUE #1337: Should we move pass / fail count to the end of the output?
[Issuehunt badges](https://issuehunt.io/r/avajs/ava/issues/1337)

When there are errors, the pass/fail count shows at the top. This is almost always out of view once all the error messages are displayed.

I believe most other reporters show that as the final lines of the output. Should we follow suite?

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $60.00)

### Submitted pull Requests
- [#2487 Move fail / pass counter to the bottom of the test output](https://issuehunt.io/r/avajs/ava/pull/2487)
---

### Tips

</details>

novemberborn:
Calling this a work in progress is an overstatement, but I wanted to at least communicate what I'm trying to do.

#1722 did some major refactoring of the reporters. Here, I'd like to consolidate the mini and verbose reporters into a single, default reporter.

* When in a TTY (outside of a CI environment), shows a spinner with the status of the latest test, and the counts of various test states
* Only when a test file exits, results for that file are printed
  - In the default mode (*without* `--verbose`) only failures are shown. `t.log()` isn't shown for passing tests
  - In verbose mode the results of all tests are shown
* CI no longer defaults to `--verbose`. It only did that to disable the spinner, but the reporter now detects whether a spinner is supported.

Fixes #653. Fixes #1337.

### DIFF
--- a/lib/reporters/default.js
+++ b/lib/reporters/default.js
@@ -1,0 +1,252 @@
+'use strict';
+const os = require('os');
+const path = require('path');
+const stream = require('stream');
+
+const cliCursor = require('cli-cursor');
+const figures = require('figures');
+const indentString = require('indent-string');
+const ora = require('ora');
+const plur = require('plur');
+const trimOffNewlines = require('trim-off-newlines');
+const trimRight = require('trim-right');
+
+const chalk = require('../chalk').get();
+const codeExcerpt = require('../code-excerpt');
+const colors = require('./colors');
+const formatSerializedError = require('./format-serialized-error');
+const improperUsageMessages = require('./improper-usage-messages');
+const prefixTitle = require('./prefix-title');
+const whileCorked = require('./while-corked');
+
+class LineWriter extends stream.Writable {
+	constructor(dest, spinner) {
+		super();
+
+		this.dest = dest;
+		this.columns = dest.columns || 80;
+		this.spinner = spinner;
+		this.lastSpinnerText = '';
+	}
+
+	_write(chunk, encoding, callback) {
+		// Discard the current spinner output. Any lines that were meant to be
+		// preserved should be rewritten.
+		if (this.spinner) {
+			this.spinner.clear();
+		}
+
+		this._writeWithSpinner(chunk);
+		callback();
+	}
+
+	_writev(pieces, callback) {
+		// Discard the current spinner output. Any lines that were meant to be
+		// preserved should be rewritten.
+		if (this.spinner) {
+			this.spinner.clear();
+		}
+
+		const last = pieces.pop();
+		for (const piece of pieces) {
+			this.dest.write(piece.chunk);
+		}
+		this._writeWithSpinner(last.chunk);
+		callback();
+	}
+
+	_writeWithSpinner(chunk) {
+		if (!this.spinner || !this.spinner.id) {
+			this.dest.write(chunk);
+			return;
+		}
+
+		const str = chunk.toString('utf8');
+		this.lastSpinnerText = str;
+		// Ignore whitespace at the end of the chunk. We're continiously rewriting
+		// the last line through the spinner. Also be careful to remove the indent
+		// as the spinner adds its own.
+		this.spinner.text = trimRight(str).slice(2);
+		this.spinner.render();
+	}
+
+	writeLine(str) {
+		if (str) {
+			this.write(indentString(str, 2) + os.EOL);
+		} else {
+			this.write(os.EOL);
+		}
+	}
+}
+
+class DefaultReporter {
+	constructor(options) {
+		this.hideCursor = options.hideCursor === true;
+		this.rewriteLines = options.rewriteLines === true;
+		this.reportStream = options.reportStream;
+		this.stdStream = options.stdStream;
+		this.verbose = options.verbose === true;
+		this.watching = options.watching;
+
+		this.spinner = this.rewriteLines ?
+			ora({
+				enabled: true,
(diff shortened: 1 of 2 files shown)


lib/reporters/default.js
```diff
@@ -0,0 +1,252 @@
...
+		// as the spinner adds its own.
+		this.spinner.text = trimRight(str).slice(2);
+		this.spinner.render();
+	}
+
+	writeLine(str) {
+		if (str) {
+			this.write(indentString(str, 2) + os.EOL);
+		} else {
+			this.write(os.EOL);
```
sindresorhus: No need to use `os.EOL`. Windows also supports `\n` for console output. Even `console.log` in Node.js uses `\n`: https://github.com/nodejs/node/blob/34bd9f318a164d6b2ee949793146b85b3e152d5f/lib/console.js#L141-L148

lib/reporters/default.js
```diff
@@ -0,0 +1,252 @@
...
+		callback();
+	}
+
+	_writeWithSpinner(chunk) {
+		if (!this.spinner || !this.spinner.id) {
+			this.dest.write(chunk);
+			return;
+		}
+
+		const str = chunk.toString('utf8');
```
sindresorhus: Using `.toString('utf8')` on chunks is an anti-pattern, as the chunk might be split on a Unicode code point, resulting in an invalid character. The correct way is to do `setEncoding('utf8') on the stream. That is also more efficient.

lib/reporters/default.js
```diff
@@ -0,0 +1,252 @@
...
+		if (this.hideCursor) {
+			cliCursor.hide(this.reportStream);
+		}
+		this.lineWriter.writeLine();
+		if (this.spinner) {
+			this.spinner.start();
+		}
+	}
+
+	consumeStateChange(evt) { // eslint-disable-line complexity
```
sindresorhus: Can you use a more descriptive name than `evt`?

lib/reporters/default.js
```diff
@@ -0,0 +1,252 @@
...
+				}
+
+				this.stdStream.write(evt.chunk);
+				// If the chunk does not end with a linebreak, *forcibly* write one to
+				// ensure it remains visible in the TTY.
+				// Tests cannot assume their standard output is not interrupted. Indeed
+				// we multiplex stdout and stderr into a single stream. However as
+				// long as stdStream is different from reportStream users can read
+				// their original output by redirecting the streams.
+				if (evt.chunk[evt.chunk.length - 1] !== 0x0A) {
```
sindresorhus: Can you make `0x0A` a local constant, so it's clear what it is?

lib/reporters/default.js
```diff
@@ -0,0 +1,252 @@
...
+		const last = pieces.pop();
+		for (const piece of pieces) {
+			this.dest.write(piece.chunk);
+		}
+		this._writeWithSpinner(last.chunk);
+		callback();
+	}
+
+	_writeWithSpinner(chunk) {
+		if (!this.spinner || !this.spinner.id) {
```
sindresorhus: With latest Ora, you can check `this.spinner.isSpinning` instead of `this.spinner.id`: https://github.com/sindresorhus/ora/commit/3df8d0d962b3af9aa5c269ed9ac3d21734b34d32

novemberborn:
It may be worth considering not printing non-AVA output unless `--verbose` is set. https://github.com/avajs/ava/issues/1849#issuecomment-402502807

sindresorhus:
> It may be worth considering not printing non-AVA output unless --verbose is set. #1849 (comment)

👍 Seems fine. We have `t.log()` now for stuff that should actually be shown.

PR #1778, opened 2018-04-23
Update Flow type `PromiseLike` to be compatible with libdefs in Flow v0.70.0

hallettj:
The previous definition of `PromiseLike` gives the `onFulfill` argument to the `then` method as optional. This is not compatible with the libdefs for `Promise` in Flow v0.70.0. Those libdefs were changed as is explained in the Flow changelog:

> Existing Promise libdefs for the cases where null or void is passed as callbacks to then and catch caused us to miss type errors. These libdefs have now been made stricter.

As a result Flow emits an error for every asynchronous test.

The change in this commit changes the `onFulfill` argument to be mandatory so that `PromiseLike` is compatible with the definitions of `Promise` in both Flow v0.70.0 and Flow v0.69.0.

Flow's libdefs changed in v0.70.0 to use overloaded signatures for `then`. It is still possible in v0.70.0 to call `then` with no arguments. But as far as I can tell there is no way to write `PromiseLike` in a way that allows `then` to be called with no arguments that is compatible with both Flow v0.70.0 and Flow v0.69.0.

### DIFF
--- a/index.js.flow
+++ b/index.js.flow
@@ -1,6 +1,9 @@
 // @flow
 export interface PromiseLike<R> {
-	then<U>(onFulfill?: (value: R) => Promise<U> | U, onReject?: (error: any) => Promise<U> | U): Promise<U>;
+	then<U>(
+		onFulfill: (value: R) => Promise<U> | U,
+		onReject?: (error: any) => Promise<U> | U
+	): Promise<U>;
 }
 
 export interface ObservableLike {


novemberborn:
@hallettj wow, this is quite a change they're making!

> Flow's libdefs changed in v0.70.0 to use overloaded signatures for then. It is still possible in v0.70.0 to call then with no arguments. But as far as I can tell there is no way to write PromiseLike in a way that allows then to be called with no arguments that is compatible with both Flow v0.70.0 and Flow v0.69.0.

Are you saying that if we drop interoperability with Flow v0.69 we can support `.then(null, func)`? I'd rather be more fully compatible with how people write JavaScript than support older Flow versions.

It does raise SemVer major questions, but we're still in beta so we can do what we want 😇 It does make me wonder whether we should publish type defs in separate packages so we can break them without having to bump AVA versions. @sindresorhus thoughts on that?

sindresorhus:
> Are you saying that if we drop interoperability with Flow v0.69 we can support .then(null, func)? I'd rather be more fully compatible with how people write JavaScript than support older Flow versions.

Me too

> It does make me wonder whether we should publish type defs in separate packages so we can break them without having to bump AVA versions. @sindresorhus thoughts on that?

I don't think we should do that. I really like how it *just works* right now. If we need to make breaking type definition changes later on, we can either wait on the next major version (we're gonna make those more often going forward) or decide it's worth a small breakage.

hallettj:
> Are you saying that if we drop interoperability with Flow v0.69 we can support .then(null, func)? I'd rather be more fully compatible with how people write JavaScript than support older Flow versions.

Yes - but the `PromiseLike` interface in the AVA type definitions serves a narrow role. I argue that it is not important for that interface to support all valid use cases for promises.

`PromiseLike` is one of the types that AVA allows to be returned from a test functions, before and after functions, etc. AVA accepts `PromiseLike` values, but does not produce them. There is no case where a user would be given a value where they expect to be able to call `.then(null, func)` but it does not work. I think that makes `PromiseLike` a good candidate for a lowest-common-denominator interface, such as the one that I suggested in this pull request.

novemberborn:
From https://github.com/avajs/ava/commit/a76d462b4cdccd86028e8c9b1bbf84f12af8ab84#commitcomment-28919209:

> This seems to have broken using `async-await`

@jamiebuilds could you elaborate?

Also, we could just type `Promise` and remove this `PromiseLike` stuff. See https://github.com/avajs/ava/issues/1794.

emilingerslev:
Would it be possible to create a new beta release with this? That would be awesome because the current release is getting a bit old and this issue breaks stuff if you go async.

novemberborn:
Yes hoping to put one out soon. We've hit a blocker updating to the latest Babel beta, but I'm about to resolve that.

PR #1788, opened 2018-05-07
Disregard cached Babel options if JS source files are used

novemberborn:
If there are *.js source files, the cache cannot be used since these files must be loaded *into* the cache before getting options. Loading just the source files into the cache is insufficient: plugins and presets must *also* be loaded, *and* given the correct auto-generated name so they can override the same plugin / preset instances declared in parent config files.

Fixes https://github.com/novemberborn/hullabaloo-config-manager/issues/29.

@vjpr could you give this a try?


novemberborn:
Closing in favor of #1789.

PR #1795, opened 2018-05-12
Update readme.md source => sources

isnifer:
Got this error:
```sh
✖ The 'source' option has been renamed. Use 'sources' instead.
```


novemberborn:
Nice catch, thanks @isnifer!

Issue #1789, opened 2018-05-07
Revisit Babel caching


novemberborn:
Currently AVA resolves the Babel options [itself](https://github.com/novemberborn/hullabaloo-config-manager/). This gave us two benefits:

* Much better analysis of plugin / preset dependencies, and thus more accurate cache usage decisions
* Easier composition of AVA's built in Babel options and project configuration
* The ability to generate a single JS module which provided a fully realized configuration object

Back then the plan was to compile on-the-fly inside worker processes. Not having to resolve the configuration in each worker would have given a performance boost. We haven't gotten round to doing this yet, and I suspect it'd be the wrong approach for compiling TypeScript files.

Babel 7 is making Babel configuration files much more powerful. It's hard to keep up with those changes, especially if we want to create a fully realized configuration object. https://github.com/avajs/ava/pull/1788 is a necessary bug fix which further negates the attempted performance improvements if `*.js` configuration files are used.

It's time to revisit this approach:

* We should take the dependency analysis from https://github.com/novemberborn/hullabaloo-config-manager/, utilizing some of the low-level APIs made available in Babel 7
* Any invalid inputs should be ignored
* If analysis fails, we should skip cache inspection. Perhaps we could log a warning
* I think we can get rid of https://github.com/avajs/caching-transform and implement a simpler version in AVA itself. Notably `require('babel')` is no longer as slow as it used to be. As part of this we should be able to fix #1667
* We should let Babel resolve its actual configuration

Composing AVA's options may be the trickiest thing to achieve when using just Babel. Note that we have an outstanding bug there, too: #1732.

For now I propose we switch to use pure Babel and forgo the dependency analysis. We may have to make it easier to *clear* the cache, though.

@avajs/core thoughts?

sindresorhus:
👍 I'm all for simplifying this.

PR #1796, opened 2018-05-12
Implement --reset-cache

novemberborn:
#1789 decreases our ability to detect changes that impact the test & helper compilation output. It should therefore be easier to reset AVA's cache.

This PR adds a `--reset-cache` CLI flag. If set, all files in the `node_modules/.cache/ava` directory are deleted. AVA then exits. I'll include it in the Babel recipe once I make the Babel changes for #1789.

I've removed the `--no-cache` CLI flag. It's confusing in combination with `--reset-cache` before it *actually* tells AVA to use a temporary directory to store cache files. A subsequent run without the flag merely reverts to the default cache directory, which may still contain stale entries.

Caching can still be "disabled" by setting `"cache": false` in AVA's configuration.


revelt:
hi guys! `--reset-cache` is not documented in readme!

PR #1797, opened 2018-05-13
docs: add missing esm config instructions to es modules recipe

revelt:
Currently, our recipe for ES modules workflow is missing the `esm` config. I followed its instructions myself and for example, in [this](https://github.com/codsen/array-of-arrays-sort-by-col/tree/0ace6a1befeeee911524896221fd105f9bff45c3) library, ava accepted the raw, untranspiled ES modules code only when called with `nyc` in front:

```
nyc --require esm --reporter=html --reporter=text ava
```
Simply calling `ava` from command line would resulted in "Unexpected identifier" errors where first `import` is not recognised. Even though, exactly as per instructions, `ava` `require` `esm` was present!

Adding `esm` config into `package.json` solves this problem.

The `package.json` example in here is taken from a library `array-of-arrays-sort-by-col`, where I tested the code.

### DIFF
--- a/docs/recipes/es-modules.md
+++ b/docs/recipes/es-modules.md
@@ -26,6 +26,24 @@ Configure it in your `package.json` file, and add it to AVA's `"require"` option
 
 By default AVA converts ES module syntax to CommonJS. [You can disable this](./babel.md#preserve-es-module-syntax).
 
+Additionally, [configure](https://github.com/standard-things/esm#options) the `esm`, instructing it to enable top-level await in modules without ESM exports and CJS features in ESM.
+
+Your package.json will now contain a setup, similar to:
+
+```json
+"esm": {
+	"await": true,
+	"cjs": true
+},
+"ava": {
+	"compileEnhancements": false,
+	"require": [
+		"esm"
+	],
+	"verbose": true
+},
+```
+
 You can now use native ES modules with AVA:
 
 ```js


sindresorhus:
As for the rest of the changes, @jdalton would you be able to review?

jdalton:
`esm` defaults to `cjs` compat enabled _(so should require no configuration)_. Your current [master branch](https://github.com/codsen/array-of-arrays-sort-by-col/tree/master) should continue to work if you dropped the `cjs:true` from the config. I even dropped the `await` config option since you didn't appear to be using it in the tests.

>

revelt:
@jdalton the screenshot shows ava was called via `npm test` script and in my case it's "sandwitched" with `nyc`. There, `esm` is called in front. This approach seems to work OK.

I'm talking about calling `ava` directly from command line, by typing `ava`. I was expecting `esm` to pick up the `esm` config from package.json from ava's config:

```json
"ava": {
    "compileEnhancements": false,
    "require": [
      "esm"
    ],
    "verbose": true
  },
``` 

but it looks as if esm didn't perform its CJS duties.

Again, I'm talking about calling `ava` through command line by just typing `ava`.

screen shot 2018-05-15 at 05 42 39

---

Having said that, I did some more digging and found very strange things. Sindre was right, `"await": true,` does not affect anything, `ava` call will work without it in `esm` config in package.json. Strange thing is, I can remove `"cjs": true` too — as long as **empty key**:

```json
"esm": {
  },
```

**is present** in `package.json`, `ava` call will work. But if you remove it, `SyntaxError: Unexpected identifier` error pops up. That's very strange. Try yourself.

It's as if the presence of an empty key `esm` in package.json enables defaults in `esm`...

jdalton:
@revelt You're testing with ava `latest` which doesn't contain `esm` support. You'll want to move to the ava `beta` releases.

revelt:
Thanks a lot! Closing this.

PR #1798, opened 2018-05-13
Use Babel directly

### LINKED ISSUE #1667: Handle compilation of files ignored by Babel
Babel can be configured to ignore test and helper files. When AVA compiles those files it'll [receive `null` instead of the expected result object](https://github.com/avajs/ava/blob/7be7fe80234af9008700502f84e7b7439c3ed82c/lib/caching-precompiler.js#L57).

We need to bail out if this is the case, by [returning `null` before attempting to use the result object](https://github.com/avajs/ava/blob/7be7fe80234af9008700502f84e7b7439c3ed82c/lib/caching-precompiler.js#L69).

This needs to be handled upstream, to [avoid writing `null` to disk](https://github.com/avajs/caching-transform/blob/50cc9935bad0465bc1da6505e25576b59477a14e/index.js#L83).

Additionally we [must not return a file hash](https://github.com/avajs/ava/blob/7be7fe80234af9008700502f84e7b7439c3ed82c/lib/caching-precompiler.js#L39) to the API, and the API must check for the lack of the file hash when compiling [tests](https://github.com/avajs/ava/blob/7be7fe80234af9008700502f84e7b7439c3ed82c/api.js#L65) and [helpers](https://github.com/avajs/ava/blob/7be7fe80234af9008700502f84e7b7439c3ed82c/api.js#L150).

### LINKED ISSUE #1732: AVA should apply its Babel presets/plugins after user presets/plugins
## Description

When AVA@next (by default) merges its own Babel config with that of the user's, it seems to put its own config in front of the user's.

This is problematic in the context of ESM/CJS. AVA automatically injects a ESM->CJS transform plugin, seemingly in front of any user plugins. If the user includes a plugin which, for example, _adds_ an ESM import to a file, that import won't be transpiled to CJS.

It's especially odd if the user's config _already_ includes the ESM->CJS transform (in the correct order wrt other plugins). AVA seems to still reorder that transform, so a build that works when run through Babel will break when run through AVA.

### Test Source

[Here](https://github.com/billyjanitsch/ava-plugin-order)'s a repo demonstrating the latter case. After an `npm install`:
- `npm run build` will correctly transpile `test.js`, outputting `lib/test.js` for inspection
- `npm test` will incorrectly transpile `test.js` and output a syntax error:

```
  import React from "react";
  ^^^^^^
  
  SyntaxError: Unexpected token import
```

## Environment

macOS 10.13.3
node 8.9.4
npm 5.7.1

novemberborn:
See #1789 for context.

Simplify test and helper file precompilation. Remove parallel Babel options resolution, at the cost of false cache positives. Fixes #1732. Fixes #1667.

* [x] Currently the `@ava/stage-4` preset cannot be disabled. See https://github.com/babel/babel/issues/7920.
* [x] I couldn't make the `{modules: false}` option work, see https://github.com/babel/babel/issues/7923
* [x] I may be resolving the options incorrectly too, see https://github.com/babel/babel/issues/7919
* [x] Still need to tackle https://github.com/avajs/ava/issues/1667
* [x] Our plugin / preset dependencies need to be updated to the latest Babel beta
* [ ] I need to file a follow-up issue for testing with `babel.config.js` files
* [ ] I need to file a follow-up issue for cache improvements


billyjanitsch:
This is awesome!

novemberborn:
CI willing this is ready. I'm facing a busy week but hopefully I can land this next weekend or shortly after.

@billyjanitsch could you give this PR a try?

Issue #1811, opened 2018-05-25
RFC: Parallelizing across machines in CI


jamiebuilds:
Many CI systems support splitting up a single task across multiple machines. This can drastically speed up CI times even when you are duplicating some of the work compiling or whatever.

From what I've seen, they mostly work like this:

```yaml
steps:
  - label: test
    commands:
      - yarn install
      - yarn test
    parallelism: 5
```  

```sh
# on machine 1
CI_JOB=0 CI_JOB_COUNT=5 yarn install
CI_JOB=0 CI_JOB_COUNT=5 yarn test

# on machine 2
CI_JOB=1 CI_JOB_COUNT=5 yarn install
CI_JOB=1 CI_JOB_COUNT=5 yarn test

# on machine 3
CI_JOB=2 CI_JOB_COUNT=5 yarn install
CI_JOB=2 CI_JOB_COUNT=5 yarn test
...
```

For example: https://buildkite.com/docs/builds/parallel-builds

It would be awesome if there was an easy way to integrate this with Ava such that you could automatically chunk tests up so that they can be run across machines.

I imagine you'd have to discover all the tests and then have some stable way of splitting them up, either by counting the tests or by splitting the files up.

I think you could even automatically do this without any additional setup from developers. I created a module to help you do just that: https://github.com/jamiebuilds/ci-parallel-vars

novemberborn:
Landed in #1812.

PR #1812, opened 2018-05-25
Automatically split up test files in parallel CI jobs

jamiebuilds:
Starting point for #1811


jamiebuilds:
Awesome, thanks. I'll look into doing that today

jamiebuilds:
Actually, the one big problem with the hash approach is that it creates surprising groupings of tests.

It seems reasonable to expect tests in the same sub-directory to be mostly in one job instead of distributed across all of them.

What if we sorted the files ourselves instead? It'll be faster than creating hashes, especially since we can expect them to already be in a decently sorted starting state.

jamiebuilds:
Updated

harrysarson:
I had a look at https://github.com/hughsk/path-sort which is the dependancy added in this Pr.

There is a pull request there that has been open for a year without any response so I dunno if the package is maintained at all.

Just thought I should flag this up, I like the idea of split tests though :)

novemberborn:
> Actually, the one big problem with the hash approach is that it creates surprising groupings of tests.
> 
> It seems reasonable to expect tests in the same sub-directory to be mostly in one job instead of distributed across all of them.

@jamiebuilds I don't know if that's such a big problem. Even with sorting there is no guarantee of sensible grouping.

> I had a look at hughsk/path-sort which is the dependancy added in this Pr.
> 
> There is a pull request there that has been open for a year without any response so I dunno if the package is maintained at all.

I'd be happy to use lexical sort and not use a dependency at all.

jamiebuilds:
> Even with sorting there is no guarantee of sensible grouping.

Yeah, it might not be perfect, but we do this with Jest today and it is nicer than a completely random grouping.

We can even improve upon this and find a reliable way to keep folders together even when it distributes them unevenly.

> > I had a look at hughsk/path-sort which is the dependancy added in this Pr.
> > 
> > There is a pull request there that has been open for a year without any response so I dunno if the package is maintained at all.
> 
> I'd be happy to use lexical sort and not use a dependency at all.

I used it because otherwise it completely changes the order that Ava runs tests. Right now the way it discovers files recursively is already giving us files ordered by their directories.

I'm happy to fork `path-sort` and maintain it. I'll reach out to the original author about it

jamiebuilds:
Alternatively, Ava might want to implement a randomized order feature like in [RSpec](https://relishapp.com/rspec/rspec-core/docs/command-line/order).

novemberborn:
@jamiebuilds I've pushed some updates. Notably, the distribution algorithm has been changed. I didn't really like the floating point arithmetic. If I've figured out the maths correctly then earlier jobs will run 1 extra file each compared to later jobs. This leads to a nice predictable distribution.

Regarding `path-sort2`, why can't we rely on `Array<string>#sort()`?

We still need test coverage before we can land this. If you happen to have a project set up in CI you could test this with that'd be great too (in addition to the test coverage), just to get some confirmation it all works as expected.

novemberborn:
Yes.

Also, and I'll try and do this today, I think we should pass the `index` and `total` values as parameters to the `Api` constructor, and resolve them in `lib/cli`. That way there is no unexpected behavior when using `ava/api` directly (not that that's officially supported, but hey).

jamiebuilds:
`Array<string>#sort()` does produce a different order, nothing terribly dramatic though. In a single folder, files and sub-folders are mixed together:

- `file-dir-file-dir` `Array<string>#sort()` 
- `file-file-dir-dir` `pathSort()`

jamiebuilds:
If you do end up using `#sort()`, I would make sure it is case-insensitive:

```js
files.sort((a, b) => {
  return a.toUpperCase().localeCompare(b.toUpperCase());
});
```

novemberborn:
@jamiebuilds @sindresorhus I think this is ready now:

* CLI determines whether parallel runs are enabled, not `Api`
* I've removed `path-sort2`
* I've added documentation
* I've added an integration test

I've temporarily included #1874 which ensures integration tests actually run…

jamiebuilds:
Thank you so much for getting this in 😄

novemberborn:
@jamiebuilds thanks for the suggestion, kicking it off, and your patience! 😄

PR #1821, opened 2018-05-30
Release prep

novemberborn:
See commits.


novemberborn:
So weird how without any TAP concurrency the failure rate has gone up, unless it's somehow related to the dependency changes, but then I've seen AppVeyor pass and there are no issues locally.

PR #1831, opened 2018-06-05
Ensure assertion message is a string

### LINKED ISSUE #1125: Check assertion message is a string

TheDancingCode:
This adds a typecheck for the optional assertion message parameter, ensuring it's a string if it is provided. Implementation is the same as in #1174, but this time with a test. 

Fixes #1125.

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -32,6 +32,9 @@ const hasOwnProperty = (obj, prop) => Object.prototype.hasOwnProperty.call(obj, 
 
 class AssertionError extends Error {
 	constructor(opts) {
+		if (opts.message !== undefined && typeof opts.message !== 'string') {
+			throw new TypeError('Assertion message must be a string');
+		}
 		super(opts.message || '');
 		this.name = 'AssertionError';
 
--- a/test/assert.js
+++ b/test/assert.js
@@ -169,6 +169,14 @@ function eventuallyPasses(t, fn) {
 	});
 }
 
+test('AssertionError throws when message is not a string', t => {
+	const input = new Error('an error object');
+	const AssertionError = assert.AssertionError;
+	const expectedError = new TypeError('Assertion message must be a string');
+	t.throws(() => new AssertionError({message: input}), expectedError);
+	t.end();
+});
+
 test('.pass()', t => {
 	passes(t, () => {
 		assertions.pass();


novemberborn:
This only checks if the assertion fails. I think this should validate when the assertion is used, and fail the assertion when the message is not a string. Perhaps something like:

```js
// …
const fail = callbacks.fail

const getMessage = (test, args) => {
  if (args.length === 0) {
    return ''
  }

  if (typeof args[0] !== 'string') {
    fail(test, new AssertionError({message: 'Assertion message must be a string', values: [formatWithLabel('Got:', args[0])]})
  }

  return args[0]
}

const assertions = {
  is(actual, expected, ...rest) {
    const message = getMessage(this, rest)
    // …
  }
}
```

Except that `fail()` doesn't throw so we need to be a little more creative to ensure we don't run the actual assertion if the message is the wrong type.

novemberborn:
Hey no worries @TheDancingCode. Apologies for writing "a little creative", I didn't mean to imply it was easy: I don't really know what the answer is either! You can close this if you want, or we can figure out what the right solution is (I'm going to be mostly away the next couple of weeks though).

Do you agree that we should fail the assertion when a non-string message is provided, even if it would have otherwise succeeded?

TheDancingCode:
I guess so.

I'll close this for now. Who knows, if inspiration strikes, I'll get back to this again. ;)

PR #1833, opened 2018-06-09
Update typescript recipe

### LINKED ISSUE #1822: Update TypeScript recipe given new extensions support
It'd be good to update the TypeScript recipe in light of AVA's new `extensions` support.

I'm suggesting the following in the release notes, but I must admit I haven't actually tested it, and it's not mentioned in the recipe yet:

```json
{
    "ava": {
        "compileEnhancements": false,
        "extensions": ["ts"],
        "require": [
            "ts-node/register"
        ]
    }
}
```

Shout if you're up for trying the snippet and updating the recipe 😄

btkostner:
Fixes #1822

I removed the translation links, because it's such a large update. Should I undo this?

TODO:
- [x] Get code coverage working

### DIFF
--- a/docs/recipes/typescript.md
+++ b/docs/recipes/typescript.md
@@ -1,29 +1,47 @@
 # TypeScript
 
-Translations: [Español](https://github.com/avajs/ava-docs/blob/master/es_ES/docs/recipes/typescript.md), [Français](https://github.com/avajs/ava-docs/blob/master/fr_FR/docs/recipes/typescript.md), [Italiano](https://github.com/avajs/ava-docs/blob/master/it_IT/docs/recipes/typescript.md), [Русский](https://github.com/avajs/ava-docs/blob/master/ru_RU/docs/recipes/typescript.md), [简体中文](https://github.com/avajs/ava-docs/blob/master/zh_CN/docs/recipes/typescript.md)
+AVA comes bundled with a TypeScript definition file. This allows developers to
+leverage TypeScript for writing tests.
 
-AVA comes bundled with a TypeScript definition file. This allows developers to leverage TypeScript for writing tests.
+This guide assumes you've already set up TypeScript for your project. Note that
+AVA's definition has been tested with version 2.8.3.
 
-This guide assumes you've already set up TypeScript for your project. Note that AVA's definition has been tested with version 2.8.3.
+## Setup
 
-Add a `test` script in the `package.json` file. It will compile the project first and then run AVA.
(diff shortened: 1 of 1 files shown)


docs/recipes/typescript.md
```diff
@@ -1,29 +1,47 @@
...
+incorrect type errors, or other _non critical_ errors, you can use
+`ts-node/register/transpile-only` instead.
 
 ## Writing tests
 
 Create a `test.ts` file.
 
 ```ts
-import test from 'ava';
+import { test } from 'ava';
```
novemberborn: Does `import test from 'ava'` not work?

docs/recipes/typescript.md
```diff
@@ -1,29 +1,47 @@
...
+
+And last but not least, configure ava to build typescript files by adding
+this configuration to your `package.json` file.
 
 ```json
-{
-	"scripts": {
-		"test": "tsc && ava"
+	"ava": {
+		"babel": false,
```
novemberborn: Hopefully we can land https://github.com/avajs/ava/issues/1836 first, and then this should be removed.

novemberborn:
Let's keep the translation links. Hopefully people notice that they're looking at outdated documentation.

Formatting wise, we don't put in hard wraps in the Markdown files for this project. There seem to be a lot of unnecessary changes due to this. Could you revert those?

docs/recipes/typescript.md
btkostner: It does, but I felt that `import { test } from 'ava';` was more consistent considering that's how it's used later in the docs. Also, a lot of the typescript community (opinionated and no numbers to back this up) recommend against using `export default`.

btkostner:
My bad. A lot of those edits where me being OCD. I'll revert those tonight when I get a chance.

docs/recipes/typescript.md
novemberborn: We show the default export everywhere else so let’s keep doing the same here.

btkostner:
Alright. It looks good to me. One thing to note, I was unable to get `nyc` code coverage to work correctly with this setup, but that's not a show stopper for this PR.

btkostner:
Never mind on my last message. Apparently it's working on travis, so something locally is funky with my setup.

novemberborn:
> I was unable to get nyc code coverage to work correctly with this setup, but that's not a show stopper for this PR.
>
> Never mind on my last message. Apparently it's working on travis, so something locally is funky with my setup.

Yea that stuff gets real difficult real quick. Thanks for looking into it though!

PR #1834, opened 2018-06-10
log typescript errors to console

btkostner:
Fixes TS building issue not being outputted

Mentioned in https://github.com/avajs/ava/issues/1822

### DIFF
--- a/lib/reporters/mini.js
+++ b/lib/reporters/mini.js
@@ -287,6 +287,11 @@ class MiniReporter {
 			this.lineWriter.writeLine(evt.err.summary);
 		}
 
+		if (evt.err.name === 'TSError' && evt.err.object && evt.err.object.diagnosticText) {
+			this.lineWriter.writeLine();
+			this.lineWriter.writeLine(evt.err.object.diagnosticText);
+		}
+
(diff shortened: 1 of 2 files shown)


novemberborn:
This looks good!

I realize this may be a bit of a stretch, but perhaps you could try and write an integration test for this? See https://github.com/avajs/ava/blob/master/test/reporters/mini.js (and `verbose.js`) and https://github.com/avajs/ava/blob/master/test/helper/report.js for the starting points.

btkostner:
I'll take a look at that tonight and see what I can get done!

test/fixture/source-map-initial.js
```diff
@@ -1,23 +1,22 @@
-'use strict';
```
btkostner: These changes were made by the linter

btkostner:
I added tests for the reporters. They include stack traces because of the errors, but I'm not sure the best way to clear those. Assuming that `ts-node` doesn't change we should be good though.

novemberborn:
Thanks @btkostner! I've taken the liberty to push some changes as I'd like to get this into today's release. Weirdly I couldn't get compile errors out of the code you pushed.

PR #1837, opened 2018-06-11
Make links point to specific git tag

### LINKED ISSUE #1757: When linking to documentation, link to specific Git tag
hi all,

One of error messages is saying:

> Unexpected Babel configuration for AVA. See https://github.com/avajs/ava#es2017-support for allowed values.

screen shot 2018-03-31 at 12 20 03

However, the link is not valid any more. Let's fix it!

TheDancingCode:
Prevents the documentation links that AVA returns from going out of date.

Closes #1757.

### DIFF
--- a/lib/babel-pipeline.js
+++ b/lib/babel-pipeline.js
@@ -8,6 +8,7 @@ const isPlainObject = require('is-plain-object');
 const md5Hex = require('md5-hex');
 const packageHash = require('package-hash');
 const stripBomBuf = require('strip-bom-buf');
+const pkg = require('../package.json');
 const chalk = require('./chalk').get();
 
 function getSourceMap(filePath, code) {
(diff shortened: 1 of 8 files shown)


novemberborn:
Thanks!

Some of these links end up in the reporter integration tests. You can update those logs by running:

```console
$ UPDATE_REPORTER_LOG=1 npx tap --no-cov --timeout=300 --jobs=4 test/reporters/{tap,mini,verbose}.j
```

However you'll need to add a [sanitizer](https://github.com/avajs/ava/blob/b6a95b14099df88d33b12b998f41fa9bb2bbc60d/test/helper/report.js#L63) that rewrites the current version to something else, otherwise the tests will fail every time we publish a new release.

Let me know if you'd like some more pointers with this.

TheDancingCode:
Thank you for the pointers. I don't manage to update the logs however. Maybe it's a Windows thing, but I'm getting: `'UPDATE_REPORTER_LOG' is not recognized as an internal or external command,
operable program or batch file.`.

TheDancingCode:
Should be good now. :)

novemberborn:
Fantastic, thanks @TheDancingCode!

PR #1839, opened 2018-06-11
`Fixes #1767`  Update to the Compile Sources Information page

### LINKED ISSUE #1767: @babel/register is the wrong module when loading sources compiled with Babel 6
## Description

**babel options** under **ava congifure** isn't working.

### Test Source

```js
import test from 'ava';

import { add } from './index'

test('add', t => {
  t.is(add(1, 2), 3)
})

```

image

### Config

```json
{
  "ava": {
    "require": [
      "@babel/register"
    ],
    "babel": {
      "testOptions": {
        "babelrc": false,
        "presets": [
          "@babel/env"
        ]
      }
    }
  }
}
```

### Command-Line Arguments

```
ava
```

## Relevant Links

[ava-demo](https://github.com/lili21/ava-demo)

## Environment

Node.js v9.8.0
darwin 17.2.0
ava 1.0.0-beta.3
npm 5.6.0

ashimagarwal:
Fixes Issue #1767 
Correct babel-register module based on how source files are compiled.

I have added two different lines recommending programmers:
- To use the correct register module based on their preference and 
- Specify different register-modules for both Babel 6 and Babel 7 config choices

### DIFF
--- a/docs/recipes/babel.md
+++ b/docs/recipes/babel.md
@@ -153,6 +153,8 @@ You'll need to install `@babel/polyfill` yourself.
 
 AVA does not currently compile source files. You'll have to load [Babel's `register` module](http://babeljs.io/docs/usage/require/), which will compile source files as needed.
 
+To compile source files on-the-fly, use `@babel/register` else, use `babel-register`.
+
 You can enable the `register` module by adding it to AVA's `require` option:
 
 ```json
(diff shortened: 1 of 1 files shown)


docs/recipes/babel.md
```diff
@@ -153,6 +153,8 @@ You'll need to install `@babel/polyfill` yourself.
 
 AVA does not currently compile source files. You'll have to load [Babel's `register` module](http://babeljs.io/docs/usage/require/), which will compile source files as needed.
 
+To compile source files on-the-fly, use `@babel/register` else, use `babel-register`.
```
novemberborn: This line is unnecessary. However, rather than the line at the end of this section, perhaps this should say something like:

> The below example uses `@babel/register` and assumes you're using Babel 7 to compile your source files. If you're using Babel 6, use `babel-register` instead.

What do you think?

sindresorhus:
Ping @ashimagarwal

novemberborn:
Now that Babel 7 is out, I think we can assume people are upgrading. We probably shouldn't be making our documentation more complicated by covering Babel 6.

PR #1840, opened 2018-06-12
Fix CI, errors in latest beta

### LINKED ISSUE #1826: stream.clearLine is not a function
When upgrading to the beta5 I started hitting [this error](https://github.com/avajs/ava/commit/2609494f530aea1a8b7321dc8b24d4e67f3e7b90#commitcomment-29203814). It looks like it's erroring out in an Ora dep but may be related to the commit _(linked in the previous sentence)_.

I'm on MacOS 10.13.4 using iTerm 3 with oh-my-zsh.

### LINKED ISSUE #1836: Crash when using custom extensions outside of Babel pipeline with enhancement compilation disabled
From https://github.com/avajs/ava/issues/1822#issuecomment-394446632:

```json
  "ava": {
    "compileEnhancements": false,
    "extensions": [
      "ts"
    ],
    "require": [
      "ts-node/register",
      "tsconfig-paths/register"
    ]
  },
```

Because of how `ts` is specified, AVA won't apply its Babel pipeline to it. However, note that the `babel` option is not disabled.

The logic in [these lines](https://github.com/avajs/ava/blob/b6a95b14099df88d33b12b998f41fa9bb2bbc60d/api.js#L228:L246) is broken: 

```js
const {projectDir, babelConfig} = this.options;
const compileEnhancements = this.options.compileEnhancements !== false;
const precompileFull = babelConfig ?
  babelPipeline.build(projectDir, cacheDir, babelConfig, compileEnhancements) :
  filename => {
    throw new Error(`Cannot apply full precompilation, possible bad usage: ${filename}`);
  };
const precompileEnhancementsOnly = compileEnhancements && this.options.extensions.enhancementsOnly.length > 0 ?
  babelPipeline.build(projectDir, cacheDir, null, compileEnhancements) :
  filename => {
    throw new Error(`Cannot apply enhancement-only precompilation, possible bad usage: ${filename}`);
  };


this._precompiler = {
  cacheDir,
  enabled: babelConfig || compileEnhancements,
  precompileEnhancementsOnly,
  precompileFull
};
```

Note how `this._precompiler.enabled` ends up being `true`. However `precompileEnhancementsOnly` is set to a function that always throws. Consequently, [when compiling](https://github.com/avajs/ava/blob/b6a95b14099df88d33b12b998f41fa9bb2bbc60d/api.js#L144:L146) TypeScript files, `precompileEnhancementsOnly` is called and AVA crashes.

I think the logic should be such that a `() => null` function is when `compileEnhancements` is `false`, and the error-throwing function is used when it's not and `this.options.extensions.enhancementsOnly.length` is `0`.

novemberborn:
This makes tests pass on Travis again. It's not great, I think once we get a 1.0 out we should look into testing AVA with the previous stable release.

This fixes #1836. @vladimiry could you give it a try? You shouldn't need `babel: false` anymore.

This also fixes #1826. @jdalton I cloned `esm` and linked my local AVA install, removing your `CI` env workarounds. Tests passed so yay! But perhaps you could give it a try as well?

Will merge and ship tomorrow.


vladimiry:
@novemberborn Just tried version from the `fixes` branch`, works well for me without `babel: false` option set, thanks.

jdalton:
Thank you @novemberborn!

PR #1869, opened 2018-07-15
Move promise-handling code to t.throwsAsync and t.notThrowsAsync

### LINKED ISSUE #1794: Remove observable support, generators, only type native promises?
AVA currently supports observables, returned by test implementations and used in `t.throws()` / `t.notThrows()` assertions. We're not happy with how they're typed though, see #1685. Their progress through TC39 has also been slower than expected.

I wonder whether we should remove support in AVA itself? Though less graceful, it wouldn't be hard to wrap observable-returning functions and do the same translation to a promise that AVA does at the moment. As precedent there are other asynchronous concepts in Node.js that we don't support directly (streams, async iterators).

This would simplify our typings for `t.throws()`, which hurt usability in Flowtype (#1790).

I'm not convinced we need to support generator functions as test implementations either.

Similarly, we support tests that return thenables, rather than promises. Typing these "promise-like" objects is also more difficult than expected (#1778). At the very least we could change the typings to expect a `Promise`. I don't quite think it's worthwhile enforcing tests to return proper `Promise` instances rather than thenables, and properly typed promise implementations should match the `Promise` interface anyhow.

---

If you're using observables or generators, how would you feel if we'd remove first-class support?

Similarly, would it be a terrible bother if our typings were stricter on promises?

@avajs/core, any thoughts?

novemberborn:
See #1794.

---

`t.throws()` and `t.notThrows()` no longer take async functions or promises. Instead, use `await t.throwsAsync()` and `await t.notThrowsAsync()`. The async assertions fail if the function throws synchronously.

Updated tests, documentation and type definitions.

### DIFF
--- a/index.js.flow
+++ b/index.js.flow
@@ -75,6 +75,9 @@ export interface Assertions {
 	/** Assert that the function does not throw. */
 	notThrows: NotThrowsAssertion;
 
+	/** Assert that the async function does not throw, or that the promise does not reject. Must be awaited. */
+	notThrowsAsync: NotThrowsAsyncAssertion;
+
 	/** Count a passing assertion. */
 	pass: PassAssertion;
 
@@ -92,6 +95,12 @@ export interface Assertions {
 	 * Assert that the function throws [an error](https://www.npmjs.com/package/is-error). If so, returns the error value.
 	 */
 	throws: ThrowsAssertion;
+
+	/**
+	 * Assert that the async function throws [an error](https://www.npmjs.com/package/is-error), or the promise rejects
+	 * with one. If so, returns a promise for the error value, which must be awaited.
+	 */
+	throwsAsync: ThrowsAsyncAssertion;
 
 	/** Assert that `actual` is strictly true. */
 	true: TrueAssertion;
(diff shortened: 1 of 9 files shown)


index.js.flow
```diff
@@ -219,66 +233,68 @@ export interface SnapshotAssertion {
 
 export interface ThrowsAssertion {
 	/**
-	 * Assert that the function returns a promise that rejects with [an error](https://www.npmjs.com/package/is-error).
-	 * If so, returns the rejection reason. You must await the result.
+	 * Assert that the function throws [an error](https://www.npmjs.com/package/is-error). If so, returns the error value.
 	 */
-	(fn: () => PromiseLike<any>, expectations?: null, message?: string): Promise<any>;
+	(fn: () => any, expectations?: null, message?: string): any;
```
sindresorhus: Not relevant to this PR, but we still wanted to type it to return an `Error` instead of `any`, right? Per https://github.com/avajs/ava/issues/1794

novemberborn: Ah yes, I guess I should be careful not to close that issue. Ran out of time though on Sunday, barely got this PR out 😄

PR #1872, opened 2018-07-18
Copy flow's promise then() interface into PromiseLike

padmaia:
This allows Flow promise types to be compatible with what AVA is expecting

https://github.com/facebook/flow/blob/c88262be2be0b94b04bfaaf5e52570856d0b87e3/lib/core.js#L589-L597

These should no longer require extra typing:

```js
await t.throws(() => promise);
await t.throws(promise);
```

PS: This PR was made by @jamiebuilds from my computer. "HI!" – Jamie

### DIFF
--- a/index.js.flow
+++ b/index.js.flow
@@ -1,8 +1,13 @@
 // @flow
 export interface PromiseLike<R> {
+	then(onFulfill: null | void, onReject: null | void): Promise<R>;
+	then<U>(
+		onFulfill: null | void,
+		onReject: (error: any) => Promise<U> | U
+	): Promise<R | U>;
 	then<U>(
 		onFulfill: (value: R) => Promise<U> | U,
-		onReject?: (error: any) => Promise<U> | U
+		onReject: null | void | ((error: any) => Promise<U> | U)
 	): Promise<U>;
 }
 


novemberborn:
Thanks @padmaia!

Does this mean that Flow can distinguish between the following overloads?

```flow
throws(() => any): Error
throws(() => PromiseLike<any>): Promise<Error>
```

We came to the conclusion in #1794 that we had to separate them. There's a PR ready to land for that: #1869.

That may still be the right thing to do, but would be good to know if we maybe don't have to.

novemberborn:
@jamiebuilds any idea on the above?

@sindresorhus regardless of whether the overloads are possible, do you still think we should make `t.throws` synchronous, and add `t.throwsAsync`?

sindresorhus:
> do you still think we should make t.throws synchronous, and add t.throwsAsync?

Yes, I think the change makes sense regardless, as it will make it more explicit and prevent mistakes.

jamiebuilds:
```js
throws(() => any): Error
throws(() => PromiseLike<any>): Promise<Error>
```

No, this won't work as you expect it to. `any` is the parent type of all types so it won't match.

Imagine a pattern match like this:

```js
if (val is anything) { // always `true`
  // do this
} else if (val is specific thing) {
  // do that
}
```

It's always going to match the first case, so the second case doesn't really matter. But you also can't just flip them because the types don't have the same sort of control flow.

In order to say:

- If I get a promise, return a promise
- If I get an observable, return an observable
- If I get anything else, return `void`

You need to be able to distinguish between promises, observable, and "anything else". Otherwise, if it overlaps, you'll get an intersection of the types you are returning (You'll "return" promise & void)

This is what it means to have "discriminant unions":

```js
type UnionT =
  | { type: 'A' }
  | { type: 'B' }
  | { type: 'C' }
```

With a type like that, just by asserting `type` we know exactly which one of the three options we've got.


```js
type UnionT =
  | { prop: boolean }
  | { prop: any }
  | {}
```

But if we checked that `prop` is a `boolean` here, it still matches all three options.

So short answer:

```js
throws(() => void): Error
throws(() => Promise): Promise<Error>
throws(() => Observable): Observable<Error>
```

would work (void, promise, observable are all distinguishable)

```js
throws(() => any): Error
throws(() => Promise): Promise<Error>
throws(() => Observable): Observable<Error>
```

would not (any can not be distinguished from promise or observable)

novemberborn:
Thanks for clarifying @jamiebuilds.

novemberborn:
Thanks @padmaia!

PR #1875, opened 2018-07-22
Add reasonml/bucklescript's binding

godu:
As suggested [here](https://github.com/avajs/awesome-ava/pull/24), I merged [@godu/bs-ava](https://github.com/godu/bs-ava) with ava.

Annoying points:
- [ ] BuckleScript use `lib/bs` and `lib/js` as output directories. It's not a breaking point, but it's a little weird to mix them with ava's files.

> [The output directory is not configurable otherwise.](https://bucklescript.github.io/docs/en/build-configuration.html#package-specs)

ava-bucklescript

- [ ] Ignore `lib/bs` and `lib/js` on `npm pack`. I test an ava's installation using `npm pack` and `npm install ava.tgz`. `lib/bs` and `lib/js` were both in the tarball. [NPM should use .gitignore to filter files](https://github.com/npm/cli/wiki/Files-and-Ignores#npmignore). I don't know if it's only a `npm pack` command bug or if npm will do the same thing on publish.

npm-pack


novemberborn:
@godu thanks for this PR!

> BuckleScript use lib/bs and lib/js as output directories. It's not a breaking point, but it's a little weird to mix them with ava's files.

This is the `bsb -make-world` output? I think that's fine. Perhaps though don't ignore the directories themselves, but add a `README.md` which explains why they're there, and nest a `.gitignore` that ignores all other files inside.

Speaking of, could you remove the `.DS_Store` file? It's probably best if you add that to your machine's `.gitignore` file, not AVA's.

Is the `bsconfig.json` necessary for publication? If so, is it odd that the `test/bs-types` files are not published?

Could you put the definition files in a different directory than `src`? `bucklescript` maybe?

The `name` value in` bsconfig.json` should probably just be `ava`, like the `package.json`.

> Ignore lib/bs and lib/js on npm pack. I test an ava's installation using npm pack and npm install ava.tgz. lib/bs and lib/js were both in the tarball. NPM should use .gitignore to filter files. I don't know if it's only a npm pack command bug or if npm will do the same thing on publish.

It'll definitely do the same on publish. Perhaps because `lib` is in `package.json#files` it overrides. I think you could add `!lib/js` and that should fix it.

---

The definition files are quite intimidating, by the way. Hopefully you'll stick around to help keep them up to date 😄 Note that the `throws` assertion has recently changed, see #1869.

I'm excited to see this land. Let's get the ReasonML community to use AVA! 🚀

godu:
- > This is the bsb -make-world output? I think that's fine. Perhaps though don't ignore the directories themselves, but add a README.md which explains why they're there, and nest a .gitignore that ignores all other files inside.

I try to put `.gitignore` and `readme.md` inside `lib/bs` and` lib/js`, but` bsb -clean` completely removes this folders.
I fallback on add `lib/.gitignore` and `lib/readme.md` and explain this folders presence in `readme.md`.

- > It'll definitely do the same on publish. Perhaps because lib is in package.json#files it overrides. I think you could add !lib/js and that should fix it.

NPM's documentation recommends to use .npmignore (and .gitignore as fallback) instead of negative glob `files` entry.

> ["The consequences are undefined" if you try to negate any of the files entries (that is, "!foo.js"). Please don't. Use .npmignore.](https://github.com/npm/cli/wiki/Files-and-Ignores#details)

- > Is the bsconfig.json necessary for publication? If so, is it odd that the test/bs-types files are not published?

You're right, it's very odd. I create a second `bsconfig.json` in `test/bucklescript` folder. It's look like better.

---

I'll fixed the implementation to follow the new API.

Thank a lot for you help.

---

@novemberborn Did you notice comment sections in `Sync.re`, `Async.re` and `Promise.re` ? I followed ava and flow definitions and the both allow the following unimplemented APIs: 
- test.serial.failing.only
- test.serial.failing.skip
- test.serial.only.cb
- test.serial.failing.only.cb
- test.serial.skip.cb
- test.serial.failing.skip.cb

docs/recipes/reasonml.md
```diff
@@ -0,0 +1,43 @@
...
+  Js.Global.setTimeout(() => t.cb(), 1000)
+  |> ignore
+});
+```
+
+```ml
+open Promise;
+
+test("Should pass", t => {
+  t.notThrows(Js.Promise.resolve(true"));
```
novemberborn: This should be `await t.notThrowsAsync(Js.Promise.resolve(true"));`.

docs/recipes/reasonml.md
```diff
@@ -0,0 +1,43 @@
...
+  "bs-dev-dependencies": ["ava"]
+}
+```
+
+## Getting started
+
+```ml
+open Sync;
+
+test("Should pass", t => {
```
novemberborn: We don't tend to put "should" in test titles. "1 equals 1" would be more AVA-like.

docs/recipes/reasonml.md
```diff
@@ -0,0 +1,43 @@
...
+  t.deepEqual(1, 1);
+  t.pass();
```
novemberborn: `t.pass()` is unnecessary here.

docs/recipes/reasonml.md
```diff
@@ -0,0 +1,43 @@
...
+});
+```
+
+```ml
+open Async;
+
+test("Should pass", t => {
+  Js.Global.setTimeout(() => t.cb(), 1000)
```
novemberborn: This should be `test.cb()` with a `t.end()` call in the timeout.

docs/recipes/reasonml.md
godu: ReasonML returns last statement of function. It's like following in javascript

```js
test("function returns a fulfilled promise", t => {
  return t.notThrowsAsync(() => Promise.resolve(true));
});
````

File: docs/recipes/reasonml.md (hunk 4 above)
Comment (godu) [reply docs/recipes/reasonml.md]: `end` is a reserved word in ocaml. I try to find something similar. I'm open to discussion.

**Comment by godu** [comment]
@novemberborn I'm inspired by [bs-mocha](https://github.com/reasonml-community/bs-mocha/tree/master/src). They split sync, promise and callback worlds.
With ava, i create 3 namespaces: `Ava.Sync` `Ava.Async` and `Ava.Promise`. I pick all assertion that i consider in its place. I tried to implement obvious assertions and skip the weird ones. I will be happy to discussion about my choice and the next step for reasonml's ava.

**Comment by godu** [comment]
> > Is the bsconfig.json necessary for publication? If so, is it odd that the test/bs-types files are not published?

> You're right, it's very odd. I create a second bsconfig.json in test/bucklescript folder. It's look like better.

I had some issues with my solution. I rollback to my initial solution. I have no idea how to correctly setup as you suggest.

**Comment by novemberborn** [comment]
OK I've been trying to read up on BuckleScript and Reason. It's a bit of a challenge as the documentation is biased to starting a new project, rather than writing bindings for existing ones.

The idea seems to be that you write modules in Reason. They can "interop" with existing packages. This PR adds those interoperating modules to the `ava` package.

AVA's `test` interface is built on the notion that functions are objects, too. Judging by your code I imagine that's not the case in Reason. I'd like to keep the conceptual differences to a minimum though. Is there a way we can support `test()` and `test.failing()` without having to resort to underscores? Perhaps `test.declare()` and `test.failing.declare()`?

Skipping assertions poses a similar problem: `t.deepEqual()` versus `t.deepEqual.skip()`. In that scenario I guess `t.deepEqual_skip()` is preferable.

Is `Ava.Sync` used for synchronous test implementations, `Ava.Async` for `test.cb()` calls, and `Ava.Promise` for asynchronous test implementations? How does `Ava.Promise` handle promises? Are these distinctions meaningful? That is, could we get away with *always* creating asynchronous test implementations, without support for `test.cb()`?

Given that this is less a type definition than a wrapper implementation, should we perhaps publish this as `@ava/bucklescript`? It's hard enough to keep the type definitions up to date, but at least they raise fewer existential questions about how to express AVA's untyped behavior 😄

**Comment by godu** [comment]
@sindresorhus @novemberborn I agree with a separate package.

I updated [@godu/bs-ava](https://github.com/godu/bs-ava). We could move this repo as `ava/bucklescript`.

```reason
open AvaBucklescript.Sync;

test("passed", t => 
  t.pass
);
```

PR #1876, opened 2018-07-23
Modernize some code

sindresorhus:
Some things XO wouldn't catch.


novemberborn:
You may want to re-enable these rules as well: https://github.com/avajs/ava/blob/c4f607cc376846363320691f65c6e96ef401996d/package.json#L174:L178

PR #1880, opened 2018-07-29
Fix typo in endpoint-testing-with-mongoose

CanRau:
fix small typo in setup list mongod-memory-server -> mongodb-memory-server

copied and tried to install it, which of cause failed, so might safe others ;-)


novemberborn:
Thanks @CanRau!

PR #1885, opened 2018-08-02
Provide getColorDepth() shim for fake TTYs

jagoda:
The AssertionError constructor in Node 10 depends on the TTY
interface. More specifically, if isTTY is true on stdout then
getColorDepth() must also be available. Ava provides a shim for
isTTY in the test workers but does not provide a getColorDepth()
implementation. This leads to unexpected failures when code under
test is using Node's built-in assert module.

### DIFF
--- a/lib/fork.js
+++ b/lib/fork.js
@@ -39,6 +39,7 @@ module.exports = (file, opts, execArgv) => {
 		file,
 		baseDir: process.cwd(),
 		tty: process.stdout.isTTY ? {
+			colorDepth: process.stdout.getColorDepth ? process.stdout.getColorDepth() : undefined,
 			columns: process.stdout.columns || 80,
 			rows: process.stdout.rows
 		} : false
--- a/lib/worker/fake-tty.js
+++ b/lib/worker/fake-tty.js
@@ -5,6 +5,10 @@ const options = require('./options').get();
 if (options.tty) {
 	Object.assign(process.stdout, {isTTY: true}, options.tty);
 
+	if (options.tty.colorDepth !== undefined) {
+		process.stdout.getColorDepth = () => options.tty.colorDepth;
+	}
+
 	const isatty = tty.isatty;
 	tty.isatty = function (fd) {
 		if (fd === 1 || fd === process.stdout) {
--- a/test/fixture/node-assertions/assert-failure.js
+++ b/test/fixture/node-assertions/assert-failure.js
@@ -1,0 +1,6 @@
+import assert from 'assert';
+import test from '../../..';
+
+test('test', () => {
+	assert(false);
+});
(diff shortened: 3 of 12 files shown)


lib/worker/fake-tty.js
```diff
@@ -4,6 +4,9 @@ const options = require('./options').get();
 
 if (options.tty) {
 	Object.assign(process.stdout, {isTTY: true}, options.tty);
+	process.stdout.getColorDepth = function () {
+		return options.tty.colorDepth;
+	};
```
sindresorhus:
```js
process.stdout.getColorDepth = () => options.tty.colorDepth;
```

jagoda:
@sindresorhus I made the change you requested. The build failures seem unrelated to my changes and rather seem to have been introduced in 8663028de2d7a7be88c4a101dafb873c5af87937. Please correct me if I'm wrong and I will happily fix.

sindresorhus:
> The build failures seem unrelated to my changes and rather seem to have been introduced in 8663028. Please correct me if I'm wrong and I will happily fix.

Yup. Already fixed.

lib/worker/fake-tty.js
```diff
@@ -4,6 +4,7 @@ const options = require('./options').get();
 
 if (options.tty) {
 	Object.assign(process.stdout, {isTTY: true}, options.tty);
+	process.stdout.getColorDepth = () => options.tty.colorDepth;
```
novemberborn: We should only shim this in Node.js versions where it's expected to be available.

lib/fork.js
```diff
@@ -39,6 +39,7 @@ module.exports = (file, opts, execArgv) => {
 		file,
 		baseDir: process.cwd(),
 		tty: process.stdout.isTTY ? {
+			colorDepth: process.stdout.getColorDepth ? process.stdout.getColorDepth() : 1,
```
novemberborn: Perhaps leave as `undefined` if there is no `getColorDepth()`, and use that as a signal of whether to shim or not?

Or `null` if you prefer.

test/integration/custom-assertions.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
+const test = require('tap').test;
+const {execCli} = require('../helper/cli');
+
+// The AssertionError constructor in Node 10 depends on the TTY interface
+test('node assertion failures are reported to the console when running in a terminal', t => {
+	const options = {
+		dirname: 'fixture/custom-assertions',
```
novemberborn: Could you rename the fixtures (and this integration test) to `node-assertions`? "Custom" implies support for a wider range of assertions, like Chai.

jagoda:
@novemberborn thanks for the feedback! I believe that I have addressed all of your concerns. I also added some extra testing around the different possible TTY states. Please let me know if you have any further concerns.

novemberborn:
Fantastic work, thanks @jagoda!

PR #1886, opened 2018-08-04
Fixes issue #583 - Print pending tests when interrupting AVA, or after a timeout

vancouverwill:
First PR for this project so wanted to get some eyes on it before I spent too much time on it and check I am moving in right direction 😄 

### Worked on so far
* keeping track of which tests are running and updating them when finished
* updating test summary to include list of running tests with their files

### Still to do 
* run this summary after interrupting AVA

### Unknowns
* What the **Tap** and **Mini** reporters are for?
* Would you like extra tests in `test/reporters/verbose.js`, what format should these be in?


lib/reporters/verbose.js
```diff
@@ -70,6 +70,8 @@ class VerboseReporter {
 		this.failures = [];
 		this.filesWithMissingAvaImports = new Set();
 		this.knownFailures = [];
+		this.runningTestFiles = {};
```
novemberborn: Use a `Map`.

lib/reporters/verbose.js
```diff
@@ -245,6 +250,36 @@ class VerboseReporter {
 		}
 	}
 
+	addTestRunning(evt) {
+		this.runningTestsCount++;
+		if (this.runningTestFiles[evt.testFile] === undefined) {
+			this.runningTestFiles[evt.testFile] = {};
```
novemberborn: Use a `Set`.

lib/reporters/verbose.js
```diff
@@ -245,6 +250,36 @@ class VerboseReporter {
...
+
+	removeTestRunning(evt) {
+		this.runningTestsCount--;
+		delete this.runningTestFiles[evt.testFile][evt.title];
+	}
+
+	/**
+	 * Generates the summary string of tests that are still running
+	 * @returns string
+	 */
```
novemberborn: To be honest we don't really bother with these comments 😄

lib/reporters/verbose.js
```diff
@@ -137,15 +140,17 @@ class VerboseReporter {
...
 			case 'test-passed':
 				if (evt.knownFailing) {
 					this.knownFailures.push(evt);
 				}
 				this.writeTestSummary(evt);
+				this.removeTestRunning(evt);
 				break;
 			case 'timeout':
-				this.lineWriter.writeLine(colors.error(`${figures.cross} Exited because no new tests completed within the last ${evt.period}ms of inactivity`));
+				this.lineWriter.writeLine(colors.error(`${figures.cross} Exited because no new tests completed within the last ${evt.period}ms of inactivity\n${this.runningTestSummary()}`));
```
novemberborn: `timeout` applies to the set of test files that are currently running. However unless `--fail-fast` is set, AVA will continue to run remaining test files. This means multiple timeouts may occur.

We should include the timed out files when the event is emitted: https://github.com/avajs/ava/blob/55e40213398dbcfd7f9941697e556b335f5229f2/api.js#L77

Then if you pass the `evt` to `runningTestSummary()` we could print the files appropriate for each particular timeout.

test/integration/assorted.js
```diff
@@ -7,10 +7,15 @@ const stripAnsi = require('strip-ansi');
...
-test('timeout', t => {
+test('timeout while tests running', t => {
 	execCli(['long-running.js', '-T', '1s'], (err, stdout) => {
 		t.ok(err);
 		t.match(stdout, /Exited because no new tests completed within the last 1000ms of inactivity/);
+		t.match(stdout, /3 tests still running/);
+		t.notMatch(stdout, /.*long-running.js:fast.*/, 'the fast test should not still be shown as running');
+		t.match(stdout, /.*long-running.js:slow.*/);
+		t.match(stdout, /.*long-running.js:slow two.*/);
+		t.match(stdout, /.*long-running.js:slow three.*/);
```
novemberborn: This timeout test ought not to be here. I think I left it in on purpose since it was reasonably small, but obviously that's no longer the case 😄 Let's remove these additions and write a reporter test instead.

vancouverwill:
hi @novemberborn thanks for the pointers so far 😄 

I used the `set()` and `map()` like you said and created new tests in `test/reporters/verbose.js` and these are in the fixtures folder within `test/fixture/report/timeoutinmultifiles/...` and `test/fixture/report/timeoutinsinglefile/...` . 

Within `test/helper/report.js` I have updated the timeout, this caused me some issues as I was passing in straight ms i.e. `1000` but this was getting converted to a string `1s` because of the library `ms` we use and didn't notice that straight away but even when I pass in the correct format for that library
`timeout: type.substring(0,7) === 'timeout' ? "1000ms" : undefined,` it doesn't seem to get picked up.

Within `api.js` I have logged out what the given timeout argument is and this is the same in my new tests as what is passed in the original `integration/assorted.js` tests but for whatever reason the timeout never gets triggered and all the tests end up passing.

Apologies because I have put a lot of in `Api.js` and the Verbose reporter trying to determine why the timeout never finishes but so far have not been able to track it down. So I wondered if you had any pointers as to why the timeout is not getting triggered? 😕 

Many thanks, Will 😄

lib/reporters/verbose.js
```diff
@@ -245,6 +251,39 @@ class VerboseReporter {
...
+	 */
+	runningTestSummary() {
+		console.log('runningTestSummary', '🍇');
+		if (this.runningTestsCount === 0) {
+			return 'NO TESTS 🍓';
+		}
+		let smy = ` ${this.runningTestsCount} tests still running`;
+		this.runningTestFiles.forEach((tests, file) => {
+			for(let test of this.runningTestFiles.get(file)) {
+				smy += `\n${figures.circleDotted} ${file}:${test}`;
```
novemberborn: Use `this.prefixTitle(file, title)` instead.

lib/reporters/verbose.js
```diff
@@ -245,6 +251,39 @@ class VerboseReporter {
...
+	 * @returns string
+	 */
+	runningTestSummary() {
+		console.log('runningTestSummary', '🍇');
+		if (this.runningTestsCount === 0) {
+			return 'NO TESTS 🍓';
+		}
+		let smy = ` ${this.runningTestsCount} tests still running`;
+		this.runningTestFiles.forEach((tests, file) => {
+			for(let test of this.runningTestFiles.get(file)) {
```
novemberborn: Change `test` to `title`.

lib/reporters/verbose.js
```diff
@@ -245,6 +251,39 @@ class VerboseReporter {
...
+	/**
+	 * Generates the summary string of tests that are still running
+	 * @returns string
+	 */
+	runningTestSummary() {
+		console.log('runningTestSummary', '🍇');
+		if (this.runningTestsCount === 0) {
+			return 'NO TESTS 🍓';
+		}
+		let smy = ` ${this.runningTestsCount} tests still running`;
```
novemberborn: You don't need to add the space. However let's put a `\n` in front and put a colon after `running`.

novemberborn:
> I'll push a commit which restores it for the timeout tests.

@vancouverwill could you [allow edits from maintainer](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/) on this PR so I can push my fix?

vancouverwill:
Hey @novemberborn thanks for those comments, will look into them. With regard to access it looks like it already is set to that as far as I can see 😄 

image

vancouverwill:
hey @novemberborn I made the changes you suggested and the different files timeout separately with their own respective error messages.

Wasn't sure what you guys prefer to do for commit messages, whether you like to squash them or keep the major ones around for histories sake, they are in there for now but happy to squash them if you prefer.

Thanks, Will

vancouverwill:
ahh that's awesome. thanks for your patience with the back and forth, I know there has been a fair bit of chat but it is the first time me working on the project so there is bound to more questions than there would normally be.

Anyway cheers for the help, I'm definitely happy to get it out. Our team use ava all the time and being able to see which files have timed out will be really useful

novemberborn:
@vancouverwill awesome.

Was about to merge this and then I realized the PR said this would work on interrupts too, and I don't think we're testing that yet. Will have a look to see what state this PR is actually in, and then merge it with the correct commit message 😉

vancouverwill:
ahh true, I was so focused on timeouts but that was part of the orginal
plan. Should be able to reuse a lot of the same logic for timeouts tho so I
can add the code on here or better yet get this out and we can do another
pr for the interrupts?

novemberborn:
> better yet get this out and we can do another
> pr for the interrupts?

Yup that's the plan. Just need to have another look at it.

PR #1887, opened 2018-08-06
Fake TTY support for stderr

### LINKED ISSUE #1494: Correctly proxy TTY for stderr
See https://github.com/avajs/ava/issues/1494#issuecomment-325351227

---

## Description

When using [visionmedia's debug](https://github.com/visionmedia/debug) with AVA the colourized output is lost. I desire to turn on debug when running my tests to observe the code paths taken. I believe it is not working because the `process.stderr.isTTY` property is `undefined` when running under AVA. I'd like to make a feature request to support colorized output of the `debug` module by setting `process.stderr.isTTY` to `true` when the initiating `ava` process has `process.stderr.isTTY` set to `true`;

### Test Source

```js
// place this code in colorize-test.js

process.env.DEBUG='worker'
const debug = require('debug')('worker')

debug('doing lots of uninteresting work');

console.log(`stderr is tty? ${process.stderr.isTTY}`);

// if we are running in AVA create a test to avoid 'No tests found' exception
if (process.env.AVA_PATH) {
    const test = require('ava');
    test('exit', async (t) => t.is(true, true));
}
```

### Command-Line Arguments

```
$ node colorize-test.js # has colored output as stderr is tty
  worker doing lots of uninteresting work +0ms
stderr is tty? true

$ ava colorize-test.js # does not have colored output as stderr is not a tty
2017-08-24T10:57:04.500Z worker doing lots of uninteresting work
stderr is tty? undefined

  1 passed
```

## Relevant Links

The [pull request](https://github.com/visionmedia/debug/pull/2) that requires `process.stderr.isTTY` to be `true` in order to print colored output.

## Environment

```
Node.js v8.4.0
Ubuntu 16.04.2 4.10.0-32-generic
AVA 0.22.0
```

jagoda:
Fixes #1494

### DIFF
--- a/lib/worker/fake-tty.js
+++ b/lib/worker/fake-tty.js
@@ -2,19 +2,25 @@
 const tty = require('tty');
 const options = require('./options').get();
 
-if (options.tty) {
-	Object.assign(process.stdout, {isTTY: true}, options.tty);
+const fakeTTYs = new Set();
 
-	if (options.tty.colorDepth !== undefined) {
-		process.stdout.getColorDepth = () => options.tty.colorDepth;
+const isatty = tty.isatty;
+tty.isatty = fd => fakeTTYs.has(fd) || isatty(fd);
+
+const simulateTTY = (stream, {colorDepth, columns, rows}) => {
+	Object.assign(stream, {isTTY: true, columns, rows});
+
+	if (colorDepth !== undefined) {
+		stream.getColorDepth = () => colorDepth;
 	}
+};
 
-	const isatty = tty.isatty;
-	tty.isatty = function (fd) {
-		if (fd === 1 || fd === process.stdout) {
-			return true;
-		}
+if (options.tty.stderr) {
+	simulateTTY(process.stderr, options.tty.stderr);
+	fakeTTYs.add(2);
+}
 
-		return isatty(fd);
-	};
+if (options.tty.stdout) {
+	simulateTTY(process.stdout, options.tty.stdout);
+	fakeTTYs.add(1);
 }
--- a/test/fixture/tty/color-disabled.js
+++ b/test/fixture/tty/color-disabled.js
@@ -1,6 +1,9 @@
 import test from '../../..';
 
-test('stdout does not support color', t => {
-	t.true(process.stdout.isTTY);
-	t.is(process.stdout.getColorDepth(), 1);
-});
+const assertNoColor = (t, stream) => {
+	t.true(stream.isTTY);
+	t.is(stream.getColorDepth(), 1);
+};
+
+test('stderr does not support color', assertNoColor, process.stderr);
+test('stdout does not support color', assertNoColor, process.stdout);
--- a/lib/fork.js
+++ b/lib/fork.js
@@ -23,6 +23,12 @@ if (env.NODE_PATH) {
 // the presence of this variable allows it to require this one instead
 env.AVA_PATH = path.resolve(__dirname, '..');
 
+const describeTTY = tty => ({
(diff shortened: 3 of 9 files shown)


lib/worker/fake-tty.js
```diff
@@ -2,19 +2,30 @@
...
-		return isatty(fd);
-	};
+	if (options.colorDepth !== undefined) {
+		stream.getColorDepth = () => options.colorDepth;
+	}
+};
+
+if (options.tty.stderr) {
+	simulateTTY(process.stderr, options.tty.stderr);
+	fakeTTYs.push(2, process.stderr);
```
novemberborn: Why do you need to push `process.stderr` (and `stdout`, below)?

In `isatty`, `fd` is supposed to be a number according to the Node.js documentation, so pushing the numbers should be sufficient.

lib/worker/fake-tty.js
```diff
@@ -2,19 +2,30 @@
...
 
-if (options.tty) {
-	Object.assign(process.stdout, {isTTY: true}, options.tty);
+const fakeTTYs = [];
 
-	if (options.tty.colorDepth !== undefined) {
-		process.stdout.getColorDepth = () => options.tty.colorDepth;
+const isatty = tty.isatty;
+tty.isatty = fd => {
+	if (fakeTTYs.indexOf(fd) !== -1) {
```
novemberborn: `fakeTTYs` may be better as a `Set` rather than an array. You could then simplify this function to `fd => fakeTTYs.has(fd) || isatty(fd)`.

lib/worker/fake-tty.js
```diff
@@ -2,19 +2,30 @@
...
+	return isatty(fd);
+};
 
-	const isatty = tty.isatty;
-	tty.isatty = function (fd) {
-		if (fd === 1 || fd === process.stdout) {
-			return true;
-		}
+const simulateTTY = (stream, options) => {
+	Object.assign(stream, {isTTY: true}, options);
```
novemberborn: This assigns `colorDepth`. Better to destructure options and assign just `columns` and `rows`:

```js
const simulateTTY = (stream, {colorDepth, columns, rows}) => {
  Object.assign(stream, {isTTY: true, columns, rows})

  if (colorDepth !== undefined) {
```

test/fixture/tty/color-disabled.js
```diff
@@ -1,6 +1,9 @@
...
-test('stdout does not support color', t => {
-	t.true(process.stdout.isTTY);
-	t.is(process.stdout.getColorDepth(), 1);
-});
+const assertNoColor = stream => t => {
+	t.true(stream.isTTY);
+	t.is(stream.getColorDepth(), 1);
+};
+
+test('stderr does not support color', assertNoColor(process.stderr));
```
novemberborn: In all of these, you could write the assertion method like this:

```js
const assertNoColor = (t, stream) => {
  // assertions here
}

test('title', assertNoColor, process.stderr)
```

This uses [test macros](https://github.com/avajs/ava#test-macros).

jagoda: Oh, very cool. I hadn't found that feature yet—thanks for the tip!

lib/worker/fake-tty.js
jagoda: That makes sense to me. I included the `stdout` and `stderr` objects here to maintain consistency with the original code which checks both the numeric file descriptor and the stream object. If the tests pass then I'm happy to drop the stream object comparison.

jagoda:
Thanks for the feedback @novemberborn! I believe that I have addressed everything that you commented on. Please let me know if I missed anything.

novemberborn:
Thank you so much @jagoda!

PR #1889, opened 2018-08-07
Ignore node_modules/* in example

ronen:
If someone were to blindly copy & paste the example (as I did), things go bad when node spawns too many processes trying to compile files under `node_modules`.


novemberborn:
Good call, thanks @ronen!

PR #1890, opened 2018-08-10
Fix typo in assertExpectations function

ivanschwarz:


novemberborn:
Nice catch @ivanschwarz!

PR #1900, opened 2018-08-13
Include `verbose` option in Configuration section of README

ronen:
Configuration section of README says "All of the CLI options can be configured", and goes on to provide a detailed spec which doesn't include the `verbose` option


novemberborn:
Thanks!

PR #1910, opened 2018-08-18
Fix esm 3.0.75 test compat

novemberborn:
Fixes https://travis-ci.org/avajs/ava/jobs/415819693.

See also https://github.com/standard-things/esm/issues/569.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2684,9 +2684,9 @@
       "dev": true
     },
     "esm": {
-      "version": "3.0.74",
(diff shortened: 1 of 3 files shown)


novemberborn:
This `esm` behavior has now been reverted. If we can deal with the partial test failures I may just change this PR to update to the next `esm` release, without changing the helper code.

sindresorhus:
> If we can deal with the partial test failures I may just change this PR to update to the next esm release, without changing the helper code.

👍

novemberborn:
Closing in favor of #1920.

PR #1913, opened 2018-08-20
Add recipe for AVA & Puppeteer

### LINKED ISSUE #1799: Request for Documentation: AVA & Puppeteer
This is not a bug but a request for documentation.

I'd like a reasonably detailed document discussing use of AVA and Puppeteer where the tests are snippets sant to and run in the browser with ava in node testing that the results are correct.

My example: I have an agent based model that I want to run for 500 steps in the browser.  I then want to write ava tests against the state of the simulation. I'd like to be able to write snippets that are sent to the browser, and the result values sent back to ava for testing.

So there is no need to execute ava in the browser, but only the ability to message back/forth between it and puppeteer, using ava's sane assertions.

sh7dm:
Added recipe for using AVA with Puppeteer for testing web apps.

Work in progress, recipe may be incomplete or buggy. I think it's worth adding a few pictures and more examples. If you have an idea for enhancing this recipe, write it in the comments.

Fixes #1799

### DIFF
--- a/docs/recipes/puppeteer.md
+++ b/docs/recipes/puppeteer.md
@@ -46,4 +46,14 @@ test('page title should contain "Google"', async t => {
 test('page should contain an element with #hplogo selector', async t => {
   t.not(await t.context.page.$('#hplogo'), null);
 });
+
+test('full page should match the snapshot', async t => {
+  let fullHTML = await t.context.page.evaluate(() => document.innerHTML);
+  t.snapshot(fullHTML);
+});
+
+test('search form should match the snapshot', async t => {
+  let searchForm = await t.context.page.$('#searchform').innerHTML;
+  t.snapshot(searchForm);
+});
 ```
--- a/docs/recipes/puppeteer.md
+++ b/docs/recipes/puppeteer.md
@@ -42,4 +42,8 @@ setup(test);
 test('page title should contain "Google"', async t => {
   t.true((await t.context.page.title()).includes('Google'));
 });
+
+test('page should contain an element with #hplogo selector', async t => {
+  t.not(await t.context.page.$('#hplogo'), null);
+});
 ```
--- a/docs/recipes/puppeteer.md
+++ b/docs/recipes/puppeteer.md
@@ -8,20 +8,20 @@
 
 The first step is setting up a helper to configure the environment:
 
-`./test/helpers/setup.js`
+`./test/helpers/withPage.js`
 
 ```js
-const puppeteer = require('puppeteer');
+import puppeteer from 'puppeteer';
 
-module.exports = async fn => {
+export default async function withPage(t, run) {
 	const browser = await puppeteer.launch();
 	const page = await browser.newPage();
 	try {
-		await fn(page);
+		await run(t, page);
 	} finally {
 		await page.close();
 		await browser.close();
-	}
+	};
 }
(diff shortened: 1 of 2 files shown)


docs/recipes/puppeteer.md
```diff
@@ -0,0 +1,43 @@
...
+	- `npm i --save-dev puppeteer`
+
+## Setup
+
+The first step is setting up a helper to configure the environment:
+
+`./test/helpers/setup.js`
+
+```js
+const { test } = require('ava');
```
novemberborn: This no longer works. Use `const test = require('ava')`.

docs/recipes/puppeteer.md
```diff
@@ -0,0 +1,43 @@
...
+const puppeteer = require('puppeteer');
+const url = "https://google.com"; // App URL, for example, google.com
+
+global.test = test;
+
+test.before(async () => {
+  global.browser = await puppeteer.launch();
+  global.page = await browser.newPage();
```
novemberborn: We're not very keen on globals with AVA. Instead, export a `setup` function that takes a `test` object. Then use that to hook up `before` and `after`.

This function also gets a `t` object. I would assign `browser` and `page` to `t.context.browser`, etc. Then in the tests you can access them via `t.context` and you can avoid the globals.

docs/recipes/puppeteer.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Setup
+
+The first step is setting up a helper to configure the environment:
+
+`./test/helpers/setup.js`
+
+```js
+const puppeteer = require('puppeteer');
+const url = "https://google.com"; // App URL, for example, google.com
```
novemberborn: It might be neat to be able to pass the URL to the `setup` function instead.

sh7dm: Yes, it will be awesome to add some more examples

docs/recipes/puppeteer.md
```diff
@@ -0,0 +1,58 @@
...
+`./test/helpers/setup.js`
+
+```js
+const puppeteer = require('puppeteer');
+
+module.exports = (test, url) => {
+  test.before(async t => {
+    t.context.browser = await puppeteer.launch();
+    t.context.page = await t.context.browser.newPage();
+    await t.context.page.goto(url);
```
lo1tuma: Sharing the browser/page between tests might cause unexpected issues because all tests will run in parallel. What I usually do in my ava/puppeteer tests is to start a separate browser instance for each test.

```js
async function withBrowserPage(fn) {
    const browser = await puppeteer.launch();
    const page = browser.newPage();
    try {
        await fn(page);
    } finally {
        await page.close();
        await browser.close();
    }
}

test('foo', t => {
    return withBrowserPage(page => {
         page.goto('https://google.com');
         t.is(await page.title(), 'Google');
    });
});

test('bar', t => {
    return withBrowserPage(page => {
         page.goto('https://bing.com');
         t.is(await page.title(), 'Bing');
    });
});
```

novemberborn:
@sindresorhus @sh7dm happy with my changes?

sindresorhus:
@novemberborn 👍

sindresorhus:
Thanks for contributing, @sh7dm 🎉

PR #1915, opened 2018-08-21
Add version info for testing ts file directly

Asing1001:
## Why need this change?
Having version info on the document so people won't waste time on trying it with an incompatible version.

### DIFF
--- a/docs/recipes/typescript.md
+++ b/docs/recipes/typescript.md
@@ -8,7 +8,7 @@ This guide assumes you've already set up TypeScript for your project. Note that 
 
 ## Configuring AVA to compile TypeScript files on the fly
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
This is already clearly documented at the top of the readme:

> This documentation covers the 1.0 beta releases.

novemberborn:
Hey @Asing1001, sorry you ran into this. Until we place the documentation on a separate website it's hard to clearly communicate what version it applies to. Unfortunately it's not manageable for us to include versions everywhere we mention a feature.

Hopefully we can ship a 1.0 release as `latest` soon, and do more frequent releases afterwards, which should alleviate this problem.

PR #1936, opened 2018-09-13
Fix profile.js was not able to work without babel

### LINKED ISSUE #1819: profile.js script does not respect extension configuration
The `profile.js` script will apply the Babel configuration to whatever file it's called with, regardless of `extensions` configuration.

This shouldn't be too dreadful to fix, but we'd rather do #1505.

goooseman:
Fixes #1819

When we need to debug our code, we launch ava not from cli, but using profile.js.

In this file the `precompileFile` was hardcoded, so files were always transpiled through babel, which was the issue with projects on typescript, where we do not need babel at all.

I took code api.js to choose when to use babelPipeline and when not.

There is some code duplication problem now between `api.js` and `profile.js`, which can not be easily refactored.

### DIFF
--- a/profile.js
+++ b/profile.js
@@ -11,10 +11,12 @@ const Promise = require('bluebird');
 const uniqueTempDir = require('unique-temp-dir');
 const arrify = require('arrify');
 const resolveCwd = require('resolve-cwd');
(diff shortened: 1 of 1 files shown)


novemberborn:
Thanks @goooseman!

Issue #1932, opened 2018-09-09
Provide useful feedback when using t.throws() with async functions


novemberborn:
As of [1.0.0 beta 7](https://github.com/avajs/ava/releases/tag/v1.0.0-beta.7), `t.throws()` only works with synchronous functions. We should detect when it's used with a promise and provide useful feedback. I also think we should add a no-op `.catch()` handler to the promise to prevent AVA from logging an unhandled rejection.

See: https://github.com/avajs/ava/blob/bfe16300e07c12a1bad2e4f02898b1e7486daf70/lib/assert.js#L341 and compare with: https://github.com/avajs/ava/blob/bfe16300e07c12a1bad2e4f02898b1e7486daf70/lib/assert.js#L365

Thanks to @Lukenickerson for raising this in #1931.

grant37:
I'd like to work on this if that's alright. Messing around locally and trying to follow the guidelines here I had success (I think) with: 
1. Checking if the `fn` argument to `throws()` returns a promise (and adding the no-op catch here)
2. If it does, calling `fail()` like it's called on line 314 with a message about asynch functions
3. Wrapping all this in a try/catch (no-op) because I got errors on the promise check if `fn` was synchronous (probably there's a better method though?)

The result is a test failed message that refers to the line `await t.throws(asynch () => {` in a test file (from #1931), which is all I tested against so far. Would this be the correct behavior/do you see any issues?

novemberborn:
@grant37 sounds like you're nearly there! Please make the PR and we can get it in.

Issue #1931, opened 2018-09-07
When using an async function in `t.throws()`, receive errors: "Function returned: Promise {}" and "Unhandled rejection"


Lukenickerson:
## Description

In `1.0.0-beta.8` (with no special configuration or command line arguments) I'm seeing an error with essentially the same function that was posted in #1371.

### Test Source

```javascript
import test from 'ava';

test('test', async (t) => {
  await t.throws(async () => {
    throw new Error();
  });
});
```

### Error Message & Stack Trace

Message returned from running the test says something like:

> 1 test failed
> 1 unhandled rejection

> Line: await t.throws(async () => {
> Function returned: Promise {}

> Line: throw new Error();
> Unhandled rejection in _[test filename]_

## Environment

Node.js v10.9.0
darwin 17.6.0
AVA 1.0.0-beta.8
npm 6.2.0

novemberborn:
Ah sorry about that. As of [Beta 7](https://github.com/avajs/ava/releases/tag/v1.0.0-beta.7), `t.throws()` only works for synchronous errors. You'll need to change your test to use `t.throwsAsync()`.

However we should provide better feedback in the test output. I've opened #1932 with some pointers.

PR #1940, opened 2018-09-19
Providing useful feedback when using t.throws() with async functions

grant37:
This is for #1932.

Here's the overview of what the lines do:
- Checks if the `fn` argument to `t.throws()` returns a promise and adds a no-op `.catch()` handler to the return value prevent ava from logging an unhandled rejection
- If it returns a promise, calls `fail()` with the message "`t.throws()` cannot be called with an asynchronous function. Use `t.throwsAsync()` instead"
- Wraps all this in a try-catch statement in order to add the `.catch()` handler in the conditional

I used the test case from #1931:
```
import test from 'ava';

test('test', async (t) => {
  await t.throws(async () => {
    throw new Error();
  });
});
```

The output running npm test is:
```
   15: test('test', async (t) => {     
   16:     await t.throws(async () => {
   17:         throw new Error();      

  `t.throws()` cannot be called with an asynchronous function. Use `t.throwsAsync()` instead

  Called with:

  AsyncFunction []
```
I'd be happy to do this a different way if anyone has suggestions.

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -331,7 +331,7 @@ function wrapAssertions(callbacks) {
 					return;
 				}
 			} catch (_) {}
-			
+
 			try {
 				expectations = validateExpectations('throws', expectations, arguments.length);
 			} catch (error) {
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -333,7 +333,9 @@ function wrapAssertions(callbacks) {
 			try {
 				retval = fn();
 				if (isPromise(retval)) {
-					retval.catch(noop);
+					try {
+						retval.catch(noop);
+					} catch (_) {}
 					fail(this, new AssertionError({
 						assertion: 'throws',
 						message: 'Function threw asynchronously. Use `t.throwsAsync()` instead'
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -321,6 +321,17 @@ function wrapAssertions(callbacks) {
 			}
 
 			try {
+				if (isPromise(fn().catch(_ => {}))) {
+					fail(this, new AssertionError({
+						assertion: 'throws',
+						improperUsage: true,
+						message: '`t.throws()` cannot be called with an asynchronous function. Use `t.throwsAsync()` instead',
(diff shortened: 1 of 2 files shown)


lib/assert.js
```diff
@@ -320,6 +320,18 @@ function wrapAssertions(callbacks) {
 				return;
 			}
 
+			try {
+				if (isPromise(fn().catch(_ => {}))) {
```
novemberborn: You should move this condition down to after the `retval = fn()` line, and check if `isPromise(retval)`. Currently this calls `fn` twice, which is not what we want.

I'd only hook up the no-op `catch()` once we know `retval` is a promise. There's a `noop` function already declared that you can use.

lib/assert.js
```diff
@@ -320,6 +320,18 @@ function wrapAssertions(callbacks) {
...
+					fail(this, new AssertionError({
+						assertion: 'throws',
+						improperUsage: true,
+						message: '`t.throws()` cannot be called with an asynchronous function. Use `t.throwsAsync()` instead',
+						values: [formatWithLabel('Called with:', fn)]
```
novemberborn: I don't think printing the `fn` function is useful in this scenario. Let's remove `values` altogether.

grant37: Thanks for the feedback! I can work on making these changes this week.

grant37:
The check fails for Node.js:6 with
```
/assert.js:718:
    assertions.throws(async () => {
		            ^

SyntaxError: Unexpected token (
```

But using the arrow syntax seems to be required. Any thoughts on how to work around this?

novemberborn:
@grant37 Node.js 6 doesn't support async functions. Use a regular function (or arrow function) that returns a promise instead.

novemberborn:
Thanks @grant37!

I don't think the messaging is quite right. The problem is that the function returned a promise. We don't actually know if it rejected or not. I'll push a commit that changes the messaging, please let me know what you think.

I've also added a `try` / `catch` around the `retval.catch(noop)` call, in case the object isn't quite a promise.

PR #1941, opened 2018-09-26
Replace CI parallelism logic with package

jamiebuilds:
I needed this logic in a couple different places so I turned it into a package with tests and shit

### DIFF
--- a/api.js
+++ b/api.js
@@ -13,6 +13,7 @@ const getPort = require('get-port');
 const arrify = require('arrify');
 const makeDir = require('make-dir');
 const ms = require('ms');
+const chunkd = require('chunkd');
 const babelPipeline = require('./lib/babel-pipeline');
 const Emittery = require('./lib/emittery');
 const RunStatus = require('./lib/run-status');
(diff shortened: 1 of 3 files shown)


api.js
```diff
@@ -84,18 +85,15 @@ class Api extends Emittery {
...
-
 					const {currentIndex, totalRuns} = this.options.parallelRuns;
 					const fileCount = files.length;
-					const each = Math.floor(fileCount / totalRuns);
-					const remainder = fileCount % totalRuns;
 
-					const offset = Math.min(currentIndex, remainder) + (currentIndex * each);
-					const currentFileCount = each + (currentIndex < remainder ? 1 : 0);
+					// The files must be in the same order across all runs, so sort them.
+					files = files.sort((a, b) => a.localeCompare(b, [], {numeric: true}));
```
sindresorhus: Couldn't the sorting here be part of `chunkd` too? Maybe as an opt-in option.

jamiebuilds: I need to be able to configure the comparator, but I guess it could do that too.

novemberborn:
Nice!

PR #1946, opened 2018-10-06
Change babel-register to @babel/register

xxczaki:
I created this PR because:
> This documentation covers the 1.0 beta releases, **which use Babel 7.**


novemberborn:
Thanks!

Issue #1692, opened 2018-02-06
Conditionally perform assertions, allow tests to commit to the results


danr:
Please see https://github.com/avajs/ava/issues/1692#issuecomment-363750906 for the accepted proposal and https://github.com/avajs/ava/issues/1692#issuecomment-366521743 for follow-up.

---

Original issue follows:

## Description: add observeTest and commitToResult

Right now we have a function `test` of which the essence has this type signature:

```typescript
const test: (t: TestContext) => PromiseLike<void> | Iterator<any> | Observable | void
```

It would be helpful for library writers if there was a function:

```typescript
const observeTest: 
  (test: (t: TestContext) => PromiseLike<void> | Iterator<any> | Observable | void) => 
  {success: boolean, test_result: TestResult}
```
_edit: fixed the type signature_

For my purposes `TestResult` can be a completely opaque type.
The use for it is to make this a result for the main driver. For the lack of a better name I will call it `commitToResult` for now. This could either be a function exported by AVA (or a method on AVA's exported `test`), or a method on `TestContext`:

```typescript
const commitToResult: (test_result: TestResult) => void
```
or
```typescript
interface TestContext {
    ...
    // stop the execution of the test now reporting this result:
    commitToResult: (test_result: TestResult) => void
}
```

## Use case: property based testing

The use case for this is property-based testing, which was first introduced in Haskell with the package [QuickCheck](https://hackage.haskell.org/package/QuickCheck).

There are quite a few implementations for JavaScript, for example [jsverify](https://github.com/jsverify/jsverify), [testcheck-js](https://github.com/leebyron/testcheck-js). The idea is to generate random data and test if functions satsify properties run on these data to approximate universal quantification.  If the test fails it is common practice in these libraries to try to _shrink_ the generated data.  Then the user can be presented with a small (and hopefully minimal) counterexample.

The library writer of a property-based testing would then run tests written using asserts as in AVA-style, but behind the hoods it would be run with `observeTest`, and if it fails start a loop trying to shrink the randomly generated input until it cannot be shrunk anymore and then report the smallest possible failing `TestResult` using `commitToResult`.

There is intergration for AVA in
[...]

Issue #1585, opened 2017-11-13
Fail when new snapshots are created in CI


novemberborn:
When running tests in CI environments, if new snapshots are created it implies the tests are behaving differently then during development. This can hide bugs. The snapshot assertions should fail, rather than silently creating the snapshot and passing.

@avajs/core and others, what do you think?

impaler:
It seems like a reasonable requirement to prevent false positives. I think an error message for this would need to describe the reasoning and fix, something along the lines of:

>  "Snapshot {x} is missing a fixture file. To prevent false posititives, a verified fixture file is required when testing in the CI environment."

novemberborn:
How about:

> Could not find snapshot. Note that snapshots are not created automatically when in a CI environment.

(This would be shown as the message for the `t.snapshot()` assertion failure.)

sindresorhus:
👍

AJamesPhillips:
Thanks all for the excellent contributions to date, much appreciated!Any thoughts or progress on this?  I wasn't clear on how you would tell you are in a ci environment, would this behaviour just be enabled by default, and to generate the new snapshot files you would have to pass the `-u` flag?  This would be a breaking change and would not score great on the "self documenting options".  Was something different being considered.

novemberborn:
@AJamesPhillips would you be willing to help out with this?

AJamesPhillips:
Hi @novemberborn , I had a quick look at the code.  I would prefer to add a flag which if set, will fail the tests if a snapshot can not be located, as opposed to only failing if on a ci environment and a snapshot can not be located.
I just had a case now where our tests passed when they should have failed.  In this case they were run locally inside a docker container.  This would not be caught by the `is-ci` check proposed.  Any thoughts on this approach (they're not mutually exclusive so we could do both at some point)?  Thanks.

novemberborn:
> In this case they were run locally inside a docker container.

If you run tests in a Docker container, how do you update the snapshots?

The idea is that during development you'd notice the new snapshot files (since you need to commit them). I think if we implement the CI flag then you could set `CI=1` in your container and AVA will treat it as such. Would that be sufficient?

AJamesPhillips:
[...]

PR #1947, opened 2018-10-06
Add try-commit to test assertions

qlonik:
Hello. Some time ago there was a discussion in the #1692 issue. I tried to look around the code and see what could be done. I tried doing this but I'm not sure if it is the best solution.

This method creates new instance of `Test` class and runs the attempted assertion as part of that and then it updates main test if the attempted test passes or fails.
There were suggestions in the issue to create some `Host` class, which `Test` class will extend. Such `Host` class will only care about running assertions, while `Test` class will care about counting assertions and taking care of snapshots. I tried with `Host` class at first, however, I found that assertions are getting bound to the `Test`, and I was not sure which parts of `Test` need to be moved to `Host` and which can stay, without breaking assertions.
Also, creating new `Host` class, would require to re-implement the `run()` function and all of the edge cases, which are already figured out in the current implementation of `Test()`.

There were also mentions about taking care of `plan()` assertion and about recursiveness of `try()` assertion. Also, snapshots were excluded from discussion, but I think they should be talked about too.
Since the current implementation of `Test` can handle all of these cases, it means reusing `Test` class as an instance for `try()` function should allow for those three things to work. There was also a comment about supporting the `.discard()` on promise returned from `.try()`, which is also added.

With the current implementation, each attempt has to be explicitly either accepted with `.commit()` or declined with `.discard()`. When the result is committed to, it increase assertion counter and copies the log of the attempt into main test. When the result is discarded, the assertion counter is not increased. It seems that it would be possible to remove explicit discard on the attempt result. However, in that case, there will be a requirement that user need to have at least one call to `.commit()` in order to pass the test.

I'm open to some suggestions about current implementation and to guidance how to make `Host` class work, if this is more desirable way.

Example usages with current implementation:
```js
const twoRndInts = () => {
  const rnd = Math.round(Math.random() * 100);
  const x = rnd % 10;
  const y = Math.floor(rnd / 10);
  return [x, y];
};

test('test', async t => {

  const result = await t.try((t, a, b) => {
    t.is(a, b);
  }, ...twoRndInts());

  if (result.passed) {
    result.commit();
  } else {
    t.log(result.error);
    result.discard();

    const result1 = await t.try((t, a, b) => {
      t.is(a, b);
    }, ...twoRndInts());

    result1.commit();
  }
});

// EDIT: the following usage scenario with discarding incomplete attempt is not supported
test('test2', t => {
  const result = t.try((t, a, b) => {  // <-- result here is a pending promise
    return new Promise(res => setTimeout(res, 1000))
      .then(() => t.is(a, b));
  }, ...twoRndInts());

  return new Promise(res => setTimeout(res, 100))
    .then(() => {
      result.discard();
    });
});
```

## TODO:

(Updated from https://github.com/avajs/ava/pull/1947#issuecomment-462152014)

- [x] Test the context (`t.context`)
- [x] Update test assertCount with the number of assertions made within a committed attempt
- [x] Add support of macros passed to try fn
- [x] Type t.try() so it accepts macros
- [x] Investigate refactor so attempts don't need to instantiate a new Test
    - [x] If not possible, add metadata to indicate the test is "inline" and adjust handling of test.failing() and t.end() based on that
- [x] Revisit how attempt titles are generated
- [x] Ensure snapshots created in each attempt start at the correct invocation count
- [x] Fail the test if concurrent attempts that created snapshots are committed
- [x] Move try-commit tests into its own file
- [ ] Determine documented properties of AssertionError and add to type definitions
- [x] Update Flow definition
- [ ] Update ESLint plugin to recognize the assertion
- [ ] File issue to discuss suggested argument name (tt?) for inline functions, and update power-assert     
   - [ ] rewrite rules to handle it
- [x] Review test coverage
- [ ] Update documentation


novemberborn:
Oh wow, this is very exciting @qlonik!

I must admit I've kind of forgotten how `Test` and assertions work… It may take me a little while to review this and figure out what we need to do in order to land this PR. There's some other Babel related work that needs to take priority, we really need to get the 1.0 release out.

I'm terribly sorry about that. I'm rather geeked about shipping this feature!

lib/test.js
```diff
@@ -53,6 +53,64 @@ function plan(count) {
...
+
+		const commitFn = toCommit => {
+			if (commitCalled) {
+				this.saveFirstError(new Error('Either commit() or discard() was already called'));
+				return;
+			}
+			commitCalled = true;
+
+			if (ret.passed) {
+				this.countFinishedAttemptAssertion(toCommit);
```
qlonik: Here, the `pendingAttemptCounter`is decremented, which means that it is required to call either `'commit()` or `discard()` for every attempt. Instead it could be decremented as part of `addPendingAttemptAssertion()` and then we will require at least one call to `commit()` and no need to have `discard()` and no need to explicitly call either commit or discard on every attempt.

novemberborn:
Go for it. I’ve been terribly busy of late, hoping to have a look after next week.

qlonik:
Its okay if it takes time. Its holidays after the next week anyway, or something :)

I've been using snapshot of ava forced on commit that includes these changes. It worked for me quite well. It will be cool when it makes it into ava :)

lib/test.js
```diff
@@ -53,6 +53,77 @@ function plan(count) {
...
+	const args = [];
+	for (let i = 1, len = arguments.length; i < len; i++) {
+		args.push(arguments[i]);
+	}
+
+	let commitCalled = false;
+	let discarded = false;
+	let finished = false;
+
+	const commitFn = (ret, toCommit) => {
```
novemberborn: Could you remove the `toCommit` argument? It's odd to pass `false` in order to get discard behavior. That should be a separate function.

lib/test.js
```diff
@@ -53,6 +53,77 @@ function plan(count) {
...
+			this.countFinishedAttemptAssertion(toCommit);
+		} else {
+			this.countFailedAttemptAssertion(ret.error, toCommit);
+		}
+
+		if (toCommit) {
+			for (const log of ret.logs) {
+				this.addLog(log);
+			}
+		}
```
novemberborn: It might be better to have one `commitAttempt(ret)` function call which performs all the relevant actions based on `ret`'s state.

lib/test.js
```diff
@@ -53,6 +53,77 @@ function plan(count) {
...
+	};
+
+	const running = this
+		.forAttempt(t => fn.apply(null, [t].concat(args)))
```
novemberborn: `t => fn(t, ...args)` should work.

lib/test.js
```diff
@@ -53,6 +53,77 @@ function plan(count) {
...
+	this.addPendingAttemptAssertion();
+
+	const finishedRunning = running.then(ret => {
+		if (discarded) {
+			return null;
+		}
+
+		finished = true;
+		return Object.assign({}, ret, {
+			commit: () => commitFn(ret, true),
```
novemberborn: This seems to be missing a few properties that are declared in the TypeScript definition.

lib/test.js
```diff
@@ -53,6 +53,77 @@ function plan(count) {
...
+			discard: () => {
+				discarded = true;
+				return commitFn(ret, false);
+			}
+		});
+	});
+
+	finishedRunning.discard = () => {
+		if (finished) {
+			this.saveFirstError(new Error('Attempt is already resolved'));
```
novemberborn: I think that's fine, it's only a problem if it was committed rather than discarded.

lib/test.js
```diff
@@ -53,6 +53,77 @@ function plan(count) {
...
+			return;
+		}
+		if (discarded) {
+			this.saveFirstError(new Error('The discard() was already called'));
```
novemberborn: Perhaps this should be a no-op.

novemberborn:
Another thought. If we're limiting to one concurrent `t.try()` per test, we should explore reusing the `t` value inside the callback. That way you could write:

```javascript
test('inline', async t => {
  const result = await t.try(() => {
    t.pass()
  })
  result.commit()
})

const reusable = t => t.pass()
test('reuse', async t => {
  const result = await t.try(reusable)
  result.commit()
})
```

We prefer the `t` variable name is used for the execution context, so it'd be confusing if an outer context is used in an inline `t.try()` callback.

index.d.ts
```diff
@@ -369,6 +370,38 @@ export interface PlanFn {
...
+	metadata: {
+		always: boolean,
+		callback: boolean,
+		exclusive: boolean,
+		failing: boolean,
+		serial: boolean,
+		skipped: boolean,
+		todo: boolean,
+		type: 'test',
+	},
```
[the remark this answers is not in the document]
qlonik: Maybe `duration` and `metadata` are not.
However, `logs` and `title` are helpful for the case when `t.try` is used inside the macro provided by third party. (This is described in more detail in parent comment)

lib/test.js
```diff
@@ -53,6 +53,77 @@ function plan(count) {
 	this.plan(count, captureStack(this.plan));
 }
 
+function tryTest(fn) {
+	const args = [];
+	for (let i = 1, len = arguments.length; i < len; i++) {
+		args.push(arguments[i]);
+	}
```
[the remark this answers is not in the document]
qlonik: :+1:

lib/test.js
qlonik: I added this so that there is no duplicate code between `commit()` and `discard()` functions returned to the user. The user dont have access to call this function directly, it is proxied. I will remove it though, for clarity between `commit()` and `discard()` functions.

lib/test.js
```diff
@@ -53,6 +53,77 @@ function plan(count) {
...
+
+	let commitCalled = false;
+	let discarded = false;
+	let finished = false;
+
+	const commitFn = (ret, toCommit) => {
+		if (commitCalled) {
+			if (discarded) {
+				this.saveFirstError(new Error('The discard() was already called'));
```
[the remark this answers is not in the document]
qlonik: :+1:

lib/test.js
```diff
@@ -53,6 +53,77 @@ function plan(count) {
...
+			} else {
+				this.saveFirstError(new Error('The commit() was already called'));
```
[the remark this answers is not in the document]
qlonik: It could be. I'm not sure how it would be better. I was trying to be explicit so that there is not more `commit()`/`discard()` calls than needed.

Also, while trying the solution, it was possible to have discard be implicitly default behavior. So that `t.try()` would transparently be not an assertion, until `commit()` is called. This way `discard()` is not needed at all. I'm not sure if this way is better, or if it should be more explicit like it is now.

lib/test.js
qlonik: Could you elaborate?

lib/test.js
qlonik: :+1:

lib/test.js
qlonik: they come from the `ret` value.

lib/test.js
qlonik: So should it not throw an error and simply return?

lib/test.js
```diff
@@ -143,6 +218,20 @@ class Test {
...
 		this.planCount = null;
 		this.startedAt = 0;
+		this.pendingAttemptCount = 0;
+		this.attemptCount = 0;
+
+		this.forAttempt = fn => {
+			const opts = Object.assign({}, options, {
+				metadata: Object.assign({}, options.metadata, {
+					callback: false,
+					failing: false
```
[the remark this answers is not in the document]
qlonik: This way was simpler. I'll try to change

lib/test.js
```diff
@@ -143,6 +218,20 @@ class Test {
...
+				}),
+				fn,
+				title: this.title + '.A' + (this.attemptCount++)
+			});
+			return new Test(opts);
```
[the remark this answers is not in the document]
qlonik: I was adjusting the title in order to prevent any conflicts related to snapshots. Isn't it the case that if we adjust the title, then there are no problems about number of concurrent `t.try()` and both may contain snapshots?

lib/test.js
```diff
@@ -213,6 +302,34 @@ class Test {
 		this.saveFirstError(error);
 	}
 
+	addPendingAttemptAssertion() {
+		if (this.finishing) {
+			this.saveFirstError(new Error('Assertion passed, but test has already finished'));
```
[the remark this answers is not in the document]
qlonik: Should it be `Attempt is complete, but the test has already finished`?

lib/test.js
```diff
@@ -213,6 +302,34 @@ class Test {
...
+		this.pendingAttemptCount++;
+	}
+
+	decrementAttemptAssertion() {
+		this.pendingAttemptCount--;
+	}
+
+	countFinishedAttemptAssertion(includeCount) {
+		if (this.finishing) {
+			this.saveFirstError(new Error('Assertion passed, but test has already finished'));
```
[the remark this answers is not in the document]
qlonik: Should it be the same as above?

lib/test.js
```diff
@@ -213,6 +302,34 @@ class Test {
...
+		}
+
+		if (includeCount) {
+			this.assertCount++;
```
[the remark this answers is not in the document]
qlonik: :+1:

index.d.ts
```diff
@@ -369,6 +370,38 @@ export interface PlanFn {
...
+		...args: T
+	): AttemptReturnValue;
+
+	skip(...values: Array<any>): void;
+}
+
+// todo: would rather remove 'null |' from Promise definition
+// that is because in typescript, it will be required to check if it is not
+// null all the time.
+export type AttemptReturnValue = Promise<null | {
```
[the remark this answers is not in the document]
novemberborn: Interesting. I don't think we'd be able to stop execution, unless we implement some form of cancellation signaling.

Perhaps then we should remove discard-before-resolve for now.

lib/test.js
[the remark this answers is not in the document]
novemberborn: Discarding should be explicit. You could have a straight-forward assertion `t.true(enabled())` and then forget to commit the `t.try()` even if it failed. Then your test seems to pass even though it shouldn't have.

lib/test.js
novemberborn: The method would sit on `Test.prototype`, so you can call `this.commitAttempt(ret)`. That implementation can safely access Test internals and you wouldn't need the additional count methods etc.

lib/test.js
novemberborn: Yes of course.

lib/test.js
novemberborn: No, I meant that it shouldn't matter whether you discard the promise or the result, even if the promise has been fulfilled.

lib/test.js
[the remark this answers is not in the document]
novemberborn: The snapshots are indexed on the title, so if currently you only commit the third attempt, then later if the first attempt is committed you won't actually compare the snapshot. See also #1585.

lib/test.js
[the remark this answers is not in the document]
novemberborn: Not sure at the moment, sorry 😄

lib/test.js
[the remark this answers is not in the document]
novemberborn: Not sure at the moment, sorry 😄

index.d.ts
```diff
@@ -369,6 +370,36 @@ export interface PlanFn {
 	skip(count: number): void;
 }
 
+export interface TryFn<Context> {
+	<T extends any[]>(
+		title: string,
+		impl: (t: ExecutionContext<Context>, ...args: T) => ImplementationResult,
```
novemberborn: I'd like to see if we could type this to accept a macro type. Happy to try that out myself.

lib/test.js
```diff
@@ -53,6 +53,57 @@ function plan(count) {
...
+
+	this.addPendingAttemptAssertion();
+
+	const running = this
+		.forAttempt(title, t => fn(t, ...args))
+		.run();
+
+	return running.then(ret => {
+		const {passed, error, title, logs} = ret;
+		const errors = error ? Array.isArray(error) ? [...error] : [error] : [];
```
novemberborn: We don't currently return an array of errors do we?

index.d.ts
```diff
@@ -369,6 +370,36 @@ export interface PlanFn {
...
+	commit: (opts?: CommitDiscardOptions) => void,
+	discard: (opts?: CommitDiscardOptions) => void,
+	passed: boolean,
+	errors: AssertionError[],
+	title: string,
+	logs: string[],
+}>
+
+export interface CommitDiscardOptions {
+	retainLogs?: boolean
```
novemberborn: Let's not forget to add inline documentation.

lib/test.js
```diff
@@ -53,6 +53,57 @@ function plan(count) {
...
+			passed,
+			errors,
+			title,
+			logs,
+			commit: ({retainLogs = true} = {}) => {
+				if (committed) {
+					return;
+				}
+				if (discarded) {
+					this.saveFirstError(new Error('Can\'t commit the result that was previously discarded'));
```
novemberborn: "a result" rather than "the result" I think.

lib/test.js
```diff
@@ -53,6 +53,57 @@ function plan(count) {
...
+					return;
+				}
+				committed = true;
+				this.countAttemptAssertion({inc: 1, result: ret, retainLogs});
```
novemberborn: This should be the number of assertions made in the attempt, not hardcoded to `1`.

lib/test.js
```diff
@@ -245,7 +344,11 @@ class Test {
 	verifyAssertions() {
 		if (!this.assertError) {
 			if (this.failWithoutAssertions && !this.calledEnd && this.planCount === null && this.assertCount === 0) {
-				this.saveFirstError(new Error('Test finished without running any assertions'));
+				if (this.pendingAttemptCount > 0) {
+					this.saveFirstError(new Error('Test finished, but not all attempts were committed'));
```
novemberborn: Add `or discarded`.

lib/test.js
qlonik: Currently run() only returns one error, which I think is the first recorded error.
We can modify `saveFirstError(error)` to save array of errors, rather than one error.

lib/test.js
qlonik: This value is not returned from the attempt. It can be obtained from `assertCount` private field of the attempt instance. Is it acceptable?

index.d.ts
qlonik: Sounds good

index.d.ts
```diff
@@ -369,6 +370,57 @@ export interface PlanFn {
...
+	): AttemptResult;
+	<T extends any[]>(
+		impl: (t: ExecutionContext<Context>, ...args: T) => ImplementationResult,
+		...args: T
+	): AttemptResult;
+
+	skip(...values: Array<any>): void;
+}
+
+export class AssertionError extends Error {
```
qlonik: We should pick what can be exposed in AssertionError

qlonik:
> Changing the test titles will definitely have unintended consequences.

But, the code as it is now does not change the title of the running test. It creates new test with a different title. And as long as order of attempts is the same, the attempt test title will correspond. However, if they are reordered, snapshots will fail in those tests, just if they would when reordering snapshots themselves.

I'm going to add the test case with the current behavior

novemberborn:
> But, the code as it is now does not change the title of the running test. It creates new test with a different title. And as long as order of attempts is the same, the attempt test title will correspond.

You're assuming that the number of attempts required is deterministic. That won't apply to all use cases, for example using `t.try()` as a method of retrying flaky tests.

qlonik:
> You're assuming that the number of attempts required is deterministic.

Oh yea. I haven't considered that use case.

qlonik:
Oh, I had an idea yesterday. If we allow to have macro passed as the attempt function to `t.try()`, that macro might have the title generating function. We can have the behavior that if that function is not provided, then tests will be in 'conflicting' mode, where you can only have one snapshot which will be committed to. However, if the title function is provided, and whatever it generates does not conflict with the title of the running test or previous attempt, then it will be in the 'non-conflicting' mode, like it is now. Additionally, if the attempt title is generated but still conflicting with the test title or with previous attempt, it might be in the 'conflicting' mode as well.

Does it even make sense to have this feature?

novemberborn:
Hey @qlonik sorry for the delay in getting back to you. Will review now and make a checklist of whatever work is left.

Re concurrent attempts, I think we can allow them as long as we throw when both attempts save a snapshot.

novemberborn:
@qlonik I've pushed two commits. What do you think about this checklist?

* [ ] Update test `assertCount` with the number of assertions made within a committed attempt
* [ ] Type `t.try()` so it accepts macros
* [ ] Investigate refactor so attempts don't need to instantiate a new `Test`
  * [ ] If not possible, add metadata to indicate the test is "inline" and adjust handling of `test.failing()` and `t.end()` based on that
* [ ] Revisit how attempt titles are generated
* [ ] Ensure snapshots created in each attempt start at the correct invocation count
* [ ] Fail the test if concurrent attempts that created snapshots are committed
* [ ] Determine documented properties of `AssertionError` and add to type definitions
* [ ] Update Flow definition
* [ ] Update ESLint plugin to recognize the assertion
* [ ] File issue to discuss suggested argument name (`tt`?) for inline functions, and update `power-assert` rewrite rules to handle it
* [ ] Review test coverage
* [ ] Update documentation

Let me know which items you'd like me to pick up.

novemberborn:
Sorry, still haven't gotten round to this PR 😢

qlonik:
> Sorry, still haven't gotten round to this PR :cry:

Yeaa, me neither sorry. The only thing is I looked at `assert.js` file very carefully in order to understand which fields it depends on from Test class. That is so I could refactor test.js if needed.

As an attempt, I refactored assert.js and test.js in such a way that assertions dont need to be bound to the test instance. Instead ExecutionContext is extending the Assert class, and Test exposes needed functions for ExecutionContext. Would this refactor be useful to submit? Maybe as a separate pull request.

Also, as a passing question: is there interest in porting ava to TypeScript?

qlonik:
@novemberborn While working on typing the try function, I stumbled upon the solution to #1970. I found that we can get rid of that clever crafty extends and infers and conditional types. It appears that TS didn't like the fact there was no definition for macro without the title. When I added that, it worked without problems. Would you like me to file PR with that refactoring?

novemberborn:
> As an attempt, I refactored assert.js and test.js in such a way that assertions dont need to be bound to the test instance. Instead ExecutionContext is extending the Assert class, and Test exposes needed functions for ExecutionContext. Would this refactor be useful to submit? Maybe as a separate pull request.

Yes would love to see that.



> While working on typing the try function, I stumbled upon the solution to #1970. I found that we can get rid of that clever crafty extends and infers and conditional types. It appears that TS didn't like the fact there was no definition for macro without the title. When I added that, it worked without problems. Would you like me to file PR with that refactoring?

Sure that'd be great. I remember something along those lines but the trouble is that the `title` function is optional.

> Also, as a passing question: is there interest in porting ava to TypeScript?

Yes on the one hand, no on the "that sounds like a lot of work and code churn that maybe we shouldn't deal with right now" hand.

qlonik:
I tried refactoring Test class so that we do not need to instantiate new instance of Test for attempt. I tried pulling all the code that is related to executing the Test into Host class. I ended up taking out majority of classes just related to executing the test function (with all of the edge cases being handled). If I were to also pull things related to assertion counting, I would end up with Host being equal to Test. So I'm not exactly sure if it makes sense to refactor Test so that we do not instantiate new instance of Test. 

I've added inline metadata type, as it seemed to be slightly easier.

qlonik:
In order to address snapshots, could you help me come up with concrete test cases? Maybe it will be easier to figure out how to implement it, if we have concrete usage examples, edge cases and problematic cases.

novemberborn:
Hi @qlonik, sorry I haven't gotten round to picking this up again. Now that your other PRs have landed would you like to update this one?

> In order to address snapshots, could you help me come up with concrete test cases?

```js
for (let i = 0; i < 10; i++) {
  const result = await t.try(t => t.snapshot(doStuffThatSometimesBreaks())
  if (result.passed) {
    result.commit()
    break
  }
  result.discard()
}
```

Here, depending on when stuff didn't break, the test would be creating different snapshots, which then causes assertion failures in the next iteration.

novemberborn:
Hey @qlonik, I finally had some time to work on this! Have a look at the commits and let me know what you think. I haven't run the tests yet so fingers crossed 😄 

I've updated the to-do at the top of this PR. Thanks for your work on this and for all the improvements we've already landed.

novemberborn:
Turns out there's a compilation issue with `TryFn` and TypeScript 3.5.1. The fix seems to be to change the default type of `Context` to `unknown`, but that's probably is a breaking change.

Not sure whether this is due to a regression in TypeScript. It may be that this is the more correct type regardless.

qlonik:
This is really cool! Thank you for your help :) I've looked over the changes and they look very good. I will try fix errors reported by the failing tests, when I have time.

> Turns out there's a compilation issue with TryFn and TypeScript 3.5.1. The fix seems to be to change the default type of Context to unknown, but that's probably is a breaking change.

Maybe it has to do something with the changes related around `{}` and `unknown` type? The closest issue I found is this one: https://github.com/Microsoft/TypeScript/pull/30637. Maybe it is related to usage of TS 3.5 rather than switch from `{}` to `unknown`.

novemberborn:
> Maybe it has to do something with the changes related around `{}` and `unknown` type? The closest issue I found is this one: [microsoft/TypeScript#30637](https://github.com/microsoft/TypeScript/pull/30637). Maybe it is related to usage of TS 3.5 rather than switch from `{}` to `unknown`.

Yes, but what's surprising is the addition of the new assertion broke it, and the only fix was to switch to `unknown`.

novemberborn:
> This is really cool! Thank you for your help :) I've looked over the changes and they look very good. I will try fix errors reported by the failing tests, when I have time.

Thank you! I've managed to fix regressions that broke unrelated tests.

qlonik:
I had a bit of time and fixed some failing test cases related to try assertion

qlonik:
I had couple questions after working on tests:

1. Doing snapshots in concurrent attempts raises error. It seems that if the flag to update snapshots is set, then the incomplete snapshot is saved (as seen in 5e874680a7d133c4de8ddd787f0ba738f62ab29e). Should that be the case? Maybe the snapshot for the entire test case should not be saved? I'm not sure of the correct behavior in this scenario.

2. I was thinking that we should clarify the error message about snapshots in concurrent attempts (defined on https://github.com/avajs/ava/pull/1947/commits/30e522a598c3a490027649da7d35948811a679b3#diff-dd2e881af3311f834199c7b515cff0c7R369). From the code it appears that we can have parallel attempts with snapshots in them as long as only 0 or 1 attempts is committed and the rest are discarded. Should we convey the possibility of this? We could say something along the lines of `"Cannot commit to multiple concurrent attempts containing snapshots. Only can commit to one or zero concurrent attempts containing snapshot."`

test/try-snapshot.js
```diff
@@ -0,0 +1,94 @@
...
+			};
+
+			const attempt2 = t2 => {
+				t2.snapshot({foo: 'bar'});
+			};
+
+			return a.try(attempt1).then(first => {
+				first.commit();
+				return a.try(attempt2);
+			}).then(second => {
```
novemberborn: Could you change these tests to use `await`?

test/try-snapshot.js
```diff
@@ -0,0 +1,94 @@
...
+			const attempt1 = t2 => {
+				t2.snapshot(true);
+				t2.snapshot({boo: 'far'});
+			};
+
+			const attempt2 = t2 => {
+				t2.snapshot({foo: 'bar'});
+			};
+
+			return a.try(attempt1).then(first => {
```
novemberborn: It may be clearer to use `const attempt1 = try(t2 => …)` rather than setting up the functions. It took me a while to realize they weren't executing yet.

test/test-try-commit.js
```diff
@@ -0,0 +1,523 @@
+'use strict';
+require('../lib/chalk').set();
+require('../lib/worker/options').set({color: false});
+
+const {test} = require('tap');
+const delay = require('delay');
+const ContextRef = require('../lib/context-ref');
+const {ava} = require('./helper/ava-test');
+
+test('try-commit are present', t => {
```
novemberborn: This isn't necessary.

test/test-try-commit.js
```diff
@@ -0,0 +1,523 @@
...
+		a.pass();
+		t.type(a.try, Function);
+	}).run();
+});
+
+test('try-commit works', t => {
+	const instance = ava(a => {
+		return a
+			.try(b => b.pass())
+			.then(res => {
```
novemberborn: Would you mind updating these tests to use `await`?

test/test-try-commit.js
```diff
@@ -0,0 +1,523 @@
...
+		}).then(res => {
+			t.true(res.passed);
+			res.commit();
+		});
+	}).run().then(result => {
+		t.true(result.passed);
+	});
+});
+
+test('try-commit fails when plan is not reached inside the try', t => {
```
novemberborn: We need a test which checks that any assertions outside of the attempt do not count towards the plan, and similarly that any assertions inside the attempt do not count towards the plan in the main test.

test/test-try-commit.js
```diff
@@ -0,0 +1,523 @@
...
+		}).then(res => {
+			t.false(res.passed);
+			res.commit();
+		});
+	}).run().then(result => {
+		t.false(result.passed);
+	});
+});
+
+test('try-commit passes with failing test', t => {
```
novemberborn: What do you mean by this? A failure inside the attempt should cause the main test to fail if that attempt is committed.

test/test-try-commit.js
```diff
@@ -0,0 +1,523 @@
...
+			})
+			.then(() => {
+				a.end();
+			});
+	}).run().then(result => {
+		t.true(result.passed);
+	});
+});
+
+test('try-commit does not allow to use .end() in attempt when parent is callback test', t => {
```
novemberborn: We should also test what happens if you use `.end()` when the parent is a regular test.

test/test-try-commit.js
```diff
@@ -0,0 +1,523 @@
...
+			b.pass();
+			return delay(200);
+		}).then(result => result.commit());
+	}).run().then(result => {
+		t.is(result.passed, false);
+		t.match(result.error.message, /timeout/);
+	});
+});
+
+test('try-commit refreshes the timeout on commit/discard', t => {
```
novemberborn: Should we test that it does not reset when assertions are run within the attempt?

test/test-try-commit.js
```diff
@@ -0,0 +1,523 @@
...
+				results[0].commit();
+				results[1].discard();
+			})
+		]);
+	}).run().then(result => {
+		t.true(result.passed);
+	});
+});
+
+test('try-commit abides timeout', t => {
```
novemberborn: We should add a test for the attempt configuring its own timeout.

test/test-try-commit.js
```diff
@@ -0,0 +1,523 @@
...
+		setTimeout(() => a.try(b => b.pass()).then(result => result.commit()), 5);
+		setTimeout(() => a.try(b => b.pass()).then(result => result.commit()), 10);
+		setTimeout(() => a.try(b => b.pass()).then(result => result.commit()), 15);
+		setTimeout(() => a.end(), 20);
+	}).run().then(result => {
+		t.is(result.passed, true);
+	});
+});
+
+test('try-commit can access parent test context', t => {
```
novemberborn: Could you rename to `try-commit inherits the test context`?

test/test-try-commit.js
```diff
@@ -0,0 +1,523 @@
...
+		return a.try(b => {
+			b.pass();
+			t.strictDeepEqual(b.context, data);
+		}).then(res => res.commit());
+	}, context).run().then(result => {
+		t.is(result.passed, true);
+	});
+});
+
+test('try-commit cannot set parent test context', t => {
```
novemberborn: Could you rename to `assigning context in try-commit does not affect parent`?

novemberborn:
I've extracted the `unknown` context change in https://github.com/avajs/ava/pull/2186.

test/test-try-commit.js
qlonik: The title is not clear at all. What I meant is that the test that is marked as failing (expected to fail) will pass, even if try-commit within the test will fail.

test/test-try-commit.js
qlonik: Added these tests, found that assertions inside the attempt in fact do count towards the plan. Is it an error which should be fixed?

test/test-try-commit.js
qlonik: I've come up with the test, but it's hard to reason if the test is correct.

qlonik:
About concurrent snapshots. The file with snapshots contains snapshots for assertion outside of attempts and assertions from within the first attempt, but not second. Maybe none of attempts should be recorded?

```markdown
## concurrent

> Snapshot 1

    'hello'

> Snapshot 2

    true

> Snapshot 3

    {
      boo: 'far',
    }
```

Thinking about names for assertions in attempt made me realize that we do not handle the case when inside the attempt we are calling assertions from parent scope. So, maybe we should figure out what to do in that case, and if cannot how to avoid that, we could recommend to shadow the variable with parent assertions. Other than this issue, those names seem good.

About properties of AssertionError, we could go safer way and not expose anything unless there are requests to have properties exposed. Another thought is that we could expose properties which will help with debugging. For example, ones like `assertion`, `raw`, and `improperUsage`, which show which assertion threw an error, which values it received (in case that computed values are passed as parameters), and whether the usage of given assertion clause was correct or not.

With respect to attempt titles, I think the titles are okay, when we are not dealing with snapshots, as they are similar to temporary ids within the test. However, those might cause issues when snapshots are involved, as you mentioned.  I actually forgot already: do attempt titles matter for snapshots at all? It seems that snapshot file contains numbered snapshots for the entire test case and they do not have any references to the title of attempt.

Also I think that naming one attempt with snapshots inside might break all following attempts with snapshots, since indexes might shift. It makes sense to force the title when snapshots inside the attempt are involved. But on the other hand, regular tests without attempts do not force the snapshot to have an id, since they are always ordered and if snapshot order changes then the failure of the test suite is expected. So, similar to that, snapshots inside attempts could be expected to be ordered and change in the attempt order would fail the test. 

We should document somewhere, that if we want to change how the default title for attempt is generated, then it will be a breaking change since it will break existing snapshots (that is in the case that snapshots actually depend on the attempt title).

novemberborn:
> About concurrent snapshots. The file with snapshots contains snapshots for assertion outside of attempts and assertions from within the first attempt, but not second. Maybe none of attempts should be recorded?

Snapshots from all committed attempts should be recorded. Are you describing a scenario in which this is not the case?

> Thinking about names for assertions in attempt made me realize that we do not handle the case when inside the attempt we are calling assertions from parent scope. So, maybe we should figure out what to do in that case, and if cannot how to avoid that, we could recommend to shadow the variable with parent assertions. Other than this issue, those names seem good.

Hmm. I suppose we could fail assertions on the parent context when attempts are pending, other than `t.try()` of course. Would be good to avoid that kind of confusion.

> About properties of AssertionError, we could go safer way and not expose anything unless there are requests to have properties exposed. Another thought is that we could expose properties which will help with debugging. For example, ones like `assertion`, `raw`, and `improperUsage`, which show which assertion threw an error, which values it received (in case that computed values are passed as parameters), and whether the usage of given assertion clause was correct or not.

I like this, but now I'm tempted to bike-shed these properties. Let's go with an empty interface for now and file a follow-up issue.

> With respect to attempt titles, I think the titles are okay, when we are not dealing with snapshots, as they are similar to temporary ids within the test. However, those might cause issues when snapshots are involved, as you mentioned. I actually forgot already: do attempt titles matter for snapshots at all? It seems that snapshot file contains numbered snapshots for the entire test case and they do not have any references to the title of attempt.

No I think you're right, we include the snapshots in the parent test not the attempt. So this is fine for now, Let's make sure we document that we may change the algorithm for generated titles.

novemberborn:
> > About concurrent snapshots. The file with snapshots contains snapshots for assertion outside of attempts and assertions from within the first attempt, but not second. Maybe none of attempts should be recorded?
> 
> Snapshots from all committed attempts should be recorded. Are you describing a scenario in which this is not the case?

@qlonik do you mean that the snapshots from the first attempt are recorded, and then the test fails? I don't think that's any different from regular snapshots which are recorded before an assertion fails.

qlonik:
Yes, I meant that. I think it's okay then.

novemberborn:
@qlonik I've hopefully fixed the tests and made it so you can't run assertions outside of an active attempt. `t.try()` now counts as a single assertion as far as `t.plan()` is concerned. Do these changes look good to you?

I'm planning to make `t.try()` opt-in for now (see https://github.com/avajs/ava/pull/2218) so we can land this at least, and then do follow-up work in `master`. Thank you so much for your work on this!

qlonik:
The changes look good to me. The only thing I have to add and I'm not sure if it exists is the `@experimental` tag for jsdoc. If it does, we could add that tag into doc for those functions (but the word experimental in the beginning should still be kept, in case the ide doesn't support experimental tag)

novemberborn:
> The only thing I have to add and I'm not sure if it exists is the `@experimental` tag for jsdoc.

Oh I don't really know about JSDoc. But if it comes up we can update the definition for that. I'll prefix the comments with "Requires opt-in." though.

novemberborn:
@qlonik I'll file the follow-up issues later, but for now 🎉 let's merge this! Thank you so much for your hard work on this, and for bearing with me over all these months.

qlonik:
Yay 🎉! Thank you for helping with the feature too.

novemberborn:
See https://github.com/orgs/avajs/projects/1 for the follow-up work. Let me know if I missed anything 😄

qlonik:
I think the case when someone tries to use parent's assertions while inside the attempt is not addressed. 

This is the only discussion we had about it so far:


>> Thinking about names for assertions in attempt made me realize that we do not handle the case when inside the attempt we are calling assertions from parent scope. So, maybe we should figure out what to do in that case, and if cannot how to avoid that, we could recommend to shadow the variable with parent assertions. Other than this issue, those names seem good.
>
> Hmm. I suppose we could fail assertions on the parent context when attempts are pending, other than t.try() of course. Would be good to avoid that kind of confusion.

novemberborn:
No I dealt with that: https://github.com/avajs/ava/pull/1947/commits/bb458be5353dc42bd48b53786e6302b58e499399

qlonik:
Ah, okay, I didn't notice. :+1:

PR #1948, opened 2018-10-08
Show correct coverage badge in the readme

lo1tuma:
I had a look at the coverage report on coveralls that is linked in the readme and I was very confused that the last build is from 2016.
It seems like that coveralls has been replaced with codecov in 2ed94853ddd2ca4614a82b2a79cc251c09212331, but the badge has never been updated.


novemberborn:
Ha! I always overlook badges. Oops!

PR #1949, opened 2018-10-09
Configuration file validation

### LINKED ISSUE #1943: Ensure ava.config.js export does not contain an "ava" property
`ava.config.js` files should include the AVA configuration, without wrapping it inside an AVA field. In other words, this is wrong:

```js
export default {
  ava: {
  }
}
```

To prevent confusion, we should check if the configuration has an `ava` key and throw an error.

I realize the documentation is confusing, since we're not consistent in whether we're showing a partial `package.json`, but this will at least help nudge in the right direction.

The config is loaded in this file: https://github.com/avajs/ava/blob/master/lib/load-config.js

wmik:
- Adds title to [configuration snippet](https://github.com/avajs/ava#configuration) to mark it as part of `package.json` and
- fixes #1943


havenchyk:
@novemberborn please, take a look

wmik:
Thanks for the review @novemberborn I suppose `.hasOwnProperty` was overkill 😅 Otherwise the updates look great. Nothing more to add on my side.

PR #1954, opened 2018-10-13
Restore Babel cache invalidation

### LINKED ISSUE #1817: Restore Babel cache invalidation
#1798 caused drastic regressions in our ability to invalidate the precompilation cache. This was intentional, as we wanted to upgrade to newer Babel 7 betas.

There's a few `TODO` comments in `lib/babel-pipeline.js` with pointers on how to improve our cache invalidation:

https://github.com/avajs/ava/blob/a5de3691e76a2a999ac3392830d1e26d279d60ab/lib/babel-pipeline.js#L134

https://github.com/avajs/ava/blob/a5de3691e76a2a999ac3392830d1e26d279d60ab/lib/babel-pipeline.js#L111

With regards to plugins &  presets, we should figure out what package the resolved plugin / preset is from, and then use [`package-hash`](https://github.com/novemberborn/package-hash/) to compute a hash. If the plugin / preset is a file in the current project (and not a dependency) we should just hash the file. The same if it's from outside of the current project.

We're now resolving Babel options for every file that's being compiled. We should therefore have a lookup of resolved plugins / presets and config files to previously computed hashes.

`package-hash` has an asynchronous mode which we could use, though we'd have to change the pipeline to be asynchronous itself, so that'd be a bit more work.

`userOptions` may come from an `ava.config.js` file (at least when #1761 lands). Consequently they may include non-serializable values. Perhaps we should serialize it using Concordance before taking the hash.

novemberborn:
Fixes #1817.

### DIFF
--- a/lib/babel-pipeline.js
+++ b/lib/babel-pipeline.js
@@ -136,7 +136,7 @@ function build(projectDir, cacheDir, userOptions, compileEnhancements) {
 	const containsStage4 = makeValueChecker('../stage-4');
 	const containsTransformTestFiles = makeValueChecker('@ava/babel-preset-transform-test-files');
 
-	const loadOptions = (filename, inputSourceMap) => {
+	const loadOptions = filename => {
 		const partialTestConfig = babel.loadPartialConfig(Object.assign({
 			babelrc: false,
 			babelrcRoots: [projectDir],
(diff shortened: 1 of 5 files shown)


lib/babel-pipeline.js
```diff
@@ -99,6 +102,41 @@ function wantsStage4(userOptions, projectDir) {
...
+		inputs.push(babelrc);
+
+		const filename = path.basename(babelrc);
+		if (filename === 'package.json') {
+			inputs.push(JSON.stringify(pkgConf.sync('babel', {cwd: path.dirname(filename)})));
+		} else {
+			inputs.push(stripBomBuf(fs.readFileSync(babelrc)));
+		}
+	}
+	// TODO: If present, hash config contents
```
sindresorhus: If *what* is present?

Issue #1893, opened 2018-08-11
Better typing for throws assertions


novemberborn:
This is the last piece of work that's come out of the discussion in #1794.

The `t.throws()` and `t.throwsAsync()` assertions are currently typed to return `any` and `Promise<any>` respectively. This is because they only return an error if one was actually thrown. If the assertion fails they return `undefined` and execution continues.

Whilst theoretically correct, we don't feel it's the pragmatic choice. Instead we should type these assertions to return `Error` / `Promise<Error>`. Additionally it'd be great if the assertions could be typed as generics so you can provide your own `Error` subclass (`const err = t.throws<MyError>(fn)` or `const err: MyError = t.throws(fn)`). Note that the assertion requires that an `Error` instance is thrown, so the generic shouldn't take any value.

This should come with some example code in `test/flow-types` and `test/ts-types`, and be documented in our Flow and TypeScript recipes.

We can land implementations for Flow and TypeScript separately, though keep in mind that we like to keep diffs between the `index.d.ts` and `index.js.flow` files to a minimum.

okyantoro:
Good idea @novemberborn. It will give more information about returned value by `t.throws()` and `t.throwsAsync()` assertions. Exciting issue to work on.

PR #1956, opened 2018-10-13
Better TypeScript typing for throws assertions

novemberborn:
See #1893.

I might do Flow tomorrow if I have the time.

### DIFF
--- a/docs/recipes/typescript.md
+++ b/docs/recipes/typescript.md
@@ -135,3 +135,32 @@ test('foo is bar', macro, 'bar');
 ```
 
 Note that, despite the type cast above, when executing `t.context` is an empty object unless it's assigned.
+
+## Typing `throws` assertions
+
+The `t.throws()` and `t.throwsAsync()` assertions are typed to always return an Error. You can customize the error class using generics:
+
+```ts
+import test from 'ava'
+
+class CustomError extends Error {
+	parent: Error
+
+	constructor (parent) {
+		super(parent.message);
+		this.parent = parent;
+	}
+}
+
+test('throws', t => {
+	const err = t.throws<CustomError>(myFunc);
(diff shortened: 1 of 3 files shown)


test/ts-types/throws.ts
```diff
@@ -0,0 +1,26 @@
...
+
+	constructor() {
+		super();
+		this.foo = 'foo';
+	}
+}
+
+test('throws', t => {
+	const err1: Error = t.throws(() => {});
+	// t.is(err1.foo, 'foo');
```
sindresorhus: ?

docs/recipes/typescript.md
```diff
@@ -135,3 +135,32 @@ test('foo is bar', macro, 'bar');
...
+
+	constructor(parent) {
+		super(parent.message);
+		this.parent = parent;
+	}
+}
+
+test('throws', t => {
+	const err = t.throws<CustomError>(myFunc);
+	t.is(err.parent.name, 'TypeError');
```
sindresorhus: Maybe also include the contents of `myFunc` somewhere as it's not that clear where `TypeError` comes from.

test/ts-types/throws.ts
novemberborn: It's a type exception. I think maybe there's a way to indicate that to `tsc`?

docs/recipes/typescript.md
novemberborn: Done.

novemberborn:
I've added the Flow typing too.

test/ts-types/throws.ts
sindresorhus: You can use the `// @ts-ignore` comment to ignore the below line.

PR #1961, opened 2018-10-21
Pass an undefined title argument to macro-title functions

novemberborn:
Previously, AVA would pass the empty string. By passing an undefined value, the title function can use argument defaults which is more natural.

---

This is a breaking change for users who were concatenating `title`, under the assumption it was an empty string.

### DIFF
--- a/docs/recipes/typescript.md
+++ b/docs/recipes/typescript.md
@@ -66,7 +66,7 @@ import test, {Macro} from 'ava';
 const macro: Macro = (t, input: string, expected: number) => {
 	t.is(eval(input), expected);
 };
-macro.title = (providedTitle: string, input: string, expected: number) => `${providedTitle} ${input} = ${expected}`.trim();
+macro.title = (providedTitle = '', input: string, expected: number) => `${providedTitle} ${input} = ${expected}`.trim();
 
 test(macro, '2 + 2', 4);
 test(macro, '2 * 3', 6);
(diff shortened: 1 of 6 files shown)


novemberborn:
I need to update the type definitions to make it clear that the first argument to the `title` function may now be `undefined`.

novemberborn:
I need to update the type definitions to make it clear that the first argument to the `title` function may now be `undefined`.

novemberborn:
I need to update the type definitions to make it clear that the first argument to the `title` function may now be `undefined`.

novemberborn:
I need to update the type definitions to make it clear that the first argument to the `title` function may now be `undefined`.

novemberborn:
I need to update the type definitions to make it clear that the first argument to the `title` function may now be `undefined`.

novemberborn:
I need to update the type definitions to make it clear that the first argument to the `title` function may now be `undefined`.

novemberborn:
I need to update the type definitions to make it clear that the first argument to the `title` function may now be `undefined`.

novemberborn:
I need to update the type definitions to make it clear that the first argument to the `title` function may now be `undefined`.

novemberborn:
I need to update the type definitions to make it clear that the first argument to the `title` function may now be `undefined`.

novemberborn:
I need to update the type definitions to make it clear that the first argument to the `title` function may now be `undefined`.

novemberborn:
I need to update the type definitions to make it clear that the first argument to the `title` function may now be `undefined`.

novemberborn:
Heh, turns out GitHub did post some of my attempts to comment. And when I deleted the copies, it aggressively deleted the last one too… so one last time:

I need to update the type definitions to make it clear that the first argument to the `title` function may now be `undefined`.

PR #1966, opened 2018-10-23
Link to French translation of puppeteer recipe

forresst:


sindresorhus:
Thanks for your awesome work on keeping things translated, @forresst 🙌

PR #1970, opened 2018-10-28
Type additional arguments used by macros

novemberborn:
This is an attempt at inferring the types of the additional arguments when you use macros.

```typescript
const hasLength: Macro<[string, number]> = (t, expected, length) => {
	t.is(expected.length, length);
};

test('bar has length 3', hasLength, 'bar', 3);

const hasCodePoints: Macro<[string, number]> = (t, expected, length) => {
	t.is(Array.from(expected).length, length);
};
test('bar has length 3', [hasLength, hasCodePoints], 'bar', 3);
```

See `test/ts-types/macro.ts` for more examples.

This all works quite well as long as you provide the correct arguments. However, the feedback when you don't is atrocious. Due to the way we overload the `test()` function TypeScript assumes you're trying to use `test(macro, ...args)` if it can't match `test('title', macro, ...args)`, which means it's complaining that the `'title'` string isn't a macro….

I feel like there should be a way for TypeScript to narrow the overloads based on whether a title is passed. We may need to open an issue for that (though I don't know the correct terms) but even then it'll take a while to ship any improvements.

One solution would be to make the `title` parameter required. We could still accept an empty string, of course. Alternatively we could require a typecast before `test()` can be used without a `title`.

I still need to document the new generics. I should write some negative typechecks as well.

I'm not sure to what extend this is possible in Flow.

@sindresorhus, @SamVerschueren thoughts?

### DIFF
--- a/docs/recipes/typescript.md
+++ b/docs/recipes/typescript.md
@@ -58,15 +58,27 @@ test(async t => {
 
 ## Using [macros](../01-writing-tests.md#reusing-test-logic-through-macros)
 
+Macros can receive additional arguments. AVA can infer these to ensure you're using the macro correctly:
+
+```ts
+import test, {ExecutionContext} from 'ava';
+
+const hasLength = (t: ExecutionContext, input: string, expected: number) => {
+	t.is(input.length, expected);
+};
+
+test('bar has length 3', hasLength, 'bar', 3);
+```
+
 In order to be able to assign the `title` property to a macro you need to type the function:
 
 ```ts
 import test, {Macro} from 'ava';
 
-const macro: Macro = (t, input: string, expected: number) => {
+const macro: Macro<[string, number]> = (t, input, expected) => {
 	t.is(eval(input), expected);
 };
-macro.title = (providedTitle = '', input: string, expected: number) => `${providedTitle} ${input} = ${expected}`.trim();
+macro.title = (providedTitle = '', input, expected) => `${providedTitle} ${input} = ${expected}`.trim();
 
 test(macro, '2 + 2', 4);
 test(macro, '2 * 3', 6);
@@ -78,7 +90,7 @@ You'll need a different type if you're expecting your macro to be used with a ca
 ```ts
 import test, {CbMacro} from 'ava';
 
-const macro: CbMacro = t => {
+const macro: CbMacro<[]> = t => {
 	t.pass();
 	setTimeout(t.end, 100);
 };
@@ -123,7 +135,7 @@ interface Context {
 
 const test = anyTest as TestInterface<Context>;
 
-const macro: Macro<Context> = (t, expected: string) => {
+const macro: Macro<[string], Context> = (t, expected: string) => {
 	t.is(t.context.foo, expected);
 };
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -394,7 +394,7 @@ export interface Macro<Args extends any[], Context = {}> {
 }
 
 /** Alias for a single macro, or an array of macros. */
-export type OneOrMoreMacros<Context> = Macro<any, Context> | [Macro<any, Context>, ...Macro<any, Context>[]];
+export type OneOrMoreMacros<Args extends any[], Context> = Macro<Args, Context> | [Macro<Args, Context>, ...Macro<Args, Context>[]];
 
 /** A reusable test or hook implementation, for tests & hooks declared with the `.cb` modifier. */
 export interface CbMacro<Args extends any[], Context = {}> {
@@ -403,7 +403,7 @@ export interface CbMacro<Args extends any[], Context = {}> {
 }
 
 /** Alias for a single macro, or an array of macros, used for tests & hooks declared with the `.cb` modifier. */
-export type OneOrMoreCbMacros<Context> = CbMacro<any, Context> | [CbMacro<any, Context>, ...CbMacro<any, Context>[]];
+export type OneOrMoreCbMacros<Args extends any[], Context> = CbMacro<Args, Context> | [CbMacro<Args, Context>, ...CbMacro<Args, Context>[]];
 
 /** Infers the types of the additional arguments the macro implementations should be called with. */
 export type InferArgs<OneOrMore> =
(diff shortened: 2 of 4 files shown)


qlonik:
Hey.

I stumbled across this pr and I thought it was interesting.
I tried simply moving definition which only has macro and args (without title) to be above the one with title. However, in that case, the outlined problem shifts to other function call. So that does not work well as solution.

I gave it a shot though. It is possible to merge those two definitions. I ended up having the following. Also, props to `tycho01/typical` for bunch of awesome TS types. `Tail` here is his definition: https://github.com/tycho01/typical/blob/65b015307eb4fb35f3771494be09a9ee00dd285d/src/array/Tail.ts

```typescript
export type Tail<T extends any[]> =
	((...args: T) => any) extends ((head: any, ...tail: infer R) => any) ? R : never;

export interface TestInterface<Context = {}> {
	/** Declare a concurrent test. */
	(title: string, implementation: Implementation<Context>): void;

	/** Declare a concurrent test that uses one or more macros. Additional arguments are passed to the macro. */
	<
		OneOrMore extends OneOrMoreMacros<Context>,
		X extends string | OneOrMore,
		Y extends (
			X extends string ? OneOrMore :
			X extends OneOrMore ? InferArgs<X>[0] :
			never
		)
	>(
		titleOrMacro: X,
		macroOrArg: Y,
		...rest:
			Y extends OneOrMore ? InferArgs<Y> :
			X extends OneOrMore ? Tail<InferArgs<X>> :
			never
	): void;

	// replace both definitions of macro based fn with the definition above.
	// after, afterEach and so on
}
```

While it is possible, this solution has problems.
1. It works the best when title is passed. It can detect if not proper amount of parameters is passed. It gets confused however when zero parameters is passed. It says that Macro is not assignable to Implementation. However, if you pass at least one additional argument, then it will say 'expected _x_, but got _y_<_x_'
2. When title is not passed, it can detect improper amount of parameters passed, but then you need to pass at least 2 parameters. If you pass zero parameters, it says 'needs at least 2 params', if only one parameter is passed, it says 'macro is not assignable to string'
3. There is weirdness when macro has zero parameters to it. I think macro has to have at least one parameter? if it has zero parameters, it is actually just a test implementation. Maybe test implementation could be generalized somehow?

One more important thing I should add. I was exploring at some point these rather complicated types that could be done with TS. TS has a lot of power, and a lot can be done. You can actually put TS compiler into infinite loop, or even make it exceed the stack. It is possible to be careful with types and not let TS explode like this. However, all of this power comes at a usability price. TS will take time to determine what are the parameters to the function supposed to be, so that while user is typing, the text editor might not keep up and show the appropriate type information while typing. It seems that the implementation is only using `extends` and `?:` and `infer`, which should be handled well by TS compiler, and there are no loops/recursions, which is good. But the complexity of types is something that should be kept in mind.

P.S. I had some fun doing this haha :smile:

novemberborn:
@qlonik wow that's brilliant! I managed to solve the zero parameter case with another overload. We should support it.

Have a look at my commit, what do you think?

qlonik:
It looks good. I have 2 comments:

1. In the case of overloads, TS requires to have most precise/specific type on the top followed by less specific types. I'm not sure if the function with 3 arguments is more specific than the one with 2 arguments or vice versa. Because of this, functions might need to be reordered in the `TestInterface` (and others)

2. Since in TS types are matched based on structure, it seems like it does not make distinction between `Implementation<Context>` and `Macro<[], Context>`. This is evident from the case of `test.after()`. 
If in the `index.d.ts` you reorder such that macro definition above implementation definition (without titles), then call `test.after()` with either test implementation or macro with no parameters, and if you ctrl-click in VS Code on the `.after` part, then both of them will link to the same type declaration, regardless of whether its macro or implementation. It seems that because of this TS does not make a distinction of empty macro and test implementation. I think it might be good to merge those two somehow, so that there are no unneeded type declarations in the TestInterface.
Similarly, the declaration `(title: string, fn: Implementation<Context>): void` should be equivalent to `(titleOrMacro: ToM, macroOrArgs: MoA, ...rest: Stuff): void`, when macro is empty, and TS might not be making any distinction between two types. 
The same test works - have the type definition with implementation above the big test definition, and pass empty macro to test with title, (e.g. `test('title, macro)`) and then ctrl-click `test`, the VS Code will link to test declaration with title and implementation and not to the big type declaration.

qlonik:
I was trying to see if Macro can be merged with Implementation somehow.

I tried declaring Macro as following (notice default declaration to Args as empty tuple). It becomes equivalent to Implementation and could be used interchangeably.
```ts
interface Macro<Args extends any[] = [], Context = {}> {
	(t: ExecutionContext<Context>, ...args: Args): ImplementationResult;

	/**
	 * Implement this function to generate a test (or hook) title whenever this macro is used. `providedTitle` contains
	 * the title provided when the test or hook was declared. Also receives the remaining test arguments.
	 */
	title?: (providedTitle: string | undefined, ...args: Args) => string;
}

{
  const testImpl: Macro = t => t.pass()
  test('test title', testImpl)
}
```

However, I noticed that it is slightly awkward, that this case:
```ts
test('title', t => {
  t.pass()
})
```
Results in `t` being of type `any`, which is not acceptable.


I tried few more things, but it ends up being awkward that implementation passed to test need to have `t` typed as ExecutionContext. It is very hard to merge those two together, it might be easier to make sure that Macro has at least one value passed to it, and Implementation always has zero values passed to it.

I discovered another problematic case:
```ts
	test('string', (t, a: string, b: number) => t.pass(), '', 0);
	test((t, a: string, b: number) => t.pass(), '', 0);
```
I'm not sure if someone might use it like this.

novemberborn:
> 1. In the case of overloads, TS requires to have most precise/specific type on the top followed by less specific types. I'm not sure if the function with 3 arguments is more specific than the one with 2 arguments or vice versa. Because of this, functions might need to be reordered in the `TestInterface` (and others)

@qlonik I don't think that's the case with interfaces, rather it only matters when defining multiple overloads followed by the implementation.

Regarding the `Implementation` type, I think that's more stylistic than anything. We explain macros as being "a bit more" than a regular test implementation, even though there are few practical differences. It's definitely valid to write a macro that takes no additional arguments and does not implement the `title` function, just as it's valid to provide an inline "implementation" that takes the additional arguments (which makes it a macro, of course).

I'd prefer to keep the type definitions as they are so that the (perhaps more common) use case of inline implementations is more easily discerned.

qlonik:
About the order - okay, I think that it might matter for the case of interfaces. But sure, this is good. If it will become a problem, we will see then.

For `Implementation`, I was thinking about duplicate definitions (one with implementation and one with empty macro). I was worried that they are unnecessary and might come in the way. Again, if they will become a problem, we will see then.


What about the case of untyped macros functions being passed to the test? As shown in my  last message. Those do not pass type check. So what should be the recommendation, for people using test function like that?

> ```ts
> test('string', (t, a: string, b: number) => t.pass(), '', 0);
> test((t, a: string, b: number) => t.pass(), '', 0);
> ```

novemberborn:
> What about the case of untyped macros functions being passed to the test? As shown in my last message. Those do not pass type check. So what should be the recommendation, for people using test function like that?

They seem to work fine, as long as you type the `t` argument. I think that's OK. Will add it to the documentation.

novemberborn:
> Will add it to the documentation.

Or actually, this is such an edge-case, perhaps not.

novemberborn:
@sindresorhus any thoughts?

PR #1974, opened 2018-11-04
Reorganize documentation

novemberborn:
Extract various sections from the readme into their own files. I've tried to keep other changes to a minimum.


sindresorhus:
I think we should list the docs files directly in the readme instead of just linking to the file browser. That one less click and makes it easier to Ctrl+F.

PR #1975, opened 2018-11-04
CI updates

novemberborn:
Use Travis CI to test on Windows. Currently, upgrading the `npm` version doesn't work, so I'm skipping tests in Node.js 6.

I've upped the parallel test execution in TAP to two. We may have to adjust this on a per-directory basis if we get intermittent test failures.

I've also added Node.js 11 to the test matrix.


.gitattributes
```diff
@@ -1,4 +1,6 @@
 * text=auto
 *.js text eol=lf
+*.log text eol=lf
+*.md.expected text eol=lf
```
sindresorhus: Btw, Git now supports setting `lf` for *all* text files (since 2016, but I didn't realize until recently):

```
* text=auto eol=lf
```

novemberborn: Would that interfere with people's editor setup?

novemberborn:
Bizarrely, the Push build passed, and the PR build failed due to the reporter logs including crlf-style linebreaks.

novemberborn:
> Bizarrely, the Push build passed, and the PR build failed due to the reporter logs including crlf-style linebreaks.

Going to merge and hope for the best.

PR #1983, opened 2018-11-18
Update the tagline and readme intro

novemberborn:
Something I've been mulling over for a while. @sindresorhus what do you think?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
 	"name": "ava",
 	"version": "1.0.0-rc.2",
-	"description": "Futuristic test runner 🚀",
+	"description": "Testing can be a slog. AVA's here to help you get it done.",
 	"license": "MIT",
 	"repository": "avajs/ava",
 	"homepage": "https://ava.li",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
 	"name": "ava",
 	"version": "1.0.0-rc.2",
-	"description": "Testing can be a slog. AVA's here to help you get it done.",
+	"description": "Testing can be a drag. AVA helps you get it done.",
 	"license": "MIT",
 	"repository": "avajs/ava",
 	"homepage": "https://ava.li",
--- a/readme.md
+++ b/readme.md
@@ -3,7 +3,7 @@
 [![Build Status](https://travis-ci.org/avajs/ava.svg?branch=master)](https://travis-ci.org/avajs/ava)  [![Coverage Status](https://codecov.io/gh/avajs/ava/branch/master/graph/badge.svg)](https://codecov.io/gh/avajs/ava/branch/master) [![XO code style](https://img.shields.io/badge/code_style-XO-5ed9c7.svg)](https://github.com/xojs/xo) [![Join the community on Spectrum](https://withspectrum.github.io/badge/badge.svg)](https://spectrum.chat/ava)
 [![Mentioned in Awesome Node.js](https://awesome.re/mentioned-badge.svg)](https://github.com/sindresorhus/awesome-nodejs)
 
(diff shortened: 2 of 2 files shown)


readme.md
```diff
@@ -1,20 +1,19 @@
...
-
-Even though JavaScript is single-threaded, IO in Node.js can happen in parallel due to its async nature. AVA takes advantage of this and runs your tests concurrently, which is especially beneficial for IO heavy tests. In addition, test files are run in parallel as separate processes, giving you even better performance and an isolated environment for each test file. [Switching](https://github.com/sindresorhus/pageres/commit/663be15acb3dd2eb0f71b1956ef28c2cd3fdeed0) from Mocha to AVA in Pageres brought the test time down from 31 to 11 seconds. Having tests run concurrently forces you to write atomic tests, meaning tests don't depend on global state or the state of other tests, which is a great thing!
-
-![](media/mini-reporter.gif)
+[![Mentioned in Awesome Node.js](https://awesome.re/mentioned-badge.svg)](https://github.com/sindresorhus/awesome-nodejs)
 
-*Read our [contributing guide](contributing.md) if you're looking to contribute (issues/PRs/etc).*
+Testing can be a slog. AVA's here to help you get it done. Its small API, detailed error output, embrace of new language features and process isolation let you write tests more effectively. So you can ship more awesome code 🚀
```
sindresorhus: Nitpick, but I think the sentence should still end in a `.` even though there's an emoji.
```suggestion
Testing can be a slog. AVA's here to help you get it done. Its small API, detailed error output, embrace of new language features and process isolation let you write tests more effectively. So you can ship more awesome code. 🚀
```

sindresorhus: I think we should also mention that it's a JavaScript test runner. Imagine someone stumbling upon this project knowing nothing about AVA.

package.json
```diff
@@ -1,7 +1,7 @@
 {
 	"name": "ava",
 	"version": "1.0.0-rc.2",
-	"description": "Futuristic test runner 🚀",
+	"description": "Testing can be a slog. AVA's here to help you get it done.",
```
sindresorhus: This is almost perfect. Maybe:

> Testing can be a slog. AVA helps you get it done.

Maybe we could also add something about AVA making it more fun / faster / less painful to do tests.

sindresorhus: I had never heard about the word `slog` before. Is there any other more common word we could use?

> Writing tests can be boring. AVA helps you get it done.

> Testing can be a bore. AVA helps you get it done.
> Testing can be a pain. AVA helps you get it done.
> Testing can be a drag. AVA helps you get it done.

If not, `slog` is fine. Just thought I'd bring it up.

readme.md
sindresorhus: `small API` => `small and intuitive API` ? Or maybe: `concise API`

package.json
sholladay: Isn't it a bit negative? Like the first thing you want me to think about is that testing is painful? I guess what you're going for is empathizing with the user. But I think it could be done with a more positive tone.

> AVA makes testing quick and painless

novemberborn: Jest used to have "painless". They're using "delightful" now. Mocha is "Simple, flexible, fun."

Perhaps I'm cynical but to me testing is something I need to get done. I may get some satisfaction out of covering difficult scenarios, but I don't need it to be fun or delightful.

novemberborn: I like "can be a drag," by the way.

novemberborn:
@sindresorhus pushed changes.

package.json
sholladay: How about "chore"?

> Testing can be a chore. AVA helps you get it done.

Chores aren't necessarily fun, but they're something everyone knows they have to do.

PR #1985, opened 2018-11-19
Fix snapshot skip interface

jamiebuilds:
I'm assuming this is what you meant and it was just a copy and paste error


havenchyk:
@novemberborn ^^

PR #1993, opened 2018-12-02
Updated docs to clarify the usage of CLI options

### LINKED ISSUE #1989: Clarify that not all configuration options can be overridden by CLI flags
[Prompted](https://github.com/avajs/ava/issues/1982#issuecomment-439713094) by @christilut, it's not entirely clear that not all configurations can be overridden by CLI flags. This line should be updated:

> Arguments passed to the CLI will always take precedence over the configuration in package.json.

See https://github.com/avajs/ava/blob/master/docs/06-configuration.md.

itskolli:
This PR Fixes #1989 . Updated the docs to clarify the usage of CLI options.

### DIFF
--- a/docs/06-configuration.md
+++ b/docs/06-configuration.md
@@ -42,7 +42,7 @@ To ignore a file or directory, prefix the pattern with an `!` (exclamation mark)
 }
 ```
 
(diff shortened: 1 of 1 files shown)


novemberborn:
Thank you @itskolli!

PR #1998, opened 2018-12-16
Re-enable some XO lint rules

sindresorhus:


api.js
```diff
@@ -175,6 +175,7 @@ class Api extends Emittery {
 						if (apiOptions.concurrency > 0) {
 							concurrency = apiOptions.concurrency;
 						}
+
```
sindresorhus: This is for the next XO version. See: https://github.com/xojs/eslint-config-xo/pull/52

PR #2002, opened 2018-12-23
Improve Babel and Vue recipes

pearofducks:
Related: #1011 

- Improves the Vue recipe by updating it with Babel v7 support. Adds exclusion of `node_modules` to `require-extension-hooks` which seems to provide a significant speedup. Also adds a note about webpack-aliases here, since this is likely a common use-case (but has one 'gotcha' involved).
- Adds webpack-aliases to the Babel recipe

### DIFF
--- a/docs/recipes/babel.md
+++ b/docs/recipes/babel.md
@@ -208,3 +208,15 @@ Note that loading `@babel/register` in every worker process has a non-trivial pe
 
 
 [Babel options]: https://babeljs.io/docs/en/options
+
+## Use webpack aliases in babel-transpiled files
+
+You'll need to install `babel-plugin-webpack-alias-7` as a devDependency. Then you can add the plugin to your babel config, likely under either `ava.babel.testOptions` or `babel.env.test`.
+
+```json
+{
+	"plugins": [
+		[ "babel-plugin-webpack-alias-7", { "config": "./configs/webpack.config.test.js" } ]
+	]
+}
+```
--- a/docs/recipes/vue.md
+++ b/docs/recipes/vue.md
@@ -5,10 +5,15 @@ Translations: [Français](https://github.com/avajs/ava-docs/blob/master/fr_FR/do
 ## Dependencies
 
 - [Require extension hooks](https://github.com/jackmellis/require-extension-hooks):
-	- `npm i --save-dev require-extension-hooks require-extension-hooks-vue require-extension-hooks-babel`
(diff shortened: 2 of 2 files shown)


docs/recipes/babel.md
```diff
@@ -208,3 +208,15 @@ Note that loading `@babel/register` in every worker process has a non-trivial pe
 
 
 [Babel options]: https://babeljs.io/docs/en/options
+
+## Use webpack aliases in babel-transpiled files
```
sindresorhus: This should also document why one would want this and what problem it solves.

docs/recipes/vue.md
```diff
@@ -5,11 +5,16 @@ Translations: [Français](https://github.com/avajs/ava-docs/blob/master/fr_FR/do
 ## Dependencies
 
 - [Require extension hooks](https://github.com/jackmellis/require-extension-hooks):
-	- `npm i --save-dev require-extension-hooks require-extension-hooks-vue require-extension-hooks-babel`
+	- babel v7: `npm i --save-dev require-extension-hooks require-extension-hooks-vue require-extension-hooks-babel@beta`
+	- babel v6: `npm i --save-dev require-extension-hooks require-extension-hooks-vue require-extension-hooks-babel`
```
sindresorhus: We only document Babel 7.

docs/recipes/vue.md
```diff
@@ -40,9 +45,11 @@ Vue.config.productionTip = false;
 // Setup vue files to be processed by `require-extension-hooks-vue`
 hooks('vue').plugin('vue').push();
 // Setup vue and js files to be processed by `require-extension-hooks-babel`
-hooks(['vue', 'js']).plugin('babel').push();
+hooks(['vue', 'js']).exclude(({filename}) => filename.includes('node_modules')).plugin('babel').push();
```
sindresorhus: This is a fragile way to test for a `node_modules` directory. It will also match paths like `foo-node_modules-bar`.

pearofducks:
@sindresorhus - fixed per your suggestions. Let me know if anything could use further improvement.

Travis and code coverage seem to be randomly failing - unrelated to the changes in this PR.

e.g. Travis fails in `1b8a5ba`, but then codecov fails in `3b5fb9d`.

novemberborn:
@pearofducks I've pushed a commit with edits to your Babel changes. Let me know what you think.

pearofducks:
@novemberborn - looks good, thanks for helping get this in shape!

PR #2008, opened 2019-01-07
Storing error object instead of capturing stack trace

### LINKED ISSUE #1399: Explore storing an error object, rather than directly capturing a stack trace
[Issuehunt badges](https://issuehunt.io/r/avajs/ava/issues/1399)

For asynchronous assertions we capture a stack trace before the assertion is run. Inspired by this comment by @alxandr we should explore storing an error object, and reading the stack trace later if it's actually necessary:

> … instead of passing the stack trace itself, an error object could be passed. This has less performance impact, because then the stack trace is only generated on demand (this might even be something that could improve the performance of your existing async tests).

See [`lib/test.js`](https://github.com/avajs/ava/blob/f6a7b18ad26d988fc85ce51892fd55ce4329edad/lib/test.js#L18) and [`lib/assert.js`](https://github.com/avajs/ava/blob/f6a7b18ad26d988fc85ce51892fd55ce4329edad/lib/assert.js#L32).

> There is a $40.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/avajs/ava/issues/1399).

Ullauri:
Fixes #1399.

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -47,8 +47,10 @@ class AssertionError extends Error {
 		// Reserved for power-assert statements
 		this.statements = [];
 
-		if (opts.stack) {
+		if (typeof opts.stack === 'string') {
 			this.stack = opts.stack;
+		} else if (opts.stack instanceof Error) {
+			this.stack = opts.stack.stack;
 		} else {
 			const limitBefore = Error.stackTraceLimit;
 			Error.stackTraceLimit = Infinity;
@@ -62,10 +64,9 @@ exports.AssertionError = AssertionError;
 function getStack() {
 	const limitBefore = Error.stackTraceLimit;
 	Error.stackTraceLimit = Infinity;
-	const obj = {};
-	Error.captureStackTrace(obj, getStack);
+	const error = new Error();
 	Error.stackTraceLimit = limitBefore;
-	return obj.stack;
+	return error;
 }
 
 function validateExpectations(assertion, expectations, numArgs) { // eslint-disable-line complexity
--- a/lib/test.js
+++ b/lib/test.js
@@ -13,13 +13,12 @@ function formatErrorValue(label, error) {
 	return {label, formatted};
 }
 
-const captureStack = start => {
+const captureStack = () => {
 	const limitBefore = Error.stackTraceLimit;
 	Error.stackTraceLimit = 1;
-	const obj = {};
-	Error.captureStackTrace(obj, start);
+	const error = new Error();
 	Error.stackTraceLimit = limitBefore;
-	return obj.stack;
+	return error;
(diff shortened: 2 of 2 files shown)


lib/assert.js
```diff
@@ -47,8 +47,10 @@ class AssertionError extends Error {
 		// Reserved for power-assert statements
 		this.statements = [];
 
-		if (opts.stack) {
+		if (typeof opts.stack === 'string') {
```
novemberborn: I'd think this would no longer be the case. Is it still needed?

lib/assert.js
```diff
@@ -47,8 +47,10 @@ class AssertionError extends Error {
...
 			this.stack = opts.stack;
+		} else if (opts.stack instanceof Error) {
+			this.stack = opts.stack.stack;
```
novemberborn: It may make sense to rename `opts.stack`. Perhaps `opts.objectWithStack`? Same for the `stack` variable and `getStack()` function.

lib/assert.js
```diff
@@ -62,10 +64,9 @@ exports.AssertionError = AssertionError;
 function getStack() {
 	const limitBefore = Error.stackTraceLimit;
 	Error.stackTraceLimit = Infinity;
-	const obj = {};
-	Error.captureStackTrace(obj, getStack);
+	const error = new Error();
```
novemberborn: I'm not sure how this impacts the captured stack trace. Perhaps the `AssertionError` constructor needs to do this `stackTraceLimit` dance, and explicitly capture the stack trace excluding the `getStack()` function?

lib/test.js
```diff
@@ -13,13 +13,12 @@ function formatErrorValue(label, error) {
 	return {label, formatted};
 }
 
-const captureStack = start => {
+const captureStack = () => {
 	const limitBefore = Error.stackTraceLimit;
 	Error.stackTraceLimit = 1;
-	const obj = {};
-	Error.captureStackTrace(obj, start);
+	const error = new Error();
```
novemberborn: Same here, I reckon we should still capture the stack trace without the `start` function.

lib/assert.js
Ullauri: I was looking at [this](https://github.com/avajs/ava/blob/master/lib/test.js#L280) when including that condition.

lib/assert.js
Ullauri: Agreed.

lib/assert.js
Ullauri: That would mean getting rid of `getStack()` no?

novemberborn:
Hey what happened @Ullauri? Anything I can do to help?

Issue #1428, opened 2017-06-28
Run in worker threads


Qix-:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%24300%20Funded-%2300A156.svg" />](https://issuehunt.io/r/avajs/ava/issues/1428)

**See https://github.com/avajs/ava/issues/1428#issuecomment-633456574**

---

<del>

Right now it appears the process is forked for each test and then the test is ran. This is unideal for a vast amount of reasons.

Upon discussion with Sindre, it was decided that we should explore moving into a pooled approach (where instances of the following classes shared a common interface and abstracted away the strategy for executing individual tests):

- `SingleProcessTestPool` would run all tests in the main process without forking
- `ForkTestPool` would fork the process `N` times (where `N` is the number of processors) and distribute tests out between them as they become idle

In the future, this could be expanded to a `WorkerTestPool` to implement support for [workers](/nodejs/worker).

This would give us the ability to decide upon the execution strategy based on a number of factors:

- For `<N` tests, use `SingleProcessTestPool`; otherwise, use `ForkTestPool`
- When `--no-fork` or equivalent flag is provided, force `SingleProcessTestPool`
- When `--inspect` or `--inspect-brk` is specified, force `SingleProcessTestPool` (ref #1108)

I'm not entirely sure how much work this would require, but consider this issue a 'jumping off' point for discussion.

</del>

**EDIT:** This might be of interest as well: https://nodejs.org/api/worker_threads.html

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $300.00)

- [<img src='https://avatars.githubusercontent.com/u/170270?v=4' alt='sindresorhus' width=24 height=24> sindresorhus](https://issuehunt.io/u/sindresorhus) ($60.00)
- [<img src='https://avatars3.githubusercontent.com/u/44827199?v=4' alt='issuehunt' width=24 height=24> issuehunt](https://issuehunt.io/u/issuehunt) ($140.00)
- [<img src='https://avatars1.githubusercontent.com/u/4330357?v=4' alt='futpib' width=24 height=24> futpib](https://issuehunt.io/u/futpib) ($100.00)

### Submitted pull Requests
- [#2011 Migrate forking to a pooled worker design #1428](https://issuehunt.io/r/avajs/ava/pull/2011)
- [#2690 feat: run in worker threads](https://issuehunt.io/r/avajs/ava/pull/2690)
---

### Tips

[...]

Issue #1451, opened 2017-07-13
Run tests with Electron


bokuweb:
Hi. i'm developing `Avaron`(https://github.com/bokuweb/avaron), which forked from `vdbwouter/electron-ava`, to run tests with electron .(plase see also, `why`? (https://github.com/bokuweb/avaron#why )).It works with AVA v0.21+.

    
`Avaron` is not forked from `AVA`, but copy and edit below files from `AVA` to launch electron.

- avaron/lib/cli.js (https://github.com/bokuweb/avaron/blob/master/lib/cli.js#L166-L169)
- avaron/lib/fork.js (https://github.com/bokuweb/avaron/blob/master/lib/fork.js#L59-L66)
   
If I can hook into the below point, I may be able to do Electron support without having to fork/copy from AVA. Is there a good idea?

#### avaron/lib/cli.js

I needed to add options for `electron` (`renderer`, `windowOptions` and `fixture` for now).

#### avaron/lib/fork.js 

I needed to modify `child_process.spawn` to launch electron process.

### related

https://github.com/avajs/awesome-ava/pull/21

novemberborn:
@bokuweb this is really cool!

I'm wondering how we can best support this. I'm not convinced AVA itself should support Electron as a process target. Perhaps though the "test runner pool" could be abstracted so that other pools can be provided by users. In your case that would be an Electron pool. We'd have to determine how to provide CLI flags and `package.json` options as well. We'd also have to standardize the communication interface between the main AVA process and the test workers.

This is lower-level than #1043 but it could be quite interesting. It may work for Wallaby.js as well.

@avajs/core, @ArtemGovorov, thoughts?

ArtemGovorov:
@novemberborn I like the idea of the abstracted "test runner pool". It might make it easier for wallaby (as well as others) to integrate with AVA. I think it is generally a good thing to separate the testing framework part from the test runner part, makes it easier to reuse/hook into/integrate with those parts.

sindresorhus:
Relevant issue: https://github.com/avajs/ava/issues/1428

novemberborn:
> Relevant issue: #1428

Heh I wonder if that was still in by subconscious 👍

novemberborn:
We're focusing on running just Node.js code for the foreseeable future.

PR #2011, opened 2019-01-10
Migrate forking to a pooled worker design #1428

ijjk:
Here's the start of my implementation for migrating forking to a pooled worker design. My implementation has a `SingleProcessTestPool`, `SharedForkTestPool`, and `ForkTestPool` which share a common interface to try to help abstract away the method of executing. 

~~Currently, it picks a `TestPool`  to use based on the factors set out on the issue~~:

- ~~For <N tests (where N is the number of processors), use SingleProcessTestPool; otherwise, use ForkTestPool~~
- ~~When --no-fork or equivalent flag is provided, force SingleProcessTestPool~~
- ~~When --inspect or --inspect-brk is specified, force SingleProcessTestPool~~

~~In the thread, the `ForkTestPool` was described as distributing the tests to the different forks as they become idle, and I thought it might simplify the execution to divide the tests beforehand and then create the forks to run them. Doing it this way allowed me to re-use the `SingleProcessTestPool` implementation inside of the forked process~~.

~~To achieve the testing inside the single process I updated the import of the `runner` to use a map with the test file as the key and the runner as the value. I also, utilized `vm` to provide isolation between running the tests in the same thread~~. Fixes: #1428 Fixes: #1332 

After review by @novemberborn, there are now no breaking changes and use of the new test pools are opt-in. Also, the `SingleProcessTestPool` does not run inside of a sandbox e.g. `vm` as to avoid having to maintain a sandboxing library since `vm2` does not work for this context. You can opt-in to a test pool with below configs/flags

- `SingleProcessTestPool` using the `--single-process` flag or `singleProcess` config
- `SharedForkTestPool` using the `--share-forks` flag or `shareForks` config

You can also make use of `worker_threads` now on supported node versions (>= 11.8 or > 10.x with `--experimental-worker` node flag) using the `--worker-threads` flag or `workerThreads` config

TODO:
- [x] Add note for `--no-fork` flag and `fork` config in docs
- [x] Clean up timeouts
- [x] Fix failing test cases


novemberborn:
Hey @ijjk thanks for picking this up! There's a lot of changes here so I may not be able to have a proper look until next weekend. Going by the commit messages it looks like you're really far along already!

ijjk:
@novemberborn okay thanks for the reply! Yeah, the changes started to add up. I'm still sorting out some test cases, and then hopefully it will be good to go.

lo1tuma:
I’ve tried to run this against the test-suite of one of our projects with 1608 tests in total. While ava v1 takes round about 37 seconds, with this changes it seems to be extremely fast (my gut feeling < 5 seconds with the ForkTestPool). The problem is that after `1222` tests have been completed the runner completely hangs and I have to manually abort to process. I see some warnings on the console:

```
 MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 25 unhandledRejection listeners added. Use emitter.setMaxListeners() to increase limit
```

The same happens when I use the `--no-fork` option.

Apart from that the reporting seems to behave strange, does it show a spinner for every process? Here is  a screenshot:
screen shot 2019-01-23 at 16 53 18

As you can see in the screenshot there is also a warning from react which complains about missing `requestAnimationFrame`. I don’t get this warning with ava v1. I would guess this is related to sandboxing. In some test files I set `global.window` to an instance of `jsdom`.

As a follow up feature it would be also nice to provide an option which disables sandboxing in favor of performance.

index.js
```diff
@@ -1,8 +1,21 @@
...
+	// Show error
+	/* eslint-disable-next-line import/no-unassigned-import */
+	require('./lib/worker/ensure-forked');
+}
+
+// Get name of test file
+const {filename} = module.parent;
+
+if (filename.indexOf('node_modules/ava') > -1) {
+	// If request was routed by a different ava install to
```
novemberborn: `AVA` please 😄

index.js
```diff
@@ -1,8 +1,21 @@
...
+	// this one just export our worker/main
 	module.exports = require('./lib/worker/main');
+} else if (process.env.AVA_PATH && process.env.AVA_PATH !== __dirname) {
+	// Ensure the same AVA install is loaded by the test file as by the test worker
+	module.exports = require(process.env.AVA_PATH)();
```
novemberborn: In the original code, the required module is not invoked. Why is it invoked here?

index.js
```diff
@@ -1,8 +1,21 @@
...
+} else {
+	module.exports = require('./lib/worker/main')();
```
novemberborn: Should the `main` module be invoked?

lib/babel-pipeline.js
```diff
@@ -44,6 +43,8 @@ function validate(conf) {
 		return {testOptions: defaultOptions};
 	}
 
+	const chalk = require('./chalk').get();
```
novemberborn: Why did this need moving?

profile.js
```diff
@@ -69,97 +36,9 @@ if (cli.input.length === 0) {
...
-process.argv[2] = JSON.stringify(opts);
-process.argv.length = 3;
-
-if (console.profile) {
-	console.profile('AVA test-worker process');
+// Prefer the local installation of AVA
+if (importLocal(__filename)) {
+	debug('Using local install of AVA');
+} else {
+	require('./lib/cli').run({fork: false});
```
novemberborn: I think we can remove `profile.js` with this PR. It's pretty awkward and you could now run `npx ava --no-fork --inspect my-test.js` instead.

Then we can also remove the `overrideConfig` from the `lib/cli`.

lib/fork.js
```diff
@@ -1,132 +1,107 @@
...
-				emitStateChange(message.ava);
-			}
-		});
+const subprocessPath = require.resolve('./worker/subprocess.js');
 
-		subprocess.on('error', err => {
-			emitStateChange({type: 'worker-failed', err});
-			finish();
-		});
+const failTypes = {
```
novemberborn: Use a `Set` instead.

lib/fork.js
```diff
@@ -1,132 +1,107 @@
...
+			});
+
+			this.subprocess.stderr.on('data', chunk => {
+				runStatus.emitStateChange({type: 'worker-stderr', chunk});
+			});
+
+			// Listen for messages
+			this.subprocess.on('message', msg => {
+				if (msg.forkMsg === 'new-file') {
+					runStatus.observeWorker({onStateChange: () => {}}, msg.file);
```
novemberborn: Looks like `run-status` needs updating?

lib/fork.js
```diff
@@ -1,132 +1,107 @@
...
+				} else if (msg.type) {
+					runStatus.emitStateChange(msg);
+					// Check if we need to trigger exits
+					if (failFast && failTypes[msg.type]) {
+						reject(new Error('notify-fail'));
```
novemberborn: This should be decided higher-up I think.

lib/testpool/fork-test-pool.js
```diff
@@ -0,0 +1,118 @@
...
+const debug = require('debug')('ava:pool');
+const {Fork} = require('../fork');
+const TestPool = require('./test-pool');
+
+// Break up files into chunks and
+// distribute to forks to run as SingleProcess
+class ForkTestPool extends TestPool {
+	bail() {
+		debug('bailing forks');
+		this.forks.map(fork => fork.exit());
```
novemberborn: Use a `for`/`of` loop.

lib/testpool/fork-test-pool.js
```diff
@@ -0,0 +1,118 @@
...
+			options.precompiled = {};
+		}
+
+		options.concurrency = concurrency;
+		this.forks = [];
+
+		return Bluebird.map(fileChunks, curFiles => {
+			return this._computeForkExecArgv().then(execArgv => {
+				this.forks.push(
+					new Fork(curFiles, runStatus, this.env, this.initOptions, execArgv)
```
novemberborn: This… gives a single `Fork` the responsibility to run multiple test files? I can't quite tell how `Fork` implements this but it also seems a little odd. You could end up with a slow fork that still has two pending files while other slots are idle.

lib/testpool/fork-test-pool.js
```diff
@@ -0,0 +1,118 @@
...
+				);
+			});
+		}).then(() => {
+			return Bluebird.map(this.forks, fork => {
+				return fork.run().catch(() => {
+					if (failFast) {
+						this.forks.map(fork => fork.exit());
```
novemberborn: Use `for`/`of`.

lib/testpool/fork-test-pool.js
```diff
@@ -0,0 +1,118 @@
...
+						return Promise.reject(new Error('clean-exit'));
+					}
```
novemberborn: Need to rethrow other errors.

lib/testpool/fork-test-pool.js
```diff
@@ -0,0 +1,118 @@
...
+				this.forks.push(
+					new Fork(curFiles, runStatus, this.env, this.initOptions, execArgv)
+				);
+			});
+		}).then(() => {
+			return Bluebird.map(this.forks, fork => {
+				return fork.run().catch(() => {
+					if (failFast) {
+						this.forks.map(fork => fork.exit());
+						return Promise.reject(new Error('clean-exit'));
```
novemberborn: You can throw, no need to return a rejected promise.

lib/testpool/fork-test-pool.js
```diff
@@ -0,0 +1,118 @@
...
+					}
+				});
+			});
+		})
+			.catch(err => {
+				// Exit cleanly if it's handled, if not re-throw
+				if (err.message === 'clean-exit') {
```
novemberborn: I'm not a fan of this approach. I'd rather build on the state change notification mechanism to then interact with specific forks or the pool.

lib/testpool/fork-test-pool.js
```diff
@@ -0,0 +1,118 @@
...
+					return;
+				}
+
+				throw new Error(err);
```
novemberborn: `throw err`

lib/testpool/single-process-test-pool.js
```diff
@@ -0,0 +1,135 @@
...
+const {failTypes} = require('../fork');
+const precompilerHook = require('../worker/precompiler-hook');
+const TestPool = require('./test-pool');
+const isolatedRequire = require('./isolated-require');
+
+const preTestPath = require.resolve('./test-pre-require.js');
+const postTestPath = require.resolve('./test-post-require.js');
+
+const preTestContent = fs.readFileSync(preTestPath, 'utf8').toString();
+const postTestContent = fs.readFileSync(postTestPath, 'utf8').toString();
```
novemberborn: Since you provided the `'utf8'` encoding `readFileSync()` already returns strings.

lib/testpool/single-process-test-pool.js
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	bail() {
+		debug('bailing single process');
+		this.resolves.forEach(resolve => resolve());
+	}
+
+	run() {
+		const {options, runOnlyExclusive} = this;
+		const concurrency = options.serial ? 1 : 2;
```
novemberborn: I'm not sure we need to have any test file concurrency within the process. That would also simplify the implementation.

lib/testpool/single-process-test-pool.js
```diff
@@ -0,0 +1,135 @@
...
+
+		if (maxListeners < concurrency + 2) {
+			process.setMaxListeners(maxListeners + concurrency);
+		}
+
+		precompilerHook.install(precompiled);
+
+		return Bluebird.map(this.files, file => {
+			// Resolve realpath in case it's symlink
+			file = fs.realpathSync(file);
```
novemberborn: I don't think we do this any other time? Is it because when you launch a child process that already takes care of resolving the symlink?

lib/testpool/single-process-test-pool.js
```diff
@@ -0,0 +1,135 @@
...
+
+			// Check if we are bailing
+			if (this.bailed) {
+				return Promise.resolve();
```
novemberborn: You can just return.

lib/testpool/single-process-test-pool.js
```diff
@@ -0,0 +1,135 @@
...
+			}
+
+			// Let fork know to create file in runStatus
+			if (typeof process.send === 'function') {
+				process.send({forkMsg: 'new-file', file});
```
novemberborn: Why communicate through `process`?

lib/testpool/single-process-test-pool.js
```diff
@@ -0,0 +1,135 @@
...
+							emitStateChange: change => this.emitStateChange(change, file),
+							handleExitTest: type => this.handleExitTest(type, resolve, reject)
+						}
+					});
+
+					this.emitStateChange({type: 'dependencies', dependencies: [...dependencies]}, file);
+				} catch (err) {
+					let error = err;
+					if (error.message === 'unexpected-exit') {
+						error = {type: 'worker-failed'};
```
novemberborn: Errors should be errors. Let's find a different way of communicating these events.

lib/testpool/test-pool.js
```diff
@@ -0,0 +1,88 @@
+const path = require('path');
+const debug = require('debug')('ava:pool');
+const serializeError = require('../serialize-error');
+
+if (!process.env.NODE_ENV) {
+	process.env.NODE_ENV = 'test';
```
novemberborn: This changes the current behavior. If `process.env.NODE_ENV === ''`, that should be preserved.

lib/testpool/test-pool.js
```diff
@@ -0,0 +1,88 @@
...
+	bail() {
+		throw new Error('not implemented');
+	}
+
+	emitStateChange(evt, file) {
+		if (!evt.testFile && file) {
+			evt.testFile = file;
+		}
+
+		// Debug('emitting', evt)
```
novemberborn: ?

ijjk:
@novemberborn 

> What do you mean by having to use "`require" in the host context"?

`vm2`'s `NodeVM` has the option to require in the `sandbox` or in the `host` context. When using the `sandbox` require mode, modules that need to modify the `module.constructor._extensions` like `@babel/register` fail since `vm2` doesn't expose a full `module` instance. 

You can use the `host` require mode and it works but this implementation breaks pretty much any sandboxing since any changes made globally or to the module on require are done in the host and not the sandbox. 

So for example, in `host` require mode, if you require a module let's call it `browser-setup.js` with below contents, the changes will be made in the `host` context and not the `sandbox` pretty much breaking the isolation.

```js
const { JSDOM } = require('jsdom')
const { window, document } = new JSDOM(`<!DOCTYPE html><p>Hello world</p>`)
global.window = window
global.document = document
global.fetch = require('node-fetch')
```

I couldn't find any existing modules that handle this.

- `vm2` has above problem

- `isolated-vm` runs in a fork and according to the [README](https://github.com/laverdet/isolated-vm#requirements) mainly works with node >= 8.11.2

- `napajs` runs on seperate threads and doesn't support all built-in node_modules

This is why I started working out a bare minimum custom implementation of `vm`.

novemberborn:
I see.

What do you think of https://nodejs.org/api/worker_threads.html? That should solve the sandboxing and the performance issues. I still see a need for properly forked processes, so the pool switching will be necessary. Perhaps we could support non-sandboxed single process as an opt-in which only works for some use cases.

ijjk:
@novemberborn `WorkerThreads` look interesting although still experimental. I re-worked my branch to support a non-isolated single process mode, the current mode of creating a new fork for each test, and a shared fork mode which reuses the forks and distributes tests to them as they become available. 

The single process mode and shared fork modes are opt-in currently. I could try out a `WorkerThread` implementation that's opt-in too since it's still in the experimental stage. Was this re-work what you had in mind with the previous comment?

novemberborn:
> I could try out a `WorkerThread` implementation that's opt-in too since it's still in the experimental stage. Was this re-work what you had in mind with the previous comment?

Yea. It no longer requires a flag in the recent Node.js 11 release(s), so it's definitely an option for us. It'll be in Node.js 12 soon enough!

ijjk:
@novemberborn I added the option to use Worker Threads instead of forking. The API was pretty similar so I was able to implement it using `./lib/fork`.

It's also set up so that Worker Threads can be shared in the same way that forks are with the `--share-forks` flag since creating new Worker Threads can be expensive on performance similar to creating new forks. 

On one project's test suite that has around 632 tests across 59 files, that is using AVA, [I found on github](https://github.com/freaktechnik/justintv-stream-notifications), the suite had below run times. Note, by no-cache I mean the cache under `node_modules/.cache` has been cleared for the project.

| Mode | Run duration in seconds |
| ---- | ----------------------- |
| `ForkTestPool` no cache | 48.67s |
| `ForkTestPool` with cache | 38.34s |
| `SharedForkTestPool` no cache | 22.09s |
| `SharedForkTestPool` with cache | 14.91s |
| `WorkerThreads` no cache | 48.43s |
| `WorkerThreads` with cache | 35.76s |
| `SharedWorkerThreads` no cache | 21.51s |
| `SharedWorkerThreads` with cache | 15.60s |
| `SingleProcessTestPool` no cache | Suite can't run in single process |
| `SingleProcessTestPool` with cache | ^ |


I also ran this on [styled-jsx's](https://github.com/zeit/styled-jsx) test-suite which is around 174 tests across 16 files and got below run times.

| Mode | Run duration in seconds |
| ---- | ----------------------- |
| `ForkTestPool` no cache | 10.73s |
| `ForkTestPool` with cache | 6.34s |
| `SharedForkTestPool` no cache | 8.80s |
| `SharedForkTestPool` with cache | 6.30s |
| `WorkerThreads` no cache | 10.84s |
| `WorkerThreads` with cache | 6.96s |
| `SharedWorkerThreads` no cache | 9.17s |
| `SharedWorkerThreads` with cache | 6.37s |
| `SingleProcessTestPool` no cache | 8.11s |
| `SingleProcessTestPool` with cache | 4.36s |

I didn't run these under a super controlled environment so they might not be perfect estimates but thought it might be helpful to compare on actual projects using AVA.

lib/babel-pipeline.js
ijjk: This was left over from some previous debugging sorry

profile.js
ijjk: I was thinking it could be removed also. I removed it and updated the docs with the new command.

lib/fork.js
ijjk: After re-working the branch, I don't think `run-status` need updating anymore. Do you think it still does?

lib/testpool/single-process-test-pool.js
ijjk: This ended up getting removed, but good to know.

lib/test-pools/single-process-test-pool.js
```diff
@@ -0,0 +1,175 @@
...
+		global.getRunner = () => {
+			accessedRunner = true;
+			return runner;
+		};
+
+		// Store value in case to prevent required modules from modifying it.
+		const testPath = file;
+
+		// Give runStatus a chance to setup listener
+		setTimeout(() => {
```
iwilson-r7: I believe this should use `nowAndTimers.setTimeout`, otherwise use of `sinon.useFakeTimers` and forgetting to restore the original timer functions in a test can cause the test run to either hang inexplicably or exit without completing (a difficult to debug problem as I discovered!)

ijjk: Good catch, I actually updated it to use `nowAndTimers.setImmediate` since it just needs to run after `runStatus.observeWorker` has been called.

novemberborn:
Hey @ijjk sorry for the delay in getting back to you. It's quite a lot to take in! 

I've left some more comments, and I'm adding a checklist below. I think we may be getting to a point where I need to play around with the code and perhaps make some more changes. If you could point out the parts you're not comfortable with yet that'd be a great help.

Since this is such an involved PR it may take a while before I can really sit down for this. I really appreciate the work you're putting into this though!

Checklist:

* [ ] Review debug() output
* [ ] Review documentation (probably needs a dedicated section / recipe)
* [ ] Bikeshed CLI flags
* [ ] Review error message for when worker threads are unavailable
* [ ] "Shared fork" reusing fork leads to complicated code paths, both in the pool code and in the `Fork` class, the latter since it's used both in shared and non-shared modes
* [ ] Consider possible code re-use between subprocess code and single process test pool
* [ ] Review tests

SneakyMax:
I wanted to try out this branch, so If you'd like some more anecdotal numbers (Node 11.9.0):

### Vue Project

We have a Vue project that has 180 tests over 101 files. We use the `t.snapshot` feature a lot:

- `ava` - 1:36, all tests pass
- `ava --worker-threads` - Cannot find any tests in any of the files.
- `ava --share-forks` - Hangs after 0:25 with 166 tests passed, 2 failed. Error in the Vue runtime file.
- `ava --single-process` - Exits after 0:28 with 68 tests passed, 0 failed. Same error as before, except this time it quits the whole process.

### Node Project

We have a GraphQL API that extensively clears the require cache during its tests, and all tests run serially in each file because they use global state. It has 1447 tests over 96 files:

- `ava` - 2:52, all tests pass
- `ava --worker-threads` - Cannot find any tests in any of the files
- `ava --share-forks` - Froze my VM
- `ava --single-process` - JavaScript heap out of memory after 333 tests

ijjk:
@SneakyMax thanks for trying out the branch! If you have a chance could you try running them again with the latest changes? 

I'm not able to reproduce the `--worker-threads` not being able to find tests error so any more info would be appreciated. I was able to reproduce the hanging and running out of memory errors and it appears there was a memory leak in the way I was clearing `require.cache` between test files which should be fixed now. 

You might need to clear the cache under `node_modules/.cache/ava` before trying again.

ijjk:
@novemberborn I understand it is a lot. I started looking over the checklist and tried addressing some of the points below.

> Review documentation (probably needs a dedicated section / recipe)

I can add more verbose info to the documentation just have been waiting until everythings been agreed on. 

> Bikeshed CLI flags

I was thinking it might be good to instead of having cli flags for each test pool to just have a test pool flag e.g. `--test-pool singleProcess` or `--test-pool sharedForks`. With the test pool flag we could also allow passing in a path to a custom test pool to achieve something like #1451 I'm open to thoughts on this though.

> "Shared fork" reusing fork leads to complicated code paths, both in the pool code and in the `Fork` class, the latter since it's used both in shared and non-shared modes

Are you thinking it would be better to have a seperate fork module for shared forks? I was mainly trying to avoid duplicating code. 

> Consider possible code re-use between subprocess code and single process test pool

I just pushed up some updates and was able to re-use the code in subprocess with single process test pool.

SneakyMax:
Sure thing @ijjk!

### Vue Project

- `ava` - 1:37, all tests pass
- `ava --worker-threads` - 1:49, all tests pass. Slower :(
- `ava --share-forks` - Hangs after 169/180 tests pass. A couple tests errored (`Error in nextTick: "TypeError: Failed to execute 'appendChild' on 'Node': parameter 1 is not of type 'Node'."`) (jsdom?)
- `ava --single-process` - Error after 68 tests pass: `node_modules/vue/dist/vue.runtime.common.js:7005
  var transitionDelays = styles[transitionProp + 'Delay'].split(', ');`

`--share-forks` is significantly slower than the previous time I tested, unfortunately. Whatever cache changes you made seem to have undone any performance benefit.

Looks like you fixed `--worker-threads` somehow though.

### Node Project

- `ava` - 2:43, all tests pass.
- `ava --worker-threads` - 3:14, all tests pass. Slower.
- `ava --share-forks` - VM froze after 196 tests passed.
- `ava --single-process` - Heap out of memory after 333 tests.

The out of memory problems might actually be that whatever require cache busting I'm using has a memory leak, and it never happens during normal ava because the worker processes shut down, taking the memory leaks with them.

ijjk:
@SneakyMax Interesting 🤔on the node project you might be right about having a memory leak in your require cache clearing also. I had to make sure to remove the module from the parent module's children to get it to free up. It's weird that the shared forks froze up faster than the single process though since the shared forks should have more memory available unless your environment has restricted memory so the shared forks hit it faster maybe.  

The Vue errors have me curious, I wonder if something is being modified globally that I'm not catching somehow. Is either of these able to be shared by any chance or possibly a minimum example of the set up so I can try investigating a little more?

api.js
```diff
@@ -204,43 +206,50 @@ class Api extends Emittery {
...
+						} else {
+							workerOptions.precompiled = {};
+						}
+
+						if (runtimeOptions.updateSnapshots) {
+							// Don't use in Object.assign() since it'll override options.updateSnapshots even when false.
+							workerOptions.updateSnapshots = true;
+						}
+
+						this.Fork = Fork;
```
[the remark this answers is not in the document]
ijjk: This is mostly for mocking during testing. I thought it would be easier only mock here instead of each `test-pool` that needs `Fork`

api.js
```diff
@@ -204,43 +206,50 @@ class Api extends Emittery {
...
-								worker.promise.then(() => { // eslint-disable-line max-nested-callbacks
-									pendingWorkers.delete(worker);
-								});
-								restartTimer();
-
-								return worker.promise;
-							});
-						}, {concurrency});
+						// Default to using a new fork for each test (provides most isolation)
+						let ProcessPool = ForkTestPool;
```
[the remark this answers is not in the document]
ijjk: Yeah that's better 👍

lib/cli.js
```diff
@@ -38,6 +38,9 @@ exports.run = () => { // eslint-disable-line complexity
 		  --timeout, -T           Set global timeout
 		  --serial, -s            Run tests serially
 		  --concurrency, -c       Max number of test files running at the same time (Default: CPU cores)
+		  --single-process        Run tests in single process (not isolated)
+		  --share-forks           Run tests in shared forks
```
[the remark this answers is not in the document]
ijjk: Yup, updated

lib/cli.js
```diff
@@ -38,6 +38,9 @@ exports.run = () => { // eslint-disable-line complexity
...
+		  --worker-threads        Run tests in worker threads
```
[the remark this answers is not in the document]
ijjk: Updated

lib/cli.js
```diff
@@ -208,7 +223,10 @@ exports.run = () => { // eslint-disable-line complexity
 		color: conf.color,
 		workerArgv: cli.flags['--'],
 		parallelRuns,
-		ranFromCli: true
+		ranFromCli: true,
+		shareForks: conf.shareForks,
+		singleProcess: conf.singleProcess,
+		workerThreads: conf.workerThreads
```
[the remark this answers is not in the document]
ijjk: Yeah that makes sense, combining them can add a lot of complexity

lib/test-pools/single-process-test-pool.js
```diff
@@ -0,0 +1,184 @@
...
+		}, {concurrency: 1});
+	}
+
+	_worker(file, options) {
+		// Clear any global variables from previous test
+		revertGlobal();
+
+		const runner = new Runner({
+			file,
+			scheduledStart: true,
```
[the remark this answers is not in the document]
ijjk: Updated

lib/test-pools/single-process-test-pool.js
```diff
@@ -0,0 +1,184 @@
...
+			});
+			this.restartTimer();
+
+			return worker.promise;
+		}, {concurrency: 1});
+	}
+
+	_worker(file, options) {
+		// Clear any global variables from previous test
+		revertGlobal();
```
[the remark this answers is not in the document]
ijjk: Updated

lib/fork.js
```diff
@@ -31,107 +31,157 @@ const describeTTY = tty => ({
...
-		emitStateChange({type: 'worker-stdout', chunk});
-	});
+// Fork manages creating a subprocess for running tests
+class Fork {
+	constructor(file, opts, execArgv) {
+		this.file = file;
+		this.finished = false;
+		this.forcedExit = false;
+		this.emitter = new Emittery();
+		this.promise = new Promise((resolve, reject) => {
```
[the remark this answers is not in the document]
ijjk: This was named `promise` to match the current `fork.promise`. We can rename it just we would also have to update it in `./api.js` and everywhere else.

lib/test-pools/shared-fork-test-pool.js
```diff
@@ -0,0 +1,63 @@
...
+					this.forks[i].promise.then(() => resolve());
+				}));
+			}));
+		}
+
+		debug('using ' + this.forks.length + ' forks, for ' + files.length + ' files');
+
+		// Wait for forks to be initialized
+		return Bluebird.all(this.forkPromises).then(() => {
+			return Bluebird.map(files, file => {
```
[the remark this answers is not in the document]
ijjk: Yeah that makes sense, updated

lib/test-pools/single-process-test-pool.js
```diff
@@ -0,0 +1,184 @@
...
+		};
+
+		process.on('unhandledRejection', unhandledRejectionHandler);
+
+		runner.on('error', error => {
+			statusListener({type: 'internal-error', err: serializeError('Internal runner error', false, error)});
+			exit(1);
+		});
+
+		/* istanbul ignore next */
```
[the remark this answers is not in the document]
ijjk: Removed all except those related to `worker_threads`, since I can't test them on unsupported node versions it brings down overall code coverage

lib/test-pools/single-process-test-pool.js
```diff
@@ -0,0 +1,184 @@
...
+
+		// Give runStatus a chance to setup listener
+		nowAndTimers.setImmediate(() => {
+			try {
+				require(testPath);
+				// Emit dependencies
+				const {children} = require.cache[file] || {};
+				if (children) {
+					const dependencies = children.map(mod => mod.id);
+					statusListener({type: 'dependencies', dependencies});
```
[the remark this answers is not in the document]
ijjk: This has been updated now that I'm re-using subprocess for `SingleProcessTestPool`

lib/test-pools/test-pool.js
```diff
@@ -0,0 +1,99 @@
...
+
+	/**
+   * Worker that is added to pendingWorkers, is not called externally
+   * @returns {Object} worker - shared worker interface
+   * @returns {string} worker.file - name of file run in worker
+   * @returns {Promise} worker.promise - promise that resolve on worker finish
+   * @returns {Function} worker.exit - function to exit worker
+   * @returns {Function} worker.onStateChange - function called to add stateChange callback
+   * @returns {Function} worker.notifyOfPeerFailure - function to try to stop currently scheduled tests.
+   */
```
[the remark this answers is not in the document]
ijjk: Not really, mainly there to be verbose. Removed.

lib/worker/main.js
```diff
@@ -1,5 +1,12 @@
 'use strict';
-const runner = require('./subprocess').getRunner();
+// If getRunner is missing AVA was probably required
+// directly instead of in a test
+if (typeof global.getRunner === 'undefined') {
```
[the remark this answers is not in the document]
ijjk: I updated this to only be used in `SingleProcess` mode to fix an issue with `esm` causing the `getRunner` export not to be set here.

SneakyMax:
@ijjk sorry about the late reply - unfortunately I can't share either of the projects :(, I'm happy to keep testing the branch though. I hope we can get the speedup I first experienced, without any errors.

lib/test-pools/single-process-test-pool.js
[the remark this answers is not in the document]
novemberborn: I'd have expected `codecov` to combine coverage reports across all the Node.js versions we test on. This isn't happening?

ijjk:
@SneakyMax I understand. I added an option `cacheRequire` that defaults to `true` which allows you to either require `options.require` everytime before new tests or to only require them once before all tests in that process. 

Note: this is only used in `sharedForks` and `singleProcess` mode. This might help increase performance a bit if your tests are set up to be able to allow `cacheRequire` to be enabled.

lib/revert-global.js
```diff
@@ -0,0 +1,47 @@
+// Track global and require.cache keys and
+// removes any new keys when revert is called
+function revertGlobal() {
+	const storedKeys = {};
+	const storedCache = Object.assign({}, require.cache);
+	const storedExtensions = Object.assign({}, module.constructor._extensions);
+	const preventCache = ['bluebird', '../index.js', './worker/main.js'];
```
Retsam: Why is Bluebird being prevented from caching? 

I've tried running my tests against this branch, and all of our tests that involve bluebird cancellation are failing: we currently have a 'testBootstrap.js" file the we import before all of our tests which enables bluebird cancellation, but it's not being respected, likely due to this logic here.

Retsam: Perhaps `.getNewLibraryCopy()` would be useful so that the test framework can have its own copy of Bluebird without stepping on the application code?

novemberborn:
Hi @ijjk, sorry I haven't gotten round to picking this up again.

lib/revert-global.js
ijjk: I was clearing bluebird there to try to not interfere with the test's usage of bluebird but you might be right that clearing it there ended up interfering anyways. I moved it to only clear initially so hopefully it doesn't conflict now.

Retsam: Yup, it looks like that fixed it.  All of my cancellation tests are running correctly now.

ijjk:
@novemberborn I just updated the branch to resolve conflicts, it should be good to review when you have a chance.

thasmo:
Awesome work @ijjk! Hope this gets reviewed soon. 👍

novemberborn:
> Hope this gets reviewed soon. 👍

Yes me too 😛 

This is the next big PR I'm meaning to look at once we land https://github.com/avajs/ava/pull/1947. Apologies for taking so long. It's really hard to find the time (several uninterrupted hours) to go through PRs with this kind of impact.

novemberborn:
Once more, apologies for taking so long in looking at this.

The bad news is, this is still a really large PR! But, the good news is that we're now supporting experimental features. See this commit for an example: https://github.com/avajs/ava/pull/1947/commits/2b8ba3ac99f7ed9ea6ea5297c915a16bbcf28a17

@ijjk if you're still keen, perhaps you could make a PR for just one of the options we discussed here. Preferably not anything that needs sandboxes and cleanups. If we put that behind a feature flag we can ship it as a new, opt-in feature without being a breaking change. We can then build on that feature, one PR at a time. It'll be a lot easier to review multiple smaller changes than one big one. We'll also be free to change the implementation as we learn more.

If you're burned out by this experience I totally understand, and again I'm sorry it got to this point. Let us know and perhaps somebody else is interested in picking this up.

This PR is still too daunting for me so I'm going to go ahead and close it. But I'd really like to see these new worker modes 😍

PR #2012, opened 2019-01-12
fix t.log in hooks

### LINKED ISSUE #1536: Ava logging with does not work outside a test
## Bug Description

Ava logging with `t.log` only work inside a test function, but not during setup (`before`, `beforeEach`) or teardown (`after.*`) functions.

This means that meaningful setup / teardown data, which is very useful for debugging and reproducing, is lost. This happens both for successful and failed tests, and with and without the `--verbose` flag.

### Test Source

```js
import test from 'ava';

test.before((t) => {
  // This runs before all tests
  t.log('before - 1');
});

test.before((t) => {
  // This runs after the above, but before tests
  t.log('before - 2');
});

test.after('cleanup', (t) => {
  // This runs after all tests
  t.log('after');
});

test.after.always('guaranteed cleanup', (t) => {
  // This will always run, regardless of earlier failures
  t.log('after always');
});

test.beforeEach((t) => {
  // This runs before each test
  t.log('beforeEach');
});

test.afterEach((t) => {
  // This runs after each test
  t.log('afterEach');
});

test.afterEach.always((t) => {
  // This runs after each test and other test hooks, even if they failed
  t.log('afterEachAlways');
});

test('Passing test', (t) => {
  t.log('Log from the passing test');
  t.pass();
});


test('Failing test', (t) => {
  t.log('Log from the failing test');
  t.fail();
});```

Only the log output from the `test` function appear:

```
  ✔ Passing test
    ℹ Log from the passing test
  ✖ Failing test Test failed via `t.fail()`
    ℹ Log from the failing test

  1 test failed [14:00:40]

  Failing test
    ℹ Log from the failing test

  /Users/adam/Personal/tmp/ava-bug-log-in-before-each/run.js:46

   45:     t.log('Log from the failing test');
   46:     t.fail();
   47: });

  Test failed via `t.fail()`
```

### Command-Line Arguments

```bash
ava run.js --verbose
```

## Relevant Links

- [Original SO question for reference](https://stackoverflow.com/q/46395020/51197).
- [Minimal reproduction github example](https://github.com/adamatan/ava-logging-bug).

## Environment

```
Node.js v8.1.2
darwin 16.7.0
0.22.0 // ava
5.4.2 . // npm
```

dflupu:
Fixes #1536

I've looked into grouping per-test hook output with the test output, and while the changes necessary were rather trivial, there were a few downsides such as:

- it would be inconsistent with the per-file hooks that would still not be grouped
- it's rather confusing and unnecessary in practice
- test results would not be printed until all of the after hooks finish


novemberborn:
> I've looked into grouping per-test hook output with the test output, and while the changes necessary were rather trivial, there were a few downsides such as:
> 
> * it would be inconsistent with the per-file hooks that would still not be grouped
> * it's rather confusing and unnecessary in practice
> * test results would not be printed until all of the after hooks finish

Fair enough. I'd really like to revisit the reporters, we could improve this later.

PR #2015, opened 2019-01-13
Add `t.timeout()`

### LINKED ISSUE #1565: Test specific timeout with `t.timeout(ms)`
### Description

I am using AVA (together with `got`) to run E2E tests on my API to make sure that a new version of a service doesn't break something.

I have an API which generates zip files. This action is asynchronous and returns a unique ID. That ID can then be used to poll the status of the zip file. I know for sure that that action shouldn't take longer than 10 seconds and if it does, something is wrong.

I could use the global `timeout` option, but I might want to be able to set a different timeout on different API endpoints.

Feel free to reject the idea though, my current workaround works but maybe other people are also interested in this. Like to hear some opinions.

### Example

```js
import test from 'ava';

test('foo bar', t => {
    t.timeout(5000);

    await delay(6000);
});
```

> `foo bar` timed out after 5000 ms

### Current workaround

This is how I do it today

```ts
export const timeout = (t, ms: number) => {
	setTimeout(() => {
		t.fail('Timeout exceeded');
	}, ms);
};
```

```js
import test from 'ava';
import { timeout } from './fixtures/utils';

test('foo bar', t => {
    timeout(t, 5000);

    await delay(6000);
});
```

dflupu:
Fixes #1565 

As suggested in the issue, the timeout is refreshed whenever an assertion is made.
I'm not sure if the typescript or flow definitions need to be updated.

### DIFF
--- a/docs/07-test-timeouts.md
+++ b/docs/07-test-timeouts.md
@@ -16,9 +16,9 @@ npx ava --timeout=100 # 100 milliseconds
 
 Timeouts can also be set individually for each test. These timeouts are reset each time an assertion is made.
 
-```javascript
+```js
 test('foo', t => {
 	t.timeout(100); // 100 milliseconds
-	// write your assertions here
+	// Write your assertions here
 });
 ```
--- a/docs/07-test-timeouts.md
+++ b/docs/07-test-timeouts.md
@@ -13,3 +13,12 @@ npx ava --timeout=10s # 10 seconds
 npx ava --timeout=2m # 2 minutes
 npx ava --timeout=100 # 100 milliseconds
 ```
+
+Timeouts can also be set individually for each test. These timeouts are reset each time an assertion is made.
+
+```javascript
+test('foo', t => {
+	t.timeout(100); // 100 milliseconds
+	// write your assertions here
+});
+```
--- a/lib/test.js
+++ b/lib/test.js
@@ -266,7 +266,13 @@ class Test {
 	}
 
 	refreshTimeout() {
-		if (this.timeoutTimer) {
+		if (!this.timeoutTimer) {
+			return;
+		}
+
+		if (this.timeoutTimer.refresh) {
+			this.timeoutTimer.refresh();
+		} else {
 			this.timeout(this.timeoutMs);
 		}
 	}
--- a/lib/test.js
+++ b/lib/test.js
@@ -145,6 +145,7 @@ class Test {
 		this.endCallbackFinisher = null;
 		this.finishDueToAttributedError = null;
 		this.finishDueToInactivity = null;
+		this.finishDueToTimeout = null;
 		this.finishing = false;
 		this.pendingAssertionCount = 0;
 		this.pendingThrowsAssertion = null;
--- a/lib/test.js
+++ b/lib/test.js
@@ -53,6 +53,10 @@ function plan(count) {
 	this.plan(count, captureStack(this.plan));
 }
 
+function timeout(ms) {
+	this.timeout(ms);
+}
+
 const testMap = new WeakMap();
 class ExecutionContext {
 	constructor(test) {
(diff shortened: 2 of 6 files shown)


docs/07-test-timeouts.md
```diff
@@ -13,3 +13,12 @@ npx ava --timeout=10s # 10 seconds
...
 ```
+
+Timeouts can also be set individually for each test. These timeouts are reset each time an assertion is made.
+
+```javascript
+test('foo', t => {
+	t.timeout(100); # 100 milliseconds
+	...
+});
+```
```
novemberborn: Could you duplicate this section in https://github.com/avajs/ava/blob/master/docs/02-execution-context.md?

docs/07-test-timeouts.md
```diff
@@ -13,3 +13,12 @@ npx ava --timeout=10s # 10 seconds
 npx ava --timeout=2m # 2 minutes
 npx ava --timeout=100 # 100 milliseconds
 ```
+
+Timeouts can also be set individually for each test. These timeouts are reset each time an assertion is made.
+
+```javascript
+test('foo', t => {
+	t.timeout(100); # 100 milliseconds
```
novemberborn: `// 100 milliseconds`, rather than `# …`

docs/07-test-timeouts.md
```diff
@@ -13,3 +13,12 @@ npx ava --timeout=10s # 10 seconds
...
+	...
```
novemberborn: Perhaps `// Write your assertions here`

lib/test.js
```diff
@@ -66,12 +70,18 @@ class ExecutionContext {
 		boundPlan.skip = () => {};
 
 		Object.defineProperties(this, assertionNames.reduce((props, name) => {
-			props[name] = {value: assertions[name].bind(test)};
+			const refreshTimeoutOnAssertion = assertion => (...args) => {
+				test.refreshTimeout();
+				return assertion.call(test, ...args);
+			};
+			const assertion = refreshTimeoutOnAssertion(assertions[name]);
```
novemberborn: Rather than wrapping the assertions, I think it would be sufficient to add a call to `refreshTimeout()` in `countPassedAssertion()`, `addPendingAssertion()` and `addFailedAssertion()`. The timer should be cleared once again when the pending assertion resolves.

lib/test.js
```diff
@@ -234,6 +244,23 @@ class Test {
 		this.planStack = planStack;
 	}
 
+	timeout(ms) {
+		clearTimeout(this.timeoutTimer);
```
novemberborn: This should probably be a no-op if the test has already finished (or failed).

lib/test.js
```diff
@@ -234,6 +244,23 @@ class Test {
...
+		this.timeoutTimer = nowAndTimers.setTimeout(() => {
+			this.saveFirstError(new Error('Test timeout exceeded'));
+
+			if (this.finishDueToTimeout) {
+				this.finishDueToTimeout();
+			}
+		}, ms);
```
novemberborn: When the test finishes or fails, the timeout should be cleared to ensure it doesn't keep the worker process alive unnecessarily. Though, at the moment the process is exited when all tests have completed, see #1718 for a proposal to change that.

dflupu:
Hi @novemberborn, thanks a lot for the quick feedback! I've tried to address everything you've mentioned. Please look over the updated changes when you find the time.

dflupu:
I'm rather curious why you'd go through the trouble of checking for Timer#refresh. Not that I have anything against it.

novemberborn:
> I'm rather curious why you'd go through the trouble of checking for Timer#refresh. Not that I have anything against it.

It's the modern API and should be more efficient than recreating the timer, closure, etc.

PR #2016, opened 2019-01-13
Change serialization of Error summary due to #1995

### LINKED ISSUE #1995: Errors thrown outside of tests have truncated message
## Description

AVA is truncating multline error messages before displaying the error in the console when that error is thrown outside of a test (e.g. inside of a helper that is being imported). This doesn't seem to happen if the error is thrown in the test itself.

In the minimal example below, where AVA shows the source code lines with a hardcoded message, this may not seem like a big deal. But in practice, I stumbled upon this with some argument validation logic in a helper. The output in the terminal was super confusing and vague because the actual and exepected values were not on the first line (it was a [joi](https://github.com/hapijs/joi) error) and basically all it said was `ValidationError: {`.

### Test Source

```js
// test/hello-world.js
import test from 'ava';

throw new Error('I will be seen\nBut I will not be seen');

test('hello', () => {
    t.true(true);
});
```

```
❯ npx ava test/hello-world.js


  ✖ No tests found in test/hello-world.js

  1 uncaught exception

  Uncaught exception in test/hello-world.js

  /Users/sholladay/Code/noirdoor/chords/test/hello-world.js:3

   2:
   3: throw new Error('I will be seen\nBut I will not be seen');
   4:

  Error: I will be seen
```

### Config

N/A

### Command-Line Arguments

```
npx ava
```

## Environment

```
Node.js v8.14.0
darwin 17.7.0
ava 1.0.0-rc.2
npm 6.4.1
```

CrispusDH:
1. Fixes #1995
2. Add `lines[index].startsWith('    at')` for fixing tests from `test/reports/*.js`
3. Change `skips esm enhancement lines when finding the summary` test in `test/serialize-error.js` for checking my changes
4. Firstly, I change `serialize standard props` test in in `test/serialize-error.js` 
from:
`t.is(serializedError.summary, 'Error: Hello');`
to:
`t.is(serializedError.summary, error.stack)`;
But after adding fix for `test/reports/*.js` I return it to initial 
`t.is(serializedError.summary, 'Error: Hello');` and it also pass

### DIFF
--- a/lib/serialize-error.js
+++ b/lib/serialize-error.js
@@ -107,7 +107,7 @@ function trySerializeError(err, shouldBeautifyStack) {
 		} else {
 			// Skip the source line header inserted by `esm`:
 			// <https://github.com/standard-things/esm/wiki/improved-errors>
-			const start = lines.indexOf(lines.find(line => !/:\d+$/.test(line)));
+			const start = lines.findIndex(line => !/:\d+$/.test(line));
 			retval.summary = '';
 			for (let index = start; index < lines.length; index++) {
 				if (lines[index].startsWith('    at')) {
--- a/lib/serialize-error.js
+++ b/lib/serialize-error.js
@@ -110,7 +110,7 @@ function trySerializeError(err, shouldBeautifyStack) {
 			const start = lines.indexOf(lines.find(line => !/:\d+$/.test(line)));
 			retval.summary = '';
 			for (let index = start; index < lines.length; index++) {
-				retval.summary += index + 1 !== lines.length ? lines[index] + '\n' : lines[index];
+				retval.summary += index + 1 === lines.length ? lines[index] : lines[index] + '\n';
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


CrispusDH:
@novemberborn or @sindresorhus , could you take a look at it? It's first PR in AVA so I'm not sure what are next steps

lib/serialize-error.js
```diff
@@ -107,7 +107,16 @@ function trySerializeError(err, shouldBeautifyStack) {
 		} else {
 			// Skip the source line header inserted by `esm`:
 			// <https://github.com/standard-things/esm/wiki/improved-errors>
-			retval.summary = lines.find(line => !/:\d+$/.test(line));
+			const start = lines.indexOf(lines.find(line => !/:\d+$/.test(line)));
```
novemberborn: You can use `findIndex()` for this.

CrispusDH:
@novemberborn thanks for review, commit with requested change has been done :)

novemberborn:
Thank you @CrispusDH.

PR #2022, opened 2019-01-19
print pending tests on interrupt

### LINKED ISSUE #583: Print pending tests when interrupting AVA, or after a timeout

dflupu:
Fixes #583

### DIFF
--- a/lib/run-status.js
+++ b/lib/run-status.js
@@ -6,6 +6,8 @@ class RunStatus extends Emittery {
 	constructor(files, parallelRuns) {
 		super();
 
+		this.pendingTests = new Map();
+
 		this.stats = {
 			byFile: new Map(),
 			declaredTests: 0,
@@ -43,6 +45,8 @@ class RunStatus extends Emittery {
 			uncaughtExceptions: 0,
 			unhandledRejections: 0
 		});
+
+		this.pendingTests.set(testFile, new Set());
 		worker.onStateChange(data => this.emitStateChange(data));
 	}
 
@@ -55,6 +59,7 @@ class RunStatus extends Emittery {
 			case 'declared-test':
 				stats.declaredTests++;
 				fileStats.declaredTests++;
+				this.addPendingTest(event);
 				break;
 			case 'hook-failed':
 				stats.failedHooks++;
@@ -87,6 +92,7 @@ class RunStatus extends Emittery {
 				fileStats.failedTests++;
 				stats.remainingTests--;
 				fileStats.remainingTests--;
+				this.removePendingTest(event);
 				break;
 			case 'test-passed':
 				if (event.knownFailing) {
@@ -99,9 +105,16 @@ class RunStatus extends Emittery {
 
 				stats.remainingTests--;
 				fileStats.remainingTests--;
+				this.removePendingTest(event);
 				break;
 			case 'timeout':
+				event.pendingTests = this.pendingTests;
+				this.pendingTests = new Map();
 				stats.timeouts++;
+				break;
+			case 'interrupt':
+				event.pendingTests = this.pendingTests;
+				this.pendingTests = new Map();
 				break;
 			case 'uncaught-exception':
 				stats.uncaughtExceptions++;
@@ -149,5 +162,18 @@ class RunStatus extends Emittery {
 
 		return 0;
 	}
+
+	addPendingTest(event) {
+		if (this.pendingTests.has(event.testFile)) {
+			this.pendingTests.get(event.testFile).add(event.title);
+		}
+	}
+
+	removePendingTest(event) {
+		if (this.pendingTests.has(event.testFile)) {
+			this.pendingTests.get(event.testFile).delete(event.title);
+		}
+	}
 }
+
 module.exports = RunStatus;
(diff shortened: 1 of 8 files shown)


dflupu:
Can't get that test working when ran during `npm test`. `tap --no-cov test/integration/assorted.js` works fine. I assume the event loop is being blocked by something else.

lib/run-status.js
```diff
@@ -99,10 +105,16 @@ class RunStatus extends Emittery {
 
 				stats.remainingTests--;
 				fileStats.remainingTests--;
+				this.pendingTests.get(event.testFile).delete(event.title);
 				break;
 			case 'timeout':
+				event.tests = this.pendingTests;
+				this.pendingTests = new Map();
```
novemberborn: I'm not at all opposed to tracking this stuff here, pending a larger overhaul of the reporters. However there may be race conditions where we time out one worker, and receive events from another right before we make it exit. I don't think you can reset `pendingTests` like this without adding guards in the other case statements, and that starts to feel iffy.

That's probably why the logic was in the (verbose) reporters.

`event.tests` is perhaps not the best naming either. Perhaps `activeTests` or `concurrentTests`? But then again it may be best to revert to the existing logic. We can backport to the mini reporter later.

lib/watcher.js
```diff
@@ -244,12 +244,17 @@ class Watcher {
...
-	trackFailures(api) {
+	trackFailures(api, reporter) {
 		api.on('run', plan => {
 			this.pruneFailures(plan.files);
 
 			const currentVector = this.runVector;
 			plan.status.on('stateChange', evt => {
+				if (evt.type === 'interrupt') {
+					reporter.endRun();
+					process.exit(1); // eslint-disable-line unicorn/no-process-exit
```
novemberborn: Why does this explicitly exit (with an exit code), but when AVA runs without the watcher it doesn't?

test/integration/assorted.js
```diff
@@ -10,11 +10,22 @@ const {execCli} = require('../helper/cli');
...
 	execCli(['long-running.js', '-T', '1s'], (err, stdout) => {
 		t.ok(err);
-		t.match(stdout, /Exited because no new tests completed within the last 1000ms of inactivity/);
+		t.match(stdout, /Timed out/);
 		t.end();
 	});
 });
 
+test('interrupt', t => {
+	const proc = execCli(['long-running.js', '-T', '5s'], (err, stdout) => {
```
novemberborn: Perhaps remove the `-T` argument to make it clearer this only deals with `SIGINT`.

test/reporters/verbose.timeoutinmultiplefiles.log
```diff
@@ -5,9 +5,9 @@
   [32m✔[39m a [90m[2m›[22m[39m a passes two
 ---tty-stream-chunk-separator
   [31m[39m
-  [31m✖ Exited because no new tests completed within the last 10000ms of inactivity[39m
+  [31m✖ Timed out while running tests[39m
```
novemberborn: The original copy seems better for the `--timeout` scenario. What made you change it?

lib/watcher.js
dflupu: Good catch. I'll make it exit(1) if ran without the watcher as well.

test/reporters/verbose.timeoutinmultiplefiles.log
dflupu: It seemed a bit long, repetitive and unclear.
Exited - not always true
no new tests - why the `new`?
10000ms - do I have to pull out a calculator if its longer duration? do I actually not know AND care about what the duration was?
of inactivity - repetitive

lib/run-status.js
dflupu: I'll look into it and revert back if I can't find a way to work around the race condition you mentioned. Wanted to avoid duplicating the code in the reporters.

test/integration/assorted.js
dflupu: I've removed this test altogether since it didn't work during `npm test`. For reference, here it was:

```
test('interrupt', t => {
	const proc = execCli(['long-running.js'], (_, stdout) => {
		t.match(stdout, /SIGINT/);
		t.end();
	});

	setTimeout(() => {
		proc.kill('SIGINT');
	}, 2000);
});
```

lib/run-status.js
dflupu: I've fixed the naming and added some checks. The reset shouldn't create problems. It's basically the same implementation as before, except its in RunStatus instead of the reporter. I think that it's more fitting for it to be in RunStatus, and it spares us of some code duplication.

Take a look at the diffs and if you still want it reverted, it's no big deal either way.

dflupu:
@novemberborn Can you take a look at the replies above? I don't know if you get notifications for them so I figured I'd poke you in a separate comment. :)

sindresorhus:
@dflupu You need to submit the PR URL to IssueHunt to claim the bounty: https://issuehunt.io/repos/26820798/issues/583

PR #2027, opened 2019-01-25
Wrote about a common problem in typescript recipe

mikob:
problem and potential solution documented.


qlonik:
Could you provide minimal reproduction of the problem?

My impression was that `allowJs` is used during the transition period when porting the project from javascript to typescript and because of that it is not so common. Is it the case?

novemberborn:
Closing due to inactivity.

PR #1751, opened 2018-03-26
Watcher no longer runs all tests when a snap file is deleted (issue #1511)

mfainshtein2:
This pull request has made the change suggested by [novemberborn](https://github.com/avajs/ava/issues/1511#issuecomment-328351206) to resolve issue #1511. 

**Testing Changes:**
    Starting the watcher with two test files:
    image
    deleting test2.snap:
    image
    deleting test.snap:
    image
    Changing Test2.js
    image

PR #2036, opened 2019-02-08
Delete snapshots files after running tests

### LINKED ISSUE #1511: Deleting a snap file runs all tests

itaisteinherz:
This fixes #1511 by recognizing snapshot (`.snap`) files as source files.

Note that I didn't add an integration test as requested in https://github.com/avajs/ava/pull/1751#pullrequestreview-108522952 since the two assertions I added to the `ava-files.js` unit tests should be enough. If you still think we need an integration test for this, let me know and I'll add one.

// cc @novemberborn

### DIFF
--- a/lib/ava-files.js
+++ b/lib/ava-files.js
@@ -171,7 +171,7 @@ class AvaFiles {
 
 		// Same defaults as used for Chokidar
 		if (!hasPositivePattern) {
-			mixedPatterns = ['package.json', '**/*.js'].concat(mixedPatterns);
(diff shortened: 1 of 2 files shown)


novemberborn:
Thanks @itaisteinherz!

PR #2037, opened 2019-02-08
Update esm package detection

jdalton:
This PR updates ava's `esm` detection to use a simplified symbol value.

### DIFF
--- a/lib/worker/subprocess.js
+++ b/lib/worker/subprocess.js
@@ -111,7 +111,8 @@ ipc.options.then(options => {
 			const required = require(mod);
 
 			try {
-				if (required[Symbol.for('esm\u200D:package')]) {
+				if (required[Symbol.for('esm:package')] ||
+						required[Symbol.for('esm\u200D:package')]) {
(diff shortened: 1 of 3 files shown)


lib/worker/subprocess.js
```diff
@@ -111,7 +111,7 @@ ipc.options.then(options => {
 			const required = require(mod);
 
 			try {
-				if (required[Symbol.for('esm\u200D:package')]) {
+				if (required[Symbol.for('esm:package')]) {
```
novemberborn: I think we need to check for both symbols, to maintain compatibility with projects using older `esm` versions. Can check for the new symbol first of course.

jdalton: Ah gross. I didn't intend to make more work for y'all. Okay. I'll mod it to check box and y'all can drop the legacy one when you all do a major bump.

jdalton: Updated ✨

novemberborn:
Thanks @jdalton!

PR #2042, opened 2019-02-14
fix: fix babel recipe link

kagawagao:
fix babel recipe link in configuration doc


novemberborn:
Nice find, thanks @kagawagao!

Issue #1454, opened 2017-07-14
Convert documentation to a more accessible format


jottr:
Currently the documentation is available in a single README.  
I would like to suggest to convert the documentation to a format which can be consumed by other tools, e.g. [Dash](https://kapeli.com/dash) or [devdocs.io](http://devdocs.io). 
I found that publishing documentation via [readthedocs.io](https://readthedocs.io), which uses sphinx to convert markdown files to the final documentation ([tutorial](https://docs.readthedocs.io/en/latest/getting_started.html)), yields great results. 
Sphinx-generated documentation can be consumed by Dash, and should be easily scrapable for devdocs. 

On a side note, I found the [API documentation](https://github.com/avajs/ava#api) too sparse, some methods are not explained in that section, but must be searched for in other sections (e.g. `test.serial`, `test.cb`, and so on).  

I found it rather difficult, at times even confusing, when trying to look up a certain method.

novemberborn:
I'm definitely in favor of moving the API documentation out of the README, into something that's better organized and more complete.

I'm not sure how great a fit these documentation websites are. We'd have to see if hosting Markdown files on GitHub is sufficient, or whether we should use `devdocs` / `readthedocs` / etc, in lieu of building an AVA website.

@avajs/core?

sindresorhus:
I'm totally 👍 for this, just needs to be a solution with little overhead. I don't want a big build step.

jottr:
There's a bunch of tools specific to API documentation, e.g. [swagger](https://swagger.io/) and [the likes](http://nordicapis.com/ultimate-guide-to-30-api-documentation-solutions/). 
I'd love to help improve the documentation. 
@sindresorhus do you have any preference?

sindresorhus:
>  e.g. swagger and the likes.

Seems like these are for web APIs, not libraries.

sindresorhus:
I'm not really up to date on what people use for API docs these days, but I would prefer something where we could write it manually instead of it parsing JSDoc comments. I like the freedom of hand-crafting docs, as generated ones are always bad. Happy to hear suggestion on what we should use.

jamiebuilds:
Maybe all the documentation could be pulled into the `docs` directory in separate files and published to a very simple markdown-to-html website using GitHub Pages (with Jekyll) or [Netlify](https://www.netlify.com) (with anything).

Netlify is cool cause it will automatically run on PRs and give you a unique URL to preview the website on every change

sindresorhus:
[...]

PR #2051, opened 2019-02-24
Create documentation site with vuepress (#1454)

nklayman:
Adds a documentation/info site, powered by [vuepress](https://vuepress.vuejs.org/). Pages are written in markdown, and then converted to static html. Start dev server with `npm run docs:dev`. Build site with `npm run docs:build` (outputs to `docs/.vuepress/dist`). Versioning is handled by requesting the `releases.json` (hosted by the docs site, located in `docs/.vuepress/public`). It creates a `versions` dropdown with links to `url/v${version}`. Each version of docs would need to be published to the proper path. Hosting may need extra configuration as well, and the [base configuration](https://vuepress.vuejs.org/config/#base) likely needs to be set.


nklayman:
It also has a `version` script that will update the `releases.json` automatically.

novemberborn:
Thanks for making the PR @nklayman. It may take me a little while to get round to reviewing this and considering how we'd publish to https://zeit.co/now.

novemberborn:
This looks rather nice @nklayman!

Does the theme need to be committed? It looks to be the default theme so won't it come with the `vuepress` dependency?

Do we need to nest the Markdown files or can they stay in `docs`? You made a fair amount of changes to the Markdown syntax. Is that for compatibility with Vuepress? The `index.md` files look like duplicates of other content?

Could you remind me how we need to lay out the versions? I'm thinking that if we go with ZEIT's Now, we could have a build process that uses the GitHub API to select the Markdown files for each release, puts them in the right location and then builds a static site that includes the documentation of all versions. Does that sound about right to you?

docs/guide/01-writing-tests.md
```diff
@@ -57,7 +62,7 @@ test('resolves with unicorn', t => {
 AVA comes with built-in support for [async functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function).
 
 ```js
-test(async function (t) {
+test(async function(t) {
```
sindresorhus: Don't do unrelated changes.

docs/guide/01-writing-tests.md
```diff
@@ -73,7 +78,7 @@ test('promises the truth', async t => {
 
 AVA comes with built-in support for [observables](https://github.com/zenparsing/es-observable). If you return an observable from a test, AVA will automatically consume it to completion before ending the test.
 
-*You do not need to use "callback mode" or call `t.end()`.*
+_You do not need to use "callback mode" or call `t.end()`._
```
sindresorhus: Don't do unrelated changes.

nklayman: Whoops, didn't mean to change that.

nklayman:
@novemberborn The reason I had to eject the default theme was to add the version selector dropdown. https://github.com/vuejs/vuepress/issues/100 is a similar issue.

The docs are nested so they fall under different sections for navigation. There is the homepage, the guide section, and the recipes section.

The markdown changes were not intentional. I think prettier changed some things, I have it set to auto-format my code.

The root `index.md` is the regular readme, but without the install instructions. Those were moved to `guide/index.md`. The `recipes/index.md` is just a table of contents.

As for versioning, my original plan was to keep an archive of all the docs versions in `ava.li/vx.x.x`. The [vuepress base config option](https://vuepress.vuejs.org/config/#base) needs to be set to `vx.x.x` for each archive build, however. The navigation bar has a dropdown that lists all versions of ava that link to the proper page. The versions get pulled from the `docs/.vuepress/public/releases.json`. This file is updated automatically by a `version` script. Your option might be easier to deploy as you won't need the `releases.json` or need to set the `base` config option. However, the builds might get really long if you are building multiple versions worth of docs. Unfortunately, [vuepress does not have versioning support](https://github.com/vuejs/vuepress/issues/1018), so this is all kind of a hack. Vuepress v1.x might support it (or support could be added with a plugin), so that might fix this.

novemberborn:
Hi @nklayman, sorry I haven't gotten round to picking this up again.

novemberborn:
Hey @nklayman, I'm so sorry for ignoring this for so very long. I think the time commitment required on my part to review and then ship this turned out to be too much. That's happened with some other AVA PRs as well. I'm trying to encourage a more iterative approach that should result in smaller changes that are easier to ship.

It's probably for the best to close this PR now. Sorry to disappoint.

PR #2052, opened 2019-02-25
Added info for shortcut commands in watch mode (#1525)

KompKK:
For #1525 issue
I've added info printing functionality for shortcut commands

### DIFF
--- a/lib/watcher.js
+++ b/lib/watcher.js
@@ -6,8 +6,8 @@ const chokidar = require('chokidar');
 const flatten = require('arr-flatten');
 const union = require('array-union');
 const uniq = require('array-uniq');
-const AvaFiles = require('./ava-files');
 const chalk = require('chalk');
+const AvaFiles = require('./ava-files');
 
 function rethrowAsync(err) {
 	// Don't swallow exceptions. Note that any
--- a/lib/watcher.js
+++ b/lib/watcher.js
@@ -7,6 +7,7 @@ const flatten = require('arr-flatten');
 const union = require('array-union');
 const uniq = require('array-uniq');
 const AvaFiles = require('./ava-files');
+const chalk = require('chalk');
 
 function rethrowAsync(err) {
 	// Don't swallow exceptions. Note that any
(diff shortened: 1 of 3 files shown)


lib/watcher.js
```diff
@@ -6,6 +6,7 @@ const chokidar = require('chokidar');
 const flatten = require('arr-flatten');
 const union = require('array-union');
 const uniq = require('array-uniq');
+const chalk = require('chalk');
```
novemberborn: You should require chalk from `./chalk`. That way you get a properly configured instance.

test/integration/watcher.js
```diff
@@ -101,7 +103,7 @@ test('watcher does not rerun test files when they write snapshot files', t => {
 				killed = true;
 			}, 500);
 		} else if (passedFirst && !killed) {
-			t.is(buffer.replace(/\s/g, ''), '');
+			t.is(buffer.replace(/\s/g, '').replace(END_MESSAGE, ''), '');
```
novemberborn: It'd be clearer if you replace whitespace from `END_MESSAGE` here rather than at the top, since the string assignment looks so much like the one in `lib/watcher.js`.

lib/watcher.js
```diff
@@ -18,6 +19,7 @@ function rethrowAsync(err) {
 
 const MIN_DEBOUNCE_DELAY = 10;
 const INITIAL_DEBOUNCE_DELAY = 100;
+const END_MESSAGE = chalk.gray('Type `r` and press enter to rerun tests\nType `u` and press enter to update snapshots');
```
novemberborn: I think it'd be good to write another `\n` at the end. That way if the tests re-rerun (because you change files) there is an empty line before the test run separator.

KompKK:
I don't know why node 8 on windows fails..

novemberborn:
> I don't know why node 8 on windows fails.. Have you got any ideas?

Looks like an intermittent failure. I've restarted it.

novemberborn:
Thanks @KompKK!

PR #1902, opened 2018-08-13
Add support for `code` in `expected` t.throws() and t.throwsAsync() t…

okyantoro:
Fixes https://github.com/avajs/ava/issues/1901

PR #2053, opened 2019-02-26
Bug fix: t.throws() does not fail when expected code is a number

qlonik:
A tiny bug with a tiny fix. Title describes it all.

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -215,7 +215,7 @@ function assertExpectations({assertion, actual, expectations, message, prefix, s
 		});
 	}
 
-	if ((typeof expectations.code === 'number' || typeof expectations.code === 'string') && actual.code !== expectations.code) {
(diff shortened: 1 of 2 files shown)


qlonik:
I dont understand why is it that it did not fail in CI on the first commit (looking at detailed results, it passed for node on linux for 11, 10 and other versions). It fails on my machine when I'm running `test/assertions.js` without the fix.

I also dont understand why after the fix is pushed, windows is failing in integration test.

How  are all of those issues are related?

novemberborn:
Wow, nice find! Looks like this never worked since number `code` expectations where introduced in #1902.

Some of the tests can be a little flaky, that's probably what's happened here.

I'm pushing a simplification (there's no need to check `code` since at this point the `expectations` object has already been validated). If CI passes I'll ship this. Thank you @qlonik!

Issue #1808, opened 2018-05-19
Improve the ergonomics of power-assert


chocolateboy:
> Major +1 on promoting power-assert. If only I could do t(result === null) :wink:
>
> -- @novemberborn https://github.com/avajs/ava/pull/663#issuecomment-200040513

The last time I used AVA, I thought power-assert had been disabled by default, as I couldn't see evidence of its helpful diagnostics in the output of various failed assertions. I've only just noticed recently that this is because it's only available via `t.true` (**edit**: and `t.truthy`), which I don't use that often, and which is not a particularly obvious way to unlock it.

Other test frameworks force you to use an ever-growing list of [assertions](http://www.chaijs.com/api/assert/) and [matchers](http://www.chaijs.com/api/bdd/), but power-assert renders almost all of these methods redundant. It would be nice to bring it center stage rather than hiding it behind the facade of its less idiomatic alternatives.

I'd like to suggest two solutions to this:

1. make power-assertions more **explicit** by adding an `assert` function:

```javascript
t.assert(foo >= bar)
```

2. make power-assertions more **convenient** by making `t` itself the `t.assert` function:

```javascript
t(foo >= bar)
```

The former can be achieved by adding `assert` as an alias for <s>`true`</s> `truthy` (or vice versa). The latter can be achieved by making the `t` object a function, which is trivial if it's a plain object and not hard if it's an instance, e.g. via:

* https://github.com/CGamesPlay/node-callable-instance
* https://github.com/shesek/callable-klass
* https://github.com/substack/funstance
* https://github.com/za-creature/callable-object

novemberborn:
Interesting. I do agree that `t.true()` communicates an assertion on a property value, not so much an expression result. The power-assert output is not always expected or welcomed either. I wonder if it may be an improvement to restrict power-assert to one purposefully named assertion, e.g. `t.expr(foo >= bar)`. That restriction may also make it easier for us to integrate the desired behavior more closely with AVA itself.

@avajs/core?

chocolateboy:
@novemberborn I don't think `t.expr` (assert that its argument is an expression?) is clearer or more descriptive than `t.assert`. What other test frameworks use that?

(P.S. I think this is a feature/enhancement request rather than (or as well as) a question :-)

chocolateboy:
> As a general rule we don't alias assertions, so since we already have t.truthy() we wouldn't introduce t.assert().

[...]

PR #2056, opened 2019-03-02
Add power-assert enabled t.assert() assertion; remove from others

eemed:
Initial for #1808 . I still have those 3 failing tests but could you @novemberborn take a look at this first.
I moved these functions away from enhanced assert.
-    t.truthy()
-    t.falsy()
-    t.true()
-    t.false()
-    t.regex()
-    t.notRegex()

Added a new function t.assert() to enhanced assert and copied over the t.truthy() tests.
Then I tried to update the tests that kept failing. I also updated the patterns in babel-preset-transform-test-files.


test/assert.js
```diff
@@ -1505,6 +1505,7 @@ test('.regex() fails if passed a bad value', t => {
 		assertions.regex(42, /foo/);
 	}, {
 		assertion: 'regex',
+		imporoperUsage: true,
```
novemberborn: There's a typo here.

test/fixture/enhanced-assertion-formatting.js
```diff
@@ -4,16 +4,16 @@ const foo = 'foo';
...
 });
 
-test('fails with "instanceof" expression', t => {
-	// eslint-disable-next-line no-new-object
-	t.false(new Object(foo) instanceof Object);
-});
+// test('fails with "instanceof" expression', t => {
+// 	// eslint-disable-next-line no-new-object
+// 	t.false(new Object(foo) instanceof Object);
+// });
```
novemberborn: Let's keep this test case, but then with `t.assert()`.

test/fixture/report/regular/test.js
```diff
@@ -26,8 +26,8 @@ test('formatted', t => {
 });
 
 test('power-assert', t => {
-	const foo = 'bar';
-	t.falsy(foo);
+	const foo = '';
+	t.assert(foo);
```
chocolateboy: To match the original test, shouldn't this be:

```js
t.assert(!foo);
```

eemed:
Updated tests and used .assert() in the test case. PR for babel-preset-transform-files is [here.](https://github.com/avajs/babel-preset-transform-test-files/pull/15)

test/fixture/report/regular/test.js
eemed: Don't really know. Is it important that they match?

novemberborn: I guess @chocolateboy's point is that you also changed the value of `foo`. But, in the end, the test still fails, so it doesn't change too much.

novemberborn:
Looks great @eemed. Would you want to have a stab at updating the documentation?

eemed:
I can take a look. Do I just commit doc updates to this pr aswell?

novemberborn:
> Do I just commit doc updates to this pr aswell?

Yep!

eemed:
I'm not sure if there's more updating to be done. I tried to look through all the files but didn't find anything else.

novemberborn:
@eemed I've pushed some documentation updates, let me know what you think.

Let's hold off merging until we can address https://github.com/avajs/eslint-plugin-ava/issues/218. Perhaps you'd like to tackle that one too? 😄

eemed:
I can take a look at it this weekend.

PR #2062, opened 2019-03-06
Don't report filtered tests as pending

### LINKED ISSUE #2055: Don't report filtered tests as pending
## Description

When tests fail when using a `-m` title matching filter, Ava says that some tests are pending, even if they are not matched. For example:

```
$ ava --concurrency=8 --timeout=2m -m='!remote*' -m='*aws*'
...
✖ Timed out while running tests
 5 tests were pending in [...]build/test/basic.test.js
 ◌ basic › remote aws basic calls { mode: 'queue', childProcess: true }
 ◌ basic › remote google basic calls { mode: 'https', childProcess: false }
 ◌ basic › remote google basic calls { mode: 'https', childProcess: true }
 ◌ basic › remote google basic calls { mode: 'queue', childProcess: false }
 ◌ basic › remote google basic calls { mode: 'queue', childProcess: true }

```

The match criteria is the union of all non-remote tests with all tests matching `*aws*`. The tests matching `remote google` should be filtered out (and indeed are not executed), however they are reported as pending. This is confusing because:

(1) It's hard to tell which test timed out, especially if there are many tests being filtered out.
(2) It makes it seem that the matching filter isn't working even though it is.

Proposed behavior: filter the pending test report before printing, so the output above would be:

```
✖ Timed out while running tests
 1 tests were pending in [...]build/test/basic.test.js
 ◌ basic › remote aws basic calls { mode: 'queue', childProcess: true }
```

This would pinpoint the test that is actually timing out.

### Config

```json
{
 "ava": {
        "files": [
            "build/test/*.test.js"
        ]
    }
}
```

### Command-Line Arguments

```
$ ava --concurrency=8 --timeout=2m -m='!remote*' -m='*aws*'
```

## Environment

```
Node.js v8.14.0
darwin 18.2.0
$ ava --version
1.2.1
$ npm --version
6.4.1
```

vancouverwill:
Fixes #2055 

New test to verify that when match filter is used, pending test report only prints matched pending tests not all pending tests


novemberborn:
Thanks @vancouverwill!

PR #2071, opened 2019-03-15
Add module-resolution example to esm recipe

IgnusG:
Custom module paths are quite common when using webpack or even babel.
It can be helpful to support them using the module-alias package when using esm with ava

### DIFF
--- a/docs/recipes/es-modules.md
+++ b/docs/recipes/es-modules.md
@@ -47,6 +47,30 @@ test('2 + 2 = 4', t => {
 });
 ```
 
+If you want to use module resolution with `esm` relative to a specific directory (such as if you use webpack's resolve.modules) you cant do so by using the `module-alias` package.
+
+First, install [`module-alias`](https://github.com/ilearnio/module-alias):
+
+```
+$ npm install module-alias
+```
+
+Then you need to configure it inside your ava configuration:
+
+**`package.json`:**
+
+```json
+{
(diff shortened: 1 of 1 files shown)


novemberborn:
Thanks for the PR @IgnusG.

This isn't `esm` specific, is it? I'm not sure how much value there is in documenting this in AVA itself if it's just adding it to AVA's `require` option.

IgnusG:
It was just a problem I had when I followed this recipe. 

For this project I used webpack to bundle my code, but I don't use babel. Because of that, I rely on webpack to resolve the import/export paths from my files, but that's obviously missing when ava is running these tests.

When I followed this recipe, using the `esm` extension helped compile the imports/exports via node but it still didn't allow custom module paths - only relative ones.
You can achieve this by using babel and ditching the `esm` extension, or some other magic with webpack integration (pre-compiling sources etc.), but if you're using `esm` and still want module paths - this is the simplest way.

I found that package in a discussion on a different repo - took a while so I thought I'd include it as a remark here :)

Maybe it could be just a side-note instead? Something like
```
If you want to use custom module paths, you can do so with `module-alias`.
```

novemberborn:
I think it could go into a webpack recipe (how to run your front-end code in Node.js without a webpack build) but we don't have such a recipe at the moment.

Besides that I don't think this recipe is the right place. I understand how it's useful, but editorially speaking there should be a good reason for promoting packages.

PR #2077, opened 2019-03-22
feature request: ava configs' factory function promise support 

### LINKED ISSUE #1934: feature request: ava configs' factory function promise support.
## Description

> The config file must have a default export, using ES modules. It can either be a plain object or a factory function which returns a plain object.

vs

> We highly recommend the use of async functions. They make asynchronous code concise and readable, and they implicitly return a promise so you don't have to.

I propose to add support for a promise from a factory function, because in my case i need to look around the environment I am going to run tests for and read some files. I can use synchronous versions of fs module, but it is contrary to node's guidelines and Ava's spirit.

### Test Source

```js
export default async ({ projectDir }) => {
  /* await stuff */
  return { /* … */};
}
```

```
☯ yarn test
yarn run v1.9.4
$ ./bin test

✖ Factory method exported by ava.config.js must not return a promise
```

### Config

N/A, see above.

### Command-Line Arguments

N/A

## Relevant Links

- Sorry, project is not public

## Environment

```
Node.js v8.11.4
darwin 17.7.0
npm 5.6.0
yarn 1.9.4
ava 1.0.0-beta.8
```

a-deeb:
Fixes #1934.


novemberborn:
Hey @a-deeb how are you getting on with this?

novemberborn:
Closing due to inactivity.

Issue #1485, opened 2017-08-08
Interrupting test execution on first exception (through assertion extension API)


jugglinmike:
As of version 0.21.0, Ava's built-in assertions do not influence the execution
of the test body. Specifically, when an assertion is violated, the test
continues to execute.

This behavior was previously reported as a bug (via gh-220). It was considered
"fixed" (via gh-259) with a patch that modified Ava's output.

This is a feature request for a behavior change: forcibly interrupt test
execution when an assertion fails via a runtime exception.

My use case is taking screen shots to assist debugging of integration test
failures. I have been able to react to failing tests programatically through a
combination of the `afterEach` and `afterEach.always` methods. When a test
fails, I would like to capture an image of the rendered application as this
can be very useful in identifying the cause of the failure (especially when the
tests run remotely on a continuous integration server).

Because the test body continues to execute following the failure, by the time
the `afterEach.always` method is invoked, the rendered output may no longer
reflect the state of the application at the moment of failure.

For unit tests, this might be addressed by making test bodies shorter and more
direct. Reducing tests to contain only one meaningful interaction would avoid
the effect described above. Because integration tests have high "set up" costs
and because they are typically structured to model complete usage scenarios,
this is not an appropriate solution for my use case.

Ava supports the usage of a general-purpose assertion library (e.g. as
Node.js's built-in `assert` module), and I've found that because these
libraries operate via JavaScript exceptions, they produce the intended results.
In the short-term, I am considering switching to one of these libraries.
However, Ava's built-in assertions have a number of advantages over generic
alternatives. In addition, restricting the use of Ava's API in my test suite
will be difficult moving forward--even with documentation in place,
contributors may not recognize that certain aspects of the API are considered
"off limits", especially considering that their usage does not directly effect
test correctness.

I haven't been able to think of a use case that would be broken by the change I
am requesting. But if there is such a use case, then this behavior could be
made "opt-in' via a command-line flag.

Thanks for your time, and thanks for the great framework!

novemberborn:
Hi @jugglinmike, thanks for your suggestion. I'll respond more in depth later, but for now, for context, could you share how the integration tests happen within an AVA `test()`?

jugglinmike:
Sure. Here's an example test for a "Todo list" application:

```js
test.afterEach(t => { t.context._passed = true; });
test.afterEach.always(function(t) {
    if (t.context._passed) {
      return;
    }

    return t.context.driver.saveScreenshot(t.title);
  });

[...]

PR #2078, opened 2019-03-22
added Test stats to execution context

pierreh:
#1485 
#1692 

I would like to propose this pull request. I built a framework on top of AVA that allows you to execute individual steps within a Test. However I wanted to have more control in order to stop steps from executing when previous steps have failed and to give debug feedback about which steps have executed and which have not. In order to do this I had to hack Object.bind to get to the Test object, however this patch would expose the necessary information that tests can use to inspect the current state of testing.

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -115,7 +115,7 @@ class ExecutionContext {
 	set context(context) {
 		testMap.get(this).contextRef.set(context);
 	}
-	
+
 	_throwsArgStart(assertion, file, line) {
 		testMap.get(this).trackThrows({assertion, file, line});
 	}
--- a/lib/test.js
+++ b/lib/test.js
@@ -57,6 +57,19 @@ function timeout(ms) {
 	this.timeout(ms);
 }
 
+function stats() {
+	return {
+		assertCount: this.assertCount,
+		assertError: this.assertError,
+		calledEnd: this.calledEnd,
+		duration: this.duration,
+		finishing: this.finishing,
+		pendingAssertionCount: this.pendingAssertionCount,
+		planCount: this.planCount,
+		startedAt: this.startedAt
+	};
+}
+
 const testMap = new WeakMap();
 class ExecutionContext {
 	constructor(test) {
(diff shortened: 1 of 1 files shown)


novemberborn:
Hi @pierreh, thanks for the PR. It's an interesting approach, could you share more about your framework? I want to be careful about how much internal state we expose to the test implementations. 

`t.try()` (#1692) is how we're looking to address use cases like yours. Would that work for you?

(I'm closing this issue for housekeeping purposes, but let's keep the conversation going. Either in this PR or in a new issue.)

pierreh:
Hi @novemberborn, thanks for your reply. I think the `t.try()` solution can work for my use case. Is there any implementation for it yet?
What I have created uses ava-spec to create features and scenario's, but on top of that each scenario is built up from individual steps on order to create a BDD style spec file. The steps are implemented in a separate file and are reusable. The steps are a collection of `given` `when` `then` actions. So for my particular case I would need to run `t.try()` on each step. So the implementation will need to allow me to execute it multiple times during a test. It also integrates supertest/superagent because we are using it to test our API's.

This is what a spec file looks like:
`feature("Log in ›", scenario => {
	scenario("a bureau service user", t => steps(t)
		.step(given.aBureauServiceUser)
		.step(when.theUserLogsIn)
		.step(expect.loginIsSuccessful)
		.step(when.theUserLogsOut)
		.step(expect.userIsLoggedOut));
}`
This is what the results look like (with my patch)
`Log in › a bureau service user
    ℹ step: aBureauServiceUser
    ℹ step: theUserLogsIn
    ℹ failed: loginIsSuccessful
    ℹ skipped: userIsABureauService`

pierreh:
The reason why I like my solution is that you can query the results at any time during testing. Also the `t.try()` solution would have to spawn a nested test context which seems somewhat illogical. The stats don't necessarily need to expose that much, perhaps just `assertCount` and `assertError`

novemberborn:
@pierreh there's an open PR, it's just taking a while to land it: https://github.com/avajs/ava/pull/1947

I like your use case, and I think `t.try()` is a good fit for it especially since it comes with its own DSL. I'd like to try that first before seeing if we need to expose other details within the main test implementation.

pierreh:
Thank you, look forward to it landing!

PR #2080, opened 2019-03-24
Manage assertions through a class

qlonik:
This changes assertions to be properties on the class. Before,
assertions were functions which needed to be bound to the test instance.
This led to the fact that snapshot() assertion accessed property which
only existed on Test - to compare snapshots. Now, the functions which
assertions access are clearly defined as those that need to be
implemented by child class.

This change also updates how ExecutionContext is created, as well as
minor changes to the test/assert.js.

Additionally, if this was to be changed to typescript, properly typing `_skippable`, `_enhanced` and parameters to all assertion functions, we would get type information of all assertions methods on `t` for free. Also adding doc comments to all assertions would give documentation which lives near the source code.

There are also some minor type related changes:
1. In throw assertions, there is no longer arguments.length passed to `validateExpectations()`. Instead validateExpectations checks if second argument null or undefined.
2. In some places there is `isPromise()` being used, however it only checks if .then on that object exists and it is a function. So, in those cases that PromiseLike object is cast to Promise with `.resolve()`.


lib/assert.js
```diff
@@ -69,12 +70,12 @@ function getStack() {
...
 }
 
-function validateExpectations(assertion, expectations, numArgs) { // eslint-disable-line complexity
+function validateExpectations(assertion, expectations) { // eslint-disable-line complexity
 	if (typeof expectations === 'function') {
 		expectations = {instanceOf: expectations};
 	} else if (typeof expectations === 'string' || expectations instanceof RegExp) {
 		expectations = {message: expectations};
-	} else if (numArgs === 1 || expectations === null) {
+	} else if (expectations === null || expectations === undefined) {
```
novemberborn: Checking the argument length was an explicit change to address https://github.com/avajs/ava/issues/1676. We need to retain that behavior.

lib/assert.js
```diff
@@ -229,95 +230,195 @@ function assertExpectations({assertion, actual, expectations, message, prefix, s
...
 }
 
-function wrapAssertions(callbacks) {
-	const {pass, pending, fail} = callbacks;
-	const noop = () => {};
+const noop = () => {};
+class Assertions {
+	/**
+	 * @protected
+	 */
```
novemberborn: I'm really not a fan of these kinds of comments, sorry. No need to repeat the parameters, and the `_` already indicates it's non-public.

lib/enhance-assert.js
```diff
@@ -27,6 +27,7 @@ const formatter = context => {
 		.reverse();
 };
 
+// Todo Should be removed
```
novemberborn: Is this something to address in this PR?

lib/assert.js
```diff
@@ -327,23 +428,25 @@ function wrapAssertions(callbacks) {
...
+			let actual = null;
 			try {
 				retval = fn();
 				if (isPromise(retval)) {
 					try {
-						retval.catch(noop);
+						// In reality isPromise returns if something is PromiseLike
+						// i.e. has a .then() function. Wrapping in Promise.resolve()
+						// casts it to actual promise
+						Promise.resolve(retval).catch(noop);
```
novemberborn: Nice catch. We can get rid of the `try` / `catch` too now.

lib/test.js
```diff
@@ -22,68 +22,39 @@ const captureStack = start => {
...
-		boundPlan.skip = () => {};
-
-		Object.defineProperties(this, assertionNames.reduce((props, name) => {
-			props[name] = {value: assertions[name].bind(test)};
-			props[name].value.skip = skip;
-			return props;
-		}, {
-			log: {value: log.bind(test)},
-			plan: {value: boundPlan},
-			timeout: {value: timeout.bind(test)}
```
novemberborn: Doesn't look like these are still bound?

lib/assert.js
qlonik: :+1: Sorry, was not aware.

lib/assert.js
qlonik: As you mentioned in the main comment, should we try to put this into weak map too?

lib/test.js
qlonik: I'm not sure if they need to be bound to test. These functions now exist as part of the ExecutionContext and are bound to ExecutionContext, similarly to all other assertions. If they need to access the test instance, they can do so via WeakMap. 

Is it acceptable behaviour or would you rather keep those functions bound to the Test instance?

lib/assert.js
qlonik: Also, the reason for the parameters in comment, is that I wanted to document what implementing method should expect as parameters passed to the given function. Is there some better way to document those values?

qlonik:
> I'm reluctant to expose the protected methods on the execution context. Folks will discover them and use them and I'd rather they didn't. Perhaps we can utilize a WeakMap instead?

This is a good idea. How about those protected methods on Assertions class? Since ExecutionContext now extends that class, the private methods starting with _ might also be discovered and used. Should those private methods be hidden via WeakMap too? If we hide them, then it is slightly more difficult to extend the base Assertions class, as we would have to pass those private functions to the constructor of Assertions and children classes cannot simply overwrite those methods.

lib/enhance-assert.js
qlonik: We can remove this function. I was not sure if it is being used anywhere else, where I haven't seen it.

test/assert.js
```diff
@@ -1245,28 +1258,40 @@ test('.snapshot()', t => {
...
+				});
+				this.title = title;
+				this.snapshotInvocationCount = 0;
+			}
+		}(_title);
 	};
 
 	passes(t, () => {
-		const testInstance = setup('passes');
-		assertions.snapshot.call(testInstance, {foo: 'bar'});
```
qlonik: Oops, the `'bar'` was changed by accident. I wonder why didnt test case fail?

lib/enhance-assert.js
novemberborn: No this should be the only place.

lib/test.js
novemberborn: By bound, I mean that this should work:

```js
const { log } = t
someArray.forEach(log)
```

Looks like we don't have test coverage for this but that should keep working.

lib/test.js
```diff
@@ -22,68 +22,55 @@ const captureStack = start => {
...
+		});
+
+		if (args.length > 0) {
+			testMap.get(this).addLog(args.join(' '));
+		}
+	}
+
+	plan(count) {
+		const test = testMap.get(this);
+		test.plan(count, captureStack(test.plan)); // Or captureStack(this.plan) ?
```
novemberborn: I guess it used to do `test.plan`. Though `this.plan` should also be fine.

lib/assert.js
novemberborn: TypeScript, I suppose 😉 I think the code is clear enough without the comments.

novemberborn:
This is pretty close to landing, nice work @qlonik!

lib/test.js
qlonik: Oh, I see.

I was going to add the test which verifies that these functions are properly bound, but then I noticed that this test already exists - https://github.com/avajs/ava/blob/master/test/test.js#L685 
I could add another more explicit test, if you would like?

qlonik: Turns out,
```js
const { log } = t
someArray.forEach(log)
```
does not work

qlonik: Fixed by moving into constructor.

Another solution would be to implement log, plan and timeout as getters (like end and title), because those seem to be binding to instance properly. However that way it seemed weird, I decided to put into constructor.

novemberborn:
🎉 

I'll land https://github.com/avajs/ava/pull/2096 tomorrow and fix the merge conflicts here, and then I'll land this one.

PR #2086, opened 2019-03-29
Simplify macro types

qlonik:
I have noticed that with the types before the proposed change (current types), the following type checks (but it should not). I'm not sure if version of typescript matters here or not, but I'm using 3.3.4000. However, if the title is removed from the test definition, it will fail as it should and even with proper error.
```ts
const hasLength: Macro<[string, number]> = (t, input, expected) => {
	t.is(input.length, expected);
};

test('bar has length 3', [hasLength], 2, 3); // should fail
//                                    ^ Wrong type
test([hasLength], 2, 3);
//                ^ error TS2345: Argument of type '2' is not assignable to parameter of type 'string'.
```

----

The change removes all of the crafty and clever type stuff in favor of simpler solution. Additionally, this change introduces the distinction between Title Macro and Untitled macro. Such distinction helps with the inference of parameter types, when we are passing macro inline without any types specified.

Essentially, in the following example we do not need to explicitly type parameters `t`, `input` and `expected`. Previously, in this case we had to explicitly specify types on parameters of test function.
```ts
test('has length 3', (t, input, expected) => {
	t.is(input.length, expected);
}, 'foo', 3);
```

However, this PR has a flaw that it does not give the user a very useful type error, when it fails. Consider the following case.
```ts
const hasLength: Macro<[string, number]> = (t, input, expected) => {
	t.is(input.length, expected);
};

test('bar has length 3', hasLength, 3, 'bar'); // useless error
//   ^ error TS2345: Argument of type '"bar has length 3"' is not assignable to parameter of type 'OneOrMoreMacros<[Macro<[string, number], {}>, number, string], {}>'.

test(hasLength, 3, 'bar'); // useful error
//              ^ error TS2345: Argument of type '3' is not assignable to parameter of type 'string'.
```
In this case, when title is provided, it is not obvious what the error means. However, if we swap around lines 436 and 439 in the changed file, then the errors swap around - the function call with the title has a useful error message and the one without the title has a useless one! This is exactly the kind of error and useless message that the original change was intended to address. However, it was not very well done.

Maybe the best approach would be to keep types simple and prefer unreadable error over no error at all. Also, if the case where title is provided is more frequently used over the macro without the title case, then we can reorder those two lines, so that there is more readable error for the more used scenario, with the sacrifice that the less used case will receive less readable error.

----

P.S. These types turned out to be very complicated and tedious to work with - ordering changes the error message completely, having "optional" title messes things up for TypeScript. Clearly the best type annotation would be something along the lines of `<T extends any[]>(title?: string, macro: Macro<T, Context>, ...rest: T): void;`, which merges the existence or non-existence of title into one definition, but this is not supported. 

Maybe we should consider asking for help from people familiar with these types of issues.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -413,7 +413,7 @@ export type Macro<Args extends any[], Context = {}> = UntitledMacro<Args, Contex
 	title?: (providedTitle: string | undefined, ...args: Args) => string;
 }
 
-type EitherMacro<Args extends any[], Context> = Macro<Args, Context> | UntitledMacro<Args, Context>;
+export type EitherMacro<Args extends any[], Context> = Macro<Args, Context> | UntitledMacro<Args, Context>;
 
 /** Alias for a single macro, or an array of macros. */
 export type OneOrMoreMacros<Args extends any[], Context> = EitherMacro<Args, Context> | [EitherMacro<Args, Context>, ...EitherMacro<Args, Context>[]];
@@ -425,7 +425,7 @@ export type CbMacro<Args extends any[], Context = {}> = UntitledCbMacro<Args, Co
 	title?: (providedTitle: string | undefined, ...args: Args) => string;
 }
 
-type EitherCbMacro<Args extends any[], Context> = CbMacro<Args, Context> | UntitledCbMacro<Args, Context>;
+export type EitherCbMacro<Args extends any[], Context> = CbMacro<Args, Context> | UntitledCbMacro<Args, Context>;
 
 /** Alias for a single macro, or an array of macros, used for tests & hooks declared with the `.cb` modifier. */
 export type OneOrMoreCbMacros<Args extends any[], Context> = EitherCbMacro<Args, Context> | [EitherCbMacro<Args, Context>, ...EitherCbMacro<Args, Context>[]];
--- a/index.d.ts
+++ b/index.d.ts
@@ -419,8 +419,7 @@ export type OneOrMoreMacros<Args extends any[], Context> = _Macro<Args, Context>
 
 export type UntitledCbMacro<Args extends any[], Context = {}> = (t: CbExecutionContext<Context>, ...args: Args) => ImplementationResult
 /** A reusable test or hook implementation, for tests & hooks declared with the `.cb` modifier. */
-export interface CbMacro<Args extends any[], Context = {}> {
-	(t: CbExecutionContext<Context>, ...args: Args): ImplementationResult;
+export type CbMacro<Args extends any[], Context = {}> = UntitledCbMacro<Args, Context> & {
 	title?: (providedTitle: string | undefined, ...args: Args) => string;
 }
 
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -402,67 +402,41 @@ export type ImplementationResult = PromiseLike<void> | ObservableLike | void;
 export type Implementation<Context = {}> = (t: ExecutionContext<Context>) => ImplementationResult;
 export type CbImplementation<Context = {}> = (t: CbExecutionContext<Context>) => ImplementationResult;
 
+export type UntitledMacro<Args extends any[], Context = {}> = (t: ExecutionContext<Context>, ...args: Args) => ImplementationResult;
```
novemberborn: Should repeat the `/** A reusable test or hook implementation. */` above this line.

index.d.ts
```diff
@@ -402,67 +402,41 @@ export type ImplementationResult = PromiseLike<void> | ObservableLike | void;
...
-
+export type Macro<Args extends any[], Context = {}> = UntitledMacro<Args, Context> & {
 	/**
 	 * Implement this function to generate a test (or hook) title whenever this macro is used. `providedTitle` contains
 	 * the title provided when the test or hook was declared. Also receives the remaining test arguments.
 	 */
 	title?: (providedTitle: string | undefined, ...args: Args) => string;
 }
 
+type _Macro<Args extends any[], Context> = Macro<Args, Context> | UntitledMacro<Args, Context>;
```
novemberborn: How about `EitherMacro`?

index.d.ts
```diff
@@ -402,67 +402,41 @@ export type ImplementationResult = PromiseLike<void> | ObservableLike | void;
...
+
 /** Alias for a single macro, or an array of macros. */
-export type OneOrMoreMacros<Args extends any[], Context> = Macro<Args, Context> | [Macro<Args, Context>, ...Macro<Args, Context>[]];
+export type OneOrMoreMacros<Args extends any[], Context> = _Macro<Args, Context> | [_Macro<Args, Context>, ..._Macro<Args, Context>[]];
 
+export type UntitledCbMacro<Args extends any[], Context = {}> = (t: CbExecutionContext<Context>, ...args: Args) => ImplementationResult
```
novemberborn: Here too let's duplicate the description from `CbMacro`.

index.d.ts
```diff
@@ -402,67 +402,41 @@ export type ImplementationResult = PromiseLike<void> | ObservableLike | void;
...
-	TitleOrMacro extends OneOrMoreMacros<any, Context> ? InferArgs<TitleOrMacro>[0] :
-	never
-
-export type TitleOrCbMacro<Context> = string | OneOrMoreCbMacros<any, Context>
-
-export type CbMacroOrFirstArg<TitleOrMacro, Context> =
-	TitleOrMacro extends string ? OneOrMoreCbMacros<any, Context> :
-	TitleOrMacro extends OneOrMoreCbMacros<any, Context> ? InferArgs<TitleOrMacro>[0] :
-	never
+type _CbMacro<Args extends any[], Context> = CbMacro<Args, Context> | UntitledCbMacro<Args, Context>;
```
novemberborn: `EitherCbMacro`? 😄

PR #2093, opened 2019-04-17
Update TypeScript recipe to allow ESM

NicoCevallos:
Added `esm` module to require parameter to avoid `SyntaxError: Unexpected identifier` error when using ECMAScript module loader(`import` or `export`).

### DIFF
--- a/docs/recipes/typescript.md
+++ b/docs/recipes/typescript.md
@@ -8,7 +8,7 @@ This guide assumes you've already set up TypeScript for your project. Note that 
 
 ## Configuring AVA to compile TypeScript files on the fly
 
-You can configure AVA to recognize TypeScript files. Then, with `ts-node` installed, you can compile them on the fly.
+You can configure AVA to recognize TypeScript files. Then, with `ts-node` installed, you can compile them on the fly. Also you will need to install `esm` to add support for `import` & `export`.
 
 **`package.json`:**
 
@@ -20,7 +20,8 @@ You can configure AVA to recognize TypeScript files. Then, with `ts-node` instal
 			"ts"
 		],
 		"require": [
-			"ts-node/register"
+			"ts-node/register",
+			"esm"
 		]
 	}
 }
--- a/docs/recipes/typescript.md
+++ b/docs/recipes/typescript.md
@@ -8,7 +8,7 @@ This guide assumes you've already set up TypeScript for your project. Note that 
 
 ## Configuring AVA to compile TypeScript files on the fly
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
I don't think this is correct. TypeScript already supports ESM out of the box.

nicocevallos:
Hi @sindresorhus sorry for my vague explanation.
The problem with `import` is when you use `"module": "esnext` in the `tsconfig.json` file.
Also I think it would have a warning because there is an issue with esm@3.2.21+, so needs to install the version 3.2.20 only.

I just updated the PR mentioning this 2 things.

nicocevallos:
Also I created a little repo to test it.
https://github.com/nicocevallos/ava-ts_needs_esm
If you remove `esm` from the `ava.config.js` file, you will see the error

novemberborn:
Heya @NicoCevallos, I understand the problem you're running into, but presumably you have the same issue when you try to run the compiled code? Which I imagine you solve using `esm`. I'm not sure whether this needs to be in the recipe.

NicoCevallos:
Hi @novemberborn you are right, I need to so the same running my code. But the difference is that what I added to the recipe is like a reminder that you need to add `esm` in the `nyc` config file.

novemberborn:
OK so maybe we can have a line that says "if you're relying on `esm` when running your code, don't forget to use it with AVA" and then link to https://github.com/avajs/ava/blob/master/docs/recipes/es-modules.md?

NicoCevallos:
That's ok. But the link doesn't mention that you do need to add esm in the NYC configuration.
What do you think to relate this link too?
https://github.com/istanbuljs/nyc/blob/master/README.md#require-additional-modules

novemberborn:
> But the link doesn't mention that you do need to add esm in the NYC configuration.

But this is AVA, not nyc. Happy to include *that* in the code coverage recipe.

NicoCevallos:
Sorry I mixed what I faced and have to resolve: add `esm` for both. But in summary, if I need to use `esm` to run the code, I'll need to do the same for `AVA` and `nyc`.

If you agree to mention this like a reminder, I could update the PR adding that reminder in the TypeScript and nyc recipes. I guess it is not necessary for using AVA with JS, right?

Hope it clarify the goal.

novemberborn:
> If you agree to mention this like a reminder, I could update the PR adding that reminder in the TypeScript and nyc recipes

Yup sounds great.

> I guess it is not necessary for using AVA with JS, right?

No… unless of course you use `esm` in your JS project 😄

kutyel:
This was just what I needed to make my library work (https://github.com/kutyel/linq.ts/commit/05eb294690188249613ab74d3a0803f696fa6d6f), please, make it happen 🙏

novemberborn:
Closing due to inactivity.

PR #2095, opened 2019-04-19
Move api.js into lib/

novemberborn:
Originally the goal may have been for this to be accessible
programmatically, but it's unclear whether we'd ever end up supporting
that. For now moving it into the lib/ directory better communicates it's
internal.


sindresorhus:
It's technically not breaking as it's an undocumented implementation detail.

novemberborn:
> It's technically not breaking as it's an undocumented implementation detail.

Yea I know, but hey. I'm trying to land some breaking changes this weekend since we can drop Node.js 6 anyway.

PR #2100, opened 2019-04-21
Fix sample test in TypeScript recipe

StoneCypher:
The getting started example contains a minor defect.  Ava requires a label, but the example given does not have one.

```
test(async t => {
	t.is(await fn(), 'foo');
});
```

Result:

```
$ npx ava ./src/test/testset.ts

  Uncaught exception in src\test\testset.ts

  (censored)\src\test\testset.ts:12

   11:
   12: test(async t => {
   13:   t.is(await fn(), 'foo');

  TypeError: Tests must have a title

  Object.<anonymous> (src/test/testset.ts:12:5)
  Module.m._compile (node_modules/ts-node/src/index.ts:439:23)
  Object.require.extensions.(anonymous function) [as .ts] (node_modules/ts-node/src/index.ts:442:12)
```

Proposed replacement:

```
test('receive foo asynchronously', async t => {
	t.is(await fn(), 'foo');
});
```

### DIFF
--- a/docs/recipes/typescript.md
+++ b/docs/recipes/typescript.md
@@ -49,7 +49,7 @@ Create a `test.ts` file.
 ```ts
 import test from 'ava';
 
-const fn = () => Promise.resolve('foo');
+const fn = () => 'foo';
 
 test('fn() returns foo', t => {
 	t.is(fn(), 'foo');
--- a/docs/recipes/typescript.md
+++ b/docs/recipes/typescript.md
@@ -51,7 +51,7 @@ import test from 'ava';
 
 const fn = async () => Promise.resolve('foo');
 
-test(async t => {
+test('receive foo asynchronously', async t => {
 	t.is(await fn(), 'foo');
 });
 ```
--- a/docs/recipes/typescript.md
+++ b/docs/recipes/typescript.md
@@ -49,10 +49,10 @@ Create a `test.ts` file.
 ```ts
 import test from 'ava';
 
-const fn = async () => Promise.resolve('foo');
+const fn = () => Promise.resolve('foo');
 
-test('receive foo asynchronously', async t => {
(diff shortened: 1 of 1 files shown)


docs/recipes/typescript.md
```diff
@@ -51,7 +51,7 @@ import test from 'ava';
 
 const fn = async () => Promise.resolve('foo');
 
-test(async t => {
+test('receive foo asynchronously', async t => {
```
novemberborn: Perhaps change this to `'fn() returns foo'`, and then make `fn` and the test synchronous. I'm not sure why this `fn` was ever asynchronous but `async () => Promise.resolve()` is rather odd.

StoneCypher: Okay, is now

```ts
import test from 'ava';

const fn = () => Promise.resolve('foo');

test('fn() returns foo', t => {
	t.is(fn(), 'foo');
});
```

StoneCypher: > I think the example may be better if fn() is synchronous, what do you think?

I actually think maybe it should have both.  `async` is still pretty hard for a lot of devs, and it might be nice for novices looking for help to get it here.  Novices have a hard time locating things, and in my opinion benefit from extra examples peppered everywhere

But also I don't care very much, I was just fixing a bug 😀

StoneCypher: I'll do the flow recipe in a different PR

StoneCypher: specifically [here](https://github.com/avajs/ava/compare/master...StoneCypher:patch-2)

StoneCypher: um apparently i forgot to make the actual pr

[fixed](https://github.com/avajs/ava/pull/2104)

novemberborn:
Thanks @StoneCypher!

Issue #1264, opened 2017-02-15
Change in source file doesn't trigger test run in watch mode when started with test selection


hoschi:
run `ava -w src/widgets/**/*.test.js` and AVA starts in watch mode and should rerun depending tests when changing a source file (e.g. `src/widgets/MyWidgets.js`). 

This only works on Mac and not on Linux/Windows because of [a chockidar bug](https://github.com/paulmillr/chokidar/issues/561). In short the bug here is that the pattern which says which tests should be run is more special as default pattern for source files or configured in package.json (e.g. source: 'src/**/*.js').

Before version 0.18.0 users could add additional sources by the `--sources` flag:
`ava -w -S src/widgets/**/*.js src/widgets/**/*.test.js`
If you run into that problem you can probably downgrade AVA to 0.17.0 and run with that solution.

Complete discussion on gitter: https://gitter.im/avajs/ava?at=58a405adf045df0a224cfb18

novemberborn:
The default `src/**/*.js` pattern is ignored by Chokidar because of the `src/widgets/**/*.test.js` pattern. This is fundamentally a Chokidar bug, so if anybody can [help out](https://github.com/paulmillr/chokidar/issues/561) that'd be great.

In the meantime I suppose we could clean up the patterns we pass to Chokidar and remove more specific ones if they're already covered by a generic pattern. I'm not sure if tools for that already exist, @sindresorhus? This may be easier than fixing Chokidar though.

sindresorhus:
> I'm not sure if tools for that already exist, @sindresorhus?

Not that I'm aware of.

PR #2103, opened 2019-04-22
Improve file and helper selection

### LINKED ISSUE #736: Harmonize test file & source pattern handling

### LINKED ISSUE #909: Only exclude `helpers` directory when inside a test directory

### LINKED ISSUE #1173: ava incorrectly ignores test files when the project itself is contained in node_modules directory
## Description

I have two projects:

* https://github.com/gajus/isomorphic-webpack
* https://github.com/gajus/isomorphic-webpack-demo

Due to the nature of the projects (`process.pwd()` aware dependency resolution), it is not possible to link `isomorphic-webpack-demo` inside `isomorphic-webpack`. Therefore, I need to do:

```bash
git clone https://github.com/gajus/isomorphic-webpack
git clone https://github.com/gajus/isomorphic-webpack-demo
(cd ./isomorphic-webpack && npm install)
(cd ./isomorphic-webpack-demo && npm install)
rm -fr ./isomorphic-webpack-demo/node_modules/isomorphic-webpack
cp ./isomorphic-webpack ./isomorphic-webpack-demo/node_modules
```

The problem happens when I try to run `ava` inside of `isomorphic-webpack`, i.e.

```bash
cd ./isomorphic-webpack-demo/node_modules/isomorphic-webpack
npm run test
```

I am getting error:

```
  1 exception

  ✖ Couldn't find any files to test
```

My guess (without looking at the source code), is that `ava` detects that `./isomorphic-webpack-demo/node_modules/isomorphic-webpack/test` path includes `node_modules` and therefore ignores this path.

Expected behaviour: ava should detect the immediate `./test` directory and run the tests.

## Environment

```
ava --version
npm --version
Node.js v7.3.0
darwin 16.3.0
0.17.0
3.10.10
```

### LINKED ISSUE #1228: Crashes while looking for helper files in huge project subdirectories

### LINKED ISSUE #1288: Ava slows when there are many non-run files

### LINKED ISSUE #1418: Too slow in large projects
## Description

I am using XO and Ava in a large project. Both packages are too slow to be used. I never actually tried to let them finish. I usually give up after 10+ minutes.
I first thought this might be related to globby: https://github.com/sindresorhus/globby/issues/43

As mentioned in that issue I created a test source that reproduces this issue.
There is only one test file in the test-project.
The glob targets a folder that exists only once.

I will open a separate issue in XO if needed.

### Test Source

* https://github.com/pixelass/ava-xo-test
* https://github.com/pixelass/ava-xo-test/blob/master/one/a/b/c/d/e/f/g/h/i/j/this-is-it/tests/one.js

Keeps doing this thing: 🤣

```
⠧
```

If using `--verbose` nothing happens

### Config

https://github.com/pixelass/ava-xo-test/blob/master/package.json#L41-L59

```json
{
  "ava": {
    "files": [
      "**/this-is-it/**/tests/**/*.js"
    ],
    "source": [
      "**/this-is-it/**/*.js"
    ],
    "presets": [
      "@ava/stage-4",
      "@ava/transform-test-files"
    ],
    "failFast": true,
    "tap": false,
    "require": [
      "babel-register",
      "ignore-styles"
    ],
    "babel": "inherit"
  }
}
```

### Command-Line Arguments

https://github.com/pixelass/ava-xo-test/blob/master/package.json#L10

```json
{
"scripts": {
    "nyc:ava": "nyc ava",
    "ava": "ava",
    "test": "npm run nyc:ava",
  },
}
```

```
ava
```

## Relevant Links

**Minimal reproduction:**  
https://github.com/pixelass/ava-xo-test/

## Environment

```
OS X 10.12.5 (16F73)  
Node 6.11.0  
```
Requested log:

```
Platform: darwin 16.6.0
Ava: 0.19.1
node: 6.11.0
npm: 3.10.10
```

### LINKED ISSUE #1501: Ava does not ignore nested node modules directory
## Description
**Edit:** I found what is causing the issue. Ava does not ignore `node_modules` directories that are nested but the README seems to state it does:
> Directories named fixtures, helpers and node_modules are always ignored.

The project structure is a monorepo with `modules/<module-name>/node_modules` directories, to make Ava ignore them appending a glob is needed: `ava test/*.js !modules/*/node_modules`.

*Original description for people searching at ENOSPC:*
> On Linux running Ava in a directory with more than the [inotify limit](https://github.com/guard/listen/wiki/Increasing-the-amount-of-inotify-watchers#the-technical-details) causes Ava to stop with an unhandled error event `ENOSPC`. It looks like it tries to watch files in `node_modules` because the error comes from a different file in the node modules directory.

```bash
Error: watch <random-file-in-node-modules> ENOSPC
    at _errnoException (util.js:1041:11)
    at FSWatcher.start (fs.js:1382:19)
    at Object.fs.watch (fs.js:1408:11)
    at createFsWatchInstance (/home/siilwyn/myproject/node_modules/chokidar/lib/nodefs-handler.js:37:15)
    at setFsWatchListener (/home/siilwyn/myproject/node_modules/chokidar/lib/nodefs-handler.js:80:15)
    at FSWatcher.NodeFsHandler._watchWithNodeFs (/home/siilwyn/myproject/node_modules/chokidar/lib/nodefs-handler.js:228:14)
    at FSWatcher.NodeFsHandler._handleFile (/home/siilwyn/myproject/node_modules/chokidar/lib/nodefs-handler.js:255:21)
    at FSWatcher.<anonymous> (/home/siilwyn/myproject/node_modules/chokidar/lib/nodefs-handler.js:473:21)
    at FSReqWrap.oncomplete (fs.js:153:5)
```

### Command-Line Arguments

```
ava test.js --watch
```

## Environment

```
Node.js v8.4.0
linux 4.4.0-93-generic
```

### LINKED ISSUE #1540: Expose files dirs seperately
Currently apart from passing in my own files list there doesn't seem to be any way to set the helper's directory apart from overriding the whole `files` array. I'd like to suggest adding a way to change each of the arrays that makes up the`files` option.

This would include `defaultExcludePatterns`, `defaultIncludePatterns ` and `defaultHelperPatterns `. My idea would be to use an object for files which has the fields`excludes`, `includes` and `helpers`.  
https://github.com/avajs/ava/blob/87eef846392d04b727b57981a16f7e8429a04195/lib/ava-files.js#L76-L95

The reason I'm suggesting this is I want to use `__helpers__` for my helper directory as I currently have a `helpers` directory for a test with my helpers. To add that I'd need to replace the whole files array instead of just doing something like `{ files: { helpers: ['**/__helpers'] } }`.

### LINKED ISSUE #1793: Scan for test helpers causes error when non-accessible folders present
## Description

When there is a directory in your project that is unaccessible by the user runnig the tests, ava runs into an error during startup and exits.

In my case this is a volume created by the postgresql docker image and it resets the directory permissions on every start.

The problem is that the defaultHelperPatterns are set in a way that tries to read all directories and cannot be changed afaict.

```
glob error { Error: EACCES: permission denied, scandir '/home/maritz/{redacted}/db-volume/test'
  errno: -13,
  code: 'EACCES',
  syscall: 'scandir',
  path: '/home/maritz/{redacted}/db-volume/test' }
/home/maritz/{redacted}/node_modules/ava/lib/cli.js:206
					throw err;
					^

Error: EACCES: permission denied, scandir '/home/maritz/{redacted}/db-volume/__tests__/helpers'
    at handlePaths (/home/maritz/{redacted}node_modules/ava/lib/ava-files.js:14:18)
    at AvaFiles.findTestHelpers (/home/maritz/{redacted}/node_modules/ava/lib/ava-files.js:140:10)
    at Api._precompileHelpers (/home/maritz/{redacted}node_modules/ava/api.js:140:5)
    at _setupPrecompiler.then (/home/maritz/{redacted}/node_modules/ava/api.js:167:21)
    at <anonymous>

```

### Making it work locally

When going manually into https://github.com/avajs/ava/blob/a84c96f5da36635cbc5fb942e0c26607ad00b570/lib/ava-files.js#L90 and changing the defaultHelperPatterns to only scan my src folder it works fine.

```javascript
const defaultHelperPatterns = () => [
  'src/**/__tests__/helpers/**/*.js',
  'src/**/__tests__/**/_*.js',
  'src/**/test/helpers/**/*.js',
  'src/**/test/**/_*.js',
];
```

## Fix options?

I see two ways to properly fix this (aside from removing inaccessible directories):

1. Add an option to change the defaultHelperPattern (related to https://github.com/avajs/ava/issues/1418#issuecomment-327771126)
or
2. Catch EACCES errors and print out a warning and continue

### LINKED ISSUE #2085: Ava rerunning all tests when changing single test file, when using .ts files.
## Description

Ava rerunning all tests when changing single test file, when using `.ts` files.

```
ava:watcher Detected change of path/to/testFile.ts +0ms
ava:watcher Sources remain that cannot be traced to specific tests: [ 'path/to/testFile.ts' ] +103ms
ava:watcher Rerunning all tests +1ms
```

### Config

```json
{
  "ava": {
    "compileEnhancements": false,
    "extensions": [
      "ts"
    ],
    "require": [
      "ts-node/register"
    ],
    "files": [
      "path/to/testFile.ts",
    ],
    "verbose": true
  }
}
```

### Command-Line Arguments

```
DEBUG=ava:watcher npm test -- --watch --verbose
```

## The cause (I believe)

`node_modules/ava/lib/watcher.js:370` filters `dirtyTests` using the `isTest` function (`node_modules/ava/lib/ava-files.js:208`) which requires files to have a `.js` extension to resolve.

## Environment

```
Node.js v11.4.0
darwin 18.2.0
ava: 1.4.0
npm: 6.4.1
```

novemberborn:
I've run out of time today and can't finish this PR:

* [x] Update documentation
* [x] Figure out why reporter integration tests changed
* [ ] Check behavior on Windows, especially with watch mode
* [x] Ensure files starting with an underscore are recognized as source files, not changed test files (watch mode)
* [ ] Test on some actual projects
* [x] ~Support `helpers` globs (see #1319)~ filed https://github.com/avajs/ava/issues/2105
* [x] Reference the open issues this solves
* [x] Does this solve #1264?
* [x] Does this solve #1173?
* [x] Does this solve #2085?
* [x] ~Ensure our linter plugin recognizes the new syntax~ filed https://github.com/avajs/eslint-plugin-ava/issues/238
* [x] ~"Global" excludes which ensures globs are excluded by the watcher too~ filed https://github.com/avajs/ava/issues/2106

Here's a quick summary of changes. I can expand on rationale later:

* Require globs to match files, not directories
* `files` and `sources` must be an array of one or more patterns, not a string
* Only accept files via the CLI, not globs or directories
* Upgrade globby
* At this stage, only test files starting with `_` are treated as helpers.
  Helper globs should be made configurable.
* It's no longer possible to override an exclusion pattern. We may
  reintroduce this at a later stage

Fixes #1228
Fixes #909 
Fixes #1418 
Fixes #1793 
Fixes #1501
Fixes #736
Fixes #1540
Fixes #1288
Fixes #1173
Fixes #2085


novemberborn:
> I can expand on rationale later

I'm thinking it's actually quite confusing that `files` may match directories. It's easier to understand you need to "specify glob patterns that match test files". Similarly, it's easier to understand that the CLI arguments must be actual file paths rather than glob patterns.

Any helper matched as a test file will still be treated as a helper. For now that's just files starting with an underscore. But before we can ship this as a 2.0 we need to make it possible to specify glob patterns for helper files. These then can be tuned to user's project setup. Any test file that also matches the helper patterns will be considered a helper, not a test file.

If you disable the Babel pipeline, AVA won't need to *find* helper files, just exclude test files that are helpers.

With this PR, test files are still found based on the configured extensions. It just takes a lot less code.

You can specify solely exclusion patterns, which will be applied together with the default patterns.

Overriding the default exclusion patterns is no longer possible. If this becomes a problem we can introduce a setting to customize the default exclusion patterns.

My plan is to get this working and then ship it as a 2.0 beta.

novemberborn:
With regards to our ESLint plugin, my idea's to expose `ava/eslint-helper` which can provide the necessary functionality, without needing to expose the full AVA configuration to the plugin, or re-implement functionality there.

sindresorhus:
> But before we can ship this as a 2.0 we need to make it possible to specify glob patterns for helper files.

I was hoping we could drop the concept of helpers (which users generally find surprising and awkward) in favor just transpiling everything.

novemberborn:
> I was hoping for v2 we could drop the concept of helpers (which users generally find surprising and awkward) in favor just transpiling everything.

That's not going to happen in the next say two weeks.

Even then, presumably we need the ability to compile tests differently from sources, and this helper work will still be necessary.

Hopefully with these changes helpers *are* easier to understand though. It'll be test files that start with an underscore. Everything else will be opt-in.

sindresorhus:
> Does this solve #1264?

Does it?

novemberborn:
> Does it?

No, it's a Chokidar limitation.

PR #2108, opened 2019-04-29
cache error

janat08:

### DIFF
--- a/docs/recipes/debugging-with-vscode.md
+++ b/docs/recipes/debugging-with-vscode.md
@@ -4,6 +4,8 @@ Translations: [Français](https://github.com/avajs/ava-docs/blob/master/fr_FR/do
 
 ## Setup
 
+You first need to cache the tests, so run the test file outside debugging run beforehand.
+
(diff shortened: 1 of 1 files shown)


sindresorhus:
Can you elaborate? I don't think this update is correct. What kind of error did you get?

janat08:
It's somehow gone from the terminal, but it was referring to some hashed file with map in it with cache being somewhere in dir possibly. I was following vscode recipe on ubuntu:
```
		{
			"type": "node",
			"request": "launch",
			"name": "Run AVA test",
			"program": "${workspaceFolder}/node_modules/ava/profile.js",
			"args": ["${file}",
				  "--serial"],
			"skipFiles": ["<node_internals>/**/*.js"]
		}
```

janat08:
image

novemberborn:
It's hard to say what caused that error, but I don't think it's something that can easily be worked around.

(I'm closing this PR for housekeeping purposes, but let's keep the conversation going.)

janat08:
the workaround is to run tests before doing debugging.

PR #2111, opened 2019-05-02
Update invalid require path in babel recipe

LukasHechenberger:
When using a separate file for `@babel/register` a relative path must be required.

### DIFF
--- a/docs/recipes/babel.md
+++ b/docs/recipes/babel.md
@@ -190,7 +190,7 @@ require('@babel/register')({
 });
 ```
 
-Now instead of requiring `@babel/register`, require `test/_register` instead.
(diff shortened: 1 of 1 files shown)


LukasHechenberger:
I guess this one's just a typo in the docs, but it's really annoying to find, ran into it twice...

I don't get where I should post the patch file, so I'll attach it:

[0001-Update-invalid-require-path-in-babel-recipe.patch.txt](https://github.com/avajs/ava/files/3137191/0001-Update-invalid-require-path-in-babel-recipe.patch.txt)

> (renamed to '.txt', as GitHub doesn't allow .patch attachments)

novemberborn:
Thanks @LukasHechenberger!

PR #2117, opened 2019-05-05
[WIP] Bind JSON builtin functions

### LINKED ISSUE #2081: Bind JSON built-ins to allow stubbing
**See conclusion here: https://github.com/avajs/ava/issues/2081#issuecomment-489430603**

---

## Description

Running a simple test where I'm stubbing the `JSON.parse()` function to return a specific output. In case a failure occurs, ava outputs a "Test failure: Could not serialize error" and then hangs. Same hang appears even if a test passes, but I don't call `sinon.restore()`.

### Test Source

```js
import test from 'ava'
import * as sinon from 'sinon'

let sandbox: sinon.SinonSandbox

test.beforeEach(() => {
    sandbox = sinon.createSandbox()
})

test.afterEach(() => {
    sandbox.restore()
})

test('getServerPort', (t) => {
    let parse = sandbox.stub(JSON, 'parse')
    parse.returns({})
    t.is(3,4)
})
```

```
▶ npm run test
> ava build/tests/lib/*.js --verbose

  ✖ getServerPort Test failure: Could not serialize error
^C
  ✖ Exiting due to SIGINT

  1 test failed

  getServerPort


  Test failure: Could not serialize error

  Error: Test failure: Could not serialize error
      at serializeError (/home/mihai/workspace/github/sherry/node_modules/ava/lib/serialize-error.js:143:23)
      at runSingle.then.result (/home/mihai/workspace/github/sherry/node_modules/ava/lib/runner.js:337:11)
      at process._tickCallback (internal/process/next_tick.js:68:7)
```

### Config
Nothing configured for ava. Just transpiling some TS files and then running ava for the js files

### Command-Line Arguments
```
  "scripts": {
    "test": "ava build/tests/lib/*.js --verbose",
    "prepare": "tsc"
  }
```

## Environment

```
Node.js v10.15.1
Ubuntu 18.04
Kernel: linux 4.18.0-16-generic
ava v1.4.0
npm v6.4.1
```

mihai-dinu:
Fixes #2081

There are still some things to do, but most of the functionality is already in the PR.

- [x] Moved `now-and-timers.js` to `bound-builtins.js`
   - [x] Export `timers` and `Date.now` using a `timers` object
   - [x] Export `JSON.parse` and `JSON.stringify` using a `json` object
   - [x] Updated all modules that required `now-and-timers` to use the new module
   - [x] Updated all `JSON.parse` and `JSON.stringify` references to use the `json` bound functions (except for the tests - those still use the built in `JSON` functions) 
- [ ] Add documentation about what built in functions can be stubbed
- [ ] ? Add tests for the `bound-builtins.js` module


mihai-dinu:
I would like to test that stubbing `JSON.parse()` doesn't affect any of `ava's` functionality. Is it possible to use a local instance of `ava` in another project somehow? Or what's the usual procedure for this?

novemberborn:
> I would like to test that stubbing `JSON.parse()` doesn't affect any of `ava's` functionality. Is it possible to use a local instance of `ava` in another project somehow? Or what's the usual procedure for this?

See this (rather timely) message on Spectrum: https://spectrum.chat/ava/general/testing-ava-code-changes~88e21813-56b7-4fc2-852d-06919c905a0e

mihai-dinu:
It looks like I'm going to need some help from someone to point me in the right direction with this PR. After binding the JSON parse and stringify functions, I've tried running a simple test with my local changes to `ava`, but the test hangs if there's an assertion failure.
I've tried digging into the problem a bit more by following the source code and adding some debug logs, but I ended up running around in circles trying to understand how everything is tied together. 

This is the test that I've tried to run against this PR branch:
```javascript
import test from 'ava'
import sinon from 'sinon'

let sandbox

test.beforeEach(() => {
    sandbox = sinon.createSandbox()
})

test.afterEach(() => {
    sandbox.restore()
})

test('test with json parse stub', (t) => {
    let parse = sandbox.stub(JSON, 'parse')
    parse.returns({})
    t.is(3,4, 'Hello')
})
```

mihai-dinu:
Did some more digging, started to understand the code base a bit better as well. 

One hypothesis on why I'm seeing this "hanging" problem is that stubbing any of the `JSON` functions will also affect modules that `ava` is importing and for those there is no way of telling them to use the bound function from the new `bound-builtins.js` module.
So when `JSON.parse()` is called somewhere in the `emittery` module (just a hypothetical example), that function will return what the stub was set to return instead of the correct object. 

I'm not too sure this is the case, since this should have happened for the built in timers as well (right?). 

So, @novemberborn, @sindresorhus, any comments regarding the above hypothesis?

mihai-dinu:
First off, sorry if I'm spamming this PR with comments.

Secondly, I've spent some more time on the problem and isolated where the subprocess hang happens. It's in the [`ipc.flush()` function](https://github.com/avajs/ava/blob/master/lib/worker/ipc.js#L43) - which, to be completely honest, I don't really understand what it does. Commenting it out in the [`subprocess.exit()` function](https://github.com/avajs/ava/blob/master/lib/worker/subprocess.js#L23) lets the process terminate.

The second problem is that the [above test](https://github.com/avajs/ava/pull/2117#issuecomment-490274432) does not fail because of an assertion error, but because of a serialization error: 
```
  ✖ test with json parse stub Test failure: Could not serialize error

  1 test failed

  test with json parse stub


  Test failure: Could not serialize error

  Error: Test failure: Could not serialize error
      at serializeError (/home/mihai/workspace/github/ava/lib/serialize-error.js:144:23)
      at Runner.runTest (/home/mihai/workspace/github/ava/lib/runner.js:337:11)
      at process._tickCallback (internal/process/next_tick.js:68:7)
```
If I print the actual error in the catch block from the [`serializeError` function](https://github.com/avajs/ava/blob/master/lib/serialize-error.js#L131), it prints out as:
```
  ✖ Internal error when running test.js
  Error: "version" is a required argument.
      at Object.getArg (/home/mihai/workspace/github/ava/node_modules/source-map-support/node_modules/source-map/lib/util.js:24:11)
      at new BasicSourceMapConsumer (/home/mihai/workspace/github/ava/node_modules/source-map-support/node_modules/source-map/lib/source-map-consumer.js:294:22)
      at new SourceMapConsumer (/home/mihai/workspace/github/ava/node_modules/source-map-support/node_modules/source-map/lib/source-map-consumer.js:22:7)
      at mapSourcePosition (/home/mihai/workspace/github/ava/node_modules/source-map-support/source-map-support.js:190:14)
      at wrapCallSite (/home/mihai/workspace/github/ava/node_modules/source-map-support/source-map-support.js:358:20)
      at /home/mihai/workspace/github/ava/node_modules/source-map-support/source-map-support.js:399:26
      at Array.map (<anonymous>)
      at Function.prepareStackTrace (/home/mihai/workspace/github/ava/node_modules/source-map-support/source-map-support.js:398:30)
      at formatError (internal/util/inspect.js:817:16)
      at formatRaw (internal/util/inspect.js:616:14)
```

It looks like the source map module uses `JSON.parse` in parsing the source map (https://github.com/mozilla/source-map/blob/master/lib/util.js#L161) and I think this might be the culprit since stubbing `JSON.parse` will not return the correct values for the source map. I need to read more into how source maps work and when they are generated, but that will be left for another comment :)

novemberborn:
> One hypothesis on why I'm seeing this "hanging" problem is that stubbing any of the `JSON` functions will also affect modules that `ava` is importing and for those there is no way of telling them to use the bound function from the new `bound-builtins.js` module.

Yup.

> I'm not too sure this is the case, since this should have happened for the built in timers as well (right?).

It does.

> It looks like the source map module uses `JSON.parse` in parsing the source map ([mozilla/source-map:lib/util.js@`master`#L161](https://github.com/mozilla/source-map/blob/master/lib/util.js#L161) [](/mozilla/source-map/blob/HEAD@{2019-05-09T20:41:57Z}/lib/util.js#L161)) and I think this might be the culprit since stubbing `JSON.parse` will not return the correct values for the source map.

Aye I can see that. I suppose you could make your stubs calculate a stack trace and see if it's called from a Node module and then perform normally…?

I'm not sure whether it's worth continuing with this PR given this obstacle?

mihai-dinu:
> Aye I can see that. I suppose you could make your stubs calculate a stack trace and see if it's called from a Node module and then perform normally…?

I'm not really sure how to do that. I will need some more details.

Regarding the hang, I've attached a debugger and tracked it down even further and indeed it happens in the [`ipc.flush() function`](https://github.com/avajs/ava/blob/master/lib/worker/ipc.js#L43), more specifically in the `await emitter.once('pong')`. When flushing, the subprocess sends a `ping` event and expects a `pong` response. The `pong` response is sent, but the message received for [`process.on('message')`](https://github.com/avajs/ava/blob/master/lib/worker/ipc.js#L5) is json parsed  (https://nodejs.org/api/process.html#process_event_message) and instead of the `pong` reply, the subprocess receives the stubbed `JSON.parse` return value.

> I'm not sure whether it's worth continuing with this PR given this obstacle?

This does look increasingly more difficult. To be honest, I wouldn't mind spending the time to fix this somehow, but the problem that I'm having is that I don't really know how at this point. 

If we don't fix this, I don't think it's such a bit deal - for my specific case at least. I will have to modify my code to not stub `JSON.parse` anymore, but that's doable.

novemberborn:
> > Aye I can see that. I suppose you could make your stubs calculate a stack trace and see if it's called from a Node module and then perform normally…?
> 
> I'm not really sure how to do that. I will need some more details.

Yea I'm just thinking out loud:

```js
const { parse } = JSON
JSON.parse = (...args) => {
  const { stack } = new Error()
  if (stack.split('\n')[0].includes('node_modules')) return parse.apply(this, args)

  return stub()
}
```

But I don't know if that'll work.

> Regarding the hang, I've attached a debugger and tracked it down even further and indeed it happens in the [`ipc.flush() function`](https://github.com/avajs/ava/blob/master/lib/worker/ipc.js#L43) [](/avajs/ava/blob/HEAD@{2019-05-12T15:33:36Z}/lib/worker/ipc.js#L43), more specifically in the `await emitter.once('pong')`. When flushing, the subprocess sends a `ping` event and expects a `pong` response. The `pong` response is sent, but the message received for [`process.on('message')`](https://github.com/avajs/ava/blob/master/lib/worker/ipc.js#L5) [](/avajs/ava/blob/HEAD@{2019-05-12T15:33:36Z}/lib/worker/ipc.js#L5) is json parsed ([nodejs.org/api/process.html#process_event_message](https://nodejs.org/api/process.html#process_event_message)) and instead of the `pong` reply, the subprocess receives the stubbed `JSON.parse` return value.

Yea that makes sense. I'm reluctant to handle that in AVA, there'll just be something else that ends up accessing `JSON.parse()`.

mihai-dinu:
> Yea that makes sense. I'm reluctant to handle that in AVA, there'll just be something else that ends up accessing JSON.parse(). 

I understand. In that case, I guess we can close this PR as won't fix / won't do. Does it make sense to add any doc mentioning that stubbing JSON functions (or any built-in functions I guess) is not something that you should do?

novemberborn:
> > Yea that makes sense. I'm reluctant to handle that in AVA, there'll just be something else that ends up accessing JSON.parse().
> 
> I understand. In that case, I guess we can close this PR as won't fix / won't do.

I wonder how Jest deals with this though. Perhaps they expose a different `JSON` object to the required modules? That's a much bigger change however, and one I'm not comfortable with either.

> Does it make sense to add any doc mentioning that stubbing JSON functions (or any built-in functions I guess) is not something that you should do?

Yes that'd be great. Perhaps you could add a section to https://github.com/avajs/ava/blob/master/docs/08-common-pitfalls.md?

sindresorhus:
> I wonder how Jest deals with this though. Perhaps they expose a different JSON object to the required modules? That's a much bigger change however, and one I'm not comfortable with either.

I think they run tests in a separate [`vm`](https://nodejs.org/api/vm.html) context, so they can inject the globals themselves.

PR #2121, opened 2019-05-11
Make the object printing depth configurable

grnch:
When printing objects or exceptions, Ava restricts the depth of the printed objects to 3 levels. Some libraries that like to throw exceptions with deeply nested data in them (e.g. Axios) would benefit from increasing the printing depth when debugging failing tests, but unfortunately this depth is hardcoded in Ava and it's not possible to change it by the user.

I would propose instead to tie this depth to [`util.inspect.defaultOptions.depth`](https://nodejs.org/api/util.html#util_util_inspect_defaultoptions), which is a setting that NodeJS uses to control the depth of printing for `console.log()` and similar. Most NodeJS developers who have had to debug deeply nested objects by printing them with `console.log()` are already familiar with this setting, so it would come natural to use the same setting to control Ava's printing depth.

There are other ways this could be implemented, of course, for example by exposing this option as a command line argument or through some Ava-specific environment variable, but I think tying it to NodeJS's printing depth is the most practical solution. This way a NodeJS developer could kill two birds with one stone, by setting both the NodeJS and Ava printing depth in one go, rather than hunting for Ava-specific docs separately.

### DIFF
--- a/lib/concordance-options.js
+++ b/lib/concordance-options.js
@@ -1,4 +1,5 @@
 'use strict';
+const util = require('util');
 const ansiStyles = require('ansi-styles');
 const stripAnsi = require('strip-ansi');
 const cloneDeepWith = require('lodash.clonedeepwith');
@@ -124,6 +125,6 @@ const plainTheme = cloneDeepWith(colorTheme, value => {
 });
 
 const theme = chalk.enabled ? colorTheme : plainTheme;
-exports.default = {maxDepth: 3, plugins, theme};
+exports.default = {maxDepth: util.inspect.defaultOptions.depth, plugins, theme};
 exports.diff = {maxDepth: 1, plugins, theme};
 exports.snapshotManager = {plugins, theme: plainTheme};
--- a/lib/concordance-options.js
+++ b/lib/concordance-options.js
@@ -125,6 +125,15 @@ const plainTheme = cloneDeepWith(colorTheme, value => {
 });
 
 const theme = chalk.enabled ? colorTheme : plainTheme;
-exports.default = {maxDepth: util.inspect.defaultOptions.depth, plugins, theme};
+
+exports.default = {
+	// Use Node's object inspection depth, clamped to a minimum of 3
+	get maxDepth() {
+		return Math.max(3, util.inspect.defaultOptions.depth);
+	},
+	plugins,
+	theme
+};
+
 exports.diff = {maxDepth: 1, plugins, theme};
 exports.snapshotManager = {plugins, theme: plainTheme};
--- a/docs/06-configuration.md
+++ b/docs/06-configuration.md
@@ -107,3 +107,20 @@ export default ({projectDir}) => {
 ```
 
 Note that the final configuration must not be a promise.
+
+## Object printing depth
+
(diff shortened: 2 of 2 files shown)


lib/concordance-options.js
```diff
@@ -124,6 +125,6 @@ const plainTheme = cloneDeepWith(colorTheme, value => {
 });
 
 const theme = chalk.enabled ? colorTheme : plainTheme;
-exports.default = {maxDepth: 3, plugins, theme};
+exports.default = {maxDepth: util.inspect.defaultOptions.depth, plugins, theme};
```
CrispusDH: What is default value for this variable? I think you should update also documentation.

grnch: The existing documentation never mentions what is the current depth (`git grep -i depth docs` returns empty), so I am not sure where I should update this information. Any suggestions where I could include it?

novemberborn: The default is `2`, but it was `20` for some of the Node.js 11 releases.

I don't think we should go lower than 3, but we could use `Math.min(3, util.inspect.defaultOptions.depth)`. More recent Node.js 11 releases seem to be using `2` again so that's OK — we don't need to support early versions of Node.js 11.

grnch:
Good point, @novemberborn!

I had been changing the depth before Ava is loaded so that's why it worked for me, but using a getter like you suggested is clearly superior because then it works even after Ava is loaded.

I have also clamped the depth to a minimum of 3, as you suggested.

Thanks!

P.S.
In case you're wondering how I got the depth to change before Ava is loaded, when the first thing the Ava CLI does is `require('ava')`:

On my dev machine I have this env var permanently set:

```
export NODE_OPTIONS="-r ~/.noderc"
```

This allows me to put whatever Node initialization code I want in `~/.noderc` and it always gets loaded by the Node process before anything else, regardless of how the `node` executable is started.

grnch:
Is there anything  I should be doing about the Windows build failure in Travis? It seems like a random failure not related to the changes in this PR.

I could change the time stamp on the commit and force push it again, to trigger another Travis build.

novemberborn:
> You may be wondering how I was changing the depth before Ava is loaded, when the first thing the Ava CLI does is `require('ava')`.
> 
> On my dev machine I have this env var permanently set:
> 
> ```
> export NODE_OPTIONS="-r $HOME/.noderc"
> ```
> 
> This allows me to put whatever Node initialization code I want in `~/.noderc` and it always gets loaded by the Node process before anything else, regardless of how the `node` executable is started.

Oh that's a neat trick!



> Is there anything I should be doing about the Windows build failure in Travis? It seems like a random failure not related to the changes in this PR.

Yea they happen. No worries.

novemberborn:
@grnch this looks good. But we should document it, too. Perhaps as another section in https://github.com/avajs/ava/blob/master/docs/06-configuration.md? What do you reckon?

grnch:
Sure thing @novemberborn.

I'll write something up when I'm near a real computer again.

grnch:
Added some docs, please let me know if wording or location needs to be changed.

lib/concordance-options.js
grnch: I believe all comments in this thread have been addressed (added docs, enforced min depth of 3, added getter).

novemberborn:
Thanks @grnch!

PR #2126, opened 2019-05-16
Define environment variables to be injected in the test file processes

### LINKED ISSUE #2112: Add ability to set environment variables for the process in the AVA config
[Issuehunt badges](https://issuehunt.io/r/avajs/ava/issues/2112)

I'd like to see a config property called `environmentVariables` that lets the user set environment variables that are injected into the test file processes. I have often needed to set some test-only environment variables, but it's not easy to do this in a cross-platform matter.

> There is a $90.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/avajs/ava/issues/2112).

alexisfontaine:
Fixes #2112

This PR adds an `environmentVariables` option to the configuration, as the issue suggested.

It enables specifying environment variables for the different processes running the test files. Note that:
- `AVA_PATH` is a reserved environment variable, setting it would have no effect
- with the exception of `NODE_PATH`, the configuration overrides `process.env`

This option should replace the need for requiring [`dotenv/config`](https://www.npmjs.com/package/dotenv) or some custom JavaScript file that loads environment variables.


alexisfontaine:
@sindresorhus Did I properly understand what you requested?

If so, should we also add some checks on the content of `environmentVariables` and/or normalize it.  [The latest Node.js documentation for `process.env`](https://nodejs.org/dist/latest-v12.x/docs/api/process.html#process_process_env) mentions:

> Assigning a property on `process.env` will implicitly convert the value to a string. **This behavior is deprecated**. Future versions of Node.js may throw an error when the value is not a string, number, or boolean.

docs/06-configuration.md
```diff
@@ -54,6 +57,7 @@ Arguments passed to the CLI will always take precedence over the CLI options con
 - `cache`: cache compiled test and helper files under `node_modules/.cache/ava`. If `false`, files are cached in a temporary directory instead
 - `failFast`: stop running further tests once a test fails
 - `failWithoutAssertions`: if `false`, does not fail a test if it doesn't run [assertions](./03-assertions.md)
+- `environmentVariables`: specifies default environment variables to be injected in the processes that run the test files
```
novemberborn: Perhaps "to be made available to the tests"?

lib/fork.js
```diff
@@ -56,7 +56,7 @@ module.exports = (file, opts, execArgv) => {
 	const subprocess = childProcess.fork(workerPath, args, {
 		cwd: opts.projectDir,
 		silent: true,
-		env,
+		env: {...opts.environmentVariables, ...env},
```
sindresorhus: The `environmentVariables` option should override the env. So this needs to be reversed. This should also be documented in the readme.

alexisfontaine: That was my initial thought. Then I realized that the `env` variable also contains `process.env` from the main process which makes it impossible to overwrite the configuration from the command line.

novemberborn: Let's look at how `env` is built up: https://github.com/avajs/ava/blob/f26634b1a82f1edc92d75935c833c6a8cad7cb93/lib/fork.js#L12:L24

* We set a default `NODE_ENV` variable, though it is overridden by `process.env`
* We rewrite `NODE_PATH`
* We set `AVA_PATH`

I don't think `AVA_PATH` and `NODE_PATH` should be overridden by `environmentVariables`. But `NODE_ENV` should. All that said I think `environmentVariables` should override `process.env`. Which means we need to break this down further:

```
{NODE_ENV: 'test', ...process.env, opts.environmentVariables, NODE_PATH, AVA_PATH}
```

alexisfontaine: Alright, I guess someone could still write the following to be able to override the configuration:

```javascript
environmentVariables: {
    MY_ENVIRONMENT_VARIABLE: process.env.MY_ENVIRONMENT_VARIABLE || 'some default value'
}
```

Or perhaps `--environment-variable, -e` could be exposed as a CLI option in the future.

lib/fork.js
```diff
@@ -9,19 +9,15 @@ if (fs.realpathSync(__filename) !== __filename) {
...
-		.join(path.delimiter);
-}
-
 // In case the test file imports a different AVA install,
 // the presence of this variable allows it to require this one instead
-env.AVA_PATH = path.resolve(__dirname, '..');
+const AVA_PATH = path.resolve(__dirname, '..');
+
+// Ensure NODE_PATH paths are absolute
+const NODE_PATH = process.env.NODE_PATH && process.env.NODE_PATH
```
novemberborn: This might be better with a `let NODE_PATH` and `if (process.env.NODE_PATH)`. Currently this will resolve to `'false'`.

lib/environment-variables.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+function validateEnvironmentVariables(environmentVariables) {
+	if (!environmentVariables) {
+		return {};
+	}
+
+	for (const key of Object.keys(environmentVariables)) {
```
novemberborn: Use `Object.values()` instead.

docs/06-configuration.md
```diff
@@ -54,6 +57,7 @@ Arguments passed to the CLI will always take precedence over the CLI options con
 - `cache`: cache compiled test and helper files under `node_modules/.cache/ava`. If `false`, files are cached in a temporary directory instead
 - `failFast`: stop running further tests once a test fails
 - `failWithoutAssertions`: if `false`, does not fail a test if it doesn't run [assertions](./03-assertions.md)
+- `environmentVariables`: specifies environment variables to be made available to the tests. The environment variables defined here overrides the ones from `process.env`
```
novemberborn: `override` rather than `overrides`.

PR #2128, opened 2019-05-16
Fix the `SkipInterface` in the TypeScript definition

alexisfontaine:
This fixes a copy/paste mistake in the TS definition preventing us from using the `skip` method with macros and without a title.


novemberborn:
Nice catch, thanks @alexisfontaine!

PR #2130, opened 2019-05-18
Store an error object in AssertionError rather than a stack trace

### LINKED ISSUE #1399: Explore storing an error object, rather than directly capturing a stack trace

mihai-dinu:
Fixes #1399.

My idea was to always try to pass a new `Error` object in the AssertionError instead of a `stack` string. It might be a bit more memory inefficient, but it would allow for a more consistent behavior overall.
I then expose the `savedError`'s stack when `serializeError` is called.
One thing that I'm not too sure about is the `Error.stackTraceLimit` of the new Error object. I'm not sure if it takes effect when calling `new Error()` or at the point of getting the stack from the `savedError`. I would assume the first case applies. If that's the case, I can do the `Error.stackTraceLimit` "dance" before creating the new Error. 

The changes need to be unit tested and I will do that as well in this PR, but I first wanted someone to take a quick look over the changes and let me know if I should continue down this path or not.

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -53,26 +53,14 @@ class AssertionError extends Error {
 		// Reserved for power-assert statements
 		this.statements = [];
 
-		if (opts.stack) {
-			this.stack = opts.stack;
-		} else {
-			const limitBefore = Error.stackTraceLimit;
-			Error.stackTraceLimit = Infinity;
-			Error.captureStackTrace(this);
-			Error.stackTraceLimit = limitBefore;
-		}
+        if (opts.savedError && isError(opts.savedError)) {
+            this.savedError = opts.savedError;
+        } else {
+            this.savedError = new Error();
+        }
 	}
 }
 exports.AssertionError = AssertionError;
-
-function getStack() {
-	const limitBefore = Error.stackTraceLimit;
-	Error.stackTraceLimit = Infinity;
-	const obj = {};
-	Error.captureStackTrace(obj, getStack);
-	Error.stackTraceLimit = limitBefore;
-	return obj.stack;
-}
 
 function validateExpectations(assertion, expectations, numArgs) { // eslint-disable-line complexity
 	if (typeof expectations === 'function') {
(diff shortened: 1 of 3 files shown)


novemberborn:
Looks good @mihai-dinu. I assume you've seen the discussion in https://github.com/avajs/ava/pull/2008, which was the previous attempt at resolving this?

mihai-dinu:
Yes, I've seen the comments in #2008. Did I miss anything there? I will start updating the tests and fixing the linting issues. 

Btw, I was thinking of opening another PR for Prettier support since I don't really like changing my IDE's tabs to actual tabs instead of spaces and having to put semicolon at the end of each line. Plus, you get consistent code formatting all over the project.

mihai-dinu:
This PR is ready for review. Not sure what unit test I could add for the changes since they're mainly covered by the existing ones.

novemberborn:
> Btw, I was thinking of opening another PR for Prettier support since I don't really like changing my IDE's tabs to actual tabs instead of spaces and having to put semicolon at the end of each line. Plus, you get consistent code formatting all over the project.

We're good with [XO](https://github.com/xojs/xo) 😄

lib/assert.js
```diff
@@ -53,25 +53,21 @@ class AssertionError extends Error {
 		// Reserved for power-assert statements
 		this.statements = [];
 
-		if (opts.stack) {
-			this.stack = opts.stack;
+		if (opts.savedError && isError(opts.savedError)) {
```
novemberborn: Can't we assume `savedError` is an error?

lib/serialize-error.js
```diff
@@ -51,6 +51,10 @@ function buildSource(source) {
 }
 
 function trySerializeError(err, shouldBeautifyStack) {
+	if (err.savedError) {
+		err.stack = err.savedError.stack;
```
novemberborn: Don't modify `err`. Instead do something like:

```js
let stack = err.savedError ? err.savedError.stack : err.stack
if (shouldBeautifyStack) {
  stack = beautifyStack(stack)
}
```

lib/assert.js
mihai-dinu: You're right. We don't pass in anything else other than an error as savedError. Will change it

Issue #2132, opened 2019-05-19
Power assert does not work with argument destructuring


trbrc:
## Description

Ava supports destructuring of the various assert methods of the `t` argument. However, using `assert` this way appears to break the `power-assert` integration.

#### This works

```js
import test from 'ava';

test('foo equals bar', t => {
  t.assert('foo' === 'bar');
});
```

The output includes the expression:

```
  1 test failed

  test › foo equals bar

  [...]/test.js:4

   3: test('foo equals bar', t => {
   4:   t.assert('foo' === 'bar');
   5: });

  Value is not truthy:

  false

  "foo" === "bar"
  => false
```

#### This breaks

```js
import test from 'ava';

test('foo equals bar', ({assert}) => {
  assert('foo' === 'bar');
});
```

The output only includes the values expression was evaluated to:

```
  1 test failed

  test › foo equals bar

  [...]/test.js:4

   3: test('foo equals bar', ({assert}) => {
   4:   assert('foo' === 'bar');
   5: });

  Value is not truthy:

  false
```

### Config

```js
// ava.config.js
export default {
  require: ['esm']
};
```

## Environment

```
Node.js v12.2.0
darwin 18.0.0
npx ava --version 1.4.1
npm --version 6.9.0
```

trbrc:
I see now that it only detects exactly `t.assert` due to the pattern defined in https://github.com/avajs/babel-preset-transform-test-files/blob/master/espower-patterns.json

This makes me think I can probably change it by customizing AVA’s Babel settings. I will see if I can figure it out. Might this be a good thing to document in the [Babel docs](https://github.com/avajs/ava/blob/master/docs/recipes/babel.md)?

trbrc:
I've modified my `ava.config.js` like this. It turns off the `@ava` preset and installs Power assert with its [default options](https://github.com/power-assert-js/babel-plugin-espower#options), which includes the `assert(value, [message])` pattern.

```js
export default {
  require: ['esm'],
  babel: {
    testOptions: {
      presets: [
        ['@ava/babel-preset-transform-test-files', false]
      ],
      plugins: [
        ['babel-plugin-espower']
      ]
    }
  }
};

```

Is there anything in this configuration that will be incompatible with AVA?

Issue #1255, opened 2017-02-11
Feature Request: Prebound assertion functions


ryardley:
It would be useful and functional to be able to destructure test functions such as `pass`, `fail`, `deepEqual`, `it` etc.  from the test object argument passed into the test function. Currently they are not bound to the test context but could probably be provided pre-bound.

```js
test(({ deepEqual }) => {
    deepEqual([1, 2], [1, 2]);
});
```

sindresorhus:
I don't really see how that is any more readable than:

```js
test(t => {
    t.deepEqual([1, 2], [1, 2]);
});
```

It actually get's very verbose when you use multiple assertions, which is common:

```js
test(({ deepEqual, is, throws }) => {
    deepEqual([1, 2], [1, 2]);
    is(1, 1);
    throws(() => foo());
});
```

Let's see what the other team members think.

Maybe you can elaborate on why you want it this way.

ryardley:
The Javascript community is moving towards a more functional style and being able to use methods functionally without depending on an object context is becoming quite common. 

I can certainly see this as a low priority change but coming to the API for the first time I think it was something I expected to be able to do but was surprised I could not and I figure it would probably not be a huge change.

Honestly probably the main reason you want to do this is that it works with tape:

```
// tape
test('timing test', ({plan, equal}) => {
    plan(2);
    equal(typeof Date.now, 'function');
    const start = Date.now();

    setTimeout(function () {
      equal(Date.now() - start, 100);
    }, 100);
});

```

novemberborn:
> The Javascript community is moving towards a more functional style and being able to use methods functionally without depending on an object context is becoming quite common.

Could you give an example of a test you would write with a functional style? I can think of this scenario:

```js
someArray.forEach(t.true)
```

But that only seems useful for assertions that have no expectations, e.g. not `t.deepEqual()`. It also wouldn't work, since the array index will be passed as the assertion message, which is (will soon be) an error. You'd have to do:

```js
someArray.forEach(value => t.true(value)
```

At which point I don't see the usefulness in binding the assertions to the `t` object.

Currently we also need the `t.` for power-assert.

ryardley:
You can get around that by using lodash's `spread` function or ramda's `apply`. A use case might be some table testing like so:

```javascript
// eg. using ramda
test('test all the things', ({ deepEquals }) => {
  thingsToTest
    .map(({ input, expected }) => [ myfunc(input), expected ])
    .map(apply(deepEquals));
});
```

[...]

PR #2133, opened 2019-05-20
Docs: Custom `babel-plugin-espower` patterns

trbrc:
Suggestion for documentation of solution to #2132 

Related discussion in #1255 suggests that I'm not the only one looking for how to do this, so might as well add it to the documentation.

### DIFF
--- a/docs/recipes/babel.md
+++ b/docs/recipes/babel.md
@@ -101,6 +101,31 @@ Note that this *does not* stop AVA from compiling your test files using Babel.
 
 If you want, you can disable the preset in your project's Babel configuration.
 
+## Change Power Assert's patterns
+
+If you prefer to name your `t` argument something else, or access `assert` with destructuring assignment, you need to tell [`power-assert`](https://github.com/power-assert-js/power-assert) which patterns to look for.
+
+**`package.json`:**
+
+```json
+{
+	"ava": {
+		"babel": {
+			"testOptions": {
+				"plugins": [
+					["babel-plugin-espower", {
+						"patterns": [
+							"assert(value, [message])",
+							"test.assert(value, [message])"
+						]
+					}]
+				]
+			}
+		}
+	}
+}
+```
+
 ## Preserve ES module syntax
 
 By default AVA's stage-4 preset will convert ES module syntax to CommonJS. This can be disabled.


docs/recipes/babel.md
```diff
@@ -101,6 +101,31 @@ Note that this *does not* stop AVA from compiling your test files using Babel.
...
+```json
+{
+	"ava": {
+		"babel": {
+			"testOptions": {
+				"plugins": [
+					["babel-plugin-espower", {
+						"patterns": [
+							"assert(value, [message])",
+							"test.assert(value, [message])"
```
novemberborn: This would be `t.assert`.

novemberborn:
I wonder if, now that only `t.assert()` has Power Assert applied to it, we can't transform `assert` out of the box. I'd rather not document these approaches because it both implies they're supported *and* that they're necessary.

docs/recipes/babel.md
trbrc: My intention was to show how to deal with an argument name other than `t`, and I thought `test` ought to be the most clear example. Perhaps the example should include `t.assert` as well.

trbrc:
Regardless of varying opinions about how AVA should be used, it’s simply unexpected that your tests behave a little different because you name an argument something else. I don’t like surprises in my tests. I believe it’s a good thing to include somewhere in the documentation, to at least make it a little less surprising. If it can be made more robust out of the box, that would be even better though.

docs/recipes/babel.md
novemberborn: It's this confusion that makes me think we should support `assert` usage by default.

novemberborn:
> it’s simply unexpected that your tests behave a little different because you name an argument something else

The tests pass or fail the same though, it's just the extra information that's no longer provided. I think people end up disabling that enhancement fairly often too.

docs/recipes/babel.md
sindresorhus: I was confused by `test.assert` too. I think you should use a totally unrelated name like `foo.assert`.

novemberborn:
> I wonder if, now that only `t.assert()` has Power Assert applied to it, we can't transform `assert` out of the box. I'd rather not document these approaches because it both implies they're supported _and_ that they're necessary.

Let's apply it to anything matching `<identifier>.assert()` and `assert()` itself.

novemberborn:
See https://github.com/avajs/babel-preset-transform-test-files/pull/18.

PR #2134, opened 2019-05-20
Include logs from successful hooks in TAP output

yovasx2:
The `t.log` outputs were not included in the TAP reporter output (`--tap`) unless they failed:
```
TAP version 13
console.log in hook
console.log inside
# suite › api › events › [C806010, monitoring, validation, xxx] [eventsApi ea2] Get Events
ok 1 - suite › api › events › [C806010, monitoring, validation, xxx] [eventsApi ea2] Get Events
  * t.log inside
  * t.log inside2

1..1
# tests 1
# pass 1
# fail 0
```

### DIFF
--- a/lib/reporters/tap.js
+++ b/lib/reporters/tap.js
@@ -2,7 +2,6 @@
 const os = require('os');
 const path = require('path');
 
-const indentString = require('indent-string');
 const plur = require('plur');
 const stripAnsi = require('strip-ansi');
 const supertap = require('supertap');
@@ -119,13 +118,6 @@ class TapReporter {
 		}) + os.EOL);
 	}
 
-	writeLogs(evt) {
-		for (const log of evt.logs) {
-			const logLines = indentString(log, 4).replace(/^ {4}/, '  * ');
-			this.reportStream.write(logLines + os.EOL);
-		}
-	}
-
 	consumeStateChange(evt) { // eslint-disable-line complexity
 		const fileStats = this.stats && evt.testFile ? this.stats.byFile.get(evt.testFile) : null;
 
@@ -137,7 +129,7 @@ class TapReporter {
 				this.writeTest(evt, {passed: false, todo: false, skip: false});
 				break;
 			case 'hook-finished':
-				this.writeLogs(evt);
+				this.writeTest(evt, {passed: true, todo: false, skip: false});
 				break;
 			case 'internal-error':
 				this.writeCrash(evt);
--- a/lib/reporters/tap.js
+++ b/lib/reporters/tap.js
@@ -120,11 +120,9 @@ class TapReporter {
 	}
 
 	writeLogs(evt) {
-		if (evt.logs) {
-			for (const log of evt.logs) {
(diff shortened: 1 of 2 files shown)


yovasx2:
@novemberborn could u please take a look to this?

novemberborn:
@yovasx2 thank you for your PR. I'm busy and won't get to this until the weekend, hopefully.

yovasx2:
Sure!

tymfear:
This looks interesting. Can you add some recipe of how to use it?

yovasx2:
just put a `t.log` inside a hook and it will be output by the logger using `TAP`
``` javascript
before("before hook ", async t =>{
    t.log("A visible log")
});
```

lib/reporters/tap.js
```diff
@@ -118,6 +119,15 @@ class TapReporter {
 		}) + os.EOL);
 	}
 
+	writeLogs(evt) {
+		if (evt.logs) {
```
novemberborn: Looking at the verbose reporter, `hook-finished` always has at least an empty array for `evt.logs` so you shouldn't need this check.

lib/reporters/tap.js
```diff
@@ -128,6 +138,9 @@ class TapReporter {
 			case 'hook-failed':
 				this.writeTest(evt, {passed: false, todo: false, skip: false});
 				break;
+			case 'hook-finished':
+				this.writeLogs(evt);
```
novemberborn: Looking at the verbose reporter, if there are log lines it prints the hook title as well as the logs. Could you do that same check here and use `this.writeTest()` instead?

yovasx2: not really cause the logs are counted as tests and they are not. I just want to print the `log`
https://github.com/avajs/ava/pull/2134/files/eb226a2ed7abc81707b059d53a61b59b79d2f343#diff-35f98ab820e4d7c63d4cc39d54e30b5bL102

also this is another problem when my hook fails it does that and the counts turn into a mess:
```
not ok 23 - api › users › createUser › [monitoring, validation] Verify a user is created
not ok 25 - api › users › createUser › Delete test user for [monitoring, validation] Verify a user is created
```
`Delete test user for` is a hook and the counter is increased because of this:
https://github.com/avajs/ava/pull/2134/files/eb226a2ed7abc81707b059d53a61b59b79d2f343#diff-35f98ab820e4d7c63d4cc39d54e30b5bR139

yovasx2: The tests count is 24 not 25

yovasx2: CC @novemberborn

yovasx2:
CC @sindresorhus

lib/reporters/tap.js
novemberborn: The counting issue aside, the idea of using `writeTest` is that it gives you information about where the log lines came from. So I'd still like to do that.

yovasx2: @novemberborn it's done

novemberborn:
Thanks @yovasx2!

PR #2135, opened 2019-05-21
Added windows match note

### LINKED ISSUE #1691: match params question
when I test doc demo
```
test('foo', t => {
	t.pass();
});

test('moo will also run', t => {
	t.pass();
});

test.only('boo will run but not exclusively', t => {
	t.pass();
});

test.only('sadfsdf', t => {
	t.pass();
});

test(function (t) {
	t.fail();
});

test(function foo(t) {
	t.fail();
});
```

and run `ava --match='*oo*' -v`,but it doesn't work (with win10)

qq 20180206143817

But it can be work in MAC.

And I find the difference between windows and mac

qq 20180206144554

get `[ '\'*oo*\'' ]` in windows
get `[ "*oo*" ]` in mac

why not remove single quotes, so it can be work both of windows and mac
`ava --match=*oo* -v`

rigwild:
Fixes #1691 
Added a note to mention that you need to use double-quotes when using the `match` CLI argument on Windows. It could be useful.

### DIFF
--- a/docs/05-command-line.md
+++ b/docs/05-command-line.md
@@ -122,6 +122,8 @@ test(function foo(t) {
 });
 ```
 
+> Note: You need to use double-quotes when using [Windows](https://github.com/avajs/ava/issues/1691#issuecomment-363355143).
+
 ## Resetting AVA's cache
 
 AVA caches the compiled test and helper files. It automatically recompiles these files when you change them. AVA tries its best to detect changes to your Babel configuration files, plugins and presets. If it seems like your latest Babel configuration isn't being applied, however, you can run AVA with the `--reset-cache` flag to reset AVA's cache. If set, all files in the `node_modules/.cache/ava` directory are deleted. Run AVA as normal to apply your new Babel configuration.


sindresorhus:
Like I commented in #1691, this is expected knowledge when using the Windows command-line. Every tool out there should not have to teach you how to use it.

rigwild:
I get your point, but the way the [matcher syntax](https://github.com/sindresorhus/matcher#usage) looks doesn't make it obvious that it is system-dependent. At first, i though it was just a simple string.
Maybe the note should be in the matcher README ?

novemberborn:
@sindresorhus single or double quote doesn't matter on Linux / MacOS right? So why don't we update our CLI docs to use double quotes everywhere?

sindresorhus:
Because single-quotes have simpler semantics: https://stackoverflow.com/questions/6697753/difference-between-single-and-double-quotes-in-bash

novemberborn:
Right. I suppose we could try and note that the samples are for non-Windows. Not sure how implied that is.

PR #2154, opened 2019-06-06
Improve the TypeScript definition

sindresorhus:


index.d.ts
```diff
@@ -1,3 +1,5 @@
+/// <reference types="node"/>
```
sindresorhus: This causes users to get a more informative error message than:

> error TS2339: Property 'observable' does not exist on type 'SymbolConstructor'.

See https://github.com/sindresorhus/typescript-definition-style-guide#checklist

package.json
```diff
@@ -135,6 +135,7 @@
 		"write-file-atomic": "^3.0.0"
 	},
 	"devDependencies": {
+		"@types/node": "^12.0.5",
```
sindresorhus: We implicitly depended on this type through a sub-dependency. Better to make it explicit.

Debug info:

```
❯ yarn why '@types/node'
yarn why v1.16.0
[1/4] 🤔  Why do we have the module "@types/node"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "@types/node@12.0.5"
info Reasons this module exists
   - "del#@types#glob" depends on it
   - Hoisted from "del#@types#glob#@types#node"
info Disk size without dependencies: "732KB"
info Disk size with unique dependencies: "732KB"
info Disk size with transitive dependencies: "732KB"
info Number of shared dependencies: 0
✨  Done in 0.49s.
```

novemberborn: Should this be `^8` given that's our minimal Node.js version?

sindresorhus: The problem is that the Node.js team often backports features and fixes to Node.js 10 and Node.js 8, but the types are only targeting latest (as far as I know), so if we use `^8`, the types might be missing features added in later Node.js 8 point versions.

novemberborn: No `@types/node` is maintained even for older versions, see the history here: https://www.npmjs.com/package/@types/node

sindresorhus: Oh, interesting. I didn't know that.

PR #2160, opened 2019-06-13
Fix grammar in readme

keyspress:
Line 125 of readme.md starts with
'We aim support all finished syntax proposals,'
Added the word 'to' after 'aim'.
This is my first ever pull request, I hope I am doing this correctly.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -125,7 +125,7 @@ AVA automatically removes unrelated lines in stack traces, allowing you to find 
 
 AVA uses [Babel 7](https://babeljs.io) so you can use the latest JavaScript syntax in your tests. There is no extra setup required. You don't need to be using Babel in your own project for this to work either.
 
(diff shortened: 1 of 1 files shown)


novemberborn:
Hey @keyspress, nice catch! Thanks for the PR!

If you look at the diff there's a lot more changes than adding the `to`. Could you undo those? I suspect your editor caused those changes: you may want to figure out why that happened or use a different editor.

It may be easier to [revert the last commit](https://help.github.com/en/desktop/contributing-to-projects/reverting-a-commit) and then make a new one with just the `to` fix.

keyspress:
Hi @novemberborn,
I believe I have the change down to just the word edit which I pushed to my fork.
I'm not sure if I need to create a new pull request.
Thank you for being so understanding of my newbie mistakes, I really appreciate the help and advice.

novemberborn:
This is great, thanks @keyspress!

PR #2161, opened 2019-06-16
Remove Flow references

langri-sha:
Amends #2098 and entirely removes Flow references, as support currently does not exist and retained information is misleading.


novemberborn:
Ah yes the claim in the README should've been removed, that's an oversight.

However I'd rather keep the recipe with a call to action for https://github.com/avajs/flow-typed/issues/1.

langri-sha:
Retained recipe :+1:.

novemberborn:
Thanks @langri-sha!

PR #2165, opened 2019-06-22
docs(readme): restore vscode debug recipe

cdaringe:
# problem

vscode recipe link is missing from readme

# solution

restore it


novemberborn:
Thanks @cdaringe!

Issue #2152, opened 2019-06-05
Not accepting globs arguments any more?


cancerberoSgx:
Just updated ava and I noticed that is not accepting glob patterns as aguments any more. Example, this worked in the previous version:

`npx ava "./dist/test/**/*Test.js"`

And now it is not. I have to manually expand the pattern :(

`npx ava ./dist/test/*Test.js ./dist/test/**/*Test.js``

I notice that the --help mention this "The above relies on your shell expanding the glob patterns." 

So I wanted to know if this i the right behavior, and if this was a conscious decision, although I'm new using the library I strongly recommend that you support globs and don't rely on the terminal since basically you are forcing users to test the patterns on every platform if using the CLI

Thanks, keep it up, great library!

novemberborn:
Per the release notes: https://github.com/avajs/ava/releases/tag/v2.0.0

> The CLI now only takes file paths, not glob patterns.

If you want to use glob patterns you should specify them in AVA's config, under the `files` key.

(I'm closing this issue for housekeeping purposes, but let's keep the conversation going.)

vladimiry:
Let's say I have 2 npm scripts (of course in reality it might be much more than 2):
```
"test:e2e": "ava --verbose \"./src/e2e/**/*.{spec,test}.ts\"",
"test:main": "ava --verbose \"./src/main/**/*.{spec,test}.ts\"",
```

How do I specify `files` for each run separately with ava@2? Does it mean I will have to list each file explicitly? That would be a very bad UX if it's so.

novemberborn:
@vladimiry for that use case, you could remove the quotes and depending on your shell it should expand to the correct files.

Alternatively you can use `ava.config.js`, perhaps in combination with an environment variable, so that you can customize `files`.

I appreciate that this use case isn't as well served now than it was before. It does greatly simplify the file selection logic however.

We have an issue somewhere about being able to specify the config file. Perhaps that would work nicely for your use case?

vladimiry:
@novemberborn Thanks for the input. Did I get it right that Ava CLI is not going to get back glob files resolving suport?

> @vladimiry for that use case, you could remove the quotes and depending on your shell it should expand to the correct files.

[...]

PR #2166, opened 2019-06-23
Implement --config flag

### LINKED ISSUE #1857: feature request: cli option of config file `--config-file [path]`
[Issuehunt badges](https://issuehunt.io/r/avajs/ava/issues/1857)

## Prerequisites

* [x] issue isn't reported

## Description

hi 👋,

I'm attempting to build nodejs flavour of react-scripts from create-react-app. But instead of jest I would love to use ava. Unfortunately, there is no way to put config inside the project itself. Currently it gotta be either in package.json or in ava.config.js. It will help a lot if I can point ava to the relevant config.

It's very common feature, i can help to implement it.

## Relevant Links

- https://github.com/iamstarkov/zen-scripts here is it

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Sponsors (Total: $60.00)

### Tips

</details>

novemberborn:
Fixes #1857.

To do:

* [x] Resolve whether the config file needs to be top-level, or whether to resolve glob patterns relative to the directory the config file is in https://github.com/avajs/ava/issues/2152#issuecomment-504902827


docs/06-configuration.md
```diff
@@ -115,6 +115,33 @@ export default ({projectDir}) => {
 
 Note that the final configuration must not be a promise.
 
+## Alternative configuration files
+
+The [CLI] lets you specify a specific configuration file, using the `--config` flag. This file is processed just like an `ava.config.js` file would be. When the `--config` flag is set, the provided file will override configuration from the `package.json` and `ava.config.js` files.
```
StarpTech: What happens when I pass a config flag `npx ava --config integration-tests.config.js` + a ava config in the package.json. Are values inherited or they will be treated as two separate configs? This is not clear to me.

novemberborn: The provided file (`integration-tests.config.js`) will override the configuration from the `package.json` file.

StarpTech: What happens when integration-tests.config.js doesn't override a default but the config in package.json? Are values inherited from the package.json?

novemberborn: Could you clarify what you mean by "overriding a default"?

All configuration is overridden. Would it help if I change the line to say:

> the provided file will override all configuration from the `package.json` and `ava.config.js` files.

StarpTech: I got it but I will try explain the other scenario.

1. I have a custom config file like `integration.config.js` with zero configuration.
2. I have a `ava` key in my package.json which defines `failFast: true`
3. I will run the tests with `ava --config integration.config.js`
4. Is `failFast` true?

novemberborn: No.

Normally, if you had config like that *and* an `ava.config.js` file we'd throw an error. The configuration is not merged.

The `--config` flag forgoes the error but the configuration is still not merged. At this stage, the idea is that you manually extend the configuration. We can explore what merging looks like separately.

StarpTech: I like it! Automatic merging could lead to crazy side-effects. Passing multiple configs would be a great option to realize merging e.g if you have a base config and a integration, unit config.

```
--config base.config.js --config int.config.js
```

novemberborn:
@vladimiry pointed out in https://github.com/avajs/ava/issues/2152#issuecomment-504899724 that users may place their config file in nested directories. We resolve glob patterns and requires relative to the project directory, however. And, since the proposed solution for merging configurations is to use object rest/spread, even if we were to resolve things relative to the config file that doesn't work when you merge with another config file.

So instead, until such time as we have a better solution for this, I've opted to require config files to be in the same directory as the `package.json` file.

@vladimiry, @sindresorhus thoughts?

vladimiry:
@novemberborn my opinion won't be relevant here since I'm already using a loader script that resolves the glob pattern and passes the absolute file paths to ava then. But if I were using ava.config.js then I could live with its location to be forced.

sindresorhus:
> So instead, until such time as we have a better solution for this, I've opted to require config files to be in the same directory as the package.json file.

👍

StarpTech:
What is blocking this PR?

novemberborn:
> What is blocking this PR?

I wanted to ship it last week, but then I noticed our tests started failing on Node.js 12. I didn't have time to work on this until now.

novemberborn:
This is now available in v2.2.0.

Issue #2138, opened 2019-05-23
Add ability to run a test residing on a given line


sindresorhus:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%24200%20Funded-%2300A156.svg" />](https://issuehunt.io/r/avajs/ava/issues/2138)

We currently have `--match` for running specific tests with titles matching a pattern, but sometimes you're in your editor and you just want to quickly run the test you're working on. `--match` comes with the overhead of having to think about the pattern and comes with the risk of matching other tests if you're not specific enough.

Would be great to just be able to take the line number and run the test from that information.

Example:

```
$ ava test/foo.js:3
```

Where `3` is the line number of the test. The line number could be be any part of the test method:

```
1: test('foo', t => {
2:   t.true(true);
3: });
```

So using both line number `1`, `2`, and `3` should work.

It should be possible to do this with multiple test files too:

```
$ ava test/foo.js:3 test/bar.js:22
```

If a path ends with `:\d+`, it should not be run through glob resolution.

I got the inspiration from https://blog.bigbinary.com/2016/01/03/test-runner-in-rails-5.html, wrote it down, and forget to open an issue about it until now.

This functionality would also open the opportunity for our editor plugins to have a "Run test at the current line" command, which would be super useful. I'll open an issue about that on the editor plugins if this issue is accepted.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $200.00)

- [<img src='https://avatars3.githubusercontent.com/u/44827199?v=4' alt='issuehunt' width=24 height=24> issuehunt](https://issuehunt.io/u/issuehunt) ($200.00)
### Submitted pull Requests
- [#2181 Add ability to run a test residing on a given line](https://issuehunt.io/r/avajs/ava/pull/2181)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/avajs/ava/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
---
IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

novemberborn:
Sounds good. I suppose we would know that a specific line is selected. How would we determine what test sits on what line though?

Would you be able to specify multiple lines?

> If a path ends with `:\d+`, it should not be run through glob resolution.

[...]

Issue #2158, opened 2019-06-10
Ensure files passed through CLI exist, and are files, before precompiling


novemberborn:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%2440%20Funded-%2300A156.svg" />](https://issuehunt.io/r/avajs/ava/issues/2158)

See #2156, where AVA was called with a test directory as its CLI argument (`ava test`). Precompilation fails because the "file" cannot be read. We should catch this and throw a more appropriate error.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $40.00)

- [<img src='https://avatars3.githubusercontent.com/u/44827199?v=4' alt='issuehunt' width=24 height=24> issuehunt](https://issuehunt.io/u/issuehunt) ($40.00)
### Submitted pull Requests
- [#2184 [Wip]: Ensure files passed through CLI exist, and are files](https://issuehunt.io/r/avajs/ava/pull/2184)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/avajs/ava/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
---
IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

MarchWorks:
On which file should I add the test if necessary?

PR #2181, opened 2019-07-08
Add ability to run a test residing on a given line

qti3e:
:unicorn: Fixes: #2138 

There are just a few edges case that I found it useful to have your comments on :)

- Currently, negative numbers are just ignored, maybe throw an error or something?
- What should happen if the `:...` is actually part of the file name, (`test.js:5` a valid filename on Linux at least)!?
- When running in the watch mode, the line number might change and results in different tests to be executed.
- Due to the implementation using line numbers that are higher than the actual number of lines in the file will result in the last test being executed, it would be nice to have a way to get line numbers in the file without the overhead of reading the file from disk.

Also, there is definitely room for tests to be improved, I couldn't actually get very useful information from `runStatus.stats` for what I had in mind, (name of executed tests) if the current test is not enough I'd be happy to hear about a way to get that information from the API.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#2138: Add ability to run a test residing on a given line](https://issuehunt.io/repos/26820798/issues/2138)
---

</details>


sindresorhus:
Negative numbers should throw a user-friendly error.

sindresorhus:
While `test.js:5` might be a valid file name on Linux, it’s not something anyone would actually use. It’s not a concern.

sindresorhus:
The range syntax should work for the `files` setting in the AVA config too. I don’t see that documented.

sindresorhus:
I’m not sure it makes sense at all to allow the range syntax when using `watch`.

qti3e:
@sindresorhus I guess everything works as expected now, I moved the whole parser logic to a new file, and also added the support for ranges in `glob` as well as in the `files` argument in `API.run()` so that this range syntax thing works everywhere the same way (configs, CLI, etc.)

That being said now we can expect a glob pattern such as `test-*.js:5` to work just fine.

There is only one option added to the `runner` named `lines`, which determines which lines are expected to be executed (an empty array means every line).

There is no negative number allowed in a range string, which means `test.js:-4` is treated as a file with the name `test.js:-4` rather than line 4 in `test.js` and well that leads to a file not found status, which I believe is best to be addressed as part of #2158 (which I'm going to take right after this PR if that's ok), and I'd rather avoid doing this error handling as part of this PR to keep it as simple as it is and keep it clean.

/c.c @novemberborn

lib/runner.js
```diff
@@ -52,11 +58,45 @@ class Runner extends Emittery {
...
 			}
 
+			if (this.lines.length > 0) {
+				this.runOnlyExclusive = true;
+				let callsiteLineNumber;
+
+				for (const jsCallsite of callsites()) {
+					const callsite = sourceMapSupport.wrapCallSite(jsCallsite);
+					const fileName = callsite.getFileName();
+					if (fileName === this.file) {
```
novemberborn: I imagine this will break when using AVA with pre-compiled test files. The selected file may be `build/test.js` but `source-map-support` may translate the callsite to `src/test.js`. You should check the files before the source map is applied.

This also needs a comment to explain why we need to check the files in the first place (because the `test()` may be invoked in a helper file — which is its own edge case that I wouldn't know how to deal with).

lib/runner.js
```diff
@@ -52,11 +58,45 @@ class Runner extends Emittery {
...
+						break;
+					}
+				}
+
+				if (lastMetadata) {
+					// `callsiteLineNumber` is the beginning of the current test, -1 to
+					// get the last line of the previous test.
+					const range = [lastCallsiteLineNumber, callsiteLineNumber - 1];
+					const includedInRange = this.lines.some(n => n >= range[0] && n <= range[1]);
+					lastMetadata.exclusive = this.match.length === 0 ?
```
novemberborn: `exclusive` will be `true` if the test is declared using `.only()`. But, crucially, tests can also be skipped.

I think the ranges should narrow down from the initial test selection. If there are exclusive tests, then only those that match the ranges should be run. If there are no exclusive tests, then skipped tests should never be run.

lib/runner.js
```diff
@@ -52,11 +58,45 @@ class Runner extends Emittery {
...
+						includedInRange :
+						includedInRange && lastMetadata.exclusive;
+				}
+
+				lastCallsiteLineNumber = callsiteLineNumber;
+				lastMetadata = metadata;
```
novemberborn: Rather than tracking the last metedata object and callsite, perhaps you could add the callsite number to the metadata, and then adjust the test selection once all tests have been declared?

lib/runner.js
```diff
@@ -52,11 +58,45 @@ class Runner extends Emittery {
...
+					const fileName = callsite.getFileName();
+					if (fileName === this.file) {
+						callsiteLineNumber = callsite.getLineNumber();
+						break;
+					}
+				}
+
+				if (lastMetadata) {
+					// `callsiteLineNumber` is the beginning of the current test, -1 to
+					// get the last line of the previous test.
```
novemberborn: This does include whitespace, but I think that's OK for now.

lib/range.js
```diff
@@ -0,0 +1,66 @@
+'use strict';
+const rangeParser = require('parse-numeric-range');
+const union = require('array-union');
+
+function parseRange(files) {
+	const any = new Set();
```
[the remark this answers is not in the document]
qti3e: @nov parseRange should work on both ranged paths and normal paths, imagine we have an input like `['a.js', 'a.js:4']` in this case the `a.js` path has a higher priority and should cancel out `a.js`, this set is used to store file paths that do not have a range syntax in them.

lib/range.js
```diff
@@ -0,0 +1,66 @@
...
+			continue;
+		}
+
+		if (rangeStr === '') {
+			any.add(actualFilename);
+			ranges[actualFilename] = [];
+			continue;
+		}
+
+		const lines = rangeParser.parse(rangeStr);
```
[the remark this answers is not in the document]
qti3e: Because of how this implementation handles globs I needed to have the range part as a string, and of course, we need another function to parse that part.

lib/range.js
```diff
@@ -0,0 +1,66 @@
...
+	};
+}
+
+exports.parseRange = parseRange;
+
+const rangeRegExp = /^(\d+(-\d+)?,?)+$/;
+const splitRangeStrCache = new Map();
+
+function splitRangeStr(file) {
+	if (splitRangeStrCache.has(file)) {
```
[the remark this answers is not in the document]
qti3e: It is called many times with the same input in different places, so I thought it's better to cache the result.

lib/range.js
```diff
@@ -0,0 +1,66 @@
...
+		return [...splitRangeStrCache.get(file)];
+	}
+
+	const colonIndex = file.lastIndexOf(':');
+	const mightHaveRange = colonIndex > -1 && file[colonIndex + 1] !== '\\';
```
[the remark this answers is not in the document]
qti3e: That regex I'm using might be a bit heavy so I use a fast and dirty check first and store the result in `mightHaveRange`, also there is no need for a test if you remove this check (and the regex), most of the tests will fail on windows.

lib/runner.js
qti3e: We can't check the file before the source map is applied after the translation line numbers are all changed, (`a.js:5` might end up running line 6.)

lib/runner.js
qti3e: I don't think there is a way to handle this case unless you parse the source file to it's AST and then extract when the function call actually ends (this information is not in the callsite data.)

qti3e:
@novemberborn Please let me know your final decision about dropping ranged globs.

sindresorhus:
> I don't see how this is useful. If you want to always run part of a test file, make a new test file. Our globbing logic is still really complex and I don't think we should add more complexity to it.

I agree. I didn't think that through when first commenting.

> Having reviewed the PR it definitely looks easier to implement and maintain if we only support this syntax on the CLI.

👍

sindresorhus:
> There is no negative number allowed in a range string, which means test.js:-4 is treated as a file with the name test.js:-4 rather than line -4 in test.js and well that leads to a file not found status

I think that should be handled here. It's clearly not a valid filename, so we should give the user a helpful and relevant error message that it's an invalid range.

lib/range.js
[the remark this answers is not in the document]
novemberborn: I see. Should that maybe print an error instead? Or perhaps we could run only the selected lines and print a warning "ignoring other tests from a.js since you selected lines" (well not quite that, but you get the idea).

lib/range.js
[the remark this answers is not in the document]
novemberborn: I'm not too fussed about performance overheads, presumably you only run this with a handful of ranges so it'll be really hard to tell what's faster or slower.

> here is no need for a test if you remove this check (and the regex), most of the tests will fail on windows.

Are you saying the test is not necessary, or it is because of Windows?

lib/range.js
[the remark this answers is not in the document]
novemberborn: I doubt that's necessary.

lib/range.js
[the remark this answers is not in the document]
novemberborn: Right, so this will be a lot simpler without globs 😄

lib/runner.js
novemberborn: > We can't check the file before the source map is applied after the translation line numbers are all changed, (`a.js:5` might end up running line 6.)

Yes I get that. My point though is that applying the source map may change the `fileName` of the call site.

lib/range.js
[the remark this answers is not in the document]
qti3e: you can look at this from when this check was not there https://travis-ci.org/avajs/ava/builds/556309917

qti3e:
@novemberborn PTAL, also please restart that Travis task.

lib/range.js
```diff
@@ -0,0 +1,53 @@
+'use strict';
+const rangeParser = require('parse-numeric-range');
+
+function parseFileSelections(files) {
+	const ranges = new Map();
+	const ignored = [];
+	const filenames = [];
```
novemberborn: Let's make this a set, too.

lib/range.js
```diff
@@ -0,0 +1,53 @@
...
+
+	for (const file of files) {
+		const [actualFilename, lines] = parseFileSelection(file);
```
novemberborn: With `filenames` a set, you can add the `actualFilename` to it here, regardless of any other condition.

lib/range.js
```diff
@@ -0,0 +1,53 @@
...
+
+		if (!ranges.has(actualFilename)) {
+			ranges.set(actualFilename, []);
+			filenames.push(actualFilename);
+		}
+
+		if (lines.length === 0 && ranges.get(actualFilename).length > 0) {
```
novemberborn: Let's move this up, and change the condition to `lines.length === 0 && ranges.has(actualFilename)`.

lib/range.js
```diff
@@ -0,0 +1,53 @@
+'use strict';
+const rangeParser = require('parse-numeric-range');
+
+function parseFileSelections(files) {
+	const ranges = new Map();
+	const ignored = [];
```
novemberborn: This should be a set, so you don't need to dedupe it later.

lib/range.js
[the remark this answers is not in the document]
novemberborn: Right, so let's add a comment explaining this handles file paths on Windows that do not include ranges.

lib/range.js
```diff
@@ -0,0 +1,53 @@
...
+
+exports.parseFileSelections = parseFileSelections;
+
+const rangeRegExp = /^(\d+(-\d+)?,?)+$/;
+
+function parseFileSelection(file) {
+	const colonIndex = file.lastIndexOf(':');
+	const mightHaveRange = colonIndex > -1 && file[colonIndex + 1] !== '\\';
+	const rangeStr = file.slice(colonIndex + 1);
+	const actualFilename = file.slice(0, colonIndex);
```
novemberborn: This isn't really a filename though, it's still a file path. Let's rename this throughout.

lib/range.js
```diff
@@ -0,0 +1,53 @@
+'use strict';
+const rangeParser = require('parse-numeric-range');
+
+function parseFileSelections(files) {
+	const ranges = new Map();
```
novemberborn: `ranges` no longer holds ranges. It holds line numbers. Shall we rename this?

lib/cli.js
```diff
@@ -280,7 +282,14 @@ exports.run = async () => { // eslint-disable-line complexity
 		});
 		watcher.observeStdin(process.stdin);
 	} else {
-		const runStatus = await api.run(files);
+		const {filenames, ranges, ignored} = rangeParser.parseFileSelections(files);
+		if (ignored.length > 0) {
+			for (const file of ignored) {
```
novemberborn: No need to add the length check.

lib/cli.js
```diff
@@ -280,7 +282,14 @@ exports.run = async () => { // eslint-disable-line complexity
...
+				console.error(`Ignoring other tests from ${file} since you selected lines.`);
```
novemberborn: We can't write to the console from here, it messes with the reporters. I think for now we can use `exit()` here and explain that if you select lines, you shouldn't then also specify the same file without a line selection, and then print the offending files.

lib/cli.js
```diff
@@ -280,7 +282,14 @@ exports.run = async () => { // eslint-disable-line complexity
 		});
 		watcher.observeStdin(process.stdin);
 	} else {
-		const runStatus = await api.run(files);
+		const {filenames, ranges, ignored} = rangeParser.parseFileSelections(files);
```
novemberborn: What do you think about executing this when `cli.input` is processed? Then in the `conf.watch` branch we can exit with an error if lines are selected.

lib/runner.js
```diff
@@ -52,11 +56,29 @@ class Runner extends Emittery {
...
 				});
 			}
 
+			if (this.lines.length > 0) {
+				let callsiteLineNumber;
+
+				for (const jsCallsite of callsites()) {
+					const callsite = sourceMapSupport.wrapCallSite(jsCallsite);
+					const fileName = callsite.getFileName();
+					if (jsCallsite === this.file || fileName === this.file) {
```
novemberborn: Did you mean `jsCallsite.getFileName()`?

I don't think this needs to be compared to the source-map converted filename. You can't pass those on the CLI since you need to provide the file that AVA can actually load.

lib/runner.js
```diff
@@ -52,11 +56,29 @@ class Runner extends Emittery {
...
+					this.callsiteLineNumbers.sort((a, b) => a - b);
 					hasStarted = true;
 					this.start();
 				});
 			}
 
+			if (this.lines.length > 0) {
+				let callsiteLineNumber;
+
+				for (const jsCallsite of callsites()) {
```
novemberborn: Let's rename `jsCallsite` to `callsite`.

lib/runner.js
```diff
@@ -52,11 +56,29 @@ class Runner extends Emittery {
...
+					const callsite = sourceMapSupport.wrapCallSite(jsCallsite);
```
novemberborn: And rename this to `wrapped`.

lib/runner.js
```diff
@@ -52,11 +56,29 @@ class Runner extends Emittery {
 			if (!scheduledStart) {
 				scheduledStart = true;
 				process.nextTick(() => {
+					this.callsiteLineNumbers.push(Infinity);
+					this.callsiteLineNumbers.sort((a, b) => a - b);
```
novemberborn: Let's do this work inside the `start()` method instead.

We should deduplicate the line numbers as well. It's an edge case but you could specify two tests on the same line.

lib/api.js
```diff
@@ -51,7 +51,12 @@ class Api extends Emittery {
 	}
 
 	async run(files = [], runtimeOptions = {}) {
-		files = files.map(file => path.resolve(this.options.resolveTestsFrom, file));
+		const ranges = new Map(runtimeOptions.ranges || []);
+		files = files.map(file => {
+			const filename = path.resolve(this.options.resolveTestsFrom, file);
+			ranges.set(filename, ranges.get(file));
+			return filename;
+		});
```
novemberborn: Why not do this processing in the CLI instead?

novemberborn:
Closing due to inactivity. @qti3e I do hope you'll find the time to come back to this!

PR #2184, opened 2019-07-13
Fixes #2158: Ensure files passed through CLI exist, and are files

### LINKED ISSUE #2158: Ensure files passed through CLI exist, and are files, before precompiling

MarchWorks:
Fixes #2158

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#2158: Ensure files passed through CLI exist, and are files, before precompiling](https://issuehunt.io/repos/26820798/issues/2158)
---

</details>

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -1,5 +1,6 @@
 'use strict';
 const path = require('path');
+const fs = require('fs');
 const del = require('del');
 const updateNotifier = require('update-notifier');
 const figures = require('figures');
@@ -269,6 +270,15 @@ exports.run = async () => { // eslint-disable-line complexity
 	});
 
 	const files = cli.input.map(file => path.relative(resolveTestsFrom, path.resolve(process.cwd(), file)));
+
+	files.forEach((file, index) => {
+		if(!fs.existsSync(file)){
+			throw new Error(`No sush file: ${cli.input[index]}`);
+		}
+		if(fs.lstatSync(file).isFile()){
+			throw new Error(`${cli.input[index]} should be a file`);
+		}
+	})
 
 	if (conf.watch) {
 		const watcher = new Watcher({
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -215,6 +215,8 @@ exports.run = async () => { // eslint-disable-line complexity
 	const match = arrify(conf.match);
 	const resolveTestsFrom = cli.input.length === 0 ? projectDir : process.cwd();
 
+	const files = cli.input.map(file => path.relative(resolveTestsFrom, path.resolve(process.cwd(), file)));
+
 	for (const file of cli.input) {
 		try {
 			const stats = fs.statSync(file);
@@ -289,13 +291,13 @@ exports.run = async () => { // eslint-disable-line complexity
 		const watcher = new Watcher({
 			api,
 			reporter,
-			files: cli.input,
+			files,
 			globs,
 			resolveTestsFrom
 		});
 		watcher.observeStdin(process.stdin);
 	} else {
-		const runStatus = await api.run(cli.input);
+		const runStatus = await api.run(files);
 		process.exitCode = runStatus.suggestExitCode({matching: match.length > 0});
 		reporter.endRun();
 	}
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -214,6 +214,21 @@ exports.run = async () => { // eslint-disable-line complexity
 
 	const match = arrify(conf.match);
 	const resolveTestsFrom = cli.input.length === 0 ? projectDir : process.cwd();
+
+	const files = cli.input.map(file => path.relative(resolveTestsFrom, path.resolve(process.cwd(), file)));
+	for (const file of files) {
+		try {
+			const stats = fs.statSync(file);
+			if (!stats.isFile()) {
+				exit(`${cli.input[files.indexOf(file)]} should be a file`);
+			}
+		} catch (error) {
+			if (error.code === 'ENOENT') {
+				exit(`No sush file: ${cli.input[files.indexOf(file)]}`);
+			}
+		}
+	}
+
 	const api = new Api({
 		babelConfig,
 		cacheEnabled: conf.cache !== false,
@@ -268,18 +283,6 @@ exports.run = async () => { // eslint-disable-line complexity
 			}
 		});
 	});
-
-	const files = cli.input.map(file => path.relative(resolveTestsFrom, path.resolve(process.cwd(), file)));
-
-	for (const file of files) {
-		if (!fs.existsSync(file)) {
-			throw new Error(`No such file: ${cli.input[files.indexOf(file)]}`);
-		}
-
-		if (!fs.lstatSync(file).isFile()) {
-			throw new Error(`${cli.input[files.indexOf(file)]} should be a file`);
-		}
-	}
 
 	if (conf.watch) {
 		const watcher = new Watcher({
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -215,6 +215,7 @@ exports.run = async () => { // eslint-disable-line complexity
 	const match = arrify(conf.match);
 	const resolveTestsFrom = cli.input.length === 0 ? projectDir : process.cwd();
 
+	const files = cli.input.map(file => path.relative(resolveTestsFrom, path.resolve(process.cwd(), file)));
 	for (const file of cli.input) {
 		try {
 			const stats = fs.statSync(file);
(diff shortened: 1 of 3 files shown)


MarchWorks:
I get it. This is not what I should do I will fix this now

MarchWorks:
In which file should I add the tests

MarchWorks:
Done

lib/cli.js
```diff
@@ -270,6 +271,16 @@ exports.run = async () => { // eslint-disable-line complexity
 
 	const files = cli.input.map(file => path.relative(resolveTestsFrom, path.resolve(process.cwd(), file)));
 
+	for (const file of files) {
+		if (!fs.existsSync(file)) {
+			throw new Error(`No such file: ${cli.input[files.indexOf(file)]}`);
+		}
+
+		if (!fs.lstatSync(file).isFile()) {
```
novemberborn: I think this should use `statSync()`, since `lstatSync()` doesn't follow symlinks.

lib/cli.js
```diff
@@ -270,6 +271,16 @@ exports.run = async () => { // eslint-disable-line complexity
 
 	const files = cli.input.map(file => path.relative(resolveTestsFrom, path.resolve(process.cwd(), file)));
 
+	for (const file of files) {
+		if (!fs.existsSync(file)) {
```
novemberborn: Using `statSync()` (see below) should be sufficient. It'll throw if the file (or its path) does not exist, but you can catch the error and use `error.code === 'ENOENT' || error.code === 'ENOTDIR'` to recognize whether the file existed. See https://nodejs.org/api/errors.html#errors_common_system_errors.

lib/cli.js
```diff
@@ -270,6 +271,16 @@ exports.run = async () => { // eslint-disable-line complexity
 
 	const files = cli.input.map(file => path.relative(resolveTestsFrom, path.resolve(process.cwd(), file)));
```
novemberborn: Let's move this, and the validation, above the `new Api()` instantiation.

lib/cli.js
```diff
@@ -270,6 +271,16 @@ exports.run = async () => { // eslint-disable-line complexity
...
 
+	for (const file of files) {
+		if (!fs.existsSync(file)) {
+			throw new Error(`No such file: ${cli.input[files.indexOf(file)]}`);
+		}
+
+		if (!fs.lstatSync(file).isFile()) {
+			throw new Error(`${cli.input[files.indexOf(file)]} should be a file`);
```
novemberborn: Use the `exit()` function instead.

lib/cli.js
```diff
@@ -270,6 +271,16 @@ exports.run = async () => { // eslint-disable-line complexity
 
 	const files = cli.input.map(file => path.relative(resolveTestsFrom, path.resolve(process.cwd(), file)));
 
+	for (const file of files) {
+		if (!fs.existsSync(file)) {
+			throw new Error(`No such file: ${cli.input[files.indexOf(file)]}`);
```
novemberborn: Use the `exit()` function instead.

lib/cli.js
MarchWorks: No need for `error.code === 'ENOTDIR'`

lib/cli.js
```diff
@@ -214,6 +214,21 @@ exports.run = async () => { // eslint-disable-line complexity
...
+	const files = cli.input.map(file => path.relative(resolveTestsFrom, path.resolve(process.cwd(), file)));
+	for (const file of files) {
+		try {
+			const stats = fs.statSync(file);
+			if (!stats.isFile()) {
+				exit(`${cli.input[files.indexOf(file)]} should be a file`);
+			}
+		} catch (error) {
+			if (error.code === 'ENOENT') {
+				exit(`No sush file: ${cli.input[files.indexOf(file)]}`);
```
novemberborn: `such`

lib/cli.js
```diff
@@ -214,6 +214,21 @@ exports.run = async () => { // eslint-disable-line complexity
...
+			}
+		}
```
novemberborn: You need to rethrow any other error.

MarchWorks: I used `throw new Error(error)` at first but I changed it to `exit` after. I am not sure which one is more appropriate in this case because this error is an unexpected one

MarchWorks:
Done hopefully

lib/cli.js
```diff
@@ -213,6 +214,23 @@ exports.run = async () => { // eslint-disable-line complexity
 
 	const match = arrify(conf.match);
 	const resolveTestsFrom = cli.input.length === 0 ? projectDir : process.cwd();
+
+	const files = cli.input.map(file => path.relative(resolveTestsFrom, path.resolve(process.cwd(), file)));
+	for (const file of files) {
+		try {
+			const stats = fs.statSync(file);
+			if (!stats.isFile()) {
+				exit(`${cli.input[files.indexOf(file)]} should be a file`);
```
novemberborn: Let's change this to:

```js
`${file} is not a test file.`
```

lib/cli.js
```diff
@@ -213,6 +214,23 @@ exports.run = async () => { // eslint-disable-line complexity
...
+			}
+		} catch (error) {
+			if (error.code === 'ENOENT') {
+				exit(`No such file: ${cli.input[files.indexOf(file)]}`);
```
novemberborn: How about this?

```js
`${file} does not exist.`
```

lib/cli.js
```diff
@@ -213,6 +214,23 @@ exports.run = async () => { // eslint-disable-line complexity
...
+			} else {
+				exit(`An unexpected error has occurred:\n${error}`);
```
novemberborn: Taking the lead from this line:

https://github.com/avajs/ava/blob/2fc7d56475ca1d77078abc3e08390341bedda58f/lib/cli.js#L145

How about:

```js
`Error accessing ${file}\n\n${chalk.gray((error && error.stack) || error)}`
```

lib/cli.js
```diff
@@ -213,6 +214,23 @@ exports.run = async () => { // eslint-disable-line complexity
 
 	const match = arrify(conf.match);
 	const resolveTestsFrom = cli.input.length === 0 ? projectDir : process.cwd();
+
+	const files = cli.input.map(file => path.relative(resolveTestsFrom, path.resolve(process.cwd(), file)));
```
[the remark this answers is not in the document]
MarchWorks: Yes you are right I tested it they give the same output 
`[folder/filename]` from `cli.input` and `[folder\\filename]` from `files`

lib/cli.js
MarchWorks: Yes better

novemberborn:
@MarchWorks don't forget to update the tests for the new error messages 😉

MarchWorks:
I completely forget that

lib/cli.js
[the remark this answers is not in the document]
MarchWorks: @novemberborn seems that it's necessary it cause some tests to fail with node 8
https://travis-ci.org/avajs/ava/builds/564659393?utm_source=github_status&utm_medium=notification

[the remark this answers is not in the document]
MarchWorks: It fail on windows with node 8 most likely it has something to do with slash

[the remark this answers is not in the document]
novemberborn: Yes I realized it was existing code. I wrote it, too! Let’s keep it then.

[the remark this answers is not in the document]
MarchWorks: Then everything is done I think

novemberborn:
Thanks @MarchWorks!

lib/cli.js
```diff
@@ -213,6 +214,24 @@ exports.run = async () => { // eslint-disable-line complexity
...
+
+	for (const file of cli.input) {
```
ulken:
```suggestion
      for (const file of files) {
```

@novemberborn Right? If not, I suggest the mapping from `cli.input` to `files` move below this block (as before).

novemberborn: @ulken Yes good catch. Would you like to open a PR?

ulken:
I’ve fixed it in my upcoming PR for #2138 (not unrelated fix, actually needed).

Should be ready for review soonish. I haven’t been able to work as much as I thought on it and ran in to a few bumps, but almost there now. Will open a PR when implementation is done and then do the docs if you want to start looking into it.

/Oskar

PR #2186, opened 2019-07-14
Change default Context type to unknown

novemberborn:
Previously, the Context was typed as an empty object. This suddenly started giving problems when typing `t.try()` (https://github.com/avajs/ava/pull/1947#issuecomment-498038318). Setting it to `unknown` seems reasonable.

This may break some code, but I'm not sure how you'd use context without typing it, and IIRC there was a non-major TypeScript change that led to similar breakage regarding the `{}` type. It seems like the kind of breakage that TypeScript users are used to, so we shouldn't need a major bump to release this. @sindresorhus?

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -353,7 +353,7 @@ export interface TruthyAssertion {
 }
 
 /** The `t` value passed to test & hook implementations. */
-export interface ExecutionContext<Context = {}> extends Assertions {
+export interface ExecutionContext<Context = unknown> extends Assertions {
 	/** Test context, shared with hooks. */
 	context: Context;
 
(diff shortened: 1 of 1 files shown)


vintprox:
Used along with TypeScript, `context` of unknown type triggers errors when I try to add property to it. Shouldn't it be of  `any` type?

novemberborn:
You can type `context` with some type casts, see https://github.com/avajs/ava/blob/master/docs/recipes/typescript.md#typing-tcontext. I'm working on a new way of building `test()` functions where the context can be inferred, keep an eye on #2435 for that.

PR #2178, opened 2019-07-05
do not report failing hooks as test

mikaoelitiana:
Fixes #2142

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

This pull request has been submitted to:
- [#2142: Hooks are counted as tests](https://issuehunt.io/repos/26820798/issues/2142)
---

IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

PR #2193, opened 2019-07-23
Not count hooks as tests

### LINKED ISSUE #2142: Hooks are counted as tests
[Issuehunt badges](https://issuehunt.io/r/avajs/ava/issues/2142)

When a hook fails it increases the count of tests:
```
not ok 23 - api › users › createUser › [C806006, monitoring, validation] Verify a user is created
not ok 25 - api › users › createUser › Delete test user for [C806006, monitoring, validation] Verify a user is created
```
**Delete test user** is a hook and it counted as a test, the current number of tests is 24, not 25

this is happening for verbose and tap formats

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $50.00)

- [issuehunt issuehunt](https://issuehunt.io/u/issuehunt) ($50.00)
### Submitted pull Requests
- [#2178 do not report failing hooks as test](https://issuehunt.io/r/avajs/ava/pull/2178)
---

### Tips

---
</details>

yovasx2:
fixes #2142 
This contains specs modifications for #2178

### DIFF
--- a/lib/reporters/tap.js
+++ b/lib/reporters/tap.js
@@ -119,9 +119,9 @@ class TapReporter {
 		}) + os.EOL);
 	}
 
-	writeLogs(evt, flags = {error: false, title: ''}) {
-		let formattedTitle = flags.title || this.prefixTitle(evt.testFile, evt.title);
-		if (flags.error) {
+	writeComment(evt, {error = false, title = this.prefixTitle(evt.testFile, evt.title)}) {
+		let formattedTitle = title;
+		if (error) {
 			formattedTitle = `Failed hook: ${formattedTitle}`;
 		}
 
@@ -142,10 +142,10 @@ class TapReporter {
 				// Ignore
 				break;
 			case 'hook-failed':
-				this.writeLogs(evt, {error: true});
+				this.writeComment(evt, {error: true});
 				break;
 			case 'hook-finished':
-				this.writeLogs(evt);
+				this.writeComment(evt, {});
 				break;
 			case 'internal-error':
 				this.writeCrash(evt);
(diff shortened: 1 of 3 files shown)


yovasx2:
@novemberborn @sindresorhus  please take a look

lib/reporters/tap.js
```diff
@@ -118,6 +119,21 @@ class TapReporter {
 		}) + os.EOL);
 	}
 
+	writeLogs(evt, flags = {error: false, title: ''}) {
+		let formattedTitle = flags.title || this.prefixTitle(evt.testFile, evt.title);
+		if (flags.error) {
+			formattedTitle = `Failed hook: ${formattedTitle}`;
+		}
+
+		this.reportStream.write(`# ${stripAnsi(formattedTitle)}${os.EOL}`);
```
novemberborn: What's the reason for not using `supertap` here?

lib/reporters/tap.js
```diff
@@ -118,6 +119,21 @@ class TapReporter {
 		}) + os.EOL);
 	}
 
+	writeLogs(evt, flags = {error: false, title: ''}) {
+		let formattedTitle = flags.title || this.prefixTitle(evt.testFile, evt.title);
```
novemberborn: This would be better as `writeLogs(evt, { error = false, title = this.prefixTitle(evt.testFile, evt.title) } = {})`.

lib/reporters/tap.js
```diff
@@ -118,6 +119,21 @@ class TapReporter {
 		}) + os.EOL);
 	}
 
+	writeLogs(evt, flags = {error: false, title: ''}) {
```
novemberborn: In my mind, "logs" here refers to `evt.logs`, but that doesn't quite cover the scenarios in which it's called. What are you referring to? Should this be `writeTapComment()`?

lib/reporters/tap.js
yovasx2: `supertap`'s API is just having **start**, **finish** and **test**, methods, I cannot use start or finish, the only remaining one is **test** but it's injecting prefix **(not)ok**, so it's counted as test, we need to modify supertap to have another method available called hook to just write the content of the hook without outputting the test format.

lib/reporters/tap.js
yovasx2: I will use: `writeLogs(evt, { error = false, title = this.prefixTitle(evt.testFile, evt.title) }) {` without default value to prevent bad behavior, in that case if you don;t pass the object deconstruction fails and you are acknowledged

lib/reporters/tap.js
yovasx2: I agree the proposed name is better, but since any function inside the reporter uses TAP, I will choose `writeComment`

lib/reporters/tap.js
novemberborn: Fair enough. I suppose it'd be good to contribute a comment method to `supertap` but that's up to you.

PR #2196, opened 2019-07-25
extend --match to support hash (generated from unique title)

timoweave:
1. generate hash from title, which is unique
2. use --matching or -m to find the hash to run

Screen Shot 2019-07-25 at 1 42 17 PM


timoweave:
need to fix lint stuff....

sindresorhus:
There's no need to open a new PR. That just creates needless noise.

Issue #1472, opened 2017-07-26
Add `--only` flag for running `test.only` tests


sindresorhus:
From https://github.com/avajs/ava/pull/1467#issuecomment-317859343

https://github.com/avajs/ava/pull/1467 makes `test.only()` not be exclusive with multiple test files, so would be nice to have this flag.

novemberborn:
> #1467 makes test.only() not be exclusive with multiple test files, so would be nice to have this flag.

To be exact, AVA won't switch into exclusive mode until the first `test.only()` file is encountered. For subsequent files only exclusive tests will be run.

revelt:
For the record, the new `.only` behaviour is quite annoying and actually impacts my productivity. I often separate the secondary library's functionality into `util.js` and therefore, unit tests go separately too. When I troubleshoot code, I rely on `console.log`'s and it is especially important that I should be able to isolate that one particular unit test, otherwise, I'll get multiple `console.log`s from multiple unit tests.

Now, if we'll end up with functionality as Mark says, where `t.only` kicks in only after first `only` is encountered, it's pretty much a _futile feature_ – I just checked, nearly always my `util` unit tests run first in the queue and therefore would contaminate the console. We need something more.

Maybe it's a stupid idea, but why can't we just run a glob against all recognised test files and look for any `.only`s in the unit tests? Then, have a think what to do, run `.only`s or run everything as it is now? Basically, a one, additional, separate round in the execution chain??? Because we're already globb'ing all `test` files, we just need to scan them first, before action commences.

What do you think?

novemberborn:
@revelt,

> why can't we just run a glob against all recognised test files and look for any .onlys in the unit tests? Then, have a think what to do, run .onlys or run everything as it is now? Basically, a one, additional, separate round in the execution chain??? Because we're already globb'ing all test files, we just need to scan them first, before action commences.

We did consider this previously. There's many ways of using AVA though, and scanning for the occurrence of `.only(` is not sufficient.

> if we'll end up with functionality as Mark says, where t.only kicks in only after first only is encountered, it's pretty much a futile feature – I just checked, nearly always my util unit tests run first in the queue and therefore would contaminate the console. We need something more.

[...]

PR #2197, opened 2019-07-25
extend --match to support hash (generated from unique title)

timoweave:
extend --match to support hash key, which is generated from title (which ava ensure title to be unique).

Screen Shot 2019-07-25 at 1 42 17 PM

> What problems does this solve?

enhancement (how to add label)?

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -39,7 +39,7 @@ exports.run = async () => { // eslint-disable-line complexity
 
 		Options
 		  --watch, -w             Re-run tests when tests and source files change
-		  --match, -m             Only run tests with matching title (Can be repeated)
+		  --match, -m             Only run tests with matching title/hash (Can be repeated)
 		  --update-snapshots, -u  Update snapshots
 		  --fail-fast             Stop after first test failure
 		  --timeout, -T           Set global timeout (milliseconds or human-readable, e.g. 10s, 2m)
--- a/lib/reporters/colors.js
+++ b/lib/reporters/colors.js
@@ -12,5 +12,6 @@ module.exports = {
 	errorSource: chalk.gray,
 	errorStack: chalk.gray,
 	stack: chalk.red,
+	hash: chalk.dim,
 	information: chalk.magenta
 };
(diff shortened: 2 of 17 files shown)


sindresorhus:
What problems does this solve?

timoweave:
request enhancement (and for other new flag --ignore, --match, --find
(https://github.com/avajs/ava/issues/1472)

novemberborn:
Hi @timoweave, impressive work on the PR!

To avoid disappointment though, I'd advise filing an issue with your proposal first before doing all the work. As evidenced by @sindresorhus's question above, I don't think this solves an actual problem. It puts more noise in the reporter output and it's not really easier than using `--match` or adding `.only` or `.skip` modifiers inside your test file.

I'm looking forward to other PRs from you though! 😄 

timoweave:
Oh, what kinda feature do you need help on?

novemberborn:
@timoweave have a look at https://github.com/avajs/ava/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3A"help+wanted" — there'll be more esoteric issues without that label if you want to venture really deep into the weeds.

timoweave:
Need to prioritize 140 issues...
What are the top 25?

novemberborn:
> Need to prioritize 140 issues...

There's "only" 82 with a help wanted label 😉 https://github.com/avajs/ava/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3A"help+wanted"

We don't have a priority list as such, but you could use the [Funded on IssueHunt](https://github.com/avajs/ava/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3A"help+wanted"+label%3A"%3Adollar%3A+Funded+on+Issuehunt") labels as a proxy. Bugs tend to be more important to solve than feature requests.

Generally speaking I find it's most fun to solve a problem you run into yourself, or add a feature you would see yourself using. You can also have a look at the `scope:` labels to get a sense of where in the codebase the issue is.

timoweave:
Yeah, it's most fun to solve a problem you run into yourself, or add a feature (but you turn it down last time) you would see yourself using.

PR #2199, opened 2019-08-01
Added gutter descriptors

### LINKED ISSUE #1558: Explain diff gutter symbols
[Issuehunt badges](https://issuehunt.io/r/avajs/ava/issues/1558)

Please see this comment: https://github.com/avajs/ava/issues/1558#issuecomment-469031454

Original issue follows.

---

## Description

When using `t.snapshot(foo)` and `foo` does not match the snapshot, the diff that is shown uses `-` to indicate the new `foo` values and `+` to show the old `foo` values.

This is unintuitive.

I believe the problem may be at https://github.com/avajs/ava/blob/14f7095d25abc5ffbff7efd7db962eaf5e86daab/lib/assert.js#L321

### Test Source

```js
import test from 'ava'
test('snapshot diff', t => {
  t.snapshot(`\n${Date.now()}\n`, 'the - value should be less than the + value')
})
```

## Environment

Latest stable ava on OS X

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $60.00)

### Submitted pull Requests
- [#2199 Added gutter descriptors](https://issuehunt.io/r/avajs/ava/pull/2199)
---

### Tips

---
</details>

kire73:
Just modified the concordance-options.js
It seemed like the most straightforward way to achieve this enhancement. I'm unsure if there is a reason this may not be a desirable approach.

Fixes #1558.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#1558: Explain diff gutter symbols](https://issuehunt.io/repos/26820798/issues/1558)
---

</details>

### DIFF
--- a/lib/concordance-options.js
+++ b/lib/concordance-options.js
@@ -21,7 +21,7 @@ const colorTheme = {
 		value: ansiStyles.blue
 	},
 	diffGutters: {
-		actual: forceColor.red('- Recieved') + '    ',
+		actual: forceColor.red('- Received') + '    ',
 		expected: forceColor.green('+ Expected') + '    ',
 		padding: '  '
 	},
--- a/lib/concordance-options.js
+++ b/lib/concordance-options.js
@@ -21,8 +21,8 @@ const colorTheme = {
 		value: ansiStyles.blue
 	},
 	diffGutters: {
-		actual: forceColor.red('-') + ' ',
-		expected: forceColor.green('+') + ' ',
+		actual: forceColor.red('- Recieved') + '    ',
+		expected: forceColor.green('+ Expected') + '    ',
 		padding: '  '
 	},
 	error: {


novemberborn:
Hey @kire73 thanks for the PR! My initial reaction is that it might be better to explain the symbols above the output, rather than adding the additional text on each line. What do you think?

kire73:
Sure, seems like it would look nicer. Could be solved easily on the same lines with:
```
		actual: forceColor.red('Received\n-') + '    ',
		expected: forceColor.green('Expected\n+') + '    ',
```
Changing the resulting output to:
image

Should I add the commit for this?

novemberborn:
> Sure, seems like it would look nicer. Could be solved easily on the same lines with:
> 
> ```
> 		actual: forceColor.red('Received\n-') + '    ',
> 		expected: forceColor.green('Expected\n+') + '    ',
> ```

That would put `Received\n` in front of every line in the diff. We should output it once before the diff is shown.

novemberborn:

mesqueeb:
I feel that in front of every line is very clear. Then I'll never forget or get confused.
+1 to adding it to every line for me!!

PR #2200, opened 2019-08-05
Center the logo in the readme

Illu:
Small adjustment to the readme, it looks better imho 😉

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1,5 +1,5 @@
 <h1 align="center">
-  <img src="media/header.png" title="AVA" alt="AVA logo" width="530">
+	<img src="media/header.png" title="AVA" alt="AVA logo" width="530">
 </h1>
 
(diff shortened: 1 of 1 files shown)


novemberborn:
Sure, not opposed to this. Though I'm not 100% sure how this would render on npm. Do you know of any other packages that use this kind of markup?

Illu:
You're right, I tried it with one of my packages and it doesn't seem to render correctly on npm. I'm closing this PR.

sindresorhus:
We should not care how it renders on npm. Their Markdown rendering is not good anyway.

novemberborn:
@Illu when you say it didn't render correctly, do you mean it wasn't centered, or it didn't show up / HTML tags were visible?

Illu:
@novemberborn the logo didn't show up at all. However, I found [this package](https://www.npmjs.com/package/@nozbe/watermelondb) that uses the same kind of markup to render its logo.

novemberborn:
@Illu that uses:

```html
<p align="center">
  <img src="https://github.com/Nozbe/WatermelonDB/raw/master/assets/logo-horizontal2.png" alt="WatermelonDB" width="539" />
</p>
```

We could put something like that at the top of the README. @gurrrung opened https://github.com/avajs/ava/pull/2214 which wraps that in an H1. I wonder if that would work?

novemberborn:

PR #2204, opened 2019-08-07
docs: add comment about global and concurrent test execution

jeremenichelli:
It happened to me several times of having issues with global varible injections, sometimes this is really necesssary for packages that rely on browser globals like `window` or `document`, and I lost a lot of time until I realize that my _cleaning_ was doing things right but the concurrent nature of AVA was causing tests not running on _cleaned_ contexts.

I don't know if this is the best place or best approach to inform this but I found it nowhere and needed to crawl into old issues. I would have loved to have this in the setup section.

I also prefered to do this than just opening an issue, but you contributors, authors, members let me know if you agree with adding this comment ❤️

### DIFF
--- a/docs/recipes/test-setup.md
+++ b/docs/recipes/test-setup.md
@@ -20,6 +20,8 @@ You could do all these things using plain setup functions, but there are tradeof
 | ✅ &nbsp; failure has friendly output| ⛔️ &nbsp; errors are attributed to the test
 | ✅ &nbsp; corresponding `afterEach` and `afterEach.always` for cleanup| ⛔️ &nbsp; cannot easily clean up
 
+An important consideration if you are modifying or adding Node global variables (and cleaning them after each test), it might introduce undesired effects in global test contexts since tests are run concurrently by default. In this edge cases you can pass the `--serial` flag, this will ensure that any modification inside hooks like `beforeEach` are preparing the setup for each test correctly.
+
 ## Complex test setup
 
 In this example, we have both a `beforeEach()` hook, and then more modifications within each test.


jeremenichelli:
Hmm... shouldn't markdown files be ignored by codecov?

jeremenichelli:
About markdown putting down coverage, I just did the change directly through GitHub UI, so I?m pretty sure I was with the latest when I made the change.

About the _Writing tests_ section, I didn't see it! So thanks, it is actually a little bit more clear. 

The overall feeling I have is that this details is really buried down in the documentation. Because, first there's no mention of serial in the README, I would also expect this content more in the _Setup_ section than in the _Writing Tests_ one given the titles.

Would be good if this warning on patching global variables could be easier to reach than going throguh various documents and clicks. If you feel _Common pitfalls is that place_ I'll write a small piece in another PR and tag you. Thanks a lot for the feedback :)

jeremenichelli:
@novemberborn I opened a new one with the changes you recommended here: https://github.com/avajs/ava/pull/2211

Closing this one.

novemberborn:
> The overall feeling I have is that this details is really buried down in the documentation. Because, first there's no mention of serial in the README, I would also expect this content more in the _Setup_ section than in the _Writing Tests_ one given the titles.

Yes that's fair enough. We could do with a dedicated documentation website but I haven't been able to give that the necessary attention.

jeremenichelli:
Yeah, it was weird to me that such a cool project didn't have a website. The search indexation could also help discovering these kind of topics that are hard to find in this docs.

Let me know if there's anything I could do to help you folks set that up. I'm to have a discussion/call about it and see if I can contribute somehow to it :)

novemberborn:
See https://github.com/avajs/ava/pull/2051 — it's just a lot of stuff to deal with when you only have a couple hours at a time.

PR #2192, opened 2019-07-22
Consistent paths in watcher

bobthekingofegypt:
Fixes #2040 

Change the path handling for watcher, now all internal paths should be
absolute and chokidar paths will be converted to absolute before
handling.  This should fix the inconsistency that caused some path 
comparisons to fail resulting in behaviour like #2040.

* [ ] review how to test the actual fix.
* [ ] see if this also solves #2069

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

This pull request has been submitted to:
- [#2040: "previous failure in test files that were not rerun" not always true with ava --watch](https://oss.issuehunt.io/repos/26820798/issues/2040)
- [#2069: "Previous failure in test files that were not rerun" persists even though the failing test file was rerun](https://oss.issuehunt.io/repos/26820798/issues/2069)
---
</details>

PR #2206, opened 2019-08-09
fix #2205 duplicate runs of `.only` test cases

### LINKED ISSUE #2205: Issues with .only - tests double run and tests do not recover
## Description

I've been experiencing 2 issues with `.only` since 2.0.0 (or maybe earlier, I can't remember for sure). I believe the issues persisted in 2.1.0 and 2.2.0.

My typical workflow is:

1. Run tests in a specific file, e.g. `ava test/test-tiny-atom.js -vw`
2. Pick one test and add `.only`
3. Once I'm done with that test, remove .only

**Expected behaviour**:
1. All tests run in the file
2. Only the one tests with .only runs
3. All tests run in the file OR I can hit r+enter to rerun all.

**Actual behaviour**:
1. All tests run in the file
2. Only the one tests with .only runs the first time, and then that test starts running twice on each rerun.
3. No tests are run after removing .only, r+enter doesn't do anything.

### Test Source

Reproducable tests in the following repo: https://github.com/KidkArolis/tiny-atom. But I've been experiencing this issue with all of my repos using ava.

* See how `does not mutate the state object` test runs twice
* See how r+enter doesn't do anything

image

### Config

```json
  "ava": {
    "files": [
      "test/test-*.js"
    ],
    "require": [
      "@babel/register",
      "./test/helpers/setup.js"
    ]
  },
```

### Command-Line Arguments

```
ava test/test-tiny-atom.js -vw
```

## Environment

```
os darwin 18.7.0
node 10.15.3
ava 2.1.0
npm 6.10.0
```

vmlf01:
Fixes #2205

When watcher triggers re-run of test cases, it would re-run `.only` declared test case twice.
Seems to only happen when test file contains other test cases without `.only`

### DIFF
--- a/lib/watcher.js
+++ b/lib/watcher.js
@@ -229,7 +229,7 @@ class Watcher {
 				}
 
 				const fileStats = plan.status.stats.byFile.get(evt.testFile);
-				this.updateExclusivity(evt.testFile, fileStats.declaredTests > fileStats.selectedTests);
+				this.updateExclusivity(nodePath.relative(process.cwd(), evt.testFile), fileStats.declaredTests > fileStats.selectedTests);
 			});
 		});
 	}
--- a/test/fixture/watcher/with-only/test-only.js
+++ b/test/fixture/watcher/with-only/test-only.js
@@ -1,0 +1,9 @@
+import test from '../../../..';
+
+test.only('works', t => {
+	t.pass();
+});
+
+test('also works', t => {
+	t.pass();
+});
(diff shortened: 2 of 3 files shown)


novemberborn:
Hi @vmlf01 thanks for the PR!

A similar path normalization just landed in https://github.com/avajs/ava/pull/2192. I'm going to close this PR since I think it has the same effect. There's still a bug though, see my comment here https://github.com/avajs/ava/issues/2205#issuecomment-520237673

Hopefully you're keen on fixing that one 😄

PR #2208, opened 2019-08-11
Update fake-tty for changes in Node.js 12.7.0

novemberborn:
See the release notes for `tty`: https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V12.md#12.7.0:

We should really check if we can just use the built-in modules. I'm not sure why we're faking TTY anymore! 😄


novemberborn:
@sindresorhus mind double-checking this? Our tests don't actually cover the method invocations 😄

PR #2211, opened 2019-08-12
Add global pollution and serial case to common pitfalls

jeremenichelli:
As discussed in https://github.com/avajs/ava/pull/2204 I opened a PR adding global value resetting case and use of `serial` recommendation for these cases to _Common pitfalls_ docs.

In addition I added the list of pitfalls to the README so it's easier to discover from the homepage, to me it was really hard to get to inner sections from the README or trying to guess where the answer to my issue was.

Let me know what you folks think!


novemberborn:
> In addition I added the list of pitfalls to the README so it's easier to discover from the homepage, to me it was really hard to get to inner sections from the README or trying to guess where the answer to my issue was.

This will be hard to keep up-to-date. The section itself is listed and the idea is that you will click through to the dedicated document.

I've pushed some changes. Note that the problem is with asynchronous tests only. If everything is synchronous then it'll work fine. I've also tried to simplify the example to use a shared local variable, rather than a global variable. What do you think?

jeremenichelli:
Yeah, this looks good @novemberborn thanks for the corrections. I still feel the ToC would have been great on the README because it makes it easier to find or uncover, but I leave that decision to core maintainers :)

👍

PR #2213, opened 2019-08-15
fix the “subprocess.kill"

xiejunfeng:
The “subprocess.kill" only sends a termination signal to child process, but may not actually terminate the process. In the example below, because puppeteer was using child processes, the “subprocess.kill" triggered by timeout will actually not terminate the child process.
```js
import test from 'ava';
import delay from 'delay';
const puppeteer = require('puppeteer');

// the test timeout config is 2000
test.beforeEach(async t => {
  t.context.browser = await puppeteer.launch();
});

test('foo', async t => {
  const browser = t.context.browser;
  const page = await browser.newPage();
  await page.goto('https://xx.xx.com/teach', {timeout: 3000});
  await page.screenshot({path: 'example.png'});
  await t.context.browser.close();
  // t.pass();
});
````

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ node_modules
 .nyc_output
 coverage
 bench/.results
+.idea/*
--- a/lib/fork.js
+++ b/lib/fork.js
@@ -119,10 +119,25 @@ module.exports = (file, opts, execArgv) => {
 		});
 	});
 
+	function killProcess() {
+		if (subprocess.pid && !subprocess.killed && !forcedExit) {
+			// Force kill the subprocess.
+			try {
+				if (process.platform === 'win32') {
+					childProcess.execSync(`taskkill /pid ${subprocess.pid} /T /F`);
+				} else {
+					process.kill(subprocess.pid, 'SIGKILL');
+				}
+				forcedExit = true;
+			} catch (e) {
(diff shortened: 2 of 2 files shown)


novemberborn:
This is quite the workaround. I'm not sure this is AVA's responsibility. I noticed this issue https://github.com/GoogleChrome/puppeteer/issues/1825 so maybe it's something Puppeteer can fix?

@ehmicky @sindresorhus any insights you have given your experience with `execa`?

sindresorhus:
> This is quite the workaround. I'm not sure this is AVA's responsibility. I noticed this issue GoogleChrome/puppeteer#1825 so maybe it's something Puppeteer can fix?

I actually think this is something Node.js should fix. It's quite silly that there's no reliable way to kill processes. Someone should open an issue on Node.js about this.

> any insights you have given your experience with execa?

We have an option for this in `execa`: https://github.com/sindresorhus/execa#optionsforcekillaftertimeout

novemberborn:
> We have an option for this in `execa`: [sindresorhus/execa#optionsforcekillaftertimeout](https://github.com/sindresorhus/execa#optionsforcekillaftertimeout)

Maybe we should start using `execa` in `fork.js` then.

novemberborn:

PR #2214, opened 2019-08-16
bring logo to center in readme

gurrrung:


novemberborn:
We have a PR open for this already in #2200. Perhaps you could join the discussion there on making this work on npm?

PR #2222, opened 2019-08-21
Explain how to use custom HTTP clients when endpoint testing

tryzniak:
Some people may want to use a more familiar client to them like fetch, got and etc when endpoint testing. Let's explain how to do that!


novemberborn:
Yes that's fair enough. I wonder if we should shake up this recipe a bit more. E.g. explain how to start a server in a `.before()` hook, use `test-listen` or `get-port` and share the URL via `t.context`. Then we can show an example that uses `got()` and AVA's assertions and link to tools like `supertest` at the end of the recipe.

What do you think @tryzniak?

novemberborn:
@tryzniak I've made some edits, what do you think?

tryzniak:
> @tryzniak I've made some edits, what do you think?

I love them :) Don't know if I can anything to it :)

novemberborn:
@tryzniak cool, merged!

PR #2223, opened 2019-08-23
Pass test's args to beforeEach/afterEach hooks

### LINKED ISSUE #2070: Feature request: pass test additional arguments to beforeEach function
[Issuehunt badges](https://issuehunt.io/r/avajs/ava/issues/2070)

Would be cool if we had access in `beforeEach` hook to additional arguments that can be passed to each test. Kinda like it's demonstrated in [macros](https://github.com/avajs/ava/blob/master/docs/01-writing-tests.md#reusing-test-logic-through-macros) docs. This would allow `beforeEach` hook parametrisation. Like in the example below:

```js

test.beforeEach(async (t, args) => {
  if (args.initDb) {
    // initialise db
  }
});

test('my test with args', async t => {

}, {
  initDb: true
});

```

> There is a $40.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/avajs/ava/issues/2070).

tryzniak:
Now we can do stuff below:
```js
test.beforeEach(async (t, args) => {
  if (args.initDb) {
    // initialise db
  }
});

test.afterEach(async (t, args) => {
  if (args.initDb) {
    // close db connection
  }
})

test('my test with args', async t => {

}, {
  initDb: true
});
```
I think this can simplify a lot of setup/teardown code flows.

Closes #2070

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -266,13 +266,11 @@ class Runner extends Emittery {
 		return result;
 	}
 
-	async runHooks(tasks, contextRef, titleSuffix) {
+	async runHooks(tasks, contextRef, testArgs = [], titleSuffix) {
 		const hooks = tasks.map(task => new Runnable({
 			contextRef,
 			failWithoutAssertions: false,
-			fn: task.args.length === 0 ?
-				task.implementation :
-				t => task.implementation.apply(null, [t].concat(task.args)),
+			fn: t => task.implementation.call(null, t, ...task.args, ...testArgs),
 			compareTestSnapshot: this.boundCompareTestSnapshot,
 			updateSnapshots: this.updateSnapshots,
 			metadata: task.metadata,
@@ -303,9 +301,14 @@ class Runner extends Emittery {
 
 	async runTest(task, contextRef) {
 		let hooksAndTestOk = false;
-
 		const hookSuffix = ` for ${task.title}`;
-		if (await this.runHooks(this.tasks.beforeEach, contextRef, hookSuffix)) {
+		let testArgs = task.args || [];
+		// Is it a macro? Then skip its args
+		if (task.implementation && task.implementation.title) {
+			testArgs = testArgs.slice(2);
+		}
+
+		if (await this.runHooks(this.tasks.beforeEach, contextRef, testArgs, hookSuffix)) {
 			// Only run the test if all `beforeEach` hooks passed.
 			const test = new Runnable({
 				contextRef,
@@ -328,7 +331,7 @@ class Runner extends Emittery {
 					knownFailing: result.metadata.failing,
 					logs: result.logs
 				});
-				hooksAndTestOk = await this.runHooks(this.tasks.afterEach, contextRef, hookSuffix);
+				hooksAndTestOk = await this.runHooks(this.tasks.afterEach, contextRef, testArgs, hookSuffix);
 			} else {
 				this.emit('stateChange', {
 					type: 'test-failed',
@@ -342,7 +345,7 @@ class Runner extends Emittery {
 			}
 		}
 
-		const alwaysOk = await this.runHooks(this.tasks.afterEachAlways, contextRef, hookSuffix);
+		const alwaysOk = await this.runHooks(this.tasks.afterEachAlways, contextRef, testArgs, hookSuffix);
 		return hooksAndTestOk && alwaysOk;
 	}
 
--- a/test/runner.js
+++ b/test/runner.js
@@ -592,6 +592,66 @@ test('macros: Additional args will be spread as additional args on implementatio
 	});
 });
 
+test('Test\'s additional args will be spread and passed down to beforeEach, afterEach implementation functions', t => {
+	t.plan(5);
+
+	return promiseEnd(new Runner(), runner => {
+		runner.on('stateChange', evt => {
+			if (evt.type === 'test-passed') {
+				t.pass();
+			}
+		});
+
+		runner.chain.beforeEach((a, ...rest) => {
+			t.deepEqual(rest, ['foo', 'bar']);
+			a.pass();
+		});
+		runner.chain.afterEach((a, ...rest) => {
+			t.deepEqual(rest, ['foo', 'bar']);
+			a.pass();
+		});
+		runner.chain.afterEach.always((a, ...rest) => {
+			t.deepEqual(rest, ['foo', 'bar']);
+			a.pass();
+		});
+
+		runner.chain('test1', (a, ...rest) => {
+			t.deepEqual(rest, ['foo', 'bar']);
+			a.pass();
+		}, 'foo', 'bar');
+	});
+});
+
+test('Test\'s additional args will be concatenated with args of beforeEach, afterEach args', t => {
+	t.plan(5);
+
+	return promiseEnd(new Runner(), runner => {
+		runner.on('stateChange', evt => {
+			if (evt.type === 'test-passed') {
+				t.pass();
+			}
+		});
+
+		runner.chain.beforeEach((a, ...rest) => {
+			t.deepEqual(rest, ['baz', 'foo', 'bar']);
+			a.pass();
+		}, 'baz');
+		runner.chain.afterEach((a, ...rest) => {
+			t.deepEqual(rest, ['baz', 'foo', 'bar']);
+			a.pass();
+		}, 'baz');
+		runner.chain.afterEach.always((a, ...rest) => {
+			t.deepEqual(rest, ['baz', 'foo', 'bar']);
+			a.pass();
+		}, 'baz');
+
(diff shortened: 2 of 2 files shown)


lib/runner.js
```diff
@@ -266,13 +266,13 @@ class Runner extends Emittery {
...
-	async runHooks(tasks, contextRef, titleSuffix) {
+	async runHooks(tasks, contextRef, testArgs = [], titleSuffix) {
 		const hooks = tasks.map(task => new Runnable({
 			contextRef,
 			failWithoutAssertions: false,
 			fn: task.args.length === 0 ?
-				task.implementation :
-				t => task.implementation.apply(null, [t].concat(task.args)),
+				t => task.implementation.apply(null, [t].concat(testArgs)) :
+				t => task.implementation.apply(null, [t].concat(task.args).concat(testArgs)),
```
novemberborn: We should clean this up and use rest syntax: `task.implementation.call(null, t, ...testArgs)`.

lib/runner.js
```diff
@@ -303,9 +303,8 @@ class Runner extends Emittery {
 
 	async runTest(task, contextRef) {
 		let hooksAndTestOk = false;
-
 		const hookSuffix = ` for ${task.title}`;
-		if (await this.runHooks(this.tasks.beforeEach, contextRef, hookSuffix)) {
+		if (await this.runHooks(this.tasks.beforeEach, contextRef, task.args, hookSuffix)) {
```
novemberborn: I think we should pass the `task.implementation` as well. That way, the hook can identify the macro that's being used.

tryzniak:
I've made some changes. Could not figure out where we should change typescript signatures though. 
Other than that now we don't pass macros related args to beforeEach/afterEach hooks:
```js
test.beforeEach(async (t, ...args) => {
  t.deepEqual(args, [{shouldLog: true}, {initDB: true}]);
}, {shouldLog: true});

function macro1(t, input, expected) {
  t.is(eval(input), expected);
}

macro1.title = (providedTitle = '', input, expected) => `${providedTitle} ${input} = ${expected}`.trim();

test(macro1, '2 + 2', 4, {initDB: true});
test(macro1, '2 * 3', 6, {initDB: true});
test('providedTitle', macro1, '3 * 3', 9, {initDB: true});

test('some test 1', async t => {
  t.is(1, 1);
}, {initDB: true});


```

lib/runner.js
```diff
@@ -303,9 +301,14 @@ class Runner extends Emittery {
 
 	async runTest(task, contextRef) {
 		let hooksAndTestOk = false;
-
 		const hookSuffix = ` for ${task.title}`;
-		if (await this.runHooks(this.tasks.beforeEach, contextRef, hookSuffix)) {
+		let testArgs = task.args || [];
+		// Is it a macro? Then skip its args
+		if (task.implementation && task.implementation.title) {
```
novemberborn: Macros are not required to implement `.title`. They're not really distinguishable from regular test implementations.

Gwenio:
> I think this needs to some change in the type definition as well. Unfortunately we won't be able to infer the argument types, but we should be able to type them as unknown[].

If it was a single object as an argument, then it should be possible assigned a type similar to how the test context is currently.

So:

```typescript
// example test file with typed hook argument
import anyTest, { TestInterface } from 'ava'

interface Context {
	/* whatever. */
}

interface HookOptions {
	/* whatever. */
}

const test = anyTest as TestInterface<Context, HookOptions>
// currently the above const would be:
// const test = anyTest as TestInterface<Context>
// for people upgrading from older versions, TSC will flag the need for a change with a compile error

test.beforeEach(async (t, options /* is type HookOptions */): Promise<void> => {
  /* hook body */
}

test('testing', async (t): Promise<void> => {
  /* test body */
}, {
  /* hook options */
})

test('another', async (t): Promise<void> => {
  /* test body */
}) // the hook options defaults to an empty object
// or TS has a compile error if HookOptions is not compatible with an empty object
```

In such a case if the basic untyped test is used, the hook options type would be:

```typescript
// theoretical default hook options type
interface HookOptions {
  [key: string]: unknown
}
```

Doing the hook arguments this way would work fine for regular Javascript tests as well, they can use destructuring to get the options a particular hook cares about.

If the worry is about being more verbose, for test files with lots of hook options using an object will make setting up defaults easier.

```javascript
const defaultHookOptions = {
/* most common values *.
}

test('test', async (t) => {/* test body */}, {
  ...defaultHookOptions,
  /* overrides specific to the test */
}
```

With a rest parameter, overriding the defaults for individual tests would be harder as it would be based on position.

In test files with few tests there is likely to also be few hook options. In files with many tests it is likely that for any particular option most tests will set it to the same value.

novemberborn:
I'm struggling a little with the type definition implications. But we can ship this without improving those. I've just landed [support for experimental features](https://github.com/avajs/ava/pull/2218) which we could use to make this an opt-in behavior until we figure out everything else.

There's this comment left as an outstanding issue though: https://github.com/avajs/ava/pull/2223#discussion_r319752987

lib/runner.js
tryzniak: I think I tried to nail down the very last argument of a test (`{initDB: true} in the example`) and only pass it to the hook, since stuff like `'3 * 3', 9` won't be useful. Also document it and explain users that the last argument is always passed to the beforeEach/afterEach hooks.
```
test('providedTitle', macro1, '3 * 3', 9, {initDB: true});
```

But due to inability to really detect when something is a macro I'm puzzled for now. 
And with my implementation args are and array, need to think about @Gwenio response.

Gwenio: > I think I tried to nail down the very last argument of a test (`{initDB: true} in the example`) and only pass it to the hook, since stuff like `'3 * 3', 9` won't be useful. Also document it and explain users that the last argument is always passed to the beforeEach/afterEach hooks.
> 
> ```
> test('providedTitle', macro1, '3 * 3', 9, {initDB: true});
> ```
> 
> But due to inability to really detect when something is a macro I'm puzzled for now.
> And with my implementation args are and array, need to think about @Gwenio response.

If both hook args and macro args can both be of variable length then finding where to split them is not possible.
This means to support macros with rest or optional parameters that the hook arg count needs to be fixed.
Having them be fixed not only excludes a rest parameter, but also means the hook arg(s) cannot be optional.
And requiring a hook arg would be an interface breaking change.

To avoid this the hook arg would need to come before the test implementation(s).
Then it could be extracted when getting the optional base title and the implementation(s).
It would also need the type of the arg limited such that it is distinct from the implementation(s) (and the title, if it can come before that too).

The relevant part of the runner for extracting title and implementation(s) is line 60 to 65 of runner.js.
The hook arg should then be added to the task record when it is created on line 129.

novemberborn:
@tryzniak @Gwenio what do you think of adding these additional arguments to the `t` object instead? Wouldn't necessary make it easier to type but it may reduce some confusion when hooks are called with additional arguments.

Gwenio:
> @tryzniak @Gwenio what do you think of adding these additional arguments to the `t` object instead? Wouldn't necessary make it easier to type but it may reduce some confusion when hooks are called with additional arguments.

The question is about how to identify the the value(s) intended for hooks from those passed to `test(...)`.
Once they are found, there is no trouble passing them to hooks.
The fact that there are so many optional parameters already make identification tricky.

For AVA 2.4.0, this mean changing 'lib/parse-test-args.js' so that its exported function includes the hook arg(s) in the return value.

So

```javascript
function parseTestArgs(args) {
	const rawTitle = typeof args[0] === 'string' ? args.shift() : undefined;
	const receivedImplementationArray = Array.isArray(args[0]);
	const implementations = receivedImplementationArray ? args.shift() : args.splice(0, 1);

	const buildTitle = implementation => {
		const title = implementation.title ? implementation.title(rawTitle, ...args) : rawTitle;
		return {title, isSet: typeof title !== 'undefined', isValid: typeof title === 'string', isEmpty: !title};
	};

	return {args, buildTitle, implementations, rawTitle, receivedImplementationArray};
}
```

would need to become something like the following

```javascript
function parseTestArgs(args) {
	const rawTitle = typeof args[0] === 'string' ? args.shift() : undefined;
	const hookArgs = /* check for hook args, and if found remove from 'args' and assign them */ ;
	const receivedImplementationArray = Array.isArray(args[0]);
	const implementations = receivedImplementationArray ? args.shift() : args.splice(0, 1);

	const buildTitle = implementation => {
		const title = implementation.title ? implementation.title(rawTitle, ...args) : rawTitle;
		return {title, isSet: typeof title !== 'undefined', isValid: typeof title === 'string', isEmpty: !title};
	};

	return {args, buildTitle, hookArgs, implementations, rawTitle, receivedImplementationArray};
}
```

Then it is just a matter of adding `hookArgs` to the task object passed to `Runner.runTest` which would then simply need to pass it to `Runner.runHooks`.

novemberborn:

PR #2224, opened 2019-08-23
Minor code refactor

jamesgeorge007:
Refactor code to result in a more concise version.


novemberborn:
Hi @jamesgeorge007. Thanks for the PR. I'd rather we stick to the old style though, especially in if-conditions. Hopefully you'll find something else to contribute to here 😄

PR #2226, opened 2019-08-25
Gitignore vim's swp files

tryzniak:
Let's ignore *.swp files so we don't commit them accidentally. Those files are usually created by vim.


sindresorhus:
That’s not project specific and should be in your own global gitignore file, not here.

Issue #1277, opened 2017-02-20
Support magic-assert with other assertion libraries


novemberborn:
It'd be great if AVA could recognize assertion errors from other libraries and show similar output as it does for AVA's own assertions. See https://github.com/avajs/ava/issues/1256#issuecomment-280407908.

danny-andrews:
I second this. I'm using the fantastic [expect](https://github.com/mjackson/expect) library, and I get unhappy error messages like the following:
```
  /Users/danny/Projects/circleci-weigh-in/node_modules/expect/lib/assert.js:29

  Error thrown in test:

  Error {
    message: 'Expected { threshold: { targets: \'yoyo.js\', maxSize: 2, strategy: \'any\' }, offendingAssets: [ \'yoyo.js\' ] } to match { message: \'"yoyo.js" (35B) must be less than or equal to 2B!\', threshold: { targets: \'yoyo.js\', maxSize: 2, strategy: \'any\' }, offendingAssets: [ \'yoyo.js\' ] }',
  }

  assert (node_modules/expect/lib/assert.js:29:9)
  Expectation.toMatch (node_modules/expect/lib/Expectation.js:138:28)
  _ramda2.default.forEach (tests/bundle-size-utils.test.js:115:33)
  forEach (node_modules/ramda/src/forEach.js:43:5)
  Object.f2 [as forEach] (node_modules/ramda/src/internal/_curry2.js:25:16)
  Test.fn (tests/bundle-size-utils.test.js:113:9)
```
when using `expect.toMatch`

tryzniak:
Are we still after this one? I'd like to tackle it, but need some hints :)

tryzniak:
Took a look at it and made some changes. What do you think about generation output as below?
New version:
<img width="493" alt="Screen Shot 2019-08-21 at 19 04 36" src="https://user-images.githubusercontent.com/10567153/63448767-5340f980-c447-11e9-812f-5062a1cc1ee2.png">

Old version:
<img width="563" alt="Screen Shot 2019-08-21 at 19 08 52" src="https://user-images.githubusercontent.com/10567153/63448769-550abd00-c447-11e9-91d1-7ba6b36da32b.png">

novemberborn:
I think you're on to something here @tryzniak. Would you mind opening a PR for this?

tryzniak:
Sure! I'll do soon!

PR #2230, opened 2019-08-26
WIP: Format failed external asserts nicer

tryzniak:
Hello. This for #1277 . I'd love to show you what I've been up to, here are the new output
image
[testfile](https://gist.github.com/tryzniak/56ab30808be43db4891aa67f9275160f)

I don't really like the way `supertest` errors are shown, I have no idea how we could detect what line of the test its expect failed.
Other errors have inconsistent formatting, which is not very nice.
I'm thinking of capturing and detecting a type of an error earlier.
Any input would be useful for me! Thanks


lib/test.js
```diff
@@ -8,6 +8,14 @@ const assert = require('./assert');
 const nowAndTimers = require('./now-and-timers');
 const concordanceOptions = require('./concordance-options').default;
 
+function isExternalAssertError(error) {
+	if (error.matcherResult) {
+		return error.matcherResult.actual && error.matcherResult.expected;
+	}
+
+	return error.actual && error.expected;
```
novemberborn: Could you add comments clarifying which errors from which frameworks you're looking for here?

tryzniak: Absolutely!

tryzniak:
> Oh that's an interesting approach. You're looking at the thrown error to see if it may have come from an assertion library? And then the hope is that its `message` is useful?

Yeah! Judging by a quick look, they usually have an already formatted `message`.

novemberborn:
> > Oh that's an interesting approach. You're looking at the thrown error to see if it may have come from an assertion library? And then the hope is that its `message` is useful?
> 
> Yeah! Judging by a quick look, they usually have an already formatted `message`.

Sounds great!

lib/test.js
novemberborn: Nah those can be added in a future PR if necessary.

novemberborn:

PR #2238, opened 2019-09-04
Report failed hooks to tap

### LINKED ISSUE #2237: Test should be reported as failed if hook failed
## Description

The tap reporter swallows failures which occur within the hooks:
https://github.com/avajs/ava/blob/v2.3.0/lib/reporters/tap.js#L144

The failure is being recorded as a comment, which will be ignored by a tap parser.  This means that after parsing it the run will be considered a pass if there was a problem in the hooks that prevented a test from running.

In our use case tests need to setup accounts before executing.  When a PR is made and the tests run, we don't want the test to pass if it can't be executed, it should be marked as a failure for us to look into.

### Test Source

```js
test.serial.beforeEach('Auto failure in hook', () => {throw new Error('Simulated failure');});
```

```
# Failed hook: Setup accounts for The test name
```

### Config

```json
{
  "ava": {
        "compileEnhancements": false,
        "extensions": [
            "ts"
        ],
        "require": [
            "ts-node/register"
        ]
    }
}
```

### Command-Line Arguments

```
$(npm bin)/ava --tap --timeout=3m -c 8 -t dist/test/*.js
```

## Relevant Links

- If your project is public, link to the repo so we can investigate directly.
- **BONUS POINTS:** Create a [minimal reproduction](http://stackoverflow.com/help/mcve) and upload it to GitHub. This will get you the fastest support.

## Environment

```
local: darwin 18.6.0
Same issue in other linux based environments
ava --version
v2.3.0
npm --version
6.4.1
```

nicholas-eden:
When a test fails in the hooks we don't want the test to be marked as passing for a tap parser.
This fixes #2237

### DIFF
--- a/lib/reporters/tap.js
+++ b/lib/reporters/tap.js
@@ -142,7 +142,7 @@ class TapReporter {
 				// Ignore
 				break;
 			case 'hook-failed':
-				this.writeComment(evt, {error: true});
+				this.writeTest(evt, {passed: false, todo: false, skip: false});
 				break;
 			case 'hook-finished':
 				this.writeComment(evt, {});


nicholas-eden:
Closing this as it isn't the ideal solution.  It looks like tests would be reported as passed then subsequently failed.

PR #2255, opened 2019-09-20
Document module path mapping in TypeScript recipe

tymfear:
`ts-node` has issues with usage of the path aliases. The only solution, that I've found so far is to use the [tsconfig-paths](https://github.com/dividab/tsconfig-paths#readme), which is dedicated for address this problem. So decided to add another typescript recipe


sindresorhus:
> ts-node has issues with usage of the path aliases.

Can you include links to the relevant `ts-node` issues in the recipe so users can let us know when it's fixed?

tymfear:
@sindresorhus , they are not going to fix it. Added a note to the recipe.

novemberborn:
Thanks @tymfear!

LucasIcarus:
@novemberborn 

Recently create a project with ava, ts-node and tsconfig-paths..

Introduction of tsconfig-paths will destroy dependency tracking function within ava's watch mode.

Any further hint?

PR #2272, opened 2019-10-17
Allow Node arguments to be configured

### LINKED ISSUE #2090: Add a `--node-arguments` flags
[Issuehunt badges](https://issuehunt.io/r/avajs/ava/issues/2090)

## Description

It could be useful to be able to de define the flags of the Node process running AVA. [A comment](https://github.com/avajs/ava/issues/1810#issuecomment-394367019) by @novemberborn scratch the surface, but apparently it never got raised as a request.
Any compatible Node flag could be listed.

Adding the ~~`--nodeArgs``--node-args`~~`--node-arguments` to AVA could allow the user to execute Node with some needed flags.
(eg: `ava --node-arguments --experimental-report`)
It should also be available as an AVA config property:
```json
{
  "ava": {
    "nodeArguments": [
      "--experimental-report",
      "- --custom-arg"
    ]
  }
}
```

### Test Source

```js
process.execArgv.includes("--experimental-policy"); // <= Never true
```

## Environment

 - Ubuntu v16.04
 - Node v11.9.0
 - AVA v1.4.1

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $60.00)

### Tips

---
</details>

maximelkin:
Fixes #2090

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

</details>


novemberborn:
Hey @maximelkin sorry about the delay in reviewing. At first glance this looks great, though I'd like to play with it a little first.

lib/api.js
```diff
@@ -311,7 +311,7 @@ class Api extends Emittery {
 	}
 
 	async _computeForkExecArgv() {
-		const execArgv = this.options.testOnlyExecArgv || process.execArgv;
+		const execArgv = this.options.nodeArguments;
```
novemberborn: I think `testOnlyExecArgv` was used by AVA's internal tests, though I don't see a reference to it anymore. So removing that is fine.

`process.execArgv` would inherit the arguments used when starting AVA itself. I think we should concatenate that with `nodeArguments` more clearly *here*, rather than through a separate module.

lib/node-arguments.js
```diff
@@ -0,0 +1,30 @@
...
+
+	const detectedInspect = parsed.find(arg => /^--inspect(-brk)?(=|$)/.test(arg));
+
+	if (detectedInspect && detectedInspect.includes('=')) {
+		throw new Error('The \'nodeArguments\' configuration must not contain inspect with port.');
+	}
+
+	const mainProcessArgs = process.execArgv.filter(arg => !detectedInspect || !/^--inspect(-brk)?(=|$)/.test(arg));
+
+	return parsed.concat(mainProcessArgs);
```
novemberborn: `mainProcessArgs` should be overridden by the arguments, not the other way around.

lib/node-arguments.js
```diff
@@ -0,0 +1,30 @@
...
+function normalizeNodeArguments(nodeArguments) {
+	const parsed = Array.isArray(nodeArguments) ? nodeArguments : parseCliParameter(nodeArguments);
+
+	const detectedInspect = parsed.find(arg => /^--inspect(-brk)?(=|$)/.test(arg));
+
+	if (detectedInspect && detectedInspect.includes('=')) {
+		throw new Error('The \'nodeArguments\' configuration must not contain inspect with port.');
+	}
+
+	const mainProcessArgs = process.execArgv.filter(arg => !detectedInspect || !/^--inspect(-brk)?(=|$)/.test(arg));
```
novemberborn: I'm not sure it's such a problem if these flags are provided twice, or with a port.

lib/node-arguments.js
```diff
@@ -0,0 +1,30 @@
...
+// --param=" --arg1 --arg2 param " --abc
+//         |_this part___________|
+function parseCliParameter(str) {
+	str = str[0].startsWith('"') && str.endsWith('"') ? str.slice(1, str.length - 1) : str;
+	return trim(str)
+		.split(' ');
+}
+
+function normalizeNodeArguments(nodeArguments) {
+	const parsed = Array.isArray(nodeArguments) ? nodeArguments : parseCliParameter(nodeArguments);
```
novemberborn: If the value comes from a configuration file, we should enforce that it's an array.

When it comes from the CLI, begrudgingly, I agree we need to parse it. Is there an npm package that can take care of this for use?

docs/05-command-line.md
```diff
@@ -21,6 +21,7 @@ $ npx ava --help
     --color                 Force color output
     --no-color              Disable color output
     --reset-cache           Reset AVA's compilation cache and exit
+    --node-arguments        Configure Node.js arguments used to launch worker processes
```
novemberborn: We should include sample usage before, given that you need to provide the arguments as a single string.

lib/node-arguments.js
maximelkin: This is problem, because ava run multiple workers, and they cannot use same port

novemberborn: Oh I see, fair point. My hunch is to not correct for this, for now. We should have a better solution for debugging regardless.

maximelkin: If this not needed, we can not parse args, and just concatenate from process.execArgv and config

novemberborn: Exactly, much easier!

Although we also have a way of providing the flags to the AVA CLI. So we still need some parsing I suppose.

maximelkin:
Not sure, why tests getting SIGKILL on local machine

lib/api.js
```diff
@@ -15,6 +16,8 @@ const makeDir = require('make-dir');
 const ms = require('ms');
 const chunkd = require('chunkd');
 const Emittery = require('emittery');
+const minimist = require('minimist');
```
novemberborn: Could we use https://www.npmjs.com/package/yargs-parser instead? It's already in our dependency tree as part of `meow`.

(Though it looks like so is `minimist`, because Node.js modules, but for less apparent reasons.)

lib/api.js
```diff
@@ -320,28 +323,28 @@ class Api extends Emittery {
...
+		const args = {
+			...mainProcessArgs,
+			...configExecArgs
+		};
 
-		forkExecArgv[inspectArgIndex] = `${flagName}=${port}`;
+		const hasInspect = args.inspect || args['inspect-brk'];
+		if (hasInspect) {
+			if (args.inspect && args['inspect-brk']) {
+				throw new Error('Flags inspect and inspect-brk provided simultaneously');
```
novemberborn: Note to self: check how this ends up being reported.

lib/cli.js
```diff
@@ -115,6 +116,10 @@ exports.run = async () => { // eslint-disable-line complexity
 				type: 'boolean',
 				default: false
 			},
+			'node-arguments': {
+				type: 'string',
+				default: ''
```
novemberborn: Can we remove this default? See my comment for `normalizeNodeArguments()`.

lib/node-arguments.js
```diff
@@ -0,0 +1,18 @@
...
+
+const minimist = require('minimist');
+const omit = require('lodash/omit');
+
+/**
+ * @param {string[]} confParams
+ * @param {string} cliParams
+ * @return {object}
+ */
+function normalizeNodeArguments(confParams, cliParams) {
```
novemberborn: If we can assume the arguments may be `undefined`, we can use default parameter values:

```suggestion
function normalizeNodeArguments(confParams = [], cliParams = '') {
```

And we can get rid of the JSDoc comment too.

We do need to enforce `confParams` is an array.

lib/node-arguments.js
```diff
@@ -0,0 +1,18 @@
...
+	return omit({
+		...minimist(confParams || []),
+		...minimist(cliParams.split(' '))
```
novemberborn: Are there Node.js arguments that take strings? Cause then splitting on the empty string won't work.

lib/node-arguments.js
```diff
@@ -0,0 +1,18 @@
...
+	}, '_');
```
novemberborn: Is the `_` returned by `minimist`? I'd prefer rest-spread for this:

```js
const { _, ...normalized } = { ...minimist(), ...minimist() };
return normalized;
```

lib/api.js
```diff
@@ -320,28 +323,28 @@ class Api extends Emittery {
...
-			return Promise.resolve(execArgv);
-		}
+		const {nodeArguments: configExecArgs} = this.options;
 
-		// --inspect-brk is used in addition to --inspect to break on first line and wait
-		const inspectArgIndex = execArgv.findIndex(arg => /^--inspect(-brk)?($|=)/.test(arg));
-		if (inspectArgIndex === -1) {
-			return Promise.resolve(execArgv);
-		}
+		const mainProcessArgs = minimist(process.execArgv);
```
novemberborn: Could we cache this? This function runs for every test file.

lib/api.js
maximelkin: I choose minimist, because it has reverse package, maybe there is another package for it?

novemberborn: I reckon `dargs` may work with `yargs-parser` too, assuming it also outputs straight-forward objects.

niftylettuce:
Great work!  Will be so nice to have since Node v12 changed TLS min version and we need `--tls-min-v1` flag as a Node arg for a project of mine.

lib/api.js
maximelkin: Not sure, what I correctly understand you, check please

novemberborn:
Hi @maximelkin, I did not have time to look at your changes yet, sorry.

maximelkin:
About passing arguments via cli
Maybe it's better to pass as --node-arguments "a b c"? 
As I see this is supported by yargs-parser (and meow) now

novemberborn:
> About passing arguments via cli
> Maybe it's better to pass as --node-arguments "a b c"?
> As I see this is supported by yargs-parser (and meow) now

@maximelkin what do you mean by this?

---

I spotted the change of `-r` to `--require` in the CLI helper. This led me of on a chase as to why `-r` wasn't working. It's because with the default configuration, `dargs` outputs `-r=foo.js` which Node.js does not support.

This then led me to think about how we're combining the arguments. I think we should simplify it a little:

* Take `process.execArgv`
* Concatenate with the `nodeArguments` array from the AVA config
* Concatenate with the value of `--node-arguments` passed to AVA on the CLI
* Provide this combined array as the `execArgv` option when forking the child process

This without parsing into objects, though we do need to parse the value to the `--node-arguments` option into the proper array.

I guess what may complicate this is the dynamic port assignment when we see `--inspect` flags. I'm hoping to do a bunch of work on AVA over the next two weeks, so maybe I can get rid of that too.

novemberborn:
I've switched the CLI flag parser to `yargs` (#2315) which will cause some conflicts here. But, #2316 adds a built-in debug command, which simplifies the `execArgv` stuff in `api.js`.

novemberborn:
@maximelkin I've been making a lot of changes over the past few weeks which has caused a lot of merge conflicts here. Sorry about that. I can try and resolve them when I get some time.

With regards to parsing `--node-arguments`, I was thinking we could maybe launch a script as a child process which outputs `console.log(JSON.stringify(process.execArgv))`. That way we can use Node.js own argument parsing logic.

novemberborn:
@maximelkin I've fixed the merge conflicts and simplified the implementation as per https://github.com/avajs/ava/pull/2272#issuecomment-568180065. Please let me know what you think!

novemberborn:
I was afraid this wouldn't work on Windows though 😢

maximelkin:
Hello, sorry for long inactivity
All is ok, very interesting hack with calling node for parsing
Also, maybe ask nodejs  to expose algorithm of doing this? (for better implementation in future)

maximelkin:
So adding `shell: true` to spawn test makes it works, but breaks watcher
Watcher breaking seems to be suspicious for me, I will research this

novemberborn:
> All is ok, very interesting hack with calling node for parsing
> Also, maybe ask nodejs to expose algorithm of doing this? (for better implementation in future)

I found https://github.com/astur/arrgv which claims to work just like Node.js does. Looking at the implementation in Node.js itself it also does a bunch of validation, which could lead to unexpected errors. So a simple parser is better. And this one gives us arrays which is what we want. Plus, it's tested using AVA 😄 

Let's see if CI passes and then let's ship this!

Issue #2257, opened 2019-09-24
Watch doesn't work 


asolopovas:
## Description

Hi, when I run ava with watch parameter it runs the first time but does not rerun if I change anything in the files. I have setup test repository https://github.com/asolopovas/ava-test

## Environment

Tell us which operating system you are using, as well as which versions of Node.js, npm, and AVA. Run the following to get it quickly:

node -e "var os=require('os');console.log('Node.js ' + process.version + '\n' + os.platform() + ' ' + os.release())"
```
Node.js v12.10.0
win32 10.0.18362
```
ava --version
```
6.10.3
```
npm --version

```
2.4.0
```

novemberborn:
Have a look at https://github.com/avajs/ava/blob/master/docs/recipes/watch-mode.md#debugging which should help you see the events that AVA is detecting.

(I'm closing this issue for housekeeping purposes, but let's keep the conversation going.)

asolopovas:
Funny enough it does not any events. Don't really understand why even after 
```
$ set DEBUG=ava:watcher
$ npm test -- --watch --verbose
```
It just simply runs the test once and does not do anything on change. I tested it on linux it works without a hitch but on windows, I can't seem to get it working.

asolopovas:
@novemberborn I can't get it to work in windows it has very strange behaviour if I create ava.config.js and edit this one during watch task running it does restart tests, if I edit files in src or tests it doesn't reload the tests.  Is there anything else I can try doing to see why it doesn't work?

asolopovas:
@novemberborn I tested it on two different computers and behaviour is the same in both instances when its run in windows.

asolopovas:
It should be reopened as its definitely a bug

asolopovas:
I have also tested chokidar and it works without problems reports all the events. Something is wrong with avajs

PR #2280, opened 2019-10-28
Update chokidar to version 3.2.1

asolopovas:
Fixes problem where watch not working with Windows 10 https://github.com/avajs/ava/issues/2257


novemberborn:
`3.2.1` is within the dependency range of `^3.0.2`. We'll update the minimum version as part of the release process.

For now, if you have a lockfile in your project, remove AVA and reinstall it. That should upgrade `chokidar` for you. Or if you don't have a lockfile, removing `node_modules` and running `npm install` should do the trick.

PR #2294, opened 2019-11-14
Correcting casing in endpoint testing example

alexdrans:


novemberborn:
Correct is in the eye of the beholder, but I do agree with the change. Thanks!

PR #2297, opened 2019-11-20
Remove array-union and array-uniq dependencies

jimmywarting:
I inlined what both array-union and array-uniq basically do so we don't have to depend on them.


novemberborn:
Oh that's neat. I think I just figured out how to make CI run for pull requests, so when that's good we can land this. Thanks!

PR #2299, opened 2019-11-22
Update mongodb-memory-server import in recipe

MoppetX:
new syntax found here: https://github.com/nodkz/mongodb-memory-server


novemberborn:
Thanks @MoppetX!

PR #2300, opened 2019-11-23
Run tests at selected line numbers

### LINKED ISSUE #2138: Add ability to run a test residing on a given line

ulken:
Only run tests at specific line number(s). Fixes #2138.

- *Format:* comma-separated list of `[X|Y-Z]`
- CLI only
	- Bail when trying to use together with `--watch`
- Single number, range or both
- Supports multiple files, merging line numbers for same file
- Deduplicate overlapping line numbers
- Deduplicate files, i.e. don't run same file more than once
	- Applies to non-file number runs as well
- Line numbers trumps running same file without line numbers
- Requires at least one test to be selected by line numbers, else error

### Approaches/considerations
A problem with the previous attempt was that it only included the start line number of each test function. It had no way of knowing on what line a function ended on and hence it had no choice but to match the preceding function, even if a line number in-between two functions was selected. This was a deal-breaker to me.

At first, I figured I could just do something like `CallSite.getFunction().toString().split('\n').length` to get the end line number. It worked perfectly... in the browser. In `node.js`, or rather in `use strict;` mode, on the other hand `CallSite.getFunction()` returns `undefined` due to security reasons (see last paragraph of [Stack trace API · V8](https://v8.dev/docs/stack-trace-api#customizing-stack-traces)).  And even if we were to drop `use strict;` for the particular files involved, it would still not be waterproof, since if the user uses `use strict;` in their test files, the same rules would apply. Also, we wouldn't want to drop `use strict;`.

Therefore, I set out to parse the test source file. The first approach I took was to drop `callsites` altogether and use the test title as an identifier to identify each test function. As a test title must be unique in AVA, it sort of worked. However, it was brittle as there was no way of ensuring it was an actual test function that was invoked. It could be any function with the test title as parameter. Also, you had to handle all different ways a test function could be called, nesting etc. Basically, it was not a good idea.

Nevertheless, if I just was able to pinpoint the test function, I could get my hands on the end line number. Time to re-introduce `callsites` again. It knew the exact function that was called to start the chain. And even though it was oblivious to the function definition (and thus end line number), it very well knew its start line number. That, if anything, sound like a good identifier to me.

So, it uses a combination of `callsites` and the test AST to figure out a test function's line numbers.

Even if start line number is a good identifier, it's not entirely unique. You can have multiple tests defined on the same line. If they were all one-liners, then all would be fine. But what if you have the following?

```js
3: test('unicorn', t => t.pass()); test('rainbow', t => {
4:     t.pass();
5: });
```

If you select `4` or `5` it would fail to run the second test as it would match the first test's start line number first (which ends on `3`). Hence, to compose a truly unique identifier for a test, we use the start line number _and_ the test title (which is guaranteed to be unique in AVA). Then it should work for all cases.

### Custom parser?
Why not use `parse-numeric-range`? Not invented here-syndrome, anyone?
* Unnecessarily complex for our needs
* Hard to read/reason about code
* Line numbers are more constrained than range
* Easier to provide sensical error messages, at the right level, without re-implementing logic on top

#### Don't run same file multiple times
**Reasoning:** To me it makes no sense to _run_ a file more than once (even if _stated_ more than once). Line numbers for the same file are merged, but the file is only run once. As a side effect, this also applies to the case when running without line numbers. To me, that's not a bad thing.

#### `lib/cli.js`: Explicitly relativize input paths from `process.cwd()` (non-breaking)
**Reasoning:** `resolveTestsFrom` is always `process.cwd()` when mapping over files apply:

```js
const resolveTestsFrom = cli.input.length === 0 ? projectDir : process.cwd();
```

As I found the overhead/indirection confusing, I made it explicit.

###### Old
```js
const files = cli.input.map(file =>
	path.relative(resolveTestsFrom, path.resolve(process.cwd(), file))
);
```

###### New
```js
const canonicalFilePaths = cli.input.map(file => path.resolve(process.cwd(), file)); // <-- Used for line numbers further down the line
const files = [...new Set(canonicalFilePaths.map(file =>
	path.relative(process.cwd() /* <-- This has changed */, stripLineNumbersSuffix(file)))
)];
```

With this in mind, I would even want to take it one step further. I don't see any reason to first `resolve` and then `relative`ize the paths in relation to `process.cwd()`. As far as my understanding goes:

```js
path.relative(process.cwd(), path.resolve(process.cwd(), file)) === path.relative(process.cwd, file)
```

That would make it:

```js
const files = [...new Set(cli.input/* <-- changed */.map(file =>
	path.relative(process.cwd(), stripLineNumbersSuffix(file)))
)];
```

In this case, the resolved/canonical paths are used elsewhere, but you get the point, I presume.

### Thoughts/future improvements
- Might be too complex, but we could support syntax like:
	- `test.js:-X` -> _from beginning of file up until line number X_
	- `test.js:X-` -> _from line number X until end of file_
- Update reporters to output selected line numbers?
- `api.js`: Include `lineNumbers` in `emit('run', plan)`?
- Include `lineNumbers` in `runner.js`'s  `meta`?

### Noticed
- `test/api.js`: The test `test file in node_modules is ignored` throws an (uncaught/unexpected) exception
	- Just thought I'd mention it as the test might succeed for other reasons than expected/desired

### Questions
I left them out as I noticed no difference in my case, but I've seen them sprinkled all over the place, so I wonder what are these used for?
- `require('../lib/chalk').set()`?
- `require('../lib/worker/options').set({})`?


lib/line-numbers.js
```diff
@@ -0,0 +1,147 @@
...
+const RANGE_REGEX = /^(?<start>-?\d+)-(?<end>-?\d+)$/;
+const LINE_NUMBERS_REGEX = /^(-?\d+(-\d+)?,?)+$/
+const PATH_DELIMITER = ':';
+
+const parseNumber = string => parseInt(string, 10);
+const range = (start, end) => Array(end - start + 1).fill(start).map((e, i) => e + i);
+const distinctArray = array => [...new Set(array)];
+const removeAllWhitespace = string => string.replace(/\s/g, "");
+const pluralizeIf = (string, condition) => condition ? `${string}s` : string;
+const getLineNumbersSuffix = path => path.split(PATH_DELIMITER).pop();
```
novemberborn: Perhaps this was discussed and we didn't consider the edge case worthwhile, but do we need to handle colons in the filename? Perhaps before the file extension?

lib/line-numbers.js
```diff
@@ -0,0 +1,147 @@
...
+  return sorted;
+};
+
+function parseLineNumbers(string) {
+  if (!string) {
+    throw new FormatError(string);
+  }
+
+  const parts = string.split(',');
+  const lineNumbers = parts.map(parsePart).flat();
```
novemberborn: Our minimal supported version will (soon) be Node.js 10 LTS, which I don't think has `flat()` available in it.

lib/line-numbers.js
```diff
@@ -0,0 +1,147 @@
...
+    const number = parseNumber(trimmed);
+    if (number <= 0) {
+      throw new Error(`Invalid line number: \`${trimmed}\`. Line numbers must be positive.`);
+    }
+    return number;
+  }
+
+  const match = RANGE_REGEX.exec(trimmed);
+  if (match) {
+    const {start, end} = Object.fromEntries(Object.entries(match.groups).map(([k, v]) => [k, parseNumber(v)]));
```
novemberborn: `Object.fromEntries` is not available on Node.js 10 LTS either.

ulken:
> You put a lot of thought into this, nice! I've had a quick look but I haven't actually tried it yet. The good news is that I fixed the CI config so tests now run. Unfortunately they're also failing…

Thanks! That's good to know. Yeah, mainly a bunch of linting errors, but also some platform specific stuff: the unsupported functions you mentioned, Windows paths and me running `yarn` instead of `npm` (`yarn.lock` vs. `package-lock.json`). My bad for not checking that. I'll make sure to run linting more diligently. I've also installed a VS Code plugin to pick up the `.editorconfig` which should avoid the obvious ones.

As for backwards compatibility with Node.js 10 (LTS):
- `Array.flat()` -> `lodash/flatten`
- RegExp named capturing groups -> Regular match + destructuring
- Removed usage of `Object.fromEntries`

I misinterpreted the [docs](https://github.com/avajs/ava#latest-javascript-support) a little and thought one could utilize all the lastest and greatest.

> Apologies if it's already the case, but could we make sure not to parse the AST unless line numbers were provided?

No worries. That's already taken care of [here](https://github.com/avajs/ava/pull/2300/files#diff-b3b53682a18f203ac8d29b0e277cad26R79): `getLineNumberRangeForTestInFile` does the actual AST parsing.

> What are your thoughts on the AST parsing when AVA is run on say TypeScript files?

Hadn't really thought about it, but it looks like there's support for it: [Using a different parser](https://github.com/benjamn/recast#using-a-different-parser). We would have to detect which one to use based on file type then, I guess?

lib/line-numbers.js
ulken: If I understand you correctly, that's already handled. See tests [`file delimiter without numbers in file name`](https://github.com/avajs/ava/pull/2300/files#diff-d4bc66da908e50f18b3a3a9cc95039ebR169) and [`file delimiter and numbers in file name`](https://github.com/avajs/ava/pull/2300/files#diff-d4bc66da908e50f18b3a3a9cc95039ebR174).

If I've missed the point, would you mind demonstrating with an example?

ulken:
@novemberborn Alright, after a series of fixes, CI checks are now green*. I've also added the documentation I found sufficient. I haven't added any _Recipe_. I'm not sure if needed at all, but if so, that could be added later, as the whole thing matures.

Anyhow, I consider the PR ready for review now.

* Just thought I'd mention: Had one flaky run where `try-commit-timeout`-smth failed on a few Windows machines. Seemed totally unrelated so didn't dig in to it.

ulken:
Also, I'd appreciate if you'd address my suggestions and questions in the PR description. Thanks!

ulken:
I can clean up the Git history, but as I've understood it you'll squash the whole thing into one commit on (potential) merge anyhow, right?

novemberborn:
> I can clean up the Git history, but as I've understood it you'll squash the whole thing into one commit on (potential) merge anyhow, right?

Yea most likely.

I did not have time to look at your changes yet, sorry.

ulken:
@novemberborn Just curious, any chance you (or anyone else) may be able to look at it anytime soon? Or should I just forget about it and it’ll happen if/when it’ll happen?

Would just be nice to know what to expect. Thanks!

novemberborn:
Hey @ulken, it's a case of it'll happen when it'll happen I'm afraid. There's life, and work, which had me travel overseas this week… still recovering! I'm hoping to have a good bunch of time to spend on AVA at the end of the year.

ulken:
Alrighty, thanks for the info. I’ll just leave it here then. Good luck!

novemberborn:
Nice work here @ulken!

I've changed a fair amount of code in `master` since you opened this PR. I've tried to resolve the conflicts in the merge commit I just pushed, but then more changes were needed in the second commit.

The big change is that AVA now takes glob patterns, not files. I've tried to adjust for that. Turns out the line number selection now also works with the watcher!

I found it quite hard to follow the `line-numbers` code. I've refactored it a little to remove some (in my opinion) unnecessary defensiveness and remove exports. I haven't updated the tests for that module, sorry. Speaking off, it looks like you're testing some very low-level behaviors. Perhaps you could simplify the tests?

> #### Don't run same file multiple times
> **Reasoning:** To me it makes no sense to _run_ a file more than once (even if _stated_ more than once). Line numbers for the same file are merged, but the file is only run once. As a side effect, this also applies to the case when running without line numbers. To me, that's not a bad thing.

This is covered by the changes to file selection already in `master`.

> #### `lib/cli.js`: Explicitly relativize input paths from `process.cwd()` (non-breaking)
> **Reasoning:** `resolveTestsFrom` is always `process.cwd()` when mapping over files apply:

I stumbled across this too, funnily enough.

---

I have a question about `recast` — how sensitive is this to test files that have new syntax (because folks use Babel), or TypeScript sources? What happens when it fails to parse a file? Should we have an option to disable it? Or perhaps should we make it opt-in or experimental?

lib/line-numbers.js
```diff
@@ -0,0 +1,138 @@
...
+		const numberPluralized = pluralizeIf('number', offender && !NUMBER_REGEX.test(offender) && offender.length !== 1);
+		super(`Invalid line ${numberPluralized}: \`${offender}\`. Expected comma-separated list of \`[X|Y-Z]\`.`);
+		this.name = 'FormatError';
+	}
+}
+
+const parseNumber = string => parseInt(string, 10);
+const removeAllWhitespace = string => string.replace(/\s/g, '');
+
+const parseLineNumbers = suffix => {
```
novemberborn: I still need to check how these errors are presented on the CLI. Especially file paths, which would be absolute values which is jarring.

lib/line-numbers.js
```diff
@@ -0,0 +1,138 @@
...
+		if (match) {
+			const {startGroup, endGroup} = match.groups;
+			const start = parseNumber(startGroup);
+			const end = parseNumber(endGroup);
+			if (start < 0 || end < 0) {
+				throw new Error(`Invalid line number range: \`${trimmed}\`. Line numbers must be positive.`);
+			}
+
+			if (start > end) {
+				throw new Error(`Invalid line number range: \`${trimmed}\`. \`start\` must be lesser than \`end\`.`);
```
novemberborn:
```suggestion
				throw new Error(`Invalid line number range: \`${trimmed}\`. \`start\` must be less than \`end\`.`);
```

lib/line-numbers.js
```diff
@@ -0,0 +1,138 @@
...
+		const {title, startLineNumber} = test || {};
+		throw new Error(`Failed to resolve end line number for test \`${title}\` starting at line number ${startLineNumber} in ${filePath}: ${error.message}`);
+	}
+};
+
+function getLineNumberRangeForTestInFile(title, filePath) {
+	const callSite = callsites().find(callSite => callSite.getFileName() === filePath);
+
+	if (!callSite) {
+		throw new Error(`Failed to resolve line number range for test in ${filePath}: Test never called.`);
```
novemberborn: What does "Test never called" mean?

lib/test-ast.js
```diff
@@ -0,0 +1,91 @@
+'use strict';
+
+const fs = require('fs');
+const AST = require('recast');
+
+/* Lives throughout one test run as `watch` mode is not supported */
```
novemberborn: This module is loaded in a subprocess, so this comment does not apply.

Still, I wonder if we should do the AST parsing explicitly and cache the results based on the file contents.

lib/test-ast.js
```diff
@@ -0,0 +1,91 @@
...
+const fileAstCache = {};
+
+/**
+ * Retrieve AST for a test starting at line number with a given title
+ *
+ * @param test      Start line number and title of test function
```
novemberborn: Why not use object destructuring and make this self documenting?

(I have a severe dislike of these JSDoc comments, sorry.)

lib/test-ast.js
```diff
@@ -0,0 +1,91 @@
...
+	if (!startLineNumber) {
+		throw new TypeError('Start line number required.');
+	}
+
+	if (!title) {
+		throw new TypeError('Test title required.');
+	}
+
+	if (!filePath) {
+		throw new TypeError('File path required.');
```
novemberborn: Does this really need to be so defensive?

novemberborn:
@ulken would you have time to pick this up again?

ulken:
@novemberborn Hi, sorry for not getting back to you any sooner. To be honest, I got a little thrown off by the long feedback cycle (nothing personal, I totally get that you guys have other stuff to tend too) after having invested quite some time in this. Also, by the quick glance I did after your review, I wasn't too happy about some of the changes, which put me off even more. But I'll have another look with fresh eyes. Maybe I was just being grumpy. Or we just disagree and I'll leave it up to you what do with it. But I'll get back to you!

ulken:
I'll run through the comments at first and then have a look at the code changes

lib/line-numbers.js
ulken: @novemberborn Please resolve if you feel this is taken care of. Otherwise, would you mind providing more details/examples?

ulken:
>I found it quite hard to follow the line-numbers code. I've refactored it a little to remove some (in my opinion) unnecessary defensiveness and remove exports. I haven't updated the tests for that module, sorry. Speaking off, it looks like you're testing some very low-level behaviors. Perhaps you could simplify the tests?

This is what threw me off. Having a quick look at it, I didn't find the refactoring an improvement. Sure, exports is just a matter of taste, but assigning to variables in loops, naming of variables etc. I simply disapprove of and on the contrary find harder to reason about. But, as I said, I'll have another look to either confirm or invalidate my initial reaction.

Regarding low-level testing, I really found the feature easier to develop by putting together small components to stitch together. The testing of some components is lower than others, but I actually did evaluate compacting the tests afterwards, but found them to be much clunkier and therefore refrained from it.

ulken:
>I have a question about recast — how sensitive is this to test files that have new syntax (because folks use Babel), or TypeScript sources? What happens when it fails to parse a file? Should we have an option to disable it? Or perhaps should we make it opt-in or experimental?

I haven't tested this. Do you have any example functionality/syntax you want to try out?

Regarding TypeScript, I haven't looked into it, but as I have answered you above, there seems to be support using a different parser. So we would have to determine what parser to use.

lib/line-numbers.js
ulken: Verdict?

lib/test-ast.js
ulken: I'm totally with you on this, and that's what I had at first. Then the linter swung by to whine about destructuring anything but the last argument.

But if you're willing to change the linter settings, I'll be happy to revert. Still found this better than nothing as "test" can mean different things.

lib/test-ast.js
ulken: Matter of taste, I guess. Personally, I try to assume nothing and be as helpful as possible to the consumer. If TypeScript was used, you could have the best of both worlds, I guess.

ulken:
Alright, replied to comments. I'll have a look at the code changes later today and get back to you with my view on it.

docs/05-command-line.md
```diff
@@ -3,37 +3,42 @@
...
-  The above relies on your shell expanding the glob patterns.
-  Without arguments, AVA uses the following patterns:
-    **/test.js **/test-*.js **/*.spec.js **/*.test.js **/test/**/*.js **/tests/**/*.js **/__tests__/**/*.js
+ava [<pattern>[:<line-numbers>]...]
+ava debug [<pattern>[:<line-numbers>]...]
+ava reset-cache
+
+Commands:
+  ava [<pattern>...]        Run tests                                  [default]
+  ava debug [<pattern>...]  Activate Node.js inspector and run a single test
```
ulken: Mention `[:line-numbers]` here?

docs/05-command-line.md
```diff
@@ -3,37 +3,42 @@
...
+                            file
+  ava reset-cache           Reset AVA's compilation cache and exit
+
+Positionals:
+  pattern  Glob patterns to select what test files to run. Leave empty if you
+           want AVA to run all test files instead                       [string]
```
ulken: Mention `line-numbers` here?

docs/05-command-line.md
```diff
@@ -3,37 +3,42 @@
...
+  --tap, -t               Generate TAP output                          [boolean]
+  --timeout, -T           Set global timeout (milliseconds or human-readable,
+                          e.g. 10s, 2m)                                 [string]
+  --update-snapshots, -u  Update snapshots                             [boolean]
+  --verbose, -v           Enable verbose output                        [boolean]
+  --watch, -w             Re-run tests when files change               [boolean]
+
+Examples:
+  ava
+  ava test.js
```
ulken: Give example of `line-numbers` usage here?

lib/cli.js
```diff
@@ -211,6 +159,7 @@ exports.run = async () => { // eslint-disable-line complexity
 			})
 		.example('$0')
 		.example('$0 test.js')
+		.example('$0 test.js:4,7-9')
```
ulken: Documentation should match, right?

lib/line-numbers.js
```diff
@@ -17,8 +17,8 @@ module.exports = {
 };
 
 const NUMBER_REGEX = /^-?\d+$/;
-const RANGE_REGEX = /^(-?\d+)-(-?\d+)$/;
-const LINE_NUMBERS_REGEX = /^(-?\d+(-\d+)?,?)+$/;
+const RANGE_REGEX = /^(?<startGroup>-?\d+)-(?<endGroup>-?\d+)$/;
```
ulken: I did use named capturing groups at first, but you (and the integration tests) told me not to because of Node version compatibility? Has that changed?

Does the same apply to the `Array.flat()` and `Object.fromEntries()` which I also replaced by request?

lib/line-numbers.js
```diff
@@ -1,93 +1,117 @@
...
+		}
+
 		throw new FormatError(string);
 	}
 
-	const parts = string.split(',');
-	const lineNumbers = flatten(parts.map(parsePart));
-	return sortNumbersAscending(distinctArray(lineNumbers));
-}
+	return retval;
```
ulken: I genuinely think this should be replaced by/reverted to: `flatten(suffix.split(',').map(parsePart))`. Breaking out `parsePart()` is not strictly necessary. Inlining inside the `map()` works too. I just find it easier to read (and test!).

ulken:
@novemberborn Okidoki, I've looked at the changes you've made and commented on them. I won't do any further changes until you've responded to see where we stand on this.

lib/line-numbers.js
novemberborn: Apologies… it's been three months now so I don't quite recall why I rewrote it like this. Possibly because I struggled to read the previous version of the code 😊

ulken:
>I'm sorry about that. This is a hard PR to review, and I think I was trying to land a whole bunch of breaking changes at the same time.

Alright, I feel you.

>That wasn't my intent. IIRC, after landing all the breaking changes, it took some effort to make this work again. Maybe I went a little overboard.

Gotcha. Thanks for clarifying.

>I get that. I've found though that it's the low level tests in a big system like AVA that tend to be brittle. Testing the behavior is better in the long run. Though that's still hard… I'm hoping to get to a place where we can test using AVA itself (#2380).

I hear you. Makes sense. I'll run through the test suite and try to slim it down accordingly.

>JSX, maybe. Any early stage syntax that is supported by Babel or TypeScript but hasn't made its way into recast.

Sure, I'll check how it behaves on JSX.

>We now have external packages for Babel support and TypeScript support so we could implement line number parsing there. I'm happy to ship this as an experimental feature until we figure it out, just to avoid confusion / crashes for Babel / TypeScript users.

OK, I'd appreciate if you'd make that decision. How would I go about flagging it as experimental? I haven't looked into how these external packages tie into AVA. Would that mean we would need to re-implement the whole thing in each package or can they utilize/access AVA code? We would need to make the parsing bit configurable somehow if so.

**Just to be clear:** Would the above mean I can forget about Babel/TypeScript support for now and just focus on landed syntax in vanilla JS (i.e. as supported by `recast`)?

ulken:
Follow-up on `recast` and TypeScript/Babel support: It looks like there are parsers for both: [recast parsers](https://github.com/benjamn/recast/tree/master/parsers).

novemberborn:
> Would the above mean I can forget about Babel/TypeScript support for now and just focus on landed syntax in vanilla JS (i.e. as supported by `recast`)?

Yup, that's fine.

> How would I go about flagging it as experimental?

We currently don't have active experiments so it's a little harder to figure out. See this line maybe: https://github.com/avajs/ava/pull/2435/files#diff-1632eed7c39ed30bb11f055e14ffc7bdR10

And then look for `experimentalTestInterfaces` to see where and how it's accessed.

Feel free to punt this to me though.

novemberborn:
> Would that mean we would need to re-implement the whole thing in each package or can they utilize/access AVA code? We would need to make the parsing bit configurable somehow if so.

We can expose functionality to the providers. Probably we'd do it so they only need to implement the `recast` bit. And maybe we'd detect when you are trying to select lines from files that are handled by the providers, and then refuse until we have providers available that can do that. But we don't need to worry about that yet.

lib/line-numbers.js
ulken: Sure, as it's used now. But there is nothing stopping other parts of the system to try to use in the future, bypassing test declaration. The error message was (an attempt to be) for them.

I can remove the last bit. I get that it doesn't help someone not in on the details.

Thank you for enlightening me on the nomenclature. Makes sense! And yes, one of the three well-known things that is hard with programming, right? ;)

lib/test-ast.js
ulken: OK, I think I follow you now. My thought regarding caching was not having to read file contents from disk for the _same_ file multiple times, if multiple locations in a file is selected. IO is costly, right? But even if you were to select say 10 locations, it might not be a problem? 

Anyway, I'll drop the cache. Also, performance optimizations should always be accompanied by benchmarks, right?

lib/test-ast.js
ulken: You might be right. I'd prefer keeping them as separate parameters, as I feel the "test" and the file that it's enclosed in are not the same concept. Do you always use `disable-line` or just disable the specific rule? I prefer the latter as it describes intent and still captures other potential errors if the code were to change in other ways.

lib/test-ast.js
ulken: Hum, interesting view. I guess I don't separate internal and external consumers, but I'll surrender on this one and remove it (and other similar cases).

lib/line-numbers.js
ulken: I see. Thank you.

lib/line-numbers.js
ulken: OK, good. Apology accepted. I guess I owe you an apology back then for writing such un-readable code...

lib/line-numbers.js
ulken: ☝️ This one remains unanswered

lib/test-ast.js
novemberborn: > My thought regarding caching was not having to read file contents from disk for the _same_ file multiple times, if multiple locations in a file is selected.

Right, fair enough. Though the worker process would only ever load the one file.

> IO is costly, right? But even if you were to select say 10 locations, it might not be a problem?

Probably not. Though is this even called more than once, given that a process is launched for a specific test file?



> Also, performance optimizations should always be accompanied by benchmarks, right?

I guess… benchmarking is hard too! 😵

lib/test-ast.js
novemberborn: > I'd prefer keeping them as separate parameters, as I feel the "test" and the file that it's enclosed in are not the same concept.

Yea it really doesn't matter for such an internal function.



> Do you always use `disable-line` or just disable the specific rule? I prefer the latter as it describes intent and still captures other potential errors if the code were to change in other ways.

XO won't allow you to disable the line 😉

lib/line-numbers.js
novemberborn: I haven't checked yet, sorry 😄 Will resolve for now.

ulken:
Looks like `recast` just landed [support for JSX](https://github.com/benjamn/recast/pull/676) in their default parser.

ulken:
@novemberborn Alright, she's all yours.

novemberborn:
@ulken 😄 

I probably won't have time to look at this until the weekend, but at least the weekend starts on Friday, and it's not like I can go out anyway 😉

ulken:
Figured as much. No rush.

Forgot to mention: Changed your code a little. You stripped the suffix even if it wasn't valid line numbers syntax. I found that odd and instead return it untouched.

Also, before you got friendly error messages when you used invalid syntax. Now it just ignores anything invalid. Not saying one is better than the other, just bringing it to the surface.

novemberborn:
> You stripped the suffix even if it wasn't valid line numbers syntax. I found that odd and instead return it untouched.

Maybe my intent was that anything after a `:` is ignored, so that it can only be used to select line numbers. Glob patterns are confusing enough as it is.

> Also, before you got friendly error messages when you used invalid syntax. Now it just ignores anything invalid. Not saying one is better than the other, just bringing it to the surface.

That gels with disregarding any unknown `:…` suffixes.

ulken:
> Maybe my intent was that anything after a `:` is ignored, so that it can only be used to select line numbers. Glob patterns are confusing enough as it is.

Fair enough. But then you won't be able to run test files with `:` in the filename (I mean the regular filename, not as an suffix). E.g. `test:foo.js` would yield `test` as pattern and thus not find the file at all (even for a normal run, without trying to select line numbers). But maybe a non-issue/constraint you want to put on using AVA?

> That gels with disregarding any unknown `:…` suffixes.

Maybe so. I agree keeping it simple is desirable.

novemberborn:
As you can see by the commits I did start to look into this, but it's my day off as well so I'll resume later this weekend 😄

lib/runner.js
```diff
@@ -408,16 +431,30 @@ class Runner extends Emittery {
...
 			});
+
+			todoTests.push(task);
+		}
+
+		const hasRunnableTests = concurrentTests.length > 0 || serialTests.length > 0;
+
+		const hasSelectedTests = hasRunnableTests || todoTests.length > 0;
+		if (this.runOnlySelectedLines && !hasSelectedTests) {
+			this.emit('error', 'No tests selected by line numbers.');
```
novemberborn: We should handle this the same way we handle `--match`, in the reporters. But only if line numbers were specified for this test file:

https://github.com/avajs/ava/blob/cb5f9f70879ee549837c31e577e01f87970164ed/lib/reporters/mini.js#L388

https://github.com/avajs/ava/blob/cb5f9f70879ee549837c31e577e01f87970164ed/lib/reporters/verbose.js#L330

lib/line-numbers.js
```diff
@@ -0,0 +1,118 @@
...
+const sourceMapSupport = require('source-map-support');
+const flatten = require('lodash/flatten');
+const parseTestSourceInFile = require('./test-ast');
+
+const NUMBER_REGEX = /^-?\d+$/;
+const RANGE_REGEX = /^(?<startGroup>-?\d+)-(?<endGroup>-?\d+)$/;
+const LINE_NUMBERS_REGEX = /^(?:-?\d+(?:--?\d+)?,?)+$/;
+const DELIMITER = ':';
+
+const pluralizeIf = (string, condition) => condition ? `${string}s` : string;
```
novemberborn, line 14: We already use https://www.npmjs.com/package/plur for this.

lib/line-numbers.js
```diff
@@ -0,0 +1,118 @@
...
+	if (parts.length === 1) {
+		return {pattern, lineNumbers: null};
+	}
+
+	const suffix = removeAllWhitespace(parts.pop());
+	if (!LINE_NUMBERS_REGEX.test(suffix)) {
+		return {pattern, lineNumbers: null};
+	}
+
+	return {pattern: parts.join(DELIMITER), lineNumbers: parseLineNumbers(suffix)};
```
novemberborn, line 75: On the one hand, this tries to ignore patterns that don't look like they select line numbers. But then on the other hand this throw errors when the numbers or ranges look invalid.

`/^(?:\d+(?:-\d+)?,?)+$/` should match individual numbers and ranges. This just leaves us with a range where the start is after the end. Maybe we should normalize that range?

Anything else would be treated as part of the pattern.

lib/line-numbers.js
```diff
@@ -0,0 +1,118 @@
...
+		return node.loc.end.line;
+	} catch (error) {
+		throw new Error(`Failed to resolve end line number for test \`${title}\` starting at line number ${startLineNumber} in ${filePath}: ${error.message}`);
+	}
+};
+
+function getLineNumberRangeForTestInFile(title, filePath) {
+	const callSite = callsites().find(callSite => callSite.getFileName() === filePath);
+	if (!callSite) {
+		throw new Error(`Failed to resolve line number range for test \`${title}\` in ${filePath}.`);
```
novemberborn, line 102: When would this happen?

Once you start declaring tests, you must declare them all at once, but you could wait before declaring the first test. So I suppose you could have the actual test declarations in a separate file, and then use a `setTimeout()`, and then we wouldn't be able to find any call site.

And if that's not the issue, but the declarations are still happening in a different file, then line number selection is just not going to work.

I wonder if instead of throwing an error, we should print a warning and just run all tests. Here and when resolving the end line. What do you think?

lib/test-ast.js
```diff
@@ -0,0 +1,54 @@
...
+		}
+	});
+
+	if (candidates.length === 0) {
+		throw new Error(`No test starting at line number ${startLineNumber} in ${filePath}.`);
+	}
+
+	const testAst = candidates.find(candidate => {
+		const [firstArgument] = candidate.arguments;
+		return firstArgument && firstArgument.value === title;
```
novemberborn, line 29: You can declare tests without a title argument, as long as the implementation is able to provide a title for you. Or, even if there is a title argument, it may not be the exact same string. Or it may be a template literal.

In other words I don't think this will work! Why do you think it's needed? Apologies because you've probably explained this once already.

lib/test-ast.js
```diff
@@ -0,0 +1,54 @@
+'use strict';
+
+const fs = require('fs');
+const AST = require('recast');
+
+module.exports = ({startLineNumber, title}, filePath) => {
+	const fileAst = AST.parse(tryReadFile(filePath));
```
novemberborn, line 7: Turns out this code runs for every test declaration… so we'll definitely have to cache this. Maybe using `mem`? Or `runner` could parse in advance? In fact we'd probably want to cache the tree traversal.

lib/globs.js
```diff
@@ -23,23 +23,27 @@ const defaultIgnoredByWatcherPatterns = [
...
 
-		return pattern;
-	});
+	return pattern;
+}
+
+exports.normalizePattern = normalizePattern;
+
+function normalizePatterns(patterns) {
+	return patterns.map(pattern => normalizePattern(pattern));
```
ulken, line 46: Just curious: Why not?
```suggestion
	return patterns.map(normalizePattern);
```

Also, is this function worth it now?

I mean `patterns.map(normalizePattern)` vs. `normalizePatterns(patterns)`?

lib/api.js
```diff
@@ -211,10 +215,12 @@ class Api extends Emittery {
 
 				const options = {
 					...apiOptions,
+					filter,
 					providerStates,
 					recordNewSnapshots: !isCi,
 					// If we're looking for matches, run every single test process in exclusive-only mode
-					runOnlyExclusive: apiOptions.match.length > 0 || runtimeOptions.runOnlyExclusive === true
+					runOnlyExclusive: apiOptions.match.length > 0 || runtimeOptions.runOnlyExclusive === true,
+					runOnlySelectedLines: filter.some(({lineNumbers}) => lineNumbers !== null)
```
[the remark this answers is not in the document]
ulken: You might be right. Can't really see the whole line in front of me. `runOnlySelectedLines` is used in quite a few places in `runner.js`, but at the least I guess you could compute its value in there instead?

lib/runner.js
ulken: Sounds good!

lib/line-numbers.js
ulken: Ah, cool! Good to know. Although, judging by the comment below neither should be needed any longer?

lib/line-numbers.js
ulken: Yes, I came to think of this too. Reversed ranges _and_ negative numbers, right?

I don't quite get what you mean by _"normalize that range"_, though? Do you mind clarifying/giving an example?

Also, I don't think `throw new FormatError(...)` can occur/is needed anymore? On the other hand, I don't really know what to do if it were to fall through, so might be better to leave it...

lib/line-numbers.js
ulken: Hehe, that error has re-surfaced quite a few times now...

So, it's not an end user error. It's a developer error. If the function `getLineNumberRangeForTestInFile` was to be used/called without a test declaration being the initiator, in which case there would be no relevant call site to be found.

Since I've learned you're not really fond of DX errors, I think we can just drop it altogether.

lib/test-ast.js
ulken: Please re-read the description of the PR under **Approaches/considerations** for an explanation as to why.

You’re probably right about the empty case and template strings, though. So might need to rethink it – again.

lib/test-ast.js
ulken: Yeah, that’s what I tried to tell you before.

Sounds good!

lib/api.js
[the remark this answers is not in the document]
novemberborn: Yea. Or compute it in the API code, and then pass down the boolean.

lib/line-numbers.js
novemberborn: Just figured I'd share it regardless 😄

lib/line-numbers.js
novemberborn: The regular expression wouldn't match negative numbers, so it'd be treated as part of the glob pattern instead.

For reverse ranges (start > end) we could normalize the range, so swap start and end.

Now there shouldn't be any errors.

lib/api.js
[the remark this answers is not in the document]
ulken: Hum, not sure I’m following. How is that different from what’s being done already?

lib/line-numbers.js
novemberborn: Yes, but you're assuming tests are declared in the test file. That would be typical, but it's not necessary.

Rather than failing, maybe we should log warnings and run whatever tests we did manage to select by line number. And if it turns out that we couldn't match any, then at least there's warnings.

lib/line-numbers.js
ulken: > The regular expression wouldn't match negative numbers, so it'd be treated as part of the glob pattern instead.

I beg to differ. Take another look. 

> For reverse ranges (start > end) we could normalize the range, so swap start and end.

Ah, works for me.

lib/test-ast.js
novemberborn: Based on the call sites we can determine the starting line of the `test()` call. Using the AST we can determine all call expressions. So for each declaration, we can determine all call expressions at that starting line. We could then figure out the end lines of each of those expressions and consider that to be the line number range for that test.

I don't think this needs the title at all?

lib/test-ast.js
novemberborn: Yea… sorry about that.

lib/globs.js
novemberborn: It's a defensive pattern, since `Array#map()` passes two additional arguments. If ever we change `normalizePattern` to take a second argument, for a new use case, it's going to receive the array index for this use case.

lib/api.js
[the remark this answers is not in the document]
novemberborn: This checks whether there are *any* line number filters. Even if it's for different files. Files for which there are no line number selections should run all their tests.

lib/line-numbers.js
novemberborn: > > The regular expression wouldn't match negative numbers, so it'd be treated as part of the glob pattern instead.
> 
> I beg to differ. Take another look.

I meant this one: `/^(?:\d+(?:-\d+)?,?)+$/`

`/^(?:-?\d+(?:--?\d+)?,?)+$/` does match negative numbers.

lib/api.js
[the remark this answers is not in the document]
ulken: Got it. Thanks.

lib/line-numbers.js
ulken: Right

lib/line-numbers.js
ulken: Wasn’t aware of that use case, but sure. Makes sense.

lib/test-ast.js
ulken: I realize it’s an extreme edge case, but I’m pretty sure it wouldn’t suffice for the use case I referred to in the PR description.

Like the example states: if you were to select lines 4-5 it would either run both tests or the wrong one (as it would match the one declared first).

But maybe you’re already aware and accept it or you realize something I don’t. 

Either way, your flaws are yet to be handled...

lib/test-ast.js
ulken: Just to make sure: When you say _”cache the tree traversal”_, you mean the parsed test AST, right?

Also, have you decided on whether to have this done up front by `runner.js` or on-the-fly?

lib/test-ast.js
novemberborn: But the AST would have the start and end lines of the first declaration (line 3), and the start and and lines of the second declaration (3 and 5), so then 4-5 would only run the second one, no?

lib/test-ast.js
novemberborn: If we know, in the runner, that we're selecting by lines *in this file*, then we can analyze up front and build a map of declaration start and end lines, indexed by the start line.

lib/test-ast.js
ulken: When you put it that way... yes, sounds perfectly reasonable. Sorry for being so slow. I'll give it a go. Thanks for bearing with me 😬

lib/test-ast.js
ulken: Aye, I'll give it a try.

lib/globs.js
ulken: Fair enough

lib/api.js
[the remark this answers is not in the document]
ulken: Better?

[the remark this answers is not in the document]
ulken: To be fair: I didn’t know how the filter worked and thought it was file specific. Also, this is how it worked before your “grand refactoring” 😉.

lib/api.js
```diff
@@ -211,10 +216,15 @@ class Api extends Emittery {
...
 					...apiOptions,
+					filter,
 					providerStates,
 					recordNewSnapshots: !isCi,
 					// If we're looking for matches, run every single test process in exclusive-only mode
-					runOnlyExclusive: apiOptions.match.length > 0 || runtimeOptions.runOnlyExclusive === true
+					runOnlyExclusive: apiOptions.match.length > 0 || runtimeOptions.runOnlyExclusive === true,
+					runOnlySelectedLines: filter.some(({pattern, lineNumbers}) =>
+						micromatch.isMatch(globs.normalizeFileForMatching(apiOptions.projectDir, file), pattern) &&
+						lineNumbers !== null
```
novemberborn, line 226: Let's check `lineNumbers` first so we don't match unnecessarily.

```suggestion
						lineNumbers !== null && micromatch.isMatch(globs.normalizeFileForMatching(apiOptions.projectDir, file), pattern)
```

lib/api.js
```diff
@@ -211,10 +216,15 @@ class Api extends Emittery {
 
 				const options = {
 					...apiOptions,
+					filter,
 					providerStates,
 					recordNewSnapshots: !isCi,
 					// If we're looking for matches, run every single test process in exclusive-only mode
-					runOnlyExclusive: apiOptions.match.length > 0 || runtimeOptions.runOnlyExclusive === true
+					runOnlyExclusive: apiOptions.match.length > 0 || runtimeOptions.runOnlyExclusive === true,
+					runOnlySelectedLines: filter.some(({pattern, lineNumbers}) =>
```
novemberborn, line 224: Rather than a boolean, what if we pass an array of line numbers? We're already checking for the filter items that match the test file anyway. And then we don't have to pass `filter` above.

ulken: Works for me!

ulken: Done. I still compute `runOnlySelectedLines` (= `lineNumbers.length > 0`) in `runner.js` as that is used in so many places.

lib/runner.js
ulken: Not quite sure how to do this efficiently, though. In the `--match` case we pass `matching` as a property of the plan when the run is started. The `run` event is emitted before each file is being processed in `api.js`. 

We would need to compute a map from file to `hasLineNumbers/runOnlySelectedLines` and pass down? How do we know which test file is running in the reports, though? 

I noticed that some hooks receive event data with the currently running test, but `reporter.endRun()` doesn't seem to receive any such info and as far as I can tell no such state is saved in the reporters.

novemberborn: You're right `matching` is for the global `--match` option.

From the API, the run status for each file is set up here:

https://github.com/avajs/ava/blob/18aeac64db2711d37853175b5e8dc1fd72c66d2a/lib/api.js#L226

You could pass whether we're hoping to select lines. Store it here:

https://github.com/avajs/ava/blob/18aeac64db2711d37853175b5e8dc1fd72c66d2a/lib/run-status.js#L34

This then becomes available in `fileStats` here:

https://github.com/avajs/ava/blob/18aeac64db2711d37853175b5e8dc1fd72c66d2a/lib/reporters/verbose.js#L196

Look for `selectedTests === 0`.

novemberborn: Just to add, I pointed you in the wrong direction with `endRun()`. That's for the entire test run, because `--match` applies to the whole thing. Here we want the `worker-finished` event.

ulken: Alright, wow. Not sure I would've been able to figure that out on my own... Thanks!

ulken: Something like that?

I assumed you wanted the `mini` reporter to be updated as well accordingly? Question: Do you want a `filesWithoutSelectedTests` as well?

ulken: Also, it felt a little weird to add the boolean to the stats object, but you lead, I follow.

lib/line-numbers.js
ulken: Uhm, how would you print a warning from here?

lib/test-ast.js
ulken: No... I think I'm gonna have to throw in the towel on this one... I thought I got it, but trying to implement it I keep coming to the same conclusion: There's no way to distinguish between two tests which starts on the same line if that's our _only_ identifier.

I guess the part I don't get is how this would work:

>We could then figure out the end lines of each of those expressions and consider that to be the line number range for that test.

This doesn't make sense to me. How can you consider the end line of another test to be the call site test's end line?

Just to do a quick walkthrough of the example at hand:
1. We parse the test source
2. We find two candidate nodes (starting at line 3)
3. ?
4. Profit*

I need help filling in `3. ?` above. How do we go from two candidate nodes starting at the same line to determining the end line of the test in question?

Maybe you have a totally different implementation/usage of the AST in your head which is why I don't follow.

If you're convinced it would work and feel like you're talking to a wall, feel free to implement it yourself.

* Just in case: https://knowyourmeme.com/memes/profit

lib/test-ast.js
ulken: I'm skipping this one until it's been determined start line number is unique enough.

lib/runner.js
novemberborn: > Alright, wow. Not sure I would've been able to figure that out on my own... Thanks!

Yes it's not great… but it works!

> I assumed you wanted the `mini` reporter to be updated accordingly as well?

Yes. Sorry about the duplicated code.

> Question: Do you want a filesWithoutSelectedTests as well?

Not sure what you mean by this?

lib/line-numbers.js
novemberborn: I think either a warning if no tests were selected, or perhaps a warning if not all lines / ranges lead to a test being selected. But that wouldn't be here, that'd be something in the runner maybe. I can give this a go if you'd like.

(It would also help us deal with Babel & TypeScript… we'd just warn that we couldn't find tests for now.)

lib/test-ast.js
novemberborn: Sure I can give this a go.

ulken: Cool 😎

Just to re-phrase where my logic breaks down: I don't see how you can determine which call expression/end line number belongs to which test (if they have the same start line number).

But maybe I'm too deeply invested into this approach that I can't see the obvious...

Good luck!

novemberborn: I don't see why it's relevant. There's two tests on this line? Fine, we'll run both.

lib/line-numbers.js
ulken: Yeah, sure. I was more thinking how do you normally (technically) communicate warnings from the non-UI code? Isn't that done in the reporters like with the other case we had. When I come to think about it, isn't it pretty much the same warning/error?

True that about Babel & TS.

lib/runner.js
ulken: > Not sure what you mean by this?

I'm referring to this guy: https://github.com/avajs/ava/blob/18aeac64db2711d37853175b5e8dc1fd72c66d2a/lib/reporters/mini.js#L209

lib/line-numbers.js
novemberborn: I… don't immediately have an answer, sorry. I can look into this.

lib/runner.js
novemberborn: Ah. Will have to dig into this.

lib/test-ast.js
ulken: If you select the starting line (3), absolutely. But in the example I present they _start_ on the same line but don't _end_ on the same line. If you were to select a line which is part of only one of the tests, the other test shouldn't run.

But I'm gonna drop this now. I'll probably just confuse/annoy you. I sincerely hope you're right and fail to see the obvious.

ulken: So, I had a revelation: instead of using the test title, we can pinpoint a specific test declaration by using its column number, in addition to line number.

When I look back at it, it’s so obvious with regards to the example case I posted. Since the two tests are on the same line, they must have different column numbers.

Now we can use a `Map` indexed by position (`{line, column}`) to cache line numbers. 

Thoughts?

Or are you still confident I’m lost and you can get away with only using start line number?

lib/test-ast.js
ulken: I went ahead and did this as well, according to the new approach

ulken:
@novemberborn Alright, apart from the open questions for you to look at, I can't find more for me to do here. Assuming you won't have time to get to it before the weekend earliest? 

Let me know if there's anything you want me to look into or if you can give me a little guidance on the unsolved stuff.

Would be nice to seal the deal on this...

novemberborn:
@ulken I'm about to push a simplification, let me know what you think. The approach is to just look for the narrowest call expression that covers the test declaration, and then use its end position to figure out if the declaration is selected.

The last commit causes the tests to fail. It's a bit confusing but it looks like the nested test declaration, according to `recast`, starts at column 4, but according to the call site it should be column 1. Or maybe a call expression was already erroneously rejected. I'm out of time for today.

Remaining issues:

* [ ] If no tests are selected, should that cause an exit code of 1?
* [ ] Work out how `@ava/babel` and `@ava/typescript` could participate in this
* [ ] Add an integration test for AST parse errors (probably using `@ava/babel` somehow)
* [ ] Fix test failure

lib/line-numbers.js
```diff
@@ -59,41 +58,91 @@ exports.splitPatternAndLineNumbers = splitPatternAndLineNumbers;
...
+	// or can't be parsed, perhaps because it's a TypeScript file.
+	let ast;
+	try {
+		ast = recast.parse(fs.readFileSync(file, 'utf8'));
+	} catch {
+		return undefined;
 	}
 
-	const sourceCallSite = sourceMapSupport.wrapCallSite(callSite);
+	// Record the start and end positions of all call expressions.
```
ulken, line 86: Find this comment redundant. I think the code speaks for itself in this case.

lib/line-numbers.js
```diff
@@ -59,41 +58,91 @@ exports.splitPatternAndLineNumbers = splitPatternAndLineNumbers;
...
+				}
+
+				if (loc.end.line === start.line && loc.end.column < start.column) {
+					return false;
+				}
+
+				return true;
+			});
+
+			// Call locations should be sorted by source order, so the last enclosing
```
ulken, line 130: I think this comment should be moved down, closer to its referred to statement.

lib/line-numbers.js
```diff
@@ -0,0 +1,148 @@
...
+			}
+
+			const testDeclaration = enclosingCalls.pop();
+			for (let {line} = testDeclaration.start; line <= testDeclaration.end.line; line++) {
+				if (selectedLines.has(line)) {
+					return true;
+				}
+			}
+
+			return false;
```
ulken, line 143: I would still prefer:

```suggestion
			return range(testDeclaration.start.line, testDeclaration.end.line).some(line => selectedLines.has(line));
```

lib/line-numbers.js
```diff
@@ -0,0 +1,148 @@
...
+	return sortNumbersAscending(distinctArray(flatten(
+		filter
+			.filter(({pattern, lineNumbers}) => lineNumbers && micromatch.isMatch(normalizedFilePath, pattern))
+			.map(({lineNumbers}) => lineNumbers)
+	)));
+}
+
+exports.getApplicableLineNumbers = getApplicableLineNumbers;
+
+function initializeLineNumberSelector(file, lineNumbers = []) {
```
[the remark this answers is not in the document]
ulken: Went ahead and did this

lib/line-numbers.js
ulken: Removed comment

lib/line-numbers.js
```diff
@@ -0,0 +1,148 @@
...
+				column: sourceCallSite.getColumnNumber() - 1 // Use 0-indexed columns.
+			};
+
+			// Find all calls that close over the test declaration.
+			const enclosingCalls = allCallLocations.filter(loc => {
+				if (loc.start.line > start.line || loc.end.line < start.line) {
+					return false;
+				}
+
+				if (loc.start.line === start.line && loc.start.column > start.column) {
```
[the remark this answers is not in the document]
ulken: Tests are passing, so guess so

lib/line-numbers.js
ulken: Done

lib/line-numbers.js
ulken: Went ahead and did this. More dense and easier to read IMO.

ulken:
> If no tests are selected, should that cause an exit code of 1?
I think so. It makes no sense to not have any tests selected when you are actively trying to pinpoint one or a few.

ulken:
> Work out how @ava/babel and @ava/typescript could participate in this

I haven't looked into how these integrate, but basically the only thing _should_ be needed is to pass the appropriate parser (Babel/TypeScript) to `recast.parse()`.

novemberborn:
Thanks for the updates @ulken!

I've figured out why the tests are failing… `recast` is adjusting the AST locations based on the tab indentation. Which can be worked around by setting the tab width, but it does not instill confidence. Will explore other options.

novemberborn:
@ulken I think we're there!

I, uhm, refactored again, sorry about that. But it's clearer now how we might integrate with `@ava/babel` and `@ava/typescript`. I'll open issues for that before this ships.

I had to switch to `acorn` which, as far as I can tell, does not add opinions to the syntax tree.

AVA now prints warnings when we can't parse the AST. Combined with the error message when no tests are selected I think that gets us to a good place.

The source map handling may break when `source-map-support` has already been loaded and is correcting stack traces. But I want to kick off some separate work to deal with that, so I've just added a comment for now.

The reporter tests now include the failure cases.

Let me know what you think, and then we can ship this!

novemberborn:
Phew tests are passing again! The one failing check is due to a dependency that broke itself.

lib/run-status.js
```diff
@@ -48,7 +48,9 @@ class RunStatus extends Emittery {
 			skippedTests: 0,
 			todoTests: 0,
 			uncaughtExceptions: 0,
-			unhandledRejections: 0
+			unhandledRejections: 0,
+			runOnlySelectedLines: false,
```
ulken, line 52: I know I added it, but with regards to the other properties, is this in correct tense?

lib/worker/line-numbers.js
```diff
@@ -0,0 +1,90 @@
+function parse(file) {
+	const fs = require('fs');
+	const acorn = require('acorn');
+	const walk = require('acorn-walk');
+
+	const ast = acorn.parse(fs.readFileSync(file, 'utf8'), {
```
ulken, line 6: Did you try to use the `acorn` parser of `recast`? I'm thinking maybe it's the `esprima` parser which handles tabs funkily. Might be wrong, though, haven't checked/tested.

The reason I'm asking is I don't see how we would handle `typescript`/`babel` now?

Plus, you wouldn't need `acorn-walk`, nor do the sorting, right?

lib/worker/line-numbers.js
```diff
@@ -0,0 +1,90 @@
...
+		}
+
+		return true;
+	});
+
+	// Locations should be sorted by source order, so the last span must be the test.
+	return spans.pop();
+}
+
+const range = (start, end) => new Array(end - start + 1).fill(start).map((element, index) => element + index);
```
ulken, line 52: Break this guy out (since it's duplicated)? Or three strikes?

lib/worker/subprocess.js
```diff
@@ -41,12 +42,26 @@ ipc.options.then(async options => {
...
+		checkSelectedByLineNumbers = () => false;
+	}
+
 	const runner = new Runner({
 		experiments: options.experiments,
 		failFast: options.failFast,
 		failWithoutAssertions: options.failWithoutAssertions,
 		file: options.file,
 		match: options.match,
+		checkSelectedByLineNumbers,
```
ulken, line 64: Alphabetical ordering? If not, move shorthand property to bottom?

ulken:
> @ulken I think we're there!

Wow, cool! 

> I, uhm, refactored again, sorry about that. But it's clearer now how we might integrate with `@ava/babel` and `@ava/typescript`. I'll open issues for that before this ships.

Not much of my original code left 😛. I had a question about that, so looking forward to see thoughts on those issues.
 
> I had to switch to `acorn` which, as far as I can tell, does not add opinions to the syntax tree.

I added a comment around that.

> AVA now prints warnings when we can't parse the AST. Combined with the error message when no tests are selected I think that gets us to a good place.

Nice!

> The source map handling may break when `source-map-support` has already been loaded and is correcting stack traces. But I want to kick off some separate work to deal with that, so I've just added a comment for now.

Good thinking.

> Let me know what you think, and then we can ship this!

Looks good! Left a few comments.

lib/run-status.js
novemberborn: Yea I'll rename.

lib/worker/line-numbers.js
novemberborn: `recast` is designed for rewriting code, but all we need is the locations of call expressions. `@ava/babel` already has access to the Babel parser, and we can do something with `@ava/typescript` as well.

lib/worker/line-numbers.js
novemberborn: Yea I didn't think it was worth extracting at this point.

lib/worker/subprocess.js
novemberborn: Good spot, thanks.

novemberborn:
> Not much of my original code left 😛.

Sometimes it's more about the journey than the code we wrote along the way 😉 We wouldn't have gotten to this point without your hard work.

lib/worker/line-numbers.js
ulken: Fair enough!

ulken:
@novemberborn Alright, so open issues for `@ava/babel` and `@ava/typescript`, then merge?

novemberborn:
Yep I'll do that. Hoping to get this release out today, along with #2459.

Issue #2212, opened 2019-08-14
ObservableLike type definition is incorrect


micaelmbagira:
## Description

As I commented in this closed issue https://github.com/avajs/ava/issues/1685#issuecomment-520885556, the `ObservableLike` definition in ava 2.2.0 is incorrect. It expects the property `[Symbol.observable]()` but this property is not in `rxjs` Observable type.

### Test Source

```ts
import { of } from 'rxjs';

test('foo', t => {
  return of(3).pipe(tap(() => t.pass()));
});
```

### Error Message & Stack Trace

``` 
  error TS2345: Argument of type '"foo"' is not assignable to parameter of type 'OneOrMoreMacros<[(t: ExecutionContext<{}>) => Observable<number>], {}>'.

  test('foo', t => {
         ~~~~~
```

Removing the `[Symbol.observable](): ObservableLike;` solves it though....

When trying to enforce the `ObservableLike` type with `return of(3).pipe(tap(() => t.pass())) as ObservableLike;`. TypeScript would throw

> Conversion of type 'Observable<number>' to type 'ObservableLike' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
    Property '[Symbol.observable]' is missing in type 'Observable<number>' but required in type 'ObservableLike'

So basically, `[Symbol.observable]()` is not in rxjs `Observable` type definition. Can we then reopen this issue?

### Expected behaviour

The returned `Observable` should match the `ObservableLike` type and the `t => ...` should not infer the type `OneOrMoreMacros`.

novemberborn:
> `t => ...` should not infer the type `OneOrMoreMacros`.

I think this is just the type checker doing its thing.

> The returned `Observable` should match the `ObservableLike` type

@sindresorhus?

sindresorhus:
> It expects the property [Symbol.observable]() but this property is not in rxjs Observable type.

This is the RxJS `ObservableLike` type:

```ts
export interface Subscribable<T> {
    subscribe(observer?: PartialObserver<T>): Unsubscribable;
    /** @deprecated Use an observer instead of a complete callback */
    subscribe(next: null | undefined, error: null | undefined, complete: () => void): Unsubscribable;
    /** @deprecated Use an observer instead of an error callback */
    subscribe(next: null | undefined, error: (error: any) => void, complete?: () => void): Unsubscribable;
    /** @deprecated Use an observer instead of a complete callback */
[...]

PR #2308, opened 2019-12-11
Type observables based on how they're consumed by AVA

micaelmbagira:
This PR aims to fix https://github.com/avajs/ava/issues/2212

**Description**
The following snippet breaks because of types conflict

```ts
import { of } from 'rxjs';

test('foo', t => {
  return of(3).pipe(tap(() => t.pass()));
});
```

The reasons are the following:

1. `ObservableLike` has the [Symbol.observable](https://github.com/ReactiveX/rxjs/blob/56cbd220dac11ec58751e15a23a56d6747964df2/src/internal/types.ts#L55-L57) tag 
2. the inferred type of `of(3)` is `Observable`
3. in RxJS, `Observable` [does not have the `[Symbol.observable]()` property](https://github.com/ReactiveX/rxjs/blob/56cbd220dac11ec58751e15a23a56d6747964df2/src/internal/Observable.ts#L303), so it breaks.

So ideally, [`ImplementationResult`](ImplementationResult type) type  should be [`SubscribableOrPromise`](https://github.com/ReactiveX/rxjs/blob/56cbd220dac11ec58751e15a23a56d6747964df2/src/internal/types.ts#L37)

**Solution**

- added a `SubscribableLike` interface similar to [RxJS type definitions](https://github.com/ReactiveX/rxjs/blob/56cbd220dac11ec58751e15a23a56d6747964df2/src/internal/types.ts#L37-L50) 
- the `ObserverLike` interface is also [similar to RxJS types](https://github.com/ReactiveX/rxjs/blob/56cbd220dac11ec58751e15a23a56d6747964df2/src/internal/types.ts#L84-L89) but as suggested by @novemberborn, I took the signature that is used when [ava subscribes to it](https://github.com/sindresorhus/observable-to-promise/blob/f886c9335595c1b9265a24d62825427e17fc45c2/index.js#L13)

@novemberborn I wanted to add a unit test but I would need to import `rxjs` as a devDependency, is it worth it?


novemberborn:
> I wanted to add a unit test but I would need to import `rxjs` as a devDependency, is it worth it?

Yes that'd be great.



> * the `ObserverLike` interface is also [similar to RxJS types](https://github.com/ReactiveX/rxjs/blob/56cbd220dac11ec58751e15a23a56d6747964df2/src/internal/types.ts#L84-L89) but as suggested by @novemberborn, I took the signature that is used when [ava subscribes to it](https://github.com/sindresorhus/observable-to-promise/blob/f886c9335595c1b9265a24d62825427e17fc45c2/index.js#L13)

We need to replace `observable-to-promise`, since that is looking for symbols and whatnot rather than just calling `subscribe`.

novemberborn:
> > * the `ObserverLike` interface is also [similar to RxJS types](https://github.com/ReactiveX/rxjs/blob/56cbd220dac11ec58751e15a23a56d6747964df2/src/internal/types.ts#L84-L89) but as suggested by @novemberborn, I took the signature that is used when [ava subscribes to it](https://github.com/sindresorhus/observable-to-promise/blob/f886c9335595c1b9265a24d62825427e17fc45c2/index.js#L13)
> 
> We need to replace `observable-to-promise`, since that is looking for symbols and whatnot rather than just calling `subscribe`.

I've pushed commits to address this.

> > I wanted to add a unit test but I would need to import `rxjs` as a devDependency, is it worth it?
> 
> Yes that'd be great.

We have existing tests using `zen-observable`, but I wouldn't mind another integration test for `rxjs`.

micaelmbagira:
@novemberborn thanks. I'll try to add the `rxjs` test tomorrow

novemberborn:
> Why v3 though? This fixes a bug introduced in v2.

It's going into the v3 release.

PR #2320, opened 2019-12-27
Add Selenium WebDriverJS recipe

### LINKED ISSUE #1023: Recipe for using AVA with Selenium
It's a common need. We should have a [recipe](https://github.com/avajs/ava/tree/master/docs/recipes) to make it easy for users to get started.

I think [Webdriver.IO](http://webdriver.io) would be the best choice.

aptester:
Hi,
This pull request fixes #1023. It adds the recipe for using Selenium WebDriverJs (official JavaScript version) with AVA for testing web applications.

### DIFF
--- a/docs/recipes/testing-with-selenium-webdriverjs.md
+++ b/docs/recipes/testing-with-selenium-webdriverjs.md
@@ -1,41 +1,39 @@
-# Setting up AVA with Selenium WebDriverJs
+# Setting up AVA with Selenium WebDriverJS
 
-This recipe shows you how to use the Selenium WebDriverJs (official JavaScript implementation) with AVA to test web applications.
+This recipe shows you how to use the Selenium WebDriverJS (official JavaScript implementation) with AVA to test web apps.
 
 ## Setup
 
-This recipe uses the following libraries:
+This recipe uses the following packages:
 
 1. [selenium-webdriver](https://www.npmjs.com/package/selenium-webdriver)
 2. [chromedriver](https://www.npmjs.com/package/chromedriver)
 
-Install the above two libraries by running the following commands:
+Install them with:
 
 ```console
-$ npm install selenium-webdriver
-$ npm install chromedriver
+$ npm install selenium-webdriver chromedriver
 ```
 
-As part of this recipe, we will use Selenium to verify web searches on [Bing](http://www.bing.com) and [Google](http://www.google.com)
+As part of this recipe, we will use Selenium to verify web searches on [Bing](https://www.bing.com) and [Google](https://www.google.com)
 
 ## Test files
 
-Create the following files
+Create the following files:
 
-`./test/bingtest.js`
+- `./test/bingtest.js`
+- `./test/googletest.js`
 
-`./test/googletest.js`
-
-In both files, let's first include the libraries
+In both files, let's first include the packages:
 
 ```js
 import test from 'ava';
-import {Builder, By, Key, until} from'selenium-webdriver';
+import {Builder, By, Key, until} from 'selenium-webdriver';
 
-require("chromedriver");
+require('chromedriver');
 ```
 
-In the `bingtest.js` file, add the following code, which tests whether searching for 'webdriver' on Bing, returns results
+In the `bingtest.js` file, add the following code, which tests whether searching for `webdriver` on Bing, returns results.
 
 ```js
 test('Bing Search', async t => {
@@ -89,7 +87,9 @@ We can run all the tests, by the following command
 ```console
 npm test
 ```
-Since I ran the above command on a laptop with 4 CPU cores, AVA ran both `bingtest.js` and `googletest.js` in parallel. See below output
+
+Since I ran the above command on a laptop with 4 CPU cores, AVA ran both `bingtest.js` and `googletest.js` in parallel. See below output.
+
 ```console
 DevTools listening on ws://127.0.0.1:49852/devtools/browser/adfcad21-9612-46ff-adc3-09adc0737f4a
 
@@ -101,6 +101,7 @@ DevTools listening on ws://127.0.0.1:49855/devtools/browser/8f6b7206-ea2b-4d41-b
 ```
 
 Dependending upon the number of CPU cores, those many test files will run at the same time. This count can be changed either via terminal (--concurrency or -c) or via the `"ava"` section in the package.json. For e.g.
+
 ```js
 "ava":{    
     "concurrency": 1,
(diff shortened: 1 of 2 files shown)


aptester:
> 
> 
> I like your receipt.

@CrispusDH Thanks. Your feedback to the earlier recipe made sense to me as the recipe should represent a 'real-world' use case for folks to find it useful.

docs/recipes/testing-with-selenium-webdriverjs.md
```diff
@@ -0,0 +1,131 @@
...
+import {Builder, By, Key, until} from 'selenium-webdriver';
+
+require('chromedriver');
+```
+
+In the `bingtest.js` file, add the following code, which tests whether searching for `webdriver` on Bing, returns results.
+
+```js
+test('Bing Search', async t => {
+	let keyword = "webdriver";
```
sindresorhus: Use `const` everywhere. And use single-quotes.

docs/recipes/testing-with-selenium-webdriverjs.md
```diff
@@ -0,0 +1,131 @@
+# Setting up AVA with Selenium WebDriverJS
```
sindresorhus: The title here should be the same as in readme.md

docs/recipes/testing-with-selenium-webdriverjs.md
```diff
@@ -0,0 +1,131 @@
...
+1. [selenium-webdriver](https://www.npmjs.com/package/selenium-webdriver)
+2. [chromedriver](https://www.npmjs.com/package/chromedriver)
+
+Install them with:
+
+```console
+$ npm install selenium-webdriver chromedriver
+```
+
+As part of this recipe, we will use Selenium to verify web searches on [Bing](https://www.bing.com) and [Google](https://www.google.com)
```
novemberborn:
```suggestion
As part of this recipe, we will use Selenium to verify web searches on [Bing](https://www.bing.com) and [Google](https://www.google.com).
```

docs/recipes/testing-with-selenium-webdriverjs.md
```diff
@@ -0,0 +1,131 @@
...
+Create the following files:
+
+- `./test/bingtest.js`
+- `./test/googletest.js`
+
+In both files, let's first include the packages:
+
+```js
+import test from 'ava';
+import {Builder, By, Key, until} from 'selenium-webdriver';
```
novemberborn:
```suggestion
const {Builder, By, Key, until} = require('selenium-webdriver');
```

docs/recipes/testing-with-selenium-webdriverjs.md
```diff
@@ -0,0 +1,131 @@
...
+
+Create the following files:
+
+- `./test/bingtest.js`
+- `./test/googletest.js`
+
+In both files, let's first include the packages:
+
+```js
+import test from 'ava';
```
novemberborn: For now let's use CJS syntax:

```suggestion
const test = require('ava');
```

docs/recipes/testing-with-selenium-webdriverjs.md
```diff
@@ -0,0 +1,131 @@
...
+	await driver.findElement(By.name("q")).sendKeys(keyword + Key.ENTER);
+	await driver.wait(until.titleIs(keyword + " - Bing"));
+	t.true((await driver.findElements(By.css("#b_content #b_results li"))).length > 0);
+	await driver.close();
+}); 
+```
+
+In the `googletest.js` file, instead of a single test, lets add two tests, one each for the terms 'webdriver' and 'avajs'.
+
+Since we would like to initialize the webdriver before each test, we use the [`beforeEach` and `afterEach`](https://github.com/avajs/ava/blob/master/docs/01-writing-tests.md#before--after-hooks) hooks to setup and teardown the driver respectively. Using these hooks, helps reduce the amount of code we would write in each test()
```
novemberborn:
```suggestion
Since we would like to initialize the webdriver before each test, we use the [`beforeEach` and `afterEach`](https://github.com/avajs/ava/blob/master/docs/01-writing-tests.md#before--after-hooks) hooks to setup and teardown the driver respectively. Using these hooks, helps reduce the amount of code we would write in each `test()`.
```

docs/recipes/testing-with-selenium-webdriverjs.md
```diff
@@ -0,0 +1,131 @@
...
+```js
+test.beforeEach(async t => {
+	t.context.driver = new Builder().forBrowser("chrome").build();
+	await t.context.driver.get("http://www.google.com");
+});
+
+test.afterEach("cleanup", async t => await t.context.driver.close());
+```
+
+Now lets add the test code,
```
novemberborn:
```suggestion
Now lets add the test code:

```

docs/recipes/testing-with-selenium-webdriverjs.md
```diff
@@ -0,0 +1,131 @@
...
+}
+```
+
+## Running the tests
+
+We can run all the tests, by the following command
+
+```console
+npm test
+```
```
novemberborn: You should assume a certain level of familiarity with AVA. So perhaps say "Now run tests (`npx ava`)".

docs/recipes/testing-with-selenium-webdriverjs.md
```diff
@@ -0,0 +1,131 @@
...
+
+Since I ran the above command on a laptop with 4 CPU cores, AVA ran both `bingtest.js` and `googletest.js` in parallel. See below output.
```
novemberborn: I'd replace all this with a reminder that test files run in parallel, and that the concurrency can be configured (linking to the configuration doc).

docs/recipes/testing-with-selenium-webdriverjs.md
```diff
@@ -0,0 +1,131 @@
...
+```js
+"ava":{    
+    "concurrency": 1,
+    "verbose": true
+  }
+```
+
+The `concurrency: 1` value will only allow AVA to run one file at a time. It however cannot control how many tests in that file can run at the same time.
+
+The `verbose: true` value will enable verbose output which helps readability IMO.
```
novemberborn: Sure, but let's keep opinion out of this :smile:

docs/recipes/testing-with-selenium-webdriverjs.md
aptester: Rephrased sentence by adding references to both the command-line and configuration(latest one) docs.

docs/recipes/testing-with-selenium-webdriverjs.md
aptester: Understood. Removed opinion.

docs/recipes/testing-with-selenium-webdriverjs.md
aptester: done

novemberborn:
I think this recipe is better without the parts on parallel execution or reporter verbosity. It should be focused on using Selenium instead.

PR #2329, opened 2020-01-02
Set a default timeout of 10 seconds

### LINKED ISSUE #2310: Set default timeout / ensure worker exits after unhandled rejections
I have tests which are executed in CI. Sometime it just hangs a lot of time cause test never end due to an uncaught promise error.

If i run the server in production it should crash ([but currently it do not](https://github.com/nodejs/node/issues/20392)).

We test in CI to prevent unexpected bug. Ava currently wait and consume CI time for unhandled async rejections.

It would save cost if ava could fail test instantly.

novemberborn:
Now that our timeout implementation is pretty good, we can set a default timeout.

10 seconds seems reasonable to me.

We exit the worker process after the timeout is reached, which should help with issues like #2310.

Fixes #2310.


novemberborn:
I'm including this in v3. In the future we could still increase the timeout in a non-breaking release.

PR #2333, opened 2020-01-03
Pooled Workers: api.js refactor

abrenneke:
Hi! I'm attempting to take over what was started in #2011 by @ijjk! I'm going to be using @ijjk's branch as a reference while I make the changes and keep in mind all of the comments in #2011.

Since the original PR was pretty large, I'm going to try doing this more piecemeal instead with multiple pull requests, and hopefully less drastic changes in some areas.

This initial pull request refactors the `api.js` file so that it doesn't always expect a forked worker process per test. That functionality has been moved to `fork-file-worker.js` which is just called by `api.js` right now. I'm relying more on `run-status.js` as an event emitter for passing information around.

Question: Why is there only one event emitted by `run-status.js`? It's an `Emittery`, so in theory it could just emit a bunch of events instead of one object with a `type` property. It seems to me that it may be simpler to use it as a normal event emitter?

I made some cleanup to `api.js` as well, I can totally revert some of it if you feel it's unnecessary. All tests are currently passing. Do you think any new ones are necessary?


novemberborn:
> Since the original PR was pretty large, I'm going to try doing this more piecemeal instead with multiple pull requests, and hopefully less drastic changes in some areas.

This sounds great, thanks for picking this up.

We should put any new worker setup behind an [experiment](https://github.com/avajs/ava/blob/33075127533e27d9eda1657271968ce00a03009f/lib/load-config.js#L9). That way we can iterate more quickly and test stuff in the wild, even if test coverage and documentation is incomplete.

> Question: Why is there only one event emitted by `run-status.js`? It's an `Emittery`, so in theory it could just emit a bunch of events instead of one object with a `type` property. It seems to me that it may be simpler to use it as a normal event emitter?

It's all kinda legacy, with tweaks made over time. Now that we're targeting Node.js 10 at a minimum I wouldn't be opposed to replacing it all by an async iterator, even.

abrenneke:
> We should put any new worker setup behind an [experiment](https://github.com/avajs/ava/blob/33075127533e27d9eda1657271968ce00a03009f/lib/load-config.js#L9). That way we can iterate more quickly and test stuff in the wild, even if test coverage and documentation is incomplete.

Yep, that's the plan. First step is a bit of refactoring so that that can be supported.

> It's all kinda legacy, with tweaks made over time. Now that we're targeting Node.js 10 at a minimum I wouldn't be opposed to replacing it all by an async iterator, even.

Cool, I'm messing around with this in a separate branch.

Are the windows errors a real problem? I can't think of any changes I made that would only affect windows but I'll double check.

novemberborn:
> > We should put any new worker setup behind an [experiment](https://github.com/avajs/ava/blob/33075127533e27d9eda1657271968ce00a03009f/lib/load-config.js#L9). That way we can iterate more quickly and test stuff in the wild, even if test coverage and documentation is incomplete.
> 
> Yep, that's the plan. First step is a bit of refactoring so that that can be supported.

👍 

> Are the windows errors a real problem? I can't think of any changes I made that would only affect windows but I'll double check.

I looked at one and it's not the general intermittent failure.

I haven't actually looked at your changes yet, sorry.

novemberborn:
@abrenneke I've fixed the merge conflicts with `master`. The changes all look good — what would you like to do next with this PR?

novemberborn:
That said the test failures on Windows look genuine.

abrenneke:
@novemberborn I'll fix up the windows problem!

novemberborn:
@abrenneke would you have time to pick this up again?

novemberborn:

PR #2336, opened 2020-01-04
Initialize Chalk in worker processes with same level as main process

novemberborn:
This Chalk upgrade is proving more tricky than expected 😉


sindresorhus:
Let me know if there’s anything we could do in Chalk itself to make it easier.

novemberborn:
> Let me know if there’s anything we could do in Chalk itself to make it easier.

Nah, it's all just subtle stuff that happened to work before and now doesn't. We'll get there.

Issue #2215, opened 2019-08-18
Do we need to fake TTY in the worker processes?


novemberborn:
We fake the TTY in worker processes, forwarding best we can the settings with which AVA is run. This code is increasingly hard to maintain.

I'm not sure why we're even doing this in the first place. What are the consequences of launching the worker processes without any TTY? Does anybody rely on this?

Presumably we'd keep forwarding color support as we do now.

sindresorhus:
I rely on this in a lot of tests. We fake TTY as the fact that AVA spawns multiple processes should just be an implementation detail.

novemberborn:
Are you suggesting that if we had an execution mode where the tests didn't run in a child process, we could remove the faking?

sindresorhus:
Yes, although I'm sure people rely on it when they need multi-process execution too, so it might be a good idea to keep it for the sake of a good developer experience.

I wonder if there's a way we could get Node.js to either have a way to fake TTY or expose more code so we don't have to duplicate as much.

novemberborn:
> I wonder if there's a way we could get Node.js to either have a way to fake TTY or expose more code so we don't have to duplicate as much.

The constructors are exposed (https://github.com/nodejs/node/blob/master/lib/tty.js) so perhaps we could?

novemberborn:
I'm adding the v3 milestone to hopefully explore whether we can make the faking opt-in instead.

novemberborn:
Our tests pass without a fake TTY. I suggest we remove the faking, and then publish our implementation as a separate package which you can load in the worker processes.

novemberborn:
TTY has been removed in #2337. I've opened https://github.com/avajs/ava/issues/2343 to deal with its replacement.

PR #2337, opened 2020-01-04
Remove the faking of the TTY in worker processes

novemberborn:
See discussion in https://github.com/avajs/ava/issues/2215.

I'll kick off the process for extracting the faking logic into a separate module, to be shipped before the final 3.0.0 release.


novemberborn:
I just realized we can replace the global `console` with one that has color mode enabled, even if the process is not a TTY. Win!

PR #2338, opened 2020-01-05
Update debugging-with-vscode.md

Menighin:
So, I was trying debugging with the last version of this and was receiveing an error. Something like `debug not found`.
I made it work with this config.
Maybe it was outdated...?


novemberborn:
Nope, it describes `master` not AVA 2. It's hard to make this clear, though there is a comment at the top.

Thanks for your PR though!

wtrocki:
Took me a while to get why README.md documentation is not working. It will be cool to update links to latest release rather than master

PR #2342, opened 2020-01-05
Remove legacy .throws() and .throwsAsync() interfaces

### LINKED ISSUE #2301: Remove legacy t.throws() interface
I (and other people) often make this mistake:

```js
t.throws(() => {...}, TypeError, 'error message here');
```

while it should be:

```js
t.throws(() => {...}, {
	instanceOf: TypeError,
	message: 'error message here'
);
```

I propose something like this:

```js
t.throws(() => {...}, TypeError, 'error message here').describe('no error received');

// Actual solution
t.is(true, false, 'check failed');

// Expected solution
t.is(true, false).describe('check failed');
```

novemberborn:
Fixes #2301.


szmarczak:
Thank you very much! 🙌

PR #2352, opened 2020-01-06
Use question mark to indicate optional argument in docs

sindresorhus:
I have switched to this in my own docs as many users were confused about the brackets `[message]` and thought it meant an array. Question mark is already commonly used to indicate optional arguments, for example, in TypeScript.

### DIFF
--- a/docs/03-assertions.md
+++ b/docs/03-assertions.md
@@ -163,51 +163,51 @@ test('custom assertion', t => {
 
 ## Built-in assertions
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
It's is what I could find in this repo, but I'll go through our other repos too.

sindresorhus:
> It's is what I could find in this repo, but I'll go through our other repos too.

✅

PR #2358, opened 2020-01-10
Ensure t.assert() counts as a passed assertion

### LINKED ISSUE #2354: Passing t.assert() does not count as a passed assertion
This is a regression in the upcoming v3 release.

The `t.assert()` implementation *without* power-assert in broken:

https://github.com/avajs/ava/blob/090884b5040a1d4b02a18939245f7c2305228e00/lib/assert.js#L856

After calling `fail()` we need to `return`. Then outside of that if-condition, we need to call `pass()`.

See https://github.com/avajs/ava/blob/090884b5040a1d4b02a18939245f7c2305228e00/lib/assert.js#L837:L840 for an example.

Our tests don't cover this but that'll be a lot harder to address.

theashraf:
a fix to `t.assert()` implementation without power-assert.

closes #2354

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -853,7 +853,10 @@ class Assertions {
 						operator: '!!',
 						values: [formatWithLabel('Value is not truthy:', actual)]
 					}));
+					return;
 				}
+
+				pass();
 			});
 		} else {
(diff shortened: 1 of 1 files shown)


novemberborn:
Thanks @theashraf!

PR #2369, opened 2020-01-17
Fix VSCode debugging instructions

novemberborn:
`configurations` is an array, not an object.


novemberborn:
While I'm here, if you're pre-compiling your tests then this doesn't work either.

Let's add a note on that. A pattern like this would work as long as that matches a single file: `"build/**/${fileBasenameNoExtension}.*"`

PR #2379, opened 2020-01-25
Rudimentary support for loading TypeScript test files

novemberborn:
This adds very rudimentary support for loading TypeScript test files, as discussed in https://github.com/avajs/ava/issues/1109#issuecomment-557805037.

Leaving this as a draft until `@ava/typescript` ships. See https://github.com/avajs/typescript/pull/1.


novemberborn:
See https://github.com/novemberborn/got/pull/1 for how this would be used.

PR #2381, opened 2020-01-26
types: fix parameter types of throws and throwsAsync

ajafff:
After upgrading to v3 I had to change every call-site of `t.throws` and `t.throwsAsync` because I only asserted the message.
I passed `undefined` as second argument as it's perfectly fine according to the type declarations.
After that all tests failed, telling me to pass `null` instead. So I had to go through all tests again and change `undefined` to `null`.

This PR changes the type declarations. Now you are no longer allowed to pass `undefined` as second argument.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -250,13 +250,13 @@ export interface ThrowsAssertion {
 	/**
 	 * Assert that the function throws [an error](https://www.npmjs.com/package/is-error). If so, returns the error value.
 	 */
-	<ThrownError extends Error>(fn: () => any, expectations?: null, message?: string): ThrownError;
+	<ThrownError extends Error>(fn: () => any): ThrownError;
 
 	/**
 	 * Assert that the function throws [an error](https://www.npmjs.com/package/is-error). If so, returns the error value.
 	 * The error must satisfy all expectations.
 	 */
-	<ThrownError extends Error>(fn: () => any, expectations: ThrowsExpectation, message?: string): ThrownError;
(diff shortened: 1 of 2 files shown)


novemberborn:
> I passed `undefined` as second argument as it's perfectly fine according to the type declarations.
> After that all tests failed, telling me to pass `null` instead.

Oh interesting. I think the original intent of requiring `null` was in case of variables left accidentally undefined. Your test would pass but it wouldn't assert what you thought it did.

Now that we're always passing an expectation object that's less of a concern. I think we should change the assertion to allow `undefined` expectations. Would you be interested in doing that instead?

---



> After upgrading to v3 I had to change every call-site of `t.throws` and `t.throwsAsync` because I only asserted the message.
> I passed `undefined` as second argument as it's perfectly fine according to the type declarations.

Without seeing your code, I don't think this does what you think it does. The message is what's shown when your assertion fails. It's **not** used to compare against the error message. Use `t.throwsAsync(promise, { message: 'My expected message' })` instead.

ajafff:
> Without seeing your code, I don't think this does what you think it does. The message is what's shown when your assertion fails. It's not used to compare against the error message.

You're right. I totally forgot it's the custom assertion failure message.

> Now that we're always passing an expectation object that's less of a concern. I think we should change the assertion to allow undefined expectations. Would you be interested in doing that instead?

I'll create a new PR with that change and close this one afterwards.

novemberborn:
I've opened a new issue to track this: https://github.com/avajs/ava/issues/2401

Issue #2344, opened 2020-01-05
Support loading .mjs test files


novemberborn:
AVA now recognizes `.mjs` test files, but refuses to load them. We should!

See the code here: https://github.com/avajs/ava/blob/54ff13036517e7dff3d95b7afe4e70254b1dbe7a/lib/worker/subprocess.js#L131:L137

Of course `import()` is only available on Node.js 13. This should fail gracefully on Node.js 10 and 12.

novemberborn:
This is now available in AVA 3.3.0. Thank you @arlac77!

PR #2382, opened 2020-01-26
Support loading ESM test files

arlac77:
Fixes https://github.com/avajs/ava/issues/2344.

Brute force try import(ref)
at least works for me

How to make xo not complain about fully qualified imports ?

```js
import test from '../../index.js';
```


sindresorhus:
You need to `await` the import call. Otherwise, the try/catch has no effect.

arlac77:
> You need to `await` the import call. Otherwise, the try/catch has no effect.

the try catch is there to prevent the unknown import() function error (node < 10?)
an fall through to the former error handling

test/integration/assorted.js
```diff
@@ -164,18 +164,30 @@ test('selects .cjs test files', t => {
 	});
 });
 
-test('refuses to load .mjs test files', t => {
+test('refuses to load .mjs test files (node < 13)', t => {
```
novemberborn: These tests should be updated to reflect what happens when ESM is not available.

novemberborn:
> How to make xo not complain about fully qualified imports ?
> 
> ```js
> import test from '../../index.js';
> ```

The XO output should contain the failing rule. Then after that line add `// eslint-disable-line the-failing-rule` and the rule should be ignored at that point.

test/fixture/esm/test.js
```diff
@@ -1,4 +1,4 @@
-import test from '../../..';
+import test from '../../../index.js';
```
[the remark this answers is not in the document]
arlac77: node build in esm loader does not support auto completion to index.js

the tests pass (only xo complains)

```sh
npm run test:tap
```

Or take a look at my playground project
[uti](https://github.com/arlac77/uti.git)
there this pr is working on a plain node (13) without any additional esm loader

lib/worker/subprocess.js
```diff
@@ -137,6 +137,10 @@ ipc.options.then(async options => {
 	const load = async ref => {
 		for (const extension of extensionsToLoadAsModules) {
 			if (ref.endsWith(`.${extension}`)) {
+				try {
+					return import(ref);
+				} catch (_) {}
```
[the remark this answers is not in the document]
arlac77: Looks like [optional catch binding](https://kangax.github.io/compat-table/es2016plus/) is only present in node >10.9

test/fixture/esm/test.js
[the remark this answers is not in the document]
novemberborn: Yup, I'm just suggesting to adjust to what Node.js expects, irrespective of how XO is configured.

lib/worker/subprocess.js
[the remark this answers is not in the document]
novemberborn: Yup. But our minimum version is 10.18 so it's safe to use.

test/integration/assorted.js
arlac77: You mean two sets of tests one for when ESM is present one without ?

I see the following matrix:

- node 10  - 'AVA cannot yet load ESM files'
- node 12  - 'AVA cannot yet load ESM files'
- node 12 with --experimental-modules - '1 test passed'
- node 13  - '1 test passed'

Or should the tests be conditionally executed based on the presence of the import() function ?

novemberborn: You can test for Node versions.

arlac77: So I did,
different outcomes based on the following expression:
```js
Number.parseFloat(process.version.slice(1)) >= 13
```

lib/worker/subprocess.js
```diff
@@ -137,6 +137,11 @@ ipc.options.then(async options => {
 	const load = async ref => {
 		for (const extension of extensionsToLoadAsModules) {
 			if (ref.endsWith(`.${extension}`)) {
+				try {
+					return await import(ref); // eslint-disable-line no-await-in-loop
+				} catch {
```
novemberborn: This swallows any errors thrown in the `ref` module itself.

Instead you should detect whether `import()` works and check that before loading `ref`. Then you can send the appropriate "ESM is not available in this Node.js version" kinda error, or have AVA report the error from the module.

test/integration/assorted.js
```diff
@@ -164,18 +164,30 @@ test('selects .cjs test files', t => {
 	});
 });
 
-test('refuses to load .mjs test files', t => {
+test('load .mjs test files (when node supports it)', t => {
 	execCli('mjs.mjs', (err, stdout) => {
-		t.ok(err);
-		t.match(stdout, /AVA cannot yet load ESM files/);
-		t.end();
+		if (Number.parseFloat(process.version.slice(1)) >= 13 && process.platform !== 'win32') {
```
novemberborn: We should figure out why it's failing on Windows. It can't be a Node.js issue.

lib/worker/subprocess.js
arlac77: I am having difficulties to understand the windows problem.
I would like to see the error coming out of the CI build.

test/integration/assorted.js
arlac77: on windows a ERR_UNSUPPORTED_ESM_URL_SCHEME is thrown maybe drive letters or backward slashes have to be removed

novemberborn:
@arlac77 great find with the path vs URL stuff!

I think there's two remaining issues:

* When `import()` is not available, the error message shown should communicate that. Currently it says AVA doesn't support ESM.
* Errors thrown in the imported module should be reported, like they are for CJS modules.

arlac77:
> * like they are for CJS modules.

Where can I find CJS module error code ?

novemberborn:
> > * like they are for CJS modules.
> 
> Where can I find CJS module error code ?

No, the issue is the `catch {}` statement. You're assuming the only thrown error is because Node.js does not support ESM. But you're inadvertently swallowing errors thrown by the module.

Instead I think the safest way to deal with this is to explicitly detect ESM support by importing a known `.mjs` file included with AVA itself. Then in the `load()` function you can check for support before calling `import()`. If it's missing, emit the internal error and exit. Otherwise import without swallowing errors.

lib/worker/subprocess.js
```diff
@@ -135,9 +136,19 @@ ipc.options.then(async options => {
 
 	let requireFn = require;
 	const load = async ref => {
+		let hasDynamicImport = false;
+		try {
+			await import('../extensions.js');
+			hasDynamicImport = true;
+		} catch {}
```
novemberborn: Let's move this out of the `load` function.

lib/worker/subprocess.js
```diff
@@ -135,9 +136,19 @@ ipc.options.then(async options => {
 
 	let requireFn = require;
 	const load = async ref => {
+		let hasDynamicImport = false;
+		try {
+			await import('../extensions.js');
```
novemberborn: Let's put in an empty `esm.mjs` file and import that instead.

lib/worker/subprocess.js
```diff
@@ -135,9 +136,19 @@ ipc.options.then(async options => {
...
+			hasDynamicImport = true;
+		} catch {}
+
 		for (const extension of extensionsToLoadAsModules) {
 			if (ref.endsWith(`.${extension}`)) {
-				ipc.send({type: 'internal-error', err: serializeError('Internal runner error', false, new Error('AVA cannot yet load ESM files.'))});
+				if (hasDynamicImport) {
+					return import(pathToFileURL(ref));
```
novemberborn:
```suggestion
					return await import(pathToFileURL(ref));
```

lib/worker/subprocess.js
```diff
@@ -135,9 +136,19 @@ ipc.options.then(async options => {
...
+				}
+
+				ipc.send({type: 'internal-error', err: serializeError('Internal runner error', false, new Error('Node version cannot load ESM files.'))});
```
novemberborn:
```suggestion
				ipc.send({type: 'internal-error', err: serializeError('Internal runner error', false, new Error('ESM files cannot be loaded in this version of Node.js.'))});
```

novemberborn:
Thanks @arlac77. This looks good at first glance, but it'll probably take until the weekend before I can take a closer look.

novemberborn:
@arlac77 this is great! I've made some tweaks and updated the documentation. Will merge when CI passes 🎊

lib/worker/subprocess.js
```diff
@@ -133,11 +134,27 @@ ipc.options.then(async options => {
...
+
+		return supportsESM();
+	};
+
 	let requireFn = require;
 	const load = async ref => {
 		for (const extension of extensionsToLoadAsModules) {
 			if (ref.endsWith(`.${extension}`)) {
-				ipc.send({type: 'internal-error', err: serializeError('Internal runner error', false, new Error('AVA cannot yet load ESM files.'))});
+				if (await supportsESM()) { // eslint-disable-line no-await-in-loop
```
fisker: I think

```js
if (cachedSupportsESMResult || await supportsESM())
```

can make this loop faster.

PR #2389, opened 2020-02-02
Improve integration with TypeScript provider

novemberborn:
See https://github.com/avajs/typescript/pull/11.


novemberborn:
It's pretty hard to add to the watcher tests, so I'm forgoing that for now. Sorry!

PR #2393, opened 2020-02-09
Upgrade @ava/babel

novemberborn:
Recognize new @ava/require-precompiled dependency when beautifying stack traces.


izelnakri:
@novemberborn thank you very much for this timely release!

PR #2394, opened 2020-02-09
t.try() title improvements

novemberborn:
@qlonik I've been using `t.try()` for a while now. This is the only thing I can think of changing. Will try and get documentation done too.


qlonik:
Looks good to me.

I am not sure if there is a test which uses `t.try()` twice and passes multiple macros in each of the invocations. Maybe in that scenario the code `test.attemptCount + 1 + index` will be printing attempts with same title?

novemberborn:
Thanks! The counter itself is increments when the attempts are started, but yes I’ll add an explicit test.

PR #2396, opened 2020-02-11
Improve error message when `test.cb()` is used with promises

### LINKED ISSUE #2386: Improve message when test.cb() is used with promises
`test.cb()` is a legacy API. It can be used for testing callback APIs, but for anything more advanced we'd prefer you use `test()` with an async function.

We should update the error message here:

https://github.com/avajs/ava/blob/56338dcb77246b5eca3c01e24d1cb6d66da8f71f/lib/test.js#L597

It should say something like:

> Use `test.cb()` for legacy callback APIs. When using promises or async functions, use `test()`.

We should replace the "promises or async function" part with "observables" as appropriate.

toddkcarlson:
Fixes #2386

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -594,7 +594,7 @@ class Test {
 		if (this.metadata.callback) {
 			if (returnedObservable || returnedPromise) {
 				const asyncType = returnedObservable ? 'observables' : 'promises';
-				this.saveFirstError(new Error(`Do not return ${asyncType} from tests declared via \`test.cb(...)\`. Use \`test.cb(...)\` for legacy callback APIs. When using promises or async functions, use \`test(...)\`.`));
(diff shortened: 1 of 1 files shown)


lib/test.js
```diff
@@ -594,7 +594,7 @@ class Test {
 		if (this.metadata.callback) {
 			if (returnedObservable || returnedPromise) {
 				const asyncType = returnedObservable ? 'observables' : 'promises';
-				this.saveFirstError(new Error(`Do not return ${asyncType} from tests declared via \`test.cb(...)\`, if you want to return a promise simply declare the test via \`test(...)\``));
+				this.saveFirstError(new Error(`Do not return ${asyncType} from tests declared via \`test.cb(...)\`. Use \`test.cb(...)\` for legacy callback APIs. When using promises or async functions, use \`test(...)\`.`));
```
novemberborn:
```suggestion
				this.saveFirstError(new Error(`Do not return ${asyncType} from tests declared via \`test.cb(...)\`. Use \`test.cb(...)\` for legacy callback APIs. When using promises, observables or async functions, use \`test(...)\`.`));
```

What do you think?

novemberborn:
Thanks @toddkcarlson!

PR #2404, opened 2020-02-18
Improve ESM support detect

fisker:
Replace `lib/esm-probe.mjs` file by a dataURL, 

~and we don't care about the `import` result, we don't have to wait~

Cache the detect result, so we don't have to wait for a microtask everytime

@arlac77

### DIFF
--- a/lib/worker/subprocess.js
+++ b/lib/worker/subprocess.js
@@ -135,12 +135,13 @@ ipc.options.then(async options => {
 	}).filter(provider => provider !== null);
 
 	// Lazily determine support since this prints an experimental warning.
-	let supportsESM = async () => {
+	let supportsESM = () => {
 		try {
-			await import('../esm-probe.mjs');
-			supportsESM = async () => true;
+			// eslint-disable-next-line no-unused-expressions
+			import('data:,');
+			supportsESM = () => true;
 		} catch {
-			supportsESM = async () => false;
+			supportsESM = () => false;
 		}
 
 		return supportsESM();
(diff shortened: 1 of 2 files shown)


lib/worker/subprocess.js
```diff
@@ -6,6 +6,16 @@ require('./ensure-forked'); // eslint-disable-line import/no-unassigned-import
 
 const ipc = require('./ipc');
 
+// Lazily determine support since this prints an experimental warning.
```
novemberborn: This comment should be moved down to the `await supportsESM()` call below.

lib/worker/subprocess.js
```diff
@@ -6,6 +6,16 @@ require('./ensure-forked'); // eslint-disable-line import/no-unassigned-import
...
+const supportsESM = async () => {
+	try {
+		await import('data:text/javascript,');
```
novemberborn: Ha that's neat. Saves us from having a weird file in the package.

lib/worker/subprocess.js
```diff
@@ -134,23 +143,18 @@ ipc.options.then(async options => {
...
 			if (ref.endsWith(`.${extension}`)) {
-				if (await supportsESM()) { // eslint-disable-line no-await-in-loop
+				if (typeof isESMSupported !== 'boolean') {
+					// Lazily determine support since this prints an experimental warning.
+					// eslint-disable-next-line no-await-in-loop
+					isESMSupported = await supportsESM();
+				}
+
+				if (isESMSupported) {
 					return import(pathToFileURL(ref));
```
fisker: Do we need this `pathToFileURL`?

fisker: Seems need for Node.js 13 on windows https://github.com/avajs/ava/pull/2404/checks?check_run_id=457268243

novemberborn: Yep 😄

PR #2407, opened 2020-02-22
support undefined as second argument to throws and throwsAsync

### LINKED ISSUE #2401: Allow `undefined` throws expectations
For historical reasons, the `expectations` object passed to `t.throws()` and `t.throwsAsync()` was not allowed to be `undefined`.

Now that we've simplified these assertions, we should allow `expectations` to be both `undefined` and `null`. See here:

https://github.com/avajs/ava/blob/55a3649000611900ba6dfbf94faefa6372a70e79/lib/assert.js#L76

This will cause some of our tests to fail, so they'd have to be changed. See here:

https://github.com/avajs/ava/blob/55a3649000611900ba6dfbf94faefa6372a70e79/test/assert.js#L941

stavalfi:
Fixes #2401

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -247,14 +247,9 @@ export interface SnapshotAssertion {
 export interface ThrowsAssertion {
 	/**
 	 * Assert that the function throws [an error](https://www.npmjs.com/package/is-error). If so, returns the error value.
-	 */
-	<ThrownError extends Error>(fn: () => any, expectations?: null, message?: string): ThrownError;
-
-	/**
-	 * Assert that the function throws [an error](https://www.npmjs.com/package/is-error). If so, returns the error value.
 	 * The error must satisfy all expectations.
 	 */
-	<ThrownError extends Error>(fn: () => any, expectations: ThrowsExpectation, message?: string): ThrownError;
+	<ThrownError extends Error>(fn: () => any, expectations?: ThrowsExpectation | null, message?: string): ThrownError;
 
 	/** Skip this assertion. */
 	skip(fn: () => any, expectations?: any, message?: string): void;
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -73,7 +73,7 @@ function getErrorWithLongStackTrace() {
 }
 
 function validateExpectations(assertion, expectations, numArgs) { // eslint-disable-line complexity
-	if (numArgs === 1 || expectations === null) {
+	if (numArgs === 1 || expectations === null || expectations === undefined) {
 		expectations = {};
 	} else if (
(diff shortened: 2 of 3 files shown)


sindresorhus:
Don't forget to update https://github.com/avajs/ava/blob/master/index.d.ts

novemberborn:
> Don't forget to update https://github.com/avajs/ava/blob/master/index.d.ts

@sindresorhus I think these definitions are fine as-is?

https://github.com/avajs/ava/blob/324e45f2f89e6fa872423c338aba20e202c5b2f5/index.d.ts#L247:L257

stavalfi:
cant we combine them?

```typescript
	/**
	 * Assert that the function throws [an error](https://www.npmjs.com/package/is-error). If so, returns the error value.
	 */
	<ThrownError extends Error>(fn: () => any, expectations?: null, message?: string): ThrownError;

	/**
	 * Assert that the function throws [an error](https://www.npmjs.com/package/is-error). If so, returns the error value.
	 * The error must satisfy all expectations.
	 */
	<ThrownError extends Error>(fn: () => any, expectations: ThrowsExpectation, message?: string): ThrownError;
```

to:

```typescript
	/**
	 * Assert that the function throws [an error](https://www.npmjs.com/package/is-error). If so, returns the error value.
	 * The error must satisfy all expectations.
	 */
	<ThrownError extends Error>(fn: () => any, expectations?: null | ThrowsExpectation, message?: string): ThrownError;
```

sindresorhus:
Oh right. I missed the question mark when I was reviewing on my phone.

stavalfi:
> cant we combine them?
> 
> ```ts
> 	/**
> 	 * Assert that the function throws [an error](https://www.npmjs.com/package/is-error). If so, returns the error value.
> 	 */
> 	<ThrownError extends Error>(fn: () => any, expectations?: null, message?: string): ThrownError;
> 
> 	/**
> 	 * Assert that the function throws [an error](https://www.npmjs.com/package/is-error). If so, returns the error value.
> 	 * The error must satisfy all expectations.
> 	 */
> 	<ThrownError extends Error>(fn: () => any, expectations: ThrowsExpectation, message?: string): ThrownError;
> ```
> 
> to:
> 
> ```ts
> 	/**
> 	 * Assert that the function throws [an error](https://www.npmjs.com/package/is-error). If so, returns the error value.
> 	 * The error must satisfy all expectations.
> 	 */
> 	<ThrownError extends Error>(fn: () => any, expectations?: null | ThrowsExpectation, message?: string): ThrownError;
> ```

done.

PR #2413, opened 2020-02-26
Add debug --host option

### LINKED ISSUE #2406: AVA debug --break is hardcoded to bind to loopback interface.
Dockerized AVA in debug mode is difficult to reach due to hardcoded binding to loopback interface (127.0.0.1).

file: [ava/lib/worker/subprocess.js:188](https://github.com/avajs/ava/blob/324e45f2f89e6fa872423c338aba20e202c5b2f5/lib/worker/subprocess.js#L188)
`require('inspector').open(options.debug.port, '127.0.0.1', true);`

Hacking it to 0.0.0.0 fixes the problem.
It would be nice if it defaults to loopback, and read, for example, 'options.debug.ip' as an override.

There are couple workarounds but none of them are easy/convenient.
`socat -d tcp-listen:xxxx,reuseaddr,fork tcp:127.0.0.1:xxxx &`
But meh...

DYefimov:
Fixes #2406

Previously AVA in debug mode was listening only on hardcoded '127.0.0.1' (loopback) interface.
Now, it defaults to loopback, with an override command line option: "--host".

Example usage:
`node ava debug --host 0.0.0.0 --break "${relativeFile}"`

### DIFF
--- a/docs/recipes/debugging-with-chrome-devtools.md
+++ b/docs/recipes/debugging-with-chrome-devtools.md
@@ -22,10 +22,10 @@ Run with the `--break` option to ensure the DevTools hit a breakpoint right befo
 npx ava debug --break test.js
 ```
 
-You can also customize the port. It defaults to `9229`:
+By default the inspector listens on `127.0.0.1:9229`. You can customize the host and the port:
 
 ```console
-npx ava debug --port 9230 test.js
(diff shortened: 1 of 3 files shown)


novemberborn:
@DYefimov PR looks great at first glance, will try and land it this weekend.

novemberborn:
Thanks @DYefimov, this is now available as v3.5!

Issue #2110, opened 2019-04-29
Display internal Node.js stacktrace if no user stacktrace is present


szmarczak:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%2440%20Funded-%2300A156.svg" />](https://issuehunt.io/r/avajs/ava/issues/2110)

### Current behavior

![image](https://user-images.githubusercontent.com/36894700/56914132-a0bde300-6ab3-11e9-8260-7c3530dc407f.png)

## Expected behavior

```js
{ Error: self signed certificate
    at TLSSocket.onConnectSecure (_tls_wrap.js:1177:34)
    at TLSSocket.emit (events.js:193:13)
    at TLSSocket._finishInit (_tls_wrap.js:668:8) code: 'DEPTH_ZERO_SELF_SIGNED_CERT' }
```

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $40.00)

- [<img src='https://avatars3.githubusercontent.com/u/44827199?v=4' alt='issuehunt' width=24 height=24> issuehunt](https://issuehunt.io/u/issuehunt) ($40.00)

### Submitted pull Requests
- [#2420 Show node internal stack trace](https://issuehunt.io/r/avajs/ava/pull/2420)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/avajs/ava/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
---
IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

sindresorhus:
Makes sense to me.

I also notice we don't show the `.code` property. We should show that if it's not empty, no matter what.

szmarczak:
It should also display `No user stacktrace present.` IMHO

novemberborn:
It's a little jarring when there is no stack trace. But, if the Node.js trace lines are useful here, why wouldn't they be useful in other situations?

Should we perhaps color code the trace lines but not filter any?

Agreed we should show all properties of the error.

sindresorhus:
It's the difference between the error originating from your own code or from Node.js itself. If it originates from your own code, you don't really care or want the Node.js code frames, but if the error originates from Node.js, you definitely want them for context or being able to report it to the Node.js team.

Issue #2412, opened 2020-02-24
Retain internal stack lines for errors, but color differently


szmarczak:
It would be quite useful because there's sometimes missing context.

novemberborn:
Could you provide more, uhm, context?

szmarczak:
Example:

```
  PromisableRequest._destroy (dist/source/core/index.js:935:21)
  PromisableRequest._beforeError (dist/source/core/index.js:855:14)
  dist/source/core/index.js:262:22
```

Full stack trace:
```
    at PromisableRequest._destroy (/home/szm/Desktop/got/dist/source/core/index.js:935:21)
    at PromisableRequest.destroy (internal/streams/destroy.js:55:8)
    at PromisableRequest._beforeError (/home/szm/Desktop/got/dist/source/core/index.js:855:14)
    at /home/szm/Desktop/got/dist/source/core/index.js:262:22
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

I think that `at PromisableRequest.destroy (internal/streams/destroy.js:55:8)` is quite important. Without this AVA implies that `_beforeError` called `_destroy` directly.

novemberborn:
Yea I don't really know why we even do this. If anything, we should use colors and separators to highlight the most useful lines, but not remove anything. Other than AVA-introduced lines.

IIRC this happens in https://github.com/avajs/ava/blob/master/lib/beautify-stack.js.

novemberborn:
@szmarczak so you filed https://github.com/avajs/ava/issues/2110 last year, which is pretty similar 😉 

There's a PR to fix that one (https://github.com/avajs/ava/pull/2420) which is great. However I think we should still consider retaining the internal stack lines for all errors, whether or not they originated outside of user code, and use color to indicate their relevance.

So we'll keep this issue open.

novemberborn:
> However I think we should still consider retaining the internal stack lines for all errors, whether or not they originated outside of user code, and use color to indicate their relevance.

I think this was addressed in #2420 as well.

PR #2420, opened 2020-03-09
Retain (most of) Node.js internals in stack traces

bunysae:
Fixes issue #2110. The node internal stack trace is shown for error originated from Node.js. If it's originated from custom code, node internal stack traces are culled out.
The `code`-property is shown in node v12.16.1 without any modification to ava-source-code. In older versions (i had v12.10.0) before, the `code`-property wasn't a enumerable property and therefore doesn't appear.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#2110: Display internal Node.js stacktrace if no user stacktrace is present](https://issuehunt.io/repos/26820798/issues/2110)
---

</details>


novemberborn:
Thanks for your PR @bunysae. I did not open my editor this weekend so I haven't had time yet to look into this, sorry.

novemberborn:
I've spent some time trying to figure out what we're trying to do here. #2110 is an older issue, we recently came to the conclusion that it'd be good to show all lines, but use colors to indicate the less relevant ones. See https://github.com/avajs/ava/issues/2412 for that discussion.

That's probably more than we need to deal with in this PR. However as you probably found the code here is rather messy.

I suspect we can remove `clean-stack`, and rely solely on `stack-utils`. We haven't kept pace with the changes in `stack-utils` so we need to make better use of its current features.

There's another mess in `lib/serialize-error.js` which ends up relying on the beautified stack trace: https://github.com/avajs/ava/blob/b6a46bd7b38a0eb1ce52e552230d066988f0402a/lib/serialize-error.js#L58

---

Going by your test, the output for a file like this:

```js
const path = require('path')
const test = require('ava')

test('foo', t => {
  path.resolve({root: '..'})
  t.pass()
})
```

Goes from:

```console
npx ava -v

  ✖ foo Error thrown in test

  1 test failed

  foo

  test.js:5

   4: test('foo', t => {
   5:   path.resolve({root: '..'})
   6:   t.pass()

  Error thrown in test:

  TypeError (NodeError) {
    code: 'ERR_INVALID_ARG_TYPE',
    message: 'The "path" argument must be of type string. Received type object',
  }
```

To:

```console
npx ava -v

  ✖ foo Error thrown in test

  1 test failed

  foo

  internal/validators.js:112

  Error thrown in test:

  TypeError (NodeError) {
    code: 'ERR_INVALID_ARG_TYPE',
    message: 'The "path" argument must be of type string. Received type object',
  }

  validateString (internal/validators.js:112:11)
  Object.resolve (path.js:980:7)
  test.js:5:8
  processTicksAndRejections (internal/process/task_queues.js:94:5)
```

Note how we're no longer showing the test code in the error output. Arguably, the error did not originate in Node.js, but it originated in the test.

https://github.com/avajs/ava/issues/2110 describes a situation where the error occurred asynchronously, outside of a test invocation.

---

Probably the right approach here is to simplify the error formatting and the various ways in which we munge the stack, and then see how errors end up being reported. What do you think @bunysae?

bunysae:
@novemberborn 
I understand your problem with this PR. When i understand the discussion in #2412 correctly, you want still cull out ava internal stack lines and give the user and node internal stack lines different colors?
I think i can do this with the `chalk`-lib.

novemberborn:
> When i understand the discussion in #2412 correctly, you want still cull out ava internal stack lines and give the user and node internal stack lines different colors?

Yes. However this is made more difficult due to mess I described above. We should try and clear that up first.

bunysae:
I refactored this PR. Now the complete stack-trace is shown except:
- ava internals,
- ava dependencies
- the node internals `internal/process/task-queues` and `internal/modules/cjs/loader` (They add noice ava assertion errors and errors thrown due to inpromper test files).
Node internals are grey. Custom-code has the default color.

How to you want the show the code-excerpt? In the path-example above, the code excerpt would be still omitted. Should ava always show a excerpt of the custom-code, if custom-code is involved?

novemberborn:
@bunysae thanks for the update. I'll try and have a look at this on the weekend.

lib/beautify-stack.js
```diff
@@ -54,19 +59,20 @@ function extractFrames(stack) {
...
 		return '';
 	}
 
-	stack = extractFrames(stack);
-	// Workaround for https://github.com/tapjs/stack-utils/issues/14
-	// TODO: fix it in `stack-utils`
-	stack = cleanStack(stack);
-
-	return stackUtils.clean(stack)
+	return stackUtils.clean(extractFrames(stack))
```
novemberborn: Do we still need `extractFrames`? Hopefully `stack-utils` deals with this now.

lib/reporters/mini.js
```diff
@@ -322,13 +323,24 @@ class MiniReporter {
...
-				this.lineWriter.writeLine(colors.errorStack(stack));
+				this.lineWriter.writeLine(this.formatErrorStack(stack));
 			}
 		}
 	}
 
+	formatErrorStack(stack) {
+		return stack.split('\n')
+			.map(line => {
+				if (StackUtils.nodeInternals().some(internal => internal.test(line))) {
```
novemberborn: `nodeInternals()` allocates a new array every time it's called. Let's assign it to a variable instead.

lib/beautify-stack.js
novemberborn: Thinking about this more, do we even need to clean the stack here? We could do it in the reporters instead. We'd have to update `serialize-error` so it finds the first call site that is in the test file, rather than assuming it's the first call site in the trace.

test/beautify-stack.js
```diff
@@ -76,3 +77,20 @@ test('beautify stack - removes uninteresting lines', async t => {
 		t.end();
 	}
 });
+
+test('beautify stack - don`t remove node internals', async t => {
```
novemberborn:
```suggestion
test('beautify stack - don\'t remove node internals', async t => {
```

bunysae:
My idea to handle the stack-traces of different nodes.js versions and
OSes is to create different log-files for every combination.
Then we wouldn't had a problem with a failing CI-build.

novemberborn:
> My idea to handle the stack-traces of different nodes.js versions and OSes is to create different log-files for every combination. Then we wouldn't had a problem with a failing CI-build.

Yes, sounds good. It'll be a bit harder to update the files, but that should still be easier than trying to string-replace away the differences.

lib/beautify-stack.js
bunysae: @novemberborn 
Do you think the parameter `shouldBeautifyStack` in function `serialize-error` (https://github.com/avajs/ava/blob/ede4f322b4fa6263c8ae14ec04282ab06bb0afd4/lib/serialize-error.js#L136) is still necessary? It's set to false for internal-errors in the modules `lib/subprocess/worker`(https://github.com/avajs/ava/blob/ede4f322b4fa6263c8ae14ec04282ab06bb0afd4/lib/worker/subprocess.js#L161) and `lib/api` (https://github.com/avajs/ava/blob/ede4f322b4fa6263c8ae14ec04282ab06bb0afd4/lib/api.js#L239).

Without this parameter, it would be much easier to move the stack-cleaning to the reporters.

novemberborn: Yea that's what I'm saying. Let's see if we can move the stack processing to the main process (reporters).

bunysae:
I moved the stack processing to the reporters and splited the log-files for the three major node versions.

novemberborn:
@bunysae awesome! It probably won't be until the weekend that I'll have a chance to look at this though.

lib/serialize-error.js
```diff
@@ -17,13 +16,35 @@ function filter(propertyName, isRoot) {
...
+	let result = null;
+	stack.split('\n').forEach(line => {
+		try {
+			const callSite = stackUtils.parseLine(line.replace('file:///', '/'));
+			// Assume the CWD is the project directory (see function `buildSource()`
+			// and assume `testFile` is a absolute path.
+			// The `callSite.file` property is platform independent (with '/' as directory separator)
+			// and `testFile` is platform dependent (on win32 with '\' as directory separator).
+			if (callSite.file === path.relative(process.cwd(), testFile).split(path.sep).join('/')) {
+				result = callSite;
```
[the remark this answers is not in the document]
bunysae: No that wasn't intentional. I didn't realize that.

lib/serialize-error.js
```diff
@@ -17,13 +16,35 @@ function filter(propertyName, isRoot) {
...
+	if (!stack || !testFile) {
 		return null;
 	}
 
-	const firstStackLine = stack.split('\n')[0];
-	return stackUtils.parseLine(firstStackLine);
+	let result = null;
+	stack.split('\n').forEach(line => {
+		try {
+			const callSite = stackUtils.parseLine(line.replace('file:///', '/'));
```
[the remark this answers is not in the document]
bunysae: `file:///` is replaced to pass the test (https://github.com/bunysae/ava/blob/7c5ebac7adb4183393d02ee6e6f5b237e5f6aa1e/test-tap/serialize-error.js#L128).

bunysae:
@novemberborn Except the open-point with the esm file-url's it looks goods. Didn't find more todos in my review.

novemberborn:
> Except the open-point with the esm file-url's it looks goods. Didn't find more todos in my review.

I couldn't reproduce that problem with the latest `esm` so I think we're good!

novemberborn:
I'm just asking some folks for feedback outside of GitHub. Stay tuned 😄

PR #2422, opened 2020-03-12
Disallow `null` expectations for `throws` and `throwsAsync` assertions

### LINKED ISSUE #2410: Disallow `null` expectations for `throws` and `throwsAsync` assertions
As a follow-up to https://github.com/avajs/ava/issues/2401, for the next major AVA release we'll remove support for `null` expectations in `throws` and `throwsAsync` assertions. If you want to specify a message without any other expectations you should set it to `undefined` instead.

This can partially land behind an [experimental flag](https://github.com/avajs/ava/blob/d0e21612390d50f084f7062b051164624ef391ff/lib/load-config.js#L10), to be unflagged when the next major is (pre-)released. The the TypeScript definition can only be updated at that point.

TokenStyle:
Fixes #2410

Second argument called `expectation` (in the documentation) of the `throws` and `throwsAsync` functions, in this change cannot be assigned as `null` (but can be assigned as `undefined`).
Also in the error throw message, I changed the text of `message`, removed `null` as one of the options for the `expectation` argument.

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -73,7 +73,7 @@ function getErrorWithLongStackTrace() {
 }
 
 function validateExpectations(assertion, expectations, numArgs) { // eslint-disable-line complexity
-	if (numArgs === 1 || expectations === null || expectations === undefined) {
+	if (numArgs === 1 || expectations === undefined) {
 		expectations = {};
 	} else if (
 		typeof expectations === 'function' ||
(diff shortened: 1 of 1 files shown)


novemberborn:
Thanks for the PR @TokenStyle.

Per #2410, this needs to be behind a flag:

> This can partially land behind an [experimental flag](https://github.com/avajs/ava/blob/d0e21612390d50f084f7062b051164624ef391ff/lib/load-config.js#L10), to be unflagged when the next major is (pre-)released. The the TypeScript definition can only be updated at that point.

novemberborn:

Issue #840, opened 2016-05-17
Is there a way to get test status on afterEach() method ?


subash-canapathy:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%2440%20Funded-%2300A156.svg" />](https://issuehunt.io/r/avajs/ava/issues/840)

mocha has a way of getting the test status on the after\* hooks to do customized teardowns. `currentTest.status` 
this is required for UI tests running on saucelabs and that feature in mocha was born out of the same need: https://github.com/mochajs/mocha/issues/797

It would be super useful to have the test status exposed as metadata OR contextual information

Proposal  A:
`t.context.result` - to contain the result object (the one which goes to the reporter)

Proposal B:
`t.getTestStatus()` - a method available only on the afterEach hook to get status of the test.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $40.00)

- [<img src='https://avatars3.githubusercontent.com/u/44827199?v=4' alt='issuehunt' width=24 height=24> issuehunt](https://issuehunt.io/u/issuehunt) ($40.00)

### Submitted pull Requests
- [#2427 Add hook `afterEachFailedTest`](https://issuehunt.io/r/avajs/ava/pull/2427)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/avajs/ava/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
---
IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

jamestalmage:
If we do this, it has to be `B`. You are allowed to completely replace the context:

```js
t.context = somThing()
```

> this is required for UI tests running on saucelabs and that feature in mocha was born out of the same need: mochajs/mocha#797

Three stars for describing a real world need for the feature! ⭐ ⭐  ⭐

novemberborn:
@subash-canapathy could you provide an example (in Mocha) of how you use this information with your Saucelabs tests? The issue you linked to doesn't provide that kind of context.

subash-canapathy:
in mocha, when you run UI tests, you typically have the afterEach block handle the session closeouts with sauceLabs depending on the status of the test. For eg:
```JS
afterEach(function(){
  // propagate test status/result to sauceLabs
  this.driver.sauceJobStatus(this.currentTest.state)  
[...]

PR #2427, opened 2020-03-22
Add hook `afterEachFailedTest`

bunysae:
I liked the idea supposed in issue #840 [introducing a new hook](https://github.com/avajs/ava/issues/840#issuecomment-524370743). So in this PR i added the hook `afterEachFailedTest`

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#840: Is there a way to get test status on afterEach() method ?](https://issuehunt.io/repos/26820798/issues/840)
---

</details>


novemberborn:
Great PR @bunysae!

I've re-read #840 and I don't think we necessarily settled on the right approach. Rather than adding a new hook, what if we added `t.passed` to the `t` object provided to `test.afterEach.always()` hooks?

What do you think?

bunysae:
I replaced this PR with #2429.

Issue #1861, opened 2018-07-13
Allow custom serialisers


OmgImAlexis:
## Description
Currently if a HTML element is passed to `t.snapshot` it copies the code as is to the `.md` file.

I was hoping to pass a function to the serialiser so I could pretty print the content before it's saved to the `.md` file.

## Relevant Links
Example: [app-header.spec.js.md](https://github.com/pymedusa/Medusa/blob/bb10a78003e9c5a2095da3b6ed0dacdea5fe1c16/themes-default/slim/test/specs/snapshots/app-header.spec.js.md
)

Example: [html serialiser for Jest](https://github.com/rayrutjes/jest-serializer-html) 

---

> It’s designed to do that, but it’s not exposed in AVA. Happy to work out details if you open an issue.

Ref: https://twitter.com/novemberborn/status/1017718535274065920

novemberborn:
The library underlying `t.snapshot()` and `t.deepEqual()` is [concordance](https://github.com/concordancejs/concordance/). It was designed to be extensible (see  [`@concordance/react`](https://github.com/concordancejs/react/), but we've never exposed an interface to do so within AVA.

Writing such extensions is a fair bit more difficult than for Jest's snapshot mechanism, but I guess it won't get easier until we start encouraging people to write their own 😄 

Question though, how do we want to configure these extensions? Programmatically?

```js
import {registerConcordancePlugin} from 'ava/extend'
```

Or in the configuration file, with a factory method pattern?

```js
export default {
  plugins: [
    ({registerConcordance}) => registerConcordance(…)
  ]
}
```

We'll also need some metadata regarding the plugin that's been registered, so that if AVA encounters a snapshot without the corresponding plugin being available, it can give a useful error message.

It's been a while since I've looked at the underlying code but these are the main questions I think.

novemberborn:
I'm tentatively adding this to the 4.0 milestone, but I reckon it won't make it. We're removing our built-in React support in AVA 4 so we'd want something like this to continue supporting React projects.

See a stalled attempt in https://github.com/avajs/ava/pull/2428.

aleclarson:
I see a fix got merged, but this issue wasn't closed. Is it documented anywhere?

PR #2428, opened 2020-03-22
Support Concordance plugins

novemberborn:
Fixes https://github.com/avajs/ava/issues/1861.

I'd like to remove our built-in support of https://github.com/concordancejs/react/, but we'd have to make Concordance configurable.

I think a top-level `plugins` configuration will work. You'd provide package names. We'd only support CJS packages for now. We'll publish our own `@ava/react` (which wraps https://github.com/concordancejs/react/, because we want to configure the colors). If you configure `plugins`, you won't get the default behavior.

To do:

* [ ] Resolve plugins
* [ ] Make this experimental
* [ ] Publish an `@ava/react`
* [ ] Documentation!


sindresorhus:
Sounds like a good idea 👍

novemberborn:
Just to note, AVA 4 is removing built-in support for React. Ideally we can land this before AVA 4 ships, but no promises.

PR #2429, opened 2020-03-23
Add t.passed to execution contexts

bunysae:
Instead of introducing the a new hook (PR #2427) the property `t.passed` is passed to the hook `afterEach.always`.

Fixes https://github.com/avajs/ava/issues/840.

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -266,7 +266,7 @@ class Runner extends Emittery {
 		return result;
 	}
 
-	async runHooks(tasks, contextRef, titleSuffix) {
+	async runHooks(tasks, contextRef, titleSuffix, testPassed) {
 		const hooks = tasks.map(task => new Runnable({
 			contextRef,
 			experiments: this.experiments,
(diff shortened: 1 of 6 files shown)


novemberborn:
Thanks @bunysae I'll try and have a look this weekend.

bunysae:
@novemberborn 
Done. Please review again.

lib/runner.js
```diff
@@ -347,7 +348,7 @@ class Runner extends Emittery {
 			}
 		}
 
-		const alwaysOk = await this.runHooks(this.tasks.afterEachAlways, contextRef, hookSuffix);
+		const alwaysOk = await this.runHooks(this.tasks.afterEachAlways, contextRef, hookSuffix, hooksAndTestOk);
```
novemberborn, line 351: This means that, if a `afterEach` hook fails, the `afterEach.always` hook sees `t.passed === false`, even if the test passed.

I think the semantics are clearer if `t.passed` is `true` only when the test passed. If the test didn't run, because a `beforeEach` hook failed, then `t.passed === false`, but if it succeeded and then an `afterEach` hook failed, it should be `t.passed === true`.

What do you think?

bunysae: This is a entitled issue. I fixed it quite now.

novemberborn:
Thanks @bunysae!

PR #2435, opened 2020-03-29
Experimental test interface

novemberborn:
This kicks off some experimental changes to AVA's test interface. That is, the `const test = require('ava')` stuff. Eventually this will be come AVA 4, but initially it'll be available through importing `ava/experimental`. You'll also have to enable the `experimentalTestInterfaces` experiment.

Summary of (planned) changes:

* AVA is now just a CJS module with a single export: the `test` function. If you're transpiling ESM, you'll no longer be able to `import { serial } from 'ava'`. This never worked anyway with proper ESM.
* Test implementations can no longer have `title` functions.
* You'll no longer be able to pass an array of macros (or implementations), when declaring tests or with `t.try()`.
* Speaking of macros, if you want them to have a `title` function you'll need to use `test.macro({ exec (t) { … }, title () { return 'title' })`.
* If you're writing tests using TypeScript, the `Macro` type has changed.
* But, on the plus side, `test.macro()` automatically uses the `t.context` type from `test`.

I tend to write service integration tests using AVA, which involve stubbing several APIs, managing database access and launching child processes. Crafting `t.context` for these tests is rather complicated.

I'm tinkering with a "forkable test interface". You'd start off calling `test.make()`. This gives you an entirely new test chain, say `test2`. You can define hooks and whatnot, and then call `test2.fork()` to create yet another test chain, which will use the hooks defined on `test2` before forking.

You'll be able to pass some options to `make()` and `fork()` to construct `t.context`, along the lines of what you'd do using `test.before()` and `test.beforeEach()`. But, for TypeScript users, the context type will be inferable.

We could also use this to create a test chain that always executes serially, or limits concurrency, or runs in isolation from other test chains in the same test file, or [executes conditionally](https://github.com/avajs/ava/issues/2602).

More to come!

### DIFF
--- a/experimental.d.ts
+++ b/experimental.d.ts
@@ -343,8 +343,24 @@ export interface TodoDeclaration {
 	(title: string): void;
 }
 
+export interface ForkableSerialInterface<Context = unknown> extends SerialInterface<Context> {
+	/** Create a new serial() function with its own hooks. */
+	fork(): ForkableSerialInterface<Context>;
+}
+
+export interface ForkableTestInterface<Context = unknown> extends TestInterface<Context> {
+	/** Create a new test() function with its own hooks. */
+	fork(): ForkableTestInterface<Context>;
+
+	/** Declare tests and hooks that are run serially. */
+	serial: ForkableSerialInterface<Context>;
+}
+
 /** Call to declare a test, or chain to declare hooks or test modifiers */
-declare const test: TestInterface;
+declare const test: TestInterface & {
+	/** Create a new test() function with its own hooks. */
+	make(): ForkableTestInterface;
+};
 
 /** Call to declare a test, or chain to declare hooks or test modifiers */
 export default test;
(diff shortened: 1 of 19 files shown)


papb:
> But, for TypeScript users, the context type will be inferable.

Hi @novemberborn, currently to get accurate type checking for `t.context`, I do the following:

```ts
import test, { TestInterface } from 'ava';

function wrapTest(test: TestInterface): asserts test is TestInterface<MyContextType> {
	test.beforeEach(t => {
		t.context = getMyContext();
	});
}

wrapTest(test);

test('foo', /* ... */);
```

And it works very well. Just in case it can help you have some ideas :)

novemberborn:
@papb that's a nice trick! Would you like to add it to our TypeScript recipe?

papb:
Hi @novemberborn, sorry for taking long to reply, I would but I don't have the time, if you like it please feel free to add it, no need to credit or anything :)

novemberborn:
I've added this to the AVA 4 milestone… I don't expect to land everything but would like to get at least the breaking changes in.

novemberborn:
Other than the "forkable" stuff, this is now in https://github.com/avajs/ava/pull/2785 instead.

PR #2436, opened 2020-03-29
Support configuration via environment variables

AdrieanKhisbe:
**Use yargs '.env()' mecanism to enable configuration via environment variables :gear:**

- Just configure `yargs` with a single option, `.env('AVA')`
- Added tests to ensure configuration is working. Made by cloning tests using the command line flags.
- A work around was needed to support both `--node-arguments` and `AVA_NODE_ARGUMENTS`

We can eventually consider hiding the feature behind some feature toggle, like using `AVA=true` or similar.

Also if this is to land, I'll be more than happy if possiblee to backport the feature on the `2.x` branch (as I still develop many libs still having node 8 support).

I'll wait for first feedbacks before to add the adequate documentation, of course 🙃

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -106,6 +106,7 @@ exports.run = async () => { // eslint-disable-line complexity
 			'strip-aliased': true,
 			'unknown-options-as-args': false
 		})
+		.env('AVA')
 		.usage('$0 [<pattern>...]')
 		.usage('$0 debug [<pattern>...]')
 		.usage('$0 reset-cache')
@@ -335,7 +336,9 @@ exports.run = async () => { // eslint-disable-line complexity
 
 	let nodeArguments;
 	try {
-		nodeArguments = normalizeNodeArguments(conf.nodeArguments, argv['node-arguments']);
+		const nodeArgs = [argv['node-arguments'], argv.nodeArguments].filter(Boolean).join(' ');
+		// Using yargs env AVA_NODE_ARGS only sets nodeArguments, while cli args set node-arguments
+		nodeArguments = normalizeNodeArguments(conf.nodeArguments, nodeArgs);
 	} catch (error) {
 		exit(error.message);
 	}
--- a/test-tap/integration/assorted.js
+++ b/test-tap/integration/assorted.js
@@ -41,9 +41,25 @@ test('--match works', t => {
 	});
 });
 
+test('--match works from env variable', t => {
+	execCli(['matcher-skip.js'], {env: {AVA_MATCH: 'tests are fun'}}, err => {
+		t.ifError(err);
+		t.end();
+	});
+});
+
 for (const tapFlag of ['--tap', '-t']) {
 	test(`${tapFlag} should produce TAP output`, t => {
 		execCli([tapFlag, 'test.js'], {dirname: 'fixture/watcher'}, err => {
+			t.ok(!err);
+			t.end();
+		});
+	});
+}
+
+for (const tapEnv of ['AVA_TAP', 'AVA_T']) {
+	test(`${tapEnv}=true should produce TAP output`, t => {
+		execCli(['test.js'], {dirname: 'fixture/watcher', env: {[tapEnv]: 'true'}}, err => {
 			t.ok(!err);
 			t.end();
 		});
--- a/test-tap/integration/concurrency.js
+++ b/test-tap/integration/concurrency.js
@@ -40,3 +40,10 @@ test('works when --concurrency is provided with a value', t => {
 		t.end();
 	});
 });
+
+test('works when AVA_CONCURRENCY is provided with a value', t => {
+	execCli(['test.js'], {dirname: 'fixture/concurrency', env: {AVA_CONCURRENCY: '1'}}, err => {
+		t.ifError(err);
+		t.end();
+	});
+});
--- a/test-tap/integration/node-arguments.js
+++ b/test-tap/integration/node-arguments.js
@@ -8,6 +8,20 @@ test('passes node arguments to workers', t => {
 		(err, stdout, stderr) => t.ifError(err, null, {stdout, stderr}));
 });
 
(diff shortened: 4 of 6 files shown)


sindresorhus:
What's the use-case for this? Why are the CLI flags / config not good enough?

sindresorhus:
> Also if this is to land, I'll be more than happy if possiblee to backport the feature on the 2.x branch (as I still develop many libs still having node 8 support).

That will unfortunately not happen. We only support the latest major. We just don't have the resources to support multiple release lines.

Even the Node.js core team doesn't support Node.js 8 anymore. It's time to move on.

novemberborn:
@AdrieanKhisbe I appreciate the work on the PR. While I am interested in your use case, I think it's unlikely we'll support configuration through environment variables.

You could however use an `ava.config.js` / `ava.config.cjs` file and resolve environment variables yourself.

AdrieanKhisbe:
Hello @sindresorhus & @novemberborn.

Be reassured, I did not fork and dive into ava codebase without having what is a real use case to me.
But indeed, I did not explained at all in the PR details what that use case is! My bad :man_facepalming:

**So here is the why:**

------

The primary need it to **configure ava when you don't invoke it directly**. Often, ava command are nested into some _npm scripts_, and npm _sub_ scripts.

In that case, you can only configure `ava` if it comes lasts, and if you provide the proper amount of `--` so that npm transfer the commands to `ava`.

Here is an example (from [update-node](https://github.com/CoorpAcademy/update-node/blob/master/package.json#L30-L34))

```json
{
    "scripts": {
        "test": "npm run lint && nyc npm run test:all",
        "test:all": "npm run test:unit && npm run test:integration",
        "test:integration": "scripts/integration-test.sh",
        "test:unit": "ava",
        "lint": "eslint src/ *.json --ext .js,.json"
    }
}
```
If you want to configure ava, you can only do it from the `test:unit` script with `npm run test:unit -- --verbose`.
The is no way to do it at `test:all` level, or even better at top level `test`

With this PR you could configure it using `AVA_VERBOSE=true npm test`.

Some alternative that would work for me would be to provide extra cli args via an env variable, `AVA_FLAGS="--verbose --watch npm test"`

----- 
I hope that now yon can have an idea of what is the idea behind this PR :)

_PS: I can update original message adding below a summary of the use case, (which should have been there in the first place)_

novemberborn:
@AdrieanKhisbe thanks for the update. I think your use case is best served by using an `ava.config.cjs` file and having that read your environment variables.

AdrieanKhisbe:
@novemberborn, Unfortunately not really. :confused:

The goal of this proposal is more to address *punctual configurations tweaks* needs like `verbose`, `watch`, `match` (the one you control with cli flags), than the *long term/standard configuration*.
Config like files pattern, concurrency,... definitely belongs in a versioned config file like `ava.config.js` or in package.json under `ava` key.

The possibility introduced by this PR _(prepending some environment variables to a (npm) command so that `ava` can interpret them)_ is just some ersatz/substitute of cli arguments, in a situation where you can't inject that flags to the one that invokes `ava`.  (cf [example in my previous comment](https://github.com/avajs/ava/pull/2436#issuecomment-606127150))

Not great for sure, :neutral_face:
but for *locals* and *ephemeral* config switches, it's likely to be faster and easier than having to tweak repeatedly with the versioned configuration file(s).

novemberborn:
My suggestion is to access `process.env` in your config file and set the appropriate flag based on the values.

AdrieanKhisbe:
_Sorry I had not answered your suggestion, having miss that the main point of that proposed solution was to put env reading logic in the config file._

------
That works sure, but that requires to write custom logic in javascript config file, repeatedly for every project.

If `ava` handles it (_just by telling `yargs` to do the work_), we would ensure consistency, and remove the need to force a javascript config file over the package.json key `ava`.

novemberborn:
> just by telling `yargs` to do the work

I'm apprehensive about adding another configuration surface which needs to be understood, maintained and documented. Sure it's easy to achieve with our current `yargs` usage, but that's not the point.

I think you've shared examples for `--verbose`, `--watch` and `--match`. Are there other flags you think you may want to customize?

What does the `integration-test.sh` script do?

AdrieanKhisbe:
It do understand the apprehension. 
Even if the options will be scoped to `AVA_` like `AVA_SOME_OPTION`, there is indeed some risks, and even some light possibility it could break some personal configurations when introduced.
Besides that also exposes to behavior changes in yargs env handling. And for the moment there is no control over env parsing other than the prefix _(something like whitelisting args options that are specifiable by env, or specifying env var names)_

-----

About the flag of interest, I would say `--fail-fast` and `--serial`. :flags:

AdrieanKhisbe:
About the `integration-test.sh` in the example shared: it's just a bash script that test out the cli bin would launch correctly.

_(it's a basic safety net that `link` the update-node cli it, `install`'s it in some sandbox folder and calling a set of command on the cli, and on git. So far now, no assertions, just checking it don't crash. We will probably move it a node test suite using ava and child_process to add assertions on output/side effects)_

novemberborn:
Big picture, I'd like for AVA to support all kinds of setup and invocations scenarios through `npx ava`. Already you can run tests from specific folders (`npx ava **/test/integration/**/*`), load different configuration files and so on. I don't think environment variables are needed for that.

I'm not sure where that leaves you though 😞

AdrieanKhisbe:
@novemberborn I'm not sure if you respond about the [second message](https://github.com/avajs/ava/pull/2436#issuecomment-607382737)

If so, that message was kind of off topic, as you asked me what this bash script was for.
(That's why I posted it as a second message)

novemberborn:
> I'm not sure if you respond about the [second message](https://github.com/avajs/ava/pull/2436#issuecomment-607382737)

No, just generally.

AdrieanKhisbe:
Okay.

PR #2444, opened 2020-04-07
Expose snapshot directory through test.meta

### LINKED ISSUE #1984: Expose config object
[Issuehunt badges](https://issuehunt.io/r/avajs/ava/issues/1984)

It would be useful for ava to expose the config object to the tests. For example I need to save some extra files with snapshots, it would be nice if I could pull the `snapshotDir` from the config.

> There is a $40.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/avajs/ava/issues/1984).

ulken:
Closes #1984


lib/runner.js
```diff
@@ -54,7 +54,12 @@ class Runner extends Emittery {
 		let hasStarted = false;
 		let scheduledStart = false;
 		const meta = Object.freeze({
-			file: options.file
+			file: this.file,
+			snapshotDirectory: snapshotManager.determineSnapshotDir({
```
novemberborn, line 58:
```suggestion
			get snapshotDirectory: () => snapshotManager.determineSnapshotDir({
```

ulken:
> Let's make `snapshotDirectory` a getter. We don't need to resolve it in test files that don't use snapshots.

Ah, that's what you meant with `get`ter – smart! I was looking for a way to only compute the path if snapshots was actually used. First turned to `options.snapshotDir`, but then learned that was just used as a "prefix". A lazy getter would solve it just perfectly.

> Regarding the CI failures, I think you need to add `mem` to this list, since it messes with some expected stack traces:

> https://github.com/avajs/ava/blob/447d37153b5af9139c45cf5b0fd97ad92bea0c14/lib/beautify-stack.js#L10

Thanks, wasn't aware. You might be right about that. I'll try it out. I was being lazy/too tired to run the whole test suite on my machine. Shouldn't have created the final PR (draft could have worked). Sorry about that.
 
> There's some other failures due to perhaps some tests that are a little too low level, and are now breaking. I haven't looked into it much yet.

Yeah, I think I might be using `mem` wrongly. I'm having a hard time debugging `tap` tests, though. `console.log` won't output anything (swallowed?). Can you run `tap` in debug mode with breakpoints in VS Code, do you know?

ulken:
Also, is there no feasible way to only run _one_ test with `tap`? Like AVA's `t.only()`? I looked at the docs, but it was so tedious...

lib/runner.js
ulken: FYI: Apparently you can't use lambdas with the `get` syntax. Must be a regular function with curly brackets, function body and `return` etc..

novemberborn:
> Also, is there no feasible way to only run _one_ test with `tap`? Like AVA's `t.only()`? I looked at the docs, but it was so tedious...

`npx tap test.js --grep=title` or something like that. Or you can add `{only: true}` as the final argument, I think.

lib/runner.js
```diff
@@ -54,7 +54,14 @@ class Runner extends Emittery {
 		let hasStarted = false;
 		let scheduledStart = false;
 		const meta = Object.freeze({
-			file: options.file
+			file: this.file,
+			get snapshotDirectory() {
+				return snapshotManager.determineSnapshotDir({
+					file: this.file,
+					fixedLocation: this.snapshotDir,
+					projectDir: this.projectDir
```
novemberborn, line 62: `this` is probably `meta`, not the runner.

lib/runner.js
novemberborn: > Apparently you can't use lambdas with the `get` syntax. Must be a regular function with curly brackets, function body and `return` etc..

Why can't we have nice things.

novemberborn:
> I'm having a hard time debugging `tap` tests, though. `console.log` won't output anything (swallowed?). Can you run `tap` in debug mode with breakpoints in VS Code, do you know?

Try `console.error()`. I've never actually tried to debug `tap` tests.

lib/runner.js
ulken: Ah, yes, you’re probably right. The lambda took care of that :p

ulken:
> Try `console.error()`. I've never actually tried to debug `tap` tests.

Will do, thanks! So your theory is that `console.error()` is not swallowed?

ulken:
> `npx tap test.js --grep=title` or something like that. Or you can add `{only: true}` as the final argument, I think.

oO, looks promising. I’ll give it a go. Thanks!

ulken:
> `npx tap test.js --grep=title` or something like that. Or you can add `{only: true}` as the final argument, I think.

Follow-up: `--grep` worked fine. The optional option object (`{only: true}`) is passed as the second parameter, not the third/last (the callback comes last). Also, `tap --only` is required to only run "only" tests. Good to know, thanks!

ulken:
> Try `console.error()`. I've never actually tried to debug `tap` tests.

No go ;( Fumbling in the dark...

So I have a test fixture which uses AVA syntax. The `tap` test for `api.js` runs that file. Any logs, `throw`s etc. inside the fixture file don't show up in the terminal. Any ideas?

lib/runner.js
ulken: Correct. This was the issue.

ulken:
Looks like I don't have to mess with `beautify-stack`/`avaDependencies` after all. At least all tests pass locally. Still have to await CI...

lib/snapshot-manager.js
```diff
@@ -405,30 +424,13 @@ function determineSnapshotDir({file, fixedLocation, projectDir}) {
...
-	return file;
-}
+exports.determineSnapshotDir = determineSnapshotDir;
 
 function load({file, fixedLocation, projectDir, recordNewSnapshots, updating}) {
-	const sourceFile = resolveSourceFile(file);
-	const dir = determineSnapshotDir({file: sourceFile, fixedLocation, projectDir});
-	const relFile = path.relative(projectDir, sourceFile);
+	const dir = determineSnapshotDir({file, fixedLocation, projectDir});
+	const relFile = path.relative(projectDir, resolveSourceFile(file));
```
ulken, line 433: Note: In order to not have to export both `determineSnapshotDir()` and `resolveSourceFile()` I changed the signature of `determineSnapshotDir()` to take the "raw" file instead of the source file (it resolves the source file on its own).

Caveat: `resolveSourceFile()` is called twice for `load()` (once more to determine the relative path). But since it's memoized, I think it should be fine, no?

lib/runner.js
```diff
@@ -54,7 +54,11 @@ class Runner extends Emittery {
 		let hasStarted = false;
 		let scheduledStart = false;
 		const meta = Object.freeze({
-			file: options.file
+			file: this.file,
```
ulken, line 57: Doesn't really matter, I guess, but should I revert this to use `options` again? I switched to `this` since I think that's what's being used elsewhere, but now it looks kinda inconsistent when `snapshotDirectory()` is using `options`. Opinion?

test-tap/fixture/meta.js
```diff
@@ -1,9 +1,14 @@
 const {default: test, meta} = require('../..');
 
+test('meta is test.meta', t => {
+	t.is(meta, test.meta);
+});
```
ulken, line 5: Is this test necessary? I added it to not have to add two tests for every property as was done for `file` (checking both `meta.property` and `test.meta.property`).

test-tap/fixture/meta.js
```diff
@@ -1,9 +1,14 @@
...
+
 test('meta.file', t => {
 	t.is(meta.file, __filename);
 });
 
-test('test.meta.file', t => {
-	t.is(test.meta.file, __filename);
+test('meta.snapshotDirectory', t => {
+	t.regex(meta.snapshotDirectory, /.*snapshot-fixture.*/);
```
ulken, line 12: Regex specific enough?

docs/01-writing-tests.md
```diff
@@ -267,14 +267,18 @@ test('context is unicorn', t => {
 });
 ```
 
-## Retrieving test meta data
+## Retrieving test metadata
 
-Helper files can determine the filename of the test being run by reading `test.meta.file`.  This eliminates the need to pass `__filename` from the test to helpers.
+Access metadata of the test being run by reading `test.meta`.
```
ulken, line 272: I try to keep this short. Found the previous text a little redundant.

ulken:
@novemberborn Okidoki, I think we're good to go here. Added some self-review comments.

docs/01-writing-tests.md
novemberborn: Yea… helper files are less emphasized within AVA core anyhow.

lib/runner.js
novemberborn: Yea I'll change it back.

lib/snapshot-manager.js
novemberborn: I think that's fine.

test-tap/fixture/meta.js
novemberborn: I guess it won't hurt 😄 

I'm looking to remove this extra export in #2435.

test-tap/fixture/meta.js
novemberborn: Yea it's fine. Though you don't need the `.*`.

ulken:
LGTM!

Issue #2449, opened 2020-04-09
TypeScript types: add `asserts` and return `never` from relevant assertions


bitjson:
As of TypeScript 3.7, it's now possible to [type assertions using the new `asserts` syntax](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions). It would be great if AVA's assertions used this syntax where possible. 

For example:

```ts
  const result = { success: true, value: 2 } as
    | { success: true; value: number }
    | { success: false; error: string };

  t.true(result.success);

  const sum = result.value + 1;
```

This produces the error:
```
Property 'value' does not exist on type '{ success: true; value: number; } | { success: false; error: string; }'.
  Property 'value' does not exist on type '{ success: false; error: string; }'.ts(2339)
```

If the `t.true` method had the following signature:
```ts
(actual: any, message?: string): asserts actual is true;
``` 
The example would not produce an error. With the current types, I have to do something like:
```ts
  const result = { success: true, value: 2 } as
    | { success: true; value: number }
    | { success: false; error: string };

  if(!result.success){
    t.fail();
    return; // required to avoid type error
  }

  const sum = result.value + 1;
```

Note also the extra required `return`  – this can also be solved with a [`never`-returning function](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#better-support-for-never-returning-functions), as is now the case with `process.exit()`.

bitjson:
Ah, I missed these related issues: https://github.com/avajs/ava/issues/2411, https://github.com/avajs/ava/issues/2330

Both issues are focused on adding the `asserts` types, which is made more complicated by this error:

```ts
test('test title', (t) => {
  const result = { success: true, value: 2 } as
    | { success: true; value: number }
    | { success: false; error: string };

  t.true(result.success); // Assertions require every name in the call target to be declared with an explicit type annotation.ts(2775)
}
```

This can be fixed by declaring the type of `t`:

```ts
test('test title', (t: ExecutionContext) => { 
  const result = { success: true, value: 2 } as
    | { success: true; value: number }
    | { success: false; error: string };

  t.true(result.success);

  const sum = result.value + 1;

}
```

As I understand it, this is a fundamental limitation of the TypeScript compiler, and assertions with inferred types are not likely to be implemented any time soon. 

Unfortunately, most AVA users probably would have trouble figure out how the resolve the error (adding the `: ExecutionContext` when using `t.true`, `t.false`, or `t.fail`). 

[...]

PR #2450, opened 2020-04-09
Adjust type of t.fail to return never

bitjson:
Implements first suggestion from https://github.com/avajs/ava/issues/2449

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -127,7 +127,7 @@ export interface DeepEqualAssertion {
 
 export interface FailAssertion {
 	/** Fail the test. */
-	(message?: string): void;
+	(message?: string): never;
 
 	/** Skip this assertion. */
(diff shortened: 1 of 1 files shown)


novemberborn:
Thanks for the PR!

Based on your example in #2449 I believe this could be a breaking change? I'd really like to avoid accidentally causing compilation errors when somebody updates AVA.

I'm working towards an updated test interface in https://github.com/avajs/ava/pull/2435. We can land breaking changes there, and cut over when we're ready to ship AVA 4.

Whilst we're discussing #2449, and since #2435 hasn't landed yet, I think it's best to close this PR.

bitjson:
Ah, and this type is actually incorrect unless `failFast` is enabled: https://github.com/avajs/ava/issues/2449#issuecomment-612491973

PR #2459, opened 2020-04-17
Add `t.teardown()`

### LINKED ISSUE #1873: Setup Functions (t.teardown)
[Issuehunt badges](https://issuehunt.io/r/avajs/ava/issues/1873)

This is going to be a bit of a more testing-philosophy proposal. I'm so sorry for writing such a long issue, but it's mostly code examples of the same test suite so hopefully it won't take very long to get through.

# TL;DR

It's very hard to write setup and teardown (beforeEach/afterEach) functions _"properly"_ in pretty much every test framework that exists today. To solve this, what if AVA had "Setup Functions":

```js
let createComponent = (t, opts) => {
  let component = new Component(opts);
  t.teardown(() => component.destroy());
  return component;
};

test('component when foo does this thing', t => {
  let component = createComponent(t, { option: true });
  // ...
});
```

# The Problem: Setting stuff up for your tests

### BDD-style

For a while now, the JavaScript community has been writing "BDD-style" tests like:

```js
describe('Component', () => {
  let component;
  
  beforeEach(() => { 
    component = new Component();
  });

  describe('when foo', () => {
    let foo;
    
    beforeEach(() => { 
      foo = new Foo();
    });

    it('should do this thing', () => {
      expect(component.doThisThing(foo)).toBe(true);
    });
  });

  describe('when bar', () => {
    let bar;
    
    beforeEach(() => { 
      bar = new Bar();
    });

    it('should do this thing', () => {
      expect(component.doThisThing(bar)).toBe(true);
    });
    
    describe('and baz', () => {
      let baz;

      beforeEach(() => { 
        baz = new Baz();
      });
      
      it('should do this thing', () => {
        expect(component.doThisThing(bar, baz)).toBe(true);
      });
    });
  });
});
```

This has some notable problems:

- Lots of refactoring when:
  - You need to add tests that use value that are setup in other `describe()`'s (i.e. a test that uses both `Foo` and `Baz` would need to hoist setup code or duplicate it)
  - You need to configure objects setup in `beforeEach()` differently in individual tests (i.e. passing a different option to `new Bar()` deep within bar + baz tests) 

### AVA-style

In contrast, this is one way you could write that same test suite in AVA:

```js
const test = require('ava');

test.beforeEach(t => { 
  t.context.component = new Component();
  t.context.foo = new Foo();
  t.context.bar = new Bar();
});

test('component when foo does this thing', t => {
  t.true(t.context.component.doThisThing(foo));
});

test('component when bar does this thing', t => {
  t.true(t.context.component.doThisThing(t.context.bar));
});

test('component when bar and baz does this thing', t => {
  t.true(t.context.component.doThisThing(t.context.bar, t.context.baz));
});
```

There are still problems here though:

- You are setting up more than you need to for every test
- You can't configure how things get setup at all.

### AVA+setup functions-style

So instead, you'd probably write it like this:

```js
let createComponent = () => new Component();
let createFoo = () => new Foo();
let createBar = () => new Bar();

test('component when foo does this thing', t => {
  let component = createComponent();
  let foo = createFoo();
  t.true(component.doThisThing(foo));
});

test('component when bar does this thing', t => {
  let component = createComponent();
  let bar = createBar();
  t.true(component.doThisThing(bar));
});

test('component when bar and baz does this thing', t => {
  let component = createComponent();
  let bar = createBar();
  let baz = createBar();
  t.true(component.doThisThing(bar, baz));
});
```

This works great because:

- Your tests and setup functions are extremely portable
- Things are only setup when you need them
- Your tests write out exactly what they require
- You can provide options to your setup functions

### Tearing setup functions down

However, there's still a problem:

- What happens when you do some cleanup of the things you created?

With `test.beforeEach` this is easy:

```js
test.always.afterEach(t => {
  t.context.component.destroy();
  t.context.foo.destroy();
  t.context.bar.destroy();
});
```

But with setup functions, it gets trickier. You end up writing something like this:

```js
let _component;
let _foo;
let _bar;

let createComponent = () => {
  let _component = new Component();
  return _component;
};

let createFoo = () => {
  let _foo = new Foo();
  return _foo;
};

let createBar = () => {
  let _bar = new Bar();
  return _bar;
};

test.always.afterEach(() => {
  _component.destroy();
  _foo.destroy();
  _bar.destroy();
});

// ...
```

### Tearing setup functions down - Part 2

Except... that doesn't work:

- `_foo` and `_bar` are not always created so you have to check if they exist first
- tests are (generally) run concurrently, so these values get re-assigned and you teardown some instances multiple times and others not at all

Instead you need to write it like:

```js
let createComponent = t => {
  t.context.component = new Component();
  return t.context.component;
};

let createFoo = () => {
  t.context.foo = new Component();
  return t.context.foo;
};

let createBar = () => {
  t.context.bar = new Component();
  return t.context.foo;
};

test.always.afterEach(t => {
  if (t.context.component) t.context.component.destroy();
  if (t.context.foo) t.context.foo.destroy();
  if (t.context.bar) t.context.bar.destroy();
});

test('component when foo does this thing', t => {
  let component = createComponent(t);
  let foo = createFoo(t);
  t.true(component.doThisThing(foo));
});

// ...
```

This is a lot to have to remember, so I want to propose an alternative:

# The (Proposed) Solution: `t.teardown()`

What if we instead tried placing the teardown code within the setup functions?

```js
const test = require('ava');

let createComponent = t => {
  let component = new Component();
  t.teardown(() => component.destroy());
  return component;
};

let createFoo = t => {
  let foo = new Foo();
  t.teardown(() => foo.destroy());
  return foo;
};

let createBar = t => {
  let bar = new Bar();
  t.teardown(() => bar.destroy());
  return bar;
};

test('component when foo does this thing', t => {
  let component = createComponent(t);
  let foo = createFoo(t);
  t.true(component.doThisThing(foo));
});

test('component when bar does this thing', t => {
  let component = createComponent(t);
  let bar = createBar(t);
  t.true(component.doThisThing(bar));
});

test('component when bar and baz does this thing', t => {
  let component = createComponent(t);
  let bar = createBar(t);
  let baz = createBar(t);
  t.true(component.doThisThing(bar, baz));
});
```

`t.teardown(cb)` would:

- Schedule `cb()` to be run after the currently running test (which it knows because it's based on TestContext)
- Always run regardless if the test passed or failed

This solves the entire problem space:

- Setup code is only run when needed
- Tests write out exactly what setup code they need
- Tests and setup code is extremely portable
- Setup code can be configured by the tests
- Setup code gets torn down correctly with async tests
- Teardown code only runs when needed

I would like to see AVA implement something like this and promote it within the docs.

## Open Questions

- If `t.teardown()` throws an error, should it fail the current test or report a separate failure?

> There is a $80.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/avajs/ava/issues/1873).

ulken:
**WIP**

Just started looking into this. Here's a somewhat working version, but unsure if right approach/interpreted specification correctly.

Any feedback greatly appreciated!

- Runs within test execution context
- Teardowns are executed sequentially, in order
- Awaits any promises
- Fails test with first error (but executes all callbacks)

**Todo**
- [x] Documentation
- [x] ~Integration tests?~
- [x] ~Separate error handling?~
- [x] Investigate assertions within `t.teardown()`

Closes #1873


index.d.ts
```diff
@@ -344,6 +345,11 @@ export interface TimeoutFn {
 	(ms: number): void;
 }
 
+export interface TeardownFn {
+	/** Declare a hook that is run after the test, regardless if it passed or failed. */
+	(): ImplementationResult;
```
novemberborn, line 350: This should take the hook as an argument, and return `void`.

index.d.ts
```diff
@@ -315,6 +315,7 @@ export interface ExecutionContext<Context = unknown> extends Assertions {
 	plan: PlanFn;
 	timeout: TimeoutFn;
 	try: TryFn<Context>;
+	teardown: TeardownFn;
```
novemberborn, line 318:
```suggestion
	teardown: TeardownFn;
	timeout: TimeoutFn;
	try: TryFn<Context>;
```

lib/test.js
```diff
@@ -699,10 +724,19 @@ class Test {
 		};
 	}
 
-	finishPromised() {
```
novemberborn, line 702: Looking at how this is called, I think we can get rid of this function entirely and just make `finish()` `async`. But that's an unrelated refactor.

lib/test.js
```diff
@@ -699,10 +724,19 @@ class Test {
...
-		return new Promise(resolve => {
+	async finishPromised() {
+		const testResult = await new Promise(resolve => {
 			resolve(this.finish());
 		});
+
+		const teardownResult = await this.runTeardowns();
```
novemberborn, line 732: We should run the teardown right after `verifyAssertions()`: https://github.com/avajs/ava/blob/276c16afc85bddc2c68592a002a96eb4cf3521cc/lib/test.js#L672

Then combined with calling `saveFirstError()` if the teardown crashes, the rest of the function will correctly record the test as having failed.

What's interesting is that the teardown function may still run assertions. The test would fail if your only assertion is run inside a teardown, which is fine. And they wouldn't count towards the plan. But we should document that, and test for it.

lib/test.js
```diff
@@ -457,6 +462,26 @@ class Test {
...
+		this.teardowns.push(callback);
+	}
+
+	async runTeardowns() {
+		const errors = [];
+		for (const teardown of this.teardowns) {
+			try {
+				await teardown(); // eslint-disable-line no-await-in-loop
+			} catch (error) {
+				errors.push(error);
```
novemberborn, line 475: (I don't know how GitHub will end up ordering these comments, so you may have to read ahead for this to make sense…)

Instead of returning all errors below, let's use `saveFirstError()`.

ulken: Fair enough. Makes sense when reading ahead.

lib/test.js
ulken: Yeah, I was considering it but didn't have enough know-how to make that call. Also, as far as I could tell, up until now `finish()` didn't really do anything `async`-ish.

When you say unrelated, do you want me to do that in a separate PR?

lib/test.js
ulken: This was the part I was most unsure of: when to actually run them. Sounds good.

Yeah, I know. It's a little weird, but that's how it is, I guess. Will do, thanks!

index.d.ts
ulken: Went with `(hook: () => PromiseLike<any> | void): void`

ulken:
@novemberborn Thanks for the feedback. Exactly what I was after. Wanted to know if I somewhat wandered down the right path or if I had totally misinterpreted something.

Regarding testing (with reference to #2300): are you fine with having the unit tests or do you want them converted to integration tests? No matter what, I will of course add integration test as well.

lib/test.js
ulken: I'm gonna go ahead and do this in a separate PR as there are so many lines affected, to not clutter this PR unnecessarily.

ulken: Done: #2463

lib/test.js
ulken: Hum, adding a test for _running assertion in `teardown()`_ revealed: 

1. `Error: Assertion <X>, but test has already finished`
2. The `assertCount` is indeed incremented

What's your take on that?

ulken: For example, see: https://github.com/avajs/ava/blob/da09b00eee34ed7304a7fb11072979cc4e29439f/lib/test.js#L324

index.d.ts
```diff
@@ -344,6 +345,11 @@ export interface TimeoutFn {
 	(ms: number): void;
 }
 
+export interface TeardownFn {
+	/** Declare a hook that is run after the test, regardless if it passed or failed. */
+	(hook: () => PromiseLike<any> | void): void;
```
novemberborn, line 350:
```suggestion
	/** Declare a function to be run after the test has ended. */
	(fn: () => void): void;
```

It's not a hook, so let's just say "a function"?

`void` sort of means that the return type is ignored. So `Promise<void> | void` is moot.

test-tap/test.js
```diff
@@ -746,6 +747,110 @@ test('timeout is refreshed on assert', t => {
...
+		t.ok(teardown.calledOnce);
+	});
+});
+
+test('only having assertion in teardown fails test', t => {
+	const teardown = sinon.spy();
+	return ava(a => {
+		a.teardown(() => {
+			teardown();
+			a.pass();
```
novemberborn, line 845: This won't be allowed anyhow, so let's remove this test.

test-tap/test.js
```diff
@@ -746,6 +747,110 @@ test('timeout is refreshed on assert', t => {
...
+		a.teardown(teardown);
+		a.pass();
+	}).run().then(result => {
+		t.is(result.passed, false);
+		t.is(result.error.name, 'TeardownError');
+		t.ok(teardown.calledOnce);
+	});
+});
+
+test('teardown throws error with previous error', t => {
```
novemberborn, line 828:
```suggestion
test('teardown errors are hidden behind assertion errors', t => {
```

test-tap/test.js
```diff
@@ -746,6 +747,110 @@ test('timeout is refreshed on assert', t => {
...
+			a.pass();
+			a.end();
+		});
+	}).run().then(result => {
+		t.is(result.passed, true);
+		t.ok(teardown.calledOnce);
+	});
+});
+
+test('teardown throws error', t => {
```
novemberborn, line 816:
```suggestion
test('teardown errors fail the test', t => {
```

test-tap/test.js
```diff
@@ -746,6 +747,110 @@ test('timeout is refreshed on assert', t => {
...
+	return ava(a => {
+		a.teardown(() => teardownPromise);
+		a.pass();
+	}).run().then(result => {
+		t.is(result.passed, true);
+		t.ok(tornDown);
+	});
+});
+
+test('teardowns run sequentially in order', t => {
```
novemberborn, line 786: This should include a scenario where a teardown throws an error. The next teardown should still run.

lib/test.js
```diff
@@ -457,6 +462,20 @@ class Test {
 		this.timeoutTimer = null;
 	}
 
+	teardown(callback) {
```
novemberborn, line 465: `addTeardown` maybe? We can be a bit more explicit here, as opposed to the `t` object which should be terser.

ulken:
> Ouch, yes, this is tricky. Let's not allow assertions then. Which is already covered 😄

Should we save an explicit error or just document that it's not allowed and leave it at that? I'm leaning towards the former, similar to how we would deal with hooks.

> One final thought: hooks are also run by creating a `lib/test.js` instance. But `t.teardown()` makes no sense in hooks. Maybe when we detect that teardowns were added in a hook, we save an error instead of running them?

Agreed. Not entirely sure how you detect that, but I'll look into it.

index.d.ts
ulken: > It's not a hook, so let's just say "a function"?

Cool

> `void` sort of means that the return type is ignored. So `Promise<void> | void` is moot.

I understand the meaning of `void` (generally), but not how `Promise` qualifies as `void`. Does it have some kind of special meaning in TypeScript?

test-tap/test.js
ulken: Assuming you won't reply today, I'm gonna go ahead and replace it with a specific error test instead. If you don't approve, I'll remove it.

lib/test.js
ulken: All for. Wouldn't the same logic apply to for example `timeout()`? `setTimeout` or similar?

test-tap/test.js
ulken: Good call

lib/test.js
ulken: Since it's an end user facing API, should we type check the input? I.e. ensure it's a function being passed in? As far as I can tell, something similar is not done to e.g. `timeout`, but could still be an improvement.

test-tap/test.js
ulken: Silly me. Forget what I said. There is already an error for that, as stated above...

ulken:
> Agreed. Not entirely sure how you detect that, but I'll look into it.

Nope, can't seem to find a current way to detect if a `Test/Runnable` instance is run as a hook or test.

Could add a boolean flag at creation to indicate that, but sounds a little blah. Assuming you have a though of way as you suggested it? Or did you just assume there was one?

lib/test.js
novemberborn: We do that in some places, usually where there's ambiguity. No harm to doing that here too, I suppose.

novemberborn:
> can't seem to find a current way to detect if a `Test/Runnable` instance is run as a hook or test.
> 
> Could add a boolean flag at creation to indicate that, but sounds a little blah. Assuming you have a though of way as you suggested it? Or did you just assume there was one?

I was thinking passing a flag at creation 😄

index.d.ts
novemberborn: > void is a little like the opposite of any: the absence of having any type at all.

https://www.typescriptlang.org/docs/handbook/basic-types.html#void

So `void` is a superset of `Promise<void>`.

ulken:
> I was thinking passing a flag at creation 😄

Hehe, OK. So be it then.

index.d.ts
ulken: Hum, I think I'm gonna have to drop this. Still don't see how _Promise\<T\>_ qualifies as nothing (`void`). In other words, how can `void` be _thenable_? Or do you just rely on `await` wrapping it for you?

novemberborn: > Or do you just rely on `await` wrapping it for you?

Yea. It just doesn't matter. You can `await` anything.

ulken: Sure, but still wouldn’t be able to do `teardownCallback().then(...)`.

lib/test.js
```diff
@@ -457,6 +463,29 @@ class Test {
...
+		if (typeof callback !== 'function') {
+			throw new TypeError('Expected a function');
+		}
+
+		this.teardowns.push(callback);
+	}
+
+	async runTeardowns() {
+		if (this.isHook && this.teardowns.length > 0) {
+			this.saveFirstError(new Error('`t.teardown()` is not allowed in hooks'));
```
novemberborn, line 476: Let's do this in `addTeardown()`.

ulken: That was my initial thought and I might have been splitting hairs, but I took your words literally:

> Maybe when we detect that teardowns were _added_ in a hook, we save an error instead of _running_ them?

To me the words _added_ and _running_ them indicates that we actually do add them and do the check right before running them.

But I'll try to be less literal and stick to what feels most reasonable next time.

ulken:
> Still need to do documentation.

Yeah, I know... not my favorite part 😉. I'm a little unsure how much of the current documentation should be removed/altered. I'm referring to usage of teardown in hooks, for example. Should we keep that intact and just add new documentation for `t.teardown()`?

I mean, they differ in that this is test specific and hooks apply to all tests, but you could choose to push for only using this anyhow because of X, Y, Z.

> Do you think we should fail the test if you add a teardown inside a teardown? Or is that pretty unlikely anyway?

Sounds farfetched to me, but I'm all for helping users not shooting themselves in the foot. How would you go about detecting that, though?

novemberborn:
I've gone ahead and done the no-teardown-while-tearing-down, with test, and added a hook test too.

I've also taken a stab at the documentation.

> I'm a little unsure how much of the current documentation should be removed/altered. I'm referring to usage of teardown in hooks, for example. Should we keep that intact and just add new documentation for `t.teardown()`?
> 
> I mean, they differ in that this is test specific and hooks apply to all tests, but you could choose to push for only using this anyhow because of X, Y, Z.

I've done the minimum, which is documenting that the function exists… what else did you have in mind? Maybe we can put out a call for a recipe and whatnot.

ulken:
Great! Yeah, some sort of example of usage would be nice. Maybe one of @jamiebuilds’s example in the initial issue?

ulken:
Maybe link/reference/promote this over `after.always()` for teardowns?

I won’t be able to have a look until tonight, though, so go ahead if you feel satisfied as is.

PR #2468, opened 2020-04-22
Node.js 14 compatibility

zackschuster:
due to [this line in `lib/globs.js`](https://github.com/avajs/ava/blob/master/lib/globs.js#L7), i get spammed with this warning on node@14:
image

this PR updates the line in question, as well as adds node@14 to the testing matrix.


zackschuster:
for some (surely cursed) reason, `lib/watcher.js` will fail forever if you don't destructure require for  `lib/glob.js` _and_ `lib/provider-manager.js`. i'm sure it's fine.

zackschuster:
with great hubris i decided to add node@14 to the testing matrix. may those above have mercy on my soul.

novemberborn:
What Node.js version is this error coming from? 14?

That fix is fine, though I'm not sure why we'd need to do the same for the import from `url`?

In order to add Node.js 14 properly, if you could merge in `latest` master and then run https://github.com/avajs/ava/blob/master/test-tap/reporters/readme.md in Node.js 14 that'll add the necessary reporter log files. We should also update: 

https://github.com/avajs/ava/blob/9a9351ddbee9c6c76601210c5dd8f5b0d81cc3b0/package.json#L10

To `>=10.18.0 <11 || >=12.14.0 <13 || >=13.5.0 <14 || >=14.0.0`.

zackschuster:
i've added the reporter logs & updated package.json.

you're correct, the warning only came when i installed 14. dropping extra destructuring fell under "why not" for me, but i can revert any changes you'd like :)

novemberborn:
> you're correct, the warning only came when i installed 14. dropping extra destructuring fell under "why not" for me, but i can revert any changes you'd like :)

Yes let's focus this on fixing the Node.js 14 compatibility issue you encountered.

zackschuster:
done :)

novemberborn:
Awesome, thanks @zackschuster!

zackschuster:
:D thank you @novemberborn!

PR #2470, opened 2020-04-24
Fix watch mode with mini reporter

### LINKED ISSUE #2458: Manual rerun doesn't work when verbose mode is off
Hello,   
I'm using AVA with watch mode, and noticed [the manual rerun][1] doesn't work if `verbose` mode is off in the config.

### Setup

related package.json part:
```jsonc
{
  /* ... */
  "scripts": {
    "test": "ava --config ava.config.verbose-off.js",
    "test:verbose": "ava --config ava.config.verbose-on.js"
  },
  "devDependencies": {
    "ava": "^3.7.0"
  }
}
```

**ava.config.verbose-off.js**:
```js
export default {
        verbose: false,
        watch: true,
}
```

**ava.config.verbose-on.js**:
```js
export default {
        verbose: true,
        watch: true,
}
```

> I made a showcase repository to give a quick example of the problem: https://github.com/pcdevil/ava-verbose-test (after cloning, `npm install` is necessary)

### Expected behaviour
When I run the `npm run test` command the tests can be manually rerun by typing `r + <Enter>`.

### Actual behaviour
The documented feature doesn't triggers AVA to rerun the tests, while using the `npm run test:verbose` task works properly.

### Potential cause
I see AVA uses the [MiniReporter][2] and  [VerboseReporter][3] respectively, and while [Watcher.observeStdin()][4] registers callback for the `"data"` event on `stdin`, it will never be called somehow when the MiniReporter is used.

Due to the lack of domain knowledge I didn't dig deeper, but I see the two reporters have very similar codebase and I suspect a modification was not applied on both classes which made the MiniReporter behave incorrectly.

The amount of code duplication and the potential cause leads me to believe a targeted refactor would be beneficial to eliminate future discrepancies.   
With some guidance from a maintainer I'd happily work on this, but I didn't want to just jump and into and endless pit :)

[1]: https://github.com/avajs/ava/blob/master/docs/recipes/watch-mode.md#manually-rerunning-all-tests
[2]: https://github.com/avajs/ava/blob/master/lib/reporters/mini.js
[3]: https://github.com/avajs/ava/blob/master/lib/reporters/verbose.js
[4]: https://github.com/avajs/ava/blob/master/lib/watcher.js#L332

pcdevil:
As title says, I solved the issue with the mini reporter + watch mode:

It turned out, the terminal spinner package, ora [doesn't pipes through the stdin by default][1], which was a [breaking change in its 4.0.0 relase][2]. I suppose during one of the dependency update the effect of this change was overlooked and caused the issue.

I ran the test locally on Linux, everything was green.

Please review and merge it accordingly!

Fixes #2458.

[1]: https://github.com/sindresorhus/ora#discardstdin
[2]: https://github.com/sindresorhus/ora/releases/tag/v4.0.0

### DIFF
--- a/lib/reporters/mini.js
+++ b/lib/reporters/mini.js
@@ -86,7 +86,7 @@ class MiniReporter {
 		this.spinner = ora({
 			isEnabled: true,
 			color: options.spinner ? options.spinner.color : 'gray',
-			discardStdin: false,
+			discardStdin: !options.watching,
 			hideCursor: false,
 			spinner: options.spinner || (process.platform === 'win32' ? 'line' : 'dots'),
 			stream: options.reportStream
--- a/lib/reporters/mini.js
+++ b/lib/reporters/mini.js
@@ -86,6 +86,7 @@ class MiniReporter {
 		this.spinner = ora({
 			isEnabled: true,
 			color: options.spinner ? options.spinner.color : 'gray',
+			discardStdin: false,
 			hideCursor: false,
 			spinner: options.spinner || (process.platform === 'win32' ? 'line' : 'dots'),
 			stream: options.reportStream


novemberborn:
@pcdevil good find! I've pushed one tweak, since we should still discard stdin if you're not using watch mode.

pcdevil:
You are right, thanks for the merge!

Issue #2048, opened 2019-02-22
Mutexes / external resources


be5invis:
Sometimes we need to limit some of the tests to run serial, or having a maximum parallel jobs, or wait some external resource to be prepared. Therefore introducing such concepts would be helpful for network-oriented tests or stress tests.

Purposed API:

```typescript
test('Some stress test', async t => {
    const handle = await t.resources('stress-test-capacity').acquire(1000);
    // The "stress-test-capacity" is a kind of resource, we need 1000 of them.
    //
    // All resources would be automatically released after the test ends.
    // Alternatively, use handle.release() to manually release a resource.
});
```

sindresorhus:
@be5invis It's not clear what `1000` means in your example. And what does `resources('memory')` actually do? Is `memory` the key? Would be nice if you could elaborate a bit more about the proposed API.

be5invis:
@sindresorhus 

`t.resource` would return a resource manager of a given resource kind (i.e., memory of DB connection), which allocates resource with a given amount.

```typescript
interface TestContext {
    resource(id: string): IResourceManager ;
}

interface IResourceManager {
    // Acquire resource with given amount
    acquire(amount: 1000): Promise<IResourceHandle>;
}

interface IResourceHanlde {
    release(): void;
}
```

sindresorhus:
But what is `1000` and who enforces/controls its usage? How would AVA enforce this? AVA could easily control concurrency, but not memory usage and other things.

be5invis:
@sindresorhus 
`1000` is simply a number, a number representing the resource quantity that the test claimed it will use. It has no link to the actual memory usage. Ah I should name it `stress` instead of `memory`.
Resource managers will take this number, and decide whether to resolve the promise: resolve, wait or throw a error (unable to allocate).

novemberborn:
Within a test file there's plenty of Node.js modules to help you achieve this. I assume you want to coordinate resources across test files. Could you elaborate on that use case?

be5invis:
@novemberborn Yes, coordinate resources across tests.
I have a lib with some stress tests, may take a lot of memory or whatever, so I want to limit the parallelism among them -- they are distributes in multiple files, belonging to multiple sub modules.

lo1tuma:
IMHO there is no need for builtin support in ava. We use [`throat`](https://www.npmjs.com/package/throat) to limit the amount of parallel runs of tests using puppeteer.

be5invis:
@lo1tuma
[...]

Issue #1366, opened 2017-04-21
Proposal: test.resource()


jeffijoe:
# The problem

When integration-testing web apps (which AVA is great for since it runs tests in parallel), I sometimes run into issues with what I think is resource constraints, even on my high-end Macbook Pro. This results in the following sadness:

![image](https://cloud.githubusercontent.com/assets/2505474/25265880/55c5d34c-266f-11e7-879d-c7785472700c.png)

To fix this, I need to run AVA with `--c 6` to lower the concurrency which is fine but if my MBP is using more resources for other things, I need to lower it even more.

The reason for this, is that I in each test suite run my HTTP server, set up a bunch of helpers for testing my API, connect to websockets, and then finally run my test. This is my pattern:

```js
test('GET /todos returns all todos', withServer(async (t, api) => {
  const todos = await api.getTodos()
  t.true(todos.length > 0)
}))
```

Where `withServer` is just a factory function that sets up the server:

```js
const axios = require('axios')
const createServer = require('../src/server')

function withServer (testFn) {
  return async function (t) {
    const server = (await createServer()).listen()
    const host = `http://127.0.0.1:${server.address().port}`
    const client = axios.create({ baseURL: host })
    const api = {
      getTodos: () => client.get('/todos').then(r => r.data)
    }
    
    await testFn(t, api)
  }
}
```

This makes it super easy for me to write integration tests for my API. Problem is, each test need their own connection to the server (in case of websockets), and even when I memoize the server (which only works per suite due to each suite being in a separate process), I still run into the above issue.

I think this is because I am using `rethinkdbdash` which manages a large connection pool. For a single server instance or 4 that might be fine, but for 10+, that might be why it starts having issues.

# The proposed solution

What I would really like to do, is set up just a single instance for all my tests to work with. 

One way to do it is to run `npm start & npm test`, but if `npm start` takes too long to actually start the server, then that's gonna be a problem.

Instead, what if we had a way to tell AVA to run a JS function before starting the test processes? For example, if I was able to do this:

```js
// integration-test.resource.js
import test from 'ava'
import createServer from '../src/createServer'

// This runs before any other test processes have been spun up
test.resource('Set up API server', async (t) => {
  const server = (await createServer()).listen()
  const host = `http://127.0.0.1:${server.address().port}`
  
  // Only JSON-serializable values can be provided to test processes
  t.resource('serverHost', host)
  
  return async function dispose () {
    // When all tests are done, dispose of the resource
    server.close()
  }
})
[...]

PR #2478, opened 2020-04-27
Proof of concept: Shared worker plugins

novemberborn:
"Shared workers" are code that runs in AVA's main process, that can communicate with test workers.

The idea is that a test file can require a module which registers itself as a plugin. The main process launches a worker thread (so this is a Node.js 12+ feature) which can communicate with all test files.

The worker thread would survive across watcher runs, under the assumption that the worker code came from a third-party and so doesn't need to be reloaded.

Advanced serialization is used for the communication between the test worker, main processes and worker thread. Consequently this requires Node.js 12.16. Though I'm also using some features introduced in Node.js 12.17.

This should give us the low-level primitives for https://github.com/avajs/ava/issues/2048 and https://github.com/avajs/ava/issues/1366.


novemberborn:
Wow I think I got this working (let's see what CI says). There's a proof of concept here: https://github.com/avajs/ava/tree/shared-worker-plugins/test/shared-workers/proof-of-concept/fixtures

But I'll try and build something real on top of this next.

novemberborn:
I'm looking to sneak this into a release, but I still have to update the issues, plan work, document things, and ship some example packages…

PR #2482, opened 2020-05-04
Split written and pronounced section

RohanTalip:


novemberborn:
Thanks for your PR @RohanTalip. I think the current text is fine, though.

PR #2488, opened 2020-05-12
Merge mini and verbose reporter implementations

### LINKED ISSUE #2217: Consolidate mini and verbose reporters
Currently we have two different reporter implementations. Code is duplicated across them which leads to various bugs. I once tried to consolidate them but [that work stalled](https://github.com/avajs/ava/pull/1776).

~~The idea is that we would have only one reporter. By default it provides minimal output. We show a spinner and the status of the last test, so you get a sense of activity, and once an entire test file completes do we print failures. In verbose mode, we'd print the results of all tests & hooks, including logs. Various counts for passing and failing tests are always shown at the end of the output.~~

~~If a TTY is not available, we wouldn't print a spinner and only write the appropriate output for a test file once it completes.~~

~~We'd use [`ink`](https://www.npmjs.com/package/ink) to construct the output.~~

Michael55555:
This pull request fixes #2217.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#653: mini reporter should surface failures faster.](https://issuehunt.io/repos/26820798/issues/653)
---
</details>


lib/cli.js
```diff
@@ -259,8 +259,8 @@ exports.run = async () => { // eslint-disable-line complexity
 
 	const ciParallelVars = require('ci-parallel-vars');
 	const Api = require('./api');
-	const VerboseReporter = require('./reporters/verbose');
-	const MiniReporter = require('./reporters/mini');
+	const VerboseReporter = require('./reporters/base-reporter');
+	const MiniReporter = require('./reporters/base-reporter');
```
novemberborn, line 263:
```suggestion
	const Reporter = require('./reporters/default');
```

lib/cli.js
```diff
@@ -396,14 +396,16 @@ exports.run = async () => { // eslint-disable-line complexity
...
+			mode: 'verbose'
 		});
 	} else {
 		reporter = new MiniReporter({
 			projectDir,
 			reportStream: process.stdout,
 			stdStream: process.stderr,
-			watching: combined.watch
+			watching: combined.watch,
+			mode: 'verbose'
```
novemberborn, line 408: This shouldn't be verbose.

Let's see if we can't merge the `if` / `else` into a single instantiation.

test-tap/reporters/mini.js
```diff
@@ -30,8 +30,10 @@ const run = (type, sanitizers = []) => t => {
 		},
 		reportStream: tty,
 		stdStream: tty,
-		watching: type === 'watch'
+		watching: type === 'watch',
+		mode: 'mini'
```
novemberborn, line 34: Before we land this, let's rename `mini` to `default`, tests included.

test-tap/reporters/mini.js
```diff
@@ -10,7 +10,7 @@ const path = require('path');
 const {test} = require('tap');
 const TTYStream = require('../helper/tty-stream');
 const report = require('../helper/report');
-const MiniReporter = require('../../lib/reporters/mini');
+const BaseReporter = require('../../lib/reporters/base-reporter');
```
novemberborn, line 13: Let's make this import consistent with `lib/cli.js`.

test-tap/reporters/verbose.js
```diff
@@ -4,7 +4,7 @@ const {test} = require('tap');
 const {restoreClock} = require('../helper/fix-reporter-env')();
 const TTYStream = require('../helper/tty-stream');
 const report = require('../helper/report');
-const VerboseReporter = require('../../lib/reporters/verbose');
+const BaseReporter = require('../../lib/reporters/base-reporter');
```
novemberborn, line 7: Let's make this import consistent with `lib/cli.js`.

lib/reporters/base-reporter.js
```diff
@@ -0,0 +1,1005 @@
...
+const figures = require('figures');
+const indentString = require('indent-string');
+const ora = require('ora');
+const plur = require('plur');
+const prettyMs = require('pretty-ms');
+const trimOffNewlines = require('trim-off-newlines');
+const beautifyStack = require('./beautify-stack');
+
+const chalk = require('../chalk').get();
+const codeExcerpt = require('../code-excerpt');
```
novemberborn, line 16:
```suggestion

const chalk = require('../chalk').get();
const codeExcerpt = require('../code-excerpt');
const beautifyStack = require('./beautify-stack');
```

lib/reporters/base-reporter.js
```diff
@@ -0,0 +1,1005 @@
...
+	}
+
+	ensureEmptyLine() {
+		if (!this.lastLineIsEmpty) {
+			this.writeLine();
+		}
+	}
+}
+
+class BaseReporter {
```
novemberborn, line 100:
```suggestion
class Reporter {
```

lib/reporters/base-reporter.js
```diff
@@ -0,0 +1,1005 @@
...
+		this.reportStream = options.reportStream;
+		this.stdStream = options.stdStream;
+		this.watching = options.watching;
+		this.relativeFile = file => path.relative(options.projectDir, file);
+		this.consumeStateChange = whileCorked(
+			this.reportStream,
+			this.consumeStateChange
+		);
+
+		if (this.mode === 'verbose') {
```
novemberborn, line 112: Maybe this should be a boolean, rather than a string?

lib/reporters/base-reporter.js
```diff
@@ -0,0 +1,1005 @@
...
+
+	ensureEmptyLine() {
+		if (!this.lastLineIsEmpty) {
+			this.writeLine();
+		}
+	}
+}
+
+class BaseReporter {
+	constructor(options) {
```
novemberborn, line 101: It'd be nice to destructure this, so the options are more explicit.

lib/reporters/base-reporter.js
```diff
@@ -0,0 +1,1005 @@
...
+
+			if (this.watching && plan.runVector > 1) {
+				this.reportStream.write(chalk.gray.dim('\u2500'.repeat(this.lineWriter.columns)) + os.EOL);
+			}
+
+			cliCursor.hide(this.reportStream);
+			this.lineWriter.writeLine();
+
+			this.spinner.start();
+		}
```
novemberborn, line 211: Most code in both branches is duplicated. Let's reduce that.

lib/reporters/base-reporter.js
```diff
@@ -0,0 +1,1005 @@
...
+				}
+			}
+		};
+
+		const fileStats = this.stats && event.testFile ? this.stats.byFile.get(event.testFile) : null;
+
+		if (STATES[event.type]) {
+			STATES[event.type](fileStats);
+		} else if (STATES[this.mode][event.type]) {
+			STATES[this.mode][event.type](fileStats);
```
novemberborn, line 514: Why do we need to duplicate these handlers?

lib/reporters/base-reporter.js
```diff
@@ -0,0 +1,1005 @@
...
+		}
+
+		if (this.previousFailures > 0) {
+			this.lineWriter.writeLine(colors.error(`${this.previousFailures} previous ${plur('failure', this.previousFailures)} in test files that were not rerun`));
+		}
+
+		if (this.watching) {
+			this.lineWriter.writeLine();
+		}
+	}
```
novemberborn, line 1003: There's still too much duplication in this function.

lib/reporters/base-reporter.js
Michael55555: Yes, this is a good idea. I have been thinking about it too but I just wasn't quite sure if it would be too much change.

lib/reporters/default.js
```diff
@@ -0,0 +1,1125 @@
...
+			},
+			'worker-finished': () => {
+				if (
+					!event.forcedExit &&
+					!this.filesWithMissingAvaImports.has(event.testFile)
+				) {
+					if (fileStats.declaredTests === 0) {
+						if (!this.isVerbose) {
+							this.filesWithoutDeclaredTests.add(event.testFile);
+						}
```
novemberborn, line 394: Here and elsewhere, I don't think there's any downside to recording this even when verbose. Let's simplify the code.

lib/reporters/default.js
```diff
@@ -0,0 +1,1125 @@
...
+	}
+
+	consumeStateChange(event) {
+		const fileStats =
+			this.stats && event.testFile ?
+				this.stats.byFile.get(event.testFile) :
+				null;
+		const write = this.isVerbose ?
+			this.lineWriter.writeLine.bind(this.lineWriter) :
+			this.writeWithCounts.bind(this);
```
novemberborn, line 220: It might be better to have a instance method which writes correctly?

lib/reporters/default.js
```diff
@@ -0,0 +1,1125 @@
...
+		} else {
+			this.dest.write(chunk);
+		}
+
+		callback();
+	}
+
+	_writev(pieces, callback) {
+		if (!this.spinner) {
+			return;
```
novemberborn, line 49: It might be better to have two writers, one with, and one without a spinner. There are a lot of unstated assumptions in the `if` / `else` conditions that are to do with how the reporter is instantiated.

lib/reporters/default.js
```diff
@@ -0,0 +1,1125 @@
...
+		}
+
+		this.reset();
+
+		this.failFastEnabled = plan.failFastEnabled;
+		this.matching = plan.matching;
+		this.previousFailures = plan.previousFailures;
+		this.emptyParallelRun = this.isVerbose ?
+			plan.status.emptyParallelRun :
+			this.emptyParallelRun;
```
novemberborn, line 184: This coupling against `verbose` is surprising. Yes, parallel runs only run in CI, and that's when the reporter is in `verbose` mode, but we should either always use `plan.status.emptyParallelRun` or have an option for this.

lib/reporters/default.js
```diff
@@ -0,0 +1,1125 @@
...
+
+		if (this.watching || plan.files.length > 1) {
+			this.prefixTitle = (testFile, title) =>
+				prefixTitle(plan.filePathPrefix, testFile, title);
```
novemberborn, line 188: I don't think we have any line-wrapping rules in our linter? So on the one hand that means this is fine, but on the other hand I find it quite jarring 😄

novemberborn:
Something came up and I haven’t yet had a chance to look at this again.

novemberborn:
@Michael55555 I've made a bunch more tweaks. I think it's good to go now.

Besides me having already closed it, I don't think this would have solved #653?

Michael55555:
@novemberborn Sounds good! Yes, I just revisited the issue, it's indeed not related to this PR. I can't remember why I referenced it.🤔

novemberborn:
Great, thank you @Michael55555!

Michael55555:
Happy to help @novemberborn

PR #2490, opened 2020-05-14
Experimentally implement `t.like()` assertion 

### LINKED ISSUE #845: Add assertion that just compares specific items in the tree
[Issuehunt badges](https://issuehunt.io/r/avajs/ava/issues/845)

This would be especially handy for AST type tests.

Consider the following

``` js
const ast = {
  loc: ... // I usually don't care about this.
  start: ... // or this
  end: ... // or this
  type: 'MemberExpression', // usually I DO care about this
  object: {
     // maybe I care about the contents of this, maybe I don't
  },
  property: {
    // again, maybe
  }
};
```

It would be nice to have a sort of "`deepStrictEqual` but only for the portion of the graph I specify"

I am going to use `t.like`:

``` js
t.like(ast, {
  type: 'MemberExpression',
  object: {
    type: 'Identifier',
    name: 'foo'
  }
});
```

In the above, I'm basically saying "I want to verify it's a `MemberExpression` Node, whose `object` property is an `Identifier` with name `foo`". `loc`, `start`, `end` are ignored throughout, as is the `property` property.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $80.00)

- [issuehunt issuehunt](https://issuehunt.io/u/issuehunt) ($80.00)

### Submitted pull Requests
- [#2490 Add `like` assertion (fix #845)](https://issuehunt.io/r/avajs/ava/pull/2490)
---

### Tips

</details>

futpib:
Fixes #845.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#845: Add assertion that just compares specific items in the tree](https://issuehunt.io/repos/26820798/issues/845)
---
</details>


novemberborn:
Thanks for your PR @futpib. I'll need some time to consider this.

Could you elaborate on the comparison algorithm and the trade-offs you made?

lib/like-pattern-concordance.js
```diff
@@ -0,0 +1,36 @@
...
+		return visited.get(likePattern);
+	}
+
+	if (isPlainObject(likePattern) && isPlainObject(actual)) {
+		const actualPicked = {};
+
+		visited.set(likePattern, actualPicked);
+
+		for (const key of Object.keys(likePattern)) {
+			if (key in actual) {
```
novemberborn, line 16: I think this should use `Reflect.has(actual, key)`, depending on the answer to the above.

lib/like-pattern-concordance.js
```diff
@@ -0,0 +1,36 @@
...
+				actualPicked[key] = pickLikePatternPaths(actual[key], likePattern[key], visited);
```
novemberborn, line 17: This is probably not an issue, but Concordance avoids recursion, whereas this will eventually hit stack limits.

lib/like-pattern-concordance.js
```diff
@@ -0,0 +1,36 @@
+'use strict';
+const concordance = require('concordance');
+const isPlainObject = require('is-plain-object');
+
+function pickLikePatternPaths(actual, likePattern, visited = new WeakMap()) {
+	if (visited.has(likePattern)) {
+		return visited.get(likePattern);
+	}
+
+	if (isPlainObject(likePattern) && isPlainObject(actual)) {
```
novemberborn, line 10: What should we do with arrays, and more complex objects like maps and sets?

lib/assert.js
```diff
@@ -386,6 +387,25 @@ class Assertions {
...
+			if (!checkMessage('like', message)) {
+				return;
+			}
+
+			const {actualPicked, result} = likePatternConcordance.compare(actual, likePattern, concordanceOptions);
+			if (result.pass) {
+				pass();
+			} else {
+				const actualDescriptor = result.actual || concordance.describe(actualPicked, concordanceOptions);
+				const expectedDescriptor = result.expected || concordance.describe(likePattern, concordanceOptions);
```
novemberborn, line 400: When would `actual` and `expected` not exist?

test/assertions/like-pattern-concordance.js
```diff
@@ -0,0 +1,21 @@
+const test = require('@ava/test');
+
+const likePatternConcordance = require('../../lib/like-pattern-concordance');
```
novemberborn, line 3: It should be possible to write an integration test, instead of requiring the module directly.

docs/03-assertions.md
```diff
@@ -203,6 +203,10 @@ Assert that `value` is not the same as `expected`. This is based on [`Object.is(
 
 Assert that `value` is deeply equal to `expected`. See [Concordance](https://github.com/concordancejs/concordance) for details. Works with [React elements and `react-test-renderer`](https://github.com/concordancejs/react).
 
+### `.like(value, likePattern, message?)`
+
+Assert that `value` is like `likePattern`. Comparison is the same as in `.deepEqual()` except only paths present in the `likePattern` object are compared. All paths not in `likePattern` are ignored and can hold any value. Use this if you want to test for some keys and ignore the others.
```
novemberborn, line 208: It's unclear what a "path" is.

lib/like-pattern-concordance.js
futpib: They are left as is to be compared by concordance, anything that's not a plain object is. I don't have a good argument for why it should be this way, I just went with examples from #845.

lib/assert.js
futpib: I just went with how `deepEqual` [is implemented](https://github.com/avajs/ava/pull/2490/files#diff-f76cd6d4ba3dad8b6b9a393af1fce15bR360). This test errors if `|| ...` part is removed: [`assertions.like('foo', 'bar')`](https://github.com/avajs/ava/pull/2490/files#diff-7ebbd1d7d5643b04618ce7d1c59ce9b8R969)

lib/like-pattern-concordance.js
futpib: I have [a test for circular pattern](https://github.com/avajs/ava/pull/2490/files#diff-7ebbd1d7d5643b04618ce7d1c59ce9b8R857). I don't have a test for more-than-stack-limit-deep pattern though. But patterns are expected to be written by hand so I hope this is not an issue too.

lib/like-pattern-concordance.js
futpib: Is there a difference between `Reflect.has` and `in`-operator? I don't think there is, so this does not matter (unless this is an issue of code style or something).

docs/03-assertions.md
futpib: A series of keys to address a deeper property of an object. I can change it to "keys", this will be easier to understand but maybe not as accurate. Should I change this?

test/assertions/like-pattern-concordance.js
futpib: There is an integration test, not in `ava` though.

futpib:
> Could you elaborate on the comparison algorithm and the trade-offs you made?

Comparison is that of concordance (same as in `deepEqual`) except before comparison the expected object is transformed deeply to keep only properties present in the `likePattern`. [`pickLikePatternPaths`](https://github.com/avajs/ava/pull/2490/files#diff-51b1fcb354b6b3b923b134c1468913e0R5) is where this is implemented. I did not find another way to make concordance only look at certain property paths.

lib/like-pattern-concordance.js
novemberborn: `in` walks the prototype chain, `Reflect.has()` does not.

futpib:
```
> Reflect.has({}, 'toString')
true
```

Did you mean `Object.hasOwnProperty`?

novemberborn: Yes, but `Reflect.has()` is the better version of that.

novemberborn:
Thinking this through, proper partial comparisons require an implementation in Concordance. However your proposal has enough value to ship now.

The algorithm then is:

* Given an assertion `t.like(actual, expected)`
* `expected` must be a plain object
* Deeply iterate through the own-keys of `expected`
  * If these are complex values that are not plain objects, such as arrays, sets, maps, or class instances, then those should be preserved and not recursed
* Construct a similarly shaped object from `actual`
* The assertion fails if we end up comparing an empty object
* The objects are compared using Concordance

What do you think?

futpib:
This accurately describes the algorithm except it does not fail if the `expected` is empty, which I will fix.

lib/assert.js
```diff
@@ -386,6 +388,35 @@ class Assertions {
 			}
 		});
 
+		this.like = withSkip((actual, likePattern, message) => {
+			if (!checkMessage('like', message)) {
+				return;
+			}
+
+			if (isPlainObject(likePattern) && Object.keys(likePattern).length === 0) {
```
novemberborn, line 396: The pattern itself *must* be a plain object, otherwise folks should be using `deepEqual`.

docs/03-assertions.md
novemberborn: We should describe the algorithm here.

test-tap/test.js
```diff
@@ -237,6 +237,67 @@ test('handle falsy testing of objects', t => {
...
+			bar: 'bar',
+			cat: 'cake'
+		});
+	});
+	return instance.run().then(result => {
+		t.is(result.passed, true);
+		t.is(instance.assertCount, 1);
+	});
+});
+
```
novemberborn, line 300: I don't think we need the additional tests here.

test/assertions/like-pattern-concordance.js
novemberborn: Yea we can remove it.

novemberborn:
I've addressed the remaining issues. I've also decided to make this an experimental feature for the time being.

fregante:
I must say, this is cool.

PR #2491, opened 2020-05-18
Update endpoint testing recipe to use correct options for got

paulrobertlloyd:
Got library uses `prefixUrl` not `baseUrl`. Also, got errors if path begins with a slash, so replaced `/user` with `user`.

### DIFF
--- a/docs/recipes/endpoint-testing.md
+++ b/docs/recipes/endpoint-testing.md
@@ -25,8 +25,8 @@ test.after.always(t => {
 });
 
 test.serial('get /user', async t => {
-	const res = await got('user', { prefixUrl: t.context.prefixUrl, json: true });
-	t.is(res.body.email, 'ava@rocks.com');
+	const {email} = await got('user', {prefixUrl: t.context.prefixUrl}).json();
+	t.is(email, 'ava@rocks.com');
 });
 ```
 
@@ -34,4 +34,3 @@ Other libraries you may find useful:
 
 - [`supertest`](https://github.com/visionmedia/supertest)
 - [`get-port`](https://github.com/sindresorhus/get-port)
-
--- a/docs/recipes/endpoint-testing.md
+++ b/docs/recipes/endpoint-testing.md
@@ -17,7 +17,7 @@ const app = require('../app');
 
 test.before(async t => {
 	t.context.server = http.createServer(app);
-	t.context.baseUrl = await listen(t.context.server);
+	t.context.prefixUrl = await listen(t.context.server);
 });
 
 test.after.always(t => {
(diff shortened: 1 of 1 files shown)


docs/recipes/endpoint-testing.md
```diff
@@ -17,15 +17,15 @@ const app = require('../app');
...
+	t.context.prefixUrl = await listen(t.context.server);
 });
 
 test.after.always(t => {
 	t.context.server.close();
 });
 
 test.serial('get /user', async t => {
-	const res = await got('/user', { baseUrl: t.context.baseUrl, json: true });
+	const res = await got('user', { prefixUrl: t.context.prefixUrl, json: true });
```
novemberborn, line 28: I think `json: true` has also changed in more recent Got releases?

sindresorhus:
I fixed the example for latest Got.

docs/recipes/endpoint-testing.md
```diff
@@ -17,21 +17,20 @@ const app = require('../app');
 
 test.before(async t => {
 	t.context.server = http.createServer(app);
-	t.context.baseUrl = await listen(t.context.server);
+	t.context.prefixUrl = await listen(t.context.server);
```
sindresorhus, line 20: It would actually be nicer to assign a new Got instance here instead. Then you could set all the common settings here and reuse it for each test. Example: https://github.com/sindresorhus/got#custom-endpoints

sindresorhus: `const {email} = await got('user', {prefixUrl: t.context.prefixUrl}).json();` could then simply be `const {email} = await t.context.fetch('user');`

paulrobertlloyd: I’d like to use `got` in my tests as that’s what I’m using in my application, but I’ve tied myself in knots trying to use it for endpoint testing. This is what I’m currently doing, using `supertest` and a setup function:

```js
import test from 'ava';
import supertest from 'supertest';

import app from '../../index.js';
const request = supertest(app);

const mockResponse = async query => {
  return request.get('/micropub')
    .set('Accept', 'application/json')
    .query(query);
};

test('Returns configuration', async t => {
  const response = await mockResponse('q=config');
  t.is(response.status, 200);
});
```

Should I be using `http.createServer` and `test-listen` in place of `supertest`? I can’t seem to get the piece in the right order to get the expected response.

novemberborn: > It would actually be nicer to assign a new Got instance here instead.

@sindresorhus yes, but this is an endpoint testing recipe, not a Got recipe 😉 

---

@paulrobertlloyd I'm not familiar with `supertest`, but just looking at your code snippet I don't see any mocks being set up.

FWIW I tend to use `http.createServer()` and wire up my fake server myself rather than learning additional libraries.

sindresorhus: > @sindresorhus yes, but this is an endpoint testing recipe, not a Got recipe 😉

But you're using Got for it, and I was showing how to my the recipe more succinct.

That's like saying I should not use `t.throws`, but instead try/catch in a recipe about testing a React app, because it's not an AVA recipe.

PR #2510, opened 2020-06-11
Remove ESM from Puppeteer recipe

Autre31415:


novemberborn:
Good spot, thanks @Autre31415!

PR #2515, opened 2020-06-14
Improve TypeScript recipe

SephReed:
Moved the instructions on setup to the top of the file.  Changed the order of `tsc` vs `ts-node`.  The latter supports all projects.  The former only works for some.


novemberborn:
I agree that the mention of `tsc` is misleading. I don't want to promote `ts-node` over a plain TypeScript setup, however.

PR #2524, opened 2020-06-21
Support assertion message in `t.timeout()`

### LINKED ISSUE #2443: Support assertion message in t.timeout()
Please provide details about:

* What you're trying to do

I'd like to display a message to folks observing test runs on a particular timeout to indicate that they need to check some configuration locally. This particular scenario arises from bad user config that is detectable only at runtime (thanks AWS). 

* Why you can't use AVA for this

I've sifted through the docs and can't find a good way to do this. I haven't taken a look at the source for a hack as I've been burned by doing that in the past and upgrading major versions.

* And maybe how you think AVA could handle this

Spitballing here, but perhaps:

```js
test('verify email token', async (t) => {
  t.timeout(2000, 'Please check yo stuffs!');
  const result = await longRunningTask();
  t.assert(result);
});

// or if this is something you all would prefer to indicate for an entire suite

test.timeout(2000, 'Please check all the things');
```

jonathansamines:
With with enormous guidance from @novemberborn I was able to add support for custom assertion messages on tests timeouts, implemented as described at #2443 

Added a tests and documentation, but please let me know if additional work is required.

**Edit:** Complimentary pull request on eslint plugin can be found at https://github.com/avajs/eslint-plugin-ava/pull/303

Fixes #2443


docs/07-test-timeouts.md
```diff
@@ -22,3 +22,12 @@ test('foo', t => {
 	// Write your assertions here
 });
 ```
+
+When timeouts are set for individual tests, you can also provide an optional message.
+
+```js
+test('foo', t => {
+	t.timeout(100, 'test exceeds the expected time'); // 100 milliseconds
```
sindresorhus, line 30: I would show an actually useful message here. `test exceeds the expected time` adds nothing over what AVA shows by default and hence doesn't show the user why this parameter might be useful.

index.d.ts
```diff
@@ -353,7 +353,7 @@ export interface TimeoutFn {
 	 * Set a timeout for the test, in milliseconds. The test will fail if the timeout is exceeded.
 	 * The timeout is reset each time an assertion is made.
```
sindresorhus, line 354: The doc comment should mention the `message` parameter and what it's useful for.

lib/test.js
```diff
@@ -431,15 +431,15 @@ class Test {
 		this.planError = planError;
 	}
 
-	timeout(ms) {
+	timeout(ms, message) {
```
sindresorhus, line 434: I think there should be some validation that ensures the `message` is a string if it's defined and error out if not. To prevent incorrect usage.

docs/07-test-timeouts.md
jonathansamines: Updated example, bit a still minimal but more realistic usage of the custom timeout feature.

index.d.ts
jonathansamines: Updated docs. Two questions though:

-  Most of existing type definitions don't document all arguments. Should we consider updating them?
- Are these regular jsdoc comments? For now I added the new argument documentation in a new line, but if jsdocs can be used I can also update the documentation to use that format.

lib/test.js
jonathansamines: Validation and tests added. I now throw errors on incorrect usage just like most of existing validations. I also used similar tests a reference to verify the corresponding behavior.

test-tap/test.js
```diff
@@ -750,6 +750,26 @@ test('timeout is refreshed on assert', t => {
 	});
 });
 
+test('timeout with custom message', t => {
```
novemberborn: I'm pushing some code so that we can test this behavior using AVA itself. Don't worry it's still early days for self-hosted tests.

lib/test.js
```diff
@@ -431,15 +431,19 @@ class Test {
...
+		if (message && typeof message !== 'string') {
+			throw new TypeError('Expected a string');
```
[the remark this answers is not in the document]
jonathansamines: Sounds great, a couple of questions:

- I based my implementation on other existing argument validations on the same module. Are we looking into moving them all to use assertion failures instead?
- What's the difference between an assertion failure and any other kind of error?

test-tap/test.js
jonathansamines: Tests are much cleaner that way. Is the `@ava/test` package the ava test runner itself?

jonathansamines:
@novemberborn I added a couple of questions, but the pull request by itself looks great. Thanks for your help.

lib/test.js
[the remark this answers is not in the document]
novemberborn: > * I based my implementation on other existing argument validations on the same module. Are we looking into moving them all to use assertion failures instead

Maybe, maybe not. Thrown exceptions can be caught, but when we save an error it is guaranteed to fail the test. Failed assertions look better in our output, too.

In this case having the message be validated in the same way as other assertion messages seems like a better outcome than being consistent _within_ the `test.js` file.

test-tap/test.js
novemberborn: Glad you like it!

Yes it's a wrapper so we get a `test-ava` bin stub.

lib/test.js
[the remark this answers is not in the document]
jonathansamines: Got it, thanks for explaining!

test-tap/test.js
jonathansamines: Nice, thanks!

jonathansamines: @novemberborn Is there any roadmap to move to move the existing test suit to use self-hosted tests? I quite enjoy working with tests, and would like to help with the migration process, if that is the goal.

novemberborn:
> Is there any roadmap to move to move the existing test suit to use self-hosted tests? I quite enjoy working with tests, and would like to help with the migration process, if that is the goal.

@jonathansamines ultimately we'd love to self host all our tests, yes. We're not necessarily expecting it for new functionality since we're still figuring out how this would work. E.g. there's additional changes I'm exploring in #2478.

That said… if you're keen and not adverse to some experimentation it'd be great if you would pick this up.

The idea is to write integration tests, hosted in a stable version of AVA, where we run the AVA version under test in a child process and then assert on its output. We expose some of the internals so we're not just asserting on `stdout` / `stderr` but have access to individual test results and other structures.

Then we can organize the tests by area, with ultimately a ton of directories and files. Which is a bit verbose but hopefully provides better encapsulation.

https://github.com/avajs/ava/tree/master/test-tap/integration would be a good place to start. https://github.com/avajs/ava/blob/master/test-tap/api.js is also full of integration style tests.

Issue #2523, opened 2020-06-21
Disable snapshots and try assertions in hooks


novemberborn:
Per https://github.com/avajs/ava/issues/2511, `t.try()` does not work in hooks. Rather than crashing we should fail the hook properly.

Snapshots work in hooks. It's OK in pre-test hooks, since those have unique titles, but the behavior is a little odd in the file hooks. I don't see how this is useful. Let's make snapshots-in-hooks fail. Unfortunately this is a breaking change, so we'll have to put this behind a flag.

okyantoro:
Hello @novemberborn 
is it enough to put the following hook checking before line https://github.com/avajs/ava/blob/master/lib/test.js#L94 
```js
if (test.isHook) {
  throw new Error(`t.try() is not allowed in hooks.`);
}
```
and for the `t.snapshot()` part, it seems the test need to pass `test.isHook` in `ExecutionContext.super()` method parameter to make the `Assertions.snapshot()` know if the test is a hook and call `fail()` properly. Or maybe there is any other suggestion in doing this properly? I am not too familiar with the source code.

novemberborn:
@okyantoro that sounds about right. It may be nice to add a more explicit `disableSnapshots` property or something like that.

Since you seem keen I'll assign this to you. Apologies if that was presumptuous.

okyantoro:
Thank you @novemberborn 
I have implemented the code. Let me know if additional work is required.

PR #2527, opened 2020-06-24
Detect t.try() usage in hooks; experimentally disable t.snapshot()

okyantoro:
This change will disable snapshots and try assertions in hooks. But I am not sure with the unit test. Let me know if additional work is required.

Fixes https://github.com/avajs/ava/issues/2523


lib/assert.js
```diff
@@ -690,6 +691,14 @@ class Assertions {
 		});
 
 		this.snapshot = withSkip((expected, ...rest) => {
+			if (disableSnapshots) {
+				fail(new AssertionError({
+					assertion: 'snapshot',
+					message: '`t.snapshot()` is not allowed in hooks',
```
novemberborn, line 697:
```suggestion
					message: '`t.snapshot()` cannot be used here',
```

lib/test.js
```diff
@@ -40,7 +40,8 @@ class ExecutionContext extends assert.Assertions {
 				return test.compareWithSnapshot(options);
 			},
 			powerAssert: test.powerAssert,
-			experiments: test.experiments
+			experiments: test.experiments,
+			disableSnapshots: test.isHook === true
```
novemberborn, line 44: This is a breaking change. We need to [add an experiment](https://github.com/avajs/ava/blob/master/lib/load-config.js#L10) for this and then _also_ check if that's been set in `test.experiments`.

test-tap/hooks.js
```diff
@@ -566,3 +567,37 @@ test('teardowns cannot be used in hooks', async t => {
...
+		await t.try(tt => tt.pass());
+
+		t.pass();
+	});
+
+	const result = await instance.run();
+
+	t.notOk(result.passed);
+	t.match(result.error.savedError.message, /not allowed in hooks/);
+});
```
novemberborn, line 603: I'd like to self-host these tests. See https://github.com/avajs/ava/tree/master/test/

I'm thinking a directory for "hooks" and then two tests in there? Have a look at the `test-timeouts` in https://github.com/avajs/ava/pull/2524/files.

lib/assert.js
okyantoro: addressed

lib/test.js
okyantoro: Addressed

lib/assert.js
```diff
@@ -690,6 +691,15 @@ class Assertions {
 		});
 
 		this.snapshot = withSkip((expected, ...rest) => {
+			if (disableSnapshots && experiments.disableSnapshotsInHooks) {
+				fail(new AssertionError({
+					assertion: 'snapshot',
+					message: '`t.snapshot()` cannot be used here',
```
tymfear, line 697: If I understand it correctly, `t.snapshot()` and `t.try` are only allowed to be used in tests.
If that's so, I would suggest changing error to something like
```
`t.snapshot()` can be used in tests only
```

test-tap/hooks.js
okyantoro: Addressed. I still use same naming convention like `test-timeouts` for the directory, but we can change it if you prefer `hooks`.

okyantoro: Addressed. I still use same naming convention like test-timeouts for the directory, but we can change it if you prefer hooks.

tymfear:
@novemberborn when something is released under the feature flag, when it is a good time to update docs?
Changes in this PR should be reflected in [04-snapshot-testing.md](https://github.com/avajs/ava/blob/master/docs/04-snapshot-testing.md) and [03-assertions.md](https://github.com/avajs/ava/blob/master/docs/03-assertions.md#trytitle-implementation--macro--macro-args).

lib/assert.js
novemberborn: Yes. Changing to "can only be used in tests"

okyantoro:
Thank you @novemberborn

PR #2540, opened 2020-07-14
Add possibility to configure how extensions should be loaded

### LINKED ISSUE #2345: Add configuration to specify how extensions should be loaded
We need a way for AVA to know whether to load a particular test file (given its extension) as `commonjs` or `module`, similar to the [`"type"` field in `package.json`](https://nodejs.org/dist/latest-v13.x/docs/api/esm.html#esm_code_package_json_code_code_type_code_field).

It must not be possible to change how `cjs` and `mjs` extensions are loaded. And I have a strong suspicion (but I have not confirmed this) that it's impossible to override the `"type"` field (which applies to `js` extensions).

Once configured as `module`, we then need to `import()` these test files. This is blocked by #2344.

macarie:
Should resolve #2345 by adding an object form for the `"extensions"` configuration.

---

> And I have a strong suspicion (but I have not confirmed this) that it's impossible to override the `"type"` field (which applies to `js` extensions).

I did some tests and it doesn't work, if `"type"` is `"module"` you can't `require()` a `js` file from that package, nor the opposite, if you try Node gets angry:

```console
require() of [...]/index.test.js from [...]/node_modules/ava/lib/worker/subprocess.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
```

I've tried to keep the edits as small and self-contained as possibile. I'm not sure if what I did and how I did it is the right approach, but it seems to be working.

This is how AVA behaves now:

1. If `"extensions"` is not defined the default module configuration for `mjs`, `cjs`, and `js` is used.
2. If `"extensions"` is an array the module types are generated automatically for each extension, keeping the default type for `mjs`, `cjs`, and `js`, only if they're defined, and using `"commonjs"` for everything else, thus maintaining backwards compatibility.
3. If `"extensions"` is an object the module types are read from the configuration, but the type for `mjs`, `cjs`, and `js` is locked and can only be set to their default type, the default type for `js` is based on the `"type"` field because it cannot be overridden.

Now, in `package.json` (or in a config file), it's possible to have:
```json
{
  "ava": {
    "extensions": {
      "ts": "module",
      "js": "commonjs",
      "cjs": "commonjs",
      "mjs": "module"
    }
  }
}
```

If you try to change `js` to `"module"` (notice the missing `"type"`) it'll throw an error:
```console
  ✖ .js files can only be configured as ’commonjs’ when the nearest parent package.json does not contain "type": "module", found ’module’ instead.
```

For `cjs` and `mjs` files:
```console
  ✖ .cjs files can only be configured as ’commonjs’, found ’module’ instead.
```

Local tests are passing, not with Node `v14.5.0` tho. I don't know why they fail (something with the reporters I guess), but they also fail on `master`, so I'm not sure what's going on.

I didn't write tests yet, and no documentation either, if what I did looks good I'll do those next 👌🏼


lib/module-types.js
```diff
@@ -0,0 +1,51 @@
...
+	};
+
+	const createModuleTypesFromExtensionsArray = extensions => {
+		const generatedModuleTypes = {};
+
+		for (const extension of extensions) {
+			if (extension in lockedModuleTypes) {
+				generatedModuleTypes[extension] = lockedModuleTypes[extension];
+			} else {
+				generatedModuleTypes[extension] = 'commonjs';
```
macarie, line 32: This is kinda useless for now, I know, but in the future _unknown_ extensions (`.ts`, `.re`, etc) could be loaded with the default type, just like `.js` is. Doing that would be as easy as changing `"commonjs"` to `defaultModuleType`. That's gonna be a breaking change tho.

novemberborn:
Thanks for the PR! It may take me a few more days before I can have a good look at this, sorry.

novemberborn:
@macarie I wasn't online much this weekend so I haven't been able to look at this yet, sorry.

lib/module-types.js
```diff
@@ -0,0 +1,51 @@
...
+		cjs: 'commonjs',
+		mjs: 'module',
+		js: defaultModuleType
+	};
+
+	const createModuleTypesFromExtensionsObject = extensionsObject => {
+		const configuredModuleTypes = {};
+
+		for (const [extension, type] of Object.entries(extensionsObject)) {
+			if (extension in lockedModuleTypes && type !== lockedModuleTypes[extension]) {
```
novemberborn, line 12: Possibly irrelevant (see below), but I prefer `Reflect.has()` these days. I find `in` very easy to overlook.

```suggestion
			if (Reflect.has(lockedModuleTypes, extension) && type !== lockedModuleTypes[extension]) {
```

lib/module-types.js
```diff
@@ -0,0 +1,51 @@
...
+				throw new TypeError(`.${extension} files can only be configured as ’${lockedModuleTypes[extension]}’${
+					extension === 'js' ?
+						` when the nearest parent package.json ${type === 'module' ? 'does not contain' : 'contains'} "type": "module"` : ''
+				}, found ’${type}’ instead.`);
```
novemberborn, line 16: Rather than requiring users to specify the correct type for the standard extensions, perhaps we can let users set them to `true` (and nothing else). That way you do not need to change your AVA configuration if you change the package type, nor do you need to be needlessly explicit. What do you think?

macarie: Yeah, I think it might work, I don't see any drawbacks with that. Should it throw if it's something else (`false | string`)? Should other extensions still only allow `'commonjs' | 'module'`?

macarie:
> You didn't mention it, but we still need to use this configuration to load the test file right?

I'm not sure I understand what you are asking, what do you mean?

lib/module-types.js
novemberborn: > Should it throw if it's something else (`false | string`)?

Yes, let's do that.

> Should other extensions still only allow `'commonjs' | 'module'`?

Indeed.

novemberborn:
> > You didn't mention it, but we still need to use this configuration to load the test file right?
> 
> I'm not sure I understand what you are asking, what do you mean?

Never mind, I'd forgotten how these module types are used.

macarie:
I did the changes you proposed, now it only accepts `true` for `'js' | 'cjs' | 'mjs'` and `'commonjs' | 'module'` for everything else.

I don't really like the error messages, something else that might work?
```
An extension's module type can only be one of ("commonjs" | "module"), found "ts": "modul" instead.
```
```
.cjs, .mjs, and .js files can only be enabled using ’true’ because their default import type cannot be overridden, found "js": "module" instead.
```

lib/module-types.js
```diff
@@ -0,0 +1,65 @@
...
+			if (type !== true) {
+				const lockedExtensions = Object.keys(lockedModuleTypes).map(key => `.${key}`);
+
+				throw new TypeError(`${lockedExtensions.slice(0, -1).join(', ')}, and ${lockedExtensions.slice(-1)} files can only be enabled using ’true’ because their default import type cannot be overridden, found ${JSON.stringify(extension)}: ${JSON.stringify(type)} instead.`);
+			}
+
+			return lockedModuleTypes[extension];
+		}
+
+		if (acceptedModuleTypes.has(type) === false) {
```
novemberborn, line 20:
```suggestion
		if (!acceptedModuleTypes.has(type)) {
```

lib/module-types.js
```diff
@@ -0,0 +1,65 @@
...
+			throw new TypeError(`An extension's module type can only be one of ("${[...acceptedModuleTypes].join('" | "')}"), found ${JSON.stringify(extension)}: ${JSON.stringify(type)} instead.`);
+		}
+
+		return type;
+	};
+
+	const createModuleTypesFromExtensionsObject = extensionsObject => {
+		const configuredModuleTypes = {};
+
+		console.log(extensionsObject);
```
novemberborn, line 30:
```suggestion
```

macarie:
@novemberborn I've tried doing some tests.

Node v10 fails because I've used an `mjs` file, not sure how to make it work tho, setting `nodeArguments: ['--experimental-modules']` in `ava-std.config.js` doesn't seem to do anything.

The `ts` test files are "fake" and not loaded as TS, `require()` doesn't care about their extension and still treats them as JS, do you want to have something _more realistic_ (= loads real TS files with types)?

I wouldn't try setting anything as `"module"` because for non-standard extensions we need a loader, and that's still not stable.

As always, any feedback is welcome 🙂

novemberborn:
> Node v10 fails because I've used an `mjs` file, not sure how to make it work tho, setting `nodeArguments: ['--experimental-modules']` in `ava-std.config.js` doesn't seem to do anything.

Just on this, have a look at this branch: https://github.com/avajs/ava/blob/shared-worker-plugins/ava.config.js

It's OK to skip these tests in Node.js 10.

novemberborn:
@macarie cool!

I'm about to push some tweaks. My apologies for the refactoring, hopefully you agree it's an improvement and not just me being silly 😄 

You took a good stab at the tests. I think it's better to use separate test files, and then use that to exclude ESM tests from Node.js 10. Snapshots should be used with caution, it's easy to think you're testing something even though you aren't. I've also renamed some files.

I've made the feature experimental for now, using the `configurableFileLoading` flag. How does that name sound to you?

Your feedback on my changes aside, I think what's left here now is to update the documentation?

macarie:
Nice work @novemberborn!

> My apologies for the refactoring, hopefully you agree it's an improvement and not just me being silly 😄

Yeah, everything's more straightforward now and a bit easier to follow, more _Go-ish_ I'd say 😝

> I've made the feature experimental for now, using the `configurableFileLoading` flag. How does that name sound to you?

Not sure about it. I think it gives more the idea that you can configure _what_ files to load and not _how_ to load them 🤔
Maybe `configurableFileFormatLoading` or something along this line would be better.

> I think what's left here now is to update the documentation?

Yes! Do you want to write it or should I give it a shot?

lib/module-types.js
```diff
@@ -0,0 +1,75 @@
...
+				moduleTypes.cjs = 'commonjs';
+				break;
+			case 'mjs':
+				moduleTypes.mjs = 'module';
+				break;
+			case 'js':
+				moduleTypes.js = defaultModuleType;
+				break;
+			default:
+				moduleTypes[extension] = 'commonjs';
```
macarie, line 50: What if this `default` behaviour was an option @novemberborn?

Defaults to `'commonjs'` — keeping backwards compatibility — but also configurable as `'module'`.

novemberborn:
> Yeah, everything's more straightforward now and a bit easier to follow, more _Go-ish_ I'd say 😝

Please don't tell my coworkers 😉 

> > I've made the feature experimental for now, using the `configurableFileLoading` flag. How does that name sound to you?
> 
> Not sure about it. I think it gives more the idea that you can configure _what_ files to load and not _how_ to load them 🤔
> Maybe `configurableFileFormatLoading` or something along this line would be better.

How about `configurableModuleFormat`?

> What if this `default` behaviour was an option?
> 
> Defaults to `'commonjs'` — keeping backwards compatibility — but also configurable as `'module'`.

I'd like to follow the `package.json` `"type"` option for the time being. If you can write your sources in ESM then you should write your tests the same way.

novemberborn:
> > I think what's left here now is to update the documentation?
> 
> Yes! Do you want to write it or should I give it a shot?

It'd be great if you could give it a shot first.

macarie:
> How about `configurableModuleFormat`?

This sounds good to me.

> I'd like to follow the `package.json` `"type"` option for the time being. If you can write your sources in ESM then you should write your tests the same way.

Sorry, I was talking about line 50, the default for "unrecognized" extensions, not the `"js"` default type. I'm not used to GitHub's review tool and apparently the line I click on is not the one in focus. 

> It'd be great if you could give it a shot first.

Sure 🙂

novemberborn:
I think `commonjs` is a good default for now. ESM is still very new.

macarie:
Sorry for the delay, it's been a busy few weeks.

Anyway... I've updated the documentation, I wasn't sure where to write about the new feature tho, so I chose `06-configuration.md`. Is there a better place?

I've also changed `configurableFileLoading` to the one you have proposed, `configurableModuleFormat`, and changed the test's folder name, if it doesn't look right I can revert the commit.

novemberborn:
3.12 is now available.

PR #2543, opened 2020-07-19
Self-host the built-in Node.js assert tests

jonathansamines:
As this discussed [here](https://github.com/avajs/ava/pull/2524#issuecomment-657223814) were are looking to move the existing test suit to use ava self-hosted tests. I am moving [node-assertions](https://github.com/avajs/ava/blob/master/test-tap/integration/node-assertions.js) tests as the first step on that direction.

Before porting additional tests, I wanted to get some feedback on this one. I'd also like to understand if porting an integration tests one by one would be ok or if we would prefer to group some of them together?

cc @novemberborn


test/README.md
```diff
@@ -1,6 +1,6 @@
 # Self-hosted tests
 
-This directory contains tests that are run using a stable version of AVA. You can run them using `npx test-ava`.
+This directory contains tests that are run using a stable version of AVA. You can run them using `npx test-ava` at the project's root.
```
jonathansamines, line 3: Maybe I am not that familiar with the project yet, but I struggled a bit to understand why `npx test-ava` wasn't working until I tried it at the project's root.

test/helpers/exec.js
```diff
@@ -25,15 +26,15 @@ const compareStatObjects = (a, b) => {
 	return 1;
 };
 
-exports.fixture = async (...args) => {
+exports.fixture = async (args, options = {}) => {
```
jonathansamines, line 29: At the existing tests for [node-assertions](https://github.com/avajs/ava/blob/master/test-tap/integration/node-assertions.js) custom environment variables where needed. I updated this code to make it simpler to provide overrides for them. A consequence of this change, is that *args* should be provided as an array now (just like it happens in *execa*) and I had to update the existing tests to also specify args that way.

Please let me know if I should update this to use rest args.

test/README.md
novemberborn: It's because there's a `node_modules` inside `test/` as well. Maybe we should symlink `test-ava` from there.

test/helpers/exec.js
novemberborn: Yea… I guess this was inevitable 😄

novemberborn:
I wasn't online much this weekend so I haven't been able to look at this yet, sorry.

test/README.md
jonathansamines: Sounds good, will give it a try.

jonathansamines: Symlink added

test/node-assertions/test.js
```diff
@@ -0,0 +1,27 @@
...
+
+// The AssertionError constructor in Node 10 depends on the TTY interface
+test('node assertion failures are reported to the console when running in a terminal', async t => {
+	const options = {
+		env: {
+			AVA_SIMULATE_TTY: true,
+			AVA_TTY_COLOR_DEPTH: 8,
+			AVA_TTY_HAS_COLORS: typeof process.stderr.hasColors === 'function'
+		}
+	};
```
novemberborn, line 12: I don't think this does anything? The TTY simulation is a `test-tap` helper.

jonathansamines: Right, I am now loading that helper into the ava self hosted tests. As a side note, I had to provide a more complete implementation to the fake TTY stream as [ora](https://github.com/sindresorhus/ora) was [relying](https://github.com/sindresorhus/ora/blob/master/index.js#L232-L236) on them for certain operations.

jonathansamines:
I have solved the existing comments and build is passing now 🔥

novemberborn:
@jonathansamines I've pushed a few more tweaks. Let's see if CI still passes… 😱 Let me know what you think and then I'll ship this. Thank you!

jonathansamines:
Changes look good @novemberborn

PR #2547, opened 2020-07-23
Detect active inspectors; update VSCode debugging recipe

### LINKED ISSUE #2542: Can't debug with Visual Code 1.47
Can't debug with Visual Code 1.47, works fine with 1.46.1

connor4312:
VS Code has a new debugger which works differently than the old debugger. Some arguments in the launch config were no longer needed or caused issues, and I also added a method using the debug terminal for users more comfortable with that.

Fixes #2542

### DIFF
--- a/docs/recipes/debugging-with-vscode.md
+++ b/docs/recipes/debugging-with-vscode.md
@@ -34,12 +34,12 @@ Open the file(s) you want to debug. You can set breakpoints or use the `debugger
 
 Now, *with a test file open*, from the *Debug* menu run the *Debug AVA test file* configuration.
 
-## Debugging with the Debug Terminal
+## Debugging with the debug terminal
 
-You can use VS Code's "JavaScript Debug Terminal" to automatically debug Ava run on the command line.
+You can use VS Code's “JavaScript Debug Terminal” to automatically debug AVA run on the command-line.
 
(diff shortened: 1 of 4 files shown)


novemberborn:
Hey @connor4312 👋 

Thanks for the PR! I've pushed some edits, but nothing major. I'm wondering if there's a page summarizing the "Debug Terminal"? I couldn't immediately find anything.

I've also pushed some code that looks for `VSCODE_INSPECTOR_OPTIONS` and assumes that AVA is being run in a VSCode debug session. That then disables timeouts and switches to the verbose reporter, like `npx ava debug` does. Does that seem appropriate to you?

connor4312:
> I'm wondering if there's a page summarizing the "Debug Terminal"? I couldn't immediately find anything.

Right now this is only in release notese -- I need to update the VS Code documentation to include it now as well, I'll PR here again when that happens.

> Does that seem appropriate to you?

Yep!

novemberborn:
> > I'm wondering if there's a page summarizing the "Debug Terminal"? I couldn't immediately find anything.
> 
> Right now this is only in release notese -- I need to update the VS Code documentation to include it now as well, I'll PR here again when that happens.

Perfect, thank you.

> Actually, a more generic way would be seeing if `require('inspector').url() !== undefined`, which will return true when the process is debuggable (unfortunately there's no way to see if the debugger is actually attached). That will work for VS Code and any other runtimes as well.

Great idea.

PR #2557, opened 2020-08-11
Use ESM syntax for ESM example

AnthumChris:
The ES modules example uses the "require()" syntax for the `test.js` file. This changes to use ES Module "import from ..." syntax in `test.js` example.


novemberborn:
@AnthumChris good spot, thank you! Must have been a little overeager converting our samples to CommonJS back when we moved Babel support out of AVA itself.

PR #2574, opened 2020-09-02
Make `t.passed` usable in tests and teardown functions

### LINKED ISSUE #2572: t.passed is undefined in teardown context
Please provide details about:

* What you're trying to do - save page screenshot in `t.teardown` callback when test has failed
* What happened - `t.passed` is `undefined`.
* What you expected to happen - `t.passed` to be defined. Since `t.teardown` might be used instead of `test.afterEach` hook it should have similar capabilities.

Code example:
```javascript
test('do some web testing', async t =>{
  t.teardown(async () => {
    if(!t.passed) {
       const imageName = `${t.title}.png`;
       t.log(`Saving screenshot "${imageName}" because test has failed`);
       await takeScreenshotAndSave(imageName);
    }
  })
})
```

tymfear:
Set the `t.passed` value before teardowns callbacks are executed to enabled logic based on test result.
Fixes #2572

### DIFF
--- a/docs/02-execution-context.md
+++ b/docs/02-execution-context.md
@@ -24,7 +24,7 @@ Contains shared state from hooks.
 
 ## `t.passed`
 
-Whether a test has passed. This value is only accurate in the `test.afterEach()` and `test.afterEach.always()` hooks.
+When used in `test.afterEach()` or `test.afterEach.always()` hooks this tells you whether the test has passed. When used in a test itself (including teardown functions) this remains `true` until an assertion fails, the test has ended with an error, or a teardown function caused an error. This value has no meaning in other hooks.
 
 ## `t.end()`
 
@@ -36,7 +36,7 @@ Log values contextually alongside the test result instead of immediately printin
 
 ## `t.plan(count)`
 
-Plan how many assertion there are in the test. The test will fail if the actual assertion count doesn't match the number of planned assertions. See [assertion planning](./03-assertions.md#assertion-planning).
+Plan how many assertions there are in the test. The test will fail if the actual assertion count doesn't match the number of planned assertions. See [assertion planning](./03-assertions.md#assertion-planning).
 
 ## `t.teardown(fn)`
 
--- a/lib/test.js
+++ b/lib/test.js
@@ -187,7 +187,8 @@ class ExecutionContext extends assert.Assertions {
 	}
 
 	get passed() {
-		return testMap.get(this).testPassed;
+		const test = testMap.get(this);
+		return test.isHook ? test.testPassed : !test.assertError;
 	}
 
 	_throwsArgStart(assertion, file, line) {
--- a/test-tap/test.js
+++ b/test-tap/test.js
@@ -956,3 +956,33 @@ test('.teardown() is bound', t => {
 		t.ok(teardownCallback.calledOnce);
 	});
 });
+
+test('t.passed value is true when teardown callback is executed for passing test', t => {
+	new Test({
+		fn(a) {
+			a.teardown(() => {
+				t.is(a.passed, true);
+				t.end();
+			});
+			a.pass();
+		},
+		metadata: {type: 'test'},
+		onResult() {},
+		title: 'foo'
+	}).run();
+});
+
+test('t.passed value is false when teardown callback is executed for failing test', t => {
+	new Test({
+		fn(a) {
+			a.teardown(() => {
+				t.is(a.passed, false);
+				t.end();
+			});
+			a.fail();
+		},
+		metadata: {type: 'test'},
+		onResult() {},
+		title: 'foo'
+	}).run();
+});


lib/test.js
```diff
@@ -725,6 +725,7 @@ class Test {
 		this.clearTimeout();
 		this.verifyPlan();
 		this.verifyAssertions();
+		this.testPassed = !this.assertError;
 		await this.runTeardowns();
```
novemberborn, line 729: I'd like to move the `error` and `passed` variable declarations up here, so we don't repeat the logic.

novemberborn (commented):
In response to your questions in #2572 I don't think there's much use in passing a (new) execution context to the teardown function, at least at this stage. You can't run assertions anyway. Using a closure is good enough I think.

lib/test.js
tymfear: The thing is that when you run teardowns, it might change the `this.assertionError` state.
So like
```javascript
		let error = this.assertError;
		let passed = !error;

		await this.runTeardowns();
// at this point if assertions made in teardowns and failed we'll not fail the test, so we have to do
                 error = this.assertError;
                 passed = !error;

		this.duration = nowAndTimers.now() - this.startedAt;
        
```

novemberborn:
@tymfear I think I've refreshed my memory on how `t.passed` is supposed to work. And yea, a teardown can still cause the test to fail.

We've documented that `t.passed` is only accurate in `afterEach` hooks. Perhaps we should change its implementation so that it behaves differently in hooks, as opposed to tests:

https://github.com/avajs/ava/blob/983ce0cb714fe0b35b48834c0911ebd4cf8e71d6/lib/test.js#L189-L191

```js
get passed () {
	const test = testMap.get(this);
	return test.isHook ? test.testPassed : !test.assertError;
}
```

That way we don't change its meaning in hooks, but when `t.passed` is accessed in a test itself you can tell whether it has already failed. What do you think?

We should also update the documentation: https://github.com/avajs/ava/blob/master/docs/02-execution-context.md#tpassed

tymfear:
Looks like a good proposal. Will do that and see if it works as expected

tymfear:
@novemberborn done.

docs/02-execution-context.md
```diff
@@ -24,7 +24,7 @@ Contains shared state from hooks.
 
 ## `t.passed`
 
-Whether a test has passed. This value is only accurate in the `test.afterEach()` and `test.afterEach.always()` hooks.
+Whether a test has passed. This value is only accurate during test execution (including teardown hooks) and in the `test.afterEach()` and `test.afterEach.always()` hooks.
```
novemberborn, line 27:
```suggestion
When used in `test.afterEach()` or `test.afterEach.always()` hooks this tells you whether the test has passed. When used in a test itself (including teardown functions) this remains `true` until an assertion fails, the test has ended with an error, or a teardown function caused an error. This value has no meaning in other hooks.
```

novemberborn (commented):
Looks great @tymfear. What do you think of my documentation suggestion?

docs/02-execution-context.md
tymfear: To be honest, now when I have more or less good understanding of how AVA works (using it quite extensively), this looks to me obvious without additional clarifications. 
Probably (can't really imagine the case) for newcomers that might be useful somehow. 
But understanding the fact that test execution stops on the first assertion error, is enough. 
But most likely I just don't have unbiased view on this.

novemberborn: > But understanding the fact that test execution stops on the first assertion error, is enough.

It doesn't though. AVA's assertions don't throw errors so execution won't stop.

tymfear: Hmm.. that's new. 
Probably I was confused by not having other assertion results for all failed assertions.
Then it's already beneficial :)
It's a topic for another ticket/conversation, but is there a sense to continue test execution if the results of that execution are unknown anyway?

tymfear:
@novemberborn build failure seems to be some kind of flakiness. Could you re-trigger the build, please?

docs/02-execution-context.md
novemberborn: This has been discussed a fair bit elsewhere — apologies for not looking up the issue numbers 👍

novemberborn:
> build failure seems to be some kind of flakiness

I think it might be a change in Node.js 14.9.

tymfear:
Do you expect me to fix that? Or there is some process for situations like this and I just wait for some general fixes and merge the latest master into my branch?

novemberborn:
> Do you expect me to fix that? Or there is some process for situations like this and I just wait for some general fixes and merge the latest master into my branch?

Sorry, am quickly responding in between things.

I can fix it. But we'll need it to start failing on Windows too…

tymfear:
@novemberborn merged with the latest master and everything seems to be working now

PR #2576, opened 2020-09-06
Experimentally disable null expectations for throws assertions

JSimoni42:
Aims to fix https://github.com/avajs/ava/issues/2410

### DIFF
--- a/docs/03-assertions.md
+++ b/docs/03-assertions.md
@@ -244,7 +244,7 @@ Assert that an error is thrown. `fn` must be a function which should throw. The 
 * `name`: the expected `.name` value of the thrown error
 * `code`: the expected `.code` value of the thrown error
 
-`expectation` does not need to be specified. If you don't need it but do want to set an assertion message you have to specify `null`.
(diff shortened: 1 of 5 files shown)


novemberborn (commented):
Thanks for picking this up @JSimoni42!

novemberborn:
Awesome.

What's left now is to update the documentation:

* https://github.com/avajs/ava/blob/master/docs/03-assertions.md#throwsfn-expectation-message
* https://github.com/avajs/ava/blob/master/docs/03-assertions.md#throwsasyncthrower-expectation-message

In both those places it says you need to specify `null`. Since we now allow `undefined` as well let's just recommend that. And then we can say something in parentheses like:

> (AVA 3 also allows you to specify `null`. This will be removed in AVA 4. You can opt in to this change early by enabling the `disableNullExpectations` experiment.)

JSimoni42:
@novemberborn Documentation has been updated. I noticed that https://github.com/avajs/ava/pull/2575 was also failing checks due to the same test error in `Install and test AVA / Node.js (^14.0.0, ubuntu-latest)`. Suffice it to say, it seems like this PR didn't introduce any additional test failures. Let me know if I should squash/merge or if you'd prefer to handle that, yourself

novemberborn:
> Let me know if I should squash/merge or if you'd prefer to handle that, yourself

Should be fine now…

novemberborn:
Thanks @JSimoni42!

Issue #1602, opened 2017-11-28
Run code before and after a group of test files


pvdlg:
## Description

The [before and after hooks](https://github.com/avajs/ava#before--after-hooks) are really useful, but they are limited to a given file.

When doing integration test we often have to do some relatively long set up tasks such as starting a database, a server or a docker container. As those tasks take often several dozen of seconds or a few minutes, we want to do them only once per test run.

As `before` and `after` are limited to the file in which they are defined, it forces us to write all the integration tests in the same file. This is not ideal as it ends up creating a giant file that is difficult to maintain.

So it would be great to be able to define have a special `before` and `after` hook that spans a group of file.

I'm not really sure what would be the best API, but I can think of two solutions.

### `t.beforeAll(scope, [function])` and `t.afterAll(scope, [function])`

Those special version of `beforeAll` and `afterAll` could be defined in any test file (even by themselves without other tests) and the `function` is optional.

All the test files that have a `beforeAll` or an `afterAll` for a given `scope` would have to be run only after the `function` of all the `beforeAll` of the same `scope` ran. And it will run the `function` of all `afterAll` only when the files having the same `scope` are completed.

```js
// integration/mocks.js
test.beforeAll('database', async () => {
  await startDatabase();
});
test.beforeAll.always('database', async () => {
  await stopDatabase();
});
```
```js
// integration/scenario1.js
test.beforeAll('database')

test.serial('First step of scenario 1' async t => {
  // Do something with the database
}

test.serial('Second step of scenario 1' async t => {
  // Do something else with the database
}

test.beforeAll.always('database');
```
```js
// integration/scenario2.js
test.beforeAll('database')

test.serial('First step of scenario 2' async t => {
  // Do something with the database
}

test.serial('Second step of scenario 2' async t => {
  // Do something else with the database
}

test.beforeAll.always('database');
```

### `after.js` and `before.js` in a diretory:
```
.
+-- test/
|   +-- integration
|       +-- before.js
|       +-- scenario1.js
|       +-- scenario2.js
|       +-- scenario3.js
|       +-- after.js
```
`ava` would run `before.js`, then all the tests in `scenario1.js`, `scenario2.js`, scenario3.js` the same way it does now (in parralel or serially is `.serial` is used) then run `after.js`.

[...]

PR #2577, opened 2020-09-08
feat(watch): add on-change cli option

dolsem:
#1602 and #1366 described the use case for executing a piece of code before all tests. I have a similar use case where I need to wipe my test database each time the tests are run. This is more of a problem when AVA is in watch mode because I need to run my logic every time it detects a change. So I propose a CLI option to run a custom shell command every time the test are re-run.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -75,6 +75,10 @@ const FLAGS = {
 		coerce: coerceLastValue,
 		description: 'Re-run tests when files change',
 		type: 'boolean'
+	},
+	'on-change': {
+		description: 'Extra command to run when files change',
+		type: 'string'
 	}
 };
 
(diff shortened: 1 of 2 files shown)


novemberborn:
We've intentionally not added this for regular test runs. I don't think we should add it for watch mode either.

Something like #1366 would be a better fit — as in: you can load code in AVA's main process and it gets some lifecycle events allowing it to do stuff before tests run. #2478 lays the groundwork for that. It's pretty close to shipping experimentally actually. Maybe you could open an issue to explore that further?

Sorry to disappoint. I do understand where you're coming from, and thank you for figuring out how to make it work and opening a pull request. It's just a question of where to fit it into AVA.

FWIW, I write a lot of service integration tests which rely on Firestore. For that database the emulator has a neat trick: if you connect using a new project ID you get a clean database. So the execution of each test file is isolated from the others.

dolsem:
@novemberborn 
Got it. #1366 seems like a much more complex way to solve this problem. In any case, is there anything I can do to help get #2478 done and merged? How soon do you think it will be finished?

novemberborn:
#2478 is close to shipping experimentally, and then we can figure out how to make it work with the run lifecycle.

I appreciate that all of this seems like more effort. Ultimately the goal is to enable many interesting use cases with code that can even be shared through the package registry.

PR #2586, opened 2020-09-19
Assertions return boolean

### LINKED ISSUE #2455: Have assertions return booleans
Spurred by conversation in https://github.com/avajs/ava/issues/2449 and https://github.com/avajs/ava/issues/1485 we'd like for our assertions to return booleans. `true` when they pass, `false` when they fail. This does not apply to the `throws()` and `throwsAsync()` assertions.

Failing assertions will fail the test, but do not throw exceptions. This means your test does not stop executing. Usually that's fine, but there are conditions where you're about to do something that uses a lot of resources, and it's already unnecessary. Returning booleans will help with this.

The assertions are defined here:

https://github.com/avajs/ava/blob/18aeac64db2711d37853175b5e8dc1fd72c66d2a/lib/assert.js#L238

The type definition should be updated to reflect the new return type:

https://github.com/avajs/ava/blob/18aeac64db2711d37853175b5e8dc1fd72c66d2a/index.d.ts#L41

Some tests should be added to (or existing test updated in) https://github.com/avajs/ava/blob/master/test-tap/assert.js.

adiSuper94:
Fixes #2455. Based on #2564.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -39,40 +39,68 @@ export type SnapshotOptions = {
 };
 
 export interface Assertions {
-	/** Assert that `actual` is [truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy). Comes with power-assert. */
+	/**
+	 * Assert that `actual` is
+	 * [truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy),
+	 * returning Boolean indicating whether the assertion passes. Comes with
+	 * power-assert.
+	 */
 	assert: AssertAssertion;
 
-	/** Assert that `actual` is [deeply equal](https://github.com/concordancejs/concordance#comparison-details) to `expected`. */
+	/**
+	 * Assert that `actual` is [deeply
+	 * equal](https://github.com/concordancejs/concordance#comparison-details) to
+	 * `expected`, returning Boolean indicating whether the assertion passes.
+	 */
 	deepEqual: DeepEqualAssertion;
 
-	/** Assert that `actual` is like `expected`. */
+	/**
+	 * Assert that `value` is like `selector`, returning Boolean indicating
+	 * whether the assertion passes.
+	 */
 	like: LikeAssertion;
 
-	/** Fail the test. */
+	/** Fail the test, always returning `false`. */
 	fail: FailAssertion;
 
-	/** Assert that `actual` is strictly false. */
+	/**
+	 * Assert that `actual` is strictly false, returning Boolean indicating
+	 * whether the assertion passes.
+	 */
 	false: FalseAssertion;
 
-	/** Assert that `actual` is [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy). */
+	/**
+	 * Assert that `actual` is
+	 * [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy), returning
+	 * Boolean whether the assertion passes.
+	 */
 	falsy: FalsyAssertion;
 
 	/**
 	 * Assert that `actual` is [the same
-	 * value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) as `expected`.
+	 * value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is)
+	 * as `expected`, returning Boolean indicating whether the assertion passes.
 	 */
 	is: IsAssertion;
 
 	/**
 	 * Assert that `actual` is not [the same
-	 * value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) as `expected`.
+	 * value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is)
+	 * as `expected`, returning Boolean indicating whether the assertion passes.
 	 */
 	not: NotAssertion;
 
-	/** Assert that `actual` is not [deeply equal](https://github.com/concordancejs/concordance#comparison-details) to `expected`. */
+	/**
+	 * Assert that `actual` is not [deeply
+	 * equal](https://github.com/concordancejs/concordance#comparison-details) to
+	 * `expected`, returning Boolean indicating whether the assertion passes.
+	 */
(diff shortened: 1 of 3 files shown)


adiSuper94:
@novemberborn I have edited test cases only for pass(), fail(), is() and not(). 
If this is fine, I'll go ahead and edit the testcases for other assertions and edit the type definitions also.

lib/assert.js
```diff
@@ -286,6 +286,7 @@ class Assertions {
 			},
 			onSuccess: () => {
 				pass();
+				return true;
```
novemberborn, line 289: Do we need to return `false` in `onError` or something?

novemberborn (commented):
@adiSuper94 thanks for redoing the PR.

I've pushed two commits, one for some weird code I spotted, and then the other to return `false` if the messages are illegal.

I think `notThrowsAsync()` should also return a boolean, just because we can. It's only the `throws()` and `throwsAsync()` assertions that return the thrown error *or undefined* (which we haven't typed because it's annoying).

With regards to the tests, I appreciate that the assertion tests are rather complicated. I'm struggling to understand the changes you've made though. Could you elaborate on your approach? For these tests too I'd prefer if we could use `tap`'s assertions to check whether an AVA assertion returned `true` or `false`.

We do have some setup in place now to use AVA to test AVA, but migrating the assertion tests to that should be separate.

sindresorhus:
@adiSuper94 Bump :)

adiSuper94:
@sindresorhus  I have been meaning to work on this for the past few weekends. Thanks for the reminder.

lib/assert.js
adiSuper94: yes, we should.

adiSuper94:
@novemberborn , sorry for the delay in response.
> 
> I've pushed two commits, one for some weird code I spotted, and then the other to return `false` if the messages are illegal.
> 
Thanks, I guess my editor is acting a little crazy.

> I think `notThrowsAsync()` should also return a boolean, just because we can. It's only the `throws()` and `throwsAsync()` assertions that return the thrown error _or undefined_ (which we haven't typed because it's annoying).
> 
And added returns for `notThrowsAsync()`

> With regards to the tests, I appreciate that the assertion tests are rather complicated. I'm struggling to understand the changes you've made though. Could you elaborate on your approach? For these tests too I'd prefer if we could use `tap`'s assertions to check whether an AVA assertion returned `true` or `false`.
> 
> We do have some setup in place now to use AVA to test AVA, but migrating the assertion tests to that should be separate.
>
The approach i took was rather crude. I have changed it now.  Let me know if it clear now.
I have edited the testcases to not use ava.

test-tap/assert.js
```diff
@@ -120,8 +128,7 @@ function eventuallyFailsWith(t, fn, subset) {
 
 function fails(t, fn) {
 	lastFailure = null;
-	fn();
-	if (lastFailure) {
+	if (!fn() && lastFailure) {
```
novemberborn, line 131: This should check more explicitly for a `false` value:
```suggestion
	if (fn() === false && lastFailure) {
```

test-tap/assert.js
```diff
@@ -146,8 +153,7 @@ function eventuallyFails(t, fn) {
 function passes(t, fn) {
 	lastPassed = false;
 	lastFailure = null;
-	fn();
-	if (lastPassed) {
+	if (fn() && lastPassed) {
```
novemberborn, line 156: And conversely:

```suggestion
	if (fn() === true && lastPassed) {
```

novemberborn (commented):
Looks good @adiSuper94. Thanks for coming back to this.

I've added two suggestions to make the tests actually look for `false` and `true` values, rather than falsy/truthy. What do you think?

Would you mind updating the type definitions as well?

novemberborn:
You might be able to cherry-pick the type definitions from https://github.com/avajs/ava/pull/2601.

adiSuper94:
@novemberborn , Sure.

adiSuper94:
> I've added two suggestions to make the tests actually look for `false` and `true` values, rather than falsy/truthy. What do you think?
> 


I accepted those, suggestions. It actually helped me improve the tests. Thanks!

test-tap/assert.js
```diff
@@ -146,10 +153,10 @@ function eventuallyFails(t, fn) {
 function passes(t, fn) {
 	lastPassed = false;
 	lastFailure = null;
-	fn();
-	if (lastPassed) {
+	if (fn() === true && lastPassed) {
 		t.pass();
 	} else {
+		lastFailure = new Error('Assertion failed');
```
novemberborn, line 159: You shouldn't need to assign / override this.

test-tap/assert.js
```diff
@@ -1047,6 +1057,7 @@ test('.throws()', gather(t => {
 		assertions.throws(() => {
 			throw new Error('foo');
 		});
+		return true;
```
novemberborn, line 1060: Why add this return statement? Same below.

test-tap/assert.js
```diff
@@ -120,8 +128,7 @@ function eventuallyFailsWith(t, fn, subset) {
 
 function fails(t, fn) {
 	lastFailure = null;
-	fn();
-	if (lastFailure) {
+	if (fn() === false && lastFailure) {
 		t.pass();
```
novemberborn, line 132: Should this be something like:

```js
const retval = fn();
if (lastFailure) {
  t.false(retval)
}
```

And something similar in `passes` below. Combining the variable check with the return value check makes it hard to know what's being tested.

index.d.ts
```diff
@@ -433,7 +509,7 @@ export interface CbExecutionContext<Context = unknown> extends ExecutionContext<
 	end(error?: any): void;
 }
 
-export type ImplementationResult = PromiseLike<void> | Subscribable | void;
+export type ImplementationResult = PromiseLike<void> | Subscribable | boolean | void;
```
novemberborn, line 512: This shouldn't change. It's the return type of the function you pass to `test()`.

novemberborn (commented):
Thanks @adiSuper94. Couple more questions. And I appreciate these assertion tests are rather cumbersome!

novemberborn:
Closing in favor of #2696.

Issue #2373, opened 2020-01-18
Improve feedback when glob patterns on the CLI did not select any test files


novemberborn:
With AVA 3, you can once again pass glob patterns on the CLI:

```console
npx ava '**/component/test/**/*'
```

However these patterns are now used to filter the tests that AVA already selected based on the configuration. In other words, you can only select from the files that are run when you invoke `npx ava`.

It would be great if we could print a message when no test files were selected, even though normally `npx ava` would run some.

Here's where the we do the selection:

https://github.com/avajs/ava/blob/61e0d05e390dd43dedb723e36f2af9bb752c0f18/lib/api.js#L121

zachesposito:
Hello I would like to take care of this issue. Is this an accurate summary of the expected behavior:
1. When running the CLI with a glob pattern, if no test files are found, but at least one would have been found without the glob pattern, then in addition to printing the `Couldn't find any files to test` message, also print something like `{N} test(s) available based on default configuration, but none matched glob pattern {globPattern}`?
2. When running the CLI with a glob pattern, if no test files are found, but the glob pattern matches at least one TypeScript file, then in addition to printing both messages from the above scenario, also print something like `{globPattern} matches {N} TypeScript file(s). To configure AVA to load TypeScript files, see https://github.com/avajs/ava/blob/master/docs/recipes/typescript.md`?

novemberborn:
Yea that sounds about right @zachesposito, thanks!

zachesposito:
@novemberborn Do you have a preference as to where these feedback messages should live? This is probably the simplest approach: https://github.com/zachesposito/ava/commit/74729a7727cc44cf65695a43182906d779870e54
Which outputs like this:
![image](https://user-images.githubusercontent.com/1486613/95454830-f5b58800-093a-11eb-81f1-fa020f5cdb31.png)
However if you'd prefer to keep `api.js` from outputting any messages directly, another option could be to emit a new event that is either handled in `cli.js` and printed using [`exit()`](https://github.com/avajs/ava/blob/master/lib/cli.js#L12), or that flows all the way through to the reporter and is printed somewhere around [here](https://github.com/avajs/ava/blob/master/lib/reporters/default.js#L667)

novemberborn:
@zachesposito this is easiest to discuss in a PR.

rauschma:
This behavior just bit me again and took me a while to debug (AVA 3.15.0).

Playing devil’s advocate:

[...]

PR #2598, opened 2020-10-09
Provide feedback when cli globs match no tests

zachesposito:
Addresses #2373 
This is the minimalist approach which outputs like this:
image

However if you'd prefer to keep api.js from outputting any messages directly, another option could be to emit a new event that is either handled in `cli.js` and printed using [`exit()`](https://github.com/avajs/ava/blob/master/lib/cli.js#L12), or that flows all the way through to the reporter and is printed somewhere around [here](https://github.com/avajs/ava/blob/master/lib/reporters/default.js#L667)

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -120,11 +120,24 @@ class Api extends Emittery {
 				if (filter.length === 0) {
 					selectedFiles = testFiles;
 				} else {
+					const globPatterns = filter.map(({pattern}) => pattern);
 					selectedFiles = globs.applyTestFileFilter({
 						cwd: this.options.projectDir,
-						filter: filter.map(({pattern}) => pattern),
+						filter: globPatterns,
 						testFiles
 					});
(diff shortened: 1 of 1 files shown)


novemberborn (commented):
Thanks for opening this PR @zachesposito!

My concern is less with where the error copy is, but with the control flow. We'd also have to figure out how this works with watch mode. So yes I think it'd be best to print this in the default reporter.

This would require us to add some additional information to the `RunStatus` object.

I think it's fine to do some globbing in the reporters.

It'd be neat to use the "warning" figure and perhaps the magenta color for this. Or yellow. I can't immediately recall what we do for comparable situations.

I'm not sure if we need to repeat the pattern in the message itself. Depending on your shell it could be expanded and be a large file list.

You can use `plur` to correctly pluralize words.

novemberborn:

PR #2610, opened 2020-11-23
Deterministic snaphot files, declaration-ordered snapshot reports

### LINKED ISSUE #2311: `*.snap` files content should be the same regardless of test execution order
## Description

When a test file has several parallel async tests and each of those tests call `t.snapshot()`, the resulting `*.snap` file is non-deterministic. The contents of the `*.snap` file depends on the order in which tests finished.

In such cases, tests pass correctly. However the `*.snap` files contents keep being updated, which means they are shown as dirty files by `git status` and they result in merge conflicts, especially in PRs.

Making `*.snap` file contents independent of test execution order would solve this issue.

### Test Source

`test.js`

```js
const test = require('ava')

const { promisify } = require('util')
const pSetTimeout = promisify(setTimeout)

test('one', async t => {
  await pSetTimeout(Math.random() * 1e3)
  t.snapshot('one')
})

test('two', async t => {
  await pSetTimeout(Math.random() * 1e3)
  t.snapshot('two')
})
```

Running `ava -u` on this file produces different `*.snap` depending on whether `one` or `two` finishes first.

```
$ ava -u test.js && sha256sum test.js.snap 

  2 tests passed

99b2f48bae8ee408643d0885635d18f1373ab17f2d3988160f5d5fc36e525f47  test.js.snap

$ ava -u test.js && sha256sum test.js.snap 

  2 tests passed

fbaf5758abca960c7b8e1daf7e06785cabed91c641d7daf595fb6939befd7daa  test.js.snap

$ ava -u test.js && sha256sum test.js.snap 

  2 tests passed

99b2f48bae8ee408643d0885635d18f1373ab17f2d3988160f5d5fc36e525f47  test.js.snap
```

### Command-Line Arguments

```
ava -u test.js
```

## Environment

```
$ node -e "var os=require('os');console.log('Node.js ' + process.version + '\n' + os.platform() + ' ' + os.release())"
Node.js v13.3.0
linux 5.3.0-24-generic
$ ava --version
2.4.0
$ npm --version
6.13.4
```

### LINKED ISSUE #2324: Store snapshots by test declaration order
#  Description
Please consider not sorting snapshots when they are recorded - especially to the markdown version. 

## My use case
I use ava with snapshots for testing "pure-ish" functions that are composed into larger workflows. 
Meaning while the functions have side effects, they are testable with input/output expectations. 
Snapshots are ideal for this. 

However, since these are functions with side-effects, for proper testing, they have to be run in serial order to simulate the workflow. This allows the functions to be tested in isolation, while
building upon previous steps. The cost of setup of partial workflows is significant enough that it's not practical to think of think of running each function as a test suite.

Snapshots are a great way to capture the expectations and track deviations over time. 
However in [snapshot-manager.js](https://github.com/avajs/ava/blob/4fdb02dd66b56f0204e09a7970416d5075f49a3f/lib/snapshot-manager.js#L107) these recordings are being sorted in title-order. 

Now my ability to manually inspect the output is severely impacted, requiring all kinds of 
up-and-down scrolling. Some of these are API calls with 10s or even 100s of lines of data.
Sometimes, out of sheer frustration, I put numbers in the test titles, but that turns out to cause
other problems in complex refactoring/requirement-change situations. 

Options I can think of:
1. Not sorting - however, this seems to be exactly counter of [this comment on #2311](https://github.com/avajs/ava/issues/2311#issuecomment-565822461)
2. Sorting only the snap-files, but not the markdown files
3. Sorting parallel tests, but not serial tests.
4. Providing a flag/option to not sort the markdown files.
5. Storing snapshots in order they are encountered in the test files. 

Strikes me that option 5 would seem to solve all problems (including #2311?), allow for side-by-side scrolling of test file and markdown, without requiring any configuration. 

Really unsure of the implementation complexity though.

ninevra:
Closes #2311
Closes #2324

Ensures that `*.snap` files are deterministic by sorting entry blocks by the hash of their test name or id.

Ensures that `*.md` snapshot report files are sorted in as close to declaration order as is reasonably possible, by 
1. emitting a `taskIndex` incremental id for each test case and hook,
2. emitting an `associatedTaskIndex` for each `beforeEach`, `afterEach`, and `afterEachAlways` hook, indicating the associated task case,
2. passing these ids and the pre-existing `snapshotCount` through to `snapshotManager`, and
3. sorting new entries (test cases, hooks, or snapshot ids) according to the `taskIndex`, `associatedTaskIndex`, and `snapshotCount` of the first snapshot therein.

## Caveats

- `*.snap` files are not sorted by declaration order, only `*.md` files. Sorting `*.snap`s by declaration order is more complicated, because when running in `appendOnly` mode (i.e. without the `-u` flag) they may contain entries which are never touched during the course of the run. These entries would then have to be sorted among the others in a deterministic fashion. While this could be accomplished with a fallback to sorting by hash, it's not clear where such entries should appear in the sort order. It's also not clear that this grants any benefit, since after encoding and compressing the order of entries is not readily apparent to a human reader in any case.
- In `appendOnly` mode, only the new report entries are sorted in declaration order, and only among themselves.
- Declaration order is not a terribly useful sort order for snapshots in hooks, particularly `beforeEach` and `afterEach` hooks. Further, `*Each` hooks are run multiple times, so `taskIndex` and `snapshotCount` don't uniquely identify them, requiring a third sort key (`associatedTaskIndex`). This should only matter until the next AVA major version, given #2523.

## Release concerns

As far as I can tell, this change won't affect any existing codebases' `*.snap` or `*.md` files until either AVA is run with `-u` or some change is made to the corresponding tests. `snapshotManager`s existing behavior is to not record anything, even internally, unless `-u` is enabled or there was no preexisting snapshot to compare to. However, this is difficult to definitively test.

No changes have been made to the snapshot file format.

## Remaining issues

When AVA is run with `-u`, projects may see unexpected snapshot updates as their snapshots and reports are resorted. Avoiding this would require reworking `snapshotManager` to: load saved snapshots even when `-u` is present; always report success from `compare()` if `-u` is present; and write updates only if some conflict or new snapshot was encountered.

Should this be placed behind a feature flag, to reduce the incidence of spurious updates?


ninevra:
Sorry, it seems like the testing approach I took doesn't work in CI. I'll have to rethink how to test this.

lib/snapshot-manager.js
```diff
@@ -104,13 +104,32 @@ function combineEntries(entries) {
 	const buffers = [];
 	let byteLength = 0;
 
-	const sortedKeys = [...entries.keys()].sort();
+	const sortedKeys = [...entries.keys()].sort((kA, kB) => {
```
sindresorhus, line 107:
```suggestion
	const sortedKeys = [...entries.keys()].sort((keyA, keyB) => {
```

Try to use descriptive variable names. No abbreviations.

test/snapshot-order/fixtures/report-declaration-order/test.js
```diff
@@ -0,0 +1,55 @@
+const test = require('ava');
+
+const id = i => `index: ${i}`;
```
sindresorhus, line 3:
```suggestion
const id = index => `index: ${index}`;
```

test/snapshot-order/fixtures/report-declaration-order/test.js
```diff
@@ -0,0 +1,55 @@
...
+
+test('E - discard some snapshots in a try()', async t => {
+	t.snapshot(id(8), 'outer');
+	(await t.try('trying', t => {
+		t.snapshot(id(9), 'inner');
+	})).discard();
+	t.snapshot(id(10), 'outer again');
+});
+
+test('D - more snapshots with ids', t => {
```
sindresorhus, line 48:
```suggestion
test('D - more snapshots with IDs', t => {
```

test/snapshot-order/test.js
```diff
@@ -0,0 +1,52 @@
...
+const exec = require('../helpers/exec');
+const fs = require('fs');
+const path = require('path');
+
+test('snapshot files are independent of test resolution order', async t => {
+	const options = {
+		cwd: exec.cwd('intertest-order')
+	};
+	// Run, updating snapshots.
+	await exec.fixture(['test.js', '-u'], options);
```
sindresorhus, line 11:
```suggestion
	await exec.fixture(['test.js', '--update-snapshots'], options);
```

Same in other places.

sindresorhus (changes requested):

novemberborn:
This looks really promising @ninevra! Let me know how I can help!

> Should this be placed behind a feature flag, to reduce the incidence of spurious updates?

No, I think this is fine. As long as existing snapshot files are accepted we're good.

ninevra:
The remaining CI failures are also present on master when running on node v15.3.0. This appears to be unrelated to this PR, and may not actually a bug; reported in #2616.

I'd still like to add more tests to this, ideally asserting that existing snapshots aren't rewritten.

test/snapshot-order/backwards-compatibility.js
```diff
@@ -0,0 +1,119 @@
...
+		fs.unlinkSync(snapshotPath);
+		fs.unlinkSync(reportPath);
+	});
+
+	const snapshotPath = path.join(options.cwd, 'test.js.snap');
+	const reportPath = path.join(options.cwd, 'test.js.md');
+
+	// Install a known-unsorted snapshot, report
+	fs.copyFileSync(unsortedSnapshotPath, snapshotPath);
+	fs.copyFileSync(unsortedReportPath, reportPath);
```
novemberborn, line 32: Write the buffers from memory, instead of copying the files? Makes the comparison below clearer I think.

test/snapshot-order/randomness.js
```diff
@@ -0,0 +1,31 @@
...
+	const snapshotPath = path.join(options.cwd, 'test.js.snap');
+	const reportPath = path.join(options.cwd, 'test.js.md');
+
+	// Run test
+	await exec.fixture(['--update-snapshots'], options);
+
+	// Assert snapshot is unchanged
+	const snapshot = fs.readFileSync(snapshotPath);
+
+	t.snapshot(snapshot);
```
novemberborn, line 24: Let's add a message so it's a little clearer what is in this snapshot. Because if we change anything else in the snapshot format, this snapshot needs updating too.

```suggestion
	t.snapshot(snapshot, 'resulting snapshot in binary encoding');
```

novemberborn (commented):
> I'd still like to add more tests to this, ideally asserting that existing snapshots aren't rewritten.

The test coverage is impressively thorough already! What do you mean by not rewriting existing snapshots? FWIW if the files are deterministic I'm not overly concerned about writing unnecessarily. Not sure how to test that either other than by looking at modification dates maybe… doesn't seem worthwhile.

I left two minor comments on the tests. Nice use of the self-testing setup.



> The remaining CI failures are also present on master when running on node v15.3.0. This appears to be unrelated to this PR, and may not actually a bug; reported in #2616.

Thanks, fixing in https://github.com/avajs/ava/pull/2618.

novemberborn:
@ninevra if you pull in latest `master` then tests should pass.

It does contain some linter related changes so there may be a bit of churn.

ninevra:
@novemberborn 
> I'd still like to add more tests to this, ideally asserting that existing snapshots aren't rewritten.

I was referring to the `backwards-compatibility` test suite I later added in 7dcb928, which demonstrates that existing unsorted snapshots won't be overwritten until AVA is run with `--update-snapshots` or the tests are modified. Mostly that was just to convince myself I wasn't reading the code wrong and missing something.

The test coverage is complete now as far as I'm concerned.

(Although, the `backwards-compatibility` suite is not of any real value after this point, and is difficult to maintain; if the snapshot file format is changed in the future, obtaining valid & unsorted snapshots will be difficult, so the suite should probably be removed at that point. We could remove it preemptively now to reduce maintainence burden.)

test/snapshot-order/backwards-compatibility.js
```diff
@@ -0,0 +1,119 @@
...
+
+	// Schedule snapshot cleanup
+	t.teardown(() => {
+		fs.unlinkSync(snapshotPath);
+		fs.unlinkSync(reportPath);
+	});
+
+	// Install a known-unsorted snapshot, report
+	fs.copyFileSync(unsortedSnapshotPath, snapshotPath);
+	fs.copyFileSync(unsortedReportPath, reportPath);
```
novemberborn, line 64:
```suggestion
	fs.writeFileSync(snapshotPath, unsortedSnapshot);
	fs.writeFileSync(reportPath, unsortedReport);
```

test/snapshot-order/backwards-compatibility.js
```diff
@@ -0,0 +1,119 @@
...
+		fs.unlinkSync(snapshotPath);
+		fs.unlinkSync(reportPath);
+	});
+
+	const snapshotPath = path.join(options.cwd, 'test.js.snap');
+	const reportPath = path.join(options.cwd, 'test.js.md');
+
+	// Install a known-unsorted snapshot, report
+	fs.copyFileSync(unsortedSnapshotPath, snapshotPath);
+	fs.copyFileSync(unsortedReportPath, reportPath);
```
novemberborn, line 102:
```suggestion
	fs.writeFileSync(snapshotPath, unsortedSnapshot);
	fs.writeFileSync(reportPath, unsortedReport);
```

novemberborn (commented):
> Although, the `backwards-compatibility` suite is not of any real value after this point, and is difficult to maintain; if the snapshot file format is changed in the future, obtaining valid & unsorted snapshots will be difficult, so the suite should probably be removed at that point. We could remove it preemptively now to reduce maintainence burden.

Yes let's do that.

There's a few more places where we could write the in-memory buffers. Other than that I think this is good to go!

If you wouldn't mind touching this up I'll try and get it out in a release soon.

ninevra:
Removed the `backwards-compatibility` test suite per https://github.com/avajs/ava/pull/2610#pullrequestreview-541540485. This incidentally removes the remaining instances of `copyFileSync` (sorry for missing those in the first place). Hopefully I'm not misunderstanding your request?

novemberborn:
Awesome, thanks @ninevra!

novemberborn:
3.14 is live with these changes.

fisker:
Wow, maybe I can remove this hack https://github.com/sindresorhus/eslint-plugin-unicorn/pull/842 👍

PR #2619, opened 2020-12-03
Upgrade supertap and print `t.log` entries according to TAP diagnostics spec

### LINKED ISSUE #2614: `t.log` entries should start with `#`(without any spaces)
Jenkins TAP reporter does not parse report correctly when `t.log` statements are present there.
E.g. tap report from ava
```
not ok 3 - test2.ts › Tomato block is not present 2
  * Log statement 1
  * Log statement 2
  * Log statement 3
  * Log statement 4
  * Log statement 5
  * Log statement 6
  ---
    name: AssertionError
    message: Tomato image still displayed
    assertion: 'true'
    values:
      'Value is not `true`:': 'false'
    at: |-
      e2e/spec/test2.ts:14:9
      initMacro (e2e/macros/initMacro.ts:55:3)
  ...
```

And the result of report is
Screenshot 2020-11-25 at 19 45 29

So I created script the replaced `  *` with `#` and it parses it correctly
Script (just in case)
```bash
sed -i 's/  \*/  #/' report.tap
```
Report 
Screenshot 2020-11-25 at 19 49 02
Screenshot 2020-11-25 at 19 49 59

According to [specification](https://testanything.org/tap-version-13-specification.html) it should be diagnostic lines and start with `#`
```text
Diagnostics
Additional information may be put into the testing output on separate lines. 
Diagnostic lines should begin with a #, which the harness must ignore, at least as far as analyzing the test results.
```
```
Anything else
Any output that is not a version, a plan, a test line, a YAML block, 
a diagnostic or a bail out is incorrect. How a harness handles the incorrect line is undefined. 
Test::Harness silently ignores incorrect lines...
```

So, of course, there is a problem with Jenkins plugin, but that's another story. Ava should just stick to specification and stream diagnostic info lines starting with `#`.

tymfear:
Fixes #2614 
@novemberborn , so I forked `supertap` and added it as git dependency so that you could take a look and we could decide on what to do next.

Check out [report.txt](https://github.com/avajs/ava/files/5635525/report.txt) for results.

Also here is result from Jenkins TAP plugin 
tapresult


novemberborn:
> I forked `supertap` and added it as git dependency so that you could take a look and we could decide on what to do next.

@tymfear I've got access to `supertap` now. Could you open it as a PR?

lib/reporters/tap.js
```diff
@@ -125,7 +125,7 @@ class TapReporter {
 		this.reportStream.write(`# ${stripAnsi(title)}${os.EOL}`);
 		if (evt.logs) {
 			for (const log of evt.logs) {
-				const logLines = indentString(log, 4).replace(/^ {4}/, '  # ');
+				const logLines = indentString(log, 4).replace(/^ {4}/gm, '#   ');
```
novemberborn, line 128: Do we need extra indentation?

tymfear: Well it was there before, and makes output more readable. Also the same indentation is in `supertap`

tymfear:
> > I forked `supertap` and added it as git dependency so that you could take a look and we could decide on what to do next.
> 
> @tymfear I've got access to `supertap` now. Could you open it as a PR?

@novemberborn please - https://github.com/vadimdemedes/supertap/pull/4

novemberborn:
@tymfear I've landed your changes to `supertap` and installed it here, as well as fixing merge conflicts. I've rebuild the reporter logs, do they look correct to you? Could you test this branch one last time? Thanks!

tymfear:
@novemberborn  - you mean on Jenkins? Will do!

tymfear:
@novemberborn, tested on Jenkins - looks correct!

novemberborn:
> One thing is - I don't have npm `6.13.4`, so cannot fix `package-lock.json`

No worries — I had to update it anyway to depend on supertap properly.

novemberborn:
Thank you @tymfear. I'm having a look at what else is close to landing and then I'll cut a release.

PR #2630, opened 2020-12-24
Fix timeout reporting in TAP reporter

### LINKED ISSUE #2612: TAP output does not say which tests (or at least which file) failed after timeout
So here is the output of timedout test report (there are passed tests as well).
```
# Exited because no new tests completed within the last 180000ms of inactivity
not ok 51 - Exited because no new tests completed within the last 180000ms of inactivity
```
It is impossible to figure out which test caused timeout. 
Would be nice at least to narrow it down to test file, to understand where to start searching.

tymfear:
Fixes #2612

### DIFF
--- a/lib/reporters/tap.js
+++ b/lib/reporters/tap.js
@@ -131,6 +131,16 @@ class TapReporter {
 		}
 	}
 
+	writeTimeout(evt) {
+		const err = new Error(`Exited because no new tests completed within the last ${evt.period}ms of inactivity`);
+
+		for (const [testFile, tests] of evt.pendingTests) {
+			for (const title of tests) {
+				this.writeTest({testFile, title, err}, {passed: false, todo: false, skip: false});
+			}
+		}
+	}
+
 	consumeStateChange(evt) { // eslint-disable-line complexity
 		const fileStats = this.stats && evt.testFile ? this.stats.byFile.get(evt.testFile) : null;
 
@@ -172,7 +182,7 @@ class TapReporter {
 				this.writeTest(evt, {passed: true, todo: false, skip: false});
 				break;
 			case 'timeout':
-				this.writeCrash(evt, `Exited because no new tests completed within the last ${evt.period}ms of inactivity`);
+				this.writeTimeout(evt);
 				break;
 			case 'uncaught-exception':
 				this.writeCrash(evt);


tymfear:
Example of output
```
not ok 2 - test23.ts › Exited because no new tests completed within the last 9000ms of inactivity
not ok 3 - dir › test.ts › Exited because no new tests completed within the last 9000ms of inactivity
```

novemberborn:
I'm struggling to understand why this needs a change in `api.js`. I had assumed the TAP reporter was lagging behind the default reporter, but this implies it's now doing something the default reporter cannot? In other words, why aren't the existing status events sufficient?

tymfear:
@novemberborn yeah, you're right. Have found the the `pendingTests` set available for event. Will make an update.

tymfear:
@novemberborn, so how about new approach - report each pending test as failed when timeout occurs?

lib/reporters/tap.js
```diff
@@ -131,6 +131,16 @@ class TapReporter {
 		}
 	}
 
+	writeTimeout(evt) {
+		const err = new Error(`Exited because no new tests completed within the last ${evt.period}ms of inactivity`);
+
+		for (const [testFile, tests] of evt.pendingTests) {
+			[...tests].forEach(title => {
+				this.writeTest({testFile, title, err}, {passed: false, todo: false, skip: false});
+			});
```
novemberborn, line 140:
```suggestion
			for (const title of tests) {
				this.writeTest({testFile, title, err}, {passed: false, todo: false, skip: false});
			}
```

lib/reporters/tap.js
```diff
@@ -209,4 +219,5 @@ class TapReporter {
 		}
 	}
 }
+
```
novemberborn, line 222:
```suggestion
```

novemberborn (commented):
@tymfear yes that's much better, thank you. What does this look like in the output?

I can't recall if we have timeout integration tests for the other reporters. If we do, could you add the same for the TAP reporter?

lib/reporters/tap.js
tymfear: Done

tymfear:
@novemberborn  - I've updated my comment with output. Probably should've rather add a new one, sorry.
I couldn't find any reporter tests with timeouts.

novemberborn (approved):

novemberborn:
Thanks @tymfear and happy new year 🎆

PR #2631, opened 2020-12-24
Clarify that assertions do not throw. Fixes #2624

### LINKED ISSUE #2624: document that ava assertions don't throw
I didn't realize until recently that ava assertions don't stop the test when they fail.

I think this could be documented better, and I'm willing to do so if that's useful

f5lee:


docs/03-assertions.md
```diff
@@ -163,53 +174,69 @@ test('custom assertion', t => {
...
+do ***not*** throw on failure.
+
 ### `.pass(message?)`
 
 Passing assertion.
 
 ### `.fail(message?)`
 
-Failing assertion.
+Failing assertion. Does not throw.
```
novemberborn, line 186: I'd rather remove this. This describes how the specific assertion behaves, not sure it aids readability to repeat that the assertion never throws.

docs/03-assertions.md
```diff
@@ -10,7 +10,15 @@ test('unicorns are truthy', t => {
...
 ```
 
-Assertions are bound to their test so you can assign them to a variable or pass them around:
+**Note**: Unlike other assertion libraries, AVA assertions do ***not***
+stop the test when they fail.
+
+When a test has multiple assertion failures,
+AVA only displays the *first* failure.
+There's currently no way to see all failures from a test
+([see issue #261](https://github.com/avajs/ava/issues/261)).
```
novemberborn, line 19: This should come at the end of the section, as it was before. The section starts with how you can access the assertions and how they're bound. An explanation of their runtime behavior shouldn't be sandwiched in between.

As for the text itself, what do you think about this?

```suggestion
AVA's assertions do not throw errors. Your test will continue executing after an assertion failure. AVA only displays the *first* failure.
```

docs/03-assertions.md
```diff
@@ -163,53 +174,69 @@ test('custom assertion', t => {
 
 ## Built-in assertions
 
+**Note**: Unlike other assertion libraries, these assertions
+do ***not*** throw on failure.
```
novemberborn, line 178: Good call on repeating it here. With my suggested copy that would become:

```suggestion
AVA's assertions do not throw errors. Your test will continue executing after an assertion failure.
```

novemberborn (commented):
Thanks for the PR @f5lee!

I've left some inline comments. Also would you mind reducing the amount of whitespace changes? We don't hard-wrap Markdown files in this project.

novemberborn:

Issue #2500, opened 2020-05-25
Improved reporter experience


novemberborn:
By default it provides minimal output. We show a spinner and the status of the last test, so you get a sense of activity, and once an entire test file completes do we print failures. In verbose mode, we'd print the results of all tests & hooks, including logs. Various counts for passing and failing tests are always shown at the end of the output.

If a TTY is not available, we wouldn't print a spinner and only write the appropriate output for a test file once it completes.

We'd use [`ink`](https://www.npmjs.com/package/ink) to construct the output.

novemberborn:
This is sort-of blocked by the consolidation of our current reporters, see https://github.com/avajs/ava/issues/2217.

stavalfi:
# Feature Request

### Background

I'm migrating a Jest project to Ava and I find it hard to search the failed tests in Ava reporter.

With jest, you can search in the terminal "fail " (with a space at the end) and you will find all the failed tests files which failed. (it won't help you to find the actual tests which failed).

With Ava, I need to search for: "✖" symbol to find the actual tests which failed (but it won't help you to find the failed test files).

-----------

### Problem

In a project with 20+- tests it can be ok. but in my project I have 150+ tests and each one has a lot of logging (with `t.log`). so the test output is too long to find what I'm looking for.

----------

### Feature Request

At the end of the in the default reporter, list the following:

1. a list of all failed test files.
2. for each failed test-file, list all the failed tests titles.
3. 
---------

### Possible workaround

working ___locally___ with custom tap reporters is annoying because I need to pipe every command with them every time.

novemberborn:
@stavalfi I think that matches with what I'm hoping we can get to. Essentially the final output is test files and the results of the tests therein, with any realtime stuff happening just at the bottom of the terminal. So you could follow the output as AVA is running.

stavalfi:
@novemberborn Thanks for your ongoing effort on Ava. You guys do amazing job. 

If that's ok from you, I would like to talk about prioritising this ticket over others.

---------

I'm considering to use Ava in my personal and company projects because of 2 reasons: `t.log` and better tests concurrency.

Ava has a killer feature: `t.log`. this feature does not exist in any other test-runner in node (As I'm aware of).
It's such a great feature that when you have tons of logs per test, you can finally do something when the CI/CD fails because the logs are organised per test-title. As to other test runners, all the logs are interleaved and un readable. as a result, some developers will try to reproduce the failed test locally but sometimes fail to do so (due to different environments and so on...)

[...]

PR #2647, opened 2021-01-15
Include failed test files and titles in default reporter

stavalfi:
Related to https://github.com/avajs/ava/issues/2500#issuecomment-758878957

@novemberborn I want to hear your opinions before fixing the tests.

##### Before:

image

##### After:

image

----

The only way to search the failed tests before this PR is to search the "✖" symbol and search for "Uncaught exception".

----

#### TODO:

@novemberborn When all of the tests of a worker passed but the worker exit with code 1, I can't find a way to get the failed-tests in the default reporter:

```
test('test1', async () => {
  process.exitCode = 1
})
```

In this situation, I would also expect that `--serial --fail-fast` won't run tests after `test1` but that's not the case. Maybe Ava needs to evaluate the process.exitCode before moving on to the next test.

wdyt?

### DIFF
--- a/lib/reporters/default.js
+++ b/lib/reporters/default.js
@@ -915,6 +915,35 @@ class Reporter {
 		if (this.watching) {
 			this.lineWriter.writeLine();
 		}
+
+		const failedFiles = [...this.stats.byFile.keys()].filter(failedFile => this.failures.some(failure => failure.testFile === failedFile));
+
+		const failedTestInFile = failedFile => this.failures.filter(failure => failure.testFile === failedFile).map(failure => failure.title);
+
+		if (failedFiles.length > 0) {
+			this.lineWriter.writeLine();
+		}
+
+		for (const failedFile of failedFiles) {
+			this.lineWriter.writeLine(colors.errorSource(this.relativeFile(failedFile)));
+			for (const failedTestTitle of failedTestInFile(failedFile)) {
+				this.lineWriter.writeLine(colors.error(`${figures.cross} ${failedTestTitle}`));
+			}
+		}
+
+		const uncaughtExceptionsTestFiles = this.uncaughtExceptions.map(failedFile => failedFile.testFile);
+
+		if (failedFiles.length > 0 || uncaughtExceptionsTestFiles.length > 0) {
+			this.lineWriter.writeLine();
+		}
+
+		for (const failedFile of uncaughtExceptionsTestFiles) {
+			this.lineWriter.writeLine(colors.title(`Uncaught exception in ${this.relativeFile(failedFile)}`));
+		}
+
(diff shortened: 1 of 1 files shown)


novemberborn:
Apologies I've lost track of what this is solving. Surely we are reprinting the failed tests already? What am I missing?

(And yes my confusion is a sure sign this code needs to be replaced.)

There should be some message for the worker crashing already I would think.

With --fail-fast, new tests shouldn't be run, though everything already in motion will continue running.

stavalfi:
@novemberborn Thanks for the great feedback.

>  Surely we are reprinting the failed tests already? What am I missing?

As you can see in the screenshots, what I have printed is not printed. 

> reprinting the failed tests

Ava does prints the failed tests __logs__ again. in case there are tons of logs for each test, it won't help that much to get an instant summary of what has failed.

Also, at least in a specific scenario, the failed tests are not printed at all:

image

I don't remember the reason behind this specific screenshot.

-----

I have tested the scenario where:
```
test('test1', async () => {
  process.exitCode = 1
})
```
The results are the following:
1. Tests are keep running (even with `--serial --fail-fast`).
2. With `--serial`, there is no clear summary of what test-title failed.
3. With or without any extra flag, there is no clear summary of what test-file failed.

---------------------------------------

I just want to mention that if you are confused about something, we can be sure that I'm missing something.

Maybe you can point me out to the right direction?

Thanks again!

novemberborn:
> I have tested the scenario where:
> 
> ```
> test('test1', async () => {
>   process.exitCode = 1
> })
> ```
> 
> The results are the following:
> 
> 1. Tests are keep running (even with `--serial --fail-fast`).
> 2. With `--serial`, there is no clear summary of what test-title failed.
> 3. With or without any extra flag, there is no clear summary of what test-file failed.
> 
> NOTE: This PR doesn't solve this. I'm not sure if it should be on this PR or not.

Let's discuss that separately.

---



> Ava does prints the failed tests **logs** again. in case there are tons of logs for each test, it won't help that much to get an instant summary of what has failed.

I was thinking of this code:

https://github.com/avajs/ava/blob/32c5425353cc58422f0a196bb0c06a4095a33825/lib/reporters/default.js#L731

But then there's other errors that are not printed in verbose mode… perhaps because it would be repetitive, perhaps because of forgotten reasons:

https://github.com/avajs/ava/blob/32c5425353cc58422f0a196bb0c06a4095a33825/lib/reporters/default.js#L749



> I just want to mention that if you are confused about something, we can be sure that I'm missing something.

Hehe, I appreciate that, but the reporting code is quite confusing. It's grown based on feedback and is no longer based on clear intent.

novemberborn:

PR #2657, opened 2021-01-29
Fix link to Discussions in issue template

jesse-savary:
As far as I can tell the link to https://github.com/avajs/ava/discussions was never showing up because of the missing `about` key.


novemberborn:
Indeed! Thanks!

jesse-savary:
Of course! Ava ended up on https://github.com/devspace/awesome-github-templates#bomb-templates-for-issues and I both discovered the issue and learned a bunch in the process. Have a good one.

Issue #2661, opened 2021-02-03
Snapshot ids can collide with test names


ninevra:
## What you're trying to do

I'm just testing edge cases in hopes of better understanding the code. This isn't likely to be important to anyone's use case.

## What happened

If a snapshot has an id that matches a test name, it compares against the first snapshot from that test.

## What you expected to happen

This should probably either:
- Throw an error, like duplicate test titles do.
- Report an error by some other means.
- Namespace snapshot ids separately from test names. That would probably require changes to the snapshot binary format.

The existing behavior _could_ be used intentionally, but I can't think of a good reason to do so.

## Sample code

```javascript
const test = require('ava');

test('a name', t => {
  t.snapshot('one', 'a snapshot with no id');
  t.snapshot('two', {id: 'a name'});
});
```

This produces:

```
  a name

  test.js:5

   4:   t.snapshot('one', 'a snapshot with no id');
   5:   t.snapshot('two', {id: 'a name'});         
   6: });                                          

  Did not match snapshot

  Difference:

  - 'two'
  + 'one'

  › test.js:5:5

  ─

  1 test failed
```

The snapshot report looks like: 

```markdown
# Snapshot report for `test.js`

The actual snapshot is saved in `test.js.snap`.

Generated by [AVA](https://avajs.dev).

## a name

> a snapshot with no id

    'one'
```

## Configuration

AVA v3.15.0.

No AVA configuration.

Invoking AVA as `npx ava` or `npx ava --update-snapshots`.

novemberborn:
Perhaps we should remove this feature?

novemberborn:
https://github.com/avajs/ava/issues/2669

PR #2663, opened 2021-02-03
Don't increment nextSnapshotIndex when skipping a snapshot with an id

### LINKED ISSUE #2662: Skipping snapshots with ids can cause other snapshots to fail or write spurious snapshots
## What you're trying to do

Skip a snapshot assertion that has an id.

(I'm still just reading/working on the code and testing edge cases. This one could affect someone in actual usage, though.)

## What happened

Normally, snapshots only increment `nextSnapshotIndex` if they don't have an `id`. Skipped snapshots, however, always increment `nextSnapshotIndex`. If a snapshot with an `id` is skipped, it effectively pushes down any subsequent snapshot assertions, causing them to compare against the expected value of the next snapshot in order. If there was a next snapshot, it will probably fail; if not, it will record a new snapshot that doesn't actually exist.

## What you expected to happen

Skipping an assertion should have no effect on the behavior of other assertions.

## Sample code

```javascript
const test = require('ava');

test('some snapshots', t => {
  t.snapshot(null, {id: 'an id'}); // <--
  t.snapshot('one');
  t.snapshot('two');
});
```

Run once, then change the first line of the test to `t.snapshot.skip(null, {id: 'an id'});` and run again. The second assertion will fail and the third will record a "Snapshot 3".

[MRE](https://github.com/ninevra/ava-snapshot-skip-id)

## Configuration

- AVA v3.15.0
- Node.js v14.15.4
- no AVA configuration
- invoking AVA as `npx ava`.

ninevra:
Fixes #2662 by checking whether the snapshot being skipped has an id.

I'm filing this as a draft because I might clean up the code a bit tomorrow.


novemberborn:
As per my comment https://github.com/avajs/ava/issues/2661#issuecomment-772378420 I'm inclined to remove this feature in the next major release (spring time, but we could start doing betas sooner). I don't think we need to fix the bug before then. What do you think?

novemberborn:
Let's leave this as a known issue and remove the feature altogether. https://github.com/avajs/ava/issues/2669

PR #2664, opened 2021-02-03
Schedule previously failing test files to run first

### LINKED ISSUE #1546: Schedule previously failing tests to run first
[Issuehunt badges](https://issuehunt.io/r/avajs/ava/issues/1546)

This would be useful when you're working on some failing tests and want to quickly check if you fixed them. You could argue that watch mode solves this, but sometimes you're just fixing a quick thing and don't bother starting watch mode.

AVA will have to persist the titles of failing tests between test runs. I think this is ok, as in the future, we'll want to, for example, save more metadata between test runs, like the time for each test file, so we can smartly schedule what files to run in parallel for optimal performance.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $80.00)

- [issuehunt issuehunt](https://issuehunt.io/u/issuehunt) ($80.00)

### Submitted pull Requests
- [#2694 Schedule previously failing tests to run first](https://issuehunt.io/r/avajs/ava/pull/2694)
---

### Tips

</details>

bunysae:
Fixes #1546. Test files with failing tests are stored in a tmp-directory and scheduled first. I decided to go with test-files to keep the current invocation order (serial tests first, concurrent ones afterwards).


lib/runner.js
```diff
@@ -400,7 +400,8 @@ class Runner extends Emittery {
 					err: serializeError('Test failure', true, result.error, this.file),
 					duration: result.duration,
 					knownFailing: result.metadata.failing,
-					logs: result.logs
+					logs: result.logs,
+					testFile: this.file
```
novemberborn, line 404: Why is this needed? IIRC the code that handles these changes knows what file they come from.

lib/scheduler.js
```diff
@@ -0,0 +1,55 @@
+const fs = require('fs');
+const path = require('path');
+const tempDir = require('temp-dir');
+
+const buildPath = projectDir => path.join(tempDir, 'ava', path.basename(projectDir));
```
novemberborn, line 5: IIRC there is a "cache dir" concept in `api.js`. Perhaps this should use that same directory?

lib/scheduler.js
```diff
@@ -0,0 +1,55 @@
...
+const FILE_NAME_FAILING_TEST = 'failing-test.json';
+
+module.exports.prepareTestResultsFile = projectDir => {
+	const temporaryResultsDir = buildPath(projectDir);
+	try {
+		fs.statSync(temporaryResultsDir);
+	} catch {
+		fs.mkdirSync(temporaryResultsDir, {recursive: true});
```
novemberborn, line 13: IIRC this doesn't fail if the directory exists, so the `statSync` shouldn't be necessary.

lib/scheduler.js
```diff
@@ -0,0 +1,55 @@
...
+
+module.exports.deleteFile = filePath => {
+	try {
+		fs.unlinkSync(filePath);
+	} catch {
+	}
+};
+
+module.exports.store = (filePath, data) => {
+	fs.writeFileSync(filePath, data);
```
novemberborn, line 27: Should this use `write-file-atomic`, like we do in the snapshot manager?

lib/scheduler.js
```diff
@@ -0,0 +1,55 @@
...
+};
+
+// Order test-files, so that files with failing tests come first
+module.exports.failingTestsFirst = (projectDir, selectedFiles) => {
+	const filePath = path.join(buildPath(projectDir), FILE_NAME_FAILING_TEST);
+	let failedTestFiles = [];
```
novemberborn, line 33: No need to initialize since you return in the `catch`.

```suggestion
	let failedTestFiles;
```

lib/scheduler.js
```diff
@@ -0,0 +1,55 @@
...
+	} catch {
+	}
+};
+
+module.exports.store = (filePath, data) => {
+	fs.writeFileSync(filePath, data);
+};
+
+// Order test-files, so that files with failing tests come first
+module.exports.failingTestsFirst = (projectDir, selectedFiles) => {
```
novemberborn, line 31: It'd be better if this would return a new, sorted array, rather than modifying the array that is passed in.

lib/scheduler.js
```diff
@@ -0,0 +1,55 @@
...
+	const filePath = path.join(buildPath(projectDir), FILE_NAME_FAILING_TEST);
+	let failedTestFiles = [];
+	try {
+		failedTestFiles = JSON.parse(fs.readFileSync(filePath));
+	} catch {
+		return;
+	}
+
+	selectedFiles.sort((f, s) => {
+		if (failedTestFiles.find(tf => tf === f) && failedTestFiles.find(tf => tf === s)) {
```
novemberborn, line 41: `includes`?

```suggestion
		if (failedTestFiles.includes(f) && failedTestFiles.includes(s)) {
```

test-tap/helper/report.js
```diff
@@ -102,6 +113,7 @@ const run = (type, reporter, {match = [], filter} = {}) => {
 
 	options.globs = normalizeGlobs({extensions: options.extensions, files: ['*'], providers: []});
 
+	prepareScheduler(projectDir);
```
novemberborn, line 116: Why is this needed?

test/scheduler/test.js
```diff
@@ -0,0 +1,17 @@
...
+		await exec.fixture(['1pass.js', '2fail.js']);
+	} catch { }
+
+	try {
+		await exec.fixture(['--concurrency=1', '1pass.js', '2fail.js']);
+	} catch (error) {
+		t.snapshot(error.stdout);
+	}
+
+	delete process.env.AVA_FORCE_CI;
```
novemberborn, line 16: No need, this runs in its own process.

lib/api.js
```diff
@@ -142,6 +144,10 @@ class Api extends Emittery {
 				runStatus = new RunStatus(selectedFiles.length, null);
 			}
 
+			if (!isCi) {
+				Scheduler.failingTestsFirst(this.options.projectDir, selectedFiles);
```
novemberborn, line 148: I'm thinking `Scheduler` should check for CI.

lib/api.js
```diff
@@ -168,7 +174,7 @@ class Api extends Emittery {
 			}
 
 			runStatus.on('stateChange', record => {
-				if (record.testFile && !timedOutWorkerFiles.has(record.testFile)) {
+				if (record.testFile && !timedOutWorkerFiles.has(record.testFile) && record.type !== 'test-failed') {
```
novemberborn, line 177: Why this additional condition?

lib/api.js
```diff
@@ -183,6 +189,10 @@ class Api extends Emittery {
 						worker.notifyOfPeerFailure();
 					}
 				}
+
+				if (record.type === 'test-failed' && !failedTestFiles.find(file => file === record.testFile)) {
+					failedTestFiles.push(record.testFile);
```
novemberborn, line 194: Make `failedTestFiles` a `Set` instead, so you can just add?

lib/api.js
```diff
@@ -271,6 +283,21 @@ class Api extends Emittery {
...
+
+		const filePath = Scheduler.prepareTestResultsFile(this.options.projectDir);
+		if (files.length === 0) {
+			Scheduler.deleteFile(filePath);
+			return;
+		}
+
+		const jsonFiles = JSON.stringify(files);
+		Scheduler.store(filePath, jsonFiles);
+	}
```
novemberborn, line 300: This logic should be in `Scheduler`.

novemberborn (commented):
Thanks for picking this up @bunysae! I've left some comments below.

Also note there's a bunch of changes coming with https://github.com/avajs/ava/pull/2672 that I'm hoping to land soon.

novemberborn:
@bunysae the diff isn't coming out right. Looks like it includes commits from `main`, perhaps due to a rebase?

bunysae:
Yes i merged the latest changes to check if it influences the scheduler-function.

novemberborn:
@bunysae fair enough, but if you look at the PR diff it contains changes already in `main`, so it's now pretty hard to review.

bunysae:
No problem. I will open another one and copy my changes manually.

bunysae:
Created #2694.

PR #2668, opened 2021-02-06
Update endpoint testing with Mongoose recipe

BrijeshBumrela:
`getConnectionString` is deprecated in newer versions, use `getUri` instead


novemberborn:
Thanks @BrijeshBumrela1

PR #2679, opened 2021-02-08
Remove support for snapshot IDs

### LINKED ISSUE #2669: Remove `id` support in `t.snapshot()`
A presumably little known feature of the `t.snapshot()` assertion is that you can provide an ID for the snapshot. I doubt it's much used and it doesn't really fit with the mental model of of snapshots belonging to tests.

Let's remove it for AVA 4.

ninevra:
Closes #2669.

Removes handling for snapshot ids. Updates documentation and typescript definitions to match. Raises a (hopefully) helpful assertion failure message when encountering a snapshot id.

[`// TODO: In a breaking change, reject non-undefined, falsy IDs and messages`](https://github.com/ninevra/ava/blob/586c0ec7fae5650bb1e08e8e243d3bf0b9179c89/lib/test.js#L228) is mostly moot now; i'd implement the rest, but I'm not quite sure what the original intent was. The only non-undefined, falsy message presently accepted is `''`; should this be rejected for all assertions, or just for `t.snapshot()`?


docs/03-assertions.md
```diff
@@ -322,9 +322,8 @@ Assert that `contents` matches `regex`.
 Assert that `contents` does not match `regex`.
 
 ### `.snapshot(expected, message?)`
-### `.snapshot(expected, options?, message?)`
 
-Compares the `expected` value with a previously recorded snapshot. Snapshots are stored for each test, so ensure you give your tests unique titles. Alternatively pass an `options` object to select a specific snapshot, for instance `{id: 'my snapshot'}`.
+Compares the `expected` value with a previously recorded snapshot. Snapshots are stored for each test, so ensure you give your tests unique titles.
```
novemberborn, line 326: I'd like to keep the documentation usable for AVA 3 users:

```suggestion
Compares the `expected` value with a previously recorded snapshot. Snapshots are stored for each test, so ensure you give your tests unique titles.

AVA 3 supports an  `options` object that lets you select a specific snapshot, for instance `{id: 'my snapshot'}`. This is buggy and will be removed in AVA 4.
```

lib/assert.js
```diff
@@ -708,17 +708,14 @@ class Assertions {
...
-				const [optionsOrMessage] = rest;
-				if (typeof optionsOrMessage === 'object') {
-					snapshotOptions = optionsOrMessage;
-				} else {
-					message = optionsOrMessage;
-				}
+			if (message && message.id !== undefined) {
+				fail(new AssertionError({
+					assertion: 'snapshot',
+					message: 'Snapshot ids were removed in AVA 4',
```
novemberborn, line 714:
```suggestion
					message: 'AVA 4 no longer supports snapshot IDs',
```

lib/test.js
```diff
@@ -222,13 +222,13 @@ class Test {
 
 		const deferRecording = this.metadata.inline;
 		this.deferredSnapshotRecordings = [];
-		this.compareWithSnapshot = ({expected, id, message}) => {
+		this.compareWithSnapshot = ({expected, message}) => {
 			this.snapshotCount++;
 
 			// TODO: In a breaking change, reject non-undefined, falsy IDs and messages.
```
novemberborn, line 228: Yes, empty messages should result in an improper use failure. Just for this assertion, I don't think it matters much for the others. The intent should be that the message is used in the snapshot report, and passing an empty string for that is nonsensical.

test-d/snapshot.ts
```diff
@@ -0,0 +1,18 @@
+import {expectError} from 'tsd';
+import test from '..';
+
+test('snapshot', t => {
+	t.snapshot({foo: 'bar'});
+	t.snapshot(null, 'a snapshot with a message');
+	expectError(t.snapshot('hello world', null));
+	// See https://github.com/avajs/ava/issues/2669
+	expectError(t.snapshot({foo: 'bar'}, {id: 'an id'}));
```
novemberborn, line 9: I appreciate adding these tests for the type definition. I don't think we need to test that a removed feature has been removed, however.

ninevra:
I'm worrying about a possible class of issues similar to #2662. Currently, if the assertion message is invalid, `nextSnapshotIndex` is not incremented, because none of the `Test`'s callbacks are ever invoked. (This behavior predates this PR, but previously affected a narrower set of messages.) This can't affect test pass/fail state, because the test will already have failed, but it can affect the pass/fail state of subsequent snapshot assertions (which I think is technically exposed by `(await t.try(/* ... */)).errors`), and potentially the recorded snapshots (in edge cases where multiple snapshots are being added/updated at once).

@novemberborn What do you think? Is this an issue?

novemberborn:
No I think that's an edge case.

Arguably we could have a different channel for communicating improper use errors, but I don't think that's terribly important to solve.

novemberborn (approved):

novemberborn:
Thank you @ninevra 😄

Issue #1666, opened 2018-01-28
Support arguments in "require" configuration


novemberborn:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%2440%20Funded-%2300A156.svg" />](https://issuehunt.io/r/avajs/ava/issues/1666)

Prompted by #1665 and inspired by Babel's plugin / preset configuration,  it'd be neat to do this:

```json
"ava": {
  "require": [
    ["@babel/register", {
      "ignore": ["test/*"]
    }]
  ]
}
```

If an array is provided we'd require the module, resolve any `__esModules` / `default` shenanigans, and invoke `.apply()` with the remaining values. We should support any number of arguments.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $40.00)

- [<img src='https://avatars3.githubusercontent.com/u/44827199?v=4' alt='issuehunt' width=24 height=24> issuehunt](https://issuehunt.io/u/issuehunt) ($40.00)

### Submitted pull Requests
- [#3184 feat: handle complex `options.require` config property](https://issuehunt.io/r/avajs/ava/pull/3184)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/avajs/ava/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

Jolo510:
Hey @novemberborn, I'd like to take on this task!

novemberborn:
@Jolo510 awesome, go for it!

amslv:
Hey, @novemberborn, I'd like to work on this task

novemberborn:
@amslv awesome, thanks!

Since this issue was opened, we added support for `ava.config.js` files. We need to make sure the arguments can be stringified as JSON and then parsed in the worker processes.

We could do the stringifying and the parsing in the main process first, and then use our "deep equal" logic to make sure the values are the same: https://github.com/avajs/ava/blob/6d12abfdff4478a1b6f4e87237764b89eb05f18d/lib/assert.js#L282

IssueHuntBot:
@issuehunt has funded $40.00 to this issue.

---
- Submit pull request via [IssueHunt](https://issuehunt.io/repos/26820798/issues/1666) to receive this reward.
- Want to contribute? Chip in to this issue via [IssueHunt](https://issuehunt.io/repos/26820798/issues/1666).
[...]

PR #2680, opened 2021-02-10
Support arguments in "require" configuration Issue#1666

yliu481:
## What
This fixes the handling of argument in require configuration   #1666 
I'm not sure if I'm doing the right thing, so could you please kindly give me some advice.

## How
- Modify the `resolveModules` function in [api.js](https://github.com/avajs/ava/blob/main/lib/api.js), so it can load array in `options.require`.
- Modify the `load` function in [subprocess.js](https://github.com/avajs/ava/blob/main/lib/worker/subprocess.js), so it can load modules with arguments. 

## Question
For the `load` function, do I need to make 
https://github.com/avajs/ava/blob/41e9473a0160bdd7dc90e33e1e3e04a9885d77df/lib/worker/subprocess.js#L189
and
https://github.com/avajs/ava/blob/41e9473a0160bdd7dc90e33e1e3e04a9885d77df/lib/worker/subprocess.js#L207
also support an array as input?


novemberborn:
Thanks for this @yliu481.

I'm thinking `resolveModules()` should return an array of `{modulePath: string, args: string[]}` (if you'll excuse the TypeScript notation). That way, in `subprocess`, you don't have to special-case the behavior.

All arguments should be applied, not just the first one.

The `resolveModules()` implementation should be less repetitive.

You don't need to change how `extensionsToLoadAsModules` or `providers` are handled.

yliu481:
Great, thanks for the reply, I'm working on it.

Question 1
Is it a good idea to let `resolveModules()` return an array of `{modulePath:string, args:object}` , which can be convenient when passing the args to `requireFn` in `subprocess`.
The implementation will look like 
``` nodejs
return requireFn(ref.modulePath)(ref.args);
```

Question 2
also if `resolveModules()` returns `object[]`, this line
https://github.com/avajs/ava/blob/6cb0944e5250bcff24846a260c748948171a0f08/lib/worker/subprocess.js#L244
will be a special case for `load()`, cause `testPath` is just a string, could you please tell me how am I supposed to deal with this.

novemberborn:
Ah! I see where the confusion lies. Sorry I didn't spot this before.

`load()` must remain unmodified.

Instead, it's these lines that need changing:

https://github.com/avajs/ava/blob/2011b92b638df42b4d050a19e9b37cbb19659970/lib/worker/subprocess.js#L194-L195

(I only just committed this on the `main` branch so you'll have to merge that into your PR.)

`options.require` is the result of `resolveModules()`.

`load()` should be used to load the module. Inspect the return value. If it is a function, call it with the arguments. If it's an object (and not `null`) with a `default` property that is a function, then call that function with the arguments. This makes it compatible with CommonJS and ES modules.

Remember to spread the arguments: `fn(...args)`.

This invocation is a change in behavior we should document. Before we didn't invoke any exported function.

yliu481:
Thanks for your comment.

I updated `resolveModule()`, it returns an array of `{modulePath:string, args:object}`,  which can be convenient when call the returned `fn` with `args`. 

Please let me know if there is anything that I did wrong. Thanks for your time and patience.

lib/api.js
```diff
@@ -21,13 +21,13 @@ const sharedWorkers = require('./plugin-support/shared-workers');
 
 function resolveModules(modules) {
 	return arrify(modules).map(name => {
-		const modulePath = resolveCwd.silent(name);
-
+		let [modulePath, args] = Array.isArray(name) ? name : [name, {}];
```
novemberborn, line 24: `args` should be an array, not an object. It might be clearer to set the default during destructuring, so even `['path/to/module']` works:

```suggestion
		let [modulePath, args = []] = typeof name === 'string' ? [name] : name;
```

lib/worker/subprocess.js
```diff
@@ -192,7 +192,14 @@ ipc.options.then(async options => {
 
 	try {
 		for await (const ref of (options.require || [])) {
-			await load(ref);
+			const fn = await load(ref.modulePath);
+			if (fn) {
+				if (fn instanceof Function) {
```
novemberborn, line 197:
```suggestion
				if (typeof fn === 'function') {
```

lib/worker/subprocess.js
```diff
@@ -192,7 +192,14 @@ ipc.options.then(async options => {
...
+					fn(ref.args);
```
novemberborn, line 198:
```suggestion
					fn(...ref.args);
```

lib/worker/subprocess.js
```diff
@@ -192,7 +192,14 @@ ipc.options.then(async options => {
...
+				} else if (typeof fn === 'object' && fn.default && fn.default instanceof Function) {
+					fn.default(ref.args);
```
novemberborn, line 200:
```suggestion
				} else if (fn && typeof fn.default === 'function') {
					fn.default(...ref.args);
```

novemberborn (commented):
Sorry for the delay in getting back to you. I've left some comments.

Have you thought about how you'd like to test this? I think ideally we add a new set of tests in https://github.com/avajs/ava/tree/main/test. Perhaps the required module (being a fixture) could declare a test, which can then be snapshotted:

```js
const test = require('ava')

module.exports = (...args) => {
  test('receives args', t => {
    t.log(args)
    t.pass()
  })
}
```

See also https://github.com/avajs/ava/blob/b6a3cb723cefe34f816f958ede7b78a1fc796b2a/test/shared-workers/workers-are-loaded-once/test.js#L6.

lib/api.js
yliu481: Thanks! I appreciate your time and correction, but this is exactly where I'm confused. 

**Why I think it's an object**
From https://github.com/avajs/ava/issues/1666#issue-292221666, I am assuming the `config` looks like this 
```
"ava": {
  "require": [
    ["@babel/register", {
      "ignore": ["test/*"]
    }]
  ]
}
```
so when destructing `require`
```javascript
let [modulePath, args] = typeof name === 'string' ? [name] : name;
```
`modulePath` is `"@babel/register"`
`args` is `{
      "ignore": ["test/*"]
    }`, which is an object.

Please correct me if I am wrong.

novemberborn: That happens to be the argument that module is expecting. Looking at this again I think we should do this:

```suggestion
		let [modulePath, ...args] = typeof name === 'string' ? [name] : name;
```

Note the `...` in front of `args`. If we're not passed a string, we assume an array. The first item must be the module path and any remaining arguments will be arguments provided when calling that module's default export.

yliu481: Oh, That makes sense, thank you, I will commit these suggestions and then work on the test part.

novemberborn:

PR #2681, opened 2021-02-12
Add Babel and TypeScript provider packages as optional peer dependencies

kiancross:
`@ava/typescript` is an optional peer dependency - this should be indicated in `package.json`, otherwise stricter package manages (such as Yarn PnP) throw errors:

> ✖ ava tried to access @ava/typescript, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -98,9 +98,13 @@
 				"node": ">=12.20 <13 || >=14.15 <15 || >=15"
 			},
 			"peerDependencies": {
+				"@ava/babel": "*",
 				"@ava/typescript": "*"
 			},
 			"peerDependenciesMeta": {
+				"@ava/babel": {
+					"optional": true
+				},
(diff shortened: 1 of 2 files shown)


novemberborn:
Why are these installers complaining about what the code does at runtime?

kiancross:
> Why are these installers complaining about what the code does at runtime?

[Yarn PnP](https://yarnpkg.com/features/pnp) also resolves packages at runtime.

novemberborn:
Such folly. But fine I guess it is an optional peer dependency after all. Could you add `@ava/babel` as well?

We’re now working towards AVA 4, so this may not be available for a while.

kiancross:
> Could you add @ava/babel as well?

Done.

> We’re now working towards AVA 4, so this may not be available for a while.

Thanks for the heads-up.

Issue #2099, opened 2019-04-20
Consider not updating snapshots automatically


novemberborn:
It may be considered odd that AVA automatically updates snapshots. From @billyjanitsch https://github.com/avajs/ava/issues/1585#issuecomment-379589219:

>  I've never liked the behavior of auto-creating new snapshots without -u: I imagine most users think of running ava as a pure function which returns a boolean describing whether their tests passed or failed. So it's odd for ava to have side effects by default, especially those which modify VCS-tracked files.

Then from @novemberborn https://github.com/avajs/ava/issues/1585#issuecomment-379672636:

> what if we'd write snapshot updates to a temporary location, exit with an instructive failure and exit code of 1, and then have a npx ava --commit-snapshots command (pending bike-shedding) to write them into [the] source tree?

ahukkanen:
I am currently writing a modification to a library that uses AVA for the tests and I found the current behavior very confusing before I bumped into the original issue (#1585) comments. There is also no mention about this behavior in the documentation:
https://github.com/avajs/ava/blob/master/docs/04-snapshot-testing.md

This change would be highly welcomed for newcomers. This is my first time ever using AVA.

sramam:
Given my extensive commenting on #1585, I'd like to argue against this. 

`ava`'s current behavior is consistent with other js-land test frameworks. 
While I'm not against doing different, it should not be done without sufficient cause. 

Many of the issues raised by  @billyjanitsch, while real, belong outside a test runner. 

> Pragmatically, it complicates scenarios like running a test suite as a commit/push/prepublish hook. I've often successfully pushed commits (relying on my git hook to fail if the tests do anything other than succeed) only to realize that the test suite passed but dirtied the repo, so my (incorrect) commit shouldn't have gone through.

IMHO the pre-commit/push/publish hooks should include a step to check for uncommitted modifications. That's not to deny @billyjanitsch's concern that it complicates things, it does. 

It'll either require `ava` to include isomorphic-git or have an implied dependency on a git executable being available. Either of these dependencies seems too onerous a price. 

I think @novemberborn's instinct on calling this bike-shedding is exactly right.

ahukkanen:
Either way, document how it works so that also people not familiar with AVA will understand how it works. Or at least when they bump into this behavior, they get the information from the documentation that it is how it is designed to work.

novemberborn:
[...]

Issue #1768, opened 2018-04-12
Exposing snapshot manager logic as a library


jamiebuilds:
## Description

I'm trying to build some tools on top of Ava's snapshots, but there's a lot of work to read/update them correctly.

I was wondering if Ava could expose some methods for working with snapshots outside of the testing process.

I need two things:

```ts
type Snapshot = {
  filePath: string,
  name: string,
  value: string,
  ...
};

export function getAllSnapshots(): Promise<Array<Snapshot>>;
export function updateSnapshot(snapshot: Snapshot, newValue: any): Promise<void>;
```

Trying to build a tool that finds all of the React related snapshots and renders stringified HTML into a browser so the visual difference can be compared. If there is no visual difference it can update them all automatically.

novemberborn:
The `.snap.md` Markdown files that AVA generates are write-only. We could support an API for parsing the binary `.snap` files. One limitation is that we currently store fixed-width hashes of test titles, so you can't get the titles back out. Tests may have multiple snapshots as well.

Values are [`concordance`](https://github.com/concordancejs/concordance) serializations,  not strings. Deserializing requires you to provide the `concordance` plugins that were used in the serialization. That's not a problem at the moment because we don't let you customize plugins, but it will be in the future. I'm not clear how you'd render these deserialized values in the browser for visual diffing though.

The deserializing-using-plugins problem is surmountable, being one of configuration. The snapshot name extraction problem requires a change to the `.snap` format. We could either add a reverse lookup or store the test titles directly, though that makes parsing more difficult.

In light of #1769, do you see this as an entirely separate process or as a complement to an AVA run? In which case we could surface the failing snapshots, you could render in the browser, then rerun a specific test and let AVA update the snapshot. Then we wouldn't need to change the `.snap` format either.

jamiebuilds:
> The snapshot name extraction problem requires a change to the .snap format. We could either add a reverse lookup or store the test titles directly, though that makes parsing more difficult.

It would be really useful to have the name back, otherwise snapshots can very easily get mixed up.

> do you see this as an entirely separate process or as a complement to an AVA run? In which case we could surface the failing snapshots, you could render in the browser, then rerun a specific test and let AVA update the snapshot.

I want it to be able to run in two modes:

- Take a diff (in like a branch or something) and display all the changed snapshots
- As Ava is running, display snapshot mismatches
  - live updating as tests re-run
  - with buttons next to the visual diffs to update them

[...]

PR #2685, opened 2021-02-14
Introducing AVA snapshots v3

### LINKED ISSUE #2634: "Could not update snapshots" message could be clearer
## What you're trying to do: 

When AVA is passed `--update-snapshots` and a test runner encounters a `skip()`ed test (or, with #2623, a `skip()`ed assertion) the reporter prints  "Could not update snapshots for the following test files: \<list of filenames\>".

This message doesn't say _why_ snapshots couldn't be updated. In practice the reason is likely to be divinable (the user just recently added a `.skip()`), but it could be a lot clearer.

Further, the message can appear for test files that never use snapshots and test files where all snaphots match. It doesn't mean much of anything in these cases, and could mislead the user into thinking something went wrong.

## How AVA could handle this

Indicating a reason: The `Runner` should know what problem prevented updating. It would have to pass this reason back with its `{cannotSave: true}` result from `Runner#saveSnapshotState()`, and then this would be propagated along though the `'snapshot-error'` event and handled by reporters.

Suppressing the message when not relevant: checking whether snapshots existed should be simple enough. checking whether snapshots changed is trickier. The snapshot manager would have to be queried about whether any snapshots failed to match, which isn't currently something it knows when run in `updating` mode, since it ignores the `.snap` file entirely in favor of regenerating it.

This issue would benefit from reworking the snapshot manager to check for and load snapshots in every run, instead of only when it's called for and only when not `updating`.

### LINKED ISSUE #2635: Allow updating snapshots for individual tests or individual assertions
## What you're trying to do

Example use case:
- a developer writes tests using `t.snapshot()` assertions
- a regression or external change causes such a test to begin failing
- the developer marks the test `.failing()`, expecting the issue to remain for some time
- later, the developer makes changes to other tests in the same file, and runs `--update-snapshots` to record the changes
- this incidentally updates the snapshots for the `.failing()` test, causing it to begin passing, even though the underlying issue remains
- developer is now stuck: they have no way to update snapshots for just _part_ of a file (not even manually, because AVA's snapshot format is an implementation detail)

## Why you can't use AVA for this

AVA appears to provide no way to select individual tests or assertions for update. Selecting files can be accomplished by passing AVA a glob, but using `--match` or line number selection with `--update-snapshots` throws.

Fundamentally, this is because AVA implements `--update-snapshots` by entirely ignoring and overwriting the snapshot files. If AVA allowed users to select specific tests, then every other test in their files would have its snapshots deleted.

## And maybe how you think AVA could handle this

From a ui perspective, AVA could handle this by providing an interactive `--review-snapshots` option, like the similar features in [`insta`](https://docs.rs/insta/1.3.0/insta/#snapshot-updating) (implemented as a cli tool, `cargo insta review`) or [`jest`](https://jestjs.io/docs/en/snapshot-testing#interactive-snapshot-mode) (seemingly only available during watch mode).

AVA could also allow the use of `--match` and other existing forms of selection with `--update-snapshots`, but this would only allow per-test granularity, not per-snapshot.

From an implementation perspective, AVA would probably need to read snapshot files from disk even when run with `--update-snapshots`, so that the unselected tests/assertions could retain their preexisting snapshots.

It _might_ be possible to implement this using AVA's currently-unused ability to append updates to `.snap` files:

https://github.com/avajs/ava/blob/3fe4c40f09fd8eef74ea998992a756525fca7b9f/lib/snapshot-manager.js#L184-L186 

https://github.com/avajs/ava/blob/3fe4c40f09fd8eef74ea998992a756525fca7b9f/lib/snapshot-manager.js#L298-L299

but I _think_ that would require (backwards-compatible!) changes to the snapshot file format.

ninevra:
Implements a new snapshot file format containing all information necessary to generate the snapshot report file.

Allows the use of `test.skip()`, `test.only()`, `--match`, line number selection, and `t.snapshot.skip()` during `--update-snapshots` runs.

Closes #2634 by removing the "Could not update snapshots" case entirely.
Closes #2635 by allowing existing test & assertion selection mechanisms to be used with `--update-snapshots`.
Hopefully will assist with #1768 and #2099.

## Current proposed snapshot file format:

- Readable prefix: `AVA Snapshot v3\n`
- Snapshot version (UInt16LE): 3
- sha256 checksum of the following
- gzip compressed data:
    - CBOR encoded data:
        ```
        {
          blocks: [
            {
              title: "A test title",
              snapshots: [
                {
                  label: "A snapshot label",
                  buffer: /* Serialized concordance descriptor */
                },
                ...
              ]
            },
            ...
          ]
        }
        ```

Notes:
- The `label` field is absent unless the user supplied a label. This allows changing the default snapshot label if desired in the future.
- The `buffer` field may be absent, if the snapshot has only ever been skipped. Such snapshots are currently rendered as `<No Data>`.
- Currently, some data still exists only in the report file, namely the filenames of the test file and the snapshot file. This information can be derived from the name of the snapshot file, but theoretically that could change in the future.

### Pros of the proposed format:
- The snapshot report can now be regenerated from the snapshot file.
    - Enables the use of test selection mechanisms during `--update-snapshots`
    - Allows snapshot reports to be kept always in order, rather than the previous approach of appending changes at the end.
    - Prerequisite for any feature that edits snapshots without overwriting them entirely.
- AVA now maintains less byte-manipulation code
- The snapshot format is easier (though not trivial, because of the gzipping and the header) for third-parties to read
- Adding new properties either to snapshots, to blocks, or to the file overall should be easy and remain readable by older AVAs (though older AVAs may or may not preserve such unrecognized data when updating the .snap file)

### Cons of the proposed format:
- AVA can't totally guarantee that the snapshot file format is deterministic; some leeway is given to the CBOR encoder. .snap files might show spurious changes e.g. between minor versions of the CBOR encoder.
- AVA's attack surface is changed. Attackers remain able to submit a malicious .snap file in a pull request, in hope of exploting potential bugs in AVA to exfiltrate repository secrets from CI runs or gain control of developers' machines. The total amount of code exposed to such an attack is significantly increased, and unless we carefully filter the result of CBOR decoding (not currently done), such an attacker can produce unexpected javascript types and values. On the other hand, the binary-parsing code now has rather more eyes on it.

### Frontend:
- Rather than append new tests to the old report, adding new tests without `--update-snaphots` now maintains declaration-order.
- When the snapshot report is changed, tests not touched in that run (e.g. removed tests, tests which no longer use snapshots) are moved to the end of the report, maintaining their previous order among themselves.

### Edge cases:
- If `--update-snapshots` is passed and the only use of snapshots is in discarded `t.try()` attempts, then the snapshot file is removed. There's no snapshot data to record in this case, so I don't see any reason to keep the file.
- If `t.snapshot.skip(); t.snapshot();` is added at the end of a test, `RangeError` is no longer thrown; instead a "blank space" is recorded for the skipped snapshot. This is currently rendered in the snapshot report as `<No Data>`.
- `snapshot-manager` now assumes `Array.prototype.sort` is stable. This is the true in all supported Node versions afaik, and is required by the standard as of ecmascript 2019, but notably is false in node v10. If `Array.prototype.sort` is not stable, the snapshot report order may not be deterministic.

### Backend:
- The snapshot file is eagerly loaded in the `Runner` constructor, rather than lazily loaded.
     - `Runner` always emits a dependency on its `.snap` file, even in `--update-snapshots` mode and even if no snapshot assertions are performed.
     - In some very narrow edge cases where the snapshot file has a valid version and checksum but unreadable data, this could result in throwing an unhelpful error. AVA@3 only threw such errors if snapshots were actually used.
- Introduces a dependency on `cbor`.
- Removes a dependency on `md5-hex`.

## TODO:
- [x] Implement a placeholder snapshot format containing the necessary information
- [x] Test that snapshot reports can be regenerated from `.snap` files 
- [x] Allow running `--update-snapshots` with `test.skip()`, `t.snapshot.skip()`, `test.only()`, `--match`, line number selection
    - [x] Finish testing this
    - [x] Document this
- [x] Refactor & clean up code
    - [x] Ordering data should be collected as-it-happens, like `skipBlock()` and `skipSnapshot()`, not once at load time 
    - [x] Now that the snapshot manager has all the information on skipping tests and snapshots, `cleanSnapshots()` fits more naturally as part of `Manager#save()`. But,
    - [x] The above requires that the snapshot manager exist when not used, which conflicts with `Runner`'s lazy-loading behavior, which in turn is relied on by existing `Runner` unit-test suites.
    - [x] Remove code for handling `cannotSave` events
- [x] Debug the performance issues somehow. Are they caused by this PR or preexisting? Are they specific to CI, specific to temp dirs, or neither?
    - Tests were just spawning too many processes in CI. Fixed.
- [x] Finalize the snapshot file format


ninevra:
Looks like the testing approach I've been using here and in `snapshot-removal` is very prone to causing timeouts in CI. Tests are taking upwards of half a minute to run in CI vs 3-4s locally. Further, since each suite contains a large number of concurrent tests in a single file, they're all kicked off at once, increasing the chance that no test finishes within AVA's global timeout period.

Most of these tests are creating a temporary directory, copying a fixture, running AVA once to set up some initial state, running AVA again to modify that state, checking that the change matches the expectation, and then cleaning up the temporary directory. (The windows runs are somehow timing out _within the first fixture execution_, which is very strange.) I'll have to rethink this & hopefully eliminate some of those steps and/or reorganize the tests to run in a timeout-friendly pattern.

ninevra:
Not using temporary directories cuts most of the runtimes back down to 2-5s. I don't think it's ideal to have tests modifying their fixtures in-place, but it'll do for now.

ninevra:
Very little of the test time appears to be spent setting up the fixtures; the same AVA invocations may be running much more slowly when in a temporary directory (which would at least explain how the windows tests were timing out during fixture execution).

novemberborn:
@ninevra this is coming along nicely! Let me know if there's specific areas you'd like me to look at.

ninevra:
@novemberborn this is nearly ready for review, the only remaining issue is tests that use temporary fixtures timing out in CI. I've got a setup now on my fork where I can ssh into a CI machine and mess around `strace`ing things, but this is pretty far outside my experience, so, can't guarantee I'll find anything.

One other question: Is it a problem for `Runner` to always declare a dependency on its `.snap` file? It's _correct_ (in the current draft, `Runner`  tries to read the `.snap` file in every run), but I'm not familiar with the precise effects of declaring a dependency.

novemberborn:
That timeout is annoying! Sorry, not sure how that would have come about.

Dependencies are used for the watcher. In theory it would know to rerun the tests if the corresponding snapshot file is changed outside of AVA, e.g. by a `git checkout`.

ninevra:
Well, I spent an embarrassingly long time looking for some kind of deadlock/bottleneck/other subtle problem with this PR's code, but it looks like the test suites are just overwhelming the CI machines in a completely normal fashion. `snapshot-removal`'s test suite is particularly bad because it launches something like 34 copies of AVA, 17 at a time.

I can fix this by breaking up the tests among more files, causing AVA to limit concurrency appropriately. Then the `snapshot-workflow` tests can go back to using temporary directories. I should be able to get the refactoring done over the weekend.

However, this PR's changes do still slow down the `snapshot-removal` tests relative to `main` somewhat. I hypothesize that this is because of the preloading of `.snap` files; the `snapshot-removal` tests previously did not read `.snap` files, and now do. If so, there should be no slowdown in the common case, since most of the time if a `.snap` file exists, it'll get used. Might still be worth benchmarking the new and old `decodeSnapshots` implementations, though?

ninevra:
Limiting concurrency succeeds in making tests pass in CI, but it looks like Windows really can't handle more than 1 test at a time, so probably I should just use `test.serial` here (or maybe set concurrency based on `os.cpus()`?).

ninevra:
The ideal here would be to hook into AVA's global maximum concurrency, so that the total number of tests running across all test files was always at the processor count. Unfortunately I don't know of a good way to do that other than putting each test in its own file, which entails a bunch of boilerplate and makes it tedious to review or edit the tests.

ninevra:
Refactored the `snapshot-removal` and `snapshot-workflow` tests to use pre-generated templates and to run in serial. All tests now take less than 3 seconds even on Windows. Slightly more maintainance will be required to keep the templates up to date, but I've tried to streamline & document it adequately.

This should now be ready for review.

novemberborn:
> The ideal here would be to hook into AVA's global maximum concurrency, so that the total number of tests running across all test files was always at the processor count.

You could do that with shared workers actually.

novemberborn:
> * md5 checksum of the following

Shall we update this to SHA-256? Just to deal with projects that rule out insecure hashes throughout the codebase.

novemberborn:
> * Adding new properties either to snapshots, to blocks, or to the file overall should be easy and remain readable by older AVAs (though older AVAs may or may not preserve such unrecognized data when updating the .snap file)

Our only concern should be being able to read snapshots written by previous minor versions (within the same major release).

A test using a new API won't pass when used with an older version either.

novemberborn:
> * AVA can't totally guarantee that the snapshot file format is deterministic; some leeway is given to the CBOR encoder. .snap files might show spurious changes e.g. between minor versions of the CBOR encoder.

The CBOR spec itself makes a big deal about determinism, so hopefully this will be rare.

novemberborn:
> The snapshot file is eagerly loaded in the `Runner` constructor, rather than lazily loaded.

Why does it need to be loaded eagerly? Could we load it synchronously when needed?

novemberborn:
> > The snapshot file is eagerly loaded in the `Runner` constructor, rather than lazily loaded.
> 
> Why does it need to be loaded eagerly? Could we load it synchronously when needed?

Although, of course, if there is no snapshot file this is a cheap operation, and if there is, it's likely needed. So this is fine, never mind 😄

docs/04-snapshot-testing.md
```diff
@@ -44,6 +44,8 @@ You can then check your code. If the change was intentional you can use the `--u
 $ ava --update-snapshots
 ```
 
+If you need to update snapshots for only a particular test, you can use `--update-snapshots` together with e.g. `--match` or `.only()` to select the test.
```
novemberborn, line 47:
```suggestion
In AVA 4, if you need to update snapshots for only a particular test, you can use `--update-snapshots` together with e.g. `--match` or `.only()` to select the test.
```

novemberborn (commented):
> * When the snapshot report is changed, tests not touched in that run (e.g. removed tests, tests which no longer use snapshots) are moved to the end of the report, maintaining their previous order among themselves.

What happens when tests are skipped? Are those sections moved to the end as well? We should have access to the declaration order.

novemberborn (commented):
I'm going to need some more time to review this. Impressive work @ninevra.

ninevra:
> > * When the snapshot report is changed, tests not touched in that run (e.g. removed tests, tests which no longer use snapshots) are moved to the end of the report, maintaining their previous order among themselves.
>
> What happens when tests are skipped? Are those sections moved to the end as well? We should have access to the declaration order.

Tests that are skipped should remain ordered properly.

Currently, ordering info is collected for tests that are skipped (or "not selected" by other means) or that use `t.snapshot()` or `t.snapshot.skip()` (if used only in a discarded `t.try()`, I think things get a little inconsistent; I need to look at that case in more detail). Ordering info is not currently collected for tests that are run but don't use snapshots, nor for `.todo()` tests, nor for hooks.

It's possible to collect ordering info for all tests that are declared, whether they use snapshots or not; I'd add a `Manager.prototype.touch(title, taskIndex)` method and call it every time a test is declared. It's of course not possible to collect ordering info for tests that have been removed entirely.

> > > The snapshot file is eagerly loaded in the `Runner` constructor, rather than lazily loaded.
> >
> >
> > Why does it need to be loaded eagerly? Could we load it synchronously when needed?
>
> Although, of course, if there is no snapshot file this is a cheap operation, and if there is, it's likely needed. So this is fine, never mind smile

`Manager` now has a substantially larger interface to `Runner` (`compare` and `save` are joined by `skipSnapshot` and `skipBlock`), some of which needs to be called _before_ the first snapshot comparison. I initially tried to keep lazyloading, which is certainly possible, but it requires basically shimming the entire `Manager` interface and replaying calls to `skipSnapshot` and `skipBlock` when the real `Manager` is finally loaded. The duplicated interface was really annoying to iterate on and debug, so I scrapped it.

Plus, in theory eager loading should make the performance characteristics of snapshot-based tests a little more consistent; before, the first `t.snapshot()` to run would perform sync file io and then parse a potentially-large file, while later assertions would be fast.

> > * AVA can't totally guarantee that the snapshot file format is deterministic; some leeway is given to the CBOR encoder. .snap files might show spurious changes e.g. between minor versions of the CBOR encoder.
>
> The CBOR spec itself makes a big deal about determinism, so hopefully this will be rare.

I expect it should be rare, yes. I'm invoking `cbor.encode` with `{omitUndefinedProperties: true, canonical: true}`, which should help. Unfortunately the `cbor` package doesn't yet appear to support "Deterministic CBOR", but if i recall correctly the main difference between that and "Canonical CBOR" is the choice of sort order for map keys, so it shouldn't really make a difference.

> > * md5 checksum of the following
>
> Shall we update this to SHA-256? Just to deal with projects that rule out insecure hashes throughout the codebase.

Sounds good to me. I'll hopefully have time to make changes in a day or two.

> > * Adding new properties either to snapshots, to blocks, or to the file overall should be easy and remain readable by older AVAs (though older AVAs may or may not preserve such unrecognized data when updating the .snap file)
>
> Our only concern should be being able to read snapshots written by previous minor versions (within the same major release).
>
> A test using a new API won't pass when used with an older version either.

Makes sense.

Should we consider including a minor and/or patch version in the `.snap` file? Could add it to the readable prefix without issue, keep the existing version field as-is so that AVA <= 3 keep reporting the right error, and add a field `version: {major: 3, minor: 0, patch: 0}` (or `version: [3, 0, 0]` or `version: "3.0.0"`) to the top level of the data structure. Then AVA@4.0 could issue a warning (probably not a hard error??) if it encountered a minor version > 0.

> > The ideal here would be to hook into AVA's global maximum concurrency, so that the total number of tests running across all test files was always at the processor count.
>
> You could do that with shared workers actually.

A shared worker could only control parallelism across the test files where it was loaded, though, right? That is, if it was loaded in `snapshot-removal` and `snapshot-workflow`, it could limit their total concurrency to `os.cpus().length`, but it wouldn't know if AVA was running tests from other suites at the same time? (Shared workers are really cool, but I'm not familiar with the interface yet, sorry.)

novemberborn:
> It's possible to collect ordering info for all tests that are declared, whether they use snapshots or not; I'd add a `Manager.prototype.touch(title, taskIndex)` method and call it every time a test is declared.

What scenario would this cover? Adding a new snapshot assertion to an existing test which is run selectively? Though that would still order correctly right?



> Should we consider including a minor and/or patch version in the `.snap` file? Could add it to the readable prefix without issue, keep the existing version field as-is so that AVA <= 3 keep reporting the right error, and add a field `version: {major: 3, minor: 0, patch: 0}` (or `version: [3, 0, 0]` or `version: "3.0.0"`) to the top level of the data structure. Then AVA@4.0 could issue a warning (probably not a hard error??) if it encountered a minor version > 0.

I don't think the warnings are worthwhile. Either it's a hard error, and then a single major version suffices, or handling in newer versions deals with missing fields. Which is needed anyway to ensure CI passes when a newer (minor / patch) AVA version is installed and snapshots haven't been updated.



> A shared worker could only control parallelism across the test files where it was loaded, though, right? That is, if it was loaded in `snapshot-removal` and `snapshot-workflow`, it could limit their total concurrency to `os.cpus().length`, but it wouldn't know if AVA was running tests from other suites at the same time? (Shared workers are really cool, but I'm not familiar with the interface yet, sorry.)

You're right, yes. Controlling overall execution could be an interesting concept, but possibly something that AVA should provide a framework for but not provide out of the box.

ninevra:
Switched to sha256 hashes and implemented the `touch()` method (the latter can be reverted easily if you think it's a detriment).

> What scenario would this cover? Adding a new snapshot assertion to an existing test which is run selectively? Though that would still order correctly right?

It covers some narrow edge cases. For instance, if a test had its snapshot assertions removed, it's snapshot data would be retained but, since it wasn't skipped and no longer used snapshots, so the `Manager` would never learn its order. This woudn't matter until a new snapshot was added to the test file, at which point the report would get regenerated and the first test shunted to the end.

Collecting ordering info for every test also just seems a little easier to reason about.

novemberborn:
@ninevra I've pushed some small tweaks, let me know what you think. In the snapshot format I renamed `buffer` to `data`, but I think everything else is the same.

Your tests seem very thorough. There's a lot of it though, so it's hard to review! Any tips? Or else let's just ship it 😄

ninevra:
@novemberborn all the tweaks look great to me! Especially the `formatEntry()` code, it looks a lot nicer now.

> Your tests seem very thorough. There's a lot of it though, so it's hard to review! Any tips? Or else let's just ship it :smile:

I mean, I don't think we should ship code you're not confident in, and I certainly don't intend to pressure you to do that. I am sorry that the diff ended up so massive.

What can I do to make the code and/or tests easier to review/maintain?

Ideas:

- I've pushed a change to the `snapshot-workflow` tests that snapshots a before/after diff of the report, rather than just the final report; hopefully that'll make it easier to see what the tests are doing. Feel free to revert it if it's just making things worse.
- I could consolidate all the `snapshot-workflow` tests in one file to make them less tedious to look through.
- I could rewrite some of the test fixtures to use descriptive messages and titles instead of just `'foo'` et al placeholders.

ninevra:
I have a crude-but-working implementation of the shared worker concurrency limiter we discussed; it's on [ninevra/ava/snapshot-regenerate-shared-worker-semaphore](https://github.com/ninevra/ava/tree/snapshot-regenerate-shared-worker-semaphore). I think landing that here would require updating [avajs/test](https://github.com/avajs/test) to expose `plugin.js`, though. Without that the plugin can't `require('@ava/test/plugin')`.

Would you be interested in a PR to add semaphores to `@ava/cooperate`, or a separate plugin? The interface I wrote could use some more work, but eventually it'd be nice to have.

novemberborn:
> Would you be interested in a PR to add semaphores to `@ava/cooperate`, or a separate plugin? The interface I wrote could use some more work, but eventually it'd be nice to have.

That'd be perfect for `@ava/cooperate`!

> I think landing that here would require updating [avajs/test](https://github.com/avajs/test) to expose `plugin.js`, though. Without that the plugin can't `require('@ava/test/plugin')`.

That'd be fine, except that shared workers are still experimental and subject to breaking changes even in patch releases of AVA. I'm not sure we should be relying on that in our own test suite shipping AVA would break its tests.

novemberborn:
> * I've pushed a change to the `snapshot-workflow` tests that snapshots a before/after diff of the report, rather than just the final report; hopefully that'll make it easier to see what the tests are doing.

It's very meta, but also very helpful. Thanks!

I think I got side tracked by all the fixtures. Once I skipped those it got a lot easier. Great work.

novemberborn:
🎊

PR #2694, opened 2021-02-20
Schedule previously failing tests to run first

bunysae:
Successor for PR #2664.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#1546: Schedule previously failing tests to run first](https://issuehunt.io/repos/26820798/issues/1546)
---
</details>


lib/api.js
```diff
@@ -131,7 +132,7 @@ class Api extends Emittery {
 				const {currentIndex, totalRuns} = this.options.parallelRuns;
 				const fileCount = selectedFiles.length;
 
-				// The files must be in the same order across all runs, so sort them.
+				// Initial sort order
```
novemberborn, line 135: I prefer the original comment.

lib/api.js
```diff
@@ -142,6 +143,8 @@ class Api extends Emittery {
 				runStatus = new RunStatus(selectedFiles.length, null);
 			}
 
+			selectedFiles = scheduler.failingTestsFirst(selectedFiles, this._createCacheDir(), this.options.cacheEnabled);
```
novemberborn, line 146: We should only create the cache dir when it's needed. Perhaps we need to separate a "get predictable cache dir path" which returns `null` when caching is disabled, and "create the cache dir for this invocation of AVA" which uses the same path, predictable or temporary. What do you think?

lib/run-status.js
```diff
@@ -194,6 +194,18 @@ class RunStatus extends Emittery {
...
+	getFailedTestFiles() {
+		const result = [];
+		const iterator = this.stats.byFile.keys();
+		for (let {value, done} = iterator.next(); !done; {value, done} = iterator.next()) {
+			if (this.stats.byFile.get(value).failedTests) {
+				result.push(value);
+			}
+		}
+
+		return result;
```
novemberborn, line 207: You can use a `for` / `of` loop when handling iterators. However this can be dealt with much more succinctly:

```suggestion
		return [...this.stats.byFile.values()].filter({ failedTests }) => failedTests)
```

novemberborn (commented):
I've left some comments inline.

Have you had a chance to look at the test failures on Windows?

lib/api.js
bunysae: For the moment ava the local cache inside the folder `node_modules/cache/ava` is only used for this function. When we want to go for cache dir in temporary directory, we create a hash of the project-dir (like sha-256 ...) and create the directory from the hash. This would be predictable.

test/scheduler/test.js
```diff
@@ -0,0 +1,48 @@
+const test = require('@ava/test');
+const exec = require('../helpers/exec');
```
novemberborn, line 2: This should give you access to the internal status objects for each test that executed, hopefully ordered. So you shouldn't need to handle stdout. Apologies for not spotting this earlier.

novemberborn (commented):
@bunysae @sindresorhus what do you think of storing the state file in the user's home directory? We could index it based on the hashed project directory. Or would `node_modules/.cache/ava` be better?

sindresorhus:
> what do you think of storing the state file in the user's home directory? We could index it based on the hashed project directory. Or would node_modules/.cache/ava be better?

I would store it in `node_modules/.cache/ava`. From experience, storing stuff in the home directory is error-prone with a high probability of encountering file permission problems.

novemberborn:
@sindresorhus do you think the cache settings should impact this feature?

sindresorhus:
If you mean the `cache` option, then yes.

novemberborn:
@bunysae this is great, thank you. Would you be able to look at the alternative test approach I commented on? I appreciate that the self-hosted tests are rather unapproachable. I can make the changes myself too if you'd prefer, just let me know.

bunysae:
@novemberborn
I will try to create some tap-tests for `api.js`:-)

novemberborn:
> I will try to create some tap-tests for `api.js`:-)

@bunysae Ideally we don't add new tests there, but instead under the `test/` directory. Those are run by AVA itself.

bunysae:
@novemberborn 
The tests are refactored. `stdout` isn't used anymore.

novemberborn:
@bunysae great, thank you!

I've pushed a few tweaks. Will merge once CI passes.

PR #2696, opened 2021-02-22
Return booleans indicating whether assertions passed

jmarkham828:
Apologies for dropping this. I just finished university! 🥳 

Hoping to still contribute to this awesome project. I picked up where [this](https://github.com/avajs/ava/pull/2586/files) left off. I totally understand if this PR is closed, given that I left it hanging for a while, but wanted to give it another go at resolving #2455, since it doesn't look like #2586 is active any longer.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -493,7 +493,7 @@ export interface CbExecutionContext<Context = unknown> extends ExecutionContext<
 	end(error?: any): void;
 }
 
-export type ImplementationResult = PromiseLike<void> | Subscribable | void;
+export type ImplementationResult = PromiseLike<void> | Subscribable | boolean | void;
 export type Implementation<Context = unknown> = (t: ExecutionContext<Context>) => ImplementationResult;
 export type CbImplementation<Context = unknown> = (t: CbExecutionContext<Context>) => ImplementationResult;
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -34,23 +34,19 @@ export type CommitDiscardOptions = {
 
 export interface Assertions {
 	/**
-	 * Assert that `actual` is
-	 * [truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy),
-	 * returning Boolean indicating whether the assertion passes. Comes with
-	 * power-assert.
+	 * Assert that `actual` is [truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy), returning a boolean
+	 * indicating whether the assertion passed. Comes with power-assert.
 	 */
 	assert: AssertAssertion;
 
 	/**
-	 * Assert that `actual` is [deeply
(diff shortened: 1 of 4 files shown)


jmarkham828:
@novemberborn Getting an error like this:
```
Error: Error loading ava.config.js: Not supported
    at loadJsConfig (/Users/johnmarkham/Desktop/ava/lib/load-config.js:33:23)
```

when trying to lint the files or check types (`xo`, `tsd`) to get the build green. Have you seen something like this before? Also not quite sure how to diagnose the failing type tests.

novemberborn:
I don't see that XO error output in CI, maybe it's a case of outdated dependencies installed on your machine? There are a lot of other linting errors though.

There should be some easier to solve TypeScript errors, like `SnapshotOptions` being missing. A lot has changed recently, might just be that your older commits are trying to reintroduce some code that is no longer meant to be there.

jmarkham828:
@novemberborn thanks for getting back to me. Weird... I have no diff with any packages, and no diff with any configuration files for `xo`, yet always get an error loading `ava.config.js` or `ava.config.mjs` locally...

jmarkham828:
Nice. The solution was quite weird: I commented [this](https://github.com/avajs/ava/blob/d8b1e89e6e8b88e1582cf02f4cf088c87c2e058f/lib/load-config.js#L88) out, then ran `npx xo --fix <file>` and it did the trick. Then even after uncommenting that line, `xo` did not throw an error anymore. Some weird setup bug maybe. 🤷

index.d.ts
```diff
@@ -417,7 +493,7 @@ export interface CbExecutionContext<Context = unknown> extends ExecutionContext<
 	end(error?: any): void;
 }
 
-export type ImplementationResult = PromiseLike<void> | Subscribable | void;
+export type ImplementationResult = PromiseLike<void> | Subscribable | boolean | void;
```
jmarkham828, line 496: `npx tsd` fails with more than a dozen errors about the type of `ImplementationResult` if this is not here. Any ideas on alternatives @novemberborn ?

novemberborn: This is a problems with the type definitions. It's fine for now, I'll address this separately for AVA 4.

novemberborn:
@jmarkham828 I've pushed some tweaks, notably in the tests to assert on the return values. The throws assertions shouldn't return booleans which I've also fixed.

Would you mind updating the documentation for the assertions that now return booleans?

jmarkham828:
Thank you so much for the help, @novemberborn. 🙏

novemberborn (approved):

novemberborn:
> Separately, how could [64ad93e](https://github.com/avajs/ava/commit/64ad93e8a0270c58b9fe6df22f12626b8211bb1e) have caused the build to fail 🤔

Just flakiness. A different run failed this time.

PR #2709, opened 2021-03-10
Update suggested AVA invocation in watch mode recipe

JerryGreen:
According to https://github.com/avajs/ava/discussions/2707#discussioncomment-461257

### DIFF
--- a/docs/recipes/watch-mode.md
+++ b/docs/recipes/watch-mode.md
@@ -98,14 +98,7 @@ You can update failing snapshots by typing <kbd>u</kbd> on the console, followed
 Sometimes watch mode does something surprising like rerunning all tests when you thought only a single test would be run. To see its reasoning you can enable a debug mode. This will work best with the verbose reporter:
 
 ```console
-$ DEBUG=ava:watcher npm test -- --watch --verbose
-```
-
-On Windows use:
-
-```console
-$ set DEBUG=ava:watcher
-$ npm test -- --watch --verbose
+$ DEBUG=ava:watcher npx ava --watch
 ```
 
 ## Help us make watch mode better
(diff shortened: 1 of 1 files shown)


docs/recipes/watch-mode.md
```diff
@@ -98,14 +98,7 @@ You can update failing snapshots by typing <kbd>u</kbd> on the console, followed
 Sometimes watch mode does something surprising like rerunning all tests when you thought only a single test would be run. To see its reasoning you can enable a debug mode. This will work best with the verbose reporter:
 
 ```console
-$ DEBUG=ava:watcher npm test -- --watch --verbose
-```
-
-On Windows use:
```
novemberborn, line 104: Do you think we need to keep this?

novemberborn (commented):
Thanks for opening this!

docs/recipes/watch-mode.md
JerryGreen: I personally use windows (and sometimes macos, though also used Linux in past like a couple years too), and I used `DEBUG=ava:watcher npx ava --watch`, so I don’t actually see why one would need to use this `set` API. Also, when installing git on windows, you get an upgrade to your terminal, so it’s becoming almost posix compliant. This command works.

And my logic is that: If you didn’t install git, what the heck would you do on GitHub? So no reason to keep it.

There’s much less incompatibilities with windows nowadays. Mainly it’s related to symlinks (bash cannot create it), and maybe few things like to use `explorer` until instead of `open` (not sure if it’s part of posix though, probably not)

novemberborn: > If you didn’t install git, what the heck would you do on GitHub?

Git or no Git, it's where our documentation lives, can't blame the reader for that 😄 

But yes happy to not mention the old-style Windows approach.

novemberborn:
Thank you @JerryGreen.

docs/recipes/watch-mode.md
JerryGreen: > But yes happy to not mention the old-style Windows approach

Glad we can agree on this 👍

PR #2723, opened 2021-03-23
Include test invocation in trace

### LINKED ISSUE #2722: Report does not contain a reference to the location of the test when using macros
When using a macro, the failure report does not include a line number that could be used to navigate to the test.

Example of the report:

```
  Number must be 5

  Difference:

  - 4
  + 5

  › macro (typescript/helper/macro.ts:4:7)
```

Reproduction: <https://github.com/MatthiasKunnen/ava-line-number-reproduction>.
AVA version: 3.15.0.
AVA configuration: See repo.

**Proposal**  
Capture the line number when `test` is invoked and either add it to the stack trace or add it above the diff, e.g. `typescript/spec/something.spec.ts:7`

**Notes**
Interestingly, the report is a bit different when the macro is inside vs outside the spec file. This can be seen on the [root commit in the reproduction repository](https://github.com/MatthiasKunnen/ava-line-number-reproduction/commit/be9eff30ceaa2019cfad8897baecb58c9fb7ce9b). Example diff with the `+` side representing a macro outside the spec file:

 ```diff
Number must be 5

-typescript/spec/something.spec.ts:4
 
-3: const macro: Macro<[number]> = (t, number) => {
-4:     t.is(number, 5);                           
-5: }
 
 Difference:
 
 ﹣ 4
 ﹢ 5
 
-› macro (typescript/spec/something.spec.ts:4:7)
+› macro (typescript/helper/macro.ts:4:7)
 ```

MatthiasKunnen:
Closes #2722.

@novemberborn, before I look into testing this, are you satisfied with this code? There are a few ways this can be accomplished with this one changing the least.

**The change**  

```js
const test = require('ava');

const {macro} = require('../helper/macro');

test('Number must be 5', macro, 4);
```

now results in the follow output

```diff
  Number must be 5

  Difference:

  - 4
  + 5

  › exports.macro (javascript/helper/macro.js:2:7)
+ › test (javascript/spec/something.spec.js:5:1)
```

### DIFF
--- a/lib/runner.js
+++ b/lib/runner.js
@@ -4,6 +4,7 @@ const matcher = require('matcher');
 const ContextRef = require('./context-ref');
 const createChain = require('./create-chain');
 const parseTestArgs = require('./parse-test-args');
+const {stackUtils} = require('./reporters/beautify-stack');
 const snapshotManager = require('./snapshot-manager');
 const serializeError = require('./serialize-error');
 const Runnable = require('./test');
@@ -88,6 +89,8 @@ class Runner extends Emittery {
 			}
 
 			metadata.taskIndex = this.nextTaskIndex++;
+			const testInvocation = stackUtils.parseLine(stackUtils.captureString().split('\n')[0]);
+			metadata.testLine = `    at test (${testInvocation.file}:${testInvocation.line}:${testInvocation.column})`;
 
 			const {args, buildTitle, implementations, rawTitle} = parseTestArgs(testArgs);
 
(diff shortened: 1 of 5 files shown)


lib/runner.js
```diff
@@ -88,6 +89,8 @@ class Runner extends Emittery {
 			}
 
 			metadata.taskIndex = this.nextTaskIndex++;
+			const testInvocation = stackUtils.parseLine(stackUtils.captureString().split('\n')[0]);
```
novemberborn, line 92: I'm not sure we should be capturing this at runtime. I'm a little concerned about any additional performance overhead, especially in files with lots of tests.

novemberborn:
Could we think through the various possible stack trace scenarios, and review which ones show up correctly and which ones don't?

Say:

```js
test('Number must be 5', (t, n) => macro(t, n), 4)
```

Does this show the invocation of `macro()` in the test file?

In other words, is this a problem only when the macro (or any other imported function) is used directly?

With AVA 4 I'm looking at having macros be defined explicitly (`test.macro((t, n) => …)`). We could recognize them and only get the stack for those arguments, not for other imported functions. What do you think?

novemberborn:

PR #2737, opened 2021-04-24
Don't silently swallow errors in puppeteer example

janosh:


sindresorhus:
It's not swallowing the error. If it throws, it will fail the test.

janosh:
That's true. But you don't get to know why the test failed.￼

novemberborn:
> But you don't get to know why the test failed

AVA should log the error.

`console.error()` is a useful debugging tool but let's not include it in the sample code. Thank you for your PR.

Issue #2719, opened 2021-03-14
Magic assert doesn't seem to work with "type": "module"


Filyus:
Just add `"type": "module"` to **package.json** and run the failing test in ES-module, you don't see any code in the logs. I don't want to install something like Babel.
ava 3.15.0
Node 12.20.0

test.js:
```js
import test from 'ava';
test("test magic assert", t => {
  t.is(1, 2); 
});
```
log:
```
  test magic assert

  Difference:

  - 1
  + 2

  » file:///test.js:3:5

  ─

  1 test failed
```

Related: https://github.com/avajs/ava/issues/2293

novemberborn:
It strikes me that the path is `file:///test.js` which implies the test file was at your file system's root. That seems odd. Assuming the file does not actually live there, this is probably the error.

I haven't had a chance yet to try and reproduce this but it's on my radar.

novemberborn:
I suspect the problem is here: https://github.com/avajs/ava/blob/b8f5685ed5b828de884d0d72c988efe837058692/lib/serialize-error.js#L25-L34

I haven't tested it directly, but this came up in my work for #2741.

jiby-gh:
@novemberborn not sure if it is fixed by #2741. I encountered the same issue today and tried with the master branch. The issue is still reproducible. Same setup as @Filyus, `"type": "module"` in package.json and everything as es modules.
![image](https://user-images.githubusercontent.com/52130524/128309099-48bb7514-a335-42ed-9460-7f8007e51fbb.png)

jiby-gh:
hmm maybe it's because of the source maps, the files are compiled from typescript and the source maps point to `.ts` files.

![image](https://user-images.githubusercontent.com/52130524/128311584-0f918feb-b081-414b-820a-8d77e9faccfd.png)

jiby-gh:
well, forgive my ignorance @novemberborn. Just found out that I would need additional configuration to work with precompiled typescript.

jiby-gh:
unfortunately, es modules do not work with the `@ava/typescript` yet. 

I was curious though. My setup has IDE watchers that auto compile the typescript to javascript in the same folder on change. This was a nice setup for me because I can use these auto transpiled codebase directly as the artifact without needing a bundle (at least in dev). I expected the same to work for `avajs` such that I just reference the transpiled test files and the source maps help with the stack traces.

[...]

PR #2741, opened 2021-04-26
Switch to ESM for internals

novemberborn:
* [x] Make sure the entrypoints remain usable in a CJS environment
* [x] Convert all other internal files to ESM


novemberborn:
Progress so far:

* `lib/` is ESM
	* I had to load `yargs` as CJS since it behaved differently
	* Line number parsing code is still CJS because it attempts to load some dependencies lazily and I didn't want to make it asynchronous
* Shared worker filenames must now be `file://` URLs, but you can provide a `URL` instance
* Shared workers are always loaded through a dynamic import
* `test/` is ESM, except for the tests that would have required top-level await or too much refactoring
* The XO config is now a JSON file
* I'm enforcing import ordering and formatting beyond what XO does
* I'm hoping code coverage isn't decreased
* [ ] `test-tap/` still needs migrating

plugin.d.ts
```diff
@@ -1,3 +1,5 @@
+import {URL} from 'url';
```
novemberborn, line 1: @sindresorhus is this idomatic?

novemberborn:
@sindresorhus feedback welcome, though I'm not expecting you to review each file 😉 To be honest neither am I, as long as the tests pass I think we'll be OK.

plugin.d.ts
sindresorhus: No, but TS forces us to do it: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/34960

lib/code-excerpt.js
```diff
@@ -1,14 +1,15 @@
-'use strict';
-const fs = require('fs');
-const equalLength = require('equal-length');
-const codeExcerpt = require('code-excerpt');
-const truncate = require('cli-truncate');
-const chalk = require('./chalk').get();
+import fs from 'fs';
```
sindresorhus, line 1:
```suggestion
import fs from 'node:fs';
```

Any thoughts on switching to this while you're at it? I plan to enforce it in XO soon.

https://github.com/nodejs/node/issues/38343

lib/provider-manager.js
```diff
@@ -1,21 +1,19 @@
-const pkg = require('../package.json');
-const globs = require('./globs');
+import * as globs from './globs.js';
+import pkg from './pkg.cjs';
 
-const levels = {
+export const levels = {
```
sindresorhus, line 4: IMHO, if the only reasonably way to import a package is `import * as foo`, then I think it should just be an object with a default export instead of single exports.

test-d/macros.ts
```diff
@@ -1,5 +1,6 @@
 import {expectType} from 'tsd';
-import test, {ExecutionContext, Macro} from '..'; // eslint-disable-line import/no-unresolved
+
+import test, {ExecutionContext, Macro} from '..';
```
sindresorhus, line 3:
```suggestion
import test, {ExecutionContext, Macro} from '../index.js';
```

You should use full name and extension for the TS stuff too.

test/config/loader.js
```diff
@@ -1,9 +1,12 @@
...
+import path from 'path';
+import {fileURLToPath} from 'url';
+
+import test from '@ava/test';
+
+import {loadConfig} from '../../lib/load-config.js';
 
 const CWD = process.cwd();
-const FIXTURE_ROOT = path.resolve(__dirname, '../../test-tap/fixture/load-config');
+const FIXTURE_ROOT = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '../../test-tap/fixture/load-config');
```
sindresorhus, line 9:
```suggestion
const FIXTURE_ROOT = fileURLToPath(new URL('../../test-tap/fixture/load-config', import.meta.url));
```

Could be slightly simplified.

test/shared-workers/lifecycle/fixtures/teardown.js
```diff
@@ -13,7 +14,7 @@ plugin.registerSharedWorker({
 });
 
 plugin.registerSharedWorker({
-	filename: require.resolve('./_worker.js'),
+	filename: new URL('./_worker.js', import.meta.url),
```
sindresorhus, line 17:
```suggestion
	filename: new URL('_worker.js', import.meta.url),
```

sindresorhus (dismissed):

sindresorhus:
> I had to load yargs as CJS since it behaved differently

New version was just released. Maybe it's better: https://github.com/yargs/yargs/releases/tag/v17.0.0

lib/code-excerpt.js
novemberborn: I'd be happy to let XO fix & enforce that when the time comes 😄

test-d/macros.ts
novemberborn: There's is no `index.js` file, but there is an exports map. Which the linter rules don't seem to deal with, so I have direct imports of `entrypoints/main.mjs` elsewhere. But then TypeScript can't find the type definition. Hence `..` here.

novemberborn:
> > I had to load yargs as CJS since it behaved differently
> 
> New version was just released. Maybe it's better: [yargs/yargs@`v17.0.0` (release)](https://github.com/yargs/yargs/releases/tag/v17.0.0)

I'm always happy once I figured out how `yargs` works… not sure I want to deal with its breaking changes in this here PR.

novemberborn:
Test file paths within shared workers are now file URLs.

XO should pass, but the `tap` tests are not passing yet.

novemberborn:
* I think I fixed #2719 
* Also, title prefixes in the reporters now correctly skip all test file extensions, not just `js`

novemberborn:
Tests pass locally… 😉

sindresorhus:
I haven't actually tested this, but the code changes look good to me.

iambumblehead:
@novemberborn I'm looking forward to using this in AVA's next update :)

novemberborn:
I believe the test failures are due to either a Node.js bug or are intermittent. I'm going to merge this.

Issue #2744, opened 2021-04-27
Unnecessary error in extractSource due to not checking for null


erights:
The line at
https://github.com/avajs/ava/blob/e9a69dca44dadb545d4698fb7df5587da2787f7d/lib/serialize-error.js#L31
should be changed to
```js
if (callsite && callSite.file === normalizedFile) {
```
since the previous line
```js
const callSite = stackUtils.parseLine(line);
```
may, and sometimes does, initialize `callSite` to `null`. Once this is fixed, if the catch at
https://github.com/avajs/ava/blob/e9a69dca44dadb545d4698fb7df5587da2787f7d/lib/serialize-error.js#L39
was there only because of this one error, then this do-nothing catch should be removed.

novemberborn:
Thanks, that makes sense. I'll try and get this into the next release if nobody else beats me to it.

Looking at the commit history, the earlier logic checked for a falsy value so I assume the `catch` isn't needed with this guard.

Fem-Fem:
Hi! Would it be possible for me to work on this?

novemberborn:
@Fem-Fem that'd be great!

Fem-Fem:
Thanks! I have a PR here: https://github.com/avajs/ava/pull/2750

PR #2750, opened 2021-05-07
Improve callsite handling

Fem-Fem:
This PR is to address https://github.com/avajs/ava/issues/2744. I also removed the try keyword, since if we were removing catch, I didn't see the value of using try - I might have missed something though.


erights (approved):
I'm not a reviewer or committer, but FWIW, LGTM

Fem-Fem:
> 
> 
> I'm not a reviewer or committer, but FWIW, LGTM

Thank you - and thanks for reporting the issue!

novemberborn:
Thanks @Fem-Fem!

Issue #2384, opened 2020-01-28
Have the global timeout take into account `t.timeout()` durations


spectranaut:
## What you're trying to do

I appreciate that having a global timeout in order to catch unexpected behavior in a test and not consume too much CI resources. However, some tests take a long time intentionally, and I would like to be able to use `t.timeout()` to explicitly override the global timeouts for just one test. Right now, you can only override the global timeouts if then `t.timeout()` is shorter than the global timeout. It would be nice if it overrides when `t.timeout()` is longer as well.

I think I am not the only one, as I found [this question](https://gitmemory.com/issue/avajs/ava/2365/574094523) referring to the same problem.

## Why you can't use AVA for this

There is no way to have a global timeout that is something like 10s and a timeout for a particular test that is longer, like 30s.

## And maybe how you think AVA could handle this

I'm not sure how the timeouts are coded, it might be a one-liner :)

novemberborn:
The global timeout is a "stop tests in case they're stuck" kind of timeout.

I suppose we could interpret that as "stop tests in case they're stuck, but don't worry for the next 30 seconds nice the test will fail if it's stuck anyway".

The workers will have to send a message to the main process whenever a per-test timeout starts, with the timeout duration. And then in the main process we'd have to make sure we don't stop tests because they're stuck until that duration has passed.

This should work with multiple `t.timeout()` calls, of course.

I don't think it's quite a one-liner, but this approach is consistent with what the global timeout is there for, so let's do it.

sramam:
How is described behavior different from removing the global timeout and setting a default timeout on every test?

Basically, 
- remove global timeout
- by default, `t.timeout(10000)`
- advertise this and let the user modify as necessary.

The net effect is the same IMHO.

novemberborn:
When tests run concurrently, within and across worker processes, their duration may vary wildly. The global timeout (misnamed as it may be) is only concerned with making sure something is happening in your test suite.

Individual timeouts can be useful in certain cases, but far from all.

sramam:
Disclaimer:  I have now found my peace with the nested timeout and can live with the timeout change.

However, it seems my initial concerns are being experienced/reported by others. 
Hopefully, the example below provides an insight.

The output below is meant to be slightly hyperbolic, but convey the confusion I sensed
with the timeout change on first blush.

```
$ npx ava dist/sdk/test/*.js

⠴ Server listening at http://127.0.0.1:54369

  ✖ Timed out while running tests

[...]

PR #2758, opened 2021-05-27
Have the global timeout take into account t.timeout() durations

OhYash:
This PR attempts to allow tests to run beyond the global limit (which is 10s default).  
See #2384 for details.

Tests cases and documentation pending until the `debounce()` wait period can be overridden.


lib/api.js
```diff
@@ -232,6 +232,13 @@ class Api extends Emittery {
 				}
 
 				const worker = fork(file, options, apiOptions.nodeArguments);
+				worker.onStateChange(data => {
+					if (data.Timeout && !apiOptions.debug)
```
novemberborn, line 236: The pattern is to identify the state changes using a `type` property. Perhaps `'test-timeout-configured'`? Then pass the timeout along in a separate property. `period` would make it match the existing `timeout` event.

lib/runner.js
```diff
@@ -65,6 +65,10 @@ class Runner extends Emittery {
 			return true;
 		};
 
+		this.emitStateChange = evt => {
+			this.emit('stateChange', Object.assign(evt));
```
novemberborn, line 69: `Object.assign()` looks unnecessary here?

lib/runner.js
```diff
@@ -289,19 +293,20 @@ class Runner extends Emittery {
 			powerAssert: this.powerAssert,
 			title: `${task.title}${titleSuffix || ''}`,
 			isHook: true,
-			testPassed
+			testPassed,
+			emitStateChange: this.emitStateChange
```
novemberborn, line 297: I wouldn't pass the entire `emitStateChange`. How about a `notifyOfTimeout()` function that calls `emitStateChange()` internally?

novemberborn (commented):
Thanks for turning this into a PR!

You're right that `debounce()` isn't a direct fit for this. It's perhaps also unnecessary now that we can use https://nodejs.org/docs/latest-v12.x/api/timers.html#timers_timeout_refresh. But maybe that's beside the point.

I don't think the timeout has to be accurate. The goal is for the global timeout to not fire when a test is waiting for a longer period. Perhaps we can think of these `t.timeout()` calls as *preventing* the global timeout from firing. We can compute the timestamp before which, if it fires, it should simply restart the timer. What do you think?

lib/api.js
OhYash: Updated. Thanks for pointing out.

lib/runner.js
OhYash: removed

lib/runner.js
OhYash: created a bespoke function for timeout: `notifyTimeoutUpdate()`

OhYash:
I've updated the `api.js` to skip global timeout from getting fired in case an individual test timeout is defined.  
+ Updated some bits for existing tests to still pass.  

If this approach looks good, I'll proceed with adding fresh tests specific to the new changes.

This still leaves `debounce()` limitations to be fixed later. So parallel running tests are more likely to succeed even beyond their timeout limits.

lib/api.js
```diff
@@ -232,6 +239,11 @@ class Api extends Emittery {
 				}
 
 				const worker = fork(file, options, apiOptions.nodeArguments);
+				worker.onStateChange(data => {
+					if (data.type === 'test-timeout-configured' && !apiOptions.debug) {
+						customTimeout = data.period;
```
novemberborn, line 245: Store the moment until which timeouts should be ignored, e.g.:

```suggestion
						ignoreTimeoutsUntil = Math.max(ignoreTimeoutsUntil, Date.now() + data.period);
```

Then in the debounced timer function above, return if `Date.now() < ignoreTimeoutsUntil`.

novemberborn (commented):
Thanks @OhYash! I've pushed some small changes (and merged `main`). CI is a bit choppy now but some runs should pass…

lib/api.js
OhYash: Thanks for the tip. I've updated accordingly.

OhYash:
Thanks @novemberborn for fixing some of the unnecessary changes I've made. I've fixed the CI tests and added one `api.js` test

OhYash:
Added a test for stateChange event emitted by notifyTimeoutUpdate

novemberborn (approved):

novemberborn:
Thanks @OhYash!

OhYash:
Finally! After me stretching it for too long &#128517;.  
Thanks for help, @novemberborn !!

mikob:
Thank you @OhYash! @novemberborn will this be included in a release soon? Seems like there hasn't been a release since January.

novemberborn:
@mikob we're working towards AVA 4. I'm hoping to put another prerelease of that out soon.

PR #2766, opened 2021-06-09
Detect more forms of snapshot file corruption

ninevra:
Fixes a logic error in `decodeSnapshots()` that caused it to mishandle snapshot files with a missing or corrupted readable prefix. This caused some invalid snapshot files, for instance blank ones, to silently crash the runner, resulting in e.g. "No tests found in test.js" or "test.js exited due to undefined" being printed.

This PR also treats all unexpected errors during `decodeSnapshots()` as corrupted snapshot files, just in case there are other such bugs.

The previous corrupted snapshot file tests used a valid readable prefix, missing this case. This PR adds a test case to cover the gap.


novemberborn (approved):
Awesome, thanks!

Issue #2769, opened 2021-06-13
Allow multiline snapshot label/message


KillyMXI:
Relevant code:
https://github.com/avajs/ava/blob/a8c908e2ab9deccd3383b311d988a1e88ff3d189/lib/snapshot-manager.js#L96

I'd expect something like `label.split(/\n/).map(line => '> ' + line).join('\n')`.
Or an option "I would like to format it myself" (Don't insert blockquote or anything - I'll pass the formatted text instead.)

Why:
I'd love to have snapshot reports as human readable and illustrative as possible to serve as a user documentation supplement.
Currently I have to produce not quite readable one-liners, as can be seen [here](https://github.com/mxxii/selderee/blob/main/packages/selderee/test/snapshots/snapshots.ts.md).

I plan to migrate a bigger project to AVA, and it should greatly benefit from the snapshots feature. But I also expect I might need to put more content into labels there.

novemberborn:
Official multiline support makes a lot of sense. Would you like to work on it?

Incidentally I think we may have a similar problem with test titles.

KillyMXI:
Hmm. Test titles? Do you mean you put code examples into test titles? And how that supposed to be rendered in markdown? Currently it's just `##` headers.

I think I can work on this as long as we agree on the design and it wouldn't grow much in complexity. One design question I've outlined already - it's whether snapshot label/message should be put inside a blockquote or whether there can/should be other way to handle it.

novemberborn:
Sometimes test titles have new lines which break formatting in the snapshot reports. Those should probably be stripped so the title is on a single line, unless Markdown let's you hard-wrap headings.

For individual snapshots we should render multi line block quotes.

KillyMXI:
Ah ok. Sanitizing titles makes sense. Although I wouldn't quite call it similar with this.

Gotcha on multi line block quotes.

novemberborn:
You're right it's different, your suggestion just reminded me of it. Happy to tackle as a separate piece of work 😊

KillyMXI:
I think `title.replace(/\s+/g, ' ')` might be a good way to sanitize the title. Unless I'm missing something and some whitespace characters have to be preserved.

What I find the most intimidating now is to find the right place where to put tests.
It looks to me that it will require new file(s) but it's not obvious to me where to put them.
[...]

PR #2773, opened 2021-06-17
Support multiline snapshot labels

KillyMXI:
Issue #2769

This PR only addresses the snapshot label formatting at the moment.
Changing the test title formatting has much bigger impact. I need more info there. Depending on how much bigger scope it gets it might be better to handle separately.


novemberborn (approved):
I've pushed a commit to fix up the tests, but other than that I think this is good to go!

KillyMXI:
Awesome, thank you!

PR #2776, opened 2021-06-20
Add StackBlitz example for macros

SamVerschueren:
This PR adds an `examples` directory with currently one example, being `macros`.

I've copied one macro case from the documentation as you will see. We can add a different one, or add multiple test files even. But maybe, the smaller the example, the easier it is to understand for new people. Open to suggestions here.

If you want to test it out how it will look like if you click the button, use this button which does the same but points to my fork.

[Open in StackBlitz](https://stackblitz.com/github/SamVerschueren/ava/tree/stackblitz/examples/macros/examples/macros?file=test.js&terminal=test&view=editor)

These are the properties I added to the URL
- Open `test.js` by default if you click the button
- Run the `test` command by default if someone opens the example
- Hide the preview pane on the right

One small sidenote, I did notice that the `view=editor` query parameter is not working and the preview pane is still open, but I will look into that to fix it asap.

Some small questions
- Do you guys want to use the `Open in StackBlitz` button? Or do you rather prefer textual links?
- Is it ok to embed the buttons in the `docs` documentation? I assume that the easier it is to play with a feature, the better. Then people don't have to skim through the examples.
- Do we want to put more information in the `readme.md` of the example? This is what it looks like now https://github.com/SamVerschueren/ava/tree/stackblitz/examples/macros/examples/macros. I didn't want to copy parts of the documentation because it can only become outdated.

### DIFF
--- a/examples/macros/test.js
+++ b/examples/macros/test.js
@@ -1,12 +1,13 @@
 'use strict';
 const test = require('ava');
-const { sum } = require('.');
+
+const {sum} = require('.');
 
 function macro(t, a, b, expected) {
 	t.is(sum(a, b), expected);
 }
 
-macro.title = (providedTitle = '', a, b, expected) => `${providedTitle} ${a}+${b} = ${expected}`.trim();
+macro.title = (providedTitle, a, b, expected) => `${providedTitle || ''} ${a}+${b} = ${expected}`.trim();
 
 test(macro, 2, 2, 4);
 test(macro, 3, 3, 6);
--- a/examples/macros/test.js
+++ b/examples/macros/test.js
@@ -1,0 +1,13 @@
+'use strict';
+const test = require('ava');
+const { sum } = require('.');
+
+function macro(t, a, b, expected) {
+	t.is(sum(a, b), expected);
+}
+
+macro.title = (providedTitle = '', a, b, expected) => `${providedTitle} ${a}+${b} = ${expected}`.trim();
+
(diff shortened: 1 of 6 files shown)


novemberborn:
I think this all great. The button is a bit attention grabby but that's fine for now.

We may want to come up with a better example though. XO rejects it.

SamVerschueren:
Yeah noticed this as well. Maybe just a `sum` method in index.js?

novemberborn:
Sounds good.

If you merge in `main` CI should pass again.

examples/macros/test.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+const test = require('ava');
+const { sum } = require('.');
+
+function macro(t, a, b, expected) {
+	t.is(sum(a, b), expected);
+}
+
+macro.title = (providedTitle = '', a, b, expected) => `${providedTitle} ${a}+${b} = ${expected}`.trim();
```
SamVerschueren, line 9: @novemberborn linting fails on the `providedTitle = ''`, you want me to remove it and use `${providedTitle || ''}` instead?

SamVerschueren: I removed the default value and used `|| ''`.

SamVerschueren:
When clicking the button, you will now see that the preview window is gone and the editor takes the entire screen. I fixed that issue upstream.

novemberborn:
Nice, thanks @SamVerschueren!

PR #2779, opened 2021-06-22
Normalize test titles

### LINKED ISSUE #2774: Gracefully handle linebreaks in test titles
If test titles include line breaks, the snapshot reports for those tests will look a bit weird since the formatting code does not handle them. Reporter output may be broken too.

Per discussion in https://github.com/avajs/ava/issues/2769#issuecomment-862664084 we could normalize whitespace in test titles at the point of declaration, which then impacts uniqueness checks, or we could do so when displaying.

My preference would actually be to do this at the point of declaration, which should be considered a breaking change given that we identify the test by its title. But that's OK since we're doing pre-releases for AVA 4 at the moment.

@KillyMXI what do you think?

KillyMXI:
Fixes #2774

I don't know how to force LF line endings so it needs a linting fix.


novemberborn:
@KillyMXI I've made some tweaks, is there anything else left outstanding?

KillyMXI:
No, I don't think so. lgtm
You've covered things I missed, so if there is anything else - it will be above my understanding of the codebase.

novemberborn:
@KillyMXI awesome, thanks!

PR #2784, opened 2021-07-04
Add StackBlitz example for snapshot testing

SamVerschueren:
This PR adds an example for snapshot testing of a React component.

I could've made the snapshot example without React, but because the docs use that React component, I figured it would be best to just follow that example instead.

You can test it out with the following button which points to my fork.

[Open in StackBlitz](https://stackblitz.com/github/SamVerschueren/ava/tree/stackblitz/snapshot-react/examples/snapshot-testing?file=index.js&terminal=test&view=editor)

image

One downside in this example is that I had to go with a `snapshotDirectory` starting with a `.`. The reason is that we don't sync this directory with our backend. Because the snapshots are binary files, I receive the PostgreSQL error `\u0000 cannot be converted to text.` which is a null-byte.

### DIFF
--- a/examples/snapshot-testing/index.js
+++ b/examples/snapshot-testing/index.js
@@ -1,0 +1,5 @@
+import React from 'react';
+
+const HelloWorld = () => <h1>Hello World...!</h1>;
+
+export default HelloWorld;
--- a/examples/snapshot-testing/index.js
+++ b/examples/snapshot-testing/index.js
@@ -1,3 +1,4 @@
+/* eslint-disable no-unused-vars */
 import React from 'react';
 
 const HelloWorld = () => <h1>Hello World...!</h1>;
--- a/examples/snapshot-testing/test.js
+++ b/examples/snapshot-testing/test.js
@@ -1,0 +1,9 @@
+import test from 'ava';
+import React from 'react';
+import render from 'react-test-renderer';
+import HelloWorld from '.';
+
+test('HelloWorld component', t => {
+	const tree = render.create(<HelloWorld />).toJSON();
+	t.snapshot(tree);
+});
--- a/examples/snapshot-testing/test.js
+++ b/examples/snapshot-testing/test.js
@@ -1,7 +1,9 @@
+/* eslint-disable no-unused-vars */
 import test from 'ava';
 import React from 'react';
 import render from 'react-test-renderer';
-import HelloWorld from '.';
+
+import HelloWorld from './index.js';
 
 test('HelloWorld component', t => {
 	const tree = render.create(<HelloWorld />).toJSON();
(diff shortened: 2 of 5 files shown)


examples/snapshot-testing/index.js
```diff
@@ -0,0 +1,6 @@
+/* eslint-disable no-unused-vars */
```
SamVerschueren, line 1: This is ugh... I have to import `React` here, although it's not used :(

examples/snapshot-testing/test.js
```diff
@@ -0,0 +1,11 @@
+/* eslint-disable no-unused-vars */
```
SamVerschueren, line 1: Same here. It complains about `React` and `HelloWorld`. Although `HelloWorld` is used. Probably because the extension is not `jsx`?

novemberborn:
Which documentation are you referring to? AVA 4 removes built-in, specific support for React.

> One downside in this example is that I had to go with a `snapshotDirectory` starting with a `.`. The reason is that we don't sync this directory with our backend. Because the snapshots are binary files, I receive the PostgreSQL error `\u0000 cannot be converted to text.` which is a null-byte.

Can't say I follow 😄

SamVerschueren:
Yes, the AVA docs use a React example to show snapshot testing. Probably because it's used very often in React. See https://github.com/avajs/ava/blob/main/docs/04-snapshot-testing.md.

> Can't say I follow 

Ok sorry about this 😂 . Let me try to explain a bit clearer because I really explained it very poorly :p.

During testing, I noticed that snapshots generated by AVA are actually binary files. On StackBlitz, binary files can't just be stored in the database (PostgreSQL), so we use S3 for that.

We currently test for binary files based on extension (not ideal and we might want to improve that). But because `.snap` is not really a binary format extension, it tries to store it's content in PostgreSQL, which fails because of a null byte.

To circumvent this issue, I use a dot-directory to store the snapshots. Currently, StackBlitz does not sync dot-directories to it's back-end, and thus to PostgreSQL.

Another issue with binary files is that currently (we're looking into this again), free users can't upload binary files. But again, as long as the snapshots are stored in a dot-directory, they should be fine. Unless there's a way of generating non-binary snapshot files? But I guess not :p.

novemberborn:
> Unless there's a way of generating non-binary snapshot files? But I guess not :p.

Nope.



> Another issue with binary files is that currently (we're looking into this again), free users can't upload binary files.

Perhaps there should be a file size limit?

---

As it stands AVA 4 does not support snapshots for React, so perhaps that's not the greatest example.

SamVerschueren:
> Perhaps there should be a file size limit?

Yes, that's probably what we're going to do.

> As it stands AVA 4 does not support snapshots for React, so perhaps that's not the greatest example.

Do you mean I should maybe just come up with a different and more simple snapshot example? Or just no snapshot example at all?

novemberborn:
A non-React example would be fine I think.

novemberborn:

PR #2785, opened 2021-07-05
Breaking changes to the test() interface

### LINKED ISSUE #2189: Non-mutating API for specifying macros?
Hi all, I love the consistency of AVA’s API (especially the lack of magically-present globals).

Would the project be open to re-evaluating how macros are defined? It’s the only thing I consistently refer back to the documentation to use rather than discovering the type via inference/hover/autocomplete. It would also be nice to avoid mutating a Macro object to define test titles (when using [tslint-immutable](https://github.com/jonaskello/tslint-immutable) or similar).

So rather than:
```ts
import test, { Macro } from 'ava';

const macro: Macro<[string, number]> = (t, input, expected) => {
	t.is(eval(input), expected);
}

// tslint:disable-next-line: no-object-mutation
macro.title = (providedTitle = '', input, expected) => `${providedTitle} ${input} = ${expected}`.trim();

test(macro, '2 + 2', 4);
test(macro, '2 * 3', 6);
test('providedTitle', macro, '3 * 3', 9);
```

Maybe something like:
```ts
import test, { Macro } from 'ava';

const macro: Macro<[string, number]> = {
    macro: (t, input, expected) => {
	t.is(eval(input), expected);
    },
    title: (providedTitle = '', input, expected) => `${providedTitle} ${input} = ${expected}`.trim();
} 

test(macro, '2 + 2', 4);
test(macro, '2 * 3', 6);
test('providedTitle', macro, '3 * 3', 9);
```

I think this might also make the macro API a little easier to discover by type inference – you'd start getting autocomplete assistance immediately after providing an object to the `test` method, e.g.:
`test({ [autocomplete offers "macro" and "title" here] })`.

Thoughts? If there’s interest, I can try to send a non-breaking PR (where the mutating API is still supported).

### LINKED ISSUE #2253: TryFn type makes ExecutionContext not covariant
## Description

When moving some utility initialization into a separate file / module, we sometimes use a base `BaseContext` interface which could be extended as needed by other tests. The issue can be worked around with by not defining functions using the `ExecutionContext` type, but sometimes would just be more verbose and in other cases less type-safe.

### Test Source

```typescript
import anyTest, { TestInterface, ExecutionContext } from 'ava';

interface BaseContext  { a: string; }
interface Context extends BaseContext { b: string; }

const test = anyTest as TestInterface<Context>;
function baseBefore(t: ExecutionContext<BaseContext>){ }
test.beforeEach(baseBefore);
```

```
TS2345: Argument of type '(t: ExecutionContext<BaseContext>) => void' is not assignable to parameter of type 'OneOrMoreMacros<[], Context>'.
  Type '(t: ExecutionContext<BaseContext>) => void' is not assignable to type 'UntitledMacro<[], Context>'.
    Types of parameters 't' and 't' are incompatible.
      Type 'ExecutionContext<Context>' is not assignable to type 'ExecutionContext<BaseContext>'.
        Types of property 'try' are incompatible.
          Type 'TryFn<Context>' is not assignable to type 'TryFn<BaseContext>'.
            Types of parameters 'fn' and 'fn' are incompatible.
              Type 'EitherMacro<any, BaseContext>' is not assignable to type 'EitherMacro<any, Context>'.
                Type 'UntitledMacro<any, BaseContext>' is not assignable to type 'EitherMacro<any, Context>'.
                  Type 'UntitledMacro<any, BaseContext>' is not assignable to type 'UntitledMacro<any, Context>'.
                    Type 'BaseContext' is not assignable to type 'Context'.
```

### Config

Default

### Command-Line Arguments

```
tsc
```

## Environment

TypeScript 3.6.3, AVA 2.4.0

### LINKED ISSUE #2716: Breaking changes to the test interface for AVA 4
See https://github.com/avajs/ava/pull/2435. This will improve context typing and simplify usage.

* [x] Add `test.macro(exec)` and `test.macro({exec, title})`. This will return a `test.Macro` instance which can be passed to `test()`
* [x] Remove ability to pass an array of macros or implementations to `test()`
* [x] Fix the typing of test implementations (or macro executors) to allow promise-returning, observable returning and any-returning functions

novemberborn:
Fixes #2716.

Remove support for passing multiple test implementations. This affects test() and t.try().

Supporting multiple implementations made both the implementation and the type definitions needlessly complex. Using for/of syntax it's easy enough to declare tests (or t.try() attempts) with the same arguments.

Change constructor type for `instanceOf` expectation. It should result in an `Error` subclass.

Add `test.macro()`, change how macros are typed. Fixes #2189. `test.macro()` returns an object that can be used with
`test()` and hooks. The `t.context` type is inherited from `test`.

Like with AVA 3, regular functions that also have a `title` property that is a string-returning function are supported. However this is no longer in the type definition.

Instead the recommended approach is to use `test.macro()` to declare macros. At a TypeScript level these are easier to discriminate from regular implementations.

Change implementation argument type constraint to unknown. This mostly impacts macros, for which we'd want to encourage explicit types.

Add test for context covariance. Fixes #2253 which no longer seems reproducible.

Add test for implementation result types.

Refactor type definition. Split across multiple files. Rename types, most noticeably `TestInterface` is now `TestFn`, but really anything that had the Interface suffix has been changed. Update some documentation.

Return null when throws() and throwsAsync() assertions fail.

Update type definitions to disallow null throws expectations.


novemberborn:
Note that `test.macro()` isn't compatible with our ESLint plugin: https://github.com/avajs/eslint-plugin-ava/issues/329

sindresorhus:
Re https://github.com/avajs/ava/pull/2785/commits/d546fe0dcb66ce8b5b63eab229aceb18c4a8cd41, why `null` and not `undefined`?

---

Context: https://github.com/sindresorhus/meta/discussions/7

novemberborn:
> Re [d546fe0](https://github.com/avajs/ava/commit/d546fe0dcb66ce8b5b63eab229aceb18c4a8cd41), why `null` and not `undefined`?

Because I like `null`. However we can change it to `undefined`.

PR #2788, opened 2021-07-19
Normalize glob patterns with trailing slashes

### LINKED ISSUE #2781: Trailing slashes break directory matching
Works:

```json
{
  "ava": {
    "files": ["!src/test"]
  }
}
```

Doesn't work:

```json
{
  "ava": {
    "files": ["!src/test/"]
  }
}
```

See https://github.com/prettier/prettier/pull/11000 for similar issue

asaid-0:
Fixes #2781


asaid-0:
@novemberborn could you please review this

novemberborn:
@asaid-0 thanks for the PR, I may have a chance to look at this on the weekend.

lib/glob-helpers.cjs
```diff
@@ -110,21 +110,26 @@ function normalizeFileForMatching(cwd, file) {
...
 
+const normalizeNegatedPattern = (cwd, pattern) => {
+	// Remove `!` from pattern
+	const patternPath = pattern.slice(1);
+	const relativePath = path.relative(cwd, patternPath);
+	return `!${relativePath}`;
+};
+
 function normalizePattern(pattern) {
+	const cwd = process.cwd();
```
novemberborn, line 121: `cwd` needs to be an argument here. AVA determines the directory to which patterns are relevant and it may not be the CWD.

lib/glob-helpers.cjs
```diff
@@ -110,21 +110,26 @@ function normalizeFileForMatching(cwd, file) {
...
 	}
 
-	if (pattern.startsWith('./')) {
-		return pattern.slice(2);
-	}
-
-	if (pattern.startsWith('!./')) {
-		return `!${pattern.slice(3)}`;
+	if (pattern.startsWith('!')) {
+		return normalizeNegatedPattern(cwd, pattern);
```
novemberborn, line 128: I don't think the helper function is all that useful, especially since it hides the critical `path.relative()` bit which is then called directly below.

novemberborn (commented):
@asaid-0 it's been a while since I've had to think about glob patterns and file paths. Why does this solution do the trick?

lib/glob-helpers.cjs
asaid-0: > @asaid-0 it's been a while since I've had to think about glob patterns and file paths. Why does this solution do the trick?

@novemberborn Thank you for your feedback, actually, I used this solution assuming that we should also solve the issue with such patterns `tests//test1`, `tests/sub/../test2` not only the trailing slash `tests/test/`

if we need to just solve the trailing slash issue I wouldn't use `path.relative()`
what do you think?

asaid-0:
> @asaid-0 it's been a while since I've had to think about glob patterns and file paths. Why does this solution do the trick?

@novemberborn Thank you for your feedback, actually, I used this solution assuming that we should also solve the issue with such patterns tests//test1, tests/sub/../test2 not only the trailing slash tests/test/

if we need to just solve the trailing slash issue I wouldn't use path.relative()
what do you think?

lib/glob-helpers.cjs
asaid-0: thank you for pointing this out, I see now I think we should receive the CWD as an argument in the `normalizeGlobs` here:
https://github.com/avajs/ava/blob/dc93f372e6a1af5f3ec4b90329d63e2f1e20e4a3/lib/globs.js#L35

Then we pass it to `normalizePatterns` call in the same function.

Also i noticed this line in cli.js
https://github.com/avajs/ava/blob/dc93f372e6a1af5f3ec4b90329d63e2f1e20e4a3/lib/cli.js#L396

which using nearly the same function path.relative, but path.resolve is being used in the second parameter to get the absolute path

novemberborn:
> Thank you for your feedback, actually, I used this solution assuming that we should also solve the issue with such patterns `tests//test1`, `tests/sub/../test2` not only the trailing slash `tests/test/`
> 
> if we need to just solve the trailing slash issue I wouldn't use `path.relative()`
> what do you think?

@asaid-0 sorry I lost track of this.

These are glob patterns so I'm not sure how safe it is to pass them to `path` functions. I don't think we have to worry about those edge cases. Trailing slashes though are easier to type, and it's surprising if that breaks the pattern.

PR #2789, opened 2021-07-22
Add StackBlitz example for test timeouts

SamVerschueren:
You can test the example on my fork by clicking the button

[Open in StackBlitz](https://stackblitz.com/github/SamVerschueren/ava/tree/stackblitz/examples/timeout/examples/timeouts?file=test.js&terminal=test&view=editor)

Example is pretty straight forward I think. Let me know if you want something differently.

### DIFF
--- a/examples/timeouts/test.js
+++ b/examples/timeouts/test.js
@@ -1,0 +1,26 @@
+'use strict';
+const test = require('ava');
+
+const {delay} = require('.');
+
+test('start server', async t => {
+	t.timeout(100);
+
+	await delay(50);
+
+	t.pass();
+});
+
+test('connect with database', async t => {
+	t.timeout(100, 'make sure database has started');
+
+	await delay(200);
+
+	t.pass();
+});
+
+test('very long and slow operation', async t => {
+	await delay(3000);
+
+	t.pass();
+});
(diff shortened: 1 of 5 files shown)


examples/timeouts/test.js
```diff
@@ -0,0 +1,26 @@
+'use strict';
+const test = require('ava');
+
+const {delay} = require('.');
```
novemberborn, line 4: What's the Node.js version used in Stackblitz? With 16 we can use `timers/promises`: https://nodejs.org/docs/latest-v16.x/api/timers.html#timers_timers_promises_api

examples/timeouts/test.js
```diff
@@ -0,0 +1,26 @@
...
+	t.pass();
+});
+
+test('connect with database', async t => {
+	t.timeout(100, 'make sure database has started');
+
+	await delay(200);
+
+	t.pass();
+});
```
novemberborn, line 20: These kind of tasks you'd typically do in a `before` hook. Which should work with `t.timeout()` but that may be more confusing to people? So perhaps just make something up?

examples/timeouts/test.js
SamVerschueren: We're currently on `14.x` unfortunately.

examples/timeouts/test.js
SamVerschueren: Yes you're right. I'll change them.

SamVerschueren: Done. Hope this is a bit better :).

PR #2792, opened 2021-07-23
Add StackBlitz examples for TypeScript

SamVerschueren:
I improved the TypeScript support on StackBlitz and added two TypeScript examples.

- [Basic example for AVA with TypeScript](https://stackblitz.com/github/SamVerschueren/ava/tree/stackblitz/examples/typescript/examples/typescript-basic?file=source%2Ftest.ts&terminal=test&view=editor)
- [TypeScript example for typing `t.context`](https://stackblitz.com/github/SamVerschueren/ava/tree/stackblitz/examples/typescript/examples/typescript-context?file=source%2Ftest.ts&terminal=test&view=editor)

### DIFF
--- a/.xo-config.json
+++ b/.xo-config.json
@@ -3,7 +3,8 @@
 		"media/**",
 		"test/config/fixtures/config-errors/test.js",
 		"test-tap/fixture/snapshots/test-sourcemaps/build/**",
-		"test-tap/fixture/report/edgecases/ast-syntax-error.cjs"
+		"test-tap/fixture/report/edgecases/ast-syntax-error.cjs",
+		"examples/typescript-*/**/*.ts"
 	],
 	"rules": {
 		"import/no-anonymous-default-export": "off",
--- a/examples/typescript-context/package.json
+++ b/examples/typescript-context/package.json
@@ -1,0 +1,21 @@
+{
+	"name": "ava-typescript-context",
+	"description": "TypeScript example for typing t.context",
+	"scripts": {
+		"test": "ava"
+	},
+	"devDependencies": {
+		"@sindresorhus/tsconfig": "^1.0.2",
+		"ava": "^3.15.0",
+		"ts-node": "^10.0.0",
+		"typescript": "^4.3.4"
+	},
+	"ava": {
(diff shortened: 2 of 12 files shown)


SamVerschueren:
@sindresorhus I use a `tsconfig.json` file which extends from `@sindresorhus/tsconfig` for the examples. But `@sindresorhus/tsconfig` does not exist after the CI runs `npm install`. So we end up with this error.

image

Can I somehow disable the `tsconfig` inference and use the built-in `parserOptions.project`? I could disable linting of those files, but it's a bit of a bummer 🤔 .

examples/typescript-context/package.json
```diff
@@ -0,0 +1,21 @@
...
+		"ts-node": "^10.0.0",
+		"typescript": "^4.3.4"
+	},
+	"ava": {
+		"extensions": [
+			"ts"
+		],
+		"require": [
+			"ts-node/register"
+		]
```
novemberborn, line 19: Can we make this work with `@ava/typescript`? Perhaps the built-in compiler mode that's in a pre-release?

SamVerschueren: Done! I totally forgot about `@ava/typescript`, my bad! This is much better. Also the `compile` flag seems to work with version 2 :).

.xo-config.json
```diff
@@ -3,7 +3,8 @@
 		"media/**",
 		"test/config/fixtures/config-errors/test.js",
 		"test-tap/fixture/snapshots/test-sourcemaps/build/**",
-		"test-tap/fixture/report/edgecases/ast-syntax-error.cjs"
+		"test-tap/fixture/report/edgecases/ast-syntax-error.cjs",
+		"examples/typescript-*/**/*.ts"
```
SamVerschueren, line 7: Disabling it for all `typescript` projects now. It's the easiest solution I guess.

SamVerschueren:
Not sure why, but sometimes some tests are flaky. Is that normal on Windows?

novemberborn:
> Not sure why, but sometimes some tests are flaky. Is that normal on Windows?

Yes 🤷

Issue #2481, opened 2020-04-30
How to clear ava cache or any temporary files?


jcubic:
I was working on my unit tests and I was running unit tests from Emacs ansi-term in watch mode using  inotifywait (so they was running in each file save) and something break because now each time I run the tests, every test is marked as pending and passing in same time (first is list all tests with pending status then list of all tests that are passing).

So I tough that this may be some kind of temporary files or cache that ava is creating that was not cleared. I've stashed all the changes I was working on got the same issues, same commit on Travis work fine.

jcubic:
Just found this but it don't work: `./node_modules/.bin/ava reset-cache`.

jcubic:
I'm running ava from nyc (too see conveage) but removing the both cache files don't work.

jcubic:
I've commited trivial change and now my clean repo also fail, I'm running `nyc ava --verbose test.js` every tests is marked as pending and passing.

jcubic:
Finally solved the issue using really long timeout: `nyc ava --timeout 10m --verbose test.js` 

Maybe there are just lot of tests and since they are all async it take time.

novemberborn:
FWIW, AVA itself no longer writes temporary files. That only happens if you use `@ava/babel`.

jcubic:
Yea it was not cache as I primary though it's just the my tests runs finish in 27 seconds (`user	0m27,328s` according to linux `time` command) and everything is async so it's probably real timeout but I'm not sure why it work fine without nyc but fail with nyc.

novemberborn:
`nyc` adds overhead.

PR #2794, opened 2021-07-27
Clarify caching

mcecode:
Hi! I'm new to AVA and while I was using it I noticed that subsequent runs didn't seem to finish faster even though I wasn't changing some of my test files. This prompted me to try setting up caching, so I set `"cache": true` in my `package.json` like the docs suggested. However, it didn't seem to make a difference and I couldn't find the specified `node_modules/.cache/ava` directory where the cache is supposed to be. Turns out, [AVA itself doesn't cache files unless used with `@ava/babel`](https://github.com/avajs/ava/issues/2481#issuecomment-622302992). This pull request aims to clarify this, both in the docs and when running `npx ava reset-cache`.

Some notes on the changes I made to the docs:

- I specified that the `cache` option is boolean because some (like I did) might think that it can be set to a directory where the cache will then be set.

Some notes on the changes I made to the `reset-cache` command:

- I removed the return statement at the end of the if check for `resetCache` because it is redundant as either `process.exit()` or `exit()` already terminates the script.
- I changed `console.error()` to `console.log()` because the message logged isn't an error.
- I removed the `nodir: true` option from `del` as it is not a valid option in `fast-glob`, [`fast-glob` instead uses `onlyFiles`](https://github.com/mrmlnc/fast-glob#compatible-with-node-glob) which is then set to `true` by default.

### DIFF
--- a/docs/05-command-line.md
+++ b/docs/05-command-line.md
@@ -224,7 +224,7 @@ AVA itself does not cache files unless used with our [`@ava/babel`](https://gith
 npx ava reset-cache
 ```
 
-This notifies you if cache files within the `node_modules/.cache/ava` directory are deleted or if there are no cache files to be deleted.
+This deletes all files in the `node_modules/.cache/ava` directory.
 
 ## Reporters
 
--- a/docs/05-command-line.md
+++ b/docs/05-command-line.md
@@ -218,7 +218,7 @@ When running a file with and without line numbers, line numbers take precedence.
 
 ## Resetting AVA's cache
 
-AVA itself does not cache files unless used with our [`@ava/babel`](https://github.com/avajs/babel) provider. However, if it seems like your latest changes aren't being picked up by AVA you can try resetting the cache by running:
+AVA itself does not cache files unless used with our [`@ava/babel`](https://github.com/avajs/babel) provider. If it seems like your latest changes aren't being picked up by AVA you can try resetting the cache by running:
 
 ```console
 npx ava reset-cache
--- a/docs/05-command-line.md
+++ b/docs/05-command-line.md
@@ -218,13 +218,13 @@ When running a file with and without line numbers, line numbers take precedence.
 
 ## Resetting AVA's cache
 
-AVA may cache certain files, especially when you use our [`@ava/babel`](https://github.com/avajs/babel) provider. If it seems like your latest changes aren't being picked up by AVA you can reset the cache by running:
+AVA itself does not cache files unless used with our [`@ava/babel`](https://github.com/avajs/babel) provider. However, if it seems like your latest changes aren't being picked up by AVA you can try resetting the cache by running:
 
 ```console
 npx ava reset-cache
 ```
 
-This deletes all files in the `node_modules/.cache/ava` directory.
+This notifies you if cache files within the `node_modules/.cache/ava` directory are deleted or if there are no cache files to be deleted.
 
 ## Reporters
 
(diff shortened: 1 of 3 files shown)


novemberborn:
Hi @mcecode, thanks for the PR! Unfortunately I haven't had a chance to look at it yet.

mcecode:
It's okay, I know you guys must be busy with the AVA 4 release and other things. So just tell me if I'm missing something or if need to change anything once you get around to review it. Thanks a lot!

docs/05-command-line.md
```diff
@@ -218,13 +218,13 @@ When running a file with and without line numbers, line numbers take precedence.
 
 ## Resetting AVA's cache
 
-AVA may cache certain files, especially when you use our [`@ava/babel`](https://github.com/avajs/babel) provider. If it seems like your latest changes aren't being picked up by AVA you can reset the cache by running:
+AVA itself does not cache files unless used with our [`@ava/babel`](https://github.com/avajs/babel) provider. However, if it seems like your latest changes aren't being picked up by AVA you can try resetting the cache by running:
```
novemberborn, line 221:
```suggestion
AVA itself does not cache files unless used with our [`@ava/babel`](https://github.com/avajs/babel) provider. If it seems like your latest changes aren't being picked up by AVA you can try resetting the cache by running:
```

`However` makes it seem like this does apply even if you don't use `@ava/babel`.

docs/05-command-line.md
```diff
@@ -218,13 +218,13 @@ When running a file with and without line numbers, line numbers take precedence.
...
 
 ```console
 npx ava reset-cache
 ```
 
-This deletes all files in the `node_modules/.cache/ava` directory.
+This notifies you if cache files within the `node_modules/.cache/ava` directory are deleted or if there are no cache files to be deleted.
```
novemberborn, line 227: I think the original copy was fine. If it logs that no files were deleted that's OK but that there's no need to mention that here.

docs/06-configuration.md
```diff
@@ -45,7 +45,7 @@ Arguments passed to the CLI will always take precedence over the CLI options con
 - `files`: an array of glob patterns to select test files. Files with an underscore prefix are ignored. By default only selects files with `cjs`, `mjs` & `js` extensions, even if the pattern matches other files. Specify `extensions` to allow other file extensions
 - `ignoredByWatcher`: an array of glob patterns to match files that, even if changed, are ignored by the watcher. See the [watch mode recipe for details](https://github.com/avajs/ava/blob/main/docs/recipes/watch-mode.md)
 - `match`: not typically useful in the `package.json` configuration, but equivalent to [specifying `--match` on the CLI](./05-command-line.md#running-tests-with-matching-titles)
-- `cache`: cache compiled files under `node_modules/.cache/ava`. If `false`, files are cached in a temporary directory instead
+- `cache`: a boolean value on whether to cache compiled files under `node_modules/.cache/ava`. If `false`, files are cached in a temporary directory instead
```
novemberborn, line 48: How about this?

```suggestion
- `cache`: defaults to `true` to cache compiled files under `node_modules/.cache/ava`. If `false`, files are cached in a temporary directory instead
```

lib/cli.js
```diff
@@ -244,18 +245,21 @@ export default async () => { // eslint-disable-line complexity
...
-			});
-			console.error(`\n${chalk.green(figures.tick)} Removed AVA cache files in ${cacheDir}`);
+			await fsPromises.access(cacheDir);
+		} catch {
+			console.log(`\n${chalk.green(figures.tick)} No cache files to remove`);
+			process.exit(0); // eslint-disable-line unicorn/no-process-exit
+		}
+
+		try {
+			await del('*', {cwd: cacheDir});
```
novemberborn, line 257: Looking at the docs for `del`, it returns the paths of the removed files. So we can check if that's empty and update the messaging based on that. I think that's simpler than having a separate check for the directory.

novemberborn (commented):
Nice, thanks for this @mcecode! I've left some inline comments. I'll also approve CI now so let's see if that flags anything.

docs/05-command-line.md
mcecode: Agreed, I was trying to be more explicit at the time but I realize now that this extra info is probably unnecessary.

lib/cli.js
mcecode: Thanks for pointing that out. I can't believe I missed that 😅, I just defaulted to using `fs`. I've made a commit fixing this.

mcecode:
Hey @novemberborn, thanks for the feedback! I've added some commits that I think resolve them already. Just looking at the logs, I think the CI failed because the lines I added we're not covered by the tests but I'm not so sure. Just tell me if there's anything else I can do.

novemberborn (approved):
Awesome, thanks @mcecode!

PR #2809, opened 2021-08-16
Return undefined when throws assertions fail

### LINKED ISSUE #2806: Change throws assertions to return undefined, not null, when the assertion fails
Per https://github.com/avajs/ava/pull/2785#issuecomment-897162682.

mcecode:
Fixes #2806


novemberborn (approved):
Great!

PR #2825, opened 2021-09-07
Ethanfann/remove trim off newlines dependency

ethanfann:
In response to https://nvd.nist.gov/vuln/detail/CVE-2021-23425, this PR removes the `https://www.npmjs.com/package/trim-off-newlines` package in favor of a utility function.

Tests have been added to ~~`test/helpers/trim-off-newlines.js`~~ `test/trim-off-newlines/test.js` with various cases related to Windows, Mac, and Linux operating systems.

Some test failures are occurring with `test-tap/integration/watcher.js`, however, I do not know if they are related to the removal of the previous trim packages library in favor of the current. Please let me know if there is a case that the new function isn't handling.


lib/trim-off-newlines.js
```diff
@@ -0,0 +1,4 @@
+export default function trimOffNewlines(text) {
+	const regex = '/^[\r\n]+|[\r\n]+$/g';
```
sindresorhus, line 2: Make it a regex literal and just put it inline in the replace call.

test/helpers/trim-off-newlines.js
```diff
@@ -0,0 +1,21 @@
+import test from '@ava/test';
+
+import trimOffNewlines from '../../lib/trim-off-newlines.js';
```
sindresorhus, line 3:
```suggestion
import test from '@ava/test';
import trimOffNewlines from '../../lib/trim-off-newlines.js';
```

lib/trim-off-newlines.js
ethanfann: I've updated the PR with your suggestion 👍

test/helpers/trim-off-newlines.js
ethanfann: I've update the PR with your suggestion 👍

KillyMXI:
Tests seem weird. As if it expected to remove all newlines while the regex is only to trim them from both ends of a string.
Looks like some expectations are wrong.

*(I'm looking from mobile atm so I might be wrong, sorry if that's the case.)*

ethanfann:
Thanks @KillyMXI! You're right. Not only was the test faulty, the test file wasn't even being picked up when the suite ran. Apologies for this and I've updated the PR with the fixed tests.

KillyMXI:
I'd still keep a test for inner newlines being unaffected though.

novemberborn:
@ethanfann thanks for the PR! I had a look at where / how we use this dependency and I reckon we can remove it entirely. See https://github.com/avajs/ava/pull/2826.

PR #2840, opened 2021-09-18
Remove Babel support, and therefore enhanced assertions

novemberborn:
This removes support for @ava/babel. Since enhanced assertions (power assert) are provided through that package this also removes support for those. Same for improper throws detection.

@ava/babel has seen no meaningful contributions since it was released at the start of 2020. It's not clear how we'd make it work with ESM. It's clearer if we can recommend tools like @babel/register and otherwise focus on plain Node.js & TypeScript support.


sindresorhus:
I very much agree 👍

PR #2843, opened 2021-09-21
Use c8 for code coverage examples instead of nyc

bompus:
As discussed in #2842 , this PR replaces mentions of using nyc for code coverage in documentation with c8.

### DIFF
--- a/lib/reporters/beautify-stack.js
+++ b/lib/reporters/beautify-stack.js
@@ -4,7 +4,7 @@ const stackUtils = new StackUtils({
 	ignoredPackages: [
 		'@ava/typescript',
 		'ava',
-		'c8',
+		'nyc',
 	],
 	internals: [
 		// AVA internals, which ignoredPackages don't ignore when we run our own unit tests.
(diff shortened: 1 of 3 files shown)


lib/reporters/beautify-stack.js
```diff
@@ -4,7 +4,7 @@ const stackUtils = new StackUtils({
 	ignoredPackages: [
 		'@ava/typescript',
 		'ava',
-		'nyc',
```
novemberborn, line 7: We can keep `nyc` here for now. `c8` uses code coverage reports generated by Node.js and shouldn't appear in the stack traces.

novemberborn:
Thanks for the PR @bompus!

lib/reporters/beautify-stack.js
bompus: Sounds good to me. I've reverted this change so this file should have no changes now.

novemberborn: Great!

Issue #2848, opened 2021-09-27
Some functions from Lodash can be replaced with native ones


live627:
For example, `lodash/flatten` can be replaced with `Array.prototype.flat()` and even combined with `Array.prototype.map()` by using `Array.prototype.flatMap()`.

Example:

https://github.com/avajs/ava/blob/e8a497696a7c0f46d92c52770dbc34b70ee493fc/lib/line-numbers.js#L52-L58

```diff
 export function getApplicableLineNumbers(normalizedFilePath, filter) {
-	return sortNumbersAscending(distinctArray(flatten(
+	return sortNumbersAscending(distinctArray(
 		filter
 			.filter(({pattern, lineNumbers}) => lineNumbers && picomatch.isMatch(normalizedFilePath, pattern))
-			.map(({lineNumbers}) => lineNumbers),
-	)));
+			.flatMap(({lineNumbers}) => lineNumbers),
+	));
 }
````

Would this be something that is of interest? Eventually being able to [remove Lodash would save ~1.3 MB](https://packagephobia.com/result?p=lodash) (over 2.2 MB on my disk)

[Here's a good resource](https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#you-dont-may-not-need-lodashunderscore)

novemberborn:
Yes that'd be great.

live627:
That's good to hear . I wasn't sure of the reaction especially after seeing [this comment](https://github.com/avajs/ava/issues/2548) and how it was closed for being "too controversial", whatever that means. Maybe it was the aggressive attitude. Anyway, I agree with the sentiment that this packages should have a smaller footprint and less dependencies if at all possible.

PR #2851, opened 2021-09-29
Replace Lodash with built-in syntax, libraries, and some code

live627:
The recursive cloning methods were retained as their individual npm packages

- Closes https://github.com/avajs/ava/issues/2848


novemberborn:
Thanks @live627! Interestingly, we previously switched to the main lodash package on purpose: https://github.com/avajs/ava/pull/2148

However I think we can remove it entirely, I'm about to push some commits let me know what you think.

live627:
Nifty. I'm  just here to remove dependency bloat 🙂

I also have a similar pull request for https://github.com/concordancejs/concordance/pull/78

PR #2858, opened 2021-10-04
Attempt to use TS 4.5 type resolution

novemberborn:
Per https://devblogs.microsoft.com/typescript/announcing-typescript-4-5-beta/#esm-nodejs this tries to tell TypeScript where to find the type definitions through the `exports` field. It doesn't seem to work though with `ava/plugin`, so I'll leave this as a draft for now.


novemberborn:
Also, sadly, it doesn't let us provide different type definitions based on the import mechanism, so it doesn't let us remove https://github.com/avajs/ava/pull/2836.

novemberborn:
TypeScript has removed these features from 4.5.

PR #2863, opened 2021-10-05
Remove `read-pkg` package from dependencies

alessandroasm:
As discussed on Issue #2863, this package was only used in one place and just to read a single property on a package.json file.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -1,4 +1,5 @@
 
+import fs from 'node:fs';
 import path from 'node:path';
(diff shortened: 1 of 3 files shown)


alessandroasm:
@live627 Please let me know if I need to change something :)

live627:
oh  didn' t realize that you were expecting a reply

Your  changeset looks good to me 😃

novemberborn:
Thanks @alessandroasm!

PR #2878, opened 2021-10-28
Update Mongoose recipe

hamidbae:
Continue previous bad pr, I really sorry about that. Actually this is the changes that I mean before.

My problem is cannot get mongo uri with your current readme on https://github.com/avajs/ava/blob/main/docs/recipes/endpoint-testing-with-mongoose.md.
After doing some search I found update rules to get mongo uri but no update on readme.
So I made this update to make it clear.
Also if I am mistaken, please don't hesitate to reach me out or just ignore this pr ya. Thanks:).

### DIFF
--- a/docs/recipes/endpoint-testing-with-mongoose.md
+++ b/docs/recipes/endpoint-testing-with-mongoose.md
@@ -52,12 +52,12 @@ Next start the in-memory MongoDB instance and connect to Mongoose:
 
 ```js
 // Start MongoDB instance
-const mongod = new MongoMemoryServer()
+const mongod = await MongoMemoryServer.create();
 
 // Create connection to Mongoose before tests are run
 test.before(async () => {
-	const uri = await mongod.getUri();
-	await mongoose.connect(uri, {useMongoClient: true});
+	const uri = mongod.getUri();
+	await mongoose.connect(uri);
 });
 ```
 
--- a/docs/recipes/endpoint-testing-with-mongoose.md
+++ b/docs/recipes/endpoint-testing-with-mongoose.md
@@ -51,11 +51,13 @@ const User = require('../models/User');
 Next start the in-memory MongoDB instance and connect to Mongoose:
 
 ```js
-// Start MongoDB instance
-const mongod = await MongoMemoryServer.create();
+// initiate MongoDB instance
+let mongod;
 
 // Create connection to Mongoose before tests are run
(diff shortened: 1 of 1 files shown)


novemberborn:
> Continue previous bad pr, I really sorry about that. Actually this is the changes that I mean before.

For future reference you don't need to close the PR, you can push more changes to your branch and they'll show up here.

---

With your changes, the recipe now assumes that top-level await is supported. I don't think we can make that assumption. I think you may need to create the server within the `test.before()` hook?

hamidbae:
Oh yes it is, I updated my code.
and for reference, I found this solution on error message while trying to use your recepe, [solution-link](https://nodkz.github.io/mongodb-memory-server/docs/guides/migrate7/#no-function-other-than-start-create-ensureinstance-will-be-starting-anything).

novemberborn (approved):
Cheers @hamidbae

docs/recipes/endpoint-testing-with-mongoose.md
```diff
@@ -51,13 +51,12 @@ const User = require('../models/User');
...
-
 // Create connection to Mongoose before tests are run
-test.before(async () => {
-	const uri = await mongod.getUri();
-	await mongoose.connect(uri, {useMongoClient: true});
+test.before(async t => {
+	// First start MongoDB instance
+	t.context.mongod = await MongoMemoryServer.create();
+	// And connect
+	await mongoose.connect(mongod.getUri());
```
hamidbae, line 59: I got error when directly call mongod to get Uri. I think it should be ```t.context.mongod.getUri()``` ?

docs/recipes/endpoint-testing-with-mongoose.md
```diff
@@ -117,8 +116,8 @@ Finally disconnect from and stop MongoDB when all tests are done:
 
 ```js
 test.after.always(async () => {
```
hamidbae, line 118: Also here I think I should add ```t``` as parameter ?

hamidbae (commented):
Call ```t.context.mongod``` instead directly call ```mongod``` and ```test.after.always``` hook missing ```t``` param ?

novemberborn:
Well spotted, thanks @hamidbae. Fixed in dc405ef3b76296f2eb2b6a1130a98fd4e889a978.

PR #2894, opened 2021-11-19
Handle path sources

### LINKED ISSUE #2889: Source maps on Windows with Node.js 14.8.1 and ts-node/esm loader return sources that are not file URLs, which causes a crash in the snapshot manager
I made [this minimal example](https://github.com/antonyfaris/sample-ava-ts).

You can see in [this gh workflow](https://github.com/antonyfaris/sample-ava-ts/runs/4119306273?check_suite_focus=true) how it only fails on windows. This error prevents tests from running.

It works fine with version 4.0.0-alpha.2

mcecode:
Fixes #2889

### DIFF
--- a/lib/snapshot-manager.js
+++ b/lib/snapshot-manager.js
@@ -398,9 +398,9 @@ const resolveSourceFile = mem(file => {
 		return file;
 	}
 
-	return payload.sources[0].startsWith('file://') ?
-		fileURLToPath(payload.sources[0]) :
-		payload.sources[0];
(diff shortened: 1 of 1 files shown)


sindresorhus:
The pull request title should describe *what* it fixes, not how ;)

mcecode:
Thanks for the feedback and sorry for the late response. I'll be more cognizant of my pull request titles from now on, just tell me if there's anything else I need to fix 😀

novemberborn:
Thanks @mcecode!

PR #2906, opened 2021-12-06
trackingDependancies config option set to false to disable dependency…

Sarveshk76:
… tracking.


novemberborn:
I'd suggest tracing how options make it from `lib/cli.js` into the `worker/base.js`. And then you need to make the installation dependent on the option.

novemberborn:

Issue #2593, opened 2020-10-03
Update TypeScript recipe for ESM support


novemberborn:
@FallingSnow [shared](https://github.com/avajs/ava/issues/2347#issuecomment-698499005) how to (experimentally) configure AVA and Node.js so that AVA can load TypeScript-based ESM files:

```json
"ava": {
    "extensions": {
      "ts": "module"
    },
    "nonSemVerExperiments": {
      "configurableModuleFormat": true
    },
    "nodeArguments": [
      "--loader=ts-node/esm",
      "--experimental-specifier-resolution=node"
    ],
    "files": [
      "test/**/*.spec.ts"
    ]
}
```

It'd be great to add this to our [TypeScript recipe](https://github.com/avajs/ava/blob/master/docs/recipes/typescript.md).

thasmo:
Trying to run TypeScript tests with ESM support, but failed to do so with the configuration above, adapted to my setup.
```
export default {
    nonSemVerExperiments: {
        nextGenConfig: true,
        configurableModuleFormat: true,
    },
    nodeArguments: [
        '--loader=ts-node/esm',
    ],
    extensions: {
        ts: 'module',
    },
    require: [
        'ts-node/register/transpile-only',
    ],
    files: [
        'packages/*/test/**/*.ts',
    ],
}
```

I'm running it with Node 14 and TypeScript 4, which results in this errors:
```
➜ npm test

> test
> ava

⚠ Experiments are enabled. These are unsupported and may change or be removed at any time.

(node:13990) ExperimentalWarning: --experimental-loader is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
(node:13996) ExperimentalWarning: --experimental-loader is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
(node:13997) ExperimentalWarning: --experimental-loader is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
(node:14010) ExperimentalWarning: --experimental-loader is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
(node:14004) ExperimentalWarning: --experimental-loader is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
(node:14025) ExperimentalWarning: --experimental-loader is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
[...]

PR #2910, opened 2021-12-14
Mention experimental specifier resolution in TypeScript recipe

rrichardson:
Using the nodeArgument `--experimental-specifier-resolution=node` makes it so one you don't have to add the `.js` extension to files when importing for your tests. (Such things are not feasible for people with large typescript workspaces)

This PR just updates the typescript recipe to suggest such a change.

### DIFF
--- a/docs/recipes/typescript.md
+++ b/docs/recipes/typescript.md
@@ -39,7 +39,8 @@ If your `package.json` has `"type": "module"`, then this is the AVA configuratio
 			"configurableModuleFormat": true
 		},
 		"nodeArguments": [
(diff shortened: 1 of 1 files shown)


rrichardson:
fixes #2593

novemberborn:
Thanks for the PR @rrichardson! I'd like to make some edits (nothing worth a back-and-forth over), hopefully I'll get to that over my break at the end of the year.

novemberborn:
I've edited this so we stick a little closer to the default behavior, but still allow people to discover these experiments.

PR #2911, opened 2021-12-15
Remove is-error dependency

drazisil:
👋 

This isn't needed at all, but seemed like an easy edit to reduce dependencies.

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -1,3 +1,5 @@
+import util from 'node:util';
+
 import concordance from 'concordance';
 import isPromise from 'is-promise';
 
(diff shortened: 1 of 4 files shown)


novemberborn:
@sindresorhus is there any nuance with `is-error` that isn't captured here?

sindresorhus:
`instanceof` will fail if the error is coming from another realm (iframe, Node.js's VM module, etc).

You could alternatively inline `Object.prototype.toString.call(error) === '[object Error]'`.

novemberborn:
Is that a scenario we should still care about?

novemberborn:
Can we use https://nodejs.org/api/util.html#utiltypesisnativeerrorvalue?

novemberborn:
`isNativeError()` didn't work, even in a situation where `instanceof Error` returned `true. The documentation is quite sparse so I'm not sure what it's meant to do!

However I don't want to be in a situation where `t.throws()` fails because it's not detecting an `Error` instance even though when we format that object it very much looks like an `Error`. We should handle all (actual) errors that Node.js can throw at us even if from a different realm. That's what `is-error` provides.

PR #2918, opened 2021-12-22
Remove dependency on `equal-length`

lightmare:
It's tiny, but still superfluous for a single point of use.


lightmare:
Sorry, accidentally included an unrelated change in the commit. The 2nd one should be clean.

novemberborn:
Thanks!

PR #2950, opened 2022-01-22
Refer to assertion inputs as "actual"

### LINKED ISSUE #2938: Assertion parameter names are inconsistent between documentation and declaration ("value" vs. "actual")
Documentation for the `is` assertion uses parameter names `value` and `expected`: https://github.com/avajs/ava/blob/6b63b1bc14a1aaa7d4ee4f2201f901624419c677/docs/03-assertions.md#L124

Declaration for the same assertion uses parameter names `actual` and `expected`: https://github.com/avajs/ava/blob/6b63b1bc14a1aaa7d4ee4f2201f901624419c677/types/assertions.d.ts#L190

This is inconsistent. I think actual/expected is more clear and more common, so suggest standardizing on that pairing.

live627:
Fix #2938

### DIFF
--- a/docs/03-assertions.md
+++ b/docs/03-assertions.md
@@ -101,51 +101,51 @@ Passing assertion. Returns a boolean indicating whether the assertion passed.
 
 Failing assertion. Returns a boolean indicating whether the assertion passed.
 
-### `.assert(value, message?)`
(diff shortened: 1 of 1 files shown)


novemberborn:
Thanks @live627!

PR #2954, opened 2022-01-24
Update examples in docs to use ES Modules

scottdotjs:
I'm trying AVA for the first time and this appears to be the case as of 4.0.

### DIFF
--- a/docs/01-writing-tests.md
+++ b/docs/01-writing-tests.md
@@ -260,7 +260,7 @@ Available properties:
 * `snapshotDirectory`: directory where snapshots are stored, as a file URL string
 
 ```js
-const test = require('ava');
+import test from 'ava';
 
 console.log('Test file currently being run:', test.meta.file);
 ```
--- a/docs/01-writing-tests.md
+++ b/docs/01-writing-tests.md
@@ -21,7 +21,7 @@ To declare a test you call the `test` function you imported from AVA. Provide th
 **Note:** In order for the [enhanced assertion messages](./03-assertions.md#enhanced-assertion-messages) to behave correctly, the first argument **must** be named `t`.
 
 ```js
-const test = require('ava');
+import test from 'ava';
 
 test('my passing test', t => {
 	t.pass();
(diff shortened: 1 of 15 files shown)


novemberborn:
Either work, it depends on whether you're using `.mjs` files or have set `"type": "module"` in your `package.json`. There's a lot of other require statements in the documentation, we should make them consistent.

I do lean towards the ESM syntax, but then I wonder if that'll trip people up who have a default setup of CommonJS. @sindresorhus what do you think?

sindresorhus:
I would use ESM syntax.

novemberborn:
@scottdotjs would you be up for converting all the docs? I'd understand if not — but it'd be good to keep the syntax consistent so I'd rather do it all in one go.

scottdotjs:
Sure! This PR only covered the first example as that was as far as I'd for at the time. 🙂

scottdotjs:
OK, I've changed the examples to use ES Modules as far as was obvious to me - in the case of other packages, I checked that it worked for them by consulting their issues if it wasn't shown in their documentation, and/or tried it myself. There were several instances I didn't change, however:

[06-configuration.md](https://github.com/avajs/ava/blob/main/docs/06-configuration.md)
```
You can require() dependencies.
[...]
const baseConfig = require('./ava.config.cjs');
```

JSON imports aren't quite there yet (as I'm sure you know) so it's your call on how to replace this if you want to. There's another JSON file being required in `recipes\when-to-use-plan.md`.

[testing-with-selenium-webdriverjs.md](https://github.com/avajs/ava/blob/main/docs/recipes/testing-with-selenium-webdriverjs.md)
```
require('chromedriver');
```

I've never used `require()` on its own like this so not sure what the `import` equivalent would be.

[vue.md](https://github.com/avajs/ava/blob/main/docs/recipes/vue.md)
```
const hooks = require('require-extension-hooks');
const Vue = require('vue');
const Component = require('component.vue');
```

I have no Vue experience either so couldn't update this section; also the Vue basic examples don't show Vue being used this way but rather all as `import { createApp } from 'vue'`, so I don't know if your example is out of date or doing something different.

Issue #2945, opened 2022-01-13
"nonSemVerExperiments.configurableModuleFormat from undefined is not a supported experiment"


broofa:
'Trying to run TS tests, following the instructions at https://github.com/avajs/ava/blob/main/docs/recipes/typescript.md#for-packages-with-type-module.  When I run `ava` I get the titular error.

'Not really sure what I'm doing wrong here.  Judging by the code, it seems like issue is with my configuration rather than any of the tests, so I haven't bothered including those here.

Here's my package.json (no "ava.config.*" file):
```json
{
  "type": "module",
  "scripts": {
    "test": "ava",
    "sh": "docker run --entrypoint \"sh\" -it cplambda-sass"
  },
  "dependencies": {
    "@broofa/asyncproxy": "^1.0.4",
    "@types/sass": "^1.43.1",
    "sass": "^1.47.0"
  },
  "devDependencies": {
    "@ava/typescript": "3.0.1",
    "ava": "^4.0.1",
    "esbuild-node-loader": "^0.6.3",
    "typescript": "4.5.4"
  },
  "ava": {
    "extensions": {
      "js": true,
      "ts": "module"
    },
    "nonSemVerExperiments": {
      "configurableModuleFormat": true
    },
    "nodeArguments": [
      "--loader=esbuild-node-loader",
      "--experimental-specifier-resolution=node"
    ],
    "files": [
      "test/**/*.spec.ts"
    ]
  }
}
```

... and my environment:
```
$ envinfo
npx: installed 1 in 0.897s

  System:
    OS: macOS 12.0.1
    CPU: (8) x64 Apple M1
    Memory: 35.51 MB / 16.00 GB
    Shell: 5.1.8 - /usr/local/bin/bash
  Binaries:
    Node: 14.17.6 - ~/.nvm/versions/node/v14.17.6/bin/node
    Yarn: 1.22.11 - ~/.nvm/versions/node/v14.17.6/bin/yarn
    npm: 6.14.15 - ~/.nvm/versions/node/v14.17.6/bin/npm
    Watchman: 2021.10.18.00 - /usr/local/bin/watchman
  Browsers:
    Chrome: 97.0.4692.71
    Firefox: 95.0.1
    Safari: 15.1
  npmPackages:
    @ava/typescript: 3.0.1 => 3.0.1
    @broofa/asyncproxy: ^1.0.4 => 1.0.4
    @types/sass: ^1.43.1 => 1.43.1
[...]

PR #2958, opened 2022-01-30
Remove outdated AVA 3 config from TypeScript recipe

sikrinick:
https://github.com/avajs/ava/issues/2945
Fix according to this [comment](https://github.com/avajs/ava/issues/2945#issuecomment-1013024157)
> Nice find, that documentation is outdated. The feature is no longer experimental in AVA 4 so this needs to be removed:
> 
> ```
>     "nonSemVerExperiments": {
>       "configurableModuleFormat": true
>     },
> ```


novemberborn:
Lovely, thanks @sikrinick!

Issue #2957, opened 2022-01-28
Watch mode not respecting ignoredByWatcher


razor-x:
When running a test that saves files to `tmp/some-dir/some-file` AVA will loop forever in watch mode even if `ignoredByWatcher` is set. Same behaviour in AVA v3 and v4.

Minimal reproducible example: https://github.com/rxbugs/ava-watch-ignore

```
Arch Linux

node version    
v16.13.2

AVA version
4.0.1 or 3.15.0
```

```json
{
  "ignoredByWatcher": ["tmp/**/*"]
}
```

```json
{
  "type": "module",
  "scripts": {
    "test:watch": "ava -w",
    "test": "ava"
  },
  "devDependencies": {
    "ava": "^4.0.1",
    "mkdirp": "^1.0.4"
  }
}
```

```js
import path from 'path'
import { promises as fsPromises } from 'fs'

import test from 'ava'
import mkdirp from 'mkdirp'

test.beforeEach(async (t) => {
  const root = 'tmp'
  await mkdirp(root)
  t.context.dst = await fsPromises.mkdtemp(path.join(root, 'foo-'))
})

test('bug', async (t) => {
  await fsPromises.writeFile(path.join(t.context.dst, 'foo'), 'bar')
  t.pass()
})
```

novemberborn:
Your configuration isn't being picked up. AVA only support JavaScript configuration files, so you'd want something like `ava.config.js`:

```js
{
  "ignoredByWatcher": ["tmp/**/*"]
}
```

razor-x:
> Your configuration isn't being picked up. AVA only support JavaScript configuration files, so you'd want something like `ava.config.js`:
> 
> ```js
> export default {
>   "ignoredByWatcher": ["tmp/**/*"]
> }
> ```


🤦🏻 Thank you. I feel I can't be the only one who missed this change. Can I suggest a waring like this? https://github.com/avajs/ava/pull/2962

PR #2962, opened 2022-02-06
Warn when unsupported ava.config.json files are encountered

razor-x:
See https://github.com/avajs/ava/issues/2957#issuecomment-1030884141


novemberborn:
Thanks for this @razor-x. `ava.config.json` files have actually never worked, it's not a v4 thing.

I'm thinking we shouldn't print to the console in `load-config` itself. Perhaps we could return an array of existing `ava.config.json` paths that were encountered while loading a configuration file, and then do the logging from `lib/cli.js`?

novemberborn:
Returning it makes it easier to print a nice list of encountered files, that's harder to do with a callback.

razor-x:
I looked at updating the return value, but that seemed kinda stressful as it's a breaking change for that function and would require updating its usage in all tests and other files. I ended up opting for the callback instead since it's a backwards compatible change. Hopefully this is ok with you. It still produces a nice list since it calls the function once with an array.

I ran a quick local test on the cli to check the warning. This is what it looks like.

novemberborn:
Cheers @razor-x. I haven't had a chance to look at this yet though.

razor-x:
No worries. I fixed the tests.

razor-x:
> @razor-x could you allow me to push to your branch? I've got some commits ready that make this return the unsupported files rather than using the callback but I don't seem to be able to push them to the PR.

Done. Looks like GH does not support that option for org-owned forks, so I just added you to repo.

novemberborn:
@razor-x it should be this checkbox on the right: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

razor-x:
image
image

novemberborn:
Strange, it's normally not an issue.

razor-x:
> Strange, it's normally not an issue.

I think it's this: https://github.com/isaacs/github/issues/1681 I got into the habit of putting all my forks on a separate org so they don't clutter up my main account.

> @razor-x what do you think of the changes?

All good, thanks.

novemberborn:
Lovely, thanks @razor-x!

Issue #2965, opened 2022-02-09
feature: allow custom comparator for parallelRuns


erezrokah:
Please provide details about:

* What you're trying to do

  Follow up to https://github.com/avajs/ava/issues/1811. Better load balance test files across multiple CI machines 

* Why you can't use AVA for this

  Ava sorts files by file name, which is not the best load balancing strategy, especially when some tests are slower than others.

* And maybe how you think AVA could handle this

  Allow setting a custom comparator to be used here:
https://github.com/avajs/ava/blob/3131ccd9f04a2cc521bc50b3533137e91f34b335/lib/api.js#L180

**Alternatives**:
Ensure number of tests per file is low enough via a lint rule.

**More context**:
https://github.com/netlify/cli/issues/4178

novemberborn:
What would you think about being able to configure the comparator function in an `ava.config.js` file?

erezrokah:
> What would you think about being able to configure the comparator function in an `ava.config.js` file?

That would be great and will solve our use can 💯

Are you open for me to contribute this change or prefer someone else to do it?

novemberborn:
All yours @erezrokah! Let me know if you have questions.

PR #2968, opened 2022-02-11
feat(parallel-runs): allow custom comparator

erezrokah:
Fixes https://github.com/avajs/ava/issues/2965

Still in draft to get feedback before adding docs.

Please note I had to disable a linting rule to allow deserialization of the comparator function

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -382,9 +382,20 @@ export default async function loadCli() { // eslint-disable-line complexity
 	}
 
 	let parallelRuns = null;
+	let parallelRunsComparator = null;
 	if (isCi && ciParallelVars) {
 		const {index: currentIndex, total: totalRuns} = ciParallelVars;
 		parallelRuns = {currentIndex, totalRuns};
+	}
+
+	if (parallelRuns) {
+		if (Reflect.has(conf, 'ciParallelRunsComparator') && typeof conf.ciParallelRunsComparator !== 'function') {
+			exit('ciParallelRunsComparator must be a comparator function.');
+		}
+
+		const defaultComparator = (a, b) => a.localeCompare(b, [], {numeric: true});
+		// The function needs to be serializable to support worker threads
+		parallelRunsComparator = (conf.ciParallelRunsComparator || defaultComparator).toString();
 	}
 
 	const match = combined.match === '' ? [] : arrify(combined.match);
@@ -413,6 +424,7 @@ export default async function loadCli() { // eslint-disable-line complexity
 		moduleTypes,
 		nodeArguments,
 		parallelRuns,
+		parallelRunsComparator,
 		projectDir,
 		providers,
 		ranFromCli: true,
--- a/lib/api.js
+++ b/lib/api.js
@@ -177,7 +177,10 @@ export default class Api extends Emittery {
 				const fileCount = selectedFiles.length;
 
 				// The files must be in the same order across all runs, so sort them.
-				selectedFiles = selectedFiles.sort((a, b) => a.localeCompare(b, [], {numeric: true}));
+				// The sorting function is a string representation of the function, so we need to deserialize it
+				// eslint-disable-next-line no-new-func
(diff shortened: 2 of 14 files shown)


lib/cli.js
```diff
@@ -382,11 +382,22 @@ export default async function loadCli() { // eslint-disable-line complexity
...
 		parallelRuns = {currentIndex, totalRuns};
 	}
 
+	if (parallelRuns) {
+		if (Reflect.has(conf, 'ciParallelRunsComparator') && typeof conf.ciParallelRunsComparator !== 'function') {
+			exit('ciParallelRunsComparator must be a comparator function.');
+		}
+
+		const defaultComparator = (a, b) => a.localeCompare(b, [], {numeric: true});
+		// The function needs to be serializable to support worker threads
```
novemberborn, line 397: It doesn't, `lib/api.js` runs in this process and the function is not needed in the worker threads. If there is a serialization error it needs to be stripped out instead.

lib/cli.js
```diff
@@ -382,11 +382,22 @@ export default async function loadCli() { // eslint-disable-line complexity
...
 
 	let parallelRuns = null;
+	let parallelRunsComparator = null;
 	if (isCi && ciParallelVars) {
 		const {index: currentIndex, total: totalRuns} = ciParallelVars;
 		parallelRuns = {currentIndex, totalRuns};
 	}
 
+	if (parallelRuns) {
+		if (Reflect.has(conf, 'ciParallelRunsComparator') && typeof conf.ciParallelRunsComparator !== 'function') {
```
novemberborn, line 392: We could make this more generic still, say `sortTestFiles`. Would let you randomize execution order as well (or, if we decide to randomize, it provides a way for others to do a deterministic sort).

novemberborn (commented):
Thanks!

lib/cli.js
erezrokah: Ah, thanks for the tip! Fixed in https://github.com/avajs/ava/pull/2968/commits/07c3eff3a3c6c6857992b5c47942ebef8a8e034c

lib/cli.js
novemberborn: > Do you mean generic for all runs, or only CI runs?
> With caching and running failed tests first, a sorter might be confusing for local runs.
> Happy to implement it, if that's the request.

For all runs, yea. I'm assuming the failure prioritization keeps the relative order of the prioritized files, I don't think that'll be too confusing. And otherwise you could detect CI in the configuration and act accordingly.

> Also, can you confirm my understanding that currently tests files are ordered based on the return value of `globby`?

I suppose, can't quite recall if there's multiple concatenations or whatnot. But the current order is not defined. I suppose that means we shouldn't set a default comparator and only sort when one is provided.

erezrokah: Done in https://github.com/avajs/ava/pull/2968/commits/9836cd7ae8b6083a2c0b4a2dc289809d5b10f6dd!

Going to add a separate test for non CI runs

erezrokah: Added another test in https://github.com/avajs/ava/pull/2968/commits/44cdb863c1efdde52e4fec3fcc62fc1c6dced5b2.
Please let me know if I put it in the right place.

I used `concurrency: 1` to make the test output predictable.
With AVA parallelism I'm not sure sorting can guarantee order of execution otherwise, WDYT?

novemberborn:
Sorry @erezrokah I haven't had a chance yet to look into this.

erezrokah:
Let me know if I should look at the fail test. It seems to be failing on the default branch too https://github.com/avajs/ava/runs/5285903957?check_suite_focus=true

novemberborn:
@erezrokah nah, it's probably some transitive dependency that has changed and is causing the tests to fail. It won't be because of your code.

novemberborn (approved):
This is great, thanks @erezrokah!

erezrokah:
>This is great, thanks @erezrokah!

Thank you for the review, appreciate the feedback ❤️

PR #2969, opened 2022-02-14
Loosen type constraints for deepEqual assertion

### LINKED ISSUE #2964: Support `Actual extends Expected` in conjunction with `Expected extends Actual`
### The Problem

The type `DeepEqualAssertion` requires that the `Expected` type extends the `Actual` type.
However, sometimes we want this relationship to be the other way around.

Consider the following example:

```ts
test("my test", (t) => {
  const expected = { foo: [1, 2, 3] } as const;
  const actual = functionToTest(1, 2, 3);

  t.deepEqual(actual, expected);
});
```

If the type of `actual` is inferred as `{ foo: number[] }` then this does not extend the type of `expected` which is `{ readonly foo: readonly number[] }`. Thus TypeScript reports an error.

### Suggested Fix

Add a overload to `DeepEqualAssertion` to handle this case:

```ts
export interface DeepEqualAssertion {
	/**
	 * Assert that `actual` is [deeply equal](https://github.com/concordancejs/concordance#comparison-details) to
	 * `expected`, returning a boolean indicating whether the assertion passed.
	 */
	<Actual, Expected extends Actual>(actual: Actual, expected: Expected, message?: string): actual is Expected;

	/**
	 * Assert that `actual` is [deeply equal](https://github.com/concordancejs/concordance#comparison-details) to
	 * `expected`, returning a boolean indicating whether the assertion passed.
	 */
	<Actual extends Expected, Expected>(actual: Actual, expected: Expected, message?: string): expected is Actual;

	/** Skip this assertion. */
	skip(actual: any, expected: any, message?: string): void;
}
```

RebeccaStevens:
fix #2964

### DIFF
--- a/test-d/deep-equal.ts
+++ b/test-d/deep-equal.ts
@@ -1,0 +1,30 @@
+import {expectType} from 'tsd';
+
+import test from '..';
+
+test('actual extends expected', t => {
+	type Expected = {foo: [1, 2, 3]};
+	const expected: Expected = {foo: [1, 2, 3]};
+	const actual = {foo: [1, 2, 3]};
+	if (t.deepEqual(actual, expected)) {
+		expectType<Expected>(actual);
+	}
+});
+
+test('expected extends actual', t => {
+	type Expected = {foo: Array<number | string>};
+	type Actual = {foo: number[]};
+	const expected: Expected = {foo: [1, 2, 3]};
+	const actual: Actual = {foo: [1, 2, 3]};
+	if (t.deepEqual(actual, expected)) {
+		expectType<Actual>(expected);
+	}
+});
+
+test('neither extends the each other', t => {
+	type Expected = {readonly foo: readonly [1, 2, 3]};
+	type Actual = {foo: number[]};
+	const expected: Expected = {foo: [1, 2, 3]};
+	const actual: Actual = {foo: [1, 2, 3]};
+	t.deepEqual(actual, expected);
+});
--- a/types/assertions.d.ts
+++ b/types/assertions.d.ts
@@ -138,6 +138,18 @@ export interface DeepEqualAssertion {
 	 */
 	<Actual, Expected extends Actual>(actual: Actual, expected: Expected, message?: string): actual is Expected;
 
+	/**
+	 * Assert that `actual` is [deeply equal](https://github.com/concordancejs/concordance#comparison-details) to
+	 * `expected`, returning a boolean indicating whether the assertion passed.
+	 */
+	<Actual extends Expected, Expected>(actual: Actual, expected: Expected, message?: string): expected is Actual;
+
+	/**
+	 * Assert that `actual` is [deeply equal](https://github.com/concordancejs/concordance#comparison-details) to
+	 * `expected`, returning a boolean indicating whether the assertion passed.
+	 */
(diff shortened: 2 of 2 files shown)


novemberborn:
Cheers @RebeccaStevens. Could you add a test for this under https://github.com/avajs/ava/tree/main/test-d?

types/assertions.d.ts
```diff
@@ -137,6 +137,12 @@ export interface DeepEqualAssertion {
 	 * `expected`, returning a boolean indicating whether the assertion passed.
 	 */
 	<Actual, Expected extends Actual>(actual: Actual, expected: Expected, message?: string): actual is Expected;
+	
+	/**
+	 * Assert that `actual` is [deeply equal](https://github.com/concordancejs/concordance#comparison-details) to
+	 * `expected`, returning a boolean indicating whether the assertion passed.
+	 */
+	<Actual extends Expected, Expected>(actual: Actual, expected: Expected, message?: string): expected is Actual;
```
RebeccaStevens, line 145: Actually this might not want to be a type guards. It might just simply want to return a boolean.

RebeccaStevens:
> Cheers @RebeccaStevens. Could you add a test for this under [`main`/test-d](https://github.com/avajs/ava/tree/main/test-d)?

Sure thing. Give me a few minutes.

types/assertions.d.ts
RebeccaStevens: I'll figure out this when writing the tests.

RebeccaStevens:
@novemberborn Done

test-d/deep-equal.ts
```diff
@@ -0,0 +1,32 @@
...
+});
+
+test('neither exectend each other', t => {
+	type Expected = {readonly foo: readonly [1, 2, 3]};
+	type Actual = {foo: number[]};
+	const expected: Expected = {foo: [1, 2, 3]};
+	const actual: Actual = {foo: [1, 2, 3]};
+	if (t.deepEqual(actual, expected)) {
+		t.pass();
+	}
```
novemberborn, line 31: The `t.pass()` isn't necessary.

```suggestion
	t.deepEqual(actual, expected);
```

types/assertions.d.ts
```diff
@@ -138,6 +138,18 @@ export interface DeepEqualAssertion {
...
+
+	/**
+	 * Assert that `actual` is [deeply equal](https://github.com/concordancejs/concordance#comparison-details) to
+	 * `expected`, returning a boolean indicating whether the assertion passed.
+	 */
+	<Actual, Expected>(actual: Actual, expected: Expected, message?: string): boolean;
```
novemberborn, line 151: Why do we need this overload? Wouldn't the assertion always fail if neither extend one another?

RebeccaStevens: If part of `Expected` extended part of `Actual`, but another part `Expected` is a super type of `Actual`, then neither extends the other but both my have the same value.

For example, imagine that one of the properties on `Expected` is typed as readonly, while the corresponding property on `Actual` isn't. And then for another property the two share, the readonlyness is declared the other way around.

novemberborn:
Thanks @RebeccaStevens!

PR #2971, opened 2022-02-17
resolveCwd doesn't work when use ava from an npm linked project

ailisp:
A common senario:
Project A depends on ava. Project B depends on project A. In Project A, call `npm link`. In project B, call `npm link ProjectA`. Now in project B tests are broken because resolveCwd cannot find ts-node/register:
```
Error: Could not resolve required module ’ts-node/register’
```

However, nodejs's `require` can find it, and also `require.resolve`. And use `require.resolve` doesn't break in original setup of ava so i use `require.resolve`.

Also with `console.log` I noticed that this `resolveModules` only tries to resolve one module: `ts-node/register`. What is the purpose of this method?

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -9,7 +9,6 @@ import commonPathPrefix from 'common-path-prefix';
 import Emittery from 'emittery';
 import ms from 'ms';
 import pMap from 'p-map';
-import resolveCwd from 'resolve-cwd';
 import tempDir from 'temp-dir';
 
 import fork from './fork.js';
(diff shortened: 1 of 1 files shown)


novemberborn:
I think it's meant to resolve absolute paths to modules provided in AVA's `require` config. But I don't know whether that's still relevant, or whether this is the best way of going about it. I'll need to think about this.

novemberborn:
I think we should resolve relative to the project directory, likely we're trying to approximate that with `resolve-cwd` but the CWD doesn't have to be the project directory. I'd classify this as a bug, actually.

This is needed so that we import the expected module, without say AVA's dependencies shadowing it.

We may be able to use https://nodejs.org/api/module.html#modulecreaterequirefilename with the project directory and use the `resolve()` method on that.

But I don't know if that'll fix your issue.

novemberborn:
Tracking in https://github.com/avajs/ava/issues/2985.

PR #2975, opened 2022-02-24
Remove accidentally committed test-ava symlink

gibson042:


novemberborn:
Why?

gibson042:
Because its target is `/Users/jsamines/dev/personal/ava/node_modules/.bin/test-ava`, which is not a path that exists on my host and probably exists _only_ for @jonathansamines: https://github.com/avajs/ava/blame/01bc445a3856838539869d5a34bab5c9b043cd0b/test/node_modules/.bin/test-ava

novemberborn:
Nice spot, thanks!

Issue #2983, opened 2022-03-05
WiP: chor(typescript4.6): Add full Typed Ava via JSDOC Annotations and implement Typescript 4.6


frank-dspeed:
Please provide details about:

* What you're trying to do
I Will upgrade ava as it is used in Projects where i am Stakeholder as i upgraded more and more of the rollup ecosystem i regonized that many tests are not fully typed because of ava and typescript now added the mjs and cjs support so i will upgrade your full project structure to support full d.ts generation and all new moduleResolve algos that are supported by Typescript 4.6
* Why you can't use AVA for this
* And maybe how you think AVA could handle this

everything should be chor i try to not break anything but as the project is at present total diffrent structured i am not sure please test as soon as you can and maybe support feedback

i keep that open as WiP (Work In Progress) till it is ready to get merged.

frank-dspeed:
This should get labled PR i will now publish the first changes and link the issues but you should not depend on that as there are also typescript upstream errors maybe that need to get fixed there before

PR #2984, opened 2022-03-05
WiP: Static Analyzeable Typescript 4.6 Support Closes: #2983

frank-dspeed:
# References
- https://github.com/avajs/ava/issues/2983

This PR Gets more details as soon as it makes more progress its a ton of work

### DIFF
--- a/entrypoints/main.mjs
+++ b/entrypoints/main.mjs
@@ -1,2 +1,3 @@
-import test from '../lib/worker/main.cjs';
+import { test } from '../lib/worker/main.cjs';
 export { test };
+export default test;
--- a/entrypoints/main.mjs
+++ b/entrypoints/main.mjs
@@ -1,1 +1,2 @@
-export {default} from '../lib/worker/main.cjs';
+import test from '../lib/worker/main.cjs';
+export { test };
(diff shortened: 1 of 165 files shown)


sindresorhus:
I honestly have no idea what this is trying to do. I would recommend trying to explain it in clearer terms in the issue before continuing work on this. In it's current form, we are unlikely to accept this. The only changes that look ok to me are adding `.js` to imports.

frank-dspeed:
@sindresorhus i need to change that more often thats why i call it iteration the new moduleResolution algos do not depend on each other for example when you choose "node" it will lookup package.json types fild when you choose node12 it will lookup exports: types: ""

when you use nodenext that also looks into exports: import: types: 

path maps like your using them are not supported but are WiP.

wait for the final result and take what ever you like if you do not like it we need that in rollup anyway.

i choosed your project because it is valid fresh ECMAScript code that is extensiv using the new extensions so i can workout all edge cases with the declarations and new extensions.

for example classic will resolve index.mjs index.d.mts

and so on also your mixed exports style is interristing edge case

novemberborn:
> i choosed your project because it is valid fresh ECMAScript code that is extensiv using the new extensions so i can workout all edge cases with the declarations and new extensions.

It sounds like you're prototyping something for `rollup`. Please use CI in your fork of AVA. When you're ready to share improvements with us here then please open a new PR.

PR #2995, opened 2022-03-26
Support message testing function in `throws` & `throwsAsync` assertions

### LINKED ISSUE #2978: Support function in `message` expectation in `t.throws` / `t.throwsAsync`
The most common way to use `t.throws` is to match against the error message. However, sometimes part of the error message varies by Node.js versions or operating systems. Currently, you would use regex matching for this. But with regex, you have to be careful to escape special symbols.

I suggest adding support for a function as the `message`. This would make it easy to:
- Check is a message contains a certain string
- Check if the error message is one of multiple messages

Examples:

```js
await t.throwsAsync(async () => {
	throw new TypeError('🦄');
}, {
	message: message => ['🦄', '🐴'].includes(message)
});

await t.throwsAsync(async () => {
	throw new TypeError(`${process.platform} is not supported`);
}, {
	message: message => message.includes('is not supported')
});
```

il3ven:
Fixes #2978 

As per the issue #2978, I have added support for function in `throws`/`throwsAsync` assertion. I have also updated the docs regarding the same. I couldn't find any tests to be updated so I have left them.

How to use?

```js
await t.throwsAsync(async () => {
	throw new TypeError('🦄');
}, {
	message: message => ['🦄', '🐴'].includes(message)
});
```

Note: This is my first PR in ava so I may have missed something. Please let me know. I will fix it. For example, I am not sure about the [assertion message](https://github.com/il3ven/ava/blob/cbb8d584b358ddbf55824d935c1a151d9c67ba1a/lib/assert.js#L238-L249) to show in case of failure.

### DIFF
--- a/docs/03-assertions.md
+++ b/docs/03-assertions.md
@@ -172,7 +172,10 @@ Assert that an error is thrown. `fn` must be a function which should throw. The 
 
 * `instanceOf`: a constructor, the thrown error must be an instance of
 * `is`: the thrown error must be strictly equal to `expectation.is`
-* `message`: either a *string*, which is compared against the thrown error's message, a *regular expression*, which is matched against this message or a *function* which is passed the thrown error message and must return a boolean. true makes the assertion pass
+* `message`: the following types are valid: 
+  * *string* - it is compared against the thrown error's message
+  * *regular expression* - it is matched against this message
+  * *function* - it is passed the thrown error message and must return a boolean for whether the assertion passed
 * `name`: the expected `.name` value of the thrown error
 * `code`: the expected `.code` value of the thrown error
 
@@ -204,7 +207,10 @@ The thrown value *must* be an error. It is returned so you can run more assertio
 
 * `instanceOf`: a constructor, the thrown error must be an instance of
 * `is`: the thrown error must be strictly equal to `expectation.is`
-* `message`: either a *string*, which is compared against the thrown error's message, a *regular expression*, which is matched against this message or a *function* which is passed the thrown error message and must return a boolean. true makes the assertion pass
+* `message`: the following types are valid: 
+  * *string* - it is compared against the thrown error's message
+  * *regular expression* - it is matched against this message
+  * *function* - it is passed the thrown error message and must return a boolean for whether the assertion passed
 * `name`: the expected `.name` value of the thrown error
 * `code`: the expected `.code` value of the thrown error
 
--- a/docs/03-assertions.md
+++ b/docs/03-assertions.md
@@ -172,7 +172,7 @@ Assert that an error is thrown. `fn` must be a function which should throw. The 
 
 * `instanceOf`: a constructor, the thrown error must be an instance of
 * `is`: the thrown error must be strictly equal to `expectation.is`
-* `message`: either a string, which is compared against the thrown error's message, or a regular expression, which is matched against this message
+* `message`: either a *string*, which is compared against the thrown error's message, a *regular expression*, which is matched against this message or a *function* which is passed the thrown error message and must return a boolean. true makes the assertion pass
 * `name`: the expected `.name` value of the thrown error
 * `code`: the expected `.code` value of the thrown error
 
@@ -204,7 +204,7 @@ The thrown value *must* be an error. It is returned so you can run more assertio
 
 * `instanceOf`: a constructor, the thrown error must be an instance of
 * `is`: the thrown error must be strictly equal to `expectation.is`
-* `message`: either a string, which is compared against the thrown error's message, or a regular expression, which is matched against this message
+* `message`: either a *string*, which is compared against the thrown error's message, a *regular expression*, which is matched against this message or a *function* which is passed the thrown error message and must return a boolean. true makes the assertion pass
 * `name`: the expected `.name` value of the thrown error
 * `code`: the expected `.code` value of the thrown error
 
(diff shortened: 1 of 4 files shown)


docs/03-assertions.md
```diff
@@ -172,7 +172,7 @@ Assert that an error is thrown. `fn` must be a function which should throw. The
 
 * `instanceOf`: a constructor, the thrown error must be an instance of
 * `is`: the thrown error must be strictly equal to `expectation.is`
-* `message`: either a string, which is compared against the thrown error's message, or a regular expression, which is matched against this message
+* `message`: either a *string*, which is compared against the thrown error's message, a *regular expression*, which is matched against this message or a *function* which is passed the thrown error message and must return a boolean. true makes the assertion pass
```
sindresorhus, line 175: I think this would be more readable if each accepted type would be on its own line.

sindresorhus, line 175:
```suggestion
* `message`: either a *string*, which is compared against the thrown error's message, a *regular expression*, which is matched against this message or a *function* which is passed the thrown error message and must return a boolean for whether the assertion passed.
```

lib/assert.js
```diff
@@ -106,7 +106,12 @@ function validateExpectations(assertion, expectations, numberArgs) { // eslint-d
 			});
 		}
 
-		if (hasOwnProperty(expectations, 'message') && typeof expectations.message !== 'string' && !(expectations.message instanceof RegExp)) {
+		if (
+			hasOwnProperty(expectations, 'message')
+			&& typeof expectations.message !== 'string'
+			&& !(expectations.message instanceof RegExp)
+			&& !(expectations.message instanceof Function)
```
sindresorhus, line 113: The best way to check for function is using `typeof`.

sindresorhus (changes requested):

sindresorhus:
You also need to update the TypeScript types.

types/assertions.d.ts
```diff
@@ -12,7 +12,7 @@ export type ThrowsExpectation = {
 	is?: Error;
 
 	/** The thrown error must have a message that equals the given string, or matches the regular expression. */
-	message?: string | RegExp;
+	message?: string | RegExp | Function;
```
sindresorhus, line 15: The type should be more specific than just `Function`.

il3ven: You are right. I didn't pay attention the first time. I have updated it now.

novemberborn:
@il3ven the code looks great, but it seems some tests are not passing.

il3ven:
@novemberborn I have fixed the test in `test-tap/assert.js`. The pipeline also passes now. However there is one test that fails in my local system. I could not debug the issue with that test. The same test also fails in the main branch. Does this also happen on your system? If no, then we can forget about the issue and move on.

**Steps to reproduce**
- Run `npx tap test-tap/api.js`
- Observe the assert at line#154 fail in case of `opts.workerThread == true`

<details>
<summary>Error output in my system</summary>

```
$ npx tap test-tap/api.js 

​ FAIL ​ test-tap/api.js
 ✖ should be equal

  --- expected    
  +++ actual      
  @@ -1,1 +1,1 @@ 
  -1              
  +2              

  test: "test-tap/api.js fail-fast mode - workerThreads: true - multiple files &
    interrupt"
  at:
    line: 154
    column: 5
    file: file:///Users/vaibhav/repos/il3ven/temp-ava/test-tap/api.js
    type: Test
  stack: |
    Test.<anonymous> (file://test-tap/api.js:154:5)

​ FAIL ​ test-tap/api.js 1 failed of 128 1m
 ✖ should be equal

​

                         
  🌈 SUMMARY RESULTS 🌈  
                         
​ FAIL ​ test-tap/api.js 1 failed of 128 1m
 ✖ should be equal

​
Suites:   ​1 failed​, ​1 of 1 completed​
Asserts:  ​​​1 failed​, ​127 passed​, ​of 128​
​Time:​   ​1m​
```

</details>

**My Machine**
OS: MacOS 12.3.1
Node: v14.18.2

novemberborn:
> However there is one test that fails in my local system. I could not debug the issue with that test.

Yea I think that test is flaky, no worries.

novemberborn:
@il3ven sorry I was AFK for nearly two weeks :) Test pass now, but it strikes me there's no test for the new behavior?

il3ven:
> @il3ven sorry I was AFK for nearly two weeks :)

No problem.

> Test pass now, but it strikes me there's no test for the new behavior?

I will add them. Can you provide the files where these tests should be added? Or maybe point me to existing tests for the throwsAsync function.

novemberborn:
IIRC it's all in https://github.com/il3ven/ava/blob/throws/test-tap/assert.js#L915=.

il3ven:
@novemberborn I have added test cases to cover the [code that I added](https://github.com/avajs/ava/blob/c712e61788fa0b5ccbacbb5c8b1909810541b458/lib/assert.js#L238-L249). I have also added a few more cases to cover [similar code](https://github.com/avajs/ava/blob/c712e61788fa0b5ccbacbb5c8b1909810541b458/lib/assert.js#L199-L236) which was uncovered before.

I haven't added a test case for `throwsAsync` because the same code is being used both in `throws` and `throwsAsync`. Let me know if I need to add it separately.

novemberborn:
@il3ven this is great!

The additional test cases would be nice to communicate intent, but then again it's a massive test file that we haven't ported to be self-hosted, so I'm not too concerned about that.

PR #2999, opened 2022-04-01
Fix function reference in t.try() assertion documentation

mefengl:
fix two function name that look like typo.


novemberborn:
Nice one, thanks @mefengl!

PR #3001, opened 2022-04-05
Allow parallel builds to be disabled

### LINKED ISSUE #2994: Parallel CI runs in GitLab CI break matrix
[Issuehunt badges](https://issuehunt.io/r/avajs/ava/issues/2994)

When using a GitLab CI configuration like this:

```yaml
test:
    image: node:$NODE_VERSION
    stage: test
    interruptible: true
    script:
        - npm run test
    parallel:
        matrix:
            - NODE_VERSION: [14, 16, lts, latest]
```

Ava assumes that the tests are being run in parallel, and thus splits the tests up into four. It then runs only a part of the tests in each node version, instead of testing everything on all node versions. This is the correct behavior when it is set like this:

```yaml
test:
    parallel: 4
```

But not when using the `matrix` keyword, since that is used

> to run a job multiple times in parallel in a single pipeline, but with different variable values for each instance of the job

(From [**`docs.gitlab.com/ci/yaml`**](https://docs.gitlab.com/ee/ci/yaml/#parallelmatrix))

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $25.00)

- [panascais panascais](https://issuehunt.io/u/panascais) ($25.00)

### Tips

</details>

il3ven:
Closes #2994 

This PR adds an option `utilizeParallelBuilds` to enable/disable parallel builds. By default, the value of `utilizeParallelBuilds` is true. Setting it to false will disable automatic parallel builds in CI.

For testing purposes I made a GitHub [repo](https://github.com/il3ven/ava-playground) and ran test cases in CI using the [branch](https://github.com/il3ven/ava/tree/parallelBuilds) of this PR.
- With `utilizeParallelBuilds` set to [undefined](https://github.com/il3ven/ava-playground/blob/6a1ec4505a53d0d8bd8fadc3014b6a302279a77d/ava.config.js). We can see test cases being executed in parallel. 
  Ref: https://github.com/il3ven/ava-playground/runs/5831541253 (Check npx ava)
- With `utilizeParallelBuilds` set to [false](https://github.com/il3ven/ava-playground/blob/b666d136c106f1de1e0741aa55e004c623461141/ava.config.js). We can see test cases *not* being executed in parallel.
  Ref: https://github.com/il3ven/ava-playground/runs/5831147820 (Check npx ava)

As usual, I am happy to incorporate any suggestions to this PR. 😄

### DIFF
--- a/docs/06-configuration.md
+++ b/docs/06-configuration.md
@@ -59,6 +59,7 @@ Arguments passed to the CLI will always take precedence over the CLI options con
 - `timeout`: Timeouts in AVA behave differently than in other test frameworks. AVA resets a timer after each test, forcing tests to quit if no new test results were received within the specified timeout. This can be used to handle stalled tests. See our [timeout documentation](./07-test-timeouts.md) for more options.
 - `nodeArguments`: Configure Node.js arguments used to launch worker processes.
 - `sortTestFiles`: A comparator function to sort test files with. Available only when using a `ava.config.*` file. See an example use case [here](recipes/splitting-tests-ci.md).
+- `utilizeParallelBuilds`: If `false`, disable [parallel builds](/docs/recipes/splitting-tests-ci.md) (default: true)
 
 Note that providing files on the CLI overrides the `files` option.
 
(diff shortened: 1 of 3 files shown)


novemberborn:
This is great, thanks @il3ven!

PR #3003, opened 2022-04-05
Improve typing of teardown methods

novemberborn:
- Explicitly type t.teardown() functions to return a promise
- Fix typing of plugin teardown. The returned method always returns Promise<void>.

The teardown functions may or may not return a promise. Support both to help with the @typescript-eslint/no-misused-promises ESLint rule.

### DIFF
--- a/plugin.d.ts
+++ b/plugin.d.ts
@@ -41,7 +41,7 @@ export namespace SharedWorker {
 		readonly file: string;
 		publish: (data: Data) => PublishedMessage<Data>;
 		subscribe: () => AsyncIterableIterator<ReceivedMessage<Data>>;
-		teardown: <TeardownFn extends () => void> (fn: TeardownFn) => TeardownFn;
+		teardown: (fn: (() => Promise<void>) | (() => void)) => () => Promise<void>;
 	};
 
 	export namespace Plugin {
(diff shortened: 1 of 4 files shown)


plugin.d.ts
```diff
@@ -41,7 +41,7 @@ export namespace SharedWorker {
 		readonly file: string;
 		publish: (data: Data) => PublishedMessage<Data>;
 		subscribe: () => AsyncIterableIterator<ReceivedMessage<Data>>;
-		teardown: <TeardownFn extends () => void> (fn: TeardownFn) => TeardownFn;
+		teardown: (fn: (() => Promise<void>) | (() => void)) => () => Promise<void>;
```
sindresorhus, line 44:
```suggestion
		teardown: (fn: (() => Promise<void> | void) => () => Promise<void>;
```

?

novemberborn: Would that be fine or would `void` "nullify" `Promise<void>`? Or would it be clearer to have overloads?

sindresorhus: Hmm: https://github.com/microsoft/TypeScript/issues/43921

I guess what you have is fine.

sindresorhus: I don't really know if overloads will make any difference.

PR #3009, opened 2022-04-11
Document how to load worker multiple times

codetheweb:
See https://github.com/avajs/ava/pull/3006 for context.

### DIFF
--- a/docs/recipes/shared-workers.md
+++ b/docs/recipes/shared-workers.md
@@ -34,6 +34,25 @@ const shared = registerSharedWorker({
 
 Within a test process you can only register one worker for each `filename`. Filenames are compared as-is, without normalization. If you call `registerSharedWorker()` a second time, the same worker instance is returned.
 
+If for some reason you want to load the same file as multiple different workers, you can append a unique hash to the end of the filename:
+
+```js
+import crypto from 'crypto';
+import {registerSharedWorker} from 'ava/plugin';
+
+const randomKey = crypto.randomBytes(20).toString('hex');
+
+const shared = registerSharedWorker<any>({
+  filename: new URL(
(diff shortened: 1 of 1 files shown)


docs/recipes/shared-workers.md
```diff
@@ -34,6 +34,25 @@ const shared = registerSharedWorker({
 
 Within a test process you can only register one worker for each `filename`. Filenames are compared as-is, without normalization. If you call `registerSharedWorker()` a second time, the same worker instance is returned.
 
+If for some reason you want to load the same file as multiple different workers, you can append a unique hash to the end of the filename:
+
+```js
+import crypto from 'crypto';
+import {registerSharedWorker} from 'ava/plugin';
+
+const randomKey = crypto.randomBytes(20).toString('hex');
```
novemberborn, line 43: It can't be random, then each test worker will end up loading a different shared worker… so it won't be shared.

docs/recipes/shared-workers.md
```diff
@@ -34,6 +34,25 @@ const shared = registerSharedWorker({
 
 Within a test process you can only register one worker for each `filename`. Filenames are compared as-is, without normalization. If you call `registerSharedWorker()` a second time, the same worker instance is returned.
 
+If for some reason you want to load the same file as multiple different workers, you can append a unique hash to the end of the filename:
```
novemberborn, line 37: It may be worth mentioning that `filename` takes a URL, and that's why you can append hashes or even use a query component.

docs/recipes/shared-workers.md
codetheweb: Whoops, my bad. 😅

novemberborn:
Cheers @codetheweb!

PR #3013, opened 2022-04-20
Remove logs from test failure summary

### LINKED ISSUE #2963: Duplicate t.log when test fails
I'm running `ava` v4.0.1, without any configuration.

Running the following code, outputs `t.log` twice on the console. Is there a reason for this?
```js
const test = require('ava');

test('application', (t) => {
  t.fail();
  t.log('1');
});
```

This is the output:
Selection_003

What I'm expecting to see is the log on the first part (where the test failed) but not on the second part (under the horizontal "line").
From my understanding, the second part of the output should just give details on why the test failed, there's no need to output the logs again.

This is extremely annoying when you're trying to `t.log` something longer than a couple of lines and your screen is flooded with duplicate text.

il3ven:
As per https://github.com/avajs/ava/issues/2963#issuecomment-1055752366, this amateur PR attempts to remove logs from the test failure summary. Closes #2963 

<details>
<summary>Output before this PR</summary>

```
npx ava

  ✔ pass
  ✖ sum Test failed via `t.fail()`
    ℹ v14.18.2
  ─

  sum

    ℹ v14.18.2

  index.test.js:10

    9:   t.log(process.version);
   10:   t.fail()               
   11: });                      

  Test failed via `t.fail()`

  › index.test.js:10:5

  ─

  1 test failed
```
</details>

<details>
<summary>Output after this PR</summary>

```
npx ava                         

  ✔ pass
  ✖ sum Test failed via `t.fail()`
    ℹ v14.18.2
  ─

  sum
  index.test.js:10

    9:   t.log(process.version);
   10:   t.fail()               
   11: });                      

  Test failed via `t.fail()`

  › index.test.js:10:5

  ─

  1 test failed
```
</details>


test-tap/reporters/default.regular.v16.log
```diff
@@ -108,6 +111,12 @@
...
   [90m› passes (test-tap/fixture/report/regular/unhandled-rejection.cjs:4:17)[39m
+  [90m› Test.callFn (file://lib/test.js:507:21)[39m
+  [90m› Test.run (file://lib/test.js:520:23)[39m
+  [90m› Runner.runSingle (file://lib/runner.js:285:33)[39m
+  [90m› Runner.runTest (file://lib/runner.js:361:30)[39m
+  [90m› async file://lib/runner.js:526:21[39m
+  [90m› async Runner.start (file://lib/runner.js:534:15)[39m
 
 ---tty-stream-chunk-separator
   [1mUnhandled rejection in unhandled-rejection.cjs[22m
```
il3ven, line 122: I tried updating the .log files by running `UPDATE_REPORTER_LOG=true npx tap test-tap/reporters/*` inside a docker container. I chose a docker container so I could change the node version. For node v16, I used the node-16 image.

The .log file generated inside the docker container is not the same as the logs generated during the workflow run. All the lines in green are not generated when the workflow runs. Thus failing the pipeline.

#### How can I update the .log files for different node versions?

I hope the above makes sense. 😅

novemberborn:
> How can I update the .log files for different node versions?

Using https://volta.sh/:

```
UPDATE_REPORTER_LOG=true volta run --node 12 npx tap test-tap/reporters/*
```

I've pushed those updates.

lib/reporters/default.js
```diff
@@ -515,9 +515,6 @@ export default class Reporter {
 
 	writeFailure(event) {
 		this.lineWriter.writeLine(colors.title(this.prefixTitle(event.testFile, event.title)));
-		if (!this.writeLogs(event, true)) {
-			this.lineWriter.writeLine();
```
novemberborn, line 519: Looking at the diff, I think we need this empty line.

il3ven: I have added the empty line. According to the existing logic, an empty line will be added when there are no logs.

novemberborn (approved):

PR #3017, opened 2022-04-27
Examples now use AVA 4

### LINKED ISSUE #2993: Examples use AVA 3
In examples dir in the ava repo, TestInterface is still used, even though it's now called TestFn

---

@novemberborn: in fact all examples are using AVA 3 and need updating: https://github.com/avajs/ava/tree/a8981540100653d99d1fe9476737751eeae50b33/examples

Timethor:
Resolves #2993

- I've updated some of the examples to "module", while others were left as "commonjs" ... thought it would be good to see both
- Seemed like I needed a `ava.config.js` file for most examples (assuming this is for monorepo behavior mentioned in the ESM Support sections of 4.X releases). If there's a more preferred way please let me know and I'll update
- Any new examples you'd like me to create for new v4 features, more than willing... still catching up, last used AVA when it was v1-v2

### DIFF
--- a/examples/endpoint-testing/ava.config.js
+++ b/examples/endpoint-testing/ava.config.js
@@ -1,0 +1,3 @@
+module.exports = { // eslint-disable-line import/no-anonymous-default-export
+	files: ['test.*'],
+};
--- a/examples/endpoint-testing/package.json
+++ b/examples/endpoint-testing/package.json
@@ -1,12 +1,13 @@
 {
 	"name": "ava-endpoint-testing",
 	"description": "Example for endpoint testing",
+	"type": "commonjs",
 	"scripts": {
 		"test": "ava"
 	},
 	"devDependencies": {
-		"ava": "^3.15.0",
-		"got": "^11.8.2",
+		"ava": "^4.2.0",
+		"got": "^12.0.4",
 		"test-listen": "^1.1.0"
 	}
 }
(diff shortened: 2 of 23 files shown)


novemberborn:
@Timethor nice!



> * I've updated some of the examples to "module", while others were left as "commonjs" ... thought it would be good to see both

I suppose, but I don't think it's helpful to see a CommonJS test that then needs to use a dynamic import() for an ES module.



> * Seemed like I needed a `ava.config.js` file for most examples (assuming this is for monorepo behavior mentioned in the ESM Support sections of 4.X releases). If there's a more preferred way please let me know and I'll update

Ha, yes that makes sense actually. You could `export default {}` though, there shouldn't be a need to override any of the defaults.



> * Any new examples you'd like me to create for new v4 features, more than willing... still catching up, last used AVA when it was v1-v2

Hmm, I'm not sure how full the coverage of v3 features was. Let's assume 100% and then we could have a look at the changelog since?

novemberborn:
Hi @Timethor, would you have time to get back to this?

novemberborn:
Line number selection may no longer work in StackBlitz with AVA 4, see https://github.com/avajs/ava/issues/3049#issuecomment-1147390189. Source maps now come from Node.js and we're seeing `https://` URLs which AVA isn't meant to deal with.

Timethor:
@novemberborn Sorry, nearly forgot about this, been a crazy few months. Will be getting some spare time in a week or two 🙃

novemberborn:
No worries @Timethor!

novemberborn:

Issue #2627, opened 2020-12-19
t.like not handling arrays as expected


awlayton:
I am trying to use the `t.like` assertion on deep objects with arrays of objects in them.

The following simplified example fails:
```ts
import test from 'ava';

test('deep t.like', (t) => {
  t.like({ a: [{ a: 1, b: 2 }] }, { a: [{ a: 1 }] });
});
```
```
> tsc -b

$ ava dist/ava.spec.js

  deep t.like

  Difference:

    {
      a: [
        {
          a: 1,
  -       b: 2,
        },
      ],
    }

  › src/ava.spec.ts:4:5

  ─

  1 test failed
error Command failed with exit code 1.
```

I expected the above assertion to pass. 
It seems that elements within arrays need to be deep equal rather than "like" for the assertion to pass.

AVA version:
```
$ ava --version
3.14.0
```

ninevra:
(Not a maintainer, but)

This is behaving as documented: ["Any values in `selector` that are not **regular objects** should be deeply equal to the corresponding values in value."](https://github.com/avajs/ava/blob/master/docs/03-assertions.md#likevalue-selector-message) What a "regular object" is in javascript is admittedly not well defined, but arrays aren't it.

It would sometimes be useful for `t.like()` to recurse into arrays, but then there would be lots of design questions to answer: Should order matter? Should extra trailing elements in the sample matter? Should `Map`s and `Set`s also be recursed into? Adding such special-cases to `t.like()` would make it much less clear what any given assertion is actually checking.

One can achieve the expected behavior either by writing one's own selector function and then using `t.deepEqual()`, or by using an outside assertion library.

awlayton:
I had interpreted it as anything that was JSON serializable. But I could see them meaning it more literally.

What do you mean about `t.deepEqual`? Is there some undocumented way to use it? It just takes 2 values to compare and an optional message.

ninevra:
Internally, what `t.like(actual, selector)` is doing is constructing a subset of `actual` and then comparing it by deep equality to `selector`; i.e. there exists some function `select(actual, selector) -> sample` such that `t.like(actual, selector)` is equivalent to `t.deepEqual(select(actual, selector), selector)` A very simplified version might look like:
```javascript
function select(actual, selector) {
  if (isPlainObject(selector) {
    const selected = {};
    for (const [key, value] of Object.entries(selector) {
      selected[key] = select(actual[key], value);
    }
    return selected;
  } else {
     return actual;
  }
}

[...]

PR #3023, opened 2022-05-12
fix: add support for use of arrays in t.like

Gozala:
fixes: #2627

### DIFF
--- a/lib/like-selector.js
+++ b/lib/like-selector.js
@@ -1,7 +1,8 @@
 export function isLikeSelector(selector) {
+	const prototype = Reflect.getPrototypeOf(selector)
 	return selector !== null
 		&& typeof selector === 'object'
-		&& Reflect.getPrototypeOf(selector) === Object.prototype
+		&& (prototype === Object.prototype || prototype === Array.prototype)
 		&& Reflect.ownKeys(selector).length > 0;
 }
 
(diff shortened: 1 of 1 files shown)


lib/like-selector.js
```diff
@@ -1,7 +1,8 @@
 export function isLikeSelector(selector) {
+	const prototype = Reflect.getPrototypeOf(selector)
 	return selector !== null
 		&& typeof selector === 'object'
-		&& Reflect.getPrototypeOf(selector) === Object.prototype
+		&& (prototype === Object.prototype || prototype === Array.prototype)
```
novemberborn, line 5: I think this could use `Array.isArray(selector)`. Note that `Reflect.ownKeys()` on an array returns `length` which messes up the next condition.

novemberborn (commented):
Thanks for the PR @Gozala! I think this could work, but it's been a while since I've thought about the logic here. It'll definitely need tests, I think the existing ones are in `test-tap/assert.js`.

novemberborn:
Hi @Gozala, do you think you would have time to get back to this?

Gozala:
> Hi @Gozala, do you think you would have time to get back to this?

I'm afraid not. I end up switching off of ava as it would not run in browser contexts. If I get back to using ava I'll definitely  make some time for this.

novemberborn:
No problem, thanks!

PR #3028, opened 2022-05-22
Document that AVA 4 cannot be run globally

### LINKED ISSUE #3026: Document that AVA 4 cannot be run globally
In commit a763592, the docs were modified to explain that AVA 4 cannot be run globally. In f09742f, this was removed ([diff](https://github.com/avajs/ava/commit/f09742f7baab7d44a48f1e3cee8f4b22deecf6f4#diff-a89aa74aa84d41f245185beac2e62ca024ba87d84740430e06aacf9b156cf5baL52-L53)). I assume the second change was a mistake. In any case, the user should know about this detail.

binyamin:
Resolves #3026

I added the note to `/readme.md`, instead of to `/docs/08-common-pitfalls.md` as discussed. I did this so I could put it directly under the installation instructions, which seemed the most appropriate place.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -66,6 +66,8 @@ Alternatively you can install `ava` manually:
 npm install --save-dev ava
 ```
 
+Note that AVA 4 cannot be run globally.
+
 Don't forget to configure the `test` script in your `package.json` as per above.
 
 ### Create your test file
--- a/readme.md
+++ b/readme.md
@@ -66,7 +66,7 @@ Alternatively you can install `ava` manually:
 npm install --save-dev ava
 ```
 
-Note that AVA 4 cannot be run globally.
+*Make sure to install AVA locally. As of AVA 4 it can no longer be run globally.*
 
 Don't forget to configure the `test` script in your `package.json` as per above.
 


novemberborn:
Do you think it may fit better in the `Running your tests` section?

binyamin:
Maybe. But then the user might (1) install it globally, (2) try to run the tests, (3) look back at the documentation, and then (4) remove it globally and install it locally. If we added it to the installation section, it would make sure this wouldn't happen. Anyhow, that was my thought process.

readme.md
```diff
@@ -66,6 +66,8 @@ Alternatively you can install `ava` manually:
 npm install --save-dev ava
 ```
 
+Note that AVA 4 cannot be run globally.
```
novemberborn, line 69:
```suggestion
*Make sure to install AVA locally. As of AVA 4 it can no longer be run globally.*
```

novemberborn (commented):
Fair enough. What do you think of my suggestion?

And then we could add your note in the `Running your tests` section so we've got our bases covered.

novemberborn:
Thanks @binyamin!

PR #3030, opened 2022-05-26
Remove redundant sentence in docs

elijahjpassmore:
Signed-off-by: Elijah J. Passmore <elijahjpassmore@elijahjpassmore.com>


novemberborn:
Thanks for the PR! I think in this case the redundancy provides extra emphasis.

PR #3038, opened 2022-06-01
apply correction to TypeScript recipe

AJStacy:
Updates the TypeScript recipe to correct TestFn to TestInterface for setting test context.


novemberborn:
`TestFn` is correct for AVA 4.

PR #3041, opened 2022-06-02
Require TypeScript 4.7

novemberborn:
- Only test against TypeScript 4.7
- Move type definitions to `entrypoints/`. Only TS 4.7 will resolve them now, but it's cleaner.
- Remove direct invocation of `tsd`, rely on `tsc --noEmit` instead
- Test with `.cts` files, containing ESM import syntax


novemberborn:
I'd like to test CJS import syntax as well but I already cannot remember how that's meant to be configured and used.

PR #3043, opened 2022-06-02
Require Node.js 14.19, 16.15 or 18

novemberborn:
Node.js 12 is end of life.

Increase the minimal versions for Node.js 14 and 16 to get a more modern feature baseline we can rely on for future development during AVA 5's lifetime.


sindresorhus:
In case it's useful, new stuff in Node.js 14:

- private class methods and accessors
- Top-level await
	+ Get rid of all `(async () => {})()`
- `crypto.randomInt` (Node.js 14.10)
- Optional chaining
- Nullish Coalescing
- `import fs from 'node:fs/promises';`
- `crypto.randomUUID()` (Node.js 14.17)
- AbortController `signal` option (Node.js 14.18)

novemberborn:
Lots of nice features we can make use of then!

PR #3044, opened 2022-06-02
Create a build process

RebeccaStevens:


RebeccaStevens:
Not sure why some tests are now failing, will investigate at a later date (hopefully tomorrow).

@novemberborn Do you know why some of the linting is playing up?
I was getting errors like [`Error: /home/runner/work/ava/ava/test-tap/assert.js: line 1415, col 2, Error - Nested block is redundant. (no-lone-blocks)`](https://github.com/avajs/ava/runs/6713077473?check_suite_focus=true) but now they seem to have fixed themselves??? (as in are no longer reported as an error, code is still the same).
I'm also getting a few `'ava/no-ignored-test-files' rule is disabled but never reported.` errors.

novemberborn:
What is the end goal with this?

novemberborn:
I don't see why we need to run `rollup` — the CJS and ESM entrypoints should be enough.

RebeccaStevens:
Hey, sorry I've been away for a bit.

> What is the end goal with this?

A few things:

- First, fixing up the CJS vs ESM entry points (explained below)
  - And resolving issues with the type files of each entry point (there are still some issue in CJS environments)
- Second, allowing the source code to be upgraded with newer ES feature and ideally be written in ESM.
  - This may also potentially involve switching to TypeScript

> I don't see why we need to run rollup — the CJS and ESM entry points should be enough.

The current ESM entry points require the CJS code which makes them not really ESM entry points at all.
They're essentially the same as not providing ESM entry points and just requiring the user to use the CJS ones in their ESM environment. Actually, it's worse than that because it will seem to the user that there are in fact ESM entry points when really there is not. This can lead to errors when the user is expecting the ESM entry points to behave as an actual ESM.

Adding rollup will give quite a few advantages; one of which is that it makes the ESM and CJS code completely separate from one another.

I don't see any downsides at all to adding rollup.

novemberborn:
> The current ESM entry points require the CJS code which makes them not really ESM entry points at all.

They are, as far as the exports are concerned. The `ava` entrypoint exports the `test()` method, and both CJS and ESM export the same exact function.

This is the key line, which exports a value set from an ESM file:

https://github.com/avajs/ava/blob/91f5254d7bfb6c658d0e89f48a852b92b70e31da/lib/worker/main.cjs#L12

> This can lead to errors when the user is expecting the ESM entry points to behave as an actual ESM.

I don't understand. You get the same function. It's an actual ESM file.

> Adding rollup will give quite a few advantages; one of which is that it makes the ESM and CJS code completely separate from one another.

But we don't want that. The `ava` export, and also `ava/plugin`, need to export functions that are defined by the worker code, and that code is ESM.

The only reason we have some CJS code is so AVA can be used in CJS test files.

live627:
> I don't see any downsides at all to adding rollup.

[The filesize is quite large.](https://packagephobia.com/result?p=rollup) Developers may have several instances of ava on possibly full and/or small hard drives, such as myself. Great strides have been made to [cut down the overall bundle size of ava](https://packagephobia.com/result?p=ava), and this pr will single-handedly undo all of that good work.

There is a physics joke about npm packages being the heaviest objects in the universe. Don't let ava contribute to that.

novemberborn:
@live627 `rollup` wouldn't be bundled. If anything, the resulting build could require fewer bytes.

That said this PR has stalled and at this time I'm not convinced it's worth having a build step, so I will close. @RebeccaStevens always happy to discuss and as always thank you for your input.

Issue #2946, opened 2022-01-16
AssertionError when debugging in VS Code


Basssiiie:
Hey I've been using the VS Code debugger launch configuration from [here](https://github.com/avajs/ava/blob/main/docs/recipes/debugging-with-vscode.md#creating-a-launch-configuration) for quite a while now, but since updating to V4 I cannot get it to work anymore. Whenever I run it, I run into this error on any of the test files:
```
  Uncaught exception in tests\temp.tests.ts

  AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:

    assert(refs.runnerChain)


  × tests\temp.tests.ts exited with a non-zero exit code: 1
  ─

  1 uncaught exception
```
The example test file used (the error occurs on any test file though):
```ts
import test from "ava";

test("ava test", t =>
{
	t.pass();
});
```
The VS Code `launch.json` (exact copy from the linked docs):
```json
{
	"version": "0.2.0",
	"configurations": [
		{
			"type": "node",
			"request": "launch",
			"name": "Debug AVA test file",
			"runtimeExecutable": "${workspaceFolder}/node_modules/.bin/ava",
			"runtimeArgs": [
				"${file}"
			],
			"outputCapture": "std",
			"skipFiles": [
				"<node_internals>/**/*.js"
			]
		}
	]
}
```
AVA configuration in `package.json`:
```json
	"ava": {
		"extensions": [
			"ts"
		],
		"files": [
			"tests/**/*.tests.ts"
		],
		"require": [
			"./tests/_setup.js",
			"ts-node/register"
		],
		"verbose": true
	}
```
Running `npx ava --version` reports 0.1.0? I have 4.0.1 installed according to the `package.json` and the `node_modules/ava` folder and both `npm install` and `npm update` do not change that.

AVA debugging still works in another project running 3.15

This sounds like a bug right?

Thank you for your time!

novemberborn:
That is peculiar. Are you breaking on (un)caught exceptions by any chance? I think this would occur if the test file is run directly, though going by the configuration I wouldn't expect that to happen.

Does debugging work through use VS Code's "debug terminal" feature?

There's a known issue where `npx ava --version` prints "unknown". No idea why it would print `0.1.0`, maybe something is wrong somehow somewhere with that.

Basssiiie:
Both "caught exceptions" and "uncaught exceptions" are unchecked. I removed all my other breakpoints as well but it still happens regardless of having breakpoints.

[...]

PR #3051, opened 2022-06-06
VSCode debug documentation fix

AlencarGabriel:
Fix issue #2946 where VSCode was no longer debugging ava tests using binstub installed by npm.

Changing the script to call directly cli.mjs worked for some users with this problem on windows.


novemberborn:
Thanks @AlencarGabriel!

PR #3055, opened 2022-06-13
Added missing banner

solomancode:


novemberborn:
A former member of the AVA team is directly impacted by the war in Ukraine and asked us to add the banner. We did so due to this personal connection.

We won't add banners for other causes, no matter how worthy our contributors or users may find them.

We won't facilitate a discussion around this either.

PR #3080, opened 2022-08-06
Detect process.exit() called from tests

### LINKED ISSUE #861: Override `process.exit` to give some context when it's called unexpectedly in a test.
[Issuehunt badges](https://issuehunt.io/r/avajs/ava/issues/861)

Just had an issue on Gitter where a users code was causing `process.exit(1)` to be called. He got the "test.js exited with a non-zero exit code message", but there was no helpful information.

I say we try and capture that somehow:

``` js
var oldExit = process.exit;
process.exit = function (code) {
  if (code > 0) {
    var stack = new Error().stack;
    // log it? send it to parent?
    // ...
  }
  oldExit.apply(process, arguments);
};
```

> There is a $60.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/avajs/ava/issues/861).

gibson042:
Fixes #861

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#861: Override `process.exit` to give some context when it's called unexpectedly in a test.](https://oss.issuehunt.io/repos/26820798/issues/861)
---
</details>

### DIFF
--- a/lib/worker/base.js
+++ b/lib/worker/base.js
@@ -19,6 +19,35 @@ import {flags, refs, sharedWorkerTeardowns} from './state.cjs';
 import {isRunningInThread, isRunningInChildProcess} from './utils.cjs';
 
 const currentlyUnhandled = setUpCurrentlyUnhandled();
+let runner = Object.create(null);
+
+// Override process.exit with an undetectable replacement
+// to report when it is called from a test (which it should never be).
+const {apply} = Reflect;
+const realExit = process.exit;
+
+async function exit(code, forceSync = false) {
+	dependencyTracking.flush();
+	if (!forceSync) {
+		await channel.flush();
+	}
+
+	apply(realExit, process, [code]);
+}
+
+process.exit = new Proxy(realExit, {
+	apply(fn, receiver, args) {
+		channel.send({type: 'internal-error', err: serializeError('Forced exit error', false, new Error('Unexpected process.exit()'), runner.file)});
+
+		// Make sure to extract the code only from `args` rather than e.g. `Array.prototype`.
+		// This level of paranoia is usually unwarranted, but we're dealing with test code
+		// that has already colored outside the lines.
+		const code = args.length > 0 ? args[0] : undefined;
+
+		// Force a synchronous exit as guaranteed by the real process.exit().
+		exit(code, true);
+	},
+});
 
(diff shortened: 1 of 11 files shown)


lib/worker/base.js
```diff
@@ -19,6 +19,35 @@ import {flags, refs, sharedWorkerTeardowns} from './state.cjs';
...
+	if (!forceSync) {
+		await channel.flush();
+	}
+
+	apply(realExit, process, [code]);
+}
+
+process.exit = new Proxy(realExit, {
+	apply(fn, receiver, args) {
+		channel.send({type: 'internal-error', err: serializeError('Forced exit error', false, new Error('Unexpected process.exit()'), runner.file)});
```
novemberborn, line 40: Could leave `runner` undefined and use `runner?.file`, that should work in the supported Node.js versions I believe.

novemberborn, line 40: I think `internal` should refer to AVA doing something wrong, but this is triggered by user code. What do you think?

novemberborn (commented):
Amazing!

lib/worker/base.js
gibson042: Done.

gibson042: I think that's valid, although there is also an argument to be made that early termination before all tests run is an internal error. But what should this be instead?

novemberborn: `process-exit` maybe? Would also allow us to print a specific message, like we do for `SIGINT`:

https://github.com/avajs/ava/blob/392ab65cd4d203643e5bb29aa7f152056fb6040b/lib/reporters/default.js#L195-L200

gibson042: Done.

novemberborn:
@gibson042 I should have some time this week to help get this over the line.

novemberborn:
@gibson042 thanks for [`96ff6ef` (#3080)](https://github.com/avajs/ava/pull/3080/commits/96ff6ef0b80cba3d42d2237cb7e3d410ce9b8776)! Let's see if it improves things.

I've pushed several updates, please let me know what you think.

gibson042:
@novemberborn The updates look good to me. My only further change was to rename the generic `proxyApply` to `handleProcessExit`.

PR #3082, opened 2022-08-06
Improve test result accessibility

### LINKED ISSUE #1558: Explain diff gutter symbols

### LINKED ISSUE #2919: Test results are inaccessible to red-green color-blind, blind and CI
## TLDR

I'm using AVA 4.0.0-rc.1 and have a few tests marked as ".failing" since they are known test failures. When looking at the test output, I expect that I can tell which tests are known failures, but known-failing tests look exactly like passing tests in some situations.

## Details

AVA's default output doesn't clearly show which of the tests are known failures because it only distinguishes between test successes and test failures using color. **When you only use color to distinguish between different things in an interface, you are making those things inaccessible to some people.**

Here's some AVA test results on my terminal. You can see that both known failures and successes have the ✔ checkmark prepended and they only differ based on their color:

ava-output-with-color

This is how those same AVA test results appear on GitHub's CI, which is similar to how red-green color-blind would see it. There's no way to tell which tests are known failures:

ava-output-on-ci

We need to re-think how these test results are displayed so they are accessible to everyone, including:
* developers who are red-green colorblind
* developers using screen readers
* developers who don't have a color-enabled terminal
* developers reading a plain text log file (like in a CI)

## Possible solutions

As a reference, here's a screenshot of skipped, failing, passing and known failing tests:
ava-all-test-types

Maybe known failures should be in "skipped" yellow with a ✖ cross instead of the checkmark. Known failures are more like skipped tests than anything else.

Also, right now failing tests only have a red ✖ with normal-colored descriptive text. Can we make the descriptive text of failing tests all in red? And maybe bold too? The bold isn't necessary since failures are printed again at the end of the test result list, but it helps the failures visually stick out from the rest of the list.

As for people who are using screen readers, I think we might need to add text to describe what `-`, `✖`, `✔` mean. Especially if we use `✖` for known and unknown fails. Maybe this?:

```
  ✔ PASS: Example › should be an example test
  - SKIP: Example › should be an example test
  ✖ FAIL: Example › should be an example test
  ✖ KNOWN FAIL: Example › should be an example test (or maybe SKIP FAIL or SKIPPED FAIL)
```

Here's a mockup with color and bold added:
ava-test-result-mockup

gibson042:
* Explain diff gutter symbols.
* Add test result labels (color covers figure and label, and indicates passed-as-expected vs. not).
* Bold the source line of errors in code excerpts.

Fixes #1558
Fixes #2919

example output


gibson042:
This produces lots of failures like the following; how do I update the [TAP] test expectations in bulk?
```
        not ok 43 - should be equal
          ---
          compare: ===
          at:
            line: 59
            column: 6
            file: file:///home/runner/work/ava/ava/test-tap/assert.js
            function: assertFailure
          stack: |
            assertFailure (file://test-tap/assert.js:59:6)
            failsWith (file://test-tap/assert.js:98:2)
            Test.<anonymous> (file://test-tap/assert.js:277:2)
            file://test-tap/assert.js:202:1
          diff: |
            --- expected
            +++ actual
            @@ -1,1 +1,1 @@
            -Difference:
            +Difference (- actual, + expected):
          ...
```

gibson042:
Found it: https://github.com/avajs/ava/blob/main/test-tap/reporters/readme.md

lib/worker/options.cjs
```diff
@@ -14,4 +20,11 @@ exports.set = newOptions => {
 	}
 
 	options = newOptions;
+	if (options.chalkOptions) {
+		if (setChalk) {
+			setChalk(options.chalkOptions);
+		} else {
+			chalk.then(chalk => chalk.set(options.chalkOptions));
+		}
```
gibson042, line 28: This is from 1b64bff382e00f85676aea888ab762b2ecc2d815, an independent experiment that can be reverted if deemed unsuccessful.
> This _seems_ to work, but the resulting interaction between CommonJS and ESM
feels a bit Zalgo-y. It may be better for modules to keep setting global
options and chalk options in parallel while the former remains CommonJS,
but I wanted to see what simplification would look like.

lib/cli.js
```diff
@@ -223,7 +223,7 @@ export default async function loadCli() { // eslint-disable-line complexity
 
 	const combined = {...conf};
 
-	for (const flag of Object.keys(FLAGS)) {
+	for (const flag of ['color', ...Object.keys(FLAGS)]) {
```
novemberborn, line 226: Why not add to `FLAGS`?

lib/concordance-options.js
```diff
@@ -85,17 +85,21 @@ const colorTheme = {
 	undefined: ansiStyles.yellow,
 };
 
-const plainTheme = JSON.parse(JSON.stringify(colorTheme), value => typeof value === 'string' ? stripAnsi(value) : value);
-
-const theme = chalk.level > 0 ? colorTheme : plainTheme;
+const plainTheme = JSON.parse(JSON.stringify(colorTheme), (_name, value) => typeof value === 'string' ? stripAnsi(value) : value);
```
novemberborn, line 88: I'm confused, how did `plainTheme` _ever_ work?!

lib/worker/options.cjs
novemberborn: Is this needed for this PR? Can we discuss it in a separate PR? It's all a little complicated I know, do we need to change it? What do you think the benefits are?

novemberborn:

novemberborn:
Thank you @gibson042, very exciting!

Note that we're seeing similar CI failures on the `main` branch, so let's not worry about those here.

sindresorhus:
I agree with almost everything. However, I'm not a big fan of the `PASS:` and `FAIL:` noise. I think the icons are clear enough for the common case and we could instead only use the prefixes for the other cases like `EXPECTED FAIL` and `SKIP`.

lib/worker/options.cjs
gibson042: I'd prefer to keep it, because otherwise the new tests don't correspond with reality (because worker output does not consistently respect color settings) and will undergo a fair amount of churn. But if you feel strongly that it should be separated, I can do that.

lib/cli.js
gibson042: Because `FLAGS` is used to define options for subcommands, whereas `color` applies at top-level: https://github.com/avajs/ava/blob/392ab65cd4d203643e5bb29aa7f152056fb6040b/lib/cli.js#L162-L183

lib/concordance-options.js
gibson042: Pretty much on accident. It ended up as an empty string, which concordance is sufficiently tolerant to accept as effectively absent, and the ava `colorTheme` stripped of its ANSI escape sequences doesn't actually deviate from the concordance default theme:
<details><summary>diff</summary>

```diff
--- concordance defaultTheme (normalized)
+++ ava corrected plainTheme (normalized)
@@ -1,11 +1,7 @@
 {
-  "bigInt": {
-    "close": "",
-    "open": ""
-  },
   "boolean": {
     "close": "",
     "open": ""
   },
   "circular": "[Circular]",
   "date": {
@@ -42,23 +38,20 @@
   },
   "global": {
     "close": "",
     "open": ""
   },
   "item": {
-    "after": ",",
-    "customFormat": null,
-    "increaseValueIndent": false
+    "after": ","
   "list": {
     "closeBracket": "]",
     "openBracket": "["
   },
   "mapEntry": {
-    "after": ",",
-    "separator": " => "
+    "after": ","
   },
   "maxDepth": "…",
   "null": {
     "close": "",
     "open": ""
   },
@@ -81,26 +74,23 @@
       "close": "",
       "open": "@"
     }
   },
   "property": {
     "after": ",",
-    "customFormat": null,
-    "increaseValueIndent": false,
     "keyBracket": {
       "close": "]",
       "open": "["
     },
-    "separator": ": "
+    "valueFallback": "…"
   },
   "regexp": {
     "flags": {
       "close": "",
       "open": ""
     },
-    "separator": "---",
     "source": {
       "close": "/",
       "open": "/"
     }
   },
   "stats": {
@@ -133,18 +123,16 @@
         "close": "",
         "open": ""
       }
     },
     "line": {
       "close": "'",
-      "escapeQuote": "'",
       "open": "'"
     },
     "multiline": {
       "end": "`",
-      "escapeQuote": "`",
       "start": "`"
     },
     "open": ""
   },
   "symbol": {
     "close": "",
```

</details>

gibson042:
> I agree with almost everything. However, I'm not a big fan of the `PASS:` and `FAIL:` noise. I think the icons are clear enough for the common case and we could instead only use the prefixes for the other cases like `EXPECTED FAIL` and `SKIP`.

@sindresorhus Without "PASS" and "FAIL", vision-impaired consumers are left to infer semantics from information like "check mark" and "multiplication symbol" (the latter improved a bit after https://github.com/sindresorhus/figures/issues/92 ). I'd prefer to keep the verbosity in support of accessibility and output consistency, but if you still disagree then let me know and I'll remove them.

novemberborn:
@gibson042 I should have some time this week to help get this over the line.

lib/worker/options.cjs
novemberborn: I had a go at removing it, and other than `test-tap/assert.js` requiring an extra `stripAnsi()` there's no impact. I think I prefer the "set once" approach rather than having code react to changes.

novemberborn:
> > I agree with almost everything. However, I'm not a big fan of the `PASS:` and `FAIL:` noise. I think the icons are clear enough for the common case and we could instead only use the prefixes for the other cases like `EXPECTED FAIL` and `SKIP`.
> 
> @sindresorhus Without "PASS" and "FAIL", vision-impaired consumers are left to infer semantics from information like "check mark" and "multiplication symbol" (the latter improved a bit after [sindresorhus/figures#92](https://github.com/sindresorhus/figures/issues/92) ). I'd prefer to keep the verbosity in support of accessibility and output consistency, but if you still disagree then let me know and I'll remove them.

@sindresorhus how strongly do you feel about this?

I do like the consistency.

sindresorhus:
> @sindresorhus how strongly do you feel about this?

I don't feel strongly about it.

sindresorhus:
> I do like the consistency.

Most assertion will be pass or fail, so there wouldn't be much inconsistency in practice. And it may even have been beneficial to make the uncommon cases more obvious. The result now is also more noisy.

novemberborn:
Project name aside, we don't really use all caps. I've pushed a commit to use lowercase and put it in square brackets.

Most of the time, most tests pass. So I've removed the prefix for passed tests.

I also noticed that the assertion message was only differentiated from the test title by color, so I've made that italic.

PR #3086, opened 2022-08-15
Recipe for executing tests inside of a browser

nicholascelestin:
I had a need to execute tests inside of a browser, and this is the solution I came up with. I thought it might be useful to other folks who had a similar need.


novemberborn:

PR #3091, opened 2022-08-25
Handle findSourceMap returning null since Node.js 18.8.0

Brooooooklyn:


iambumblehead:
this issue is causing the CI pipeline here to fail https://github.com/iambumblehead/esmock/runs/8011056406?check_suite_focus=true

lib/snapshot-manager.js
```diff
@@ -391,7 +391,10 @@ class Manager {
 
 const resolveSourceFile = mem(file => {
 	const sourceMap = findSourceMap(file);
-	if (sourceMap === undefined) {
+	// since Node.js 18.8.0, the return value of `findSourceMap` changed to null if not found
+	// it was `undefined` in previous version
+	// https://github.com/nodejs/node/pull/43875
+	if (sourceMap == undefined) {
```
sindresorhus, line 397:
```suggestion
	if (sourceMap === undefined || sourceMap === null) {
```

novemberborn:
Nice find @Brooooooklyn. Too bad the documentation doesn't define behavior when the source map cannot be found. I guess that's what we get for using an experimental feature.

novemberborn (approved):

novemberborn:
Should be fixed in 4.3.3.

PR #3095, opened 2022-08-25
Work around os.cpus() returning an empty array on unsupported platforms

catink123:
There is a bug in Node's built-in `os` module, where, in some cases, getting the `os.cpus()` value will return an empty Array. As a result, the testing process won't even start. To accommodate for this issue, I suggest adding a check for `os.cpus()` value.


lib/api.js
```diff
@@ -247,7 +247,8 @@ export default class Api extends Emittery {
 			}));
 
 			// Resolve the correct concurrency value.
-			let concurrency = Math.min(os.cpus().length, isCi ? 2 : Number.POSITIVE_INFINITY);
+			const cpuCount = os.cpus().length;
+			let concurrency = Math.min(cpuCount > 0 ? cpuCount : 1, isCi ? 2 : Number.POSITIVE_INFINITY);
```
sindresorhus, line 251: Please add a comment with a link to the Node.js issue (open one if none exist).

catink123:
Here's the Node.js issue describing the `os.cpus()` bug: https://github.com/nodejs/node/issues/38190

novemberborn:
On what platform are you running into this, and with which version of Node.js?

novemberborn:
Okay now I am curious, why run AVA on Android?

catink123:
Well, as I wanted to try web development on the go and I don't have a laptop, I decided to do it on my Android tablet, using Termux and Neovim. Most utilities just work and only some refuse to work because of bugs and poor (if any) support for Android.

novemberborn:
Fair enough!

catink123:
Thanks!

Issue #3060, opened 2022-07-10
Cap idle timeouts to the maximum possible value


novemberborn:
See discussion in https://github.com/avajs/ava/issues/3058. Both the `--timeout` feature and `t.timeout()` rely on `setTimeout()`. In Node.js there is a maximum delay of `2 ** 31 - 1` milliseconds. Any larger value causes Node.js to print a warning and use a `1ms` timeout instead.

We should cap the timeouts to the maximum possible value.

PR #3100, opened 2022-09-04
Cap idle timeouts to the maximum possible value

creestor:
**I have fixed this issue** : https://github.com/avajs/ava/issues/3060.
Was added constraint for max time of setTimeout.

### DIFF
--- a/lib/now-and-timers.cjs
+++ b/lib/now-and-timers.cjs
@@ -1,5 +1,14 @@
 'use strict';
 const timers = require('timers');
 
-Object.assign(exports, timers);
+// Slightly higher than 24 days
+const MAX_INT32 = Math.pow(2, 31) - 1;
+
+const setSafeTimeout = (callback, ms) => {
+    const avaliableMs = ms < MAX_INT32 ? ms : MAX_INT32;
+
+    return timers.setTimeout(callback, avaliableMs)
+}
+
+Object.assign(exports, timers, { setSafeTimeout });
 exports.now = Date.now;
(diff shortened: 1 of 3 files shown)


lib/now-and-timers.cjs
```diff
@@ -1,5 +1,14 @@
...
+// Slightly higher than 24 days
+const MAX_INT32 = Math.pow(2, 31) - 1;
+
+const setSafeTimeout = (callback, ms) => {
+    const avaliableMs = ms < MAX_INT32 ? ms : MAX_INT32;
+
+    return timers.setTimeout(callback, avaliableMs)
+}
+
+Object.assign(exports, timers, { setSafeTimeout });
```
novemberborn, line 13: It'd be better to assign `exports.setSafeTimeout` like we do for `exports.now` below.

lib/now-and-timers.cjs
```diff
@@ -1,5 +1,14 @@
 'use strict';
 const timers = require('timers');
 
-Object.assign(exports, timers);
+// Slightly higher than 24 days
+const MAX_INT32 = Math.pow(2, 31) - 1;
+
+const setSafeTimeout = (callback, ms) => {
+    const avaliableMs = ms < MAX_INT32 ? ms : MAX_INT32;
```
novemberborn, line 8: Perhaps `maxMs`? And use `Math.min(ms, MAX_INT32)`.

novemberborn (commented):
Thanks for the PR @creestor! I've left some comments, also you'll have to run `npx xo --fix` to fix the linting issues.

lib/now-and-timers.cjs
creestor: could you relaunch the ci? Probably it's occasional building error, cause check on node 18.0-windows. broken due to timeout.

novemberborn:
Thanks @creestor!

PR #3101, opened 2022-09-04
Update dependencies

novemberborn:
- Test with TypeScript 4.8 (but don't update the dependency)
- Update tsd and xo


novemberborn:
More strange failures in 4.8, holding off for now.

PR #3109, opened 2022-09-18
Update to the browser testing recipe to use jsdom directly

DavidVujic:
In this PR, I am suggesting to use `jsdom` directly instead of the `browser-env` tool. Even though it is a helpful tool, browser-env hasn't been updated in many years as it seems.

I also believe the suggestions here about writing unit tests for JavaScript in the browser are more explicit, by adding only the necessary globals and DOM elements for the particular test target. I like to see it as a TDD friendly approach 😄


novemberborn:
Nice one, thanks @DavidVujic!

PR #3123, opened 2022-10-23
Rewrite watcher

novemberborn:
This PR aims to rewrite the watcher implementation.

Goals:

* [x] Remove `chokidar` dependency, rely on `fs.watch()`.
* [x] This will require some additional logic for Linux (where we cannot watch recursively), but we don't need the full behavior that `chokidar` (or similar) provides
* [x] Simplify logic
* [x] Use static analysis for dependency tracking (https://github.com/vercel/nft)
* [ ] Use self-hosted integration-style tests

And as a follow-up goal, integrate with `@ava/typescript` so changes to say `src/` are detected, and depending on whether it assumes a running `tsc` wait for `build/` to be updated, or run immediately since it compiles prior to testing. But this is beyond AVA's current watch mode so can come later.


unional:
Any consideration about plugins, btw? :)

novemberborn:
Closing in favor of https://github.com/avajs/ava/pull/3218.

Issue #2780, opened 2021-06-24
Logs swallowed during test timeout/interrupt


Connor-Wood:
Existing writePendingTests is not too helpful when trying to figure out what went wrong in the test during these scenarios. You should print all logs accumulated if verbose is turned on.  Seems calling writeTestSummary before the existing code in each case would suffice

novemberborn:
Looks like this may have been done in #3125.

PR #3125, opened 2022-10-25
Output logs for tests that remain pending when AVA exits

### LINKED ISSUE #3122: t.log() does not output anything if an async test fails due to time out
First, thanks for maintaining such a great tool. So far it has addressed most of our javascript unit and integration test running needs.

**What you're trying to do**
I would like to see important test debug logs even if it times out.

**What happened**
Simple async test that will timeout because the promise resolves after 20 seconds. The test has a simple `t.log()` invocation.
```js
import test from 'ava';

test('example where logs are not provided by ava when test time out', async (t) => {
  t.log({ message: 'Very important debug statement, that developer would love to see in output after test run' });

  await new Promise((res) => {
    console.log('start promise that resolves after timeout...');
    setTimeout(res, 20000);
  });
});
```

Simple test output (note ava is run with timeout=1s configured):
```bash
 $ npx ava --timeout=1s --verbose

start promise that resolves after timeout...
  
  ✖ Timed out while running tests

  1 tests were pending in test/example.js

  ◌ example where logs are not provided by ava when test time out

  ─

  1 test remained pending after a timeout
```

**What you expected to happen**
I expected the output to contain the log: `Very important debug statement, that developer would love to see in output after test run`.
Being able to see this log will be very helpful in determining why the test timed out.

**Environment**
`package.json`
```json
{
  "name": "ava-logging-timeout-bug",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "type": "module",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC",
  "dependencies": {
    "ava": "^4.3.3"
  }
}
```
```bash
$ node -v 
v16.16.0
$ npx ava --version
4.3.3
```

kevo1ution:
Fixes #3122 , #2780

**What**
This PR plumbs test logs to be printed when tests timeout due to the global timeout set by `ava --timeout <timeout>` command.

**Why**
`t.log` is a very useful utility for debugging what happens when a test fails. When a test times out, these logs are useful for figuring out why the tests timed out.


kevo1ution:
@novemberborn 

This is the output with the same code snipping on #3122 . Before I attempt writing test cases for this, do you think this looks good implementationwise?
```
 % npm run test

> ava-logging-timeout-bug@1.0.0 test
> npx ava --timeout=1s --verbose


start promise that never resolves...
  
  ✘ Timed out while running tests

  1 tests were pending in test/example.js

  ◌ example where logs are not provided by ava when test time out
    ℹ {
        message: 'Very important debug statement, that developer would love to see in output after test run',
      }

  ─

  1 test remained pending after a timeout
```

lib/run-status.js
```diff
@@ -9,6 +9,7 @@ export default class RunStatus extends Emittery {
 		super();
 
 		this.pendingTests = new Map();
+		this.pendingTestsLogReference = new Map();
```
kevo1ution, line 12: separate map from `this.pendingTests` because they are modified and initialized in different events with different objects.

novemberborn:
@kevo1ution this is on my list to look at more closely, but at first glance it seems reasonable.

kevo1ution:
@novemberborn This is ready for review. It took some time to figure out how testing the test framework works (lol), so hopefully, I did it up to your standards. Let me know if there is anything I missed

kevo1ution:
@novemberborn bumping on this, let me know if there's anything else I need to do, thanks!

lib/reporters/default.js
```diff
@@ -370,8 +370,11 @@ export default class Reporter {
 			}
 
 			this.lineWriter.writeLine(`${testsInFile.size} tests were pending in ${this.relativeFile(file)}\n`);
+			const testTitleToLogs = evt.pendingTestsLogReference.get(file) ?? new Map();
```
novemberborn, line 373:
```suggestion
			const testTitleToLogs = evt.pendingTestsLogReference.get(file);
```

lib/reporters/default.js
```diff
@@ -370,8 +370,11 @@ export default class Reporter {
...
 			for (const title of testsInFile) {
+				const logs = testTitleToLogs.get(title) ?? [];
```
novemberborn, line 375: `writeLogs()` can handle falsey `logs` values. So with optional chaining and the above we can do:

```suggestion
				const logs = testTitleToLogs?.get(title);
```

lib/run-status.js
novemberborn: Why `Reference`?

lib/run-status.js
```diff
@@ -198,6 +209,12 @@ export default class RunStatus extends Emittery {
 		return 0;
 	}
 
+	addLogReference(event) {
+		if (this.pendingTestsLogReference.has(event.testFile)) {
+			this.pendingTestsLogReference.get(event.testFile).set(event.title, event.logs);
+		}
```
novemberborn, line 215:
```suggestion
		this.pendingTestsLogReference.get(event.testFile)?.set(event.title, event.logs);
```

novemberborn (commented):
Nice! Left some comments and suggestions around optional chaining.

lib/run-status.js
kevo1ution: oops, I think `pendingTestsLogs` sounds better

novemberborn (approved):
Lovely!

kevo1ution:
> Lovely!

Thank you! Looking forward to using this in my repos

PR #3128, opened 2022-10-27
Add main field to package.json for compatibility

Baune8D:
Linting with `eslint-plugin-import` currently breaks with the error `Unable to resolve path to module 'ava'  import/no-unresolved` when importing using CommonJS.

Since this package officially still supports CommonJS, adding back the main field will help with compatibility, and fix the above error.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
 	"bin": {
 		"ava": "entrypoints/cli.mjs"
 	},
+  	"main": "./entrypoints/main.cjs",
 	"exports": {
 		".": {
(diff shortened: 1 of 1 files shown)


sindresorhus:
`exports` works for CommonJS too. This is a problem with `eslint-plugin-import`.

novemberborn:
As far as I can tell, we've never had a `main`, and the `exports` were introduced in March 2021 https://github.com/avajs/ava/pull/2729. Why is this an issue now?

PR #3135, opened 2022-11-01
Check for --config file extensions after they fail to load, allowing custom loaders

### LINKED ISSUE #3134: allow resolve config file to use loaders
The `resolveConfigFile` function currently restricts the resolved files to .js, .cjs, or .mjs files.

https://github.com/avajs/ava/blob/7007f13411cb476d134a161d46c5d0185c1495c0/lib/load-config.js#L40-L42

What I'm hoping to use, without having to patch ava myself, is a .ts file and use a loader pushed on to node through an environment variable like so

```
export NODE_OPTIONS='--experimental-loader=@esbuild-kit/esm-loader'
```

The config file in question performs setup and having it typed helps a lot.

panva:
This pushes the configFile extension check down to `loadConfig()` to allow experimental loaders use. If there's a loader registered that will handle an arbitrary extension loading, there's no need to throw.

Resolves #3134


novemberborn:
Yes this could work. What I'd need to double check is how this interfaces with ambiguous config-file detection. 

Would you use this with the `--config` option?

panva:
> Would you use this with the --config option?

Yes.

novemberborn (approved):
@panva cool! Having refamiliarized myself with the loader logic I think this is a good addition to make loaders possible.

Let's see what CI says.

novemberborn:
https://github.com/avajs/ava/pull/3137 should fix the Node.js 16 failures.

PR #3151, opened 2022-12-19
De-register all shared workers

codetheweb:
This fixes the bug introduced in https://github.com/avajs/ava/pull/3149 that I briefly mentioned.

Since we were assigning the launched shared worker to a single variable per test worker, we inadvertently assumed that a test worker would launch a maximum of 1 shared worker.

I tested this patch with two of our internal codebases that heavily use shared workers and everything's working as expected now.

Sorry for the previous buggy PR!

### DIFF
--- a/lib/plugin-support/shared-workers.js
+++ b/lib/plugin-support/shared-workers.js
@@ -48,64 +48,63 @@ function launchWorker(filename, initialData) {
 	return launched;
 }
 
+/**
+ * Handles the shared worker lifecycle within a test worker thread.
+ */
 export async function observeWorkerProcess(fork, runStatus) {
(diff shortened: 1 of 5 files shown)


novemberborn:
Thanks for the follow-up @codetheweb! I haven't touched a keyboard much these past few weeks but hopefully I can find some time soon. Of course it's back to work so busy busy again 😄

novemberborn:
@codetheweb thoughts on the commit I pushed up?

Thanks for adding all these test cases, it's pretty hard to _think_ about the expected behavior.

codetheweb:
Looks perfect to me! Much clearer on what's going on.

PR #3170, opened 2023-02-07
Explicitly configure script name for CLI

mcecode:
Currently, the help command/option prints `cli.mjs` as the script name. This PR sets it to `ava` so that it's in line with the [CLI doc](https://github.com/avajs/ava/blob/main/docs/05-command-line.md).

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -140,6 +140,7 @@ export default async function loadCli() { // eslint-disable-line complexity
 
 	let resetCache = false;
 	const {argv} = yargs(hideBin(process.argv))
+		.scriptName('ava')
 		.version(pkg.version)
 		.parserConfiguration({
 			'boolean-negation': true,


novemberborn:
Thanks for the PR!

How are you getting the help options? `npx ava --help` prints `ava` for me.

mcecode:
> How are you getting the help options?

I also use `npx ava --help`.

> `npx ava --help` prints `ava` for me.

Hmmm. That's weird.

Testing it on different terminals on my Windows 10 machine, I can reproduce the script name being set to `cli.mjs` on Git Bash, PowerShell, and Command Prompt. Here's a screenshot from Git Bash:

ava-win10-test

However, I can't reproduce the issue on WSL running Ubuntu 22.04. The script name is correctly set to `ava`:

ava-wsl-test

Looking at yargs/yargs#538 and gatsbyjs/gatsby#9833, it seems this has something to do with the differences in how binaries are installed on Windows vs Unix-based systems. Looking at other similar issues on the Yargs repo (yargs/yargs#1010) and how Gatsby fixed the issue (gatsbyjs/gatsby#12186), the proposed solution for these kinds of problems appears to be to just set the script name using the `.scriptName()` method as this PR does.

novemberborn:
Thanks for clarifying @mcecode!

PR #3184, opened 2023-03-09
Modernize `require` config

sculpt0r:
Handled complex `options.require` in api path resolver and also in actual loader inside worker/base.

Could you also give some feedback if that is the right direction?

Closes: #1666 

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#1666: Support arguments in "require" configuration](https://issuehunt.io/repos/26820798/issues/1666)
---
</details>


lib/worker/base.js
```diff
@@ -176,7 +176,14 @@ const run = async options => {
...
 	try {
 		for await (const ref of (options.require || [])) {
-			await load(ref);
+			if (typeof ref === 'string') {
+				await load(ref);
+			} else if (Array.isArray(ref)) {
+				const [path, options] = ref;
+
+				const mod = await load(path);
+				mod.apply(null, ...options);
```
novemberborn, line 185: Need to check how this works with a mix of CJs and ESM. You may need to check if `mod` is callable, and if not, access a callable `default`.

novemberborn (commented):
@sculpt0r looks great!

lib/worker/base.js
sculpt0r: Thanks @novemberborn  - I'll correct that :)

lib/worker/base.js
```diff
@@ -176,7 +176,21 @@ const run = async options => {
...
-			await load(ref);
+			if (typeof ref === 'string') {
+				await load(ref);
+			} else if (Array.isArray(ref)) {
+				const [path, options] = ref;
+
+				const loadedModule = await load(path);
+
+				if (typeof loadedModule === 'function') {
+					loadedModule.apply(null, ...options);
```
novemberborn, line 187: Let's `await` this too, so it can do some async setup before we load the next module.

And below for `default.apply`.

Do we care about the `this` argument? Could `loadedModule(...options)` instead.

novemberborn:
> Not sure if such complication in config is needed? But if you want to load some third party module, that you can't control - you might want to call some specific part of it - without extracting it manually or making proxy file just for sake of re-export some part of it?

I think a proxy file would be most appropriate.

sculpt0r:
Hi @novemberborn, PR is ready for another verification:
- both invocations are now `await'ed`
- removed `null` as `this` argument in favour of using destructed `options`. Previously I thought that maybe we shouldn't mess with `this` at all at config level. But this solutions allows users to pass `null` if they don't want that.

lib/worker/base.js
```diff
@@ -176,7 +176,21 @@ const run = async options => {
...
-			await load(ref);
+			if (typeof ref === 'string') {
+				await load(ref);
+			} else if (Array.isArray(ref)) {
+				const [path, options] = ref;
+
+				const loadedModule = await load(path);
+
+				if (typeof loadedModule === 'function') {
+					await loadedModule.apply(...options);
```
novemberborn, line 187: This uses the first option as the `this` argument, so you'd always have to pass `null` which seems odd. I think it's better like:

```suggestion
					await loadedModule(...options);
```

sculpt0r:
@novemberborn - could you explain me, why there are two different imports `from ava` in tests?

There is `import test from '@ava/test';` and `import test from 'ava';` which one should I use?

sculpt0r:
Sorry for the long delay... I have some life-topics that needs to be handle.

I think I'll finish tests soon.

sculpt0r:
☝️ Rebased on the newest `main` branch.

test/config-require/fixtures/required-default/req.js
sculpt0r, line 1: @novemberborn  Could you lead me a little bit more?
Is this test construction valid in any way? My idea was to invoke this default exported function and at least see sth in the logs - but it seems to not working the way I thought...

novemberborn: The issue was that the test files were in a subdirectory, which means you have to specify the `cwd` option.

novemberborn:
Thanks @sculpt0r. I've done some more work on this, will push the commits soon.

This code has been a little overlooked in the past few years. For example it only works with CJS modules! I think the answer is to only support local files, and then resolve their paths from the project directories. But that's a breaking change. So we'll have to leave this pending a little while longer until we're ready to cut for AVA 6.

novemberborn:
Tracking as a draft until we're ready for AVA 6.

@sculpt0r what do you think of the changes?

lib/worker/base.js
```diff
@@ -175,22 +176,14 @@ const run = async options => {
...
-
-				const loadedModule = await load(path);
-
-				if (typeof loadedModule === 'function') {
-					await loadedModule(...options);
-				} else if (typeof loadedModule.default === 'function') {
-					const {default: fn} = loadedModule;
-					await fn(...options);
-				} else {
-					channel.send({type: 'non-invokable-require-option'});
```
sculpt0r: I'm not sure how important it could be to inform users that the `thing` they want to import wont be called at all. For example if they try to use some local file and forgot to export sth that should be called?

It could be better to see an error instead of silent run when something is not working 🤔

test/config-require/fixtures/with-arguments/test.js
```diff
@@ -0,0 +1,15 @@
...
+
+import cjs from './required.cjs';
+import {receivedArgs} from './required.mjs';
+
+test('receives arguments from config', t => {
+	t.deepEqual(receivedArgs, ['hello', 'world']);
+	t.deepEqual(cjs.receivedArgs, ['goodbye']);
+});
+
+test('ok to load for side-effects', async t => {
```
sculpt0r, line 11: It took me some time to understand that this test case take adventage of requesting the `side-effect` file before (during test loads config).

Maybe test case could be renamed? Sth like `side-effects are execute at first with test loading` or sth similar?

lib/worker/base.js
novemberborn: I don't think it's necessarily the case that you want to execute a function, so making that required behavior isn't great either.

sculpt0r:
Thanks @novemberborn for explanation and commits. That are the tests I tried to implement earlier - but I wasn't able...

> I think the answer is to only support local files, and then resolve their paths from the project directories. But that's a breaking change. So we'll have to leave this pending a little while longer until we're ready to cut for AVA 6.

We talked before about such resolution - so definitely - if that is a breaking change - let's wait for next major release.

I left two extra comments  to the latest changes. Not sure how important they are - so I didn't touch the code.

novemberborn:
We'll also want to test and document that you can pass, say, maps and uint8arrays. And verify how that works with child processes.

test/config-require/fixtures/with-arguments/test.js
novemberborn: Yea I'm fine with renaming them, I didn't think too hard about it just wanted to get the tests done in the little time I had.

novemberborn:
> > We'll also want to test and document that you can pass, say, maps and uint8arrays. And verify how that works with child processes.
> 
> Alright, then I add those tests since I think I know how :) Thanks for the tip :)

Nice!

We may find that we always JSON serialize the options, in which case `Map` won't survive the round trip. That's fine, but it'd be nicer to v8 serialize/deserialize. With worker threads that may be the default behavior, but for child processes that may require a change. In short I'm not really sure what we do at the moment. I think all other options are JSON compatible so it hasn't come up before.

With `ava.config.*` files you should be able to create maps for the `require` option.

All that said, just documenting the current behavior may suffice for now.

sculpt0r:
Rebased on the newest `main`.

sculpt0r:
> All that said, just documenting the current behavior may suffice for now.

TBH - I can't find any place that serialize config in any way? Since the config is loaded - it's passed almost unchanged to the API (besides `arrify` usage). And then the modules are loaded. So right now, I'm not sure if I should further investigate that for documenting purposes? In other words: do You think this PR requires more work? Or just until the `6.0` is ready -  just make sure it didn't break anything?

novemberborn:
> I can't find any place that serialize config in any way?

You're right, it was not explicit. With worker threads we already relied on "advanced" serialization and I've pushed a commit to enable that for child processes as well. Plus tests.

I also figured out a way to load regular dependencies, by writing a file into `node_modules` from which we can then do the `import()`.

I think this is now mostly compatible with AVA 5, except for the invocation of the default export which is new and could still have surprising side-effects.

novemberborn:
@sculpt0r let me know what you think of my changes! `main` has now dropped Node.js 14 so this can be merged.

sculpt0r:
@novemberborn  - sorry for being late... As I can see - the PR is already merged.

PR #3185, opened 2023-03-13
Support arrays in t.like() assertion

### LINKED ISSUE #2627: t.like not handling arrays as expected

tommy-mitchell:
Continues from #3023. Adds a test case (the repro from #2627) and handles the `Reflect.ownKeys(array).length` issue.

Fixes #2627.

It's been a couple months since someone last said they'd add this, so I went for it.

### DIFF
--- a/lib/like-selector.js
+++ b/lib/like-selector.js
@@ -1,14 +1,12 @@
-export function isLikeSelector(selector) {
-	const isArrayOrObject = selector !== null && typeof selector === 'object';
+const isObject = selector => Reflect.getPrototypeOf(selector) === Object.prototype;
 
-	if (!isArrayOrObject) {
+export function isLikeSelector(selector) {
+	if (selector === null || typeof selector !== 'object') {
 		return false;
 	}
 
-	const isArray = Array.isArray(selector);
-	const minimumKeysRequired = isArray ? 1 : 0;
-
-	return Reflect.ownKeys(selector).length > minimumKeysRequired;
+	const keyCount = Reflect.ownKeys(selector).length;
+	return (Array.isArray(selector) && keyCount > 1) || (isObject(selector) && keyCount > 0);
 }
 
 export const CIRCULAR_SELECTOR = new Error('Encountered a circular selector');
--- a/lib/like-selector.js
+++ b/lib/like-selector.js
@@ -1,9 +1,14 @@
 export function isLikeSelector(selector) {
-	const prototype = Reflect.getPrototypeOf(selector);
-	return selector !== null
-		&& typeof selector === 'object'
-		&& (prototype === Object.prototype || prototype === Array.prototype)
-		&& Reflect.ownKeys(selector).length > 0;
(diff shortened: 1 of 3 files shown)


tommy-mitchell:
Force pushed because my local user was setup incorrectly.

novemberborn:
Thanks @tommy-mitchell! I need to find some time to check the current behavior and your changes.

lib/like-selector.js
```diff
@@ -1,8 +1,14 @@
...
+	const isArrayOrObject = selector !== null && typeof selector === 'object';
+
+	if (!isArrayOrObject) {
+		return false;
+	}
+
+	const isArray = Array.isArray(selector);
+	const minimumKeysRequired = isArray ? 1 : 0;
+
+	return Reflect.ownKeys(selector).length > minimumKeysRequired;
```
novemberborn, line 11: I think we need to keep the check for `Object.prototype`, otherwise you could pass other non-array values like functions or maps.

Something like this may be more elegant (but probably doesn't satisfy the linter):

```suggestion
	if (selector === null || typeof selector !== 'object') return false;
	
	const keyCount = Reflect.ownKeys(selector).length;
	return Array.isArray(selector) && keyCount > 1 || Reflect.getPrototypeOf(selector) === Object.prototype && keyCount > 0;
```

novemberborn (commented):
Sorry for the long wait @tommy-mitchell. Work is pretty intense so I don't spend much time with computers outside of it.

This looks good, left a comment but the approach looks fine.

Could you also update the documentation at https://github.com/avajs/ava/blob/0574e5e6d7959b7dc506cb2204f432f73ba39451/docs/03-assertions.md#likeactual-selector-message?

The type definition also needs an overload: https://github.com/avajs/ava/blob/0574e5e6d7959b7dc506cb2204f432f73ba39451/types/assertions.d.cts#L166

lib/like-selector.js
tommy-mitchell: That's a much cleaner solution, there wasn't much to change for the linter.

novemberborn (commented):
Hi @tommy-mitchell, bit of an oopsy here. I went to check up on this PR and found a pending comment I never submitted 🤦

novemberborn (approved):

PR #3187, opened 2023-03-30
Recognize typical assertion errors and use their formatting

Irvenae:
Based on https://github.com/avajs/ava/pull/2230 which was mentioned in (incorrectly) resolved https://github.com/avajs/ava/issues/1277.


lib/test.js
```diff
@@ -7,6 +7,16 @@ import concordanceOptions from './concordance-options.js';
...
 import parseTestArgs from './parse-test-args.js';
 
+function isExternalAssertError(error) {
+	// Match errors thrown by <https://www.npmjs.com/package/expect>.
+	if (error.matcherResult) {
+		return true;
+	}
+
+	// Match errors thrown by <https://www.npmjs.com/package/chai> and <https://nodejs.org/api/assert.html>.
+	return error.hasOwnProperty('actual') && error.hasOwnProperty('expected');
```
Irvenae, line 17: I noticed the properties can have value `false` (which is the case in the test). So we should check existence of the property instead of the property. (which was the case in the original PR)

Irvenae:
I don't know what the expected way is to add the expect dependency in the test.
So all tests fail with `Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'expect' imported from /home/runner/work/ava/ava/test/external-assertions/fixtures/expect-failure.js␊`

novemberborn:
Nice one @Irvenae!

Fixture packages are not installed, those `package.json` files are mostly there so that the AVA-under-test does not walk up to the repository root.

It should be fine if you install `expect` as a dev dependency.

---

Could you share some screenshots of how these errors are then reported? And could you address the linter errors?

Irvenae:
You could see the result from the snapshot.
Before: 
Screenshot 2023-04-12 at 12 37 30

After: 
Screenshot 2023-04-12 at 12 43 57

lib/assert.js
```diff
@@ -27,7 +27,7 @@ function formatWithLabel(label, value) {
 	return formatDescriptorWithLabel(label, concordance.describe(value, concordanceOptions));
 }
 
-const hasOwnProperty = (object, prop) => Object.prototype.hasOwnProperty.call(object, prop);
+export const hasOwnProperty = (object, prop) => Object.prototype.hasOwnProperty.call(object, prop);
```
Irvenae, line 30: not sure we want to expose that here or place this in a more appropriate file (or duplicate worst case)

novemberborn: It's available as `Object.hasOwn()` in Node.js 16 which will be the minimal required version soon enough. Repeating it is fine for now, I'm sure https://github.com/sindresorhus/eslint-plugin-unicorn will flag it soon enough 😉

novemberborn:
In that screenshot, is the syntax highlighting coming from Jest itself?

test/external-assertions/test.js
```diff
@@ -0,0 +1,13 @@
+import test from '@ava/test';
+
+import {fixture} from '../helpers/exec.js';
+
+test('node assertion ', async t => {
+	const result = await t.throwsAsync(fixture(['assert-failure.js']));
+	t.snapshot(result.stdout);
```
novemberborn, line 7: Looks like the tests are failing because the error message contains carriage returns. IIRC AVA only emits line feeds which would explain why there's no automatic handling of the discrepancy.

```suggestion
	t.snapshot(result.stdout.replace(/\r/g, ''));
```

(I'd suggest `replaceAll('\r', '')` but as of right now AVA still targets Node.js 14 which doesn't support that.)

Irvenae:
> In that screenshot, is the syntax highlighting coming from Jest itself?

yes, we pass the error message as a `formatted` field in `values` of `AssertionError`

novemberborn (approved):

PR #3192, opened 2023-04-10
Update TypeScript recipe to remove ts-node recommendation

### LINKED ISSUE #2593: Update TypeScript recipe for ESM support

ondreian:
adds a recipe to use the `tsx` loader from esbuild

closes #2593;

### DIFF
--- a/docs/recipes/typescript.md
+++ b/docs/recipes/typescript.md
@@ -8,58 +8,21 @@ This guide assumes you've already set up TypeScript for your project. Note that 
 
 ## Enabling AVA's support for TypeScript test files
 
-### With precompile step
+Broadly speaking, there are two ways to run tests written in TypeScript:
 
-Out of the box AVA does not load TypeScript test files. You can use our [`@ava/typescript`] package, which is designed to work for projects that precompile TypeScript using the `tsc` command. Please see [`@ava/typescript`] for setup instructions.
+1. Build first, then test against the build output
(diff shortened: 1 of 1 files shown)


novemberborn:
I'd be happy to list `tsx`. Would it make sense to offer multiple suggestions and then have sections for each? Note that below the code example the recipe continues with `ts-node`.

ondreian:
> I'd be happy to list `tsx`. Would it make sense to offer multiple suggestions and then have sections for each? Note that below the code example the recipe continues with `ts-node`.

Good catch, I should probably remove that section.  I do not think a `ts-node` example makes sense because it doesn't work at all that I could find, is someone has a working example that would be a great addition though.

PR #3198, opened 2023-05-17
Remove use of the "del" package. This reduces total dependencies by 11%.

craigahobbs:
https://craigahobbs.github.io/npm-dependency-explorer/#var.vDirect=1&var.vName='ava'&var.vSort='Dependencies'

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -5,7 +5,6 @@ import process from 'node:process';
 
 import arrify from 'arrify';
 import ciParallelVars from 'ci-parallel-vars';
-import {deleteAsync} from 'del';
 import figures from 'figures';
 import yargs from 'yargs';
 import {hideBin} from 'yargs/helpers'; // eslint-disable-line n/file-extension-in-import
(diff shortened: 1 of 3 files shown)


novemberborn:
Thanks @craigahobbs. I don't think this materially impacts the dependency size (notice how `package-lock.json` wasn't changed) but we should use built-in methods when appropriate.

novemberborn (approved):

craigahobbs:
@novemberborn, package-lock.json is a reflection of all dependencies including devDependencies, which is a much larger dependency list. Most end-users users only require the non-devDependencies.

To measure the change in total non-devDependencies, do an "npm install --omit dev" and check the total installed packages count. I used the top-level node_modules file count as a reasonable approximation:

~~~
$ rm -rf node_modules
$ npm install --omit dev
$ ls -1 node_modules | wc -l
~~~

Here's what I get before/after:

**before: ** 143
**after: ** 127

In other words, after the change, for non-development usage, 16 fewer packages are installed.

novemberborn:
> Here's what I get before/after:
> 
> **before: ** 143 **after: ** 127
> 
> In other words, after the change, for non-development usage, 16 fewer packages are installed.

You can't compare it that way. Fewer dependencies may be hoisted because they're no longer shared, but that doesn't mean they're not installed.

craigahobbs:
Do you mean that, for any package that lists Ava as a development dependency, that just because we eliminate "del" from Ava's runtime dependencies that it might get installed by another of the package's development dependencies? And so on with sub-dependencies? If so, that's valid, but many packages will get some of the possible benefit (16).

PR #3209, opened 2023-06-05
Update t.like to support Symbol keys and ignore non-enumerable properties

### LINKED ISSUE #3208: `t.like` mishandles symbol-keyed and non-enumerable properties
This is actually two distinct but related issues affecting `t.like`, which is effectively a concordance deep comparison of a comparable value against a selector value, where the comparable is a filtered view of the actual input with respect to the selector.
1. Rejection of empty selector objects mistakenly allows objects based on the presence of non-enumerable properties that are ignored by concordance, resulting in inappropriate pass results from e.g. `t.like({}, Object.defineProperties({}, {x: {enumerable: false}}))`.
2. Use of `Object.entries(selector)` in `selectComparable` ignores symbol-keyed properties that are respected by concordance, resulting in inappropriate failure results from e.g. `t.like({[Symbol.toStringTag]: 'Custom'}, {[Symbol.toStringTag]: 'Custom'})`.

gibson042:
Fixes #3208

### DIFF
--- a/docs/03-assertions.md
+++ b/docs/03-assertions.md
@@ -165,7 +165,7 @@ t.like({
 You can also use arrays, but note that any indices in `actual` that are not in `selector` are ignored:
 
 ```js
-t.like([1, 2, 3], [1, 2])
+t.like([1, 2, 3, 4], [1, , 3])
 ```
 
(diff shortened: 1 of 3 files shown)


novemberborn (approved):
Nice one, thanks @gibson042 !

PR #3212, opened 2023-06-16
Fix circular selector detection in t.like()

### LINKED ISSUE #3205: `t.like()` selector must not contain circular references
Hi,
 
The following code works in ava version `5.2.0`:

```js
const arr = ['c1', 'c2']
t.like({a: 'a', b: 'b', c: ['c1', 'c2'], d: ['c1', 'c2']}, {b: 'b', d: arr, c: arr})
```

However, it does not work in version `5.3.0` and seems to be related to pull request #3185.

Receive this error:

```text
`t.like()` selector must not contain circular references
```

novemberborn:
The previous implementation tracked each object, even if not circular. Update to use a stack.

Fixes #3205.

### DIFF
--- a/lib/like-selector.js
+++ b/lib/like-selector.js
@@ -16,13 +16,7 @@ export function isLikeSelector(selector) {
 
 export const CIRCULAR_SELECTOR = new Error('Encountered a circular selector');
 
-export function selectComparable(actual, selector, circular = new Set()) {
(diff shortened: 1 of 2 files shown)


powru9-gobVok-nutguj (commented):
[]()

powru9-gobVok-nutguj:
[]()

PR #3229, opened 2023-07-30
Only treat native errors as errors

novemberborn:
* Remove is-error dependency
* Document edge case where `error instanceof Error` can be true, yet AVA does not recognize `error` as an error

See also https://github.com/avajs/ava/pull/2911 for an earlier attempt.

Includes a workaround for https://github.com/nodejs/node/issues/48716 so that we can treat worker errors as such.


novemberborn:
@drazisil I figured out the problem with https://github.com/avajs/ava/pull/2911!

PR #3232, opened 2023-08-13
feat: add `sourceMaps` option

aleclarson:
When this option is false, the --enable-source-maps Node.js flag is not set for worker threads.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -60,6 +60,11 @@ const FLAGS = {
 		description: 'Run tests serially',
 		type: 'boolean',
 	},
+	'source-maps': {
+		coerce: coerceLastValue,
+		description: 'Enable source maps support (true by default)',
+		type: 'boolean',
+	},
 	tap: {
 		alias: 't',
 		coerce: coerceLastValue,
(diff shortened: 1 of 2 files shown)


aleclarson:
Okay, the `execArgv` no longer contains `--enable-source-maps`, but I'm still seeing translated stacks. 😳

sindresorhus:
What problem does it solve?

aleclarson:
Tried using `ts-node` instead, but the suggested setup seems out of date? Couldn't get it working, even after removing the `configurableModuleFormat` experiment option.

```json
{
  "ava": {
    "extensions": {
      "ts": "module"
    },
    "nonSemVerExperiments": {
      "configurableModuleFormat": true
    },
    "nodeArguments": [
      "--loader=ts-node/esm"
    ]
  }
}
```

aleclarson:
> What problem does it solve?

I'm writing a compiler, so I need to see the output code to get a clearer view of what's going wrong at runtime.

aleclarson:
> Okay, the `execArgv` no longer contains `--enable-source-maps`, but I'm still seeing translated stacks. 😳
> 
> Possibly related to `--loader=tsx` 🤔

This is indeed related to the "tsx" loader.

https://github.com/esbuild-kit/core-utils/blob/develop/src/source-map.ts

novemberborn:
I agree there should be a way to not use the built-in source map support. However I don't think it should be a CLI flag, it should go into the config. I'm wary of booleans since the behavior change is quite subtle. Perhaps `sourceMaps: 'built-in' | 'manual'`.

novemberborn:
Please re-open if you have a chance to return to this.

PR #3233, opened 2023-08-14
Make `assert`, `truthy` and `falsy` typeguards

ZachHaber:
This will make ava easier to use with typescript and allow for removing redundant checks.

```ts
// with change
const actual = 3 as number | undefined;
if (t.assert(actual)) {
    // actual is `number` here
} else {
    // actual is `undefined` here
}
```
```ts
// before change
const actual = 3 as number | undefined;
t.assert(actual);
// Need a separate if branch from the assertion
if (actual) {
    // actual is `number` here
} else {
    // actual is `undefined` here
}
```

Since ava is a library and to reduce any type errors in existing projects, it's best to use the most exact approach to the typeguards as possible. For most use cases something akin to `actual is NonNullable<T>` is completely valid, but in this case it's very likely that things like `boolean` will be passed in.

Unfortunately typescript doesn't have convenient `Falsy` or `Truthy` utility types, so the best approach I've seen so far is to enumerate all possibilities (except `NaN`) to make up `Falsy` and then use that to Exclude `Falsy` types. `NaN` doesn't have it's own type, so the typeguard won't be able to differentiate between them no matter what we do.

### DIFF
--- a/types/assertions.d.cts
+++ b/types/assertions.d.cts
@@ -124,7 +124,7 @@ export type Assertions = {
 	 * Assert that `actual` is [truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy), returning a boolean
 	 * indicating whether the assertion passed.
 	 * 
-	 * Note: with typescript, an `else` clause using this as a typeguard will be subtly incorrect for string and number types and will not give `0` or `''` as a potential value in an `else` clause
+	 * Note: An `else` clause using this as a type guard will be subtly incorrect for `string` and `number` types and will not give `0` or `''` as a potential value in an `else` clause.
 	 */
 	truthy: TruthyAssertion;
 };
@@ -137,7 +137,7 @@ export type AssertAssertion = {
 	 * Assert that `actual` is [truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy), returning a boolean
 	 * indicating whether the assertion passed.
 	 * 
-	 * Note: with typescript, an `else` clause using this as a typeguard will be subtly incorrect for string and number types and will not give `0` or `''` as a potential value in an `else` clause
+	 * Note: An `else` clause using this as a type guard will be subtly incorrect for `string` and `number` types and will not give `0` or `''` as a potential value in an `else` clause.
 	 */
 	<T>(actual: T, message?: string): actual is T extends Falsy<T> ? never : T;
 
@@ -346,7 +346,7 @@ export type TruthyAssertion = {
 	 * Assert that `actual` is [truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy), returning a boolean
 	 * indicating whether the assertion passed.
 	 * 
-	 * Note: with typescript, an `else` clause using this as a typeguard will be subtly incorrect for string and number types and will not give `0` or `''` as a potential value in an `else` clause
+	 * Note: An `else` clause using this as a type guard will be subtly incorrect for `string` and `number` types and will not give `0` or `''` as a potential value in an `else` clause.
 	 */
 	<T>(actual: T, message?: string):  actual is T extends Falsy<T> ? never : T;
 
--- a/types/assertions.d.cts
+++ b/types/assertions.d.cts
@@ -27,6 +27,8 @@ export type Assertions = {
 	/**
 	 * Assert that `actual` is [truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy), returning a boolean
 	 * indicating whether the assertion passed.
+	 * 
+	 * Note: with typescript, an `else` clause using this as a typeguard will be subtly incorrect for string and number types and will not give `0` or `''` as a potential value in an `else` clause
 	 */
 	assert: AssertAssertion;
 
@@ -121,18 +123,23 @@ export type Assertions = {
 	/**
 	 * Assert that `actual` is [truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy), returning a boolean
 	 * indicating whether the assertion passed.
+	 * 
+	 * Note: with typescript, an `else` clause using this as a typeguard will be subtly incorrect for string and number types and will not give `0` or `''` as a potential value in an `else` clause
 	 */
 	truthy: TruthyAssertion;
 };
 
-type Falsy = false | 0 | 0n | '' | null | undefined;
+type FalsyValue = false | 0 | 0n | '' | null | undefined;
+type Falsy<T> = T extends Exclude<T, FalsyValue> ? (T extends number | string | bigint ? T & FalsyValue : never) : T;
 
 export type AssertAssertion = {
 	/**
 	 * Assert that `actual` is [truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy), returning a boolean
 	 * indicating whether the assertion passed.
+	 * 
+	 * Note: with typescript, an `else` clause using this as a typeguard will be subtly incorrect for string and number types and will not give `0` or `''` as a potential value in an `else` clause
 	 */
-	<T>(actual: T, message?: string): actual is Exclude<T, Falsy>;
+	<T>(actual: T, message?: string): actual is T extends Falsy<T> ? never : T;
 
 	/** Skip this assertion. */
 	skip(actual: any, message?: string): void;
@@ -194,7 +201,7 @@ export type FalsyAssertion = {
 	 * Assert that `actual` is [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy), returning a boolean
 	 * indicating whether the assertion passed.
 	 */
-	<T>(actual: T, message?: string): actual is T extends Falsy ? T :never;
+	<T>(actual: T, message?: string): actual is Falsy<T>;
 
 	/** Skip this assertion. */
 	skip(actual: any, message?: string): void;
@@ -338,8 +345,10 @@ export type TruthyAssertion = {
 	/**
 	 * Assert that `actual` is [truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy), returning a boolean
 	 * indicating whether the assertion passed.
+	 * 
+	 * Note: with typescript, an `else` clause using this as a typeguard will be subtly incorrect for string and number types and will not give `0` or `''` as a potential value in an `else` clause
 	 */
-	<T>(actual: T, message?: string): actual is Exclude<T, Falsy>;
+	<T>(actual: T, message?: string):  actual is T extends Falsy<T> ? never : T;
 
 	/** Skip this assertion. */
 	skip(actual: any, message?: string): void;
--- a/types/assertions.d.cts
+++ b/types/assertions.d.cts
@@ -125,7 +125,7 @@ export type Assertions = {
 	truthy: TruthyAssertion;
 };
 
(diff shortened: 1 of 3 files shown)


types/assertions.d.cts
```diff
@@ -125,12 +125,14 @@ export type Assertions = {
 	truthy: TruthyAssertion;
 };
 
+type Falsy = false | 0 | -0 | 0n | '' | null | undefined;
```
sindresorhus, line 128: I don't think `-0` is a unique type in TS.

ZachHaber: You're right about that! It just resolves to 0, so was basically a no-op.

novemberborn:
Neat! @sindresorhus you OK reviewing?

types/assertions.d.cts
```diff
@@ -192,7 +194,7 @@ export type FalsyAssertion = {
 	 * Assert that `actual` is [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy), returning a boolean
 	 * indicating whether the assertion passed.
 	 */
-	(actual: any, message?: string): boolean;
+	<T>(actual: T, message?: string): actual is T extends Falsy ? T :never;
```
sindresorhus, line 197: I don't think `never` is correct here.

types/assertions.d.cts
```diff
@@ -125,12 +125,14 @@ export type Assertions = {
 	truthy: TruthyAssertion;
 };
 
+type Falsy = false | 0 | 0n | '' | null | undefined;
```
sindresorhus, line 128:
```suggestion
type Falsy = false | 0 | 0n | '' | null | undefined;
type Truthy<T> = Exclude<T, Falsy>;
```

just for readability.

types/assertions.d.cts
ZachHaber: I don't think we can use anything besides `never` there and have it work correctly. The only things that typescript doesn't complain about using there are `never` and `any` or a subset of the `T` type. If you make it `any` then `falsy(true)` thinks it's falsy.

This is also the exact opposite condition I setup in `truthy` and `assert`:
`actual is Exclude<T,Falsy>` where Exclude is `Exclude<T,U> = T extends U ? never : T` becomes `actual is T extends Falsy ? never : T` which is the opposite condition of the `falsy` assertion.

I included the test-types with both the `if` case and the `else` case so you can see the types as they work. For further showcasing of the types, I setup a quick [Playground example](https://www.typescriptlang.org/play#code/C4TwDgpgBAYghgGwM4igXigM0U6AfKABigMIDsSoByKysgVwQUvrIBMJMBLMiNgbgBQgzKwDGwLgHsK2ZCAA8AFQB8ACjgT6iAFxQlASj2bg25lyT6oEAB7AI7S-HlQA-Fb28AbhABOUAG9BKBCoXwhTXwoAQhMzIQBfIVEyCWkKYF96YAALRVUNLV19Iyg4xCgLKABRGzEEeg5lABpYHBAVINCwiPooqGjYooREoTEZJGB9LOgMTPoIMYmp51x0LBxFwQB6bagAYThJqFyqmWgkKX0AZSg2KQgkMiopzNRgK4BrCAgwSqmkJAxFxuGJBOMyMcAEZSKTMDByNZHKAwuEQOBkIQ7PaI1ApNIyYQgqBqXFqJQzAwGQLBUK7fQ5KpgOC5KAAdykjDYUG8figOTgYEgZFpIQpC1FUHpAD1XIIEtZkNAuqFxRBJTK5QlBMTSe01KsIFSad1DRq9rL5Yq1iqQvSlIzLMzWRyuTyID5-AKhQ5JWbupr5TrMCSyaiEMbbSjYQhzVBLQqIEqTaFw3GE8J7VlcnjxJJCcGSfMc+TKdSo2r01rrcq4w6mSycuzOQhubyvYLhZLKwGLVrC2pi3kDZtI3XHVBnU3Xa33Z7+Z3fabNlWrUmbX6V734-3dUOQGpw2PumntwmaymQqe6X35UA) if you want to try changing what the assertions are.

Playing around with it, the current conditions don't work well for number and string primitives. Typescript seems to not recognize that `0` is a subset of number unless you use an `as const` expression with it. Similar results for an empty string (`''`) but even more wacky as `''` will make typescript think it is `truthy` with a value of `''`, but `'' as const` makes typescript recognize it as falsy. Typescript also doesn't separate out `0` or `''` from their primitive values.

So far, I think I'm having the most accurate approach using
```ts
function falsy<T>(actual: T): actual is T extends Exclude<T, Falsy>? never : T {
    return !actual;
};
function truthy<T>(actual: T): actual is Exclude<T, Falsy> {
    return !!actual;
};
```

In which case the only remaining issue is that for a type of `number` or `string`, typescript doesn't think that they could ever be falsy.

sindresorhus, line 197: Does it work to have specific overloads?
```suggestion
	(actual: 0): true;
	(actual: ''): true;
	<T>(actual: T, message?: string): actual is T extends Falsy ? T :never;
```

sindresorhus: > I don't think we can use anything besides never there and have it work correctly.

You're right. I read the signature incorrectly.

> So far, I think I'm having the most accurate approach using ([playground link](https://www.typescriptlang.org/play#code/C4TwDgpgBAYghgGwM4igXigM0U6AfKABigMIDsSoByKysgVwQUvrIBMJMBLMiNgbgBQgzKwDGwLgHsK2ZCAA8AFQB8ACjgT6iAFxQlASj2bg25lyT6oEAB7AI7SwFEbYhPQ7KANLBwgVUAD8ULwAbhAATlB6SlAA3oJQSVAREKYRFACEJmZCAL5ComQS0hTAEfTAABaKqhpauvpGUDmIUBZQLm4eEN6+8gEJySlp9BlQmdkNCPlCYjJIwPoV1SAAaoj00BgAjHMLS-DyG+7bREIA9BdQAMJwi1DVHTLQSFL6AMpQbFIQSGRUJblVDAd4AawgEDA7SWSEgYi43DEgnmZAeoiYSnAED0DAAtgAjSLoIgtSz4okRISCRFQNRKFY1E5bAzxRLJBmVJmbCDspJXKAAPUCgjy1mQ0CGHMZ6x5fKgAuFoppmDp5S5IHpMuZEAMrKlSU5qx18sVIrFEAlbOGRu5p1N1yVeRVdLkKC1Gp1eut0s9cuGZtF4twPsN2v9yUDzpdaiOKC9of68Yj-Md5uDkvlcdl9oDaeVtLU6tWsb8Xv1WbLKYV+YtVoNSZzWwdQvNMbdmuz5cTXerUYzParucjteEhYxCCxkArwwnU95edbQctIYbc+xLadMeLNTU6+nM+S+4XI6XddX8uPm7b478e8Yk+x3rXD-n1+X9cvr43i6dQA))

👍

sindresorhus, line 197:
```suggestion
	<T>(actual: T, message?: string): actual is T extends Falsy ? T : never;
```

sindresorhus (changes requested):

types/assertions.d.cts
ZachHaber: With your suggestion of multiple overloads, that partially works, but runs into an issue where union types would fall back to the base behavior of the generic rather than the behavior of the overload.

I think I've cracked it for falsy, but I can't get the same to work for truthy as Typescript has no way to just invert a type. [playground](https://www.typescriptlang.org/play#code/C4TwDgpgBAYghgGwM4gGqIK7QLxQGaJLQA+UADFKWQHaVQDk9d1GCCdG1AJhHgJbUIXANwAoUJFiEQAHgAqAPii45UCAA9gEbkigBRdQGMEGHvIA0U5GkwQlAfigAKVRq06oLALYAjCACc6JGB-AQBzOh8+MIFgKEdVADIrFHQTaAAuTwgANwCASigsuTFRPE5DYD4Ae1oCa3kFJzhKjERi-KyW4Db2Pl14BsUoAG9RKAmofwge-1oAQm7esQBfMXLqSpraEIxgAAtZRWbW9qg5TqglxCh+87VNbS4B6Ub47LzA4tHxyenZhaLU4IVZiQy1YLnfx7Q5pLDKBj7Pj0MEQuKDVK2BGMMQAelxUAAwnBIQc7rVoEhqucAMpQLjVCBIaj0OIhEBQYDUgDWEAgYFucSQkEMfH4hlE4OokPKbDk4AgWWCoWoEVI3j8gVwjCuumV4WYGF8AVKYuccmhBxs6UKY0mUJh1qwvwm+KgAD17KIVmpkNA7ZMLY64RAXVA3Z7vaIzU5dlaXJbYbZ8raw0GrSGwxGvT6IH6fvb00n0lmCZGVtG8M56igE8Hk6nC4mnaH7dnvb6iAXA83M22yznRJXnBiW43JqO+5N27n8wGJpPbKWPYOY3HDk5FzbxwvpFPXQOO3mu-OUi3l+Xh04ayBN3uG93d9Z9+HD7OT2Gt87+yuo1fZQg8qQDu+CsIBCoXjmnb+mGAFAa205vle663nBCopo+oFyhBP7ltBmFoZAkFRjGN7XmB8EYaehEIQev7vjB9o0cRFZSpCORYrgFAkp4Rqahw3C8AIQhBCEBrqmBdAAEoQOC-hcDI+qqpYnDctQ1QAO7UEoVC0KQKC+NU7CkAAgv4-hwLIcDUCAChiKR0hOBx26Yc5WBiLhUHHoxkxuRAHmIfRyHNk5D6nn5AV0Xh3mubYkWvvRQA) - As you can see from the screenshot, falsy works great, but truthy is missing `0` in the `else` clause, and I don't think there's really anything we can do about it...
image


```ts
type FalsyValue = false | 0 | 0n | '' | null | undefined;
type Falsy<T> = T extends Exclude<T, FalsyValue> ? (T extends number | string | bigint ? T & FalsyValue : never) : T;

function falsy<T>(actual: T): actual is Falsy<T> {
    return !actual;
};
function truthy<T>(actual: T): actual is T extends Falsy<T> ? never : T {
    return !!actual;
};
```

sindresorhus: This looks good 👍

sindresorhus: > but truthy is missing 0 in the else clause, and I don't think there's really anything we can do about it...

I would mention this in the doc comments.

ZachHaber:
I gave it my best shot for a reasonably concise note about the type issue for truthy/assert. The nice thing about the issues being where they are is that if they are reaching that issue, then that means the test failed anyway... 🙃

types/assertions.d.cts
```diff
@@ -27,6 +27,8 @@ export type Assertions = {
 	/**
 	 * Assert that `actual` is [truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy), returning a boolean
 	 * indicating whether the assertion passed.
+	 * 
+	 * Note: with typescript, an `else` clause using this as a typeguard will be subtly incorrect for string and number types and will not give `0` or `''` as a potential value in an `else` clause
```
sindresorhus, line 31: Metioning TypeScript here is moot as it's implied by it being in the type definition here and also the mention of type guard.

sindresorhus, line 31:
```suggestion
	 * Note: An `else` clause using this as a type guard will be subtly incorrect for `string` and `number` types and will not give `0` or `''` as a potential value in an `else` clause.
```

sindresorhus (changes requested):

PR #3234, opened 2023-08-15
Add `t.unorderedEqual()` assertion

### LINKED ISSUE #3020: deepEqual is inconsistent for containers (Object, Set, Map)
deepEqual take insertation order in account for Sets and Maps but not for Objects. Therefore `t.deepEqual(new Set([1,2]), new Set([2,1])` will fail while `t.deepEqual({a:1, b:2}, {b:2, a:1})` succeed. Container classes should obey the same rules, same assertations should work the same for all containers. 

I understand  object iteration order is undefined (arbitrary) according to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in), while Sets and Maps are OrderedSets and OrderedMaps in javascript. Probably this is the basis of the current implementation, but please be aware that order of Object.getOwnPropertyNames, Object.getOwnPropertySymbols and Reflect.ownKeys return values *are* defined in ES6 therefore elements of objects are pretty much ordered.

The current assertation rules are unfortunately inconsistent for containers, and there is no (easy) way to test (unordered) Sets and Maps. 

Sets are usually used as mathematical sets, where ordering *shouldn't matter*. Also Maps are used as generic containers, trees, caches where failing on order difference is more like a bug than a feature. To be honest there are very few, rare edge-cases where one would like to test for Map/Set ordering (eliminating duplicates from arrays is the only one came into my mind). If one would need ordering, insertion order is almost certainly won't suffice and some more advanced structure like red-black tree or alike will be implemented instead anyway). 

Ideal solution would be if ES would have both Sets and OrderedSets, but unfortunately this is not the case. Changing deepEqual implementation certainly would break some existing test cases, which is very unfortunate. Therefore probably the only viable way is to add a new assertation rule for unordered cases and document the fact Object are not considered ordered collections.

tommy-mitchell:
Closes #3020

Adds a new assertion for testing equality in an unordered manner between two `Map`s or two `Set`s/`Array`s:

```js
test('unorderedEqual', t => {
	t.unorderedEqual([1, 2, 3], new Set([2, 3, 1]));

	t.unorderedEqual(
		new Map([['a', 1], ['b', 2], ['c', 3]]),
		new Map([['b', 2], ['c', 3], ['a', 1]]),
		'Map keys should be equal!',
	);
});
```

---

Todos:

- Improve error messages, format values with `concordance`
- Improve types (e.g. `Actual extends Map | Set | Array`)
	- Add type tests
- Fix failing tests


docs/03-assertions.md
```diff
@@ -137,6 +137,10 @@ Assert that `actual` is deeply equal to `expected`. See [Concordance](https://gi
 
 Assert that `actual` is not deeply equal to `expected`. The inverse of `.deepEqual()`. Returns a boolean indicating whether the assertion passed.
 
+### `.unorderedEqual(actual, expected, message?)`
+
+Assert that all values in `actual` are in `expected`, returning a boolean indicating whether the assertion passed.
```
novemberborn, line 142: Could you clarify the differences with `deepEqual`?

types/assertions.d.cts
```diff
@@ -123,6 +123,9 @@ export type Assertions = {
 	 * indicating whether the assertion passed.
 	 */
 	truthy: TruthyAssertion;
+
+	/** Assert that all values in `actual` are in `expected`, returning a boolean indicating whether the assertion passed. */
```
novemberborn, line 127: Here too let's clarify the differences with `deepEqual`.

lib/unordered-equal.js
```diff
@@ -0,0 +1,22 @@
+export const checkValueForUnorderedEqual = value => {
+	/* eslint-disable indent, operator-linebreak, unicorn/no-nested-ternary */
+	const type = (
+		value instanceof Map ? 'map' :
+		value instanceof Set ? 'set' :
+		Array.isArray(value) ? 'array' :
+		'invalid'
+	);
+	/* eslint-enable indent, operator-linebreak, unicorn/no-nested-ternary */
```
novemberborn, line 9: This probably does read better as a series of `if` statements that are assigning to a let-bound `type` variable.

lib/assert.js
```diff
@@ -958,5 +959,100 @@ export class Assertions {
...
+				return false;
+			}
+
+			const actualInfo = checkValueForUnorderedEqual(actual);
+
+			if (!actualInfo.isValid) {
+				fail(new AssertionError({
+					assertion: 'unorderedEqual',
+					improperUsage: true,
+					message: '`t.unorderedEqual` only compares Maps, Sets, and arrays',
```
novemberborn, line 974:
```suggestion
					message: '`t.unorderedEqual()` only compares arrays, maps and sets',
```

lib/assert.js
```diff
@@ -958,5 +959,100 @@ export class Assertions {
...
+				return false;
+			}
+
+			const expectedInfo = checkValueForUnorderedEqual(expected);
+
+			if (!expectedInfo.isValid) {
+				fail(new AssertionError({
+					assertion: 'unorderedEqual',
+					improperUsage: true,
+					message: '`t.unorderedEqual` only compares Maps, Sets, and arrays',
```
novemberborn, line 986:
```suggestion
					message: '`t.unorderedEqual()` only compares arrays, maps and sets',
```

lib/assert.js
```diff
@@ -958,5 +959,100 @@ export class Assertions {
...
+					values: [formatWithLabel('Called with:', expected)],
+				}));
+				return false;
+			}
+
+			if (actualInfo.size !== expectedInfo.size) {
+				fail(new AssertionError({
+					assertion: 'unorderedEqual',
+					message: 'size must be equal',
```
novemberborn, line 995: I think this should be the `message` argument. You can then do something like this:

https://github.com/avajs/ava/blob/a39dab098154c8d4fbfd39a85a72e13e71004c7e/lib/assert.js#L332-L333

lib/assert.js
```diff
@@ -958,5 +959,100 @@ export class Assertions {
...
+			if (actualInfo.type === 'map') {
+				if (expectedInfo.type !== 'map') {
+					fail(new AssertionError({
+						assertion: 'unorderedEqual',
+						message: 'both must be maps',
+					}));
+					return false;
+				}
+
+				const comparedKeysResult = concordance.compare(actual.keys, expected.keys, concordanceOptions);
```
novemberborn, line 1009: Did you mean:

```suggestion
				const comparedKeysResult = concordance.compare(actual.keys(), expected.keys(), concordanceOptions);
```

Of course that returns an iterator, which Concordance may walk over (can't recall), but the keys won't be in the same order.

If both values have the same size, then you probably don't need this?

novemberborn (commented):
I think we could make this more generic, require both values to have an `entries()` method that returns an `Iterator`. Each result should then be an array with a key and a value.

You can make a map of the results for `expected` and then iterate through `actual`.

If we want we could then still require both values to have the same `constructor` but it may be interesting not to.

novemberborn:
Please re-open if you have a chance to return to this.

lib/assert.js
tommy-mitchell: Agh - the joys of plain JavaScript.

> If both values have the same size, then you probably don't need this?

Good catch, I think you're right as well.

tommy-mitchell:
@novemberborn I don't think I can reopen by myself.

docs/03-assertions.md
tommy-mitchell: Thoughts?

> Assert that all values in `actual` are in `expected`. This is a variant of `.deepEqual()` that does not depend
on the order of `actual`/`expected` and only compares instances of `Map`s or `Set`s/arrays.
>
> The size of `actual` and `expected` must be equal. For `Map`s, each key-value pair in `actual` must be in
`expected`, and vice-versa. For `Set`s/arrays, each value in `actual` must be in `expected`.
>
> Returns `true` if the assertion passed and throws otherwise.

novemberborn:
> I don't think I can reopen by myself.

@tommy-mitchell done!

novemberborn:

PR #3241, opened 2023-08-28
Revert "Add banner to support Ukraine"

Heartnett:
Reverts avajs/ava#2992

# Please don't bring political matters into the code or repos.

### ;TLDR

- Not picking a fight or side.
- Need to keep code and repos free of political stuff

I really want to use this library again, but neither I nor my clients want anything to do with the political agendas (past, present, or future) in other countries. 

I understand this is a controversial and polarizing subject, and so do my clients. **Like every peace lover in the world, we don't condone war and/or violence.** 

The expectation is to keep code (**and their repositories**) free from any political subject matter (wars, invasions, [world] government sponsored killings/invasions/wars/racism/discrimination/equity/genocide/ideologies/lockdowns/"vaccines"/over-reach, etc.).

### If there is a need to support and raise awareness for them, by all means go for it, but don't bring this stuff into code and/or repositories. **It is not appropriate** - just like how wars, invasions, government over-reach, etc. are not appropriate.

I'm fully aware that I am going to receive crap for this. **I am not picking a fight or a side.**

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1,5 +1,3 @@
-[![SWUbanner](https://raw.githubusercontent.com/vshymanskyy/StandWithUkraine/main/banner2-direct.svg)](https://vshymanskyy.github.io/StandWithUkraine/)
-
 # <img src="media/header.png" title="AVA" alt="AVA logo" width="530">
 
 AVA is a test runner for Node.js with a concise API, detailed error output, embrace of new language features and process isolation that lets you develop with confidence 🚀


novemberborn:
Software, especially an open source project like this one, is made by people. Sometimes people who see their country invaded as is the case for a former contributor to AVA.

Will we use this platform to call out any and all injustice? No, that's not what this project is about. Will we stand by the people we've worked closely with in the past? Yes.

If it makes you uncomfortable to be reminded about the war against Ukraine then I think the banner has achieved its purpose.

Issue #2317, opened 2019-12-23
The Debugging with WebStorm recipe needs updating for the new debug command


novemberborn:
As of #2316, the upcoming v3 release has a built-in `debug` command. I've updated two of the existing debugging recipes but I'm not a WebStorm user, so this recipe still needs updating: https://github.com/avajs/ava/blob/master/docs/recipes/debugging-with-webstorm.md

hikyaru-suzuki:
@novemberborn Hi!
Is this issue related https://github.com/avajs/ava/issues/1787 ?
I have been failing debugger after upgraded ava to v3.
But I found how to do debug on WebStorm.
I using
- WebStorm 2019.3.4
- AVA 3.5.0

You will be able to do debug mode by this capture.
![image](https://user-images.githubusercontent.com/43962447/77047037-6335b180-6a07-11ea-8975-e976cba1fdff.png)


I didn't use npm Configuration but Node.js Configuration.
You do not need to debug mode on ava. Only specify path to test file and click debug button.

ATTENTION:
- WebStorm will run 2 Debugger Listener (I don't know that reason...)
- ava have been killed process 10 seconds from v3. Because I do not ran  ava's debug mode, You should set option about timeout.
  - release note: https://github.com/avajs/ava/releases/tag/v3.0.0
  - document: https://github.com/avajs/ava/blob/master/docs/07-test-timeouts.md

@erikkemperman said this at https://github.com/avajs/ava/issues/1787. I think so too.
> JetBrains are now doing some internal cleverness around debug ports

hikyaru-suzuki:
@novemberborn Hi!
Is this issue related https://github.com/avajs/ava/issues/1787 ?
I have been failing debugger after upgraded ava to version 3.
But I found how to do debug on WebStorm.
I using
- WebStorm 2019.3.4
- AVA 3.5.0

You will be able to do debug mode by this capture.
![image](https://user-images.githubusercontent.com/43962447/77047541-3b931900-6a08-11ea-8a33-e3990bf870b4.png)

I didn't use npm Configuration but Node.js Configuration.
You do not need to debug mode on ava. Only specify path to test file and click debug button.

ATTENTION:
- WebStorm will run 2 Debugger Listener (I don't know that reason...)
- ava have been killed process 10 seconds from v3. Because I didn't run  ava's debug mode, you should set option about timeout.
  - release note: https://github.com/avajs/ava/releases/tag/v3.0.0
[...]

PR #3242, opened 2023-08-30
Update WebStorm (lack of) support

dandv:


novemberborn:
Thanks for the PR @dandv. Lobbying Jetbrains seems reasonable, but it'd be better to update the recipe (https://github.com/avajs/ava/issues/2317 has been open for three years) or if it's impossible to delete the recipe.

PR #3247, opened 2023-09-11
Expose internal events for custom reporters via config

codetheweb:
We've been using a similar approach to this internally on a patched version of AVA for the last few weeks to pipe test runs to Datadog and it's been working great.

Alternatives:

- [use a Shared Worker](https://github.com/avajs/ava/discussions/3004): I think the solution here is much simpler than introducing a third type of worker (Shared Workers, Test Workers, and Suite(?) Workers) and covers most use cases I can think of.
- use [Diagnostic Channels](https://nodejs.org/api/diagnostics_channel.html): I'm not very familiar with this built-in, but looks like it could help reduce the performance impact of this?

Happy to add TS definitions for events if this is the route we want to go.

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -471,10 +471,24 @@ export default async function loadCli() { // eslint-disable-line complexity
 		});
 	}
 
-	api.on('run', plan => {
+	api.on('run', async plan => {
+		if (combined.onInternalEvent) {
+			await combined.onInternalEvent({
+				type: 'run',
+				plan,
+			});
+		}
+
 		reporter.startRun(plan);
 
-		plan.status.on('stateChange', evt => {
+		plan.status.on('stateChange', async evt => {
+			if (combined.onInternalEvent) {
+				await combined.onInternalEvent({
+					type: 'stateChange',
+					stateChange: evt,
+				});
+			}
+
 			if (evt.type === 'end' || evt.type === 'interrupt') {
 				// Write out code coverage data when the run ends, lest a process
 				// interrupt causes it to be lost.
--- a/test/internal-events/fixtures/ava.config.js
+++ b/test/internal-events/fixtures/ava.config.js
@@ -1,0 +1,16 @@
+import fs from 'node:fs/promises';
+
+const internalEvents = [];
+
+export default {
+	files: [
+		'test.js',
+	],
+	async onInternalEvent(event) {
+		internalEvents.push(event);
+
+		if (event.type === 'stateChange' && event.stateChange.type === 'end') {
+			await fs.writeFile('internal-events.json', JSON.stringify(internalEvents));
+		}
+	},
+};
--- a/test/internal-events/fixtures/.gitignore
+++ b/test/internal-events/fixtures/.gitignore
@@ -1,0 +1,1 @@
+internal-events.json
--- a/test/internal-events/fixtures/package.json
+++ b/test/internal-events/fixtures/package.json
@@ -1,0 +1,3 @@
+{
+	"type": "module"
+}
--- a/test/internal-events/fixtures/test.js
+++ b/test/internal-events/fixtures/test.js
@@ -1,0 +1,5 @@
+import test from 'ava';
+
+test('placeholder', t => {
+	t.pass();
+});
--- a/test/internal-events/test.js
+++ b/test/internal-events/test.js
@@ -1,0 +1,38 @@
+import fs from 'node:fs/promises';
+import {fileURLToPath} from 'node:url';
+
+import test from '@ava/test';
+
+import {fixture} from '../helpers/exec.js';
+
+test('internal events are emitted', async t => {
+	await fixture();
+
+	const result = JSON.parse(await fs.readFile(fileURLToPath(new URL('fixtures/internal-events.json', import.meta.url))));
+
+	t.like(result[0], {
+		type: 'run',
+		plan: {
+			files: [
+				fileURLToPath(new URL('fixtures/test.js', import.meta.url)),
+			],
+		},
+	});
+
+	const testPassedEvent = result.find(event => event.type === 'stateChange' && event.stateChange.type === 'test-passed');
+	t.like(testPassedEvent, {
+		type: 'stateChange',
+		stateChange: {
+			type: 'test-passed',
+			title: 'placeholder',
+			testFile: fileURLToPath(new URL('fixtures/test.js', import.meta.url)),
+		},
+	});
+
+	t.like(result[result.length - 1], {
+		type: 'stateChange',
+		stateChange: {
+			type: 'end',
+		},
+	});
+});


lib/cli.js
```diff
@@ -471,10 +471,24 @@ export default async function loadCli() { // eslint-disable-line complexity
 		});
 	}
 
-	api.on('run', plan => {
+	api.on('run', async plan => {
+		if (combined.onInternalEvent) {
+			await combined.onInternalEvent({
```
novemberborn, line 476: There isn't much use in awaiting this, and I'd use optional chaining:

```suggestion
		combined.onInternalEvent?.({
```

test/internal-events/fixtures/ava.config.js
```diff
@@ -0,0 +1,16 @@
+import fs from 'node:fs/promises';
+
+const internalEvents = [];
+
+export default {
+	files: [
+		'test.js',
+	],
+	async onInternalEvent(event) {
```
novemberborn, line 9: How about passing an async generator?

```suggestion
	async observeRun(runs) {
	  for await (const event of run.events) {
```

novemberborn:
I agree that this is the easiest approach that can ship. I don't want to provide any SemVer guarantees to the internal event objects though. This should be an experiment requiring opt-in through the configuration (`nonSemVerExperiments: { observeRunsFromConfig: true } }` if I recall the syntax).



> We've been using a similar approach to this internally on a patched version of AVA for the last few weeks to pipe test runs to Datadog and it's been working great.
> 
> Alternatives:
> 
> * [use a Shared Worker](https://github.com/avajs/ava/discussions/3004): I think the solution here is much simpler than introducing a third type of worker (Shared Workers, Test Workers, and Suite(?) Workers) and covers most use cases I can think of.
> * use [Diagnostic Channels](https://nodejs.org/api/diagnostics_channel.html): I'm not very familiar with this built-in, but looks like it could help reduce the performance impact of this?
> 
> This is very much still a draft, just wanted to get some initial feedback on whether this solution is ok before I continue down the path. Still need to:
> 
> * add TS definitions for internal events (maybe exposed under `import type {AVAEvent} from "ava/internal"` or something?)
> * add additional tests
> * ensure errors in the user's event handler don't affect AVA

Shared workers survive across multiple test runs (when in watch mode) and so would be appropriate. It's just awkward that you have to launch them from a test worker and they could integrate better with the concept of a test run.

test/internal-events/fixtures/ava.config.js
codetheweb: yeah that makes sense

lib/cli.js
codetheweb: 84dba1d7

entrypoints/internal.d.ts
```diff
@@ -0,0 +1,25 @@
...
+		matching: boolean;
+		previousFailures: number;
+		runOnlyExclusive: boolean;
+		firstRun: boolean;
+	};
+};
+
+export type {StateChangeEvent} from '../types/state-change-events.d.cts';
+
+export type Run = {
```
codetheweb, line 23: open to a better name/export structure for this type

lib/api-event-iterator.js
```diff
@@ -0,0 +1,17 @@
+import {on} from 'node:events';
+
+export async function * asyncEventIteratorFromApi(api) {
+	for await (const [plan] of on(api, 'run')) {
```
codetheweb, line 4: is there a scenario where `api` will emit `run` multiple times? (other than watch mode maybe?)

codetheweb (commented):
apologies for the late replies!

happy to add more tests, but not sure what to look for

novemberborn:
Sorry @codetheweb I've had very little time for open source lately.

entrypoints/internal.d.ts
```diff
@@ -0,0 +1,25 @@
...
+		bailWithoutReporting: boolean;
+		debug: boolean;
+		failFastEnabled: boolean;
+		filePathPrefix: string;
+		files: string[];
+		matching: boolean;
+		previousFailures: number;
+		runOnlyExclusive: boolean;
+		firstRun: boolean;
+	};
```
novemberborn, line 18: Do you think this is useful? I wonder if we should just emit the state change events? And name those `Event`.

entrypoints/internal.d.ts
novemberborn: `ObservedRun`?

lib/api-event-iterator.js
novemberborn: No, just in watch mode. May be helpful for the callback function to know that watch mode is active though.

novemberborn (commented):
This is looking pretty good, I've added some suggestions. A ton of care went into typing the state change events!

When it comes to additional testing, the good news is that for these "experimental" features the bar is a little lower, so this is pretty close to done IMHO.

entrypoints/internal.d.ts
codetheweb: sure, I like that

entrypoints/internal.cjs
```diff
@@ -0,0 +1 @@
+module.exports = {};
```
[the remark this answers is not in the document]
codetheweb: sure, that seems fine to me

entrypoints/internal.d.ts
codetheweb: b7c80536d32ea4aea77c674d27ed8c46d7f9b63c

entrypoints/internal.cjs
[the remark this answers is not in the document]
codetheweb: 1d4de13885cc8e8ad97a9ab79fcad75a380116e5

entrypoints/internal.mjs
```diff
@@ -0,0 +1 @@
+export {};
```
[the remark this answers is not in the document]
codetheweb: 1d4de13885cc8e8ad97a9ab79fcad75a380116e5

lib/api-event-iterator.js
codetheweb: 👍 might be good to add as a future enhancement

lib/api-event-iterator.js
```diff
@@ -0,0 +1,17 @@
...
+		yield {
+			type: 'run',
+			plan,
+		};
+
+		for await (const [stateChange] of on(plan.status, 'stateChange')) {
```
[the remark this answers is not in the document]
codetheweb: fixed in e15de695 I think

entrypoints/internal.d.ts
codetheweb: f10831b5

lib/api-event-iterator.js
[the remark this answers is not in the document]
codetheweb: I guess I misunderstood, I understand now what you meant and fdca2edc demonstrates the issue.

I'm not super familiar with async generators/emittery and running into issues trying to fix it. I think I effectively have to race the abort signal vs. the next `run` event from the `api` class?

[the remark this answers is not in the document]
novemberborn: Turns out there's an `{type: 'end'}` event that is emitted, so it's a bit simpler. There's also an `'interrupt'` which tends to end things.

```js
		for await (const [stateChange] of on(plan.status, 'stateChange')) {
			yield stateChange;
			if (stateChange.type === 'end' || stateChange.type === 'interrupt') {
				break;
			}
```

novemberborn (commented):
Looking good, bar the failing tests :)

lib/api-event-iterator.js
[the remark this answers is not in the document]
codetheweb: thanks! I tried that before, but wasn't quite working because of some emittery quirks. Working now.

types/state-change-events.d.cts
```diff
@@ -0,0 +1,144 @@
...
+	type: 'aggregate';
+	errors: SerializedError[];
+};
+
+type NativeSerializedError = SerializedErrorBase & {
+	type: 'native';
+	source: ErrorSource | undefined;
+};
+
+// eslint-disable-next-line @typescript-eslint/naming-convention
```
codetheweb, line 25: a new rule started complaining about this, ava is always stylized AVA so I think it's ok to disable here?

novemberborn: I'd like to stick with the rule here. Will push up a commit.

novemberborn (approved):

PR #3260, opened 2023-11-12
Don't force-exit after tests have completed

### LINKED ISSUE #1718: Wait for event loop to drain before exiting workers?

novemberborn:
Fixes #1718.

With worker threads, this seems to stop AVA from detecting that the thread has exited, causing a hang.

Also remove flush logic implemented in #1722. Let's hope that current Node.js versions are better at flushing IPC before exiting.

* [ ] ~To do: add a test~ Change is mainly visible in reporters which don't have great tests


Raynos:
This is a non trivial breaking change. We relied heavily on the feature of ava where we didnt have to tear down any database connections etc and it would kill the process when the tests are done.

novemberborn:
Unfortunately it caused hard to understand errors in Node where it wouldn't exit at all.

Can you close these in an after hook? If that is problematic due to the restrictions on test declarations we could come up with another mechanism, say an event when it's safe to force the exit?

Kikobeats:
I'm facing with this too – It seems to me a mechanism should be provided in order to be possible to achieve this in a explicit way

https://github.com/avajs/ava/discussions/3259#discussioncomment-7799985

Raynos:
> Unfortunately it caused hard to understand errors in Node where it wouldn't exit at all.

I fully understand that is the case, this is why i've written or used other test frameworks that don't autoclose.

Going from autoclose to not autoclose is such a large breaking change you might as well switch test frameworks.

In fact it's probably cheaper and quicker for me to port the tests to `node:test` then figure out auto closing in the ava test suite.

novemberborn:
@Raynos @Kikobeats I think a `import { registerCompleteHandler } from 'ava'` which sets up a function to be called when AVA is done would be a decent solution. You could then `registerCompleteHandler(() => process.exit())` and AVA will just assume the test worker exited on its own. This could be in a helper file that is loaded.

Other use cases are as a cleanup routine that sits outside of AVA's test & hook logic. I'm leaning to a solution like this over configuration because configuration implies we can handle edge cases in Node.js and we can't.

novemberborn:
https://github.com/avajs/ava/pull/3283

Issue #3201, opened 2023-05-21
Make assertions throw


sindresorhus:
Currently, the AVA assertions don't throw. This has a lot of downsides:

- It's surprising behavior for new users, as other test frameworks don't do it like this. (Relevant: https://github.com/avajs/ava/issues/2624)
- It's surprising behavior for existing users too (I get surprised by it sometimes).
- It means that you have to guard anything coming after an assertion as the execution of the test would continue even after the assertion fails. It simply complicates writing tests.
- It means AVA cannot easily support third-party assertion frameworks.

Lets make them throw again. We can start by making it opt-in and then change the default in the next major version.

---

Related: https://github.com/avajs/ava/issues/1485

sindresorhus:
I would make the breaking change now and release a new major AVA version that also targets Node.js 16. I don't think the work to make it opt-in first and duplicate a lot of code is worth it. But it's up to @novemberborn.

novemberborn:
There’s a couple things around errors I want to get into AVA 6, and actually I’d then drop Node.js 16 support before release.

If this lands in time we can include it without a config option.

Implementation wise I think failed assertions should throw a `TestFailure` error. If it is caught the test should still fail. We don’t need to expose that error constructor, having a name field should suffice. For the first version let’s not expose any details about the failed assertion (hence not throwing an `AssertionError` with the implication you can catch it and pass the test).

Tests already fail if an error is thrown. I think the twist is that maybe we record the `TestFailure` errors that were thrown during execution and if we see one that did not come from a failed assertion that is a special reason to fail the test? We want people to use `t.fail()’ not these errors. But I suppose that could come later.

ZachHaber:
@novemberborn, I found out something that is problematic: Assertions in typescript require explicit type annotations for all parts of the call: https://github.com/microsoft/TypeScript/issues/45385, https://github.com/microsoft/TypeScript/pull/33622
This means that in order to setup type assertions with the existing API, users will need to explicitly annotate the ExecutionContext in every test.
```ts
test('falsy', (t)=>t.assert(1)); // Typescript Error
test('falsy', (t: ExecutionContext)=>t.assert(1)); // works correctly
```
The reason typescript complains is because `t` is technically inferred from the type definition of `test` and `ImplementationFn`, so there's no way to get it to work without a bunch of pointless boilerplate in TS, and a ton of JSDocs in JS. Which would probably make the current approach using typeguards more approachable in the JS side.

[...]

PR #3267, opened 2023-12-04
Update 03-assertions.md because of issue #2624

ChopCodes:
Adding text in assertion planning in order to inform about ava assertion not stopping even when they fail linked to issue 
#2624 Document that awa assertions don't throw |

I also have 3 of my friends using this tool, I also do. I was thinking why does it not stop and found an issue already. This minor change **will help the users very much**


novemberborn:
Thanks for the PR @ChopCodes!

This paragraph is about assertion planning, so it'd be the wrong place to add this clarification.

However, the behavior has changed in `main`, see this PR: https://github.com/avajs/ava/issues/3201. Which incidentally also adds the clarification discussed in #2624.

For next time, it can be useful to check when documentation was last changed, [through the commit history](https://github.com/avajs/ava/commits/main/docs/03-assertions.md). That would have made it apparent #2624 can be closed.

PR #3292, opened 2024-01-20
Fix import typo

codetheweb:
We finally started building on top of the API [I added last year](https://github.com/avajs/ava/pull/3247) and discovered a small typo. The `ava/internal` type import was failing until I made this change.


novemberborn:
Thanks for the fix! Good timing too, am about to do a release.

PR #3296, opened 2024-01-28
Update TypeScript recipe for loader changes

tommy-mitchell:
Loaders must be added via the new `--import` flag (previously, `--loader` was used). The updated recipe recommends `tsimp`[^1], per https://github.com/avajs/ava/discussions/3213#discussioncomment-8085427.

I added a table of supported Node.js versions for the `--import` flag, but they should probably be double checked. I also removed the section about mocking, since `tsimp` seems to work without issue.

[^1]: [npm](https://www.npmjs.com/package/tsimp), [GitHub](https://github.com/tapjs/tsimp)

### DIFF
--- a/docs/recipes/typescript.md
+++ b/docs/recipes/typescript.md
@@ -17,12 +17,21 @@ Broadly speaking, there are two ways to run tests written in TypeScript:
 
 **You can use loaders, but you're largely on your own. [Please post questions to our Discussions forum if you're stuck](https://github.com/avajs/ava/discussions/categories/q-a).**
 
+> [!NOTE]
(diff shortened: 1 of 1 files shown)


novemberborn:
Looks great, thanks for the compatibility table. Which I haven't checked, so PR welcome in case it's wrong 😄

tommy-mitchell:
@novemberborn seems like maybe it should be `v18.19.0`, but I’m not sure. The changelog mentions the ESM loader changes for that version, but says `--import` was added in `v18.18.0`:

https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V18.md#18.19.0

Issue #3306, opened 2024-02-23
registerCompletionHandler can't be used in CommonJS project


sosnovsky:
Please provide details about:

* What you're trying to do
I tried to use `registerCompletionHandler` in CommonJS project (https://github.com/FlowCrypt/flowcrypt-browser)

* What happened
I wasn't able to import `registerCompletionHandler`, as it's not added to `main.d.cts`

* What you expected to happen
I expected `registerCompletionHandler` to be available for import

novemberborn:
This is intentional. Mostly to avoid polluting the default export in CJS, which is shared with ESM. Also because usage ought to be rare enough that you can work around it in CJS using dynamic import.

acro5piano:
> you can work around it in CJS using dynamic import

This works for me:

```ts
import('ava').then(({ registerCompletionHandler }) => {
  registerCompletionHandler(() => {
    process.exit(0)
  })
})
```

PR #3307, opened 2024-02-23
Add registerCompletionHandler to main.d.cts

sosnovsky:
In 6.1.1 release it's not possible to use `registerCompletionHandler` in CommonJS project.
This PR adds `registerCompletionHandler` to `main.d.cts`, so it can be imported in CommonJS

Fixes https://github.com/avajs/ava/issues/3306.


novemberborn:
Thank you for the thorough PR @sosnovsky.

The lack of CJS here is intentional. Mostly to avoid polluting the default export, which is shared with ESM. Also because usage ought to be rare enough that you can work around it in CJS using dynamic import.

PR #3313, opened 2024-02-27
Update t.throws to accept falsy non-Error thrown values when expectation.any is true

### LINKED ISSUE #3312: `t.throws` doesn't recognize falsy thrown values
[`t.throws`](https://github.com/avajs/ava/blob/main/docs/03-assertions.md#throwsfn-expectation-message) documents that an expectation with property "any" set to `true` should result in a passing assertion when the function throws a non-error value, which doesn't work when the thrown value is falsy.

Concretely, the following test should pass:
```js
test("throws(throwsFalse)", t => {
  t.throws(() => { throw false; }, { any: true });
});
```
but instead fails against ava version 6.1.1:
```console
   2: test("throws(throwsFalse)", t => {                      
   3:   t.throws(() => { throw false; }, { any: true });
   4: });                                               

  Function returned:

  undefined
```

gibson042:
Fixes #3312

### DIFF
--- a/lib/assert.js
+++ b/lib/assert.js
@@ -417,6 +417,7 @@ export class Assertions {
 			}
 
 			let retval;
+			let threw = false;
 			let actual = null;
 			try {
 				retval = fn();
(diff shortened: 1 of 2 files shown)


novemberborn:
Nice find, thanks @gibson042!

PR #3318, opened 2024-03-07
Update readme.md

bn-l:
It has been frustrating getting typescript to work with AVA nicely. Googling information on this is a lost cause these days with the quality of the results. I have spent a lot of time and it is very frustrating to decipher just how to best set up AVA in general, today, 08/03/2024 using modern javascript (typescript). 

I think this warning would also prevent new issues and discussions. It really should be the first thing people see when going to the github.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -2,6 +2,8 @@
 
 ---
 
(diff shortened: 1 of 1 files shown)


novemberborn (commented):
I don’t think this is so important it needs to be mentioned before the logo and introduction. Is there another place a bit further down where we can more proactively direct people to the TypeScript recipe?

novemberborn:

PR #3319, opened 2024-03-07
Update code-coverage.md

bn-l:
Seeking to prevent endless frustration and time spent not knowing this very, **very** key detail.


novemberborn:
Could you add this as a caveat to the custom loader section instead?

novemberborn:

PR #3321, opened 2024-04-06
docs: add `serial` to config options

turadg:
all CLI options can be overridden but this one was omitted https://github.com/avajs/ava/blob/01ec2804ab9db0ab3ef11e3b5b0c5697d68f8bc5/lib/cli.js#L58-L63

I confirmed in code that it is observed
https://github.com/avajs/ava/blob/01ec2804ab9db0ab3ef11e3b5b0c5697d68f8bc5/lib/cli.js#L438

It wasn't clear to me what order to use so I put it before `tap` as it is in `cli.js`.

I'd suggest partitioning the options list into ones that are available in CLI and aren't, but I didn't know if that would be welcomed


turadg:
ping @novemberborn

novemberborn:
Thanks @turadg !

Re partitioning, the CLI flags are already documented elsewhere right? Would it add much value to split the option list?

PR #3333, opened 2024-06-25
ava+testcontainers mini guide

sombriks:
i wrote a small blog post about the topic recently and ava performs so well with testcontainers that i thought it would help more people doing similar setup.

please comment and let me know if it can be improved.


sindresorhus:
I think this is better maintained outside of this repo. None of us use testcontainers, so we are not the best ones to maintain such a recipe. We could potentially link to a blog post / example repo from the docs.

sombriks:
Hello @sindresorhus, that sounds nice, i'll close this one and do another one with a simple link to the article sampling the ava/testcontainers duo.

PR #3341, opened 2024-08-30
doc(watch-mode): Document file watcher crash

adamdicarlo0:
Only when writing this up did I realize that a newer version of Node.js fixes the issue.

Not sure if you want this added to the docs, but hopefully, if nothing else, this will help someone figure out the issue faster -- it took me a while to hunt down!

**Update:** Perhaps this is the Node.js bug causing the issue I encountered: https://github.com/nodejs/node/issues/52018


novemberborn:
Thanks for the write-up, glad it's fixed in newer Node.js versions. As a general rule I don't think the documentation needs to caveat for Node.js bugs.

PR #3345, opened 2024-10-26
Link to CLI options to clarify how to run tests in separate processes

thoughtsunificator:
Add link to detail how you can fall back running in separate process.


thoughtsunificator:
Not sure if running the entire CI for this is necessary and/or should  be required by the way.

novemberborn:
Thanks @thoughtsunificator!

PR #3355, opened 2024-11-23
Adds '--clear-screen' option to clear screen between watch runs

colinta:
This is very slapped together, but I wanted it for myself and figured I'd start the conversation. If you are even open to mergin it let me know of any changes.

###### Usage:
```
npm test ava --watch --clear-scren 
```

###### What does this PR change?
Between test runs, it sends `this.lineWriter.write(ansiEscapes.clearTerminal)`, which clears the screen and scrollback buffer on all the terminal apps _I_ have handy.

I [promoted this same behaviour in TypeScript](https://github.com/microsoft/TypeScript/pull/57701), I guess it's my hill to die on 🤣 (in that case, `--watch` was already clearing the screen, this just changed/improved the behaviour).

Jest also clears the screen when using `--watch`. I prefer it, FWIW, but I've read the argument for keeping all those logs... I'm not convinced, but hey that's just me. I'd be happy w/ `--clear-screen` though it isn't exactly very discoverable atm.

WDYT?

### DIFF
--- a/lib/reporters/default.js
+++ b/lib/reporters/default.js
@@ -3,6 +3,7 @@ import path from 'node:path';
 import stream from 'node:stream';
 import {fileURLToPath} from 'node:url';
 
+import ansiEscapes from 'ansi-escapes';
 import figures from 'figures';
 import indentString from 'indent-string';
 import plur from 'plur';
@@ -73,12 +74,14 @@ export default class Reporter {
 		stdStream,
 		projectDir,
 		watching,
+		clearScreen,
 		durationThreshold,
 	}) {
 		this.extensions = extensions;
 		this.reportStream = reportStream;
 		this.stdStream = stdStream;
 		this.watching = watching;
+		this.clearScreen = clearScreen;
 		this.relativeFile = file => {
 			if (file.startsWith('file://')) {
 				file = fileURLToPath(file);
@@ -128,6 +131,10 @@ export default class Reporter {
 	}
 
 	startRun(plan) {
+		if (this.clearScreen) {
+			this.lineWriter.write(ansiEscapes.clearTerminal);
+		}
+
 		if (plan.bailWithoutReporting) {
 			return;
 		}
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -91,6 +91,12 @@ const FLAGS = {
 		description: 'Re-run tests when files change',
 		type: 'boolean',
 	},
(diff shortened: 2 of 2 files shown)


colinta:
When I first wrote this, I used `process.stderr.write('\x1b[2J\x1b[H\x1b[3J');`, but that felt wrong given the use of `this.lineWriter.write` so I switched to that... but now it doesn't work, so I switched it back to `process.stderr.write`. 🤷‍♂️

sindresorhus:
I have no opinion whether this should be added, but the flag should be `--clear-screen`.

colinta:
I'm not _personally_ a fan of the more verbose `--clear-screen` – [I chose `cls` for what I thought were was a reason](https://en.wikipedia.org/wiki/CLS_(command)) – but I won't quibble. 🙂

lib/reporters/default.js
```diff
@@ -128,6 +131,10 @@ export default class Reporter {
 	}
 
 	startRun(plan) {
+		if (this.clearScreen) {
+			this.lineWriter.write(ansiEscapes.clearTerminal);
+		}
```
novemberborn, line 136: Maybe move this down into the `if (this.watching && !plan.firstRun)` line below? It better connects `clearScreen` with watch mode plus maybe there's no need to clean the terminal on the first run? Or does that get weird?

novemberborn (commented):
Yea I don't mind this. Our long-form flags are not abbreviated so 👍 on `--clear-screen`.

Do we want to keep this as a CLI flag only or also support it in the configuration?

Testing wise, this could be a permutation of https://github.com/avajs/ava/blob/024de32a15aba9fac4002aa8d3c2c076955f44f4/test-tap/reporters/default.js#L45 which should include the ansi code in the generated reporter logs. Let me know if you need a pointer on how to regenerate those, it's far from ideal.

lib/reporters/default.js
colinta: I had it somewhere down there to start, but then it occurred to me that `--clear-screen` *without* `--watch` had no effect at all, which simply felt "unhandled". Why *not* have it clear the screen? 🤔 

Since then I've been actually using it without the watch flag just as often as with it.

colinta:
Thanks for the pointers, I'll try to get those tests updated, and yes having it config makes sense 👍

novemberborn:

PR #3362, opened 2024-12-24
Enable to type `t.context` by extending module interface

acro5piano:
Thank you for the great tool.

## Summary

This Pull Request introduces a new **AvaContext** interface that can be augmented by library users. By allowing developers to declare custom fields in their test context, this change provides greater flexibility and type safety when writing tests in TypeScript.

## Details of Changes

- Added AvaContext Interface
   - Introduced a new `AvaContext` interface
   - Updated `TestFn` to default its generic `Context` to `AvaContext`
- Updated corresponding docs

Library users can now directly extend the `AvaContext` interface to add custom properties in tests, which could be impressive when using shared context:

```ts
declare module 'ava' {
  interface AvaContext {
    foo: string;
  }
}
````

## Difference between `TestFn`

There is already a guide to type `t.context`:

```ts
const test = anyTest as TestFn<{foo: string}>;
```

This PR provides another method to type the context without creating a new variable, which make users ease to use context values.

### DIFF
--- a/docs/recipes/typescript.md
+++ b/docs/recipes/typescript.md
@@ -151,6 +151,20 @@ test('an actual test', t => {
 });
 ```
 
+Alternatively, you can extend the type globally:
+
+```ts
+declare module 'ava' {
+    interface AvaContext {
+        foo: string;
+    }
+}
+
+test.beforeEach(t => {
+	t.context = {foo: 'bar'};
+});
+```
+
 Note that, despite the type cast above, when executing `t.context` is an empty object unless it's assigned.
 
 ## Typing `throws` assertions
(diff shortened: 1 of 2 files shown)


novemberborn:
@sindresorhus what are your thoughts on using extensible interfaces like this?

novemberborn:
I've noticed that XO now recommends using types, not interfaces. I also don't like how this would apply across your project, yet it's not unusual to have different contexts in different files.

Closing for now, perhaps this is better as a Discussion first.

acro5piano:
@novemberborn  Thank you for your review!

> I've noticed that XO now recommends using types, not interfaces.

Yes, I saw the lint errors. Note that this feature cannot be implemented by `type`, because `type` is not extendable while `interface` can be extended.

> I also don't like how this would apply across your project, yet it's not unusual to have different contexts in different files.

That's a fair reason to reject this PR, still users can choose explicit typing the context if they want.

> Closing for now, perhaps this is better as a Discussion first.

Agree! Thanks for considering it.

PR #3364, opened 2024-12-29
Replace arrgv with args-tokenizer

TrySound:
Recently published args-tokenizer package. We use it at Webstudio for parsing cURL on paste. Used at
https://github.com/antfu-collective/bumpp.

There are a few benefits
- published with type definitions
- parser is forgiving to unclosed quotes which can be more user friendly
- actively supported, arrgv last release was 5 years ago

The package also is esm only but used only in CLI so should not be a problem with commonjs. Though I can add cjs support if necessary.

https://github.com/TrySound/args-tokenizer


sindresorhus:
I don't think silently forgiving invalid input is an improvement.

TrySound:
Will "strict" option work for you? I can add it

TrySound:
Upgraded to new strict by default version and reverted test changes

novemberborn:
@TrySound apologies for ignoring this for so long. What would give more confidence is if you incorporate `arrgv'`s test cases, that way it's easier to treat this as a like-for-like substitution. Shipping a 1.0 would also be great.

novemberborn:
> @TrySound apologies for ignoring this for so long. What would give more confidence is if you incorporate `arrgv`s test cases, that way it's easier to treat this as a like-for-like substitution. Shipping a 1.0 would also be great.

To follow up on this, you're asking to replace a dependency that as far as I know hasn't caused problems with a modern version you wrote. That's great! But the less work for me to verify the behavior is exactly the same, the better.

TrySound:
Of course! Thanks for your response. Will update PR when release 1.0

kensmajesty-ship-it:
{
	"name": "awesome-package",
	"scripts": {
		"test": "ava"
	},
	"devDependencies": {
		"ava": "^1.0.0"
	}
}

Issue #3367, opened 2025-02-11
DOMException is not recognized as an Error by throws() and throwsAsync() in AVA v6


sholladay:
In the Ky tests, we have [a simple assertion](https://github.com/sindresorhus/ky/blob/4a427011ad7b4ab98bb5f02ecaa5375fba5addca/test/main.ts#L710) that a [Node.js `DOMException`](https://nodejs.org/api/globals.html#domexception) is thrown when a request is aborted. In AVA v5, this assertion works as expected. However, beginning with AVA v6, `DOMException` seems to no longer be detected as an error, even though `error instanceOf Error` is `true` for it.

Minimal reproduction with default config:

```js
import test from 'ava';

test('DOMException should be considered an error', async t => {
	await t.throwsAsync(Promise.reject(new DOMException('an error')));
});
```

The above test passes on AVA v5 but fails on v6. This applies to both `t.throws()` and `t.throwsAsync()`.

As an aside, the docs are not very clear on what is considered "an error". I always assumed it meant the object had to be an `instanceof Error`, but at least in AVA v6, this seems to not be the case. Is it because `error.constructor.name` doesn't include "Error"? Would be nice to have some clarification.

Prasanth2706:
`import test from 'ava';

test('DOMException should be considered an error', async t => {
	await t.throwsAsync(async () => {
		throw new DOMException('an error');
	}, {instanceOf: DOMException, message: 'an error'});
});
`

sholladay:
We don't use the assertion `instanceOf` option because the error constructor actually depends upon the environment where the code is run. Sometimes it's just a plain `Error`. So we run a separate `t.true()` assertion afterwards with an array of constructor names to check that.

novemberborn:
See https://github.com/avajs/ava/blob/main/docs/08-common-pitfalls.md#error-edge-cases (which isn't linked to from the assertions documentation, so that's not helpful).

> The throws() and throwsAsync() assertions use the Node.js built-in [isNativeError()](https://nodejs.org/api/util.html#utiltypesisnativeerrorvalue) to determine whether something is an error. This only recognizes actual instances of Error (and subclasses).
>
>Note that the following is not a native error:
>
> ```
> const error = Object.create(Error.prototype);
> ```
>This can be surprising, since error instanceof Error returns true. You can set any: true in the expectations to handle these values:
>
> ```
> const error = Object.create(Error.prototype);
> t.throws(() => { throw error }, {any: true});```

How is `DOMException` created?

[...]

PR #3370, opened 2025-03-15
Fix: Handle DOMException error in Ava tests

Prasanth2706:
This PR fixes an issue where DOMException was not correctly recognized as an error in the test.

- Updated the test case to ensure DOMException is properly caught.
- Verified the fix by running npm test and confirming all tests pass.
- No breaking changes introduced.


sholladay:
I was hoping this was a fix for #3367. But this PR doesn't actually change any behavior, right?

novemberborn:
@Prasanth2706 I don't think you pushed up all the changes, there's some submodule showing in the PR.

novemberborn:

PR #3374, opened 2025-04-03
Do not count writes to stdout/stderr as non-idling activity for timeouts

### LINKED ISSUE #3373: Timeout does not function if test is writing to the console
Please provide details about:

* What you're trying to do
I have a test which is writing debug logs to the console but not making any asserts. The timeout on that test is triggering that the test failed, but not actually stopping the test run so ava never exits.

> Timeouts can also be set individually for each test. These timeouts are reset each time an assertion is made. The test fails if it takes more than ms for an assertion to be made or the test to complete.

* What happened
```
$ npm test will-not-stop.js

> test
> ava will-not-stop.js


0s
1s
2s
3s
4s
  ✘ [fail]: will not stop stop after 5s
5s
6s
7s
...
62s
^C  
  ✘ Exiting due to SIGINT

  Failed to exit when running will-not-stop.js

  ─

  will not stop

  Error: stop after 5s

  Error: stop after 5s
      at Timeout.<anonymous> (file:///Users/matthew/spikes/repro-ava-timeout/node_modules/ava/lib/test.js:439:24)
      at listOnTimeout (node:internal/timers:594:17)
      at process.processTimers (node:internal/timers:529:7)

  ─

  1 test failed
```

* What you expected to happen
I expected the test to exit after it was marked as having failed, as in this example:

```
$ npm test will-stop.js

> test
> ava will-stop.js


  ✘ [fail]: will stop stop after 5s
  
  ✘ Timed out while running tests

  Failed to exit when running will-stop.js

  ─

  will stop

  Error: stop after 5s

  Error: stop after 5s
      at Timeout.<anonymous> (file:///Users/matthew/spikes/repro-ava-timeout/node_modules/ava/lib/test.js:439:24)
      at listOnTimeout (node:internal/timers:594:17)
      at process.processTimers (node:internal/timers:529:7)

  ─

  1 test failed
```

Sample Code:
1. Run `npm init ava` to create a skeleton project (done on 04/03/2025, ava v6.2.0)
2. Add will-not-stop.js
```javascript
import test from "ava";

test("will not stop", async (t) => {
  t.timeout(5_000, "stop after 5s");
  for (let i = 0; ; ++i) {
    console.log(`${i}s`);
    await new Promise((resolve) => setTimeout(resolve, 1000));
  }
});

```
3. Add will-stop.js (same as will-not-stop but with the write call commented out)
```javascript
import test from "ava";

test("will not stop", async (t) => {
  t.timeout(5_000, "stop after 5s");
  for (let i = 0; ; ++i) {
    // console.log(`${i}s`);
    await new Promise((resolve) => setTimeout(resolve, 1000));
  }
});

```

mdouglass:
fixes #3373


lib/api.js
```diff
@@ -218,7 +218,7 @@ export default class Api extends Emittery {
 			}
 
 			runStatus.on('stateChange', record => {
-				if (record.testFile && !timedOutWorkerFiles.has(record.testFile)) {
+				if (record.testFile && !timedOutWorkerFiles.has(record.testFile) && record.type !== 'worker-stderr' && record.type !== 'worker-stdout') {
```
mdouglass, line 221: I'd personally prefer to make this opt-in so only assertions/test state changes reset the timeout (so we hew closer to the documentation). 
But:
a) I don't know the complete list of record type's here that would apply
b) it's a bigger change from current behavior than I'm comfortable making without feedback

novemberborn: I think this change is correct. The timeout here is a guard against test workers that are no longer making progress. It's different from `t.timeout()` which limits how long an individual test can take.

Essentially with this change, we no longer count stdout/stderr as "making progress".

test/test-timeouts/test.js
```diff
@@ -14,3 +14,9 @@ test('timeout messages must be strings', async t => {
 	t.snapshot(error.message, 'error message');
 	t.snapshot(error.formattedDetails, 'formatted details');
 });
+
+test('timeouts are not blocked by console output', async t => {
+	const result = await t.throwsAsync(fixture(['console-output.js']));
+	const error = result.stats.getError(result.stats.failed[0]);
+	t.snapshot(error.message, 'timeout despite console output');
+});
```
novemberborn, line 22: This test is in the wrong place. `test-timeouts` refers to `t.timeout()` but this is modifying the overall timeout behavior. A new `idle-timeout/` folder would be more appropriate.

(Can't recall whether there's existing tests under the old `test-tap` but I'm happy not to add there.)

novemberborn (commented):
This fix is correct, but for a different reason than you might expect. The timeout in the `api.js` file is best thought of as an "idle guard". I agree that stdout/stderr shouldn't count as non-idle activity.

In #3373 the test worker isn't exiting because it's still waiting on a timer. And the idle guard doesn't force it to exit. With this change it will, though only 10 seconds after the test itself has timed out.

test/test-timeouts/test.js
mdouglass: Went ahead and moved the test, thanks.

mdouglass:
> This fix is correct, but for a different reason than you might expect. The timeout in the `api.js` file is best thought of as an "idle guard". I agree that stdout/stderr shouldn't count as non-idle activity.
> 
> In #3373 the test worker isn't exiting because it's still waiting on a timer. And the idle guard doesn't force it to exit. With this change it will, though only 10 seconds after the test itself has timed out.

Thanks for explaining that, some of the behavior I saw while fixing this makes more sense now that I see the test timeout in Test#timeout vs. the worker timeout here.

novemberborn:
Thanks @mdouglass, this looks good. I've made some smaller tweaks, mostly in the test itself. Your version took 10.8 seconds to execute in CI which is a little unnecessary 😄

novemberborn (approved):

Issue #3390, opened 2025-07-09
AVA finishes prematurely on Windows, failing with remaining tests


matz3:
In our project https://github.com/UI5/linter we are observing issues in our CI pipeline on Windows for quite some time now.
I currently suspect this to be caused by a timing issue when running tests in AVA.

First of all, I would like to check with you, whether I understood the concept of how runners are communicating with the main thread.

Runners are sending messages about tests and whether they finish via an async communication port.
When the runner is done, it fires a final "finish" message, and the process eventually exits.

The main thread receives those messages and records the status of tests, e.g. whether they are finished or not.

What we could observe in the CI and also reproduce in a VM is that AVA finishes before the main thread receives all messages from the runner. This leads to the error `× N tests remaining in test.ts`, as AVA thinks the runner exited without running all the tests. But in fact, only the information that the tests ran was not received.

This seems to only appear when `workerThreads` is `false`, which we require in order to use loaders for TypeScript, coverage and mocking.

I am happy to contribute to a solution. I believe there is no mechanism that ensures all messages are fully sent to the main thread before the runner process exits. Maybe another event ping-pong at the end could help to ensure all messages have been received. That is at least the idea a got from debugging and looking at the code, but I'm not fully aware of the whole concept of AVA or whether this is actually already ensured.

Here is a very simple example with which I could reproduce the problem most of the time.

**Config**
```js
export default {
	files: [
		"test.js",
	],
	workerThreads: false,
};
```

**Test**
```js
import test from "ava";
for (let i = 0; i < 50; i++) {
	test("Test " + i, (t) => {
		t.is(true, true);
	});
}
```

**Output**

Note that the test sometimes succeeds and the number of remaining tests differs a lot.

```txt
> ava


  ✔ Test 0
  ✔ Test 1
  ✔ Test 2
  ✔ Test 3
  ✔ Test 4
  ✔ Test 5
  ✘ 44 tests remaining in test.js
  ─

  6 tests passed
```

Exit code: 1

**System Information**

We've encountered this in our project on GitHub Actions with both, `windows-2022` and `windows-2025`.  
I could also reproduce it in a virtual machine with `Microsoft Windows 11 Enterprise` (`Version 10.0.26100 Build 26100`).

- Node.js: `22.17.0`
- npm: `10.9.2`
- AVA: `6.4.0`

novemberborn:
It's odd that occurs with 50 synchronous tests!

[...]

PR #3391, opened 2025-07-10
Keep test worker alive until explicitly freed

matz3:
This change ensures that all messages from the worker are received by the main process before the worker exits.

It solves issues on Windows that are occurring with Node.js v20 and later when `workerThreads: false` is set in the AVA configuration.

Fixes: https://github.com/avajs/ava/issues/3390


novemberborn:
Released as 6.4.1.

novemberborn:
Thanks for the fix!

PR #3437, opened 2026-04-09
Remove CommonJS support

### LINKED ISSUE #3349: Ava test() expression is not callable / has no call signatures
After updating Typescript from 5.5.4 to 5.6.3 i started to have an error 

```
error TS2349: This expression is not callable.
  Type 'typeof import("<full_path_to_project>/node_modules/ava/entrypoints/main")' has no call signatures.
``` 

Version of ava - 6.2.0
Version of Typescript - 5.6.3
Version of tsimp - 2.0.12

ava config

```
export default {
  files: ['tests/**/*.spec.ts'],
  extensions: {
    ts: 'module',
  },
  nodeArguments: ['--import=tsimp/import', '--no-warnings'],
  timeout: '2m',
  serial: true,
  concurrency: 1,
};
```

test example

```
import test from 'ava';
import type { ExecutionContext } from 'ava';

test('example', (t: ExecutionContext) => {
  t.is('8080', '8080');
});
```

 I expect not to have this warning from Typescript when using Ava.
 
 There was similar issue #2539, but it's closed and nothing from comments there helping me (first of all because I can not use require syntax).

sindresorhus:
Fixes #3349

Breaking changes:

- Removes support for `ava.config.cjs` files
- Always uses `import()` to load test files and extra modules through the `require` configuration
- No longer includes `.cjs` extension in the defaults (so will not find `test.cjs` files)
- Removes the `commonjs` value in the `extensions` configuration
- [x] Update provider protocol to `ava-8` to reflect these changes (requires new `@ava/typescript`)


sindresorhus:
Ready

novemberborn:
@sindresorhus looks good! I've tried to summarize the breaking changes in the PR description. Unfortunately this does impact the `@ava/typescript` provider protocol. I can ship a new version there that is compatible with AVA 7 and 8.

I've pushed a few further simplifications. I did add back support for configuring a `cjs` extension, but only with a `true` value. It's weird to specify `module`.

I wonder if, instead of mandating `module` for all other extensions we should just always mandate `true` values when extensions is given in object form. Or remove support for object form entirely. What do you think?

sindresorhus:
> I wonder if, instead of mandating module for all other extensions we should just always mandate true values when extensions is given in object form. Or remove support for object form entirely. What do you think?

I would just remove support for object form.

sindresorhus:
Your changes look good.

novemberborn:
https://github.com/avajs/typescript/pull/63 is green, but I need to do a bunch of config to make it publish, so that's for another evening.

novemberborn (approved):
Just waiting for CI to pass and then this can be merged.

docs/06-configuration.md
```diff
@@ -68,18 +68,18 @@ Provide the `typescript` option (and install [`@ava/typescript`](https://github.
```
xjx62xjv8s-droid, line 68: ر

docs/06-configuration.md
```diff
@@ -42,7 +42,7 @@ Arguments passed to the CLI will always take precedence over the CLI options con
 
 ## Options
```
xjx62xjv8s-droid, line 43: ت

PR #3442, opened 2026-05-09
Add seeded test order randomization

### LINKED ISSUE #595: Randomize test runs

Thoughts?

dicnunz:
Fixes #595.

IssueHunt bounty: https://issuehunt.io/r/avajs/ava/issues/595

## Summary

- Add `--shuffle` to randomize test-file order and concurrent test start order within each file using a generated seed.
- Add `--seed=<number>` to reproduce a randomized order.
- Report the seed in the default and TAP reporters.
- Preserve `.serial` tests in declaration order before concurrent tests.

## Test plan

- `npx test-ava test/randomization/test.js`
- `npx test-ava test/scheduler/test.js test/concurrency/test.js`
- `npx xo` (passes with existing TODO warnings)
- `npx tsc --noEmit`
- `git diff --check`
- `node --check lib/randomization.js && node --check test/randomization/test.js`

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#595: Randomize test runs](https://oss.issuehunt.io/repos/26820798/issues/595)
---
</details>

### DIFF
--- a/docs/05-command-line.md
+++ b/docs/05-command-line.md
@@ -38,6 +38,9 @@ Options:
       --node-arguments     Additional Node.js arguments for launching worker
                            processes (specify as a single string)       [string]
   -s, --serial             Run tests serially                          [boolean]
+      --seed               Randomize test order using a reproducible seed
+                                                                        [number]
+      --shuffle            Randomize test order using a random seed     [boolean]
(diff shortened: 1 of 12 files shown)


sindresorhus:
Closing as this is fully AI-generated spam.

Issue #2501, opened 2020-05-25
Known reporter issues


novemberborn:
This is a list of known problems and desired improvements with our default reporter. Some may be easier to solve than others, but probably we should [clean up our reporter implementation](https://github.com/avajs/ava/issues/2500) first.

* If an exception occurs when loading the test file, before it manages to declare any tests, we should say so. Originally reported in https://github.com/avajs/ava/issues/2290.

* When `--match` is used, if no matching tests were selected, the reporter shouldn't still print other errors that might have occurred. Originally reported in https://github.com/avajs/ava/issues/2191.

* Code excerpts are derived from error stack traces, but we should show the excerpt for the test itself, not wherever the error originated from. Originally reported in https://github.com/avajs/ava/issues/1743, might have already been resolved.

* We should print how many snapshots were updated. Originally reported in https://github.com/avajs/ava/issues/1583.

* Bad usage of the test interface is thrown as an internal error, which looks horrible. Originally reported in https://github.com/avajs/ava/issues/698.

* When `--match` is used, or after a `test.only()` has been encountered, AVA still needs to launch each test file to see if there are matches or exclusive tests. There is no output while this is happening. There should be. Originally reported in https://github.com/avajs/ava/issues/2246.

* We should revisit the color scheme. Originally reported in https://github.com/avajs/ava/issues/413.

* It's hard to distinguish console output between watch runs. Originally reported in https://github.com/avajs/ava/issues/1917.

* We print pending tests when AVA times out. We should print just those tests (and hooks) that have started. Originally reported in https://github.com/avajs/ava/issues/2421.

* `t.log()` output for passing tests is only available in _verbose_ mode. We should always print this. Originally reported in https://github.com/avajs/ava/issues/2201.

* The `-` & `+` diff symbols need an explanation. Originally reported in https://github.com/avajs/ava/issues/1558.

* Even when not in _verbose_ mode, we should print details for all exceptions. Originally reported in https://github.com/avajs/ava/issues/2171.

* When not in _verbose_ mode, console output from the test files messes with the reporter output. Originally reported in https://github.com/avajs/ava/issues/1953.

futpib:
Bounties from #1743, #1917, #1558 and #1953 are now dangling and should probably be moved here somehow.

novemberborn:
@futpib yes, I'm not sure to what extend that's possible. @sindresorhus?

sindresorhus:
[...]

PR #3444, opened 2026-05-11
Improve reporter message for pre-declaration failures

NoeFabris:
## Summary
- report explicitly when a test worker fails before any tests are declared
- cover the worker-failed reporter path with a focused regression test

## Test plan
- npx tap test-tap/reporters/default-worker-failed.js --disable-coverage
- npx xo lib/reporters/default.js test-tap/reporters/default-worker-failed.js

Addresses one reporter item from #2501.


sindresorhus:

PR #3445, opened 2026-05-11
Add cleanup hook API

### LINKED ISSUE #928: Make it easier to write reliable cleanup tasks
See: https://github.com/avajs/ava/pull/918#issuecomment-227016238

`after.always` has some issues when used as a cleanup task. Specifically, it won't run if:

- There are test failures and `--fail-fast` is used.
- There are uncaught exceptions thrown.

I've advocated using `.before` or `.beforeEach` to ensure state is clean before running, but that means state is left on disk after the test run. It's easy enough to get around that:

```js
function cleanup() {
  if (temDirExists()) {
    removeTempDir();
  }
}

test.before(cleanup);
test.after.always(cleanup);
```

Still, it might be nicer if we had a modifier that allowed you to do it a little cleaner:

```js
// runs as before and after
test.cleanup(cleanupFn);

// runs as beforeEach and after (not sure afterEach makes much sense?)
test.cleanupEach(cleanupFn);
```

Or maybe we introduce a `.and` modifier:

```js
test.before.and.after(cleanupFn);
test.beforeEach.and.after(cleanupFn);
test.beforeEach.and.afterEach(cleanupFn);
```

I think the second gives you a little more flexibility and is clearer without reading the docs. The first is probably simpler to implement (though I don't think the second would be very hard)

Agentriel:
Fixes #928.

This adds `test.cleanup()` as a small lifecycle convenience for idempotent cleanup code. The declaration registers the same implementation as:

- a `before` hook, so state left behind by an interrupted or crashed previous run can be removed before tests start
- an `after.always` hook, so state created by the current run is removed after tests finish, including failed runs that still reach AVA's normal hook execution

It also adds `test.serial.cleanup()`, `.cleanup.skip()`, TypeScript declarations, docs, and regression coverage for failure and skipped cleanup behavior.

Checked:
- `npx tap test-tap/hooks.js --disable-coverage`
- `npx tap test-tap/runner.js --disable-coverage`
- `npx xo lib/create-chain.js lib/runner.js test-tap/hooks.js`
- `npm exec -- tsd --files test-types/module/conditional-chains.ts`
- `git diff --check`

### DIFF
--- a/docs/01-writing-tests.md
+++ b/docs/01-writing-tests.md
@@ -178,6 +178,8 @@ AVA lets you register hooks that are run before and after your tests. This allow
 
 `test.beforeEach()` registers a hook to be run before each test in your test file. Similarly `test.afterEach()` registers a hook to be run after each test. Use `test.afterEach.always()` to register an after hook that is called even if other test hooks, or the test itself, fail.
 
+`test.cleanup()` registers a hook to be run once before the first test and again as an `after.always()` hook. Use it to make cleanup idempotent: the first run removes state left behind by a previous interrupted run, and the second run removes state created by this run.
+
 If a test is skipped with the `.skip` modifier, the respective `.beforeEach()`, `.afterEach()` and `.afterEach.always()` hooks are not run. Likewise, if all tests in a test file are skipped `.before()`, `.after()` and `.after.always()` hooks for the file are not run.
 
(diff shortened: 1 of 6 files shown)


Agentriel:
Thanks for taking a look when you have time. GitHub Actions is currently waiting for first-time contributor workflow approval on this PR.

I ran the focused checks locally against the submitted commit:

- `npx tap test-tap/hooks.js --disable-coverage`
- `npx tap test-tap/runner.js --disable-coverage`
- `npx xo lib/create-chain.js lib/runner.js test-tap/hooks.js`
- `npm exec -- tsd --files test-types/module/conditional-chains.ts`
- `git diff --check`

Happy to adjust the API shape or tests if you prefer a different direction for cleanup hooks.

sindresorhus:
We don't accept fully AI-generated PRs. This user has spammed many repos.

PR #3447, opened 2026-05-12
Add cleanup hook helpers

### LINKED ISSUE #928: Make it easier to write reliable cleanup tasks

Cosmhyccc:
Fixes #928

## Summary
- Add `test.cleanup()` and `test.cleanupEach()` hook chains, including `test.serial` variants.
- Expand cleanup hooks into existing lifecycle hooks: `before` + `after.always`, and `beforeEach` + `afterEach.always`.
- Document the helpers, update TypeScript declarations, and cover passing/failing test behavior.

## Verification
- `PATH=/Users/saifali/.cache/codex-runtimes/codex-primary-runtime/dependencies/node/bin:$PATH npx tsc --noEmit`
- `PATH=/Users/saifali/.cache/codex-runtimes/codex-primary-runtime/dependencies/node/bin:$PATH npx tap test-tap/hooks.js`
- CI: `Lint source files` passed after the switch refactor.

### DIFF
--- a/docs/01-writing-tests.md
+++ b/docs/01-writing-tests.md
@@ -178,6 +178,8 @@ AVA lets you register hooks that are run before and after your tests. This allow
 
 `test.beforeEach()` registers a hook to be run before each test in your test file. Similarly `test.afterEach()` registers a hook to be run after each test. Use `test.afterEach.always()` to register an after hook that is called even if other test hooks, or the test itself, fail.
 
+`test.cleanup()` is shorthand for registering the same hook with `test.before()` and `test.after.always()`. Use it for idempotent file-level cleanup that should run before tests start and again once they complete. `test.cleanupEach()` works the same way for each test, combining `test.beforeEach()` with `test.afterEach.always()`.
+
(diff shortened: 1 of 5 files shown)


sindresorhus:

PR #3448, opened 2026-05-12
Add seeded test order randomization

### LINKED ISSUE #595: Randomize test runs

JustABiologist:
Fixes #595.

IssueHunt bounty: https://oss.issuehunt.io/r/avajs/ava/issues/595

## Summary
- Add `--randomize` and `--seed` so AVA can shuffle selected test files and non-serial tests within each file.
- Keep `test.serial()` declarations in source order while randomizing ordinary tests.
- Report the seed in the default and TAP reporters so a randomized order can be reproduced.

## Test plan
- `npx tap test-tap/runner.js`
- `npx tap test-tap/integration/assorted.js`
- `npx xo` (passes with existing TODO warnings)
- `npx tsc --noEmit`
- `git diff --check`

### DIFF
--- a/docs/05-command-line.md
+++ b/docs/05-command-line.md
@@ -37,6 +37,8 @@ Options:
       --no-worker-threads  Don't use worker threads                    [boolean]
       --node-arguments     Additional Node.js arguments for launching worker
                            processes (specify as a single string)       [string]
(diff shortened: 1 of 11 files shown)


sindresorhus:

PR #3450, opened 2026-05-15
Add Node.js 26 support; drop 25

novemberborn:
- Sanitize line numbers for `lib/` files in reporter logs
- Remove v25 snapshots
- Test with Node.js 26
- Update support statement now that odd-numbered Node.js releases will become LTS
- Upgrade to cbor2 for Node.js 26 snapshot compatibility


lib/snapshot-manager.js
```diff
@@ -180,10 +182,36 @@ function sortBlocks(blocksByTitle, blockIndices) {
...
+			.map(([k, v]) => [k, normalizeBeforeEncoding(v)]));
+	}
+
+	return value;
+}
+
+// FIXME: This seems to be a bug in cbor2, ignoreGlobalTags shouldn't break (Uint8)Array encoding.
+const types = new TypeEncoderMap();
+types.registerEncoder(Array, writeArray);
+types.registerEncoder(Uint8Array, writeUint8Array);
```
novemberborn, line 207: @hildjj FYI, something seems wrong here. It's late, so a proper bug report will have to come later sorry 😄  There's a related issue in the decode code path.

But, `cbor2`, yay!

PR #3467, opened 2026-07-11
Add truthy falsy example

Romie321:
This PR adds a small beginner-friendly example demonstrating AVA’s t.truthy() and t.falsy() assertions. The example shows how to check whether values are considered truthy or falsy in JavaScript, helping new users understand AVA’s assertion helpers.


Romie321:
Two CI jobs failed on Node 22 and Node 26. This example doesn’t touch AVA internals, so the failures appear unrelated. Let me know if anything needs to be adjusted.

Romie321:
The code is done and passing on stable Node versions, but the snapshot check fails on the experimental Node v26 Linux runner. Since I am on windows and cannot replicate that environment locally to update the baseline, could a maintainer please run the snapshot update on review? Thanks!

PR #3472, opened 2026-07-20
fix: detect synchronous throws of falsy values in t.throwsAsync()

tsushanth:
## Bug

When the function passed to `t.throwsAsync()` throws a falsy value synchronously — for example `throw 0`, `throw false`, `throw null`, or `throw ""` — the guard that detects the synchronous throw uses a bare truthiness check:

```js
let actual = null;
try {
    retval = thrower();
} catch (error) {
    actual = error;
}

if (actual) { // ← misses falsy thrown values
```

Because `0`, `false`, `null`, and `""` are all falsy, the `if (actual)` branch is skipped. The code falls through to the `isPromise(retval)` check (retval is `undefined` since the function threw), which also fails, so AVA ultimately reports **"Function returned: undefined"** instead of the correct **"Function threw synchronously. Use `t.throws()` instead: <value>"**.

## Fix

Change the sentinel check to `actual !== null`, matching the initial value assigned before the `try/catch`.

```js
if (actual !== null) {
```

This correctly detects any synchronous throw — including falsy thrown values — without affecting the normal case where no exception is thrown (`actual` remains `null`).


tsushanth:
The `Node.js (^26, ubuntu-latest)` failure appears pre-existing and unrelated to this PR — the main branch CI also has intermittent Node 26 failures (run #25970521309). This PR passes on Node 22, 24, and Node 26 on Windows and macOS.

